PR #19, opened 2013-02-04
Substitute macro

BrantKyser:
Add substitue "macro" to allow basic find/replace within typemaps and director descriptors.


ojwb:
@wsfulton This PR seems to have stalled - what are your thoughts in light of the latest comment?

wsfulton:
I'm not keen on adding additional complexity to SWIG with these kinds of special functions as there are already too many obscure extensions to C/C++. Either using the C preprocessor or templates or doing it 'long hand' is required, like in C or C++. Sorry.

PR #68, opened 2013-07-29
Fixed %unrefobject for inherited classes

davidlongest:
Fixes the "unref" feature when used on derived classes with non-public
destructors.
I'm not sure if this is the correct method to be used, but this causes
the swig wrapper to check for a parent that supports the unref feature.


ojwb:
I've been well over a year without a response, so closing.

@davidlongest If you can provide an example (or examples) this fixes, please reopen.

PR #132, opened 2014-02-05
avoid collision with canonical definition of std::less<PyObject *>

bjoernthiel:
working with std::map<PyObject *, ...> I stumbled because of the very special definition of std::less<PyObject *> coming with pycontainer.swg. 
So I suggest to remove it or at least to put it into an extra *.i file.


ojwb:
I think by "exact output" William means on stdout/stderr, not the files produced.  To be more explicit, what's the error from the compiler?

And is this still an issue with a more recent version of SWIG?

bjoernthiel:
examplepython_wrap.cxx(4214) : error C2908: explicit specialization; 'std::less<_Ty>' has already been instantiated with [ _Ty=_object * ]
examplepython_wrap.cxx(4236) : error C2766: explicit specialization; 'std::less<_object*>' has already been defined

using SWIG 3.0.5

kyl27:
Is there any progress on this issue? It appears that specializing std::less is against convention: http://stackoverflow.com/questions/2282349/specialization-of-templateclass-tp-struct-stdless-in-different-namespace
I am having this exact same problem.
Compiler: gcc (GCC) 4.8.2 20140120 (Red Hat 4.8.2-15)
Kernel version: 2.6.32-504.38.1.el6.x86_64
OS version: redhat-release-server-6Server-6.6.0.2.el6.x86_64

wsfulton:
I doubt that changing the `std::less` specialization to `operator<` will make a difference given the error message provided as it looks like it has simply been defined twice.

@kyl27 Can you provide a simple reproducer of the error on your Redhat set up?

PR #136, opened 2014-02-11
Fix the division operator overload in Python 3 generated bindings.

ejulien:
Fix the division operator overload under Python 3.
The complete test suite passes with this changes under Python 2.7.5+.

The test suite fails on Python 3.3 because of compliance issues in the test themselves (eg. print call missing parentheses, except issues, etc...). But the division operator works and the fairly large binding we use still works with no regression in our tests.

### DIFF
--- a/Lib/python/pyopers.swg
+++ b/Lib/python/pyopers.swg
@@ -103,7 +103,7 @@
 %pybinoperator(__neg__,      *::operator-(),		unaryfunc, nb_negative);
 %pybinoperator(__neg__,      *::operator-() const,	unaryfunc, nb_negative);
 %pybinoperator(__mul__,      *::operator*,		binaryfunc, nb_multiply);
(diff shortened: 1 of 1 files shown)


ejulien:
@wsfulton @xantares After digging some more it seems the operator overload test suite was not implemented for Python. I opened #617 as a first step to work toward solving this issue.

Note that I had been bitten by another issue the test suite uncovered immediately: the += operator family is assigning None to the object it is applied on. I was able to workaround this bug by returning *this from the affected operators but it would be nice to fix that as well.

wsfulton:
@ejulien, can this be closed now that #617 has been committed? The testcase for Foo::operator/ you provided on 18 Aug 2014 works now.

ejulien:
@wsfulton Yes, #617 fixes that issue, closing this PR.

PR #155, opened 2014-03-28
Segfault in _zval_dtor_func when PHP built with ZTS

pierre-labastie:
A bunch of test in the PHP testsuite fail with an error message "tsrm_ls undefined", if PHP is compiled with zts enabled.
The proposed change allows those tests to pass.

### DIFF
--- a/Lib/php/utils.i
+++ b/Lib/php/utils.i
@@ -1,4 +1,3 @@
-
 %define CONVERT_BOOL_IN(lvar,t,invar)
   convert_to_boolean_ex(invar);
   lvar = (t) Z_LVAL_PP(invar);
@@ -92,6 +91,7 @@
 %fragment("t_output_helper","header") %{
 static void
 t_output_helper( zval **target, zval *o) {
+  TSRMLS_FETCH();
   if ( (*target)->type == IS_ARRAY ) {
     /* it's already an array, just append */
     add_next_index_zval( *target, o );


ojwb:
@pierre-labastie Did you get a chance to re-test?

ojwb:
It would be better to report those as separate issues, especially as others maintain those parts of SWIG.

pierre-labastie:
Le 11/01/2015 12:20, Olly Betts a écrit :
> It would be better to report those as separate issues, especially as others
> maintain those parts of SWIG.
> 
> 
Understood.

I have reported to swig-devel list. Do you think there is a better place?

Pierre

ojwb:
Opening a ticket is better for reporting a bug.

pierre-labastie:
Le 09/01/2015 01:50, Olly Betts a écrit :
> @pierre-labastie <https://github.com/pierre-labastie> Did you get a chance to
> re-test?

While I was with SWIG-3.0.3, I recompiled php with zts support. PHP version is
5.6.4. gcc/g++ version is 4.9.2, binutils version is 2.24.

Here is what I get:
+++++++++++++++++
$ make check-php-examples
[...]
checking Examples/php/extend
/bin/sh: line 1: 26317 Segmentation fault      php -n -q -d extension_dir=. -d
safe_mode=Off runme.php > /dev/null
../../Makefile:1132: recipe for target 'php_run' failed
make[3]: *** [php_run] Error 139
Makefile:10: recipe for target 'check' failed
make[2]: *** [check] Error 2
Makefile:246: recipe for target 'extend.actionexample' failed
make[1]: *** [extend.actionexample] Error
[..all other tests pass..]
+++++++++++++++++
$ make check-php-test-suite
[...]
checking php testcase director_abstract (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_basic (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_classic (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_enum (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
checking php testcase director_exception (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_finalizer (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_nested (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_protected (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
hecking php testcase director_thread (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_string (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[..all the other tests pass..]
++++++++++++++++

That is, only the "extend" examples, and a few "director" tests in the test
suite fail. A lot of "director" tests pass and all the other tests pass.

Pierre

ojwb:
Thanks for retesting.

Comparing with the list of failing tests from above, these are now passing (so some progress):

 - Examples/funcptr
 - Examples/variables
 - test-suite/autodoc
 - test-suite/char_binary
 - test-suite/char_strings
 - test-suite/funcptr_cpp
 - test-suite/grouping
 - test-suite/template_default
 - test-suite/typedef_funcptr 

Leaving the following failures:

 - Examples/extend
 - test-suite/director_abstract
 - test-suite/director_basic
 - test-suite/director_classic
 - test-suite/director_enum
 - test-suite/director_exception
 - test-suite/director_finalizer
 - test-suite/director_nested
 - test-suite/director_protected
 - test-suite/director_thread
 - test-suite/director_string

I notice that all the tests which still fail make use of subclassing via directors (though some passing tests do too).

Rereading the discussion, it seems the originally proposed patch made all the test pass, or is that no longer true with newer versions of PHP?

While avoiding calling `TSRMLS_FETCH()` is desirable, avoiding segfaults in generated wrapper code clearly trumps that.  I don't have any good ideas as to why the current code isn't working.  Perhaps the value we store in the `swig_zts_ctx` member of class `Director` doesn't remain valid, but that seems very surprising as it belongs to the current thread.  And at least the overhead is only incurred when using a ZTS build of PHP.

pierre-labastie:
Le 11/01/2015 22:04, Olly Betts a écrit :
> Thanks for retesting.
> 
> [...]
> I notice that all the tests which still fail make use of subclassing via
> directors (though some passing tests do too).
> 
> Rereading the discussion, it seems the originally proposed patch made all the
> test pass, or is that no longer true with newer versions of PHP?
> 
Sorry, I only meant that tests failing with "tsrm_ls undefined" passed with
the patch I proposed. There were also segfaulting tests, and the patch did not
address that.

What I have been able to investigate with gdb, is that segfaulting occurs
because some code assume that the pointer "this" is defined, while it is
actually NIL. Why it is so, I do not know. Clearly, the constructor has been
called before executing that part of the code, but somehow, "this" is lost. I
have not been able to follow the code (the call stack contains tens of
procedures).

Pierre

ojwb:
I think I've found the problem - the upcall check is currently being performed on a `NULL` object.  We get away with this in a non-ZTS build, but in a ZTS build we try to access the `swig_zts_ctx` member and the NULL `this` causes the segfault.

Can you test with SWIG commit 682b4dd?

wsfulton:
682b4dd breaks the PHP Travis test case called 'director_finalizer' - https://travis-ci.org/swig/swig/jobs/46731933
I would like to release SWIG this Wed, @ojwb can you either revert the patch or fix it before then? Thanks.

ojwb:
I added some tracing code and this isn't a bug in my change, but a bug uncovered by my change.

```
new SwigDirector_Foo -> 0x147f7e0
_wrap_Foo_orStatus : arg1 = 0x147f7e0
_wrap_Foo_orStatus : director = 0x147f7e8
delete Foo 0x147f7e0
new SwigDirector_Foo -> 0x147f830
_wrap_Foo_orStatus : arg1 = 0x147f830
_wrap_Foo_orStatus : director = 0x147f838
delete Foo 0x147f830
new SwigDirector_Foo -> 0x147f7e0
deleteFoo 0x147f7e0
_wrap_Foo_orStatus : arg1 = 0x147f7e0
Segmentation fault
```

So that last `SwigDirector_Foo` object is deleted before the `orStatus` method gets called.  Previously we'd call a method on a NULL pointer here (which I'm fairly sure is undefined behaviour, but we probably mostly get away with).  Now we use the correct `this` pointer for the object and so actually notice that the object is no longer valid.

I'll see if I can work out what's up.  Reverting seems a very unappealing option.

pierre-labastie:
Le 12/01/2015 01:57, Olly Betts a écrit :
>
> I think I've found the problem - the upcall check is currently being 
> performed on a |NULL| object. We get away with this in a non-ZTS 
> build, but in a ZTS build we try to access the |swig_zts_ctx| member 
> and the NULL |this| causes the segfault.
>
> Can you test with SWIG commit 682b4dd 
> <https://github.com/swig/swig/commit/682b4dd8434401ba57f26f3e3da6fe2571d372ca>?
>
Done. I am sorry to say thatthere are still failures, and that there are 
regressions with ZTS *not* enabled:
Without ZTS:
----
checking php testcase default_args
Parse error: syntax error, unexpected 'LL' (T_STRING), expecting ')' in 
/sources/swig/Examples/test-suite/php/default_args.php on line 50
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Error 255

checking php testcase director_finalizer (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
----
With ZTS:
$ make check-php-examples
[...]
checking Examples/php/extend
Segmentation fault
../../Makefile:1132: recipe for target 'php_run' failed
make[3]: *** [php_run] Error 139

$ make check-php-test-suite
[...]
checking php testcase default_args
Parse error: syntax error, unexpected 'LL' (T_STRING), expecting ')' in 
/sources/swig/Examples/test-suite/php/default_args.php on line 50
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Error 255
[...]
checking php testcase director_basic (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault

[...]
checking php testcase director_classic (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_enum (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_finalizer (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_protected (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
[...]
checking php testcase director_thread (with run test)
../../../Examples/Makefile:1132: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault
---------------------
But director_abstract, director_nested and director_string pass.

Pierre

pierre-labastie:
The regression with testcase default_args has been introduced by:
commit 34787ab98e86b1c82a4d41cecf58ff01d816a4ae
Author: Vadim Zeitlin <vz-swig@zeitlins.org>
Date:   Mon Jan 5 02:50:24 2015 +0100

    Python default argument test cases from issue #294

wsfulton:
@pierre-labastie Can you check you have a stable build as my local testing and Travis tests at https://travis-ci.org/swig/swig/jobs/46824614 don't show the extend example segfault, nor the director_xxxx tests. Travis does show the default_args breakage (which is down to additional tests for issue #294) and so isn't a regression, but I'd like to see it fixed so the test-suite passes for the release due out now. The most concerning thing is the segfault in director_finalizer. @ojwb, could you look at that please and I'll figure out a fix for the default_args testcase.

wsfulton:
Mmm, seems the new default arg tests aimed at testing Python default arg handling have highlighted shortcomings in the PHP default arg handling. There are quite a few problems, I'm wondering if the approach taken for Python coudl be re-used for PHP. @ojwb, I have you don't mind, but I'm going to leave this for you and disable the test for PHP.

pierre-labastie:
@wsfulton The extend example segfault, and the director_xxxx tests segfaults all occur when using PHP built *with* ZTS. I do not know whether this may be called a "stable" build, but this is how this thread started. The reasons why we used ZTS are in my above comment of March 30, 2014. Some comments above seem to imply that the ZTS is not the main target, but @ojwb asked to retest, and this is what I am doing now.

ojwb:
SWIG ought to work with PHP built with ZTS enabled, but generally ZTS is only enabled on Windows which isn't a platform I can test on, and we have had the occasional regression, such as this.

The PHP API has these macros which expand to pass the ZTS pointer around in a ZTS build, and expand to nothing otherwise, so if the use of these macros in a patch is wrong, you don't spot it until you build for ZTS.

ojwb:
@wsfulton I've fixed the default_args issue, so no need to disable that testcase.

ojwb:
@wsfulton And I'm already looking at the director_finalizer testcase.  I'd like to better understand what's going on before we decide to just revert it, as I think we've uncovered a deeper problem from what I've seen so far.

wsfulton:
@pierre-labastie sorry, I read your post too quickly, when ZTS is not enabled, you have the same results as Travis.

Is it worth finding or building an Ubuntu PPA with ZTS enabled and using this in the future on Travis? 

@ojwb Thanks for the default fix. 

@ojwb I'm more interested in releasing 3.0.4 asap without regressions than fixing bugs. If you can't fix in the next 18 hours or so, then I suggest we revert for the release and then you can apply a fixed patch at your leisure for 3.0.5.

ojwb:
If you can find a ZTS build for automatically testing with, that'd be very handy.  I've contemplated building a custom version to test with, but PHP is a bit of a beast to build, and getting a version on all the machines I work on SWIG on isn't an appealing task.

The issue with reverting is really what we revert to that's actually a better state than where we are now.  It's not as simple as that 682b4dd8 regressed us and pre-682b4dd8 is unregressed - there are different problems at different points in the increasinly epic saga that is this ticket.

Just reverting 682b4dd8 leaves us invoking undefined behaviour every time we call a director method.  It seems in a non-ZTS build we may get away with this, but that's at best brittle, and could be causing problems that just haven't been reported yet.  We could revert to an earlier state, for example before any of the fixes related to this ticket, since the ZTS build is already broken with 682b4dd8 reverted anyway.

I'd really like to understand what's going on to inform the decision about what is most sensible to do for 3.0.5, as the fix could easily be a change as small as the revert, and it would be a shame to release 3.0.4 in 18 hours and in 19 hours time to discover this is a serious problem than needs us to release 3.0.5 next week.

ojwb:
OK, 0dd685b fixes the crash in director_finalizer in a simple way (just make that method static, since it only needs the object context for the ZTS pointer, and we can just pass that in from the caller).

I think there's something screwy going on, as the object destruction seems to happen too early (compared to when I'd expect it to happen, and to when it happens under Python).  I'm dubious about how disowning is implemented for PHP (it's nothing like the implementation for Python or Perl, but instead does stuff with the "newobject" flag), but attempting to reimplement that machinery for 3.0.4 seems extremely foolish.  With this patch, we aren't calling a method on a NULL pointer or a deleted object, making it better than the situation either before or after 682b4dd.

I'm hopeful this will pass the testsuite for a ZTS build too, but that has been broken for some time, so I don't think holding up 3.0.4 to test that is worthwhile.

pierre-labastie:
About building PHP, with or without ZTS, I give below the instructions I use to build it in /usr/local (so no interference with system PHP). You then just have to change your PATH so that it begins with /usr/local/bin.
I think you need (on Debian) zlib1g-dev, libgmp-dev, libgdbm-dev, libgettextpo-dev, librealine6-dev, libbz2-dev, libxml2-dev. Not sure I do not forget some (I assume you have a development environment, with the libc and libstdc++ development versions).

```
./configure --with-zlib \
            --enable-bcmath \
            --with-bz2 \
            --enable-dba=shared \
            --with-gdbm \
            --with-gmp \
            --enable-ftp \
            --with-gettext \
            --enable-mbstring \
            --with-readline \
            --enable-maintainer-zts
```

If you do not want ZTS, just suppress the last line. I you want debugging symbols for using gdb, add --enable-debug

pierre-labastie:
@ojwb Great job: the last commits you have done allow all the tests to pass with ZTS enabled, except the *director_thread* one (all the tests pass without ZTS).

ojwb:
How to build it isn't the issue - it's that it's a slow build (https://buildd.debian.org/status/logs.php?pkg=php5&arch=amd64 suggests ~30-90 minutes, though I bet it's longer on my netbook), plus I commonly work on SWIG on at least 4 machines and would need to rebuild a newer version periodically.  I don't really want most of my SWIG hacking time to be spent building custom versions of PHP.

I take it that the director_thread testcase still fail like this:

```
checking php testcase director_thread (with run test)
[Thu Apr 3 10:00:14 2014] Script: '/sources/swig/swig-git/Examples/test-suite/php/director_thread_runme.php'
---------------------------------------
/sources/php/php-5.5.10/Zend/zend_hash.c(563) : Block 0x7f02c85f50d8 status: Invalid pointer: ((thread_id=0xC8E34700) != (expected=0xCC552700)) 
```

This error reads to me that PHP has noted down the current thread id when creating a data structure and then fails when it isn't the same on a later access.  As I said above already, we're assuming that one can call back into the target language interpreter from a different thread, which isn't necessarily going to work for all target languages.  I don't see that we can fix this - it just seems it inherently isn't supported by PHP (but in a non-ZTS build, it presumably doesn't bother to check thread_id in this way).

Having a known failure like this isn't useful, so I've disabled this runme in ecf3ab5.

Which means we can at long last close this ticket.  Thanks to everyone for your help and patience.

pierre-labastie:
@ojwb : With a core i5 at 2.6 GHz, It takes a couple of minutes to build PHP with the instructions above (using make -j5). I do not know what the Debian timings mean. I guess they build several flavours of PHP, and maybe a lot of modules, which are not built (and not useful for swig tests) with the intructions above. Anyway, it is up to you, and I congratulate you again for resolving the bug.

PR #184, opened 2014-05-31
Fix function naming conflict with class overloads.

lefticus:
This fix takes into account the classname while generating overload
handlers.

Example:

If you have two classes:

```cpp
  class A {
    public:
      void doSomething(int);
      void doSomething(double);
  };

  class B {
    public:
      void doSomething(int);
      void doSomething(double);
  };
```

Before this patch, the overload handlers for `A::doSomething` and
`B::doSomething` create conflicting names and function redefinition compile errors
are caused.

After the patch, the overload handlers are named `classname_doSomething`
and no longer conflict.

This is might not the best way to implement this, but it
solves a critical problem on large projects, and specifically can affect
operator overloads that are being wrapped.


lefticus:
@wsfulton I've cleaned up the `_wrap__wrap` naming issue, and moved the tests out into the test-suite folder.

The downside is that this seems to have uncovered issues with overloads in basically every other generator.

lefticus:
It appears that the issues here have been fixed by some other commit and no longer exist in 3.0.6.

choois:
> This fix takes into account the classname while generating overload handlers.
> 
> Example:
> 
> If you have two classes:
> 
> ```c++
>   class A {
>     public:
>       void doSomething(int);
>       void doSomething(double);
>   };
> 
>   class B {
>     public:
>       void doSomething(int);
>       void doSomething(double);
>   };
> ```
> 
> Before this patch, the overload handlers for `A::doSomething` and `B::doSomething` create conflicting names and function redefinition compile errors are caused.
> 
> After the patch, the overload handlers are named `classname_doSomething` and no longer conflict.
> 
> This is might not the best way to implement this, but it solves a critical problem on large projects, and specifically can affect operator overloads that are being wrapped.

This doesn't handle the following case:

```
class Toy {
     public: 
          Shape container_shape();
}; 

class Toy_container {
     public: 
          Shape shape();
}; 
```

The above definitions will result in
    error: redefinition of ‘PyObject* _wrap_Toy_container_shape

PR #185, opened 2014-06-04
Implement support for parts of Java annotations.

farre:
Implementation of parts of https://github.com/swig/swig/wiki/GSoC-2013-ideas#Java_annotations_support mainly annotations for constructors, methods and type declarations. What's missing is the annotations for in/out arguments. This is basically a word for word copy from the same feature in the C# module.

### DIFF
--- a/Examples/test-suite/java/java_annotations_runme.java
+++ b/Examples/test-suite/java/java_annotations_runme.java
@@ -1,5 +1,8 @@
 import java_annotations.*;
 
+import java.lang.annotation.*;
+import java.lang.reflect.*;
+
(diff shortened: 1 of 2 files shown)


wsfulton:
Hi, @farre . Are you going to get a chance to finish this off?

farre:
Unfortunately not at this time. I quit the job where I was using SWIG and
then it feel out of my mind. Sorry.

ojwb:
Essentially abandoned by the submitter, but appears to only need updating for current git master, then adding tests and docs which would be very similar to existing ones for csharp.

PR #190, opened 2014-06-17
Cffi fix

enupten:
Attached is a fix for the case when arrays are declared inside structs. For the  case:
```c
struct _vec3 {
  double v[3];}
````
SWIG currently generates the incorrect FFI code:
```lisp
(cffi:defcstruct vec3 
   (v :pointer))
```
instead of:
```lisp
(cffi:defcstruct vec3 
   (v :double :count 3))
```
This patch fixes this bug.


wsfulton:
This can't be merged without a conflict. Please re-open with a fix and include the testcases promised and test-suite.

PR #232, opened 2014-09-24
Python: Fix property access with director

diorcety:
When adding a property in a class extending a directory class, we can't set/read property correctly.
Here the fix, and the test

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -172,6 +172,7 @@ CPP_TEST_CASES += \
 	director_abstract \
 	director_alternating \
 	director_basic \
+	director_property \
 	director_binary_string \
 	director_classes \
 	director_classic \
--- a/Examples/test-suite/director_property.i
+++ b/Examples/test-suite/director_property.i
@@ -1,0 +1,151 @@
+%module(directors="1") director_property
+
+%warnfilter(SWIGWARN_TYPEMAP_THREAD_UNSAFE,SWIGWARN_TYPEMAP_DIRECTOROUT_PTR) MyClass::pmethod;
+
+%{
+ #include <string>
+
+ class Foo {
+ private:
+   std::string a;
+ public:
+   virtual ~Foo() {}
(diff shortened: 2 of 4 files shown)


wsfulton:
@diorcety Looks like this change has broken use of swig with -classic as mentioned on the swig-devel mailing list. It is pretty urgent to fix this, are you around to take a look? Probably the test-suite goes wrong when run with -classic, but I am not able to check right now.

diorcety:
i will take a look this evening

diorcety:
Ok maybe we have to add a switch in the case of modern or !classic . __setattr__ is maybe not defined in the old style class or something like that

wsfulton:
The python_nondynamic testcase reproduces the problem:

```
checking python testcase python_nondynamic (with run test)
  File "python_nondynamic_runme.py", line 46, in <module>
    cc.d = 3
  File "/home/william/swig/github/swig/Examples/test-suite/python/python_nondynamic.py", line 100, in <lambda>
    __setattr__ = lambda self, name, value: _swig_setattr(self, C, name, value)
  File "/home/william/swig/github/swig/Examples/test-suite/python/python_nondynamic.py", line 51, in _swig_setattr
    return _swig_setattr_nondynamic(self, class_type, name, value, 0)
  File "/home/william/swig/github/swig/Examples/test-suite/python/python_nondynamic.py", line 45, in _swig_setattr_nondynamic
    object.__setattr__(self, name, value)
TypeError: can't apply this __setattr__ to instance object
make[1]: *** [python_nondynamic.cpptest] Error 1
```

The classic and modern flags in python.cxx are set when -classic and -modern are passed on the command line, so you can use those and I suggest generating the old code when -classic is set. The generated .py file also contains a _newclass variable which looks like it could be used.

wsfulton:
Fix committed. With new style classes __setattr__ is used and with old style classes __dict__ is used (like it used to be). Amazingly the director_property testcase in this patch now works with old style classes and new style classes :)

PR #250, opened 2014-10-25
Fix python global object constants

ptomulik:
With current swig the following simple test case

```swig
%module object_constants

struct Type1 { };
%constant Type1 t1;

%{
  struct Type1 {};
  Type1 t1;
%}
```

does not compile, when generated with ``-builtin`` option. The compilation error is:

```console
swig -python -c++ -builtin -o object_constants_wrap.cxx object_constants.i
g++ -o object_constants_wrap.o -c -fpic -I.  -I/usr/include/python2.7 object_constants_wrap.cxx 
object_constants_wrap.cxx: In function ‘void init_object_constants()’:
object_constants_wrap.cxx:1170:83: error: ‘self’ was not declared in this scope
 #define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerObj(self, ptr, type, flags)
                                                                                   ^
object_constants_wrap.cxx:4530:70: note: in expansion of macro ‘SWIG_NewPointerObj’
   SWIG_Python_SetConstant(d, d == md ? public_interface : NULL, "t1",SWIG_NewPointerObj(SWIG_as_voidptr(&t1),SWIGTYPE_p_Type1, 0 ));
```

This patch fixes the problem.

**Advice needed**: I've set ``self`` to ``m`` (module) but maybe it should be ``NULL``? Please advise.

### DIFF
--- a/Examples/test-suite/python/constant_directive_runme.py
+++ b/Examples/test-suite/python/constant_directive_runme.py
@@ -1,0 +1,21 @@
+import constant_directive
+
+if type(constant_directive.TYPE1_CONSTANT1) != constant_directive.Type1:
+  print("TYPE1_CONSTANT1    type: {}".format(type(constant_directive.TYPE1_CONSTANT1)))
+  raise RuntimeError("fail");
+if type(constant_directive.getType1Instance()) != constant_directive.Type1:
+  print("getType1Instance() type: {}".format(type(constant_directive.getType1Instance())))
+  raise RuntimeError("fail");
+
+if constant_directive.TYPE1_CONSTANT1.val != 1:
+  print "constant_directive.TYPE1_CONSTANT1.val != 1"
+  print "constant_directive.TYPE1_CONSTANT1.val is %r" % constant_directive.TYPE1_CONSTANT1.val
+  raise RuntimeError("fail")
+if constant_directive.TYPE1_CONSTANT2.val != 2:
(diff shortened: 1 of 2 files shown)


ptomulik:
I have updated this PR following all your suggestions. I've also ran tests for some arbitraily choosen combinations of swig flags. My test run script and the results may be found [here](https://gist.github.com/3a7991f25608e69dda5d.git).  There is some problem with ``-classic``, probably the ``constant_directive_runme.py`` script has to be adjusted (I believe, the instances of classic clases are seen quite differently in python type system than instances of modern classes). I've also seen, that all the tests having runtime scripts (*_runme.py) segfault when ``-modernargs`` flag is used.

ptomulik:
Ok, I have fixed the issue with ``-classic``. The ``-modernargs`` + ``-builtin`` should be addressed in separate PR as it's not related to this particular PR. I think, this PR is ready to be merged.

wsfulton:
-builtin should not be used with many of the other flags like -modern, but then again, SWIG shouldn't seg fault if this is attempted! Thanks, for the update, I'm about to merge them in.

ptomulik:
Thank you for merging. I just want to make a remark about segfault. It happens when we use ``-builtin -modernargs``. However, if we add to this duet the ``-fastunpack``, for example, it pass tests again.

PR #251, opened 2014-10-26
Initial implementation of strong and scoped enums in python

ptomulik:
May look quite awkward for first sight, but this is a minimalistic approach to the matter. Primarily I've used the existing "class" machinery to emit classes that represent enums. The purpose of `python/enums.i` is to clear default typemaps for `enum SWIGTYPE` and define `%typemap(constcode,...) enum SWIGTYPE`. The rest of the stuff is handled by default swig typemaps.

To enable strong/scoped types, one has to use `%feature("python:enum",...)`. Obviously, this may be done per-module or per-declaration. Strong/scoped enums may be enabled globally -- new options `-strongenum`, `-scopedenum` and `-enumclass` are provided for this purpose.

I have some doubts regarding memory management. No sure if I shouldn't use `SWIG_POINTER_OWN` in the `constcode` typemap. Currently this causes warnings about memory leaks, when running the test case.


ptomulik:
Anyone to help here?

derkuci:
Just found this thread as well as the discussion emails of @vadz in 2014.  I have been looking for better support of scoped enums and thank you guys for trying to implement it, especially together with the "builtin" mode.

@ptomulik, If I may suggest, do you think it's easier/cleaner to support scoped enums instead of both scoped and strong enums?  I see the issues in your last post here were all about strong enums.

ptomulik:
Quite hard to recall, but looking at code it seems that there is not much difference. At least in both cases we need to generate a class (for scoping the enum names and eventually for giving a type for enum values). Scoping issues that prevent this PR to be pushed forward are mentioned in #316.

baldurk:
Hey @ptomulik I'm using this commit in a local fork of swig with some fixes and things for my project. It's working great, so thank you very much for that :smile:.

I wanted to post a small change I made locally to fix a compile error with enum classes. It's up to you if this makes sense or if you want it. When the generated code referenced a templated object with the `enum class` as an inner type (the obvious simple example being a `std::vector`) the generated code included `SwigValueWrapper< vector< enum MyEnum > >` which caused a compile error as it should instead be `SwigValueWrapper< vector< enum class MyEnum > >`, since you can't declare a scoped enum as an unscoped one.

This small patch does that for me, although it would possibly break if the declaration was originally `enum struct` - I didn't test if the compiler allows that as equivalent.

```patch
diff --git a/Source/Modules/typepass.cxx b/Source/Modules/typepass.cxx
index bf8028c..c6da434 100644
--- a/Source/Modules/typepass.cxx
+++ b/Source/Modules/typepass.cxx
@@ -777,6 +777,7 @@ class TypePass:private Dispatcher {

   virtual int enumDeclaration(Node *n) {
     String *name = Getattr(n, "name");
+    int scopedenum = GetFlag(n, "scopedenum");

     if (name) {
       String *scope = 0;
@@ -802,10 +803,10 @@ class TypePass:private Dispatcher {
          Setattr(n, "tdname", tdname);
        }

-       SwigType *t = NewStringf("enum %s", nname);
+       SwigType *t = NewStringf("enum %s%s", scopedenum ? "class " : "", nname);
        SwigType_typedef(t, name);
       } else {
-       SwigType *t = NewStringf("enum %s", name);
+       SwigType *t = NewStringf("enum %s%s", scopedenum ? "class " : "", name);
        SwigType_typedef(t, name);
       }
       Delete(scope);
```

sunavlis:
Hi, I just found this "old" issue. Is there any activity on it and and plan to merge them? So that the strong enum is available in python?

ojwb:
@sunavlis Strongly typed enums are *already* available in Python (they have been since SWIG 3.0.3 which was released more than 9 years ago). For Python, SWIG currently just includes the enum class name into the wrapped name exposed in Python.

AIUI this PR aims to wrap them in a better way.  Sadly it seems nobody is actively working on it any more.

PR #261, opened 2014-11-05
[Go] Do not start getters with Get

steeve:
Removes the `Get` prefix from getters in Go.

From their style guide: http://golang.org/doc/effective_go.html#Getters

> Go doesn't provide automatic support for getters and setters. There's nothing wrong with providing getters and setters yourself, and it's often appropriate to do so, but it's neither idiomatic nor necessary to put Get into the getter's name. If you have a field called owner (lower case, unexported), the getter method should be called Owner (upper case, exported), not GetOwner. The use of upper-case names for export provides the hook to discriminate the field from the method. A setter function, if needed, will ikely be called SetOwner. Both names read well in practice:

### DIFF
--- a/Source/Modules/go.cxx
+++ b/Source/Modules/go.cxx
@@ -637,8 +637,10 @@ private:
       bool is_set = Strcmp(Char(name) + Len(name) - 4, "_set") == 0;
       assert(is_set || Strcmp(Char(name) + Len(name) - 4, "_get") == 0);
 
-      // Start with Set or Get.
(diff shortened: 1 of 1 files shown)


ojwb:
@steeve Any progress on fixing the testsuite?

@ianlancetaylor Do you agree this a desirable change?  It sounds like it to me, but I don't really know Go well.

ianlancetaylor:
This will break many existing SWIG users, so I don't think it would be wise to make this change.  I suppose I should have written it that way in the first place, but I didn't.

Issue #289, opened 2014-12-24
Go doesn't track C++ object ownership


michael-schaller:
Go currently doesn't track C++ object ownership like other target languages do. Tracking the C++ object ownership would allow Go to automatically free C++ objects once they are no longer referenced and thus simplify the usage of a SWIG wrapper as objects wouldn't need to be freed manually unless the wrapper would specifically demand it.

Generated Java code for an example C++ class:
``` java
public class Test {
  private long swigCPtr;
  protected boolean swigCMemOwn;

  protected Test(long cPtr, boolean cMemoryOwn) {
    swigCMemOwn = cMemoryOwn;
    swigCPtr = cPtr;
  }

  protected static long getCPtr(Test obj) {
    return (obj == null) ? 0 : obj.swigCPtr;
  }

  protected void finalize() {
    delete();
  }

  // Call C++ destructor
  public synchronized void delete() {
    if(swigCPtr != 0 && swigCMemOwn) {
      swigCMemOwn = false;
      exampleJNI.delete_Test(swigCPtr);
    }
    swigCPtr = 0;
  }

  // Call C++ constructor
  public Test() {
    this(exampleJNI.new_Test(), true);
  }
}
```

In order for Go to track the C++ object ownership all structs that represent a C++ class would need to have an ownership boolean, the ownership boolean would need to be initialized properly (tracked or not tracked) on object creation and in case a C++ object is tracked runtime.SetFinalizer needs to be called after object creation with its delete function as finalizer function.

This topic has been previously discussed with @ianlancetaylor in issue #266.

ianlancetaylor:
I've seen existing SWIG code fail due to runtime.SetFinalizer.  To briefly sketch the problem, class C1 had a field that was a vector of items of class C2.  The C1 destructor destroyed the vector and all the C2 elements.  The Go code had a variable of type wrap.C1, and retrieved a value of type C2.  The value of type C2 was then passed to various other functions.  Although the variable of type C2 was not returned, the function made no further references to the variable of type C1.  Therefore, the C1 variable was not live, and the Go garbage collector observed that, freed it, and ran the finalizer.  Running the finalizer ran the C++ destructor, which destroyed the C1 object and the C2 object.  However, the Go code was still using the Go version of the C2 object, which in effect became a dangling pointer.

My point is that automatic destruction of SWIG objects takes great care in Go, because objects can be finalized at points where many programmers would consider them to be obviously live.  I don't know if Java has similar problems.

[...]

PR #291, opened 2014-12-26
[Go] Renamed/refactored Swigcptr<classname> (uintptr) to SwigClass<classname> (struct).

michael-schaller:
* First part of issue #289 to let Go track object ownership. This part renamed/refactored Swigcptr<classname> (uintptr) to SwigClass<classname> (struct) so that later an ownership boolean can be added to the struct.
* This change required two tests to be fixed because of hard-coded names.
* This change required one test to be disabled because a tiny bit more static type-checking is in place.


ianlancetaylor:
This will definitely break existing code.  For better or for worse, the Swigcptr type is actually documented (http://swig.org/Doc3.0/Go.html#Go_classes).  There is existing code like SwigcptrClass(0) to get a C++ NULL pointer, and there is code like SwigcptrClass1(class2.Swigcptr()) to do, in effect, an unchecked dynamic_cast.  Unfortunately I don't think this approach will work without transitional changes.

michael-schaller:
Ian, I totally agree with you that this change would break existing code. I guess we should have more discussion in issue #289 before making an actual change. Nevertheless thank you for reviewing this experimental code.

PR #294, opened 2015-01-05
Fix handling of non-first arguments default values in Python.

vadz:
This fixes the bug reported in the "Default argument not being added to wrapped Python .py File" mailing list thread.

Please review as I don't really understand the semantics of `tmap:in:next`. I hadn't touched this code at all in the commit that broke it (15b3690) exactly because of this, even though it did seem suspicious to me, but it turns out that it's more than suspicious and simply wrong, as in the case of the new unit test this result in simply not checking the second parameter at all because `tmap:in:next` is null and so we exit the loop in `is_representable_as_pyargs()` after the first iteration.

Unfortunately I don't know where does this code come from, it appears in git since 3d8ddfc4, but this commit ("Python 3.0 branch merge") is too huge to be able to extract anything useful from it.

Also notice that there is another instance of the same code in `make_autodocParmList()`, but fixing it there in the same way broke `autodoc` and `char_binary` tests, so I didn't touch it. It's worrisome that the code which seems clearly wrong to me is needed to make the tests work currently though...

Commit message:

---
Since 15b3690, non-first arguments with default values of non-primitive types
were not used in Python wrappers, resulting in invalid Python code if the
preceding arguments did have default values.

This happened because is_representable_as_pyargs() didn't check any arguments
except the first one at all in most cases because it used tmap:in:next
unconditionally, but it is usually null. Fall back on the next argument in the
function parameter list now to fix this.

### DIFF
--- a/Examples/test-suite/default_args.i
+++ b/Examples/test-suite/default_args.i
@@ -13,6 +13,12 @@
 %inline %{
   #include <string>
 
+  // All kinds of numbers: hex, octal (which pose special problems to Python), negative...
+  void lots_of_args(int pos = -1, unsigned rgb = 0xabcdef, int mode = 0644) { }
+
+  // Long long arguments are not handled at Python level currently but still work.
+  void seek(long long offset = 0LL) {}
+
   // Anonymous arguments
   int anonymous(int = 7771);
   int anonymous(int x) { return x; }
--- a/Examples/test-suite/default_args.i
+++ b/Examples/test-suite/default_args.i
@@ -199,6 +199,7 @@ namespace Space {
 struct Klass {
   int val;
   Klass(int val = -1) : val(val) {}
+  static Klass inc(int n = 1, const Klass& k = Klass()) { return Klass(k.val + n); }
 };
 Klass constructorcall(const Klass& k = Klass()) { return k; }
 
--- a/Examples/test-suite/default_args.i
+++ b/Examples/test-suite/default_args.i
@@ -35,6 +35,12 @@
       bool blah(speed s = FAST, flavor f = SWEET) { return (s == FAST && f == SWEET); };
   };
 
+  // using base class enum in a derived class
+  class DerivedEnumClass : public EnumClass {
+  public:
+    void accelerate(speed s = SLOW) { }
+  };
+
   // casts
   const char * casts1(const char *m = (const char *) NULL) {
     char *ret = NULL; 
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -17,6 +17,8 @@
 static int treduce = SWIG_cparse_template_reduce(0);
 
 #include <ctype.h>
+#include <errno.h>
+#include <stdlib.h>
 
 #define PYSHADOW_MEMBER  0x2
 #define WARN_PYTHON_MULTIPLE_INH 405
@@ -1815,19 +1817,128 @@ public:
   }
 
   /* ------------------------------------------------------------
+   *  convertDoubleValue()
+   *    Check if the given string looks like a decimal floating point constant
+   *    and return it if it does, otherwise return NIL.
+   * ------------------------------------------------------------ */
+  String *convertDoubleValue(String *v) {
+    const char *const s = Char(v);
+    char *end;
+
+    (void)strtod(s, &end);
+    if (errno != ERANGE && end != s) {
+      // An added complication: at least some versions of strtod() recognize
+      // hexadecimal floating point numbers which don't exist in Python, so
+      // detect them ourselves and refuse to convert them (this can't be done
+      // without loss of precision in general).
+      //
+      // Also don't accept neither "NAN" nor "INFINITY" (both of which
+      // conveniently contain "n").
+      if (strpbrk(s, "xXnN"))
+	return NIL;
+
+      // Disregard optional "f" suffix, it can be just dropped in Python as it
+      // uses doubles for everything anyhow.
+      for (char* p = end; *p != '\0'; ++p) {
+	switch (*p) {
+	  case 'f':
+	  case 'F':
+	    break;
+
+	  default:
+	    return NIL;
+	}
+      }
+
+      // Avoid unnecessary string allocation in the common case when we don't
+      // need to remove any suffix.
+      return *end == '\0' ? v : NewStringWithSize(s, end - s);
+    }
+
+    return NIL;
+  }
+
+  /* ------------------------------------------------------------
    * convertValue()
    *    Check if string v can be a Python value literal or a
    *    constant. Return NIL if it isn't.
    * ------------------------------------------------------------ */
   String *convertValue(String *v, SwigType *t) {
-    char fc = (Char(v))[0];
-    if (('0' <= fc && fc <= '9') || '\'' == fc || '"' == fc) {
-      /* number or string (or maybe NULL pointer) */
-      if (SwigType_ispointer(t) && Strcmp(v, "0") == 0)
-	return NewString("None");
-      else
-	return v;
+    const char *const s = Char(v);
+    char *end;
+
+    // Check if this is a number in any base.
+    (void)strtol(s, &end, 0);
+    if (end != s) {
+      if (errno == ERANGE) {
+	// There was an overflow, we could try representing the value as Python
+	// long integer literal, but for now don't bother with it.
+	return NIL;
+      }
+
+      if (*end != '\0') {
+	// If there is a suffix after the number, we can safely ignore any
+	// combination of "l" and "u", but not anything else (again, stuff like
+	// "LL" could be handled, but we don't bother to do it currently).
+	bool seen_long = false;
+	for (char* p = end; *p != '\0'; ++p) {
+	  switch (*p) {
+	    case 'l':
+	    case 'L':
+	      // Bail out on "LL".
+	      if (seen_long)
+		return NIL;
+	      seen_long = true;
+	      break;
+
+	    case 'u':
+	    case 'U':
+	      break;
+
+	    default:
+	      // Except that our suffix could actually be the fractional part of
+	      // a floating point number, so we still have to check for this.
+	      return convertDoubleValue(v);
+	  }
+	}
+      }
+
+      // Deal with the values starting with 0 first as they can be octal or
+      // hexadecimal numbers or even pointers.
+      if (s[0] == '0') {
+	if (Len(v) == 1) {
+	  // This is just a lone 0, but it needs to be represented differently
+	  // in Python depending on whether it's a zero or a null pointer.
+	  if (SwigType_ispointer(t))
+	    return NewString("None");
+	  else
+	    return v;
+	} else if (s[1] == 'x' || s[1] == 'X') {
+	  // This must have been a hex number, we can use it directly in Python,
+	  // so nothing to do here.
+	} else {
+	  // This must have been an octal number, we have to change its prefix
+	  // to be "0o" in Python 3 only (and as long as we still support Python
+	  // 2.5, this can't be done unconditionally).
+	  if (py3) {
+	    String *res = NewString("0o");
+	    Append(res, NewStringWithSize(s + 1, end - s - 1));
+	    return res;
+	  }
+	}
+      }
+
+      // Avoid unnecessary string allocation in the common case when we don't
+      // need to remove any suffix.
+      return *end == '\0' ? v : NewStringWithSize(s, end - s);
     }
+
+    // Check if this is a floating point number (notice that it wasn't
+    // necessarily parsed as a long above, consider e.g. ".123").
+    if (String *res = convertDoubleValue(v)) {
+      return res;
+    }
+
     if (Strcmp(v, "true") == 0 || Strcmp(v, "TRUE") == 0)
       return NewString("True");
     if (Strcmp(v, "false") == 0 || Strcmp(v, "FALSE") == 0)
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -1946,12 +1946,15 @@ public:
     if (Strcmp(v, "NULL") == 0 || Strcmp(v, "nullptr") == 0)
       return SwigType_ispointer(t) ? NewString("None") : NewString("0");
 
-    // This could also be an enum type, default value of which is perfectly
-    // representable in Python.
-    Node *lookup = Swig_symbol_clookup(v, 0);
-    if (lookup) {
-      if (Cmp(Getattr(lookup, "nodeType"), "enumitem") == 0)
-	return Getattr(lookup, "sym:name");
+    // This could also be an enum type, default value of which could be
+    // representable in Python if it doesn't include any scope (which could,
+    // but currently is not, translated).
+    if (!Strchr(s, ':')) {
+      Node *lookup = Swig_symbol_clookup(v, 0);
+      if (lookup) {
+	if (Cmp(Getattr(lookup, "nodeType"), "enumitem") == 0)
+	  return Getattr(lookup, "sym:name");
+      }
     }
 
     return NIL;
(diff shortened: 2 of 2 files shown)


xantares:
Hi,

I think there's another issue with the modification that landed in the 3.0.3 regarding how arguments are passed.

Here's a MNWE from the problem I encountered in my lib:
```
class Foo
{
public:
  enum ExternalFileFormat { CSV };
  Foo(double x = 8.0);
private:
  double x_;
};

class Bar : public Foo
{
public:
  Bar(double x = 8.0, double y = 9.6);

  static Bar FromFile(const std::string & fileName,
      const std::string & parameters = "",
      const ExternalFileFormat format = CSV) {return Bar(9.7); }

private:
  double y_;
};
```

With the 3.0.3 or this PR, it fails on import:
```
    def Bar_FromFile(fileName, parameters="", format=CSV):
NameError: name 'CSV' is not defined
```

because CSV is valid only in the Foo class scope:
```
class Foo(_object):
    __swig_setmethods__ = {}
    __setattr__ = lambda self, name, value: _swig_setattr(self, Foo, name, value)
    __swig_getmethods__ = {}
    __getattr__ = lambda self, name: _swig_getattr(self, Foo, name)
    CSV = _example.Foo_CSV
```

and it fails at the (new) definition of Bar::FromFile:
```
def Bar_FromFile(fileName, parameters="", format=CSV):
    return _example.Bar_FromFile(fileName, parameters, format)
Bar_FromFile = _example.Bar_FromFile
```

What do you think ?

vadz:
@xantares this is definitely a bug, but a different one, please open an issue for it. I'll try to fix this, although right now I don't really know how am I going to do it (any advice from @wsfulton would be welcome!).

xantares:
ok, I opened #296

wsfulton:
The patch is definitely an improvement. However, the way the parameters are being examined is not really right and unfortunately is a bit complicated. 

The correct way to handle them is as in PYTHON::functionWrapper, so something like:


```cpp
for (i = 0, p = l; i < num_arguments; i++) {
  while (checkAttribute(p, "tmap:in:numinputs", "0")) {
    p = Getattr(p, "tmap:in:next");
  }
  ...
  p = Getattr(p, "tmap:in:next");
}
```
and if num_arguments is not available, as in JAVA::proxyClassFunctionHandler:

```cpp
for (i = 0, p = l; p; i++) {

  /* Ignored varargs */
  if (checkAttribute(p, "varargs:ignore", "1")) {
    p = nextSibling(p);
    continue;
  }

  /* Ignored parameters */
  if (checkAttribute(p, "tmap:in:numinputs", "0")) {
    p = Getattr(p, "tmap:in:next");
    continue;
  }
  ...
  p = Getattr(p, "tmap:in:next");
}
```
The above is necessary for ignored parameters (numinputs=0) and multi-argument typemaps. In order for the above to work, the 'in' typemaps must be attached to the parameters. This is done at some point via a call:  Swig_typemap_attach_parms("in", l, NULL). 

The problem with is_representable_as_pyargs is it is assuming that the 'in' typemaps are attached. This is not always the case. I attempted to put the looping in as shown above, but then it caused further problem for cases like the following in default_args.i:

```cpp
struct Except {
  Except(bool throwException, int a = -1) throw (int) {
    if (throwException)
      throw a;
  }
};
```

precisely because the 'in' typemaps are not attached. 

Note that both make_autodocParmList and is_representable_as_pyargs calls convertValue. The parameters passed to convertValue must be identical in order to correctly determine whether to use *args or the expanded/named parameters. Note that autodocParmList calls Swig_typemap_attach_parms and is_representable_as_pyargs does not. The truly correct fix will be to ensure is_representable_as_pyargs is consistently called and the parameters are looped correctly. I'll have a go at this (and try make it not too inefficient).

I notice that -ve numbers are not accepted in convertValue and hence default values with -ve values are not accepted. We could easily enhance this easily enough. I am a bit concerned that if convertValue determines that a default value can be represented in Python and it gets it wrong, there is no easy way to alter this. For example:

```cpp
void xxx(long long abc = 8LL) {}
```

There is a similar problem for Java constants, we the default is to getting the value from C, thereby avoiding the problem. Various features will turn on the value coming from Java. All in all, I am wondering if we have turned on this new code generation style when it is not robust enough yet.

wsfulton:
Also I meant to mention that CopyParmList does a bad job of copying the parameter list. It only copies attribute values that are strings. Hence it does not copy the tmap:in:next as it contains a pointer. So the copy is bit of an abomination, however, it is better than copying the attributes that are pointers. 

The more I think about this, I think we should roll back this change for default arguments and overloaded parameters. Another case came up in #296 where default values that are octal values are not generated correctly and to complicate matters, are handled differently in python 2 and python 3.

vadz:
> The more I think about this, I think we should roll back this change for default arguments and overloaded parameters.

Rolling back everything would be pretty bad, having `*args` for all Python functions is extremely unhelpful. We would need to roll back enum parameter changes if I don't find a way to deal with the scope issue from #296, but please give me at least some time to look into it as it's definitely very nice to have defaults for enums, especially when dealing with enum-heavy APIs like ours.

Notice that octal stuff is a totally different issue and unrelated to the enum changes, it always existed and was just hidden before by the unconditional use of `*args`.

wsfulton:
I'm suggesting that wherever a C++ default value is generated as a python default value, we use *args by default as that is guaranteed to work, and is of prime concern. Making this more flexible via a new feature would meet your needs. This is the approach taken with %javaconst to solve a near identical problem. So if we have:
```swig
void xxx(long long abc = 8LL);
void yyy(int abc = 123);
```
by default the *args would be generated:
```python
def xxx(*args):
    return _example.xxx(*args)
def yyy(*args):
    return _example.yyy(*args)
```
Using a new feature called %pydefaultargs, say, would generate:
```python
def xxx(abc=8LL):
    return _example.xxx(abc)
def yyy(abc=123):
    return _example.yyy(abc)
```
which wouldn't work for abc, so it can be simply turned off for the xxx function as the feature is attached to functions/methods:
```swig
%pydefaultargs; // global feature attaches to all functions
%nopydefaultargs xxx;
void xxx(long long abc = 8LL);
void yyy(int abc = 123);
```
to give:
```python
def xxx(*args):
    return _example.xxx(*args)
def yyy(abc=123):
    return _example.yyy(abc)
```

In Java, you can use %javaconstvalue to replace 8LL with say just 8. However, I don't think this fine tuning is a good idea for python default args as currently you'd have to use a typemap for Python. This could be dangerous as a named typemaps such as: 
```swig
%typemap(pydefaultvalue) long long abc "8";
```
would affect all parameters with the name abc, and they may have different default values. It is okay for Java as it works for constants which are always uniquely named symbols, unlike parameter names. One day I'd like to add support for better typemap targeting of parameters, so you could uniquely target abc in xxx using an extension to C++ symbol scoping using the outer function as a scope, eg:
```swig
%typemap(pydefaultvalue) long long xxx::abc "8";
```
But I can't see that happening anytime soon.

Anyway, I think we need to push out swig-3.0.4 asap with the *args default as described and add in %pydefaultargs in a later release if it isn't implemented really quickly.

vadz:
> I'm suggesting that wherever a C++ default value is generated as a python default value, we use *args by default as that is guaranteed to work

This is not as bad as I feared, but still quite far from ideal. IMHO SWIG should do the right thing by default, such as using `abc = 123` without being told anything.

vadz:
I'm checking in the dumbest possible fix for #296 for now which just reverts to the old behaviour for any "scoped" enums. This is clearly not ideal, as it doesn't use default values in many cases when it would be fine to use them (e.g. inside the class itself, anywhere if the scope in the enum is a name of a namespace and not a class, ...), but it should at least fix the problem.

I still have no idea how to determine the Python expression corresponding to the given C++ enum constant in general... There doesn't seem to be a typemap for this, or am I missing something again?

wsfulton:
I know it is nicer to have the default values shown in Python, but, converting these all successfully is not easy. It is more important that the generated code actually works. This is a general philosophy in SWIG to aim for code that always compiles and loads rather than generate optimally working code as the complexity of C++ makes this fraught with pitfalls. So, the general philosophy also is to allow for user to tweak/improve the generated code. Hence, I'm about to check in the change described earlier with *args as the default for default parameters (as it will always work) and the possibility of using a new feature "python:defaultargs", renamed from the previously mentioned "pydefaultargs".

This commit is to get a release out quickly for a pretty bad regression. The idea is that PYTHON::is_representable_as_pyargs() will get better over time at converting C++ defaults to Python defaults. @vadz, please adapt your patches for the new commit. Note that the new testcase python_default_args uses the default_args testcase including the Python runtime code. We should expand default_args.i with lots of tricky default values that don't convert to Python and exclude them using %nopythondefaultargs in python_default_args.i. Over time this list of working defaults should improve.

I'll add docs for "python:defaultargs" next.

vadz:
Sorry, I just don't understand this at all. With the last version of this PR there were no more problems with any code being invalid (AFAIK), the only problem was exactly that suboptimal code was generated. I share your goal (to fix the regression a.s.a.p.) and the plans for the future (make `convertValue()` smarter), but this was *already* done.

So I just see no need for `python:defaultargs` but if you add it now it will need to be kept later because of compatibility. Could you please reconsider its addition? Again, to the best of my knowledge, there are no issues remaining with the last version of this PR...

I will try to update this patch, of course, but it's pretty demotivating to have done everything to fix the regression as quickly as possible (and doing it before your commits), only for the changes to be discarded in favour of a solution that is worse for the SWIG users.

wsfulton:
I wouldn't be de-motivated as your changes will be accepted and I appreciate the quick response. Okay, so I should have merged your patches in and then added in python:defaultargs, so I apologise for that. I've been a bit short of time, but I've had a good look at the patch and I like the improvements it makes. I'll do the merge later today before release.

I've given python:defaultargs some more thought, but would like to keep it, as in summary I'm just being rather conservative about change management.

Each of the points below by themselves might be not so important, but adding them up
overall explains my decision...
- The change change away from *args requires me to rush out a swig-3.0.4, I would like to mitigate this from happening again.
- I'm concerned about adding in corner case bugs and users not being able to work around them.
- I don't want our naivity to overlook possible new problems if we don't use
python:defaultargs:
  a) I'm a little concerned about the use of strtod and the numerous bugs reported in many different compilers. 
  b) Has the locale behaviour of strtod been thought about?
- With the "default:pythonargs" feature, we avoid problematic corner case regressions by default and provide an easy way to turn them off/on.
- We both agree that convertValue can get smarter in the future, but I really do not want to take any unnecessary risks wrt regressions. I can see many many tweaks and improvements to convertValue() being introduced over time. Each of these could easily introduce regressions. For many users who don't care about showing default values in Python (or are using -builtin), this won't matter if we introduce bugs, and for those that are using default:pythonargs, they are easy to work around by turning the feature off for a particular function while we fix bugs. Hence I think we can be more aggressive about improving default:pythonargs as there users can
easily workaround bugs.
- It fits in with the overall SWIG philisophy of working by default and users being able to easily add in improvements via SWIG directives. Adding in the "default:pythonargs" global feature is really easy.
- The current implementation of convertValue() is inconsistent, eg enums in classes don't work but global enums do and getting these to reliably work is likely to be probably non-trivial. I'd like users to have control over consistency.
- I am not 100% sure that the new implementation of convertValue(), even though it is pretty conservative, is bug free. You even state 'AFAIK', which I think means you are not absolutely sure.

ojwb:
> b) Has the locale behaviour of strtod been thought about?

That's only an issue if you call `setlocale()`, which we don't do in SWIG currently.

The PHP backend already uses `strtod()`, and `DOH/base.c` uses `atof()` which the man page says behaves like `strtod()`.

vadz:
I fundamentally disagree with the idea that SWIG should do suboptimal thing by default and require users to explicitly enable the better behaviour. It may seem to you that adding `default:pythonargs` feature is very easy, but it only is if you know about it in the first place and 99.5% of SWIG users won't.
I also think there is still some misunderstanding about the goal of these changes: they are not so much needed to put default values in Python code (although it is admittedly useful to see them there too), but chiefly to avoid having just `*args` instead of the proper arguments in the signature as this makes Python API much more difficult to use and discover.
You wouldn't advocate generating C code with all functions taking `...` by default and having to do something extra to use proper arguments, would you? So why should we be doing this for Python? Especially as I'm pretty sure that this was never done intentionally but was just result of a bug during Python 3 work.
And yes, any change is risky, and while I do apologize for breaking it once already, I can't reasonably promise that this is never going to happen again. But IMO the (small) risk is worth the big improvement.

wsfulton:
Okay, so I may well be misunderstanding the goal of the changes as I thought part of it was to add default parameters into the Python code in the process of removing *args. I can now see that these default args have actually been generated by default since version 3.0.37 (when PY3 was introduced). So if this is the case there is a lot less change to manage than I thought. Bizarrely, it seems that previously non-default args were previously using *args and default args were not using *args. 

What I don't really understand, and need to, is why so many regressions have been introduced? It seems it is firstly turning on enum default args, which was a bit too ambitious. Secondly, the bug fix in the loop means a lot more default args are now being generated. Is that the sum of it?

BTW @vadz, I've merged in your fixes now.

vadz:
Just to explain: my initial goal was just to make individual arguments appear in the generated Python code. After fixing this bug it turned out that most of the functions in our API still used `*args` because they had enum parameters with default values, so I tried to fix this as well.

The bugs appeared because of two unrelated (although affecting the same code) issues: first, the code generating the default values in Python had been already broken before I touched it, but as it was never used because of the other bug, nobody was affected by it before. And, second, my code for dealing with enums didn't take into account the differences in scope lookup rules between C++ and Python, so I also introduced a new bug and, again, I'm sorry about this.

Anyhow, thanks for merging the changes, I'll try to return to the issue of enum parameters with default values later, and if you have any hints about how to handle this (i.e. whether SWIG already has something that would allow me to generate the correct Python expression for the C++ enum element), they would be very, very welcome.

wsfulton:
Look at Language::enumLookup and how it is used in JAVA::getEnumName. It should be tied in with the language symbol tables (-debug-lsymbols option). Remember to consider enums in nested classes and c++11 enums. #251 may also be of interest.

ojwb:
Adding a new feature to address the regression doesn't seem ideal to me - it's more code to maintain and to have to fix bugs in, and it makes SWIG harder for users to get to grips with.  Once we add a feature, it's hard to remove it again because people will start to use it and rely on it.

I appreciate we want to get a fix for this out quickly, as it's affecting multiple users, but why not make the quick fix conservatively use `*args` except in cases where we're really confident that the defaults are all OK (or even just always use `*args` again for now)?

Then we can consider how best to address this in a more leisurely way rather than pushing out something we may later regret.

As a data point, the PHP backend also makes decisions about implementing default parameters in the target language vs in C++.  I implemented this when I added PHP 5 OO support in 2007, and there's not been a single user bug report about itreported bug in it.

wsfulton:
As default args were previously generated the risk of breaking new is a lot lower than I previously thought, so I'm going to drop the python:defaultargs feature. It does mean that we have to be really conservative in adding improved support for processing default arguments.

@ojbw Using *args all the time when wrapping c++ default args is precisely what @vadz is trying to avoid. So although generating default args sometimes worked previously, it often used *args when it should not have. So turning it on may break existing code, but I've had a good look at the current implementation and it looks fairly good (and has fixed previously problems like use of LL as a suffix. Is there any opportunity for using common code for Python and PHP in this area?

ojwb:
I realise the motivation for vafz's change is avoiding using `*args` as much.  However, the quickest and safest fix would be to revert the problematic change - indeed a temporary step backwards for what vadz is trying to do, but then reverting to using `*args` by default with an new option isn't much better from that point of view.

But don't get me wrong - I'm certainly happy to keep the change with the fixes too.  I just felt that quickly adding a new option here so we could get a fix out promptly was something we'd end up regretting.

I suspect having common code for this would actually be less clear and less maintainable - PHP and Python have different quirks as to exactly what number, strng, etc formats are, so you'd end up with a highly parameterised common function, and having to test any change with both backends (making it harder for anyone with knowledge of just one language to improve).

There may be scope for common support functions - e.g. I don't currently try to do anything with enums, but a function to give the correctly qualified name for an enum in the target language would make supporting those much easier.  But perhaps that's too specific to the target language.

PR #303, opened 2015-01-11
[lua/luarun] change return type from int to void on functions not return...

dertom95:
...ing anything

Using emscripten compiler to crosscompile to javascript lead to "traps" on this functions which results in a crash


ojwb:
Thanks.

Issue #304, opened 2015-01-12
[Ruby] Use of Config::CONFIG in configure does not work anymore with Ruby-2.2


pierre-labastie:
In 2006 or so, the Config module was renamed to RbConfig, but there was the instruction "Config=RbConfig" in obsolete.rb (+ a warning that Config was obsolete), until Ruby-2.1. With Ruby-2.2, "obsolete.rb" has been suppressed. Yet, the configure script in SWIG still uses Config at some places, which leads to some bad behavior (specifically, RUBYSO=., instead of .so). Because of that, any test with a run test fails (the shared library module is named example., instead of example.so, and is not found by ruby).
Patching configure.ac is easy, unless one wants to keep compatibility with versions of Ruby before 2006 (I think versions 1.6 and below).

wsfulton:
I'm sure we could detect the version of Ruby and then use RbConfig or Config as appropriate in configure.ac. Isn't that what configure.ac is trying to do anyway to do anyway:
```ruby
        # Try Ruby1.9+ first
        RUBYDIR=`($RUBY -rrbconfig -e 'print RbConfig::CONFIG[["rubyhdrdir"]] || $rubyhdrdir') 2>/dev/null`
```
Supporting 1.6 is a nice to have, but I'm not too concerned if it is no longer detected given how old 1.6 is.

kkaempf:
Yes, backwards compatibility is possible but probably not worth the effort. Also all calls in configure.ac already have "-rrbconfig", ruby 1.6 would need "-rconfig" ;-)

wsfulton:
Okey doke, I'll merge this in then. I can't test it against 2.2, but hopefully one of the newer OpenBuild platforms will have 2.2 and then turn on Ruby 2.2 testing.

wsfulton:
Merged, thanks everyone.

kkaempf:
I've tested it with a local Ruby 2.2 installation. ;-)

PR #305, opened 2015-01-12
Ruby: Replace Config::CONFIG with RbConfig::CONFIG in configure.ac

kkaempf:
The Config namespace was deprecated for a long time and Ruby 2.2
finally removed it. Adapt configure.ac accordingly.

This fixes issue #304

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1508,7 +1508,7 @@ if test -n "$RUBY"; then
         # Try Ruby1.9+ first
         RUBYDIR=`($RUBY -rrbconfig -e 'print RbConfig::CONFIG[["rubyhdrdir"]] || $rubyhdrdir') 2>/dev/null`
 	if test x"$RUBYDIR" = x"" || test x"$RUBYDIR" = x"nil"; then
-		RUBYDIR=`($RUBY -rmkmf -e 'print Config::CONFIG[["archdir"]] || $archdir') 2>/dev/null`
+		RUBYDIR=`($RUBY -rmkmf -e 'print RbConfig::CONFIG[["archdir"]] || $archdir') 2>/dev/null`
         else
                 RUBYARCH=`($RUBY -rrbconfig -e 'print RbConfig::CONFIG[["arch"]] || $arch') 2>/dev/null`
         fi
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -1508,7 +1508,7 @@ if test -n "$RUBY"; then
         # Try Ruby1.9+ first
         RUBYDIR=`($RUBY -rrbconfig -e 'print RbConfig::CONFIG[["rubyhdrdir"]] || $rubyhdrdir') 2>/dev/null`
 	if test x"$RUBYDIR" = x"" || test x"$RUBYDIR" = x"nil"; then
-		RUBYDIR=`($RUBY -rmkmf -e 'print Config::CONFIG[["archdir"]] || $archdir') 2>/dev/null`
+		RUBYDIR=`($RUBY -rmkmf -e 'print RbConfig::CONFIG[["archdir"]] || $archdir') 2>/dev/null`
```
ojwb: I suspect the use of `-rmkmf` here is wrong, but perhaps there's some subtle reason I'm missing.

kkaempf: Right, `-rmkmf` is certainly wrong. I guess it's never reached since `RbConfig::CONFIG[["rubyhdrdir"]]` is set.

ojwb: On 14 January 2015 at 19:49, Klaus Kämpf <notifications@github.com> wrote:

> In configure.ac
> <https://github.com/swig/swig/pull/305#discussion-diff-22919169>:
>
> > @@ -1508,7 +1508,7 @@ if test -n "$RUBY"; then
> >          # Try Ruby1.9+ first
> >          RUBYDIR=`($RUBY -rrbconfig -e 'print RbConfig::CONFIG[["rubyhdrdir"]] || $rubyhdrdir') 2>/dev/null`
> >  	if test x"$RUBYDIR" = x"" || test x"$RUBYDIR" = x"nil"; then
> > -		RUBYDIR=`($RUBY -rmkmf -e 'print Config::CONFIG[["archdir"]] || $archdir') 2>/dev/null`
> > +		RUBYDIR=`($RUBY -rmkmf -e 'print RbConfig::CONFIG[["archdir"]] || $archdir') 2>/dev/null`
>
> Right, -rmkmf is certainly wrong. I guess it's never reached since
> RbConfig::CONFIG[["rubyhdrdir"]] is set.
>
Thanks for confirming (and that's really RbConfig::CONFIG["rubyhdrdir"] -
the extra [] are m4 quoting.

Testing with Ruby 1.9.3, these all work:

ruby -rmkmf -e 'print RbConfig::CONFIG["archdir"]'
ruby -rrbconfig -e 'print RbConfig::CONFIG["archdir"]'
ruby -e 'print RbConfig::CONFIG["archdir"]'

With 1.8.7, the first two work, but the last fails.

It's better to ask for the module we actually want rather than rely on it
being pulled in indirectly, so I've fixed this in 4fed2e6.

PR #308, opened 2015-01-14
Fix #307.

### LINKED ISSUE #307: Invalid code generated for enum class when using directors.
This happens with rel-3.0.3
bla.h:

    enum class Color { Red, Green, Blue };
    struct Callback { virtual void f(Color c) = 0; };

bla.i:

    %module(directors="1") ColorModule
    %{
    #include "bla.h"
    %}

    %feature("director") Callback;
    %include "bla.h"

When  I generate a wrapper using `swig -c++ -csharp bla.i` from these files, the generated code does not compile. Here's the invalid piece of code:

    // ... some code ..
    void SwigDirector_Callback::f(Color c) {
      int jc  ;

      if (!swig_callbackf) {
        throw Swig::DirectorPureVirtualException("Callback::f");
      } else {
        jc = c; // <-- this code fails to compile because Color cannot be implicitly converted to int.
                // it should be jc = static_cast<int>(c); instead.
        swig_callbackf(jc);
      }
    }
    // ... some more code

As a workaround I can use the following typemap in bla.i:

    %typemap(directorin) enum class SWIGTYPE "$input = static_cast<int>($1);"

leg0:


wsfulton:
Thanks for bug report and patch. All merged in now and I've added some similar fixes.

PR #314, opened 2015-01-23
Remove unused assignment

DavidMazary:
The value of *module_head* is not used after this assignment in this branch of the function.


ojwb:
Thanks.

PR #318, opened 2015-01-31
fix pep257 warnings

xantares:
minor python docstrings syntax fixes:
- mono-line module docsstring
- rewrite autodoc parameters section in numpydoc style
- on-line summary should end with "."
- adds a blank line after class docstring


wsfulton:
Did you use a tool to check the style? I think there is a pep257 tool to check this, did you try this out? Is it worth putting this into our testing?

xantares:
Yes, I did use https://github.com/GreenSteam/pep257.
I first integrated it in the make check stage, but then I backed off because I had to disable so much warning types because of the missing docstrings and that it was only useful for the autodoc.i testcase.

wsfulton:
Can you please add to the Python.html documentation this new style that autodoc uses. A comment in the python.cxx about this should also be put in so that future developers know what the style attempted is. A reference to https://github.com/GreenSteam/pep257 would also not go amiss.

xantares:
Ok

ojwb:
@xantares Why did you close this?  It doesn't appear to have been merged yet...

xantares:
it's not clean enough, i'll resubmit

PR #334, opened 2015-02-10
Fix build on S390(x)-architecture

besser82:
A simple patch to `configure.ac` for properly adding `-fPIC`-flag on S390(x)-arches.


wsfulton:
Presumably you've tested this on both architectures and neither -fpic nor -fPIC will work on both architectures?

besser82:
@wsfulton: Yes, builds were run on both (different) arches and that patch is the only working solution.

PR #324, opened 2015-02-04
Avoid collisions with the 'check' word on apple clang

lisitsyn:
When compiling [Shogun](https://github.com/shogun-toolbox/shogun) with swig 3.0.5 on Mac OS X Yosemite with `clang -v`:

```
Apple LLVM version 6.0 (clang-600.0.56) (based on LLVM 3.5svn)
Target: x86_64-apple-darwin14.0.0
Thread model: posix
```
a few compilation errors occur: 

```
.../modshogunPYTHON_wrap.cxx:7682:38: error: expected member name or ';' after declaration specifiers
    static bool check(PyObject *obj) {
```

After some research I found that it is due to the word 'check' used here that interferes with something else (apparently no symbol named 'check' is in the scope). Changing names of static member functions that check pyobjects helps to solve the problem and compile the wrapper. Maybe some simpler reproducing example is possible - please tell me whether it is needed.

PR #335, opened 2015-02-10
Deal with Apple's deprecated 'AssertMacros.h' from Carbon-framework

besser82:
As suggested by @ojwb in https://github.com/swig/swig/pull/324#issuecomment-73623121.


ojwb:
Presumably this change does actually fix your problem?

besser82:
@ojwb: Yes, that fix does do the trick!

PR #336, opened 2015-02-10
Allow configuration of behaviour when there is a Director Error in Python

blaisegassend:
This changeset allows directors to be used in Python without using C++ exceptions by allowing the user to indicate which action to take when a DirectorException is raised.

I am not entirely satisfied with my use of %insert to describe the action to take on exception and am open to comments on how to do it better. Using %feature would be nice, and I will happily make that change if somebody points me to an example of a %feature that only applies globally. Alternatively, using a #define similar to SWIG_DIRECTOR_UEH would also work for me. (Except that that macro would expand to whatever error handling code you want rather than just be a flag.)

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -3078,6 +3078,27 @@ Swig::DirectorMethodException is thrown, Python will register the
 exception as soon as the C wrapper function returns.
 </p>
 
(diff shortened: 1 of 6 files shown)


ojwb:
Closing - #349 supercedes this.

PR #352, opened 2015-03-10
Added some information on manually compiling for R

ibell:
This was the root of my problem, and it would have been good to have this clearly explained before.


ojwb:
Thanks.

PR #357, opened 2015-03-13
Change %pythoncode {...} to %pythoncode %{...%}

r-owen:
Update .i files to use %pythoncode %{...%} instead of %pythoncode {...}

### DIFF
--- a/Lib/python/pyabc.i
+++ b/Lib/python/pyabc.i
@@ -1,7 +1,9 @@
 %define %pythonabc(Type, Abc)
   %feature("python:abc", #Abc) Type;
 %enddef
-%pythoncode {import collections};
+%pythoncode %{
+import collections
+%}
 %pythonabc(std::vector, collections.MutableSequence);
 %pythonabc(std::list, collections.MutableSequence);
 %pythonabc(std::map, collections.MutableMapping);
--- a/Lib/python/std_map.i
+++ b/Lib/python/std_map.i
@@ -176,14 +176,16 @@
 
 #else
   %extend {
-    %pythoncode {def __iter__(self):
-    return self.key_iterator()}    
-    %pythoncode {def iterkeys(self):
-    return self.key_iterator()}
-    %pythoncode {def itervalues(self):
-    return self.value_iterator()}
-    %pythoncode {def iteritems(self):
-    return self.iterator()}
+    %pythoncode %{
+      def __iter__(self):
+        return self.key_iterator()
+      def iterkeys(self):
+        return self.key_iterator()
+      def itervalues(self):
+        return self.value_iterator()
+      def iteritems(self):
+        return self.iterator()
+    %}
   }
 #endif
 
--- a/Lib/python/std_map.i
+++ b/Lib/python/std_map.i
@@ -177,14 +177,14 @@
 #else
   %extend {
     %pythoncode %{
-      def __iter__(self):
-        return self.key_iterator()
-      def iterkeys(self):
-        return self.key_iterator()
-      def itervalues(self):
-        return self.value_iterator()
-      def iteritems(self):
-        return self.iterator()
+def __iter__(self):
+    return self.key_iterator()
+def iterkeys(self):
+    return self.key_iterator()
+def itervalues(self):
+    return self.value_iterator()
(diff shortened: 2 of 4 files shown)


r-owen:
I pushed some fixes, so please make sure you have them before attempting to merge. Also, I build the test suite but don't have all optional packages my testing is not complete.

Finally, I am concerned about indentation of the %pythoncode blocks in swigged interfaces, since I've been running into some indentation issues in SWIG 3.0.2 after switching my own code from %pythoncode {...} to %pythoncode %{...%}. I hope the built in tests will show any such problems. Unindenting the python code wrapped in %pythoncod %{...%} to the left margin should fix any errors, if they do show up.

wsfulton:
Have you seen the Travis builds break with this patch? This will need fixing before consideration is given to the patch.

r-owen:
Thanks for the heads up. I had not seen the broken build. I will outdent the python (in fact I tried once but messed up one file, so am trying again).

ojwb:
@r-owen: Still failing.

I'd suggest trying keeping each `%pythoncode` directive on one line as before, and just changing the delimiters, so the first case would be:

```
%pythoncode %{import collections%}
```

r-owen:
@ojwb thanks for the suggestion. You are probably right, but I feel it is important for readability to put the python code on its own lines. The old code was extremely difficult to read. If SWIG really cannot handle that then I feel it is a misfeature or bug of SWIG.

If I am reading the logs correctly it appears that the failure is due to a PEP8 warning about extra whitespace. If so, one option is to suppress that warning or leave it as a warning instead of promoting it into an error. If there is a way to examine the python wrapper that code that the test is complaining about, that would be interesting.

ojwb:
I suspect you'll end up in a world of pain if you try to change this behaviour - because Python cares about indentiation, changing how this is handled is bound to break some interface files which currently work.

I think it's worth a little ugliness in the SWIG source in order to generate PEP8 compliant output.  PEP8 is widely accepted in the Python community as how code ought to be indented, and while the code generated by SWIG isn't intended to be read by humans, it is unhelpful to generate code which causes warnings when people run tools which check for PEP8 conformence over a source tree containing both SWIG-generated and hand-written Python code (it's the same reason we try to fix compiler warnings in the generated C/C++ code, even if they don't indicate a genuine problem).  That's the reason that the automatic PEP8 check in SWIG's testsuite was added.

I don't know if you can get travis to show you files from the failed build, but you can examine the wrapper code generated if you run the tests locally.

r-owen:
I am pretty sure that the Travls CI build is not failing due to indentation. As you know, incorrect indentation causes much more serious problems than PEP8 warnings; it is often a syntax error, and if not, then can lead to incorrect code.

The current code matches SWIG documentation (e.g. see http://www.swig.org/Doc3.0/Python.html#Python_nn42) and produces maximally readable python blocks (other than the need for unfortunate outdenting to the left margin, which is probably a bug but as you say might not be safe to fix).

The PEP8 tester is warning about extra about whitespace, which I doubt is indentation. I suspect it means that SWIG is adding blank lines that contain with spaces (likely before and/or after the code blocks), though I'd have to see the generated interface file to know for sure. There is no need need for SWIG to insert blank lines with whitespace, nor trailing whitespace for that matter, because neither of them in meaningful to the Python interpreter. So it would be safe to change SWIG to not do that.

In my opinion this patch is correct as it stands and I'm not willing to garbage it up to eliminate a nearly meaningless PEP8 warning. I think the problem should be fixed in SWIG or in the tester. However, you are welcome to adapt this patch if you disagree.

ojwb:
I've just changed `{` `}` to `%{` `%}` in c8a7322.  I verified the output was the same using `make partialcheck-python-test-suite` before and after and diffing `Examples/test-suite/python` from each.

If you want to work on changes beyond this, that's fine, but they need to not trigger additional warnings from the pep8 tool (and without simply turning off warnings).  The warning is not innocuous, because it is useless noise to users running the pep8 tool over a tree of Python scripts which include scripts generated by SWIG.

Otherwise we can close this PR now.

ojwb:
Closing now - any follow on is probably better dealt with in a fresh PR.

PR #367, opened 2015-03-26
When warning about multiple inheritance in C#, say C# in the message

jamesathey:
instead of Java


ojwb:
Thanks.

I did a quick check for other instances, and discovered this in `Source/Modules/csharp.cxx`:

```
        if (GetFlag(n, "feature:javadowncast")) {
```

@wsfulton Should that be changed to `csdowncast` or `cs:downcast` or something like that?

There was also a reference to `csbase` in `java.cxx` in an error message, which I've pushed 2c08e33 to fix.

wsfulton:
It isn't documented, but you could change it to csdowncast, but don't record it in the CHANGES file unless some documentation about it is put in place first. I'm not sure if it even works and it isn't tested.

ojwb:
OK, changed that in 084f724.

Issue #256, opened 2014-10-31
swig3.x python generated code segfaults


vigsterkr:
We are currently experiencing some problems with the swig3.x generated python interface. Namely, when a wrapped object's destructor, i.e `_wrap_delete_<object>` is being called we are getting a segfault, see my last 3 comments in shogun-toolbox/shogun#2313

Basically when we generate the swig interface with `-modern -modernargs` arguments due to commit 8998f11ca39d60ae4593a3fab2fc86ab92439003 the following line is being added, to every function that is being wrapped in case of swig3.x:
```
if(!PyArg_UnpackTuple(args,(char *)"delete_CSVFile",0,0)) SWIG_fail;
```
if `args` is actually `NULL` this call will cause a `SEGFAULT`. this does not happen in swig2.x...

besser82:
@wsfulton: It looks like it is related to `-modernargs`.  Without that switch or explitly specifying `-nomodernargs` segfaults are gone.

ptomulik:
Hi, some time ago I was working on #250 and have seen several segfaults during test suite run with ``-builtin -modernags``. I remember it was this specific combination of flags causing problems. Maybe it's the same issue?

besser82:
Thanks for your remark…  Lemme try adding `-fastunpack`…  Let's see what the outcome is…

besser82:
So…  I just experimented using ´-fastunpack´ with the following outcome:

  * Using `-builtin -modernags` results in segfaulting as described by @vigsterkr.
  * Using `-builtin -nomodernags` results in working binaries.
  * Using `-builtin -modernags -fastunpack` results in working binaries, too.

@wsfulton:  Any suggestions / ideas about how to fix that inside of SWIG?

besser82:
@ptomulik can you try to recompile swig with my patch from #372 and see if `-builtin -modernags` still crashes your testsuite for #250?

ptomulik:
Sorry, but I'm short of time and can't do it ATM.  Please ping me back in a week or two if you'll still  need assistance.

wsfulton:
Is this the normal SWIG test-suite that is being referred to? If so, I can run it with -builtin -modernargs with this patch and report back.

besser82:
Closed https://github.com/swig/swig/pull/372, because branch-name changed.  See the difference in TravisCI-Python-tests between:
* no-changes to code-base:  https://github.com/swig/swig/pull/381
* one-line change to code-base:  https://github.com/swig/swig/pull/382

[...]

PR #372, opened 2015-04-03
fix using `-builtin -modern -modernargs` results in segfaulting code (#256)

besser82:
this small change fixes all segfaults as reported in issue #256

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,17 @@ matrix:
     - compiler: gcc
       env: SWIGLANG=python SWIG_FEATURES=-builtin
     - compiler: gcc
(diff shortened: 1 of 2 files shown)


besser82:
This has been removed in https://github.com/swig/swig/commit/8998f11ca39d60ae4593a3fab2fc86ab92439003.  It was introduced with changes in swig-2.0.12, which produced flawlessly working Python-wrappers with that special combination of arguments.  The remove has been before the swig-3.0.0 release, where all trouble with the python-wrappers begun.

besser82:
Closing this one.  New PR, because of changed branch-name, is https://github.com/swig/swig/pull/382.

PR #373, opened 2015-04-03
Windows CI testing using Appveyor

wsfulton:
This adds CI testing on Windows using Appveyor - http://appveyor.com.
Appveyor have free accounts for open source projects.

The support is done using the normal autotools on Cygwin but using
the Microsoft Visual C++ compiler for compiling SWIG and running
the tests. C#, Java, Python 2 and Python 3 currently working.
The full test-suite takes too long to run though, so
make partialcheck-test-suite is run instead.

The support for Visual C++ requires the cccl script which is now
available at https://github.com/swig/cccl. The appveyor.yml script
downloads cccl-1.0 from Github.


wsfulton:
Merged

Tools/config/ax_boost_base.m4
```diff
@@ -92,8 +92,11 @@ if test "x$want_boost" = "xyes"; then
     libsubdirs="lib"
     ax_arch=`uname -m`
     case $ax_arch in
-      x86_64|ppc64|s390x|sparc64|aarch64)
-        libsubdirs="lib64 lib lib64"
+      x86_64)
+        libsubdirs="lib64 libx32 lib lib64"
+        ;;
+      ppc64|s390x|sparc64|aarch64|ppc64le)
+        libsubdirs="lib64 lib lib64 ppc64le"
```
ojwb: That addition of `ppc64le` to `libsubdirs` looks rather suspect...

(And unrelated to your change, why's `lib64` in there twice?)

wsfulton:
I just updated to the latest from the autoconf-archive (after contributing my windows fix). See history at https://github.com/peti/autoconf-archive/commits/master/m4/ax_boost_base.m4. Probably best to liaise with the authors there.

ojwb:
I've attempted to report it there.

ojwb:
Hmm, is there no way to hide the config file for this service?  At least `.travis.yml` is hidden by default.

vadz:
You can configure another file to use instead of `appveyor.yml` in their web interface but I don't think anybody but @wsfulton has access to it.

PR #381, opened 2015-04-10
added TravisCI-Python-tests for using -builtin with -modern and/or -modernargs

besser82:
_Don't merge this one, please._  The full changeset is in https://github.com/swig/swig/pull/382.

The only purpose of this is showing the difference in the testsuite.

PR #382, opened 2015-04-10
fix using `-builtin -modern -modernargs` results in segfaulting code (#256)

besser82:
* this small change fixes all segfaults as reported in issue #256
* This has been removed in 8998f11. It was introduced with changes in swig-2.0.12, which produced flawlessly working Python-wrappers with that special combination of arguments. The remove has been before the swig-3.0.0 release, where all trouble with the python-wrappers begun.

_Close https://github.com/swig/swig/pull/381 after merging this one, please._

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,67 +7,79 @@ env:
 matrix:
   include:
     - compiler: gcc
-      env: SWIGLANG=csharp
+      env: SWIGLANG="csharp"
     - compiler: gcc
-      env: SWIGLANG=d
+      env: SWIGLANG="d"
     - compiler: gcc
-      env: SWIGLANG=go
+      env: SWIGLANG="go"
     - compiler: gcc
-      env: SWIGLANG=guile
+      env: SWIGLANG="guile"
     - compiler: gcc
-      env: SWIGLANG=java
+      env: SWIGLANG="java"
     - compiler: gcc
-      env: SWIGLANG=javascript ENGINE=node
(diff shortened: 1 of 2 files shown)


besser82:
Closing this one…  New PR is https://github.com/swig/swig/pull/382

wsfulton:
@besser82, I can see that the segfaults in #381 are fixed by this patch which is a step forward. The additional builds you've added still have a few other failures meaning that they fail overall. Is that something you are going to help fix and if so, will that be in another patch or added onto this patch? I'm just trying to work out how to merge these in.

wsfulton:
@besser82, given no response, I'll presume any additional fixes will be in another patch and I'll now merge this one fix into master.

wsfulton:
This patch request needs a rethink. It reverts 8998f11 and also see a3a0535 which was added to test the former.
```
checking python testcase global_functions (with run test)
  File "global_functions_runme.py", line 16, in <module>
    raise RuntimeError("argument count check failed")
RuntimeError: argument count check failed
```
This occurs for both:
```
env SWIG_FEATURES="-modernargs" make global_functions.ctest -s
env SWIG_FEATURES="-builtin -modernargs" make global_functions.ctest -s
```
The patch removes the generation of the following line in _wrap_global_void, thus removing the argument count checking
```
   if(!PyArg_UnpackTuple(args,(char *)"global_void",0,0)) SWIG_fail;
```

Issue #376, opened 2015-04-06
Python type error with typedef'd pointer types and NULL default argument


jspreston:
When wrapping a typedef'd pointer type such as MyT in the following header:
```
typedef void *MyT;

void f3(int n,MyT t=NULL);

class MyClass{
public:
    void classprint2(int n,void *t=NULL);
    void classprint3(int n,MyT t=NULL);
};
```

the function `classprint3` is not callable without specifying an argument for `t`, it will give the following error:

```
Traceback (most recent call last):
  File "test_hello.py", line 24, in <module>
    hc.classprint3(2)
  File "/home/sci/jsam/code/Projects/Utah/tmp/swigtest/hello.py", line 119, in classprint3
    return _hello.MyClass_classprint3(self, n, t)
NotImplementedError: Wrong number or type of arguments for overloaded function 'MyClass_classprint3'.
  Possible C/C++ prototypes are:
    MyClass::classprint3(int,MyT)
    MyClass::classprint3(int)
```
This error does not occur with version 3.0.2, but does with 3.0.5 and HEAD (g6c76c7b  as of my last test). It seems to only come up with typedef'd pointer types in class member functions, I've created a full minimal example here:

https://gist.github.com/jspreston/cf702e2dae00c1c3717c

vadz:
Should be fixed by 0eae8a8efaa5ce1d5c918542d34a23afd2f92c9e

PR #388, opened 2015-04-22
Fix handling of NULL default argument values for pointer types.

### LINKED ISSUE #365: Python converts NULL default argument to 0 instead of None
https://github.com/Openwsman/openwsman/issues/59 raises an issue with default arguments in Python.

A function definition of

     WsXmlDocH invoke( client_opt_t *options, const char *resource_uri, const char *method, WsXmlDocH data = NULL) {

gets converted to the following Python code

     def invoke(self, options, resource_uri, method, data=0):

instead of

     def invoke(self, options, resource_uri, method, data=None):

vadz:
Accept not only manifest pointer types (such as e.g. "void *") but also types
that are typedefs for pointer types when checking whether C++ value of 0 must
be represented as 0 or None in Python.

Closes #365, #376.

### DIFF
--- a/Examples/test-suite/default_args.i
+++ b/Examples/test-suite/default_args.i
@@ -115,6 +115,7 @@
 %rename(renamed1arg) Foo::renameme() const;
 
 %inline %{
+  typedef void* MyHandle;
 
   // Define a class
   class Foo {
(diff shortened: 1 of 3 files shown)


vadz:
Travis errors are all bogus (two are due to a completely unrelated failure in CCache tests and two are just due to the build running too slowly), so if there are no objections I'm going to merge this soon in spite of them.

PR #390, opened 2015-04-23
Python examples pep8 cleanup

yazug:
Doing a bit of pep8 cleanup on the python examples files

### DIFF
--- a/Examples/python/callback/runme.py
+++ b/Examples/python/callback/runme.py
@@ -2,14 +2,15 @@
 
 # This file illustrates the cross language polymorphism using directors.
 
-import example 
+import example
 
 
 class PyCallback(example.Callback):
-	def __init__(self):
-		example.Callback.__init__(self)
-	def run(self):
-		print "PyCallback.run()"
+    def __init__(self):
+        example.Callback.__init__(self)
+
+    def run(self):
+        print "PyCallback.run()"
 
 # Create an Caller instance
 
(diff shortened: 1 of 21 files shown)


wsfulton:
How about adding pep8 checks to the examples like we have in the test-suite to prevent the examples regressing in this respect in the future?
I think all the runme.py changes should be squashed into one commit.

yazug:
Should I remaster the branch and squash the fixup in one commit for all the runme.py scripts? and push as a rebase to this branch or how's the best way to go about this?

yazug:
Just looked at the pep8 construct in the test-suites for python and are you thinking of integrating into each Makefile or is there an easier place to add that construct?

xantares:
I suggest we just autopep8 the example *.py source files, in one single commit :]

yazug:
I'll make another request

yazug:
sounds good for autoprep8 I hadn't see that before.

wsfulton:
Yes a rebase is what is needed. 
I presume autoprep8 will make the same changes you have already done, or at least something similar.
I'd put the pep8 checks into Examples/Makefile.in as a make target (say python_pep8) and then call that in each individual makefile in the check target before python_run. It would be nice if Examples/test-suite/Makefile.in also called the same target in Examples/Makefile.in.
Did you mean to close this issue?

yazug:
I'm working on the changes not sure if I did the wrong work flow for git but new branch will be up shortly and I'll see if I can get the pep8 checks in if I can figure it out.

wsfulton:
sure no probs, that explains why you closed the issue.

PR #391, opened 2015-04-23
Add some generated example files to the .gitignore list

yazug:
After running make -k check I was left with a bunch of generated files showing up as untracked files.  Added some rules to .gitignore to deal with them for python and perl5 to reduce how many showed up

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -187,6 +187,6 @@ Examples/python/import_template/spam.py
 # Scilab generated files
 loader.sce
 
-# Pearl Examples generated files
+# Perl5 Examples generated files
 Examples/test-suite/perl5/*.pm
 Examples/test-suite/perl5/*.h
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -157,7 +157,36 @@ Examples/test-suite/octave/*.oct
...
+Examples/python/import_packages/same_modnames2/pkg1/pkg2/foo.py
+Examples/python/import_template/bar.py
+Examples/python/import_template/base.py
+Examples/python/import_template/foo.py
+Examples/python/import_template/spam.py
 
 # Scilab generated files
 loader.sce
 
+# Pearl Examples generated files
```
ojwb: "Perl"

yazug: Good catch, my bad spelling
fixed.

yazug:
I like the clean up you did in addition to this pull request.

wsfulton:
Thanks, the only downside is it requires git 1.8 or later for the ** patterns to work.

PR #392, opened 2015-04-24
lua: push integer constants as integer

alexwarg:
This allows better compatibility with Lua 5.3. Otherwise function
overloading assuming integer parameters might not work.


wsfulton:
I presume this is closed as superceded by #398 ?

alexwarg:
Yes.

PR #396, opened 2015-04-25
Python pep8 cleanup

yazug:


yazug:
This added python_check to all of the python Examples
also does autopep8 cleanup for most of the python code under the Examples folder

yazug:
The test-suite python scripts only have the following warnings

2 - E402 module level import not at top of file
50 - E501 line too long (xxx > 79 characters)
428 - W602 deprecated form of raising exception

yazug:
was able to run make check it got through all the examples for python clean but started failing when it got to the test-suite/python for cpp_check run of pep8 on the generated python files.

yazug:
Looking into the python3 failures now.

yazug:
This Branch ended up being a lot more work than I originally imagined.  But now I can see where the coverage is to propose a few more changes for the python code generator for white space clean up.

wsfulton:
Are you saying this pull request is ready now and you'll do more clean up in another patch?

yazug:
Yes This Pull Request is ready and any other changes will be separate efforts

wsfulton:
Can you keep the following in numeric order and alphabetical order too, so the W numbers are separate to E numbers...
```patch
+PEP8_FLAGS   = --ignore=E402,E501,W602,E731,W391,E301,E302,E303,W602,W291
```
Are you using a later version of pep8 to warrant putting the additional --ignore above as we don't see these using 1.5.7 on Travis?

For the Travis testing, we probably need to update the version we use, see .travis.yml file:
```
  - if test "$SWIGLANG" = "python"; then git clone https://github.com/jcrocholl/pep8.git && pushd pep8 && git checkout tags/1.5.7 && python ./setup.py build && sudo python ./setup.py install && popd; fi

```
I'm a bit concerned as to how many simple whitespace changes are needed (besides the leading indentation). Are these absolutely necessary? I don't mind a one off cleanup of these whitespaces, but expecting new testcases to conform is a bit too pedantic and likely to break builds with simple patch requests.

yazug:
Flags sorted and ordered.

I normally use pip install pep8 but I did install the version referenced above.  I'll have to go back and look at what it was complaining about again.

The whitespace cleanup is a lot because it had accumulated a lot of technical debt.  From a developer who uses python on a regular basis have consistent and clean formatting of code (pep8 lays out an expectation of what good formatting is for python) make it much easier to read and find code that actually has issues in it.

https://www.python.org/dev/peps/pep-0008/

If there are pieces of pep8 that you don't care about for the test-suite runners, generated code and the examples then a bit of reading through that page to identify things you don't care about for SWIG should be done and noted where the ignore list is declared.

if you are ok with the overall cleanup of whitespace and pep8 cleanup but don't care about maintaining that pristine level of compliance then the set of ignored rules for pep8 should be adjusted to minimize impact.

yazug:
after testing a bit more

When I am using pep8 version 1.6.2 I see 
"W402 module level import not at top of file" in several Example and test suite files

yazug:
Question, does this branch have too many changes to be accepted as is?

If it does have too many changes in the one branch I have gotten a subset of changes that are only autopep8 cleanup whitespace cleanup.

yazug:
This branch had too many changes all in one branch.  Look for another pull request with simple cleanup in it, along with the pep8 check for the python Examples code.

wsfulton:
Yes, there were too many whitespace changes in the test-suite, which isn't clear if they are needed or not. I think the next pull request should leave the test-suite alone and just change the examples and use a common set of pep8 flags as currently used by the test-suite. A separate pull request could change the test-suite whitespace / adjust pep8 flags.

yazug:
That will be formed up shortly. Thank you for the feedback.

PR #398, opened 2015-04-27
Fix lua __getitem + inheritance

alexwarg:
The new handling of classes in Lua (not merging methods into the derived classes) breaks for classes
that provide a __getitem function. The __getitem function prevents method calls to any method defined in a base class. This fix calls __getitem only if the member is not found using  recursive lookup.

### DIFF
--- a/Examples/test-suite/lua/Makefile.in
+++ b/Examples/test-suite/lua/Makefile.in
@@ -14,6 +14,7 @@ top_builddir = @top_builddir@
 
 CPP_TEST_CASES += \
 	lua_no_module_global \
+	lua_inherit_getitem  \
 
 
 C_TEST_CASES += \
--- a/Examples/test-suite/lua/lua_inherit_getitem_runme.lua
+++ b/Examples/test-suite/lua/lua_inherit_getitem_runme.lua
@@ -1,0 +1,12 @@
+require("import")	-- the import fn
+import("lua_inherit_getitem")	-- import lib
+
+local t = lua_inherit_getitem;
+local base = t.CBase()
+local derived = t.CDerived()
+
+assert(base.Foo ~= nil)
+assert(base:Foo() == "CBase::Foo")
+assert(derived.Foo == base.Foo)
+assert(derived:Foo() == "CBase::Foo")
+
--- a/Examples/test-suite/lua_inherit_getitem.i
+++ b/Examples/test-suite/lua_inherit_getitem.i
@@ -1,0 +1,19 @@
+%module lua_inherit_getitem
+
+%inline %{
+
+class CBase {
+public:
+  const char* Foo(void) {
+    return "CBase::Foo";
+  }
+};
+
+class CDerived : public CBase {
+public:
+  void *__getitem(const char *name) const {
+    return NULL;
+  }
+};
+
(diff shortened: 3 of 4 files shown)


wsfulton:
Unless some other Lua expert has a look at this (ping @v-for-vandal) , we'll need more details to consider merging this in, due to lack of Lua knowledge. 

What do you mean by 'new handling off classes in Lua'. Is this in a new version of Lua or the changes in SWIG-3.0.0?
We need a test case that failed before the change as part of the patch.

ojwb:
@wsfulton - it was a SWIG 3.0.0 change - `-squash-bases` gives the old-style handling:

```
     -squash-bases   - Squashes symbols from all inheritance tree of a given class
                       into itself. Emulates pre-SWIG3.0 inheritance. Insignificantly
                       speeds things up, but increases memory consumption.
```

The `lua_pushinteger()` change looks sensible, though unrelated to the rest.  I don't really know enough to usefully comment on the rest of it.

v-for-vandal:
__getitem change seems ok to me.
Useful, if somewhat incorrect analogy with Python:
  before suggested commit __getitem was more like __getattr__
  after suggested commit __getitem is more like __getattribute__

IMHO, the suggested behavior is more useful for SWIG users than the former
one.



On 28 April 2015 at 23:20, Olly Betts <notifications@github.com> wrote:

> @wsfulton <https://github.com/wsfulton> - it was a SWIG 3.0.0 change -
> -squash-bases gives the old-style handling:
>
>      -squash-bases   - Squashes symbols from all inheritance tree of a given class
>                        into itself. Emulates pre-SWIG3.0 inheritance. Insignificantly
>                        speeds things up, but increases memory consumption.
>
> The lua_pushinteger() change looks sensible, though unrelated to the
> rest. I don't really know enough to usefully comment on the rest of it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/swig/swig/pull/398#issuecomment-97196059>.
>



-- 
Sincerely yours,
Artem Serebriyskiy

alexwarg:
Yes the  lua_pushinteger() change is unrelated to the __getitem change. This change is related to changed behavior in Lua 5.3 that now supports integers internally, and in my opinion swig should use the integer types of Lua 5.3 for integer constants. This change should not affect Lua < 5.3 where the lua_pushinteger() API stores the value as a Lua number. AFAIK.

I can provide separate changes for these to changes if desired.

ojwb:
Ah, I see the `lua_pushinteger()` change is a separate commit, so I've merged that one.

As William says, a regression testcase for the `__getitem` change would be good to have, so we can have confidence it's fixed and will remain fixed.

alexwarg:
ok, I'll put a testcase in place

alexwarg:
Added a test case that fails without the patch, but works with the -squash-bases option and swig-2.x

alexwarg:
Hm, there was a timeout during the build test, I've no idea how to cope with this?

alexwarg:
I've updated to the current master and included a test case, please consider merging the change.
Thanks.

wsfulton:
No-one is maintaining Lua, but given the positive comments, I'll merge this.

PR #400, opened 2015-04-27
Fix handling of default arguments after ignored ones in Python.

### LINKED ISSUE #377: Default argument handling for Python
(Excerpt from test.i)
...
#include "test.h"

%typemap( default ) double* param_opt
{
   $1 = NULL;
}

%rename( funcid ) Test::func( long** long_list_out, float min_val, float max_val,
										  long min_size, long max_size, long delta, 
										  double min_stab, long stable_dist,
										  long stable_mode, long count,
										  double* param_opt, long param_len_opt );

class Test
{
	public:
		Test( long img_id, long neigh, double thres_change );
		long func( long** long_list_out, float min_val, float max_val,
							  long min_size, long max_size, long delta, 
							  double min_stab, long stable_dist,
							  long stable_mode, long count,
							  double* param_opt, long param_len_opt );
};
...

( test.h declares:
  class Test
  {
	public:

		Test( long img_id, long neigh = 8, double thres_change = 3.0 );
		long func( long** cont_ids, float min_val, float max_val,
							  long min_size, long max_size, long delta,
							  double min_stab, long stable_dist = 0,
							  long stable_mode = 0, long count = 0,
							  double* param = NULL, long param_len = 0 );
  }
)

creates test.py by SWIG 3.0.5 on Windows (swigwin-3.0.5 installer):
...
class Test(_object):
    __swig_setmethods__ = {}
    __setattr__ = lambda self, name, value: _swig_setattr(self, Test, name, value)
    __swig_getmethods__ = {}
    __getattr__ = lambda self, name: _swig_getattr(self, Test, name)
    __repr__ = _swig_repr

    def __init__(self, img_id, neigh, thres_change):
        this = _icono.new_Test(img_id, neigh, thres_change)
        try:
            self.this.append(this)
        except:
            self.this = this

    def funcid(self, min_val, max_val, min_size, max_size, delta, min_stab, stable_dist, stable_mode, count, param_opt):
        return _test.Test_funcid(self, min_val, max_val, min_size, max_size, delta, min_stab, stable_dist, stable_mode, count, param_opt)
    __swig_destroy__ = _test.delete_Test
    __del__ = lambda self: None
Test_swigregister = _test.Test_swigregister
Test_swigregister(Test)
...

And complains in Python (too few arguments) when calling test.Test().func() with less than 10 parameters.

Matt 

PS: Works for SWIG 2.0.9

vadz:
Don't skip checking subsequent arguments just because one of them has "in"
typemap with numinputs=0 attribute.

Add a unit test showing the problem which is relatively rare as it doesn't
happen for the class methods and is hidden unless autodoc feature is used for
the global functions.

Closes #377.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -5,6 +5,9 @@ See the RELEASENOTES file for a summary of changes in each release.
 Version 3.0.6 (in progress)
 ===========================
 
+2015-04-27: vadz
+            [Python] Fix "default" typemap used after an argument with "numinputs=0" (#377).
+
 2015-04-24: wsfulton
             [Python] Fix #256. Code generated with '-builtin -modernargs' segfaults for any
             method taking zero arguments.
--- a/Examples/test-suite/default_args.i
+++ b/Examples/test-suite/default_args.i
@@ -278,3 +278,17 @@ struct ConstMethods {
     } Pointf;
   }
 %}
+
+// Default arguments after ignored ones.
+%typemap(in, numinputs=0) int square_error { $1 = 2; };
+%typemap(default, noblock=1) int def17 { $1 = 17; };
+
+// Enabling autodoc feature has a side effect of disabling the generation of
+// aliases for functions that can hide problems with default arguments at
+// Python level.
+%feature("autodoc","0") slightly_off_square;
+
+%inline %{
+  inline int slightly_off_square(int square_error, int def17) { return def17*def17 + square_error; }
+%}
+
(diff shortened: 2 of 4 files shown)


vadz:
There is still something wrong here, I don't understand why doesn't the problem happen for the class methods: debugging I see that "in" typemaps are simply not present for the methods in `is_representable_as_pyargs()` (but are present for the functions) but I have no idea why, all I can say is that calling `Swig_typemap_attach_parms("in")` doesn't help.

But I do think this change is a step in the right direction. I'm just almost certain it's not the _last_ step...

vadz:
I'm going to commit this because CI failures are bogus and this does fix a regression.

Issue #402, opened 2015-05-01
feature("director") and feature("ref") don't work together


LindleyF:
When feature("director") and feature("ref") are used together, the method named by ref (we'll call it AddRef()) is called prior to swig_connect_director(). This would be fine, since the ref feature is always going to point to an existing method, not a generated method.

However, the generated SwigDirector_T class does not initialize its swig_override array in the constructor, which means that at the time AddRef() is called, it's filled with random garbage. Usually it will think it has to look for a Java method to call, not find one, and throw "null upcast object".

This array should be zero'd in the constructor.

ojwb:
Fixed by #403.

wsfulton:
Re-opened as #403 needs further work.

ojwb:
Fixed by #413.

wsfulton:
#403 enhanced by #413.

PR #403, opened 2015-05-01
Zero-initialize swig_override in the director constructor.

LindleyF:
This pull request addresses https://github.com/swig/swig/issues/402, allowing feature("director") and feature("ref") to work together in C++ to Java mode.

In short, the director feature constructs the C++ object and then hooks up the director as two separate steps in Java. Until that second JNI call to set up the director, the swig_override array is uninitialized. Meanwhile, the ref feature specifies a method to call immediately after construction before control returns to Java; the method called is the generated director method, which tries to consult the uninitialized array.

By zero-initializing the swig_override array, we ensure that the method specified by the "ref" feature always calls the C++ method. There is no possibility of it calling an overriding Java method with this design; however, with "ref", that's probably not what you want anyway.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -198,6 +198,7 @@ CPP_TEST_CASES += \
 	director_protected \
 	director_protected_overloaded \
 	director_redefined \
+	director_ref \
 	director_smartptr \
 	director_thread \
 	director_unroll \
--- a/Examples/test-suite/director_ref.i
+++ b/Examples/test-suite/director_ref.i
@@ -1,0 +1,82 @@
+%module(directors="1") director_ref
+
+%{
+#include <string>
+
+class Foo {
+public:
+        Foo(int i = -1) : count(0) {}
+	virtual void OnDelete() {}
+	virtual ~Foo() {}
+	virtual std::string Msg(std::string msg = "default") { return "Foo-" + msg; }
+
+	std::string GetMsg() { return Msg(); }
+	std::string GetMsg(std::string msg) { return Msg(msg); }
+
+	void Ref() { ++count; }
+	void Unref() { --count; if (count == 0) { OnDelete(); delete this; } }
+	int GetRefCount() { return count; }
+private:
+	int count;
+};
+
+class FooPtr {
+public:
+        FooPtr(Foo* f = NULL) : my_f(f) { if (my_f) { my_f->Ref(); } }
+        ~FooPtr() { if (my_f) { my_f->Unref(); } }
+        void Reset(Foo* f = NULL) {
+		if (f) { f->Ref(); }
+		if (my_f) { my_f->Unref(); }
+		my_f = f;
+	}
+	int GetOwnedRefCount() {
+		if (my_f) { return my_f->GetRefCount(); }
+		return 0;
+	}
+
+private:
+        Foo* my_f;
+};
+
+%}
+
+%include <std_string.i>
+
+%feature("director") Foo;
+%feature("ref") Foo "$this->Ref();"
+%feature("unref") Foo "$this->Unref();"
+
+class Foo {
+public:
+        Foo(int i = -1) : count(0) {}
+	virtual void OnDelete() {}
+	virtual ~Foo() {}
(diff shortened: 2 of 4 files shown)


LindleyF:
appveyor appears to have flaked; I didn't touch the python test that failed there. I'm not sure how to re-trigger it.

wsfulton:
Don't worry about the Appveyor builds, especially the non-Java builds for this patch. Appveyor is unfortunately going through a rather unreliable phase and often just hangs.

LindleyF:
Well, that time it did the opposite. The Java check passed though, which is the one that matters here.

I can confirm this works in my code, although there isn't a test at the moment to verify these features work together.

ojwb:
Can you add a regression test for this?  If we  have one, we'll quickly notice if a future change would this again.

LindleyF:
I think this is ready to go in. How do I go about getting a decision on that?

wsfulton:
It needs a review but without a test case it'll take longer and so I at least will give it lower priority.

LindleyF:
I've added a test. The automated system doesn't seem very eager to run it though.

LindleyF:
Ahh---the tests finally ran and passed it seems.

LindleyF:
I'm not completely clear on when a director object would not have a swig_override array. It seems like if you're using the director feature, you probably want to override something. Is it really that common a case to be otherwise?

Normally I'd agree a helper object would be best, but I'm trying to keep the changes minimal on the theory that I don't really understand what's going on in a lot of the code generation logic.

LindleyF:
Actually, one simple solution would be to make swig_override a std::array, which works correctly even with size 0.

This is a C++11 feature, though. Is that acceptable?

LindleyF:
I found a simpler solution to address ojwb's concern. Instead of trying to manually initialize a bool array, I just turned the bool array into a std::bitset. The interface is the same, but the default constructor zeroes it.

LindleyF:
Good, everything still passes.

6 commits for a 2-line change :)

ojwb:
Requiring C++11 would be a problem, as we aim for maximal portability in the generated wrapper code.

But `std::bitset` sounds ideal.  Thanks for all your work on this.

wsfulton:
Regarding 6 commits for two lines of changes, for next time, I would rather a rebase was done before the merge.

wsfulton:
The general approach in SWIG is to avoid adding in additional STL header dependencies that aren't already in user's code. Please revert the addition of the <bitset> header?

LindleyF:
You're right, I should have rebased. That isn't normally a part of my workflow but it would have made sense here.

I'm not sure I see the point in avoiding STL headers in this case, but I've changed it to create a custom wrapper to do the initialization.

ojwb:
@wsfulton: That seems to be an undocumented policy - there are no matches for `STL` in the files under `Doc/Devel/`, and I can't find anything relevant by grepping for other likely strings.

wsfulton:
I'm not sure if this is documented, maybe it is more of an unspoken policy! I suppose I've always thought it is obvious and a natural follow on from using the conservative code generation policy of using ANSI C, mentioned in [SWIG and Freedom](http://swig.org/Doc3.0/Introduction.html#Introduction_nn13). This is mostly followed, although some exceptions to this have crept into director code and during Marcelo's rapid expansion of SWIG.

I have mentioned it in reviews in the past, eg development of Java's DirectorException https://github.com/swig/swig/pull/20#issuecomment-13330081. Any clean up to remove unrelated STL headers unless the wrapped code is depending on the STL would be good.

Given the absence of this 'policy' being clearly laid out, I think I'll write something into the documentation.

ojwb:
I can see the logic in avoiding features which are known to be problematic in one or more compiler implementations, or which have been introduced by a newer version of the language standards (e.g. C++11), but I don't see it makes sense to avoid using long-standardised language features on the pre-assumption that there might be an implementation with a buggy version, while assuming that our own alternative implementations will be bug-free.

Standard features in popular compilers will have been well exercised.  A custom variant in our own code won't get nearly that much hammering.

And to neatly illustrate my point, there is in fact a bug in the `class BoolArray` template, which I've just fixed in e796ecaa23e233db020a01af7b20c2d4f9bd9ee6.

PR #413, opened 2015-05-07
Don't include an STL header in SWIG.

LindleyF:
As requested.


LindleyF:
The java tests have passed. (Others are still running but they are not relevant.) Any objections to this one?

ojwb:
Thanks

PR #414, opened 2015-05-08
[Go] Improved Go Class Memory Management section of the Go documentation.

### LINKED ISSUE #289: Go doesn't track C++ object ownership

michael-schaller:
Fixes #289.

### DIFF
--- a/Doc/Manual/Go.html
+++ b/Doc/Manual/Go.html
@@ -436,34 +436,119 @@ for this by calling the Swigcptr() method.
 
 
 <p>
-Calling <tt>NewClassName</tt> for some C++ class <tt>ClassName</tt>
-will allocate memory using the C++ memory allocator.  This memory will
-not be automatically freed by Go's garbage collector as the object ownership is
-not tracked.  When you are done with the C++ object you must free it manually
-using <tt>DeleteClassName</tt>.
-</p>
-
-<p>
-A common technique is to store the C++ object into a Go object, and
-use the Go function <tt>runtime.SetFinalizer</tt> to free the C++ object when
-the Go object is freed. It is strongly recommended to read the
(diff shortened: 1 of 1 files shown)


ianlancetaylor:
Looks good, thanks.

michael-schaller:
Thanks, Ian. Wish you a nice weekend.

PR #426, opened 2015-06-02
Rebase COM branch on the current master.

vadz:
This PR is mostly to check that nothing is broken by the addition of COM module for the other languages, COM stuff itself is not ready for testing yet.


wsfulton:
I'm not entirely clear what your thinking here is @vadz ? Can't you check the status of code in the vadz repository by using travis to build the vadz:COM branch in the vadz repository? Or is this to start a discussion about potentially merging this branch into the main master? BTW, I'm more than happy with you  committing COM improvements directly the swig repo in the jezabek-gsoc2008 branch.

vadz:
I didn't find a way to test COM under Unix yet (at least without relying on WINE installation with all the necessary tools, notably `midl` for which there is just no replacement), so all I wanted to test so far were the other modules just to ensure I didn't break something inadvertently during the merge (or, rather, rebase, as merge was too difficult to do).

In longer term I'd indeed like to discuss merging it into trunk if people still think it's useful. I'm not sure if I'm going to use it myself yet, I need to write the `shared_ptr<>` typemaps first as without them I can't do much with my real code. But the branch is in a rather good state otherwise (incomparably better than the other GSoC branches I had looked at), Jan seems to have done some really good work here.

ojwb:
I've no interest in COM as such, but if it's still a useful thing (I really have no idea) and we have a reasonable quality branch implementing it and someone interesting in working on it then it seems worth merging.

Or if we aren't going to, we should actually decide that and close this PR and any other tickets open for COM support (github issue search is terrible but I couldn't find one with a quick try although I think I've seen one before).  Keeping this in limbo forever doesn't help anyone and the issue and pr lists would be a lot more manageable if they were less cluttered.

wsfulton:
Okay, doesn't seem useful having this open

PR #427, opened 2015-06-03
[C#] Single file mode

Sghirate:
Added "-csout <path>" parameter. If supplied all generated C# code will
be written to the given file. Makes it easier to integrate SWIG into
automated script based build processes.


vadz:
FWIW: I think it might be not the greatest idea for big projects as the enormous C# files risk running into compiler/hardware limits (we have ~150kLoC of generated code for a not so huge library), but it would definitely make things simpler for smaller libraries and testing, so I'm all for it, thanks!

wsfulton:
Mostly looks okay to me.

Does -outdir still work?
I'm wondering if it should be -csout \<filename\> instead of -csout \<path\>.
Does it fail if the path given does not exist (it should)?

Can you change the name of -csout to -outfile ? I would like this option to be
expanded to other languages in the future and having a common name is helpful.

Some formatting problems: Set your tab spacing to 8 characters and reformat the
lines you added which don't line up. Style for pointers is wrong, eg use
'File \*f' instead of 'File\* f'. getFile() needs a description and a better name.

It would be nice if the -M option started to work with this change.

Can you add a few lines into CSharp.html about this including @vadz's comments
about it not being advised for large projects.

Sghirate:
Changed formatting, argument name and documented it. If outdir and outfile are both supplied and the uotdir doesn't exist, swig will fail.

Hope I didn't miss anything . . .

PR #429, opened 2015-06-05
Add a space between literal and string macro

ngladitz:
In C++11 a space between a literal and string macro is required.

This was pointed out by the following gcc warning when trying to build in C++11 mode:
Source/Modules/php.cxx:1736:23: warning: invalid suffix on literal; C++11 requires a space between literal and string macro [-Wliteral-suffix]
        Printf(output, "\t\t\t$c->"SWIG_PTR" = $r;\n");

### DIFF
--- a/Source/Modules/php.cxx
+++ b/Source/Modules/php.cxx
@@ -1722,7 +1722,7 @@ public:
 	      Printf(output, "\t\t\treturn new %s%s($r);\n", prefix, Getattr(classLookup(d), "sym:name"));
 	    } else {
 	      Printf(output, "\t\t\t$c = new stdClass();\n");
-	      Printf(output, "\t\t\t$c->"SWIG_PTR" = $r;\n");
+	      Printf(output, "\t\t\t$c->" SWIG_PTR " = $r;\n");
 	      Printf(output, "\t\t\treturn $c;\n");
 	    }
 	    Printf(output, "\t\t}\n\t\treturn $r;\n");


ojwb:
Looks good (though probably deserves a `CHANGES.current` entry noting that SWIG can now be built cleanly with a C++11 compiler).

Did you try running the testsuite to see if there are any instances of this in generated code?

ngladitz:
Thanks. I wasn't using swig's own build system when I found this and I was only using one particular C++ compiler which may not be representative for C++11 cleanliness in general.

I was performing a super-build of a project which includes swig in its own build system which enabled C++11 for the entire build. Swig's testsuite is not part of that super-build.

The swig wrapper that was generated by the project itself (for lua) normally is not build with warnings enabled since there are a lot of (unrelated) warnings from it otherwise ... but enabling warnings did not show any instances of this particular diagnostic here.

ojwb:
It doesn't actually seem all that easy to force SWIG to build as C++11 with GCC - configure adds `-ansi` to `CXXFLAGS`, which seems to be equivalent to `-std=c++98` in C++ mode and overrides `-std=c++11` earlier in `CXXFLAGS`.  I think you'd have to use something like `make CXXFLAGS=-std=c++11`.

I tried running the testsuite with `-std=c++11` hand-added to `CXXFLAGS` in `Examples/Makefile` for languages csharp java lua ocaml perl5 php python ruby tcl - this results in a couple of compilation errors for tcl:

```
checking tcl testcase array_typedef_memberin
array_typedef_memberin_wrap.cxx: In function ‘int _wrap_new_ExampleDetail(ClientData, Tcl_Interp*, int, Tcl_Obj* const*)’:
array_typedef_memberin_wrap.cxx:2099:75: error: use of deleted function ‘ArrayExample::ExampleDetail::ExampleDetail()’
   result = (ArrayExample::ExampleDetail *)new ArrayExample::ExampleDetail();
                                                                           ^
array_typedef_memberin_wrap.cxx:1695:11: note: ‘ArrayExample::ExampleDetail::ExampleDetail()’ is implicitly deleted because the default definition would be ill-formed:
     class ExampleDetail 
           ^
array_typedef_memberin_wrap.cxx:1695:11: error: uninitialized const member in ‘class ArrayExample::ExampleDetail’
array_typedef_memberin_wrap.cxx:1700:21: note: ‘const short int ArrayExample::ExampleDetail::node_list2 [8]’ should be initialized
         const Eight node_list2; 
                     ^
array_typedef_memberin_wrap.cxx:1695:11: error: uninitialized const member in ‘class ArrayExample::ExampleDetail’
     class ExampleDetail 
           ^
array_typedef_memberin_wrap.cxx:1701:20: note: ‘const short int ArrayExample::ExampleDetail::node_list3 [8]’ should be initialized
         ConstEight node_list3; 
                    ^
../../../Examples/Makefile:183: recipe for target 'tcl_cpp' failed
```

```
checking tcl testcase default_constructor
default_constructor_wrap.cxx: In function ‘int _wrap_new_FFF(ClientData, Tcl_Interp*, int, Tcl_Obj* const*)’:
default_constructor_wrap.cxx:2615:27: error: use of deleted function ‘FFF::FFF()’
   result = (FFF *)new FFF();
                           ^
default_constructor_wrap.cxx:1790:7: note: ‘FFF::FFF()’ is implicitly deleted because the default definition would be ill-formed:
 class FFF : public F { 
       ^
default_constructor_wrap.cxx:1777:4: error: ‘F::~F()’ is private
    ~F() { }
    ^
default_constructor_wrap.cxx:1790:7: error: within this context
 class FFF : public F { 
       ^
../../../Examples/Makefile:183: recipe for target 'tcl_cpp' failed
```

Those look like they are probably the same issue, but I'm not sure what's going on there.

I'm using GCC 5.1.1 (`g++-5 (Debian 5.1.1-5) 5.1.1 20150507` says `g++ --version`).

wsfulton:
Thanks for the patch, it was added for swig-3.0.6 as 925cec3a8290dc666257d4d20d1a4e7825229caf.

Since swig-3.0.6 was released, I have turned on the c++11 testing in Travis and fixed the test-suite for c++11 as a few problems had crept in as it was not tested regularly on Travis.

PR #432, opened 2015-06-09
Fix a race in java swig_connect_director. Lazily initialized baseclas…

sshrdp:
Fix a race in java swig_connect_director. Lazily initialized baseclass can be 64bit wide which is not atomic. So a thread can see an invalid jclass value.

### DIFF
--- a/Source/Modules/java.cxx
+++ b/Source/Modules/java.cxx
@@ -440,6 +440,9 @@ public:
       Printf(f_directors, "/* ---------------------------------------------------\n");
       Printf(f_directors, " * C++ director class methods\n");
       Printf(f_directors, " * --------------------------------------------------- */\n\n");
+#if defined(__cplusplus) && __cplusplus >= 201103L
+      Printf(f_directors, "#include <atomic>\n");
+#endif
       if (outfile_h) {
 	String *filename = Swig_file_filename(outfile_h);
 	Printf(f_directors, "#include \"%s\"\n\n", filename);
@@ -4451,7 +4454,11 @@ public:
     // .'s to delimit namespaces, so we need to replace those with /'s
     Replace(internal_classname, NSPACE_SEPARATOR, "/", DOH_REPLACE_ANY);
 
+#if defined(__cplusplus) && __cplusplus >= 201103L
+    Wrapper_add_localv(w, "baseclass", "static std::atomic<jclass> baseclass", "= {0}", NIL);
+#else
     Wrapper_add_localv(w, "baseclass", "static jclass baseclass", "= 0", NIL);
+#endif
     Printf(w->def, "void %s::swig_connect_director(JNIEnv *jenv, jobject jself, jclass jcls, bool swig_mem_own, bool weak_global) {", director_classname);
 
     if (first_class_dmethod != curr_class_dmethod) {


sshrdp:
If c++11 features are not allowed yet, let me know your preferred way.

wsfulton:
C++11 isn't used by most folk, so this isn't the ideal solution. I also don't think the test for c++11 using 201103L is portable. However, I know that solving this in C without the C++11 standard isn't so easy, so I suggest using a portable solution. The easiest would be to use Java/JNI code to implement the locking. This might be slow, so how about using an atomic variable (bool or int) and the double checked locking pattern to indicate if baseclass is set. Something like:

static bool baseclass_set = false;
static jclass baseclass = 0;

if (!baseclass_set) {
  // JNI code to start lock in this block of code
  if (!baseclass_set) {
   ... same as now finishing with setting baseclass
    baseclass_set = true;
  }
  // JNI call to free lock
}

sshrdp:
Without a memory fence the code is still racy.

wsfulton:
Sigh, yes.

wsfulton:
Any ideas for a portable/non-c++11 solution?

vadz:
There won't be any simple portable solution, even to support just the most common platforms you need ugly stuff like [this](https://github.com/wxWidgets/wxWidgets/blob/master/include/wx/atomic.h).

I think using C++11 is the only simple solution and it would be nice to provide it for at least those who can afford it. But it should be governed by a new SWIG `-c++11` option (or perhaps `-c++=11` variant of the existing one? This would be easier to extend to C++14, C++17 etc) and not by checking whether SWIG itself is being compiled with a C++11 compiler.

Alternatively, the `#if` checks should be put into the _generated_ code instead of being in SWIG itself. But then they would need to be more accurate and account for MSVC which doesn't predefine `__cplusplus` properly but does provide atomics.

ojwb:
While there's no good simple answer currently, if we aim to generate code with vadz's suggested preprocessor conditionals which locks when compiled as C11 or C++11 (and has the current failings otherwise) then things will actually be fairly rosy in the near future.

C11 added `<stdatomic.h>` and GCC >= 4.8 supports it;  GCC >= 5 defaults to `-std=gnu11` when compiling C code so it'll be available by default there.  I know clang also announced they'd default to C11, but I don't know what version that happened in.  `<stdatomic.h>` isn't required to be supported (if it isn't, `__STDC_NO_ATOMICS__` should be defined by a C11 compiler, so checking the value of `__STDC__` plus that should determine if it's safe to use).

For C++ code, GCC 6 will switch to `-std=gnu++14` by default.  I'd imagine clang will soon follow suit if they haven't already.

So we can easily deliver a fix for those with a new enough compiler to have switched the default (which will be more and more with time) and many more can currently benefit if they enable C11/C++11 mode explicitly when compiling the generated wrappers (and if you're wrapping a C++ library which makes use of C++11 features in the API headers, you'll already need to).

wsfulton:
I suppose fixing this for C++11 enabled compilers is better than not fixing it.

```#if (__cplusplus >= 201103L)``` might detect c++11 in gcc, but I don't know that this is portable? I've just looked at the standard and this is actually specified in it which is great news for detecting c++11 compatibility. Anyone have any experience whether this is actually implemented and works? Probably we should add a macros with our best effort at c++11 detection in swig.swg so that it can be simply used in the rest of the swig code.

I don't see std_atomics.h in the c++11 standard, only <atomic> in 29.2. Which section in the standard are you looking at? Also I don't see any mention of ```__STDC_NO_ATOMICS__```.

vadz:
As I wrote in my previous comment, at the very least we need to test for MSVC separately as it still defines `__cplusplus` to its C++98 value, even in the latest versions.

C++ standard only has `<atomic>` header, but not `stdatomic.h` which is in C11 only.

ojwb:
Yes, C11 != C++11 (a bit confusing that they both happened to get an update the same year.)

The value of `__cplusplus` is an all-or-nothing indicator, so this won't detect compilers with partial C++11 support which do support this (I guess that's why MSVC hasn't updated the value yet), but even if it only kicked in for GCC and clang currently, that's better than being racy everywhere, and the situation should improve with time without people even having to regenerate their wrappers.

sshrdp:
There is also another issue with my patch (in case anyone wants to pick it up):
baseclass should be only assigned after the static global variable methods[] is assigned otherwise we might return swig_connect_director while another thread is still filling in the methods[] and some of them are still null or invalid pointers.

  if (swig_set_self(jenv, jself, swig_mem_own, weak_global)) {
    if (!baseclass) {
      baseclass = jenv->FindClass("CLASS");
      if (!baseclass) return;
      baseclass = (jclass) jenv->NewGlobalRef(baseclass);
    }
    bool derived = (jenv->IsSameObject(baseclass, jcls) ? false : true);
    for (int i = 0; i < 21; ++i) {
      if (!methods[i].base_methid) {
        methods[i].base_methid = jenv->GetMethodID(baseclass, methods[i].mname, methods[i].mdesc);
        if (!methods[i].base_methid) return;
      }
      swig_override[i] = false;
      if (derived) {
        jmethodID methid = jenv->GetMethodID(jcls, methods[i].mname, methods[i].mdesc);
        swig_override[i] = (methid != methods[i].base_methid);
        jenv->ExceptionClear();
      }
    }
  }

ojwb:
@sshrdp Why did you close this?  Was this fixed meanwhile?

sshrdp:
I don't think it has been fixed, but the proposed patch had issues (see https://github.com/swig/swig/pull/432#issuecomment-159047898) that needed more work.

wsfulton:
Should be fixed now, see issue #1862.

PR #438, opened 2015-06-15
Use "mixed" path to source directory under Cygwin.

vadz:
This allows build to work with both native and Cygwin builds of SWIG and
doesn't restrict us to building in the source directory as was the case
previously because SWIG_LIB was explicitly not set under Windows.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2790,6 +2790,17 @@ AC_SUBST(SKIP_ANDROID)
 
 ABS_SRCDIR=`(cd ${srcdir} && pwd)`
 
+dnl Under Cygwin, we may need native absolute path as it is used by SWIG, which
+dnl may be a native, and not a Cygwin, program (this is the case when it's
+dnl built using MinGW or cccl compiler in Cygwin environment). However it may,
+dnl although this is probably more rare, also be built as a Cygwin program.
+dnl Using "mixed" path like we do here allows the path to work in both cases.
+case $host in
+*-*-cygwin* )
+    ABS_SRCDIR=`cygpath --mixed $ABS_SRCDIR`
+    ;;
+esac
+
 # Root directory
 ROOT_DIR=`pwd`
 case $host in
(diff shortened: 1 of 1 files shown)


vadz:
For the record, this (with more COM-specific changes) allows me to run `make check-com-examples` successfully when building in a directory different from the source one, while previously this failed because of wrong `SWIG_LIB`.

AFAICS this shouldn't break anything and so should, IMHO, be applied independently of COM stuff.

wsfulton:
Good news about COM and I'm all for the test-suite working out-of-source for Windows. However, I'd like to keep this check:
```
# For testing - Windows builds of SWIG do not need SWIG_LIB set
AC_EGREP_CPP([yes],
[#ifdef _WIN32
  yes
#endif
```
for in-source builds Windows builds. I think this is important because this is how the Windows source code uses _WIN32 for SWIG_LIB. The Windows distribution is the only distribution that should work out of the box without SWIG_LIB being set and is thus important for testing. I've used the '#ifdef _WIN32' in the test because this is how the code works and should remain working irrespective of what the testing environment is. In the windows world the host and target platforms get mixed up between mingw/cygwin/native and I'm basically doing a target platform check that will always match the code. 

Adding in support for ming would also be nice and you can probably utilise something like the code that sets SWIG_LIB_WIN_UNIX in configure.ac for this. Or use the Tools/convertpath tool which does something similar. ${srcdir} is seems to be the way to detect out-of-source builds in configure.ac.

vadz:
Hmm, I don't see how can we combine the two requirements: mine, that swig works after building it but without installing, and yours, that it doesn't depend on `SWIG_LIB` being defined. The only possibility I see is to copy the entire `Lib` directory to the build directory but this is not really ideal... Am I missing some other solution?

Maybe we could just add a separate test running `swig.exe -swiglib` without `SWIG_LIB` and checking that it produces the expected result?

wsfulton:
I think we've got our wires crossed in some way. From what I understand you want to build SWIG out-of-source for which you need to set SWIG_LIB. To clarify my suggestion, I think you should change the build so that SWIG_LIB is set for all platforms (including mingw & cygwin) when built out-of-source. If you leave the in-source build as is for _WIN32 then we can still test the native Windows build as we currently do; Appveyor tests in-source and I don't propose to change it.

vadz:
Ah, sorry, I didn't understand that you only wanted to avoid `SWIG_LIB` for the in-source builds, my fault. I'll update the patch.

vadz:
Finally done, please let me know if I still managed to misunderstand you.

vadz:
FYI: Travis broke because of "No space left on device" errors, i.e. really nothing to do with this patch.

wsfulton:
Yes, the latest changes is more of what I had in mind. I don't think mingw will work as the unix type paths that 'pwd' outputs (eg /c/swig) needs to be turned into a windows path (C:\swig). SWIG_LIB_WIN_UNIX attempts to do this right. Next time I start windows, I'll investigate and see what what needs doing.

vadz:
I don't really understand what kind of magic is the code for `SWIG_LIB_WIN_UNIX` using :-( I believe that passing the path to `cmd` auto-converts it to Windows format as described [here](http://www.mingw.org/wiki/Posix_path_conversion), but I have no idea where do the trailing spaces come from. I guess I could just copy this, of course, but I feel uncomfortable copying things I don't understand.

At the very least I would need to test this before making this change. Which version of MSYS do you use for building SWIG? I'd like to test with the same one...

wsfulton:
I've been using Cygwin rather than MinGW for many years now, so that quirk would have been from a version I used when I made that commit 9 years ago. It may or may not still exist, I havn't checked. I used MinGW on Appveyor earlier this year, but that has been upgraded recently and I have not tried it since. I havn't looked yet, but will do (but not high priority right now).

wsfulton:
I currently have MinGW 3.21 and it doesn't generated that trailing space. No harm leaving it in. I'll make the changes to this patch for MinGW. I'll put the path conversion code into the Tools/pathconvert tool. 

I rebuilt the failing Travis failures, they have passed now.

wsfulton:
There is still some more work to be done for MinGW out of source builds, something going wrong during the bison calls, but at least SWIG_LIB is correctly set to the full path now. Something for another day.

PR #439, opened 2015-06-15
Enable variable and typemap substitution in typemap kwargs

LindleyF:
Typemap substitution in typemap kwargs may be needed whenever $javaclassname is not the right type for the directorin JNI descriptor, and the necessary type cannot be hard-coded (such as in a macro).

kwargs are processed in both Swig_typemap_lookup_impl (for director method return types) and Swig_typemap_attach_parms (for director method parameters). The latter method calls a helper function typemap_attach_kwargs; the former does not call this method. I feel like it should, but there are some slight differences in handling of the 'type' attribute for reasons I don't understand, so I left them separate and instead updated both code blocks with equivalent changes.

The call to replace_embedded_typemap() is where the magic happens. Most of the rest is just getting parameters for typemap_replace_vars(), and I'm honestly not certain how many of them are really needed. It seems like the only necessary line in typemap_replace_vars() is:
Replaceall(s, "$typemap", "$TYPEMAP");
since this is a prerequisite for replace_embedded_typemap. I call the entire method for the sake of consistency and because it should make other common substitutions work too, although I don't have a good use-case for anything else right now.

I've added a new test that verifies this works for directorin:descriptor. It's based on the idea that it should be possible to write typemaps to make pointer-like classes convert to Java references to the underlying type.

### DIFF
--- a/Examples/test-suite/java/Makefile.in
+++ b/Examples/test-suite/java/Makefile.in
@@ -27,6 +27,7 @@ CPP_TEST_CASES = \
 	java_director_assumeoverride \
 	java_director_exception_feature \
 	java_director_exception_feature_nspace \
+	java_director_ptrclass \
 	java_enums \
 	java_jnitypes \
 	java_lib_arrays_dimensionless \
--- a/Examples/test-suite/java/java_director_ptrclass_runme.java
+++ b/Examples/test-suite/java/java_director_ptrclass_runme.java
@@ -1,0 +1,47 @@
+
+import java_director_ptrclass.*;
+
+public class java_director_ptrclass_runme {
+
+  static {
+    try {
+      System.loadLibrary("java_director_ptrclass");
+    } catch (UnsatisfiedLinkError e) {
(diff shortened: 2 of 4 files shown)


LindleyF:
Ping?

wsfulton:
I've been looking at this pull request today. It looks okay. I think I can make one small improvement to support the expansion of special variables within typemap attributes and am testing it now. This is done by replacing 0 with 1 in typemap_replace_vars() which then kicks in the majority of the functionality in this function, ie replacing special variables such as $1_type. I just am not sure if multi-argument typemaps will work.

The patch ought to be rebased so that there is just one commit. I'll do that when I merge if you don't do it here.

LindleyF:
Rebased. Any luck with the additional expansions?

wsfulton:
I've made some changes for special variable expansions to work in typemap attributes as well. These work in combination with your special variable macros (including multi-argument typemaps. Also added some docs and tests. See 2f00f6c8cc6377210312290913662ddacf9e8b69 and f482adc6d1a4b1cc27b4ecbb6b461d1f747906b6.

Issue #225, opened 2014-09-14
runme.rb:15: [BUG] object allocation during garbage collection phase



andreas-schwab:
Examples/ruby/mark_function fails if pointer values don't fit in ruby fixnums

```
(gdb) bt
#0  0xc02daed8 in __GI_raise (sig=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
#1  0xc02dc1d8 in __GI_abort () at abort.c:78
#2  0xc009bfb6 in rb_bug (
    fmt=0xc01c8e6a "object allocation during garbage collection phase")
    at error.c:341

#3  0xc00b8db6 in newobj_of (klass=2147521340, flags=42, v1=0, v2=0, v3=0)
    at gc.c:1290
#4  0xc00b8df8 in rb_newobj_of (klass=2147521340, flags=42) at gc.c:1353
#5  0xc006b1a8 in bignew_1 (klass=2147521340, len=1, sign=1) at bignum.c:2985
#6  0xc0073eb0 in rb_uint2big (n=2144845944) at bignum.c:3148
#7  0xc0073f16 in rb_int2big (n=-2144845944) at bignum.c:3180
#8  0xc0520754 in rb_long2num_inline (v=-2144845944)
    at /usr/include/ruby-2.1.0/ruby/ruby.h:1307
#9  0xc0520f2e in SWIG_RubyPtrToReference (ptr=0x80283f88)
    at example_wrap.cxx:1228
#10 0xc0520fcc in SWIG_RubyInstanceFor (ptr=0x80283f88)
    at example_wrap.cxx:1272
#11 0xc0521f68 in mark_Zoo (ptr=0x8028bd48) at example_wrap.cxx:1993
#12 0xc00b639a in gc_mark_children (objspace=0x80005778, ptr=2150076460)
    at gc.c:3883
#13 0xc00b6b94 in gc_mark_stacked_objects (objspace=0x80005778) at gc.c:3968
#14 gc_marks_body (objspace=0x80005778, full_mark=1) at gc.c:4152
#15 0xc00b7e8c in gc_marks (full_mark=255, objspace=0x80005778) at gc.c:4504
#16 garbage_collect_body (objspace=0x80005778, full_mark=255,
    immediate_sweep=1, reason=<optimized out>) at gc.c:5011
#17 0xc00b803a in garbage_collect (objspace=0x80005778, full_mark=1,
    immediate_sweep=1, reason=1024) at gc.c:5064
[...]

PR #444, opened 2015-06-18
Avoid conversion of C pointers to Ruby Bignum.

voxik:
The trackings hash is split into two hashes now. One is for pointers
which can be converted to Ruby's Fixnum, while the other is used for
non-convertable. This is needed, since C pointers are 32/64 bit values,
while Ruby's Fixnum can store only 31/63 bit values.

This avoid issues desribed in #225, i.e. allocation of object during GC
phase.


voxik:
I should note, that instead of calling SWIG_RubyPtrToReference, which can allocate "Bignum" at background, I am using LONG2FIX, which forces the conversion of pointer into Fixnum, no matter what is its content, but since it is used just as a key for hash, that should be perfectly fine.

wsfulton:
Please see #225 for the fix.

Issue #437, opened 2015-06-15
[C#] Directors and non-RTTI compilation


Sghirate:
Currently the director implementation for C# relies on dynamic_casts, thus requiring RTTI to be enabled, when compiling the generated cpp code. However, I do not understand the need for dynamic_cast (instead of static_cast) in that context (to me it seems like the type in *_connect can't be much else than the derived director class) and with static_casts it is compiling and seems to function perfectly fine.

So my question would be:
Why do directors use *dynamic_cast* instead of *static_cast*? 
Am I missing/misunderstanding something?

If there is no reason against using static_cast, I would like to change that in order to allow for non-RTTI compilation.

ojwb:
Closing in favour of PR #449.

PR #449, opened 2015-06-27
[C#] Fix allow non-RTTI compilation

Sghirate:
Used static_cast instead of dynamic_cast for directors - allowing generated native code to be compiled without RTTI flag (which can negatively impact application performance).

_Related Issue_: https://github.com/swig/swig/issues/437

_Note_: I do not see, why dyanmic_cast would be necessary in the case of directors and in my test cases it worked without a problem. But, if anybody wants to raise an objection - feel free!


vadz:
Just out of curiosity, did you benchmark the difference in performance with static vs dynamic cast? How much is it?

Source/Modules/csharp.cxx
```diff
@@ -2044,7 +2044,7 @@ class CSHARP:public Language {
 	Printf(dcast_wrap->code, "  Swig::Director *director = (Swig::Director *) 0;\n");
 	Printf(dcast_wrap->code, "  jobject jresult = (jobject) 0;\n");
 	Printf(dcast_wrap->code, "  %s *obj = *((%s **)&jCPtrBase);\n", norm_name, norm_name);
-	Printf(dcast_wrap->code, "  if (obj) director = dynamic_cast<Swig::Director *>(obj);\n");
+	Printf(dcast_wrap->code, "  if (obj) director = static_cast<Swig::Director *>(obj);\n");
 	Printf(dcast_wrap->code, "  if (director) jresult = director->swig_get_self(jenv);\n");
```
vadz: The `if (director)` test could be removed now, `director` is not going to be `NULL` if `obj != NULL` with static cast.

ezralanglois: This looks good, but may be it should be a conditional compile time choice? Some people might want the run time safety. You could add a compile flag or check compiler specific variables like __GXX_RTTI.

johnb003:
+1. I think this is important. We ended up working around this by post-processing the files and replacing the dynamic casts. A built in solution would be nice.

devluz:
This would be really great to have! I am wrapping a non rtti library currently and mixing the library with rtti wrapper causes a lot of trouble!

wsfulton:
I think this would be okay. I'd like it implemented uniformly across all the languages at the same time though.

devluz:
The python version seem to use a -nortti flag according to the python.cxx.

ojwb:
The `-nortti` option which SWIG's Python backend supports defines `SWIG_DIRECTOR_NORTTI` which appears to use a `std::map<void *, Director *>` in place of `dynamic_cast` - not sure that's really an improvement overall.  The Perl backend supports the same define, but there's no `-nortti` option there.

wsfulton:
I've extended this to Java and D in b90a8d7444360abbc338875eb69b52cfb2972b5a as these two language's director designs are very similar. I can't see how this can be easily done for all the scripting language's implementation of directors, so has been left for another day!

PR #452, opened 2015-06-29
lua: push characters as unformatted 1-character strings

ngladitz:
Since Lua 5.3 the "%c" format character in lua_pushfstring will produce
the string "<\XXX>" (XXX being a decimal code sequence) when
given unprintable characters.

Use lua_pushlstring instead to reproduce the old behavior.

### DIFF
--- a/Examples/test-suite/chartest.i
+++ b/Examples/test-suite/chartest.i
@@ -1,0 +1,15 @@
+%module chartest
+
+%inline %{
+char printable_global_char = 'a';
+char unprintable_global_char = 0x7F;
+
+char GetPrintableChar() {
+  return 'a';
+}
+
+char GetUnprintableChar() {
+  return 0x7F;
+}
+
+%}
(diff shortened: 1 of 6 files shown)


ojwb:
The change seems sensible to me (though I'm no Lua expert).

It would be good to have a regression test for this - currently the testsuite for Lua doesn't seem to generate *any* calls to `lua_pushfstring()`.

`git grep` shows that `SWIG_Lua_InstallConstants()` in `Lib/lua/luarun.swg` and `%typemap(throws) char` in `Lib/lua/lua.swg` also both use `lua_pushfstring(L,"%c",foo)` so I suspect those will also need a similar fix.

ngladitz:
Thank you for looking this over.

I changed the other two locations you mentioned and tried adding a new test case "chartest" to cover this.

ojwb:
Thanks.

wsfulton:
Does the use of lua_pushlstring change the minimum version of Lua that is required? When was it introduced? Hopefully it is an old version, I see some usage in std_string.i before the patch was applied, although these typemaps aren't used by default.

ngladitz:
lua_pushlstring() seems to be older than lua_pushfstring().

I checked e.g. 4.0.1 which has the former but not the later.

PR #457, opened 2015-07-04
Fix unitialized variable warning 2

xantares:


wsfulton:
The array won't be initialized in C. Compile in gcc with:
```
-std=c89 -pedantic
```
and you'll get:
```
warning: ISO C forbids empty initialiser braces [-pedantic]
```
Other compiler will probably reject it outright. I suggest instead using 
```
PyObject *argv[%d] = {0}
```
Please don't create any new pull requests. Just fix it in your current branch, rebase and push it.

wsfulton:
What was the reason for closing? I think we should merge this in after rebasing the two commits into one.

wsfulton:
Looks like it was closed as #477 has a new patch for this.

PR #458, opened 2015-07-06
Set OCAMLC to empty if Ocaml compiler not found

kkaempf:
OCAMLC is tested for 'empty string' when setting SKIP_OCAML
(configure.ac:2677), setting it to ':' will not skip ocaml.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1891,7 +1891,7 @@ else
 
     AC_MSG_CHECKING(for Ocaml compiler)
 	if test -z "$OCAMLC"; then
-	AC_CHECK_PROGS(OCAMLC, ocamlc, :)
+	AC_CHECK_PROGS(OCAMLC, ocamlc, )
     fi
 
     AC_MSG_CHECKING(for Ocaml toplevel creator)


wsfulton:
Looks like this may have started since @ojwb commit 6d80e406999c2074d307d51e612daab683c6efea . I wonder what the point of the ':' as the 3rd argument to AC_CHECK_PROGS. Wouldn't it make sense to remove all of them?

ojwb:
It was actually like that before my change, but...

`:` is a shell built-in which evaluates but then ignores any arguments, and returns status 0 - it's much 
like `true`, except that some `true` implementations don't ignore things like `--version`:

```
$ /bin/true --version
true (GNU coreutils) 8.23
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jim Meyering.
```

The advantage of setting a tool we haven't found to `:` instead of empty is that if something tries to use that tool we won't end up trying to run its arguments as a command.  At best that usually results in a confusing error message, and at worst it has the potential to run an unintentional command.

As a particularly nasty example: `$(ECHO) rm -rf /` is much worse after `ECHO=` than after `ECHO=:`.

Issue #459, opened 2015-07-08
octave 4.0 hdf5 support


opoplawski:
WIth octave 4.0.0 with hdf5 support I get:
~~~~~
/builddir/build/BUILD/plplot-5.11.0/fedora/bindings/octave/plplot_octaveOCTAVE_wrap.cxx:1638:18: error: 'hid_t' has not been declared
       save_hdf5 (hid_t loc_id, const char *name, bool save_as_floats) {
                  ^
~~~~~
swig source is:
~~~~~
#if defined (HAVE_HDF5)
    virtual bool
      save_hdf5 (hid_t loc_id, const char *name, bool save_as_floats) {
      return true;
    }
~~~~~
This now appears to be caused by the update to octave 4.0.0.  With 3.8.2:

```
/usr/include/octave-3.8.2/octave/ov.h:#include "oct-hdf5.h"
/usr/include/octave-3.8.2/octave/oct-hdf5.h:#include <hdf5.h>
```

so ov.h included `<hdf5.h>`.  But now:

```
/usr/include/octave-4.0.0/octave/oct-hdf5.h:#include <hdf5.h>
/usr/include/octave-4.0.0/octave/ls-hdf5.h:#include "oct-hdf5.h"
```

only ls-hdf5.h or oct-hdf5.h will bring in <hdf5.h>.

Also, octave's definition of save_hdf5 has changed from:

```
bool save_hdf5 (hid_t loc_id, const char *name, bool save_as_floats)
```

to

```
bool save_hdf5 (octave_hdf5_id loc_id, const char *name, bool save_as_floats);
```

So it looks like Lib/octave/octrun.swg needs to get updated to handle that?

ojwb:
Looks like this was fixed by PR #460, so closing.  If not, please reopen.

Issue #356, opened 2015-03-13
Octave 4.0 compatibility


rdmitry:
Swig uses OCTAVE_ALLOCATOR, but it was deleted:

> *\* The configure option --enable-octave-allocator has been removed.
> The internal class <octave_allocator> declared in oct-alloc.h has been removed.  The header remains, but is deprecated.  The macros to access the class (DECLARE_OCTAVE_ALLOCATOR, DEFINE_OCTAVE_ALLOCATOR, and DEFINE_OCTAVE_ALLOCATOR2) are now unconditionally defined to be empty.

ojwb:
Looks like this was addressed by PR #460, please reopen if not.

PR #460, opened 2015-07-08
Support for octave 4.0.0

opoplawski:
Initial attempt at octave 4.0.0 support.  Addresses issues #459 and #356.  I'm still not getting a clean build on Fedora rawhide:

https://kojipkgs.fedoraproject.org//work/tasks/4841/10324841/build.log

~~~~~
checking Examples/octave/variables
In file included from /usr/include/c++/5.1.1/x86_64-redhat-linux/bits/os_defines.h:39:0,
                 from /usr/include/c++/5.1.1/x86_64-redhat-linux/bits/c++config.h:2216,
                 from /usr/include/c++/5.1.1/cstdlib:41,
                 from example_wrap.cxx:141:
/usr/include/features.h:328:4: warning: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Wcpp]
 #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)
    ^
error: octave_base_value::print (): wrong type argument 'swig_ref'
error: called from
    runme.m at line 40 column 1
../../Makefile:453: recipe for target 'octave_run' failed
~~~~~


opoplawski:
Second commit gets me past the print() issue.  Now getting:
~~~~~

checking octave testcase default_arg_values
error: function delete_B requires at least 1 arguments
error: called from
    ./default_constructor_runme.m at line 17 column 7
~~~~~
and several messages like this:
~~~~~
checking octave testcase refcount (with run test)
refcount_wrap.cxx: In static member function 'static octave_value swig::traits_from_stdseq<Seq, T>::from(const sequence&)':
refcount_wrap.cxx:3477:30: error: 'INT_MAX' was not declared in this scope
       if (size <= (size_type)INT_MAX) {
                              ^
../../../Examples/Makefile:444: recipe for target 'octave_cpp' failed
~~~~~

opoplawski:
It now sees that octave is pickier about needed to pass an argument to functions that are declared to have one.  I tried:

 del_b = dc.delete_B(b)

but that results in:

 error: 'b' undefined near line 17 column 21

So I'm not sure what to do here or how this test is supposed to work.  If 'b' is gone though, perhaps there is no need to call delete_B?

opoplawski:
This is my best guess at a fix for the default_constructor test.  swig now compiles and tests run for me.

wsfulton:
Can you add ver 4.0.0 to .travis.yml so that it gets tested on the Travis infra with the appropriate version? Removing code from the  default_constructor_runme.m test seems wrong and this ought to be better understood. I'm hoping @kwwette can shed some light and review the patch.

opoplawski:
I'm not at all familiar with travis configuration.  Hope this is correct.

kwwette:
It looks like the call to delete_B, etc. isn't actually meaning to call the function, but merely test that it exists. To do this, however, it seems to be using an Octave bug/feature - calling the function without arguments but *with* a semicolon! - which evidently no longer exists. A possible replacement might be

-del_b = dc.delete_B;
+assert(exist("delete_B") == 3)

since delete_B will also by default be exported as a global function.

Travis uses Ubuntu Precise which only has Octave 3.2 - newer versions are being installed from a private PPA of mine. I would need to package and upload Octave 4.0 to this PPA first for the Travis build to work. I'll see if I can remember how to do this...

kwwette:
I've added Octave 4.0.0 to the PPA, so the Travis build should now be able to install it.

kwwette:
I would also recommend updating section 32.1 (SWIG and Octave: Preliminaries) in the documentation to reflect the Octave versions currently being tested, e.g.:

As of SWIG 3.0.x, the Octave module is regularly tested against Octave versions 3.2.4, 3.8.1, and 4.0.0.

wsfulton:
I restarted the Octave tests for this pull request, see https://travis-ci.org/swig/swig/jobs/70484945. Looks good now.

Issue #445, opened 2015-06-20
[Python] set docstring in tp_doc slot for classes with -builtin


m7thon:
Currently, the docstring for wrapped classes in python when using -builtin is always set to be the class name (ignoring `%feature("autodoc")` or `%feature("docstring")`). I.e., there is currently no way to specify the docstring for wrapped classes in python when using -builtin, which is annoying.

The `tp_doc` slot for the python class is set in [Sources/Modules/python.cxx](https://github.com/swig/swig/blob/master/Source/Modules/python.cxx) as:
```printSlot(f, quoted_rname, "tp_doc");```

However, a mechanism is in place to extract autodoc and or docstring information and place it in the attribute `feature:python:tp_doc` of the node in the swig parse tree.

Is there any reason that this is not used? To fix this, shouldn't line 3873 essentially(+) be instead:
```printSlot(f, getSlot(n, "feature:python:tp_doc"), "tp_doc");```?
(+) The string needs to be surrounded by quotes, of course, and then freed appropriately.

wsfulton:
Fixed by #461.

PR #461, opened 2015-07-09
[Python] Set docstring in tp_doc slot for classes with -builtin (issue #445)

m7thon:
This fixes issue #445, i.e.:

Docstrings provided via `%feature("docstring")` are now quoted and added to the `tp_doc` slot when using python builtin classes. When no docstring is provided, the `tp_doc` slot is set to the "quoted rname" of the class as before.


wsfulton:
This looks fine. Thanks.

PR #462, opened 2015-07-11
Octave: use correct mkoctfile executable

mtmiller:
Allow a versioned Octave executable to be configured via --with-octave
and the correct corresponding mkoctfile executable to be used.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1030,7 +1030,7 @@ fi
 
 if test -n "$OCTAVE"; then
    AC_MSG_CHECKING([for mkoctfile])
-   mkoctfile="`dirname ${OCTAVE}`/mkoctfile"
+   mkoctfile="$(dirname $OCTAVE)/$(basename $OCTAVE | sed -e 's/octave/mkoctfile/')"
    AS_IF([test -x "${mkoctfile}"],[
       AC_MSG_RESULT([${mkoctfile}])
    ],[


wsfulton:
Another one for @kwwette to look at.

kwwette:
This looks fine to me.

wsfulton:
@mtmiller, would you mind giving some details of the files that this now fixes. I'm finding it hard to follow without the particular version of Octave installed.

mtmiller:
@wsfulton, this fix allows swig to be configured to a specific Octave if more than one version are co-installed. For example I have `/opt/gnu/octave/bin/octave-3.8.2` and `/opt/gnu/octave/bin/octave-4.0.0`.

Without this change, as it is currently, `./configure --with-octave=$mybindir/octave-3.8.2` will set `mkoctfile=$mybindir/mkoctfile`, which may or may not be the right version for the Octave I want to use.

With this change, we have:


`./configure --with-octave=` | `mkoctfile=`
--- | ---
`$dir/octave` | `$dir/mkoctfile`
`$dir/octave-3.8.2` | `$dir/mkoctfile-3.8.2`
`$dir/octave-4.0.0` | `$dir/mkoctfile-4.0.0`

I think this fix continues to support the most common case of `/usr/bin/octave` and `/usr/bin/mkoctfile` as well as allowing developers like me to pick from multiple versions of Octave in parallel.

PR #463, opened 2015-07-15
Fix typo on SWIG Java

chibi929:
Fixed a typo in the variable name when using the Director.


wsfulton:
I don't see this as a typo, I don't think the name is any better. You probably found the name confusing as is often the case with variable names which are not fully spelt out. A better name would be director_method_ids instead of director_methids. Perhaps change it to that. Please rebase your patch if you resubmit.

chibi929:
Thank you for explanation.
I understood it. `director_methids` was understood that it was an abbreviation of `director_method_ids`.

PR #467, opened 2015-07-18
Remove callback function from autodoc unit test.

vadz:
It doesn't seem to belong there at all, there is a dedicated callback unit
test for it and it was added to the initial version of autodoc.i back in
124253d69828b4323f939bf118254ee96aefcdc7 without any explanation, so just
remove it.

----
I'm submitting this separately, it's one of commits from my COM branch and is needed there to allow the COM test suite to pass the autodoc test as it doesn't have callbacks (yet), but it also just seems to globally make sense.

### DIFF
--- a/Examples/test-suite/autodoc.i
+++ b/Examples/test-suite/autodoc.i
@@ -114,12 +114,6 @@
   }
 %}
 
-%callback("%(uppercase)s_CALLBACK") func_cb;
-
-%inline {
-  int func_cb(int c, int d) { return c; }
-}
-
 // Bug 3310528
 %feature("autodoc","1") banana; // names + types
 %inline %{
--- a/Examples/test-suite/callback.i
+++ b/Examples/test-suite/callback.i
@@ -1,10 +1,18 @@
 %module callback
 
+// Not specifying the callback name is only possible in Python.
+#ifdef SWIGPYTHON
 %callback(1) foo;
 %callback(1) foof;
 %callback(1) A::bar;
 %callback(1) A::foom;
-%callback("%s_Cb_Ptr") foo_T;  // old style, still works.
+#else
+%callback("%s") foo;
+%callback("%s") foof;
+%callback("%s") A::bar;
+%callback("%s") A::foom;
+#endif
+%callback("%s_Cb_Ptr") foo_T;  // this works in Python too
 
 %inline %{
 
--- a/Examples/test-suite/callback.i
+++ b/Examples/test-suite/callback.i
@@ -12,7 +12,7 @@
 %callback("%s") A::bar;
 %callback("%s") A::foom;
 #endif
-%callback("%s_Cb_Ptr") foo_T;  // this works in Python too
+%callback("%(uppercase)s_Cb_Ptr") foo_T;  // this works in Python too
 
 %inline %{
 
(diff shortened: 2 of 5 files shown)


vadz:
It turns out (thanks Travis) that this callback was used in a PHP test, so to remove it while keeping the PHP run test I had to move PHP code to `callback` test which, in turn, required fixing the latter test to pass with PHP first, as it didn't pass before.

@ojwb Could you please have a look at my PHP changes? I've never touched it before so I'm hardly knowing what I'm doing, I just cut-and-pasted the old test there. TIA!

wsfulton:
Splitting this test out of autodoc.i into callback.i makes sense. I suggest you keep the %(uppercase) test:
```
%callback("%(uppercase)s_CALLBACK") func_cb;
```
to be identical. Without knowing more details about this test when it was added (the misnaming may have been due to uppercasing, it would be prudent to keep the test the same.

vadz:
Updated the patch to use `%(uppercase)s` inside `%callback`.

vadz:
NB: AppVeyor failures are spurious.

wsfulton:
Thanks, looks okay to me now and ready for merge unless @ojwb has any comments.

ojwb:
I am on vacation currently so am unlikely to be able to look for at least 3
weeks.

ojwb:
> it was added to the initial version of autodoc.i back in 124253d without any explanation

There's no explicit explanation, but looking at the commit it seems pretty clear to me its intended to be testing that autodoc works for callback functions, i.e. it's test coverage for this part of that commit:

https://github.com/swig/swig/commit/124253d69828b4323f939bf118254ee96aefcdc7#diff-3cc91cccc4a3a707448c1f5993015108R882

Testing with the code prior to this removal, there is indeed an autodoc for `func_cb` (though `autodoc_runme.py` didn't actually check this docstring):

```
def func_cb(c, d):
    """func_cb(int c, int d) -> int"""
    return _autodoc.func_cb(c, d)
```

Removing this seems wrong - instead it should be disabled for a backend which doesn't support callbacks (using `#ifndef SWIGCOM` or similar).  Any objections to me reinstating it and fixing `autodoc_runme.py`?  Then the com branch can disable this as above.

I'd agree that the test PHP is doing makes much more sense where it's been relocated to - I probably just added it where I did because the more obvious `callback` example wasn't set up to work with PHP.

vadz:
OK, I didn't realize this test was supposed to do it. I have no objections to restoring it either here or in the callback test, whichever you prefer. Unfortunately my work on the COM branch got preempted by something else anyhow, so it's nowhere close to being in a mergeable state yet in any case :-(

ojwb:
So I tried to reinstate this and add a test for it, but the test fails because the autodoc has some bogus extra stuff at the end:

```
Expected: [func_cb(int c, int d) -> int]
Got     : [func_cb(int c, int d) -> int
swig_ptr: _1ca2cae3e47f0000_p_f_int_int__int]
```

The cause seems to be the last line of this quoted section from the generated `autodoc.py`:

```
def func_cb(c, d):
    """func_cb(int c, int d) -> int"""
    return _autodoc.func_cb(c, d)
func_cb = _autodoc.func_cb
```

That line seems to be superfluous since the function before it forwards calls, but I completely failed to find the code which generates it, so I can't see how to get rid of it to try to test my hypothesis.

Any ideas?

ojwb:
Ah, never mind, I finally found it.  SWIG has too many ways to print a string...

PR #470, opened 2015-07-21
missing feature:ignore

wkalinin:
explicitly %ignore'd symbol does not get feature:ignore if it is only added to C symbol table

### DIFF
--- a/Source/CParse/parser.y
+++ b/Source/CParse/parser.y
@@ -466,6 +466,8 @@ static void add_symbols(Node *n) {
     if (only_csymbol || GetFlag(n,"feature:ignore")) {
       /* Only add to C symbol table and continue */
       Swig_symbol_add(0, n);
+      if (strncmp(Char(symname),"$ignore",7) == 0)
+	SetFlag(n,"feature:ignore");
     } else if (strncmp(Char(symname),"$ignore",7) == 0) {
       char *c = Char(symname)+7;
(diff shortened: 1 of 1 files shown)


Source/CParse/parser.y
```diff
@@ -466,6 +466,8 @@ static void add_symbols(Node *n) {
     if (only_csymbol || GetFlag(n,"feature:ignore")) {
       /* Only add to C symbol table and continue */
```
vadz: Isn't this comment out of date now?

wkalinin:
No, why? The symbol indeed is only added to the C symbol table in that case (protected nested class, when nested classes are not supported).

vadz:
I interpreted this comment as saying that this branch of code does _not_ call `SetFlag("feature:ignore")` unlike the one just below and thought it didn't make sense any more now that it can set this flag as well. But to be honest I don't fully understand what's going on here, so I could be completely wrong about the meaning of this comment...

wkalinin:
Actually, as I understand, the comment applies to both branches, as opposed to the last, where the symbol receives sym:name and is added to the real symbol table (not only C). I think these 2 "if" branches might be merged for clarity.

wkalinin:
refactored: https://github.com/swig/swig/pull/472
I left the warning about invalid $ignore string, but I'm not sure if it is necessary, Who cares what is behind the "$ignore" if it is never used anyway?

PR #473, opened 2015-07-22
Don't use preinst-swig in the test suite makefile.

vadz:
This is unnecessary if we export SWIG_LIB from this makefile instead and prevents SWIGTOOL=gdb from working as gdb can't be used to debug a shell script -- but only the real SWIG binary.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -56,8 +56,8 @@ endif
 COMPILETOOL=
 SWIGTOOL   =
 
-SWIG       = $(SWIGTOOL) $(top_builddir)/preinst-swig
-SWIG_LIB   = $(top_srcdir)/Lib
+SWIG       = $(SWIGTOOL) $(top_builddir)/swig
+export SWIG_LIB   = $(top_srcdir)/Lib
 TEST_SUITE = test-suite
 EXAMPLES   = Examples
 CXXSRCS    = 


vadz:
FYI: the appveyour failure is spurious, due to nuget timeout.

wsfulton:
I don't care for preinst-swig either and would rather invoke swig directly. However, I really am not keen on exporting variables as replicating what make shows on the terminal becomes quite hard and requires working out what env have set and is often not easy to follow. i'd rather see the env variables required being set in one line such as:
```
env SWIG_LIB=../../../Lib swig -c++ -python blah.i
```
so it can be easily copy pasted to replicate. It used to work like this before preinst-swig got introduced, but I wasn't quite sure what was wrong with the old way.

vadz:
I think requiring to set `SWIG_LIB` is not a big deal, SWIG errors ("unable to find swig.swg") are relatively clear in this case. Also, the main point of this patch, for me, is that you won't need to copy-and-paste from the terminal so much because 90% of the time I was doing this in order to run swig under gdb and this can now be done much more easily with `make SWIGTOOL=gdb`.

As for getting rid of `preinst-swig` entirely, it's not that simple because of the whole baroque structure of SWIG test suite makefiles `SWIG` variable is being passed around a lot and I don't know how to simply use `env SWIG_LIB=... swig` instead of `preinst-swig` for it...

I hope you can apply this for now as it does simplify debugging SWIG a lot, and maybe someone (maybe even me, but maybe not) can deal with `preinst-swig` some time later.

ojwb:
> I think requiring to set SWIG_LIB is not a big deal, SWIG errors ("unable to find swig.swg") are relatively clear in this case.

If there's an installed version of SWIG you may not get an error, but instead quiet use of the older library files from that installed version.

I'd like the `preinst-swig` script to continue to exist even if it isn't used inside SWIG - it's very convenient for testing other projects with a build of SWIG from git, in a way which having to juggle the `SWIG_LIB` setting as well wouldn't be.  If it gets removed, I'll probably end up having to maintain something equivalent to it outside of SWIG.

vadz:
Personally I use `preinst-swig` sometimes as well but wouldn't have any problem with using `SWIG_LIB=... swig` instead, it's still a one-liner.

However the original patch doesn't prevent the use of `preinst-swig` at all, it just stops using it for the test suite. I'd still like to see it applied, whatever other improvements can be done in the future. It's a simple change which helps me a lot with debugging right now.

wsfulton:
As convenient as it might be, I **really really really** don't like mysteries where what you see is not what you need to replicate something, I don't ever want to have any export statements in the Makefile. It might be obvious to a more experienced SWIG developer/user that SWIG_LIB needs setting, it is just plain user unfriendly and I want to remove obstacles to developing SWIG rather than add to them. Sorry, but there is a better solution although I appreciate that it will take a lot more Makefile lines of code to implement. I'll see if I can prototype it in one of the languages and report back.

wsfulton:
I've prototyped it in ca1431f4cf1671961a1e012c4042bb90c152181f which has the Java test-suite and Java class example. I'll extend to all the other languages unless there are any objections.

ojwb:
> Personally I use preinst-swig sometimes as well but wouldn't have any problem with using `SWIG_LIB=... swig` instead, it's still a one-liner.

That's OK for running it by hand, but it doesn't work so well for specifying which swig to use to another project's build system.

wsfulton:
8e2bc59 removed preinst-swig from the build system.

PR #474, opened 2015-07-23
fix #453

### LINKED ISSUE #453: GCC warning about uninitialized variable in overload functions dispatcher
Generated overload functions dispatcher causes warning under g++ 4.9.2, like `warning: ‘argv[0]’ may be used uninitialized in this function [-Wmaybe-uninitialized]`. My SWIG version is 3.0.2.

I propose a patch to fix this error: https://gist.github.com/oliora/b19f20c4b374e99be4af

PS. I did not tested the patch and I know nothing about how SWIG generates the wrapper, so probably this patch doesn't do what it should. But at least you can got my idea through it :)

xantares:
and add the right number of zeros, seems Wrapper_add_local adds unexpected eols between braces, but it's valid:
```
WIGINTERN PyObject *_wrap_new_Box(PyObject *self, PyObject *args) {
 9622   int argc;
 9623   PyObject *argv[2] = {
 9624     0, 0
 9625   };
 9626   int ii;
 9627 
 9628   if (!PyTuple
```

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -2385,7 +2385,12 @@ public:
     Printv(f->def, linkage, builtin_ctor ? "int " : "PyObject *", wname, "(PyObject *self, PyObject *args) {", NIL);
 
     Wrapper_add_local(f, "argc", "int argc");
-    Printf(tmp, "PyObject *argv[%d]", maxargs + 1);
+    String *zeros = NewString("0");
+    for (int i = 0; i < maxargs; ++ i) {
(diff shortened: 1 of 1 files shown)


xantares:
FYI travis is broken by beabf61, but ok for python and this PR: https://travis-ci.org/swig/swig/builds

vadz:
Why bother with the right number of zeroes? Generating `PyObject *argv[N] = { 0 };` does exactly the same thing for any `N` and is simpler...

xantares:
you're right

PR #476, opened 2015-07-24
incorrect warning for private symbols

wkalinin:


wsfulton:
Looks okay to me. I've rebased these 4 commits into one and committed to master.

wkalinin:
Thanks for looking into this.

PR #477, opened 2015-07-25
fix #453

### LINKED ISSUE #453: GCC warning about uninitialized variable in overload functions dispatcher

xantares:


vadz:
I've committed this with a slightly more informative commit message as 812f789db6ec21ebe817521424dff32ada8be00c, thanks.

PR #478, opened 2015-07-25
Complain if any new shift/reduce conflicts are introduced in the grammar.

vadz:
Currently there are 6 of them, suppress warning about them but do complain if
any new ones are introduced.

----

This basically formalizes what you wrote in a comment to PR #170.

### DIFF
--- a/Source/CParse/parser.y
+++ b/Source/CParse/parser.y
@@ -13,11 +13,9 @@
  * some point.  Beware.
  * ----------------------------------------------------------------------------- */
 
-/*
-Removed until we know more about the min versions of Bison and Yacc required for this
-to work, see Byacc man page: http://invisible-island.net/byacc/manpage/yacc.html
+/* There are 6 known shift-reduce conflicts in this file, fail compilation if any
+   more are introduced. */
 %expect 6
-*/
 
 %{
 #define yylex yylex


vadz:
P.S. Tested with 2.3 (under OS X 10.8) too, works as well.

vadz:
I've tested this with 3.0, works just fine too. IMO the comment is overcautious (d357bec8 doesn't even say that any such version of Bison was actually observed) and the benefits outweigh the risks.

vadz:
The crash in C# test is worrisome but I don't think it can be related to the change in this PR.

vadz:
Can this trivial change be merged?

ojwb:
It seems good to me.  Without this enabled, it's not clear to people that the existing SR conflicts are OK, but that a patch introducing new ones needs needs careful inspection.

The only evidence of bugs with `%expect` in bison seems to be a rather vague allusion in the documentation of byacc, which is a rival project to bison, so not an impartial source of information.  There's no version or time indicated for these alleged issues - even if correct when written, the information could be decades out of date.  The only way to really find out if there are any compatibility issues which are relevant is to enable this and try it.

Source distributions include the bison-generated files, so any problems won't affect end users building SWIG, only those working on SWIG.

vadz:
I am taking [this comment](https://github.com/swig/swig/commit/d357bec829d08ee96afe8ca8bf6d8631f6b983c9#commitcomment-17154766) as an approval and will commit this soon.

PR #481, opened 2015-07-27
Use JAVA_HOME to detect Java in configure.

vadz:
This conforms to the common Java practice and works better by default.

These commits are part of doxygen branch (see #170), but not really related to it, so I'm extracting them in a separate PR.


wsfulton:
Using JAVA_HOME is a good idea and this implementation looks good and it allows the old way to still work if JAVA_HOME is not set. I suggest a couple of minor tweaks:

- Convert JAVA_HOME to forward slashes on mingw using 'Tools/convertpath -u'.
- Only set JAVA_HOME on osx if /usr/libexec/java_home actually exists.
- Display JAVA_HOME during configure if this env variable is set and is a valid directory.

A point of note is if JAVA_HOME points to the non-jdk version, then javac will be missing for JAVA_HOME will be missing and if javac is actually in the path, it might be pointing to a different version to the one found in $JAVA_HOME/bin/java. I have actually come across set ups like this before! So potentially java and javac in the path might be matching versions and now we'd have a version mismatch when prior to the patch the selected java and javac versions might match. This will be an uncommon set up though and the solution would be to try parse the versions of the tools and I don't think that can be successfully done easily as javac doesn't always support -version. Anyway, after all that waffle, I don't think we can improve this aspect, and supporting JAVA_HOME is a step forward.

vadz:
OK, I addressed the issues mentioned (MinGW change is untested though) and also actually made it work under OS X where `java_home` doesn't return the headers location (but [Apple docs](https://developer.apple.com/library/mac/documentation/Java/Conceptual/Java14Development/05-CoreJavaAPIs/CoreJavaAPIs.html) explicitly say to hard code it to the value used, so who am I to argue...).

wsfulton:
Okay, nearly there I think. Couple more issues on OSX and the include dir. Even if Apple say it should be hard coded, I think a user should still have the flexibility of setting it to their own jdk if they have installed one elsewhere...

- The ```--with-javaincl``` option on OSX will now be ignored.
- Would be nice if this line used this location from earlier in the configure script
```
    *-*-darwin*) JAVAINCDIR="/System/Library/Frameworks/JavaVM.framework/Headers $JAVAINCDIR";;
```

vadz:
>  I think a user should still have the flexibility of setting it to their own jdk if they have installed one elsewhere...

Yes, sure, and AFAICS they have not one but two ways of doing it: either set `JAVA_HOME` to the location of their custom JDK (preferred) or use `--with-javaincl` option. Why do you say that the latter is ignored? It still sets `JAVAINCDIR` which overrides `JAVA_HOME`. Am I missing some bug here? This certainly was the intention...

> Would be nice if this line used this location from earlier in the configure script

Sorry, I'm lost. Which line is "this"?

wsfulton:
Ah sorry, it was late last night, I got it wrong and ```--with-java-incl``` will work. I was also trying to say that we should specify "/System/Library/Frameworks/JavaVM.framework/Headers" in just one place... not terribly important, we can leave as is.

PR #484, opened 2015-08-01
Make --without-alllang option more useful and use it for Travis.

vadz:
The last commit ("use it for Travis") is optional in the sense that I don't really care about it, I just wanted to speed up the Travis builds a little bit. But the main point is that this option is currently quite useless and counter-intuitive, I keep doing `--without-alllang --with-java` (for example) to test just Java stuff but it doesn't work like this. Or, rather, it didn't -- and with this change it does, which, I think, makes much more sense.


vadz:
NB: rebased the original commit on master to allow merging it automatically, there was a (trivial) conflict for the changes to `CHANGES.current`.

PR #487, opened 2015-08-05
Make (char*, size_t) typemap usable for strings of other types in Java.

vadz:
Notably it now works for "unsigned char*" strings.

Add a test to check that it now works in Java and also showing that it already
worked for the other languages with support for this typemap.

---
Note that I didn't find any other way to get rid of `const` than using `$1_basetype*`: if I used just `$1_type`, we had problems with `const T*` as `$1_type` had const while the actual type of the local variable did not (wrongly?).

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -4,3 +4,7 @@ See the RELEASENOTES file for a summary of changes in each release.
 
 Version 3.0.8 (in progress)
 ===========================
+
+2015-08-05: vadz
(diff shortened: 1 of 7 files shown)


vadz:
NB: AppVeyor errors are, as usual I'm tempted to say, bogus.

wsfulton:
I've asked appveyor for some help working around the 'nuget install' problems as every other build seems to fail due to this.

vadz:
The second version uses `$1_ltype` instead of `$1_basetype`.

PR #488, opened 2015-08-06
Extend the export feature in the CFFI module to support exporting to a particular package.

lyze:


lyze:
Added an explanation to the relevant part of the documentation.

The goal of this is to improve the usability of the "export" feature.

Currently, the only way to group the functions into some package for CFFI output is to do something along the lines of
````
%insert(lisphead) %{
(defpackage #:my-package)
(in-package #:my-package)
%}

%feature("export");
````
It is much more difficult to export a select few symbols to another package, since there is no way to change the package in the middle of the CFFI output. One might argue that it's probably better to split the wrapped functions in this case, but this is not always feasible. So I used the [export function in Common Lisp](http://clhs.lisp.se/Body/f_export.htm), which can has the ability to specify in which package the symbol should be external, so instead of `(cl:export '#.(my-lispify "lispsort_double"))`, we can now have generated output like this:
````
(cl:export '#.(my-lispify "lispsort_double" 'function) 'some-other-package)
````

PR #489, opened 2015-08-06
Enum r2015

richardbeare:
This is a modification to support use of tricky enumerations in R. It includes the addition of a _runme for an existing test - preproc_constants that was previously not run. That tests includes a preprocessor based setting of an enumeration which is ignored by the existing r enumeration infrastructure. The new version correctly reports the enumeration value as 4 - previous versions set it to 0. Traditional enumerations are unchanged.

One test that previously passed now fails - namely the R sourcing of preproc_constants.R, as the enumeration code requires the shared library, which isn't loaded by that script.

This work was motivated by problems generating the SimpleITK binding, specifically with some of the more complex enumerations.

The approach I've taken is similar to other languages. Enumerations that cannot be reliably detected by swig are initialised via a function call. Other languages model enumerations in a similar way.

This approach does have some issues wrt to code in packages, but I can't see an alternative. The problem with packages is that the R code setting up the enumeration structures requires the shared library so that the C functions returning enumeration values can be called. The enumeration setup code thus needs to be moved to the package initialisation section. For SimpleITK I do this using an R script, which I think is an acceptable solution. The core part of the process is the following function. I dump all the enumeration stuff into a .onload function. This is only necessary if some of the enumerations are tricky.

{
  p1 <- parse(file=filename)

  getdefineEnum <- function(X)
  {
  }

  enums <- p1[dd]

  ## put the function def around this

  everythingelse <- p1[!dd]
  writeLines(enums, onloadfile)

}


wsfulton:
Copying of unittest.R... this has fixed most of the Travis build problems for R which is excellent news. However, isn't there a better way... can't R find code using a path or something similar? I'd prefer not to see this kind of checking and copying of an R file for every test. If there is no path R can use, I think this file should be generated at configure time, so there would be a unittest.R.in file and configure would copy it to unittest.R in the build tree.

richardbeare:
Possibly - I wasn't sure of the test architecture. Will look into it in
more detail.

wsfulton:
Could you also please rebase this patch, it contains lots of reverts. Also the whitespace changes in r.cxx need to be in one separate commit to any code changes. I'm not sure why you have whitespace changes? If this file has whitespace that is not conforming, please make sure you use the SWIG beautifier...

```
cd Source
mkdir ../IndentBaks/Modules
make beautify-file INDENTFILE=Modules/r.cxx
```

and then there are usually some manual corrections to make.

richardbeare:
Thankyou - I was wondering about whitespace standards - I think I must have
hit emacs reindent at some point, and it is probably set up for ITK.

Source/Modules/r.cxx
```diff
@@ -1183,51 +1201,131 @@ int R::OutputArrayMethod(String *className, List *el, File *out) {
...
+  Delete(cppcode);
+
+  Delete(namespaceprefix);
+
+  Printv(scode, "defineEnumeration('", mangled_tdname, "'",
+         ",\n",  tab8, tab8, tab4, ".values = c(\n", NIL);
+
   Node *c;
   int value = -1; // First number is zero
+  bool NeedEnumFunc=false;
```
wsfulton: Variable names are usually lowercase.

richardbeare:
Not sure if I'll get a rebase correct since there's already been a merge -
will squashing commits do what you want?

ojwb:
This PR seems to be superseded by #500 - should it be closed now?

ojwb:
Closing as a duplicate - if I'm wrong, please reopen.

PR #491, opened 2015-08-07
Supress perl5 testcase warnings in clang

talby-:
clang finds some cast range checks "tautological" when Perl integer
sizes happen to line up during type conversions.

### DIFF
--- a/Examples/test-suite/perl5/wrapmacro_runme.pl
+++ b/Examples/test-suite/perl5/wrapmacro_runme.pl
@@ -1,7 +1,7 @@
 #!/usr/bin/perl
 use strict;
 use warnings;
-use Test::More tests => 5;
+use Test::More tests => 27;
 BEGIN { use_ok('wrapmacro') }
 require_ok('wrapmacro');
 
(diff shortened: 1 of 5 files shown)


talby-:
I'm going through testcases now to verify the behavior before and after this patch is fully covered.  If not I will append some tests to this work.

wsfulton:
Thanks. Let's leave merging until you have completed this then.

talby-:
I'm hitting a few surprises on master mostly emerging from the C strtol() function family:
 * an empty string is treated as zero.
 * a string containing only whitespace is a TypeError.
 * leading (but not trailing) whitespace in a string containing digits is ignored.
 * For the unsigned conversions, -1 is a TypeError, but '-1' is treated as ~0.

I think these should all produce a TypeError, but I can preserve existing behavior if a change seems dangerous.

wsfulton:
Presumably this is when a string is passed where a C numeric type is expected?  If so I think all of your cases should be a type error.

Except for the last one, surely it is either a type error or a number, it can't be both??

Keeping the behaviour the same has some merits, but the old behaviour strikes me as being quite wrong.
Given the changes you highlight, you'd better check C++ overloading where a method takes either a string or a numeric type as detecting the type correctly in the typecheck typemaps is important to get right.

talby-:
Great, I agree completely.  [overload_simple.i](https://github.com/swig/swig/blob/master/Examples/test-suite/overload_simple.i) and [wrapmacro.i](https://github.com/swig/swig/blob/master/Examples/test-suite/wrapmacro.i) seem to have the all the overloading situations I need to get good coverage.  I've been enhancing the run tests in those interfaces.

Examples/test-suite/perl5/wrapmacro_runme.pl
```diff
@@ -12,3 +12,40 @@
...
+
+# regression cases
+check("-11, ''", undef);
+check("0, ' 11'", undef);
+check("0, ' 11.0'", undef);
+check("-13, ' -11.0'", undef);
+check("0, \"11\x{0}\"", undef);
+check("0, \"\x{0}\"", undef);
+check("0, \"\x{9}11\x{0}this is not eleven.\"", undef);
+check("0, \"\x{9}11.0\x{0}this is also not eleven.\"", undef);
```
talby-: This block of tests are the ones that pass on master at the moment, plus `check("11, '-13'", 11)` due to the amusing behavior of strtoull("-1", NULL, 0).

talby-:
Well those test results were surprising.  I guess I shouldn't touch Lib/typemaps/fragments.swg.

PR #492, opened 2015-08-08
Surgical perlprimtypes fix

talby-:
this just adds additional coverage and the sizeof check short circuiting with none of the functional fixes

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -5,6 +5,9 @@ See the RELEASENOTES file for a summary of changes in each release.
 Version 3.0.8 (in progress)
 ===========================
 
+2015-08-07: talby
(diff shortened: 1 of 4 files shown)


wsfulton:
Shouldn't the CHANGES.current entry also mention the fixes in the conversion of incorrect numeric conversions mentioned in #491. If so let me know and I can make the changes during the merge.

talby-:
This patch does not include any functional changes.  #491 got failures in the AppVeyor checks so tried this smaller patch to see if that would help.  It apparently did not so I am trying to reproduce to failures AppVeyor is getting.

wsfulton:
Which Appveyor build are you concerned with? I don't see any Appveyor failures that are related to your changes. Note that Appveyor is generally unreliable and sometimes just hangs for no reason or has network problems, so the failures need a bit of interpretation.

PR #494, opened 2015-08-10
replacement for enumR2015 - tidied and rebased

richardbeare:
This is a modification to support use of tricky enumerations in R. Itincludes the addition of a _runme for an existing test - preproc_constants
that was previously not run. That tests includes a preprocessor based
setting of an enumeration which is ignored by the existing r enumeration
infrastructure. The new version correctly reports the enumeration value
as 4 - previous versions set it to 0. Traditional enumerations are unchanged.

The approach used to deal with these enumerations is similar to that of
other languages, and requires a call to a C function at runtime to return
the enumeration value. The previous approach figured out the values statically
and this is still used where possible. The need for a runtime call leads to
changes in when swig code is used in packages - see below.

One test that previously passed now fails - namely the R sourcing of
preproc_constants.R, as the enumeration code requires the shared library,
which isn't loaded by that script.

There is also a modification to the way the R _runme.R files are used.
The call to R CMD BATCH now includes a --args option that indicates
the source folder for the unittest.R file, and the first couple
of lines of the _runme.R files deal with correctly locating this.
Out of source tests now run.

This work was motivated by problems generating the SimpleITK binding,
specifically with some of the more complex enumerations.

This approach does have some issues wrt to code in packages, but I can't
see an alternative. The problem with packages is that the R code setting
up the enumeration structures requires the shared library so that the C
functions returning enumeration values can be called. The enumeration
setup code thus needs to be moved to the package initialisation section.
For SimpleITK I do this using an R script, which I think is an acceptable
solution. The core part of the process is the following function. I dump
all the enumeration stuff into a .onload function. This is only necessary
if some of the enumerations are tricky.

{
p1 <- parse(file=filename)

getdefineEnum <- function(X)
{
}

enums <- p1[dd]

everythingelse <- p1[!dd]
writeLines(enums, onloadfile)

}

### DIFF
--- a/Examples/test-suite/r/Makefile.in
+++ b/Examples/test-suite/r/Makefile.in
@@ -5,7 +5,7 @@
 LANGUAGE     = r
 SCRIPTSUFFIX = _runme.R
 WRAPSUFFIX   = .R
-RUNR         = R CMD BATCH --no-save --no-restore
+RUNR         = R CMD BATCH --no-save --no-restore '--args $(SCRIPTDIR)'
 
 srcdir       = @srcdir@
 top_srcdir   = @top_srcdir@
@@ -43,6 +43,7 @@ include $(srcdir)/../common.mk
 	$(setup)
 	+$(swig_and_compile_multi_cpp)
 	$(run_multitestcase)
+
 
 # Runs the testcase.
 #
--- a/Examples/test-suite/r/arrays_dimensionless_runme.R
+++ b/Examples/test-suite/r/arrays_dimensionless_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("arrays_dimensionless", .Platform$dynlib.ext, sep=""))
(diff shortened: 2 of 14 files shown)


wsfulton:
For the record, this patch is a follow on from #489. 

I asked for the whitespace and formatting changes in r.cxx, to be in a separate commit but now everything is in one big commit. The improved out-of-source test-suite changes should be in a separate commit too. I was also going to suggest that the enum / constant wrapping is controllable like it is in the other languages, but now I can't make out the actual code changes. Can you please point me to a patch with just the code changes for review?

Although the out-of-source test-suite fixes are a step forward, the test-suite is now failing for me in-source for a few tests (cpp_enum enum_thorough rename_pcre_enum as well as preproc_constants_c). These need fixing. The failures can be seen in the Travis tests that were run for this patch before it was merged (see 'View Details' above). Has this got something to do with the requiring of the shared library for enums now?

This was a premature merge given these problems.

richardbeare:
OK,
Back to the drawing board - Are we undoing the merge so I can resubmit, or
using some other strategy?

joequant:
Let me unmerge the fix.

wsfulton:
Further development of this patch is in #500

PR #500, opened 2015-08-11
Enum r2015 d

richardbeare:
This is a modification to support use of tricky enumerations in R. It
includes the addition of a _runme for an existing test - preproc_constants
that was previously not run. That test includes a preprocessor based
setting of an enumeration which is ignored by the existing r enumeration
infrastructure. The new version correctly reports the enumeration value
as 4 - previous versions set it to 0. Traditional enumerations are unchanged.

The approach used to deal with these enumerations is similar to that of
other languages, and requires a call to a C function at runtime to return
the enumeration value. The previous approach figured out the values statically
and this is still used where possible. The need for a runtime call leads to
changes in when swig code is used in packages - see below.

Several tests that previously passed now fail- such as the R sourcing of
preproc_constants.R, as the enumeration code requires the shared library,
which isn't loaded by that script. Old style enumerations always looked like this - 
with assignments using statically determined literals:
defineEnumeration("_itk__simple__KernelEnum", .values = c(sitkAnnulus = 0, 
    sitkBall = 1, sitkBox = 2, sitkCross = 3, sitkPolygon3 = 4, 
    sitkPolygon4 = 5, sitkPolygon5 = 6, sitkPolygon6 = 7, sitkPolygon7 = 8, 
    sitkPolygon8 = 9, sitkPolygon9 = 10))

New ones can look like this - with values determined using function calls to the wrapped code :
defineEnumeration('_MyEnum',
                    .values = c(
                        'kValue' = R_swigenum__MyEnum_kValue_get()
))

This kind of call will fail if the library isn't loaded, which is the case for the failing tests.

---

There is also a modification to the way the R _runme.R files are used.
The call to R CMD BATCH now includes a --args option that indicates
the source folder for the unittest.R file, and the first couple
of lines of the _runme.R files deal with correctly locating this.
Out of source tests now run correctly.

This work was motivated by problems generating the SimpleITK binding,
specifically with some of the more complex enumerations.

This approach does have some issues wrt to code in packages, but I can't
see an alternative. The problem with packages is that the R code setting
up the enumeration structures requires the shared library so that the C
functions returning enumeration values can be called. The enumeration
setup code thus needs to be moved to the package initialisation section.
For SimpleITK I do this using an R script, which I think is an acceptable
solution. The core part of the process is the following function. I dump
all the enumeration stuff into a .onload function. This is only necessary
if some of the enumerations are tricky.

{
p1 <- parse(file=filename)

getdefineEnum <- function(X)
{
}

enums <- p1[dd]

everythingelse <- p1[!dd]
writeLines(enums, onloadfile)

}

### DIFF
--- a/Examples/test-suite/r/Makefile.in
+++ b/Examples/test-suite/r/Makefile.in
@@ -5,7 +5,7 @@
 LANGUAGE     = r
 SCRIPTSUFFIX = _runme.R
 WRAPSUFFIX   = .R
-RUNR         = R CMD BATCH --no-save --no-restore
+RUNR         = R CMD BATCH --no-save --no-restore '--args $(SCRIPTDIR)'
 
 srcdir       = @srcdir@
 top_srcdir   = @top_srcdir@
@@ -43,6 +43,7 @@ include $(srcdir)/../common.mk
 	$(setup)
 	+$(swig_and_compile_multi_cpp)
 	$(run_multitestcase)
+
 
 # Runs the testcase.
 #
--- a/Examples/test-suite/r/arrays_dimensionless_runme.R
+++ b/Examples/test-suite/r/arrays_dimensionless_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("arrays_dimensionless", .Platform$dynlib.ext, sep=""))
 source("arrays_dimensionless.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/funcptr_runme.R
+++ b/Examples/test-suite/r/funcptr_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("funcptr", .Platform$dynlib.ext, sep=""))
 source("funcptr.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/ignore_parameter_runme.R
+++ b/Examples/test-suite/r/ignore_parameter_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("ignore_parameter", .Platform$dynlib.ext, sep=""))
 source("ignore_parameter.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/integers_runme.R
+++ b/Examples/test-suite/r/integers_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("integers", .Platform$dynlib.ext, sep=""))
 source("integers.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/overload_method_runme.R
+++ b/Examples/test-suite/r/overload_method_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("overload_method", .Platform$dynlib.ext, sep=""))
 source("overload_method.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/preproc_constants_runme.R
+++ b/Examples/test-suite/r/preproc_constants_runme.R
@@ -1,0 +1,11 @@
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
+dyn.load(paste("preproc_constants", .Platform$dynlib.ext, sep=""))
+source("preproc_constants.R")
+cacheMetaData(1)
+
+v <- enumToInteger('kValue', '_MyEnum')
+print(v)
+unittest(v,4)
+q(save="no")
--- a/Examples/test-suite/r/r_copy_struct_runme.R
+++ b/Examples/test-suite/r/r_copy_struct_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("r_copy_struct", .Platform$dynlib.ext, sep=""))
 source("r_copy_struct.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/r_legacy_runme.R
+++ b/Examples/test-suite/r/r_legacy_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("r_legacy", .Platform$dynlib.ext, sep=""))
 source("r_legacy.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/r_sexp_runme.R
+++ b/Examples/test-suite/r/r_sexp_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("r_sexp", .Platform$dynlib.ext, sep=""))
 source("r_sexp.R")
 cacheMetaData(1)
--- a/Examples/test-suite/r/rename_simple_runme.R
+++ b/Examples/test-suite/r/rename_simple_runme.R
@@ -1,4 +1,6 @@
-source("unittest.R")
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+
 dyn.load(paste("rename_simple", .Platform$dynlib.ext, sep=""))
 source("rename_simple.R")
 cacheMetaData(1)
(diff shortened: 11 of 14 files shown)


richardbeare:
I have not addressed the new failing tests. Awaiting discussion.

wsfulton:
The output in the director_protected testcase has changed and the enum is now missing. This doesn't seem right to me, can you check please?

```patch
--- ../../partialgoodts/r/director_protected.R	2015-08-11 19:11:56.361256241 +0100
+++ Examples/test-suite/r/director_protected.R	2015-08-11 19:13:10.313622951 +0100
@@ -361,11 +361,6 @@
 attr(`Bar_used`, "inputTypes") = c('_p_Bar')
 class(`Bar_used`) = c("SWIGFunction", class('Bar_used'))
 
-defineEnumeration('_Bar__Hello',
-                    .values = c(
-                        'hola' = 0, 
-                        'chao' = 1
-))
 # Start of new_Bar
 
 `Bar` = function()
```

I notice looking at the differences in enum_thorough.R that there is a problem
with %ignore and enum items. For example this has no effect in IgnoreA:

```swig
%ignore ignoreA_one;
```

Previously ignoreA_three was set to 0, now it should be set to ignoreA_zero + 3
(see csharp as reference output):

```
  public enum IgnoreA {
    ignoreA_zero,
    ignoreA_three = (ignoreA_zero) + 3,
    ignoreA_ten = 10,
    ignoreA_eleven,
    ignoreA_thirteen = (ignoreA_eleven) + 2,
    ignoreA_fourteen,
    ignoreA_twenty = 20,
    ignoreA_thirty = 30,
    ignoreA_thirty_two = (30) + 2,
    ignoreA_thirty_three
  }
```


```enum Foo``` in rename_enum_pcre.i is no longer being ignored now
with the change. Not only that, but it sets 'Foo_Max' = 1 when the actual
value is 2.

Another problem case:
```
enum Global { global0, global1, global2 = 1+2+3, global3 };
```
will wrap global2 = 1 which is wrong.

I suggest you convert Examples/test-suite/java/enum_thorough_runme.java to R
to ensure all enums are working.

If you use a GUI directory diff tool like meld of the output from 'make partialcheck-r-test-suite'
comparing before and after your change you'll see these.

I think you should use the same kind of logic that Java and C# do for wrapping
constants, that is, the value should always come from the C layer by default (if the enumvalue is set).
This can be changed via some equivalent to %javaconst(1) or %csconst(1)
... say %rconst(1) and similarly needs to apply to both constants and enum values. Also an equivalent to %javaconstvalue or %csconstvalue is needed for user control.

The names used in the C layer do not have the correct mangling either and can
lead to potential name clashes.  You have used a new function
getNamespacePrefix() when the correct approach is to use:
```c
Node *parent = parentNode(n);
String *symname = Getattr(n, "sym:name");
newsymname = Swig_name_member(0, Getattr(parent, "sym:name"), symname)
```
to set the name of the enum value. The C++ namespace should NOT be mangled in
unless %nspace is supported (which it isn't in R), but the class name should
be mangled in.

Use JAVA::enumValueDeclaration() as a reference.

Then, use this code from r.cxx:

```
String *wname = Swig_name_wrapper(iname);
Replace(wname, "_wrap", "R_swig", DOH_REPLACE_FIRST);
```
and tweak for your use, something like this:
```
String *wname = Swig_name_member(NSPACE_TODO, class_name, symname)
Replace(wname, "_wrap", "R_swigenum", DOH_REPLACE_FIRST);
```

As an aside, this isn't quite the way the name mangling should be working in r.cxx, it
should be using Swig_name_register("wrapper", "R_swig%f") and Swig_name_wrapper() instead of
Swig_name_wrapper() and Replace(). That could be
fixed in a separate patch and the output will probably be unchanged.

Anyway, if you want to check that the mangling of the C names is correct, use
the mangling that is similar to the other languages. Here is one testcase you
should use for checking the mangled C names:
```swig
%{
#define TWO 2
%}
%include <enumtypesafe.swg>
%inline %{
enum Global { global0, global1, global2 = TWO, global3 };
namespace Space {
  enum Spaced { spaced0, spaced1, spaced2 = TWO, spaced3 };
  struct KlassSpace {
    enum KlassySpace { klassyspace0, klassyspace1, klassyspace2 = TWO, klassyspace3 };
  };
}
struct Klass {
  enum Klassy { klassy0, klassy1, klassy2 = TWO, klassy3 };
};
%}
```
You'll see these C functions being generated for C#:
```
SWIGEXPORT int SWIGSTDCALL CSharp_global2_get() {
SWIGEXPORT int SWIGSTDCALL CSharp_spaced2_get() {
SWIGEXPORT int SWIGSTDCALL CSharp_KlassSpace_klassyspace2_get() {
SWIGEXPORT int SWIGSTDCALL CSharp_Klass_klassy2_get() {
```
R should be the same, but CSharp_ replaced with R_swigenum_.

You'll see JAVA::enumValueDeclaration is quite complex, but it does work
well, I'm afraid you'll need to use this logic. It will be simpler as you won't be supporting all the different enum wrapping approach (enumsimple.swg, enumtypesafe.swg, enumtypeunsafe.swg, enums.swg).

wsfulton:
I'm not sure what to do about testing the R tests that don't have a runme.R file. @joequant might have some ideas. My best suggestion is to just keep the behaviour the same and provide _runme.R files for the few failing files as a workaround.

richardbeare:
I have not addressed the new failing tests. Awaiting discussion.

Thanks for the thorough review! I hadn't realised how much was missing when I started trying to fill in the gaps.

Apart from the lack of emission of a protected enum (I was emitting only public ones - should have checked more carefully), I don't think I've broken anything that was working previously, as the examples that have changed were wrong anyway.

I'll start working through correcting the structure - will probably take a while.

wsfulton:
I have cherry picked f12481c, committed as 1a6d952094c47b1f2fcc2463615646db2770dc62 and finished off the out of source builds. This means that Travis should now successfully run the R tests for the first time ever, which is great as I've been wanting to do get this working for a while. You will need to re-enable the test in preproc_constants_runme.R for your reworked patch as I had to disable it to get master to pass.

richardbeare:
I'm working my way through the enum support in java and c# to see if a direct translation of parts of those is the best way to hit this. I'll probably be asking questions soon..

richardbeare:
I am slowly getting an understanding of how the java and csharp versions work. I am having trouble with the name mangling at present - I must be missing a step, but not sure what.

The call to Swig_name_member is always after a check for "scopedenum":

```{c}
if  ((enum_feature == SimpleEnum) && GetFlag(parent, "scopedenum"))
newsymname = Swig_name_member(0, Getattr(parent, "sym:name"), symname)
```
In my skeleton code the GetFlag call from inside enumvalueDeclaration never returns true. Is there a setting elsewhere that I need to make for this to happen for the R code?

wsfulton:
scopedenum is only set for c++ scoped enums, such as
```
enum class A { a, aa };
```

richardbeare:
I think I need a bit of advice regarding the bigger picture. I've managed to test the ideas behind the generation of mangled names. I've mostly been looking at java.cxx. The key methods are enumDeclaration, enumvalueDeclaration and enumValue.

They appear to mostly communicate via the enum_code variable which is used to store the java code for each enumeration as it is processed.

Seems that it isn't necessary to generate any special C code (I don't know java, so perhaps that is obvious).

I'm playing with getting r.cxx to emit C bindings to enumeration elements using the variableWrapper interface (obviously should be constantWrapper, but that isn't available for R yet). My problem is setting the names for the binding functions using the name mangling approach you outlined. What are the steps that should be used - I've played with the following, pieced together from bits of the java interface code.

```{c}
  if (!addSymbol(wname, n, scope)) {
    Printf(stdout, "error adding symbol\n");
    return SWIG_ERROR;
  }
  Setattr(n, "name", Getattr(n, "value"));
  variableWrapper(n);
```

I'm clearly missing something - not sure if I'm using the appropriate strategy at all, but hopefully close.
Thanks

richardbeare:
Still having trouble with emitting c wrappers from enumvalueDeclaration. At the moment my problem is getting an appropriately scoped reference. For example, if I use variableWrapper(n) from inside enumvalueDeclaration, the generated _get function looks like this (for the enum_thorough test):

SWIGEXPORT SEXP
R_swig_fast_get ( SEXP s_swig_copy)
{
  int result;
  unsigned int r_nprotect = 0;
  SEXP r_ans = R_NilValue ;
  VMAXTYPE r_vmax = vmaxget() ;
  
  result = (int)fast;
  r_ans = Rf_ScalarInteger(result);
  vmaxset(r_vmax);
  if(r_nprotect)  Rf_unprotect(r_nprotect);
  
  return r_ans;
}

I have no luck getting any class/scoping information into wrapper function. I'm unclear on whether I should be writing a new generator function, or whether I can use variableWrapper in conjunction with some magic, like adding symbols to scopes etc.

Apologies - still learning the right way of doing things.

wsfulton:
Have you tried debugging Java for this enum? The default enum handling will use the class name in the scoping. Although the full name will actually be Java_enum_1thorough_enum_1thoroughJNI_SpeedClass_1fast_1get, it doesn't matter, you'll see it uses the Swig_name functions to get the C name and you just need to set up a more simple name mangling.

richardbeare:
Making progress now. Haven't had success changing the wrapped names, but don't think it matters. Have used the right Swig_name functions to retrieve the names, I think. Should have something to look at soon.

richardbeare:
Hi,
Getting back to this now, but running into issues when using C compilers. I think the name mangling is at least approximately correct. My problem is that the _get function generated using variableWrapper from inside enumValue looks like this (for enums_wrap):

```{c}
SWIGEXPORT SEXP
R_swig_foo3_ABCDE_get ( SEXP s_swig_copy)
{
  foo2 result;
  unsigned int r_nprotect = 0;
  SEXP r_ans = R_NilValue ;
  VMAXTYPE r_vmax = vmaxget() ;
  
  result = (foo2)ABCDE;
  r_ans = SWIG_From_int((int)(result));
  vmaxset(r_vmax);
  if(r_nprotect)  Rf_unprotect(r_nprotect);
  
  return r_ans;
}
```
but I need it to be using C forms:

```{c}
enum foo2 result;

result = (enum foo2)ABCDE;
```
It seemed that using variableWrapper to create the wrapper functions was a sensible thing to do, but perhaps it isn't.

So, is there a step I'm missing to get the correct type of code emitted, or should I be using an alternative strategy.

Thanks

richardbeare:
Superseded by PR #1328

Issue #418, opened 2015-05-10
[Go] How to call a C++ method from within a Go method on a director?


michael-schaller:
Maybe I'm missing something essential here but how do I call a C++ method from within a Go method on a director?

According to the SWIG documentation (23.4.7) I would assume that this is possible. Quote from the documentation:
*"The Go code may of course call other methods on itself, and those methods may be defined either in Go or in C++."*

I've created a simple example. The full source can be found here:
https://github.com/michael-schaller/go-swig-examples/tree/master/director-example1

The example is based around this abstract C++ class:
``` C++
class Progress
{
	public:
		Progress(int s) : total_steps(s) { steps_done = 0; };
		virtual ~Progress() {};

		void StepDone() { steps_done++; OnStepDone(); };
		float Percent() { return float(steps_done) / float(total_steps) * 100.0; };
		virtual void OnStepDone() = 0;

	private:
		int total_steps;
		int steps_done;
};
```

If I test this code in C++ I end up with this C++ code:
``` C++
class CppProgress : public Progress
{
	public:
		CppProgress(int s) : Progress(s) {};
		virtual ~CppProgress() {};

		virtual void OnStepDone() { printf("Current percent: %f\n", this->Percent()); fflush(stdout); };
};

void ProgressTest(Progress *p, int steps) {
	for (int step = 0; step < steps; step++) {
		printf("Step %d done.\n", step + 1);
		fflush(stdout);
		p->StepDone();
	}
}
```

The output of the C++ test is:
```
Step 1 done.
Current percent: 20.000000
Step 2 done.
Current percent: 40.000000
Step 3 done.
Current percent: 60.000004
Step 4 done.
Current percent: 80.000000
Step 5 done.
Current percent: 100.000000
```

If I try to implement similar Go test code then I wonder how I would access the C++ Percent() method. My example Go code looks like this:
``` Go
type GoProgress struct{}

func (p *GoProgress) OnStepDone() {
	// How to call Percent()? Percent is not a method on GoProgress.
	fmt.Printf("Current percent: %f\n", 0.0)
}

func MakeProgress(steps int) Progress {
	return NewDirectorProgress(&GoProgress{}, steps)
}
```

@ianlancetaylor 
Ian, could you shed some light on this? How can I call a C++ method from within a Go method on a director? Is the documentation wrong or did I miss something essential?

michael-schaller:
I've played a bit with this problem and the cleanest solution I could come up with so far is this one:
``` Go
[...]

PR #502, opened 2015-08-21
[Go] Reworked beginning of the documentation.

michael-schaller:
* Removed link to examples in the Go source tree as discussed in issue #418.
* Reworded occurences of the 'gc tool' as it has been removed with Go 1.5.
* Reworked chapter 23.3.  This should make it easier for users to get started
  with SWIG as the chapter starts with how to use SWIG with the go tool.
* Added helpful links.


michael-schaller:
@ianlancetaylor Please review.

Doc/Manual/Go.html
```diff
@@ -60,38 +60,44 @@
...
 
 <p>
-There are (at least) two different Go compilers.  One is the gc
-compiler, normally invoked via the go tool.  The other
-is the gccgo compiler, which is a frontend to the gcc compiler suite.
-The interface to C/C++ code is completely different for the two Go
-compilers.  SWIG supports both, selected by a command line option.
+There are (at least) two different Go compilers.  The first is the Go compiler
+of the <a href="https://golang.org/doc/install">Go distribution</a>.
+Since Go 1.5 the Go compiler is part of the <a href="https://golang.org/cmd/go/">
```
ianlancetaylor: These lines aren't accurate.  The only change in 1.5 is a change in the name of the compiler, which is an unimportant detail for most people.  I think the existing sentence is accurate: "One is the gc compiler, normally invoked via the go tool."  You can rework that to refer to the Go distribution, but there shouldn't be any notes about the compiler changing in 1.5.

Doc/Manual/Go.html
```diff
@@ -99,12 +105,83 @@
...
+options the go tool uses run <tt>go build -x</tt>.  To access the automatically
+generated files run <tt>go build -work</tt>.  You'll find the files under the
+temporary WORK directory.
+</p>
+
+<p>
+To manually generate and compile C/C++ wrapper code for Go, use the <tt>-go</tt>
+option with SWIG.  By default SWIG will generate code for the Go compiler of the
+Go distribution.  To generate code for gccgo, you should also use the <tt>-gccgo
+</tt> option.
```
ianlancetaylor: Should this </tt> be at the end of the preceding line?

Doc/Manual/Go.html
michael-schaller: The issue for me is that the 'gc' tool/binary is gone with Go 1.5.  There is no more 'go tool gc' and the link https://golang.org/cmd/gc/ is dead.  'gc' itself still exists but it is now part of the internal package under 'cmd/compile/internal'.  So from an user perspective 'gc' is gone and IMHO we should no longer talk about 'gc' if it is something that is no longer usable by users.  ;-)

Anyhow... What would you write?  How about "One is the Go compiler of the Go distribution." and leave the "Since Go 1.5 ..." part out?  I personally would like to give a link with additional information and so a link to 'go build' seems to be most appropriate to me.  What do you think?

Should I update the Examples/go/index.html document as well as it still mentions the gc compiler?
https://htmlpreview.github.io/?https://github.com/swig/swig/blob/master/Examples/go/index.html

ianlancetaylor: Well, there was never a gc binary.  There was a 6g binary, an 8g binary, etc.  In 1.5 those have all been combined into a single compile binary.  So where you used to run "go tool 6g" you can now run "go tool compile".

But when discussing Go compilers, the overall name of the compiler that is part of the standard Go distribution is "gc", whereas the name of the compiler that is part of the GCC distribution is "gccgo".

I feel like the original text of this paragraph is accurate.  It's fine to add some links but I don't see a great need to change the text.  In particular nothing in the original paragraph is different in the 1.5 release.

michael-schaller: Done.

Doc/Manual/Go.html
michael-schaller: Done.

michael-schaller:
@ianlancetaylor PTAL

ianlancetaylor:
LGTM.  Thanks!

michael-schaller:
Thanks, Ian.

PR #506, opened 2015-08-25
Removed golang stringing for signed/unsigned char

### LINKED ISSUE #505: Default golang typemaps treat signed/unsigned char* as string
I'm wrapping a library that makes heavy use of unsigned char arrays as arrays of 8-bit numeric values which often include zeroes.

The default golang handling of these is to treat them as a zero-terminated string: the same special handling that a char* gets. This leads to a lot of trouble wrapping it properly, as these are not really strings and the normal response of "tell swig to treat it like an explicitly signed or unsigned char" doesn't help.

It seems to me that the least surprising (and likely intended) default behavior would be to treat a char* as a string (as it already does) but do no special string handling for an unsigned char* or signed char*.

All tests still appear to pass when the behavior is changed to how I suggest, which makes me believe the current behavior was a minor oversight rather than intended functionality.

demi-rluddy:
With this change, generated code for golang treats char* as a string but
treats signed char* and unsigned char* as normal pointers. This seems to
fit better with the expected behavior, as the latter are more often used
as non-string data.

Resolves #505

### DIFF
--- a/Lib/go/go.swg
+++ b/Lib/go/go.swg
@@ -426,57 +426,41 @@
 /* Strings.  */
 
 %typemap(gotype)
-	char *, char *&, char[ANY], char[],
-	signed char *, signed char *&, signed char[ANY], signed char[],
-	unsigned char *, unsigned char *&, unsigned char[ANY], unsigned char[]
-"string"
+	char *, char *&, char[ANY], char[] "string"
 
 /* Needed to avoid confusion with the way the go module handles
    references.  */
-%typemap(gotype) char&, unsigned char& "*byte"
-%typemap(gotype) signed char& "*int8"
+%typemap(gotype) char& "*byte"
 
 %typemap(in)
-	char *, char[ANY], char[],
-	signed char *, signed char[ANY], signed char[],
-	unsigned char *, unsigned char[ANY], unsigned char[]
+	char *, char[ANY], char[]
 %{ $1 = ($1_ltype)$input.p; %}
 
-%typemap(in) char *&, signed char *&, unsigned char *&
+%typemap(in) char *&
 %{ $1 = ($1_ltype)$input.p; %}
 
 %typemap(out,fragment="AllocateString")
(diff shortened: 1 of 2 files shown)


wsfulton:
I think this patch should be applied as corrects the handling to be consistent with the rest of SWIG, as per my comments in #505. @ianlancetaylor can you merge if you agree please? It'll require a note in the CHANGES file with a * POTENTIAL INCOMPATIBILITY * note and info on how to restore the old behaviour (using %apply).

Lib/go/go.swg
```diff
@@ -426,57 +426,41 @@
...
 %typemap(gotype)
-	char *, char *&, char[ANY], char[],
-	signed char *, signed char *&, signed char[ANY], signed char[],
-	unsigned char *, unsigned char *&, unsigned char[ANY], unsigned char[]
-"string"
+	char *, char *&, char[ANY], char[] "string"
 
 /* Needed to avoid confusion with the way the go module handles
    references.  */
-%typemap(gotype) char&, unsigned char& "*byte"
```
ianlancetaylor: This part of the change doesn't seem right.  I understand the point about signed char *, but this is signed char&.  It's not being treated as a string, it's being treated as a pointer.

ianlancetaylor:
@demi-rluddy Can you add a note to CHANGES as wsfulton suggests?

Lib/go/go.swg
demi-rluddy: Good catch, it looks like a mistake - I'll remove that part.

demi-rluddy:
Updated as mentioned about signed char& and unsigned char&; CHANGES updated with note about potential incompatibility.

ianlancetaylor:
Sorry for letting this sit.  This looks good but it needs to be updated to avoid a merge conflict.

wsfulton:
There was just a simple conflict in the CHANGES.current file which I resolved and so I've merged this.

PR #509, opened 2015-08-28
Do not use bare exception in generated Python code.

sjoerdjob:
By using the 'except:', you can catch all kinds of exceptions, including
the KeyboardInterrupt and SystemExit exceptions. From the generated
code, it is quite obvious that it is not these cases that should be
caught, but more specific ones like AttributeError and TypeError. To be
on the safe side, I decided to keep using 'Exception' for now.


wsfulton:
Looks okay to me.

PR #511, opened 2015-08-28
Add new but ignored warning option -Wdate-time.

Bluehorn:
Rationale: This would be helpful for the Debian reproducible builds
effort, described at https://wiki.debian.org/ReproducibleBuilds

Basically, the warning is used to detect where the time macros
`__TIME__`, `__DATE__` etc. are used, but this makes SWIG fail because
of the unrecognized option.

Please see my email on swig-devel and [this feature request](https://bugs.debian.org/790102) for more background.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -661,6 +661,8 @@ MULTI_CPP_TEST_CASES += \
 
 # Custom tests - tests with additional commandline options
 wallkw.cpptest: SWIGOPT += -Wallkw
+wdate_time.ctest: SWIGOPT += -Wdate-time
+
 preproc_include.ctest: SWIGOPT += -includeall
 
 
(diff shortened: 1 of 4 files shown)


wsfulton:
@ojwb, Agreed, but this is a non-ideal but pragmatic change, see the swig-devel discussion on this at http://thread.gmane.org/gmane.comp.programming.swig.devel/23717. Could you respond to that discussion. Also, as you know Debian quite well, perhaps you can guide the appropriate project to sort it out properly?

ojwb:
I'd be much more sold on a wrong but pragmatic change if there were many affected packages, but only two have been mentioned so far.  So to me, the appropriate fix seems to be patching the build systems of the Debian subversion and unbound packages to not do `swig $(CPPFLAGS) [...]` (or equivalent).

Will follow up in that thread.

ojwb:
Done.

Bluehorn:
Closing this pull request, I concur with Ollys arguments.

PR #512, opened 2015-08-29
Integrate osx .travis.yml into master branch using multi-os feature.

DarthGandalf:
http://docs.travis-ci.com/user/multi-os/


vadz:
FWIW I think the `before_install` commands need to be extracted into their own separate script file, it was already messy before but it is even worse now. Keeping everything in `.travis.yml` makes it completely unmaintainable.

wsfulton:
I'm okay with all the code remaining in .travis.yml, personally I think it is easier to follow.

I can't see why the OS X builds failed. Also puzzling me is why the Ruby build has director_exception errors when the last time it ran on OS X it didn't (using the same versions of Ruby and clang from what I can see).

wsfulton:
I think it is clearly an improvement to run both OS X and Linux tests on each commit but, yikes, with this change a whopping 49 Travis builds are run!

DarthGandalf:
I've moved `install` to separate files, I think it looks better now too, rather than inline.

DarthGandalf:
The builds are pretty small, so some of them can be combined. E.g. test perl and octave (or whatever) together in a single build.

DarthGandalf:
I've fixed everything except ruby. I don't know what's happening to it, but one of builds randomly succeeded. https://travis-ci.org/swig/swig/jobs/77838778

wsfulton:
That build didn't actually work as intended. There is an error:

```
0.00s$ if test -n "$SWIGLANG"; then export cxxflags=$(Tools/testflags.py --language $SWIGLANG --cxxflags --std=$CPPSTD --compiler=$CC) && echo $cxxflags; fi

  File "Tools/testflags.py", line 77, in <module>

    print(get_flags(args.language, args.std, args.compiler))

  File "Tools/testflags.py", line 52, in get_cxxflags

    cflags["guile"] += " -Wno-attributes" # -Wno-attributes is for clang LLVM 3.5 and bdw-gc < 7.5 used by guile

NameError: global name 'cflags' is not defined
```
Clearly this needs fixing, but more importantly, this failure must result in an exit code that fails the build on Travis. Probably your splitting of .travis.yml needs some more thought fine tuning.

I'll see if I can fix the Ruby problem, note that the build didn't fail because it was missing -Werror, the warning is still there.

The builds for each language should remain separate, I'd rather it was nicely modularised so that it is easier to see what has failed and combined builds wouldn't do this.

DarthGandalf:
> Clearly this needs fixing

Ah, yes, I didn't realize why it succeeded. But this error was fixed already anyway.

> this failure must result in an exit code that fails the build on Travis

Of course. The same error could happen before this change though.
Fixed.

DarthGandalf:
Eh... https://travis-ci.org/swig/swig/jobs/77912889 proves that error in that script causes travis to fail.
So I've added R to testflags.py too.

wsfulton:
I just spotted PY3 on OS X isn't running... https://travis-ci.org/swig/swig/jobs/77923988
```
skipping python test-suite check
```

DarthGandalf:
> skipping python test-suite check

Fixed. It was `PY3=1` instead of `PY3=3`.
Also now it will fail such builds. https://travis-ci.org/swig/swig/jobs/78140814

wsfulton:
Can you restore the gcc-4.2 testing on osx? I don't believe it is an alias to clang as can be seen from the builds in travis-osx branch:

```
$ $CC --version
i686-apple-darwin11-gcc-4.2.1 (GCC) 4.2.1 (Apple Inc. build 5666) (dot 3)
Copyright (C) 2007 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

wsfulton:
ensure-$SWIGLANG-enabled is a good idea. Let's rename it check-$SWIGLANG-enabled to be consistent as just about everything else in that makefile is a check too.

DarthGandalf:
Ah, I didn't realize it's 4.2.
`gcc` is alias for clang, but gcc-4.2 is not

wsfulton:
I see clang is still being used instead of gcc-4.2. This feels like something we ought ask Travis about how it should be configured. In the meantime I guess we'll have to use an environment variable like GCC5 to work around this.

DarthGandalf:
Yep, it overrided CC from env with gcc. Now it should be working, with a separate variable.

DarthGandalf:
ping

wsfulton:
Mmm, strange when I looked a few days ago gcc-4.2 still wasn't working, but it is now. Maybe you rebased it? Anyway, we seem to have two different approaches to a non-standard compiler. We should use just one approach, so either use GCC5 or the new SWIG_CC approach. Probably SWIG_CC is better for consistency and adding other compilers compared to defining various special env variables for different compilers.

DarthGandalf:
The last time travis for some reason didn't pick up an update, so I amended last commit with no change, pushed again, and travis saw it. Probably you were looking at the old build which was built before the last fix...
Ok, I'll switch GCC5 to SWIG_CC.

DarthGandalf:
Done.

Issue #519, opened 2015-09-14
[javascript][node] incorrect exception handing with function overload


microcai:
in generated code,
swig uses  args.Length() to dispatch different overloaded functions, but when exception occurs, 
```
    if(errorHandler.err.IsEmpty()) {
      return;
    }
```
and the execution falls to 

```
  SWIG_exception_fail(SWIG_ERROR, "Illegal arguments for function xxxx");
```

therefore, when ever exceptions thrown, it's always  "Illegal arguments" .

ojwb:
Now covered by PR #520, closing.

ojwb:
Oops, didn't actually close this - doing so now.

PR #520, opened 2015-09-14
[javascript][node] incorrect exception handing with function overload

microcai:
this fix https://github.com/swig/swig/issues/519:

in generated code,
swig uses  args.Length() to dispatch different overloaded functions, but when exception occurs, 

```
    if(errorHandler.err.IsEmpty()) {
      return;
    }
```

and the execution falls to 

```
  SWIG_exception_fail(SWIG_ERROR, "Illegal arguments for function xxxx");
```

therefore, when ever exceptions thrown, it's always  "Illegal arguments" .

### DIFF
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -334,9 +334,13 @@ fail:
 #else
     $jswrapper(args, errorHandler);
     if(errorHandler.err.IsEmpty()) {
-      return;
+      SWIGV8_RETURN(jsresult);
+    }
+    else {
+      SWIGV8_RETURN(SWIGV8_UNDEFINED());
     }
 #endif
+    SWIGV8_RETURN(SWIGV8_UNDEFINED());
   }
 %}
 


wsfulton:
This pull request has broken the javascript tests on Travis. Please address these with an update. Can you please also put in a description of what this fixes in the title and description of the pull request. A testcase is also needed to show what it fixes.

ojwb:
I've closed #519 (having both an issue and a PR open isn't useful), and copied over its title and description to here.

microcai:
updated ;)

Lib/javascript/v8/javascriptcode.swg
```diff
@@ -334,9 +334,13 @@ fail:
 #else
     $jswrapper(args, errorHandler);
     if(errorHandler.err.IsEmpty()) {
-      return;
+      SWIGV8_RETURN(jsresult);
+    }
+    else {
+      SWIGV8_RETURN(SWIGV8_UNDEFINED());
```
ojwb: I think you don't want the `else` here, as it makes the addition below unreachable code in the case where this `#else` branch is used.

ojwb:
@microcai Any response to the comments above?

wsfulton:
Possibly related to #832.

ojwb:
There is a problem here but the patch here doesn't seem right and the submitter seems to have abandoned it so closing.  We have #772 open for tracking this bug.

PR #522, opened 2015-09-15
Remove unused support for typemap scopes

vadz:
While debugging some typemap problem, I noticed that the code in `typemap.c` was unnecessarily complicated as it supported typemap scopes which are, in fact, never used, so I've just removed them to simplify it.


vadz:
NB: AppVeyor C# errors are bogus, the Python/x64 one is real but doesn't seem to have anything to do with these changes.

wsfulton:
The patch looks fine to me, I'll merge. 

I've restarted Appveyor tests for this patch but C# is still failing, but I'm not sure why. The python x64 windows build is marked an expected failure.

PR #524, opened 2015-09-16
lua: add typemap support for ssize_t

dafugg:
Simple addition of ssize_t support.

### DIFF
--- a/Lib/lua/luatypemaps.swg
+++ b/Lib/lua/luatypemaps.swg
@@ -391,6 +391,9 @@ parameters match which function
 %apply unsigned long { size_t };
 %apply const unsigned long & { const size_t & };
 
+// ssize_t (which is just a signed long)
+%apply long { ssize_t };
+%apply const long & { const ssize_t & };
 
 /* -----------------------------------------------------------------------------
  *                          Specials


Lib/lua/luatypemaps.swg
```diff
@@ -391,6 +391,9 @@ parameters match which function
 %apply unsigned long { size_t };
 %apply const unsigned long & { const size_t & };
 
+// ssize_t (which is just a signed long)
```
vadz: This is not really correct, under Win64 `ssize_t` is 64 bits while long is 32.

dafugg:
@vadz it looks like the existing size_t typemap suffers from a similar size mismatch under win64. Can you suggest a method for handling this? Perhaps tests are needed however I don't have a windows system with which to develop them.

vadz:
I think the simplest would be to just always map `[s]size_t` to `[u]int64_t`. This is not optimal for 32 bit architectures, of course, but the only way to avoid it would be to use conditional compilation in the generated C code. And while I guess this could be done for Lua (there is no Lua generated component at all for it, is there?), it could be problematic for the other languages...

BTW, it looks like no other language handles `ssize_t` neither currently, it would probably make sense to add it to some common file under `Lib`.

wsfulton:
I don't see ssize_t in either of the C or C++ standards, so I don't know that SWIG needs to be provide typemaps for this. It looks like it is a posix type. I couldn't find the docs on this. As @vadz said, it would need to go into a  new .i library file for other posix types. The file ought to be common across all target languages.

I'm not sure what to do about the incorrect mapping of size_t on windows 64.

ojwb:
@wsfulton POSIX types are documented here http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html (at least those in `sys/types.h` are).

FWIW, that says:

> The implementation shall support one or more programming environments in which the widths of blksize_t, pid_t, size_t, ssize_t, and suseconds_t are no greater than the width of type long.

Mapping `size_t` to a wider type than necessary doesn't seem a great plan - that would be punishing other platforms for Microsoft's choices.

vadz:
`ssize_t` is indeed POSIX and not ANSI C, but it's still pretty common, so I think it could be worth including typemaps for it.

As for `sizeof(size_t)`, I really don't see how else can it be handled for the backends that generate files in non-C languages if we want to preserver (the nice, IMO) property of SWIG that it produces the same output independently of the platform it runs on. E.g. taking example of Java, SWIG must generate either `int` (32 bits) or `long` (64 bits) in the Java code and only the latter will work under all platforms.

ojwb:
Providing `posix.i` as somewhere for people to contribute typemaps for the common POSIX types certainly seems useful to me.

As for the `sizeof(size_t)` issue, I don't currently see a nice alternative, but forcing wider types for the benefit of (AIUI) one proprietary platform seems a shame.  If we do that we should at least have a simple way to say you don't care about supporting it.

wsfulton:
Please adapt the patch as suggested, or close it. Thanks.

dafugg:
From my reading of the comments this would need to be part of a larger POSIX layer with an additional method to tell SWIG that we don't care about generating platform-agnostic code. I'll close this and let the mailing list sort out design.

PR #528, opened 2015-09-22
Doc/Manual/CSharp.html: remove irrelevant text

xypron:
The reference "for warning 845" does not make any sense in this
context.

Signed-off-by: Heinrich Schuchardt <xypron.glpk@gmx.de>

### DIFF
--- a/Doc/Manual/CSharp.html
+++ b/Doc/Manual/CSharp.html
@@ -1272,7 +1272,7 @@ the C# code can be generated into the intermediary class using the <tt>imclassco
   }
 
   // Note that SWIG detects any method calls named starting with
-  // SWIG_CSharpSetPendingException for warning 845
+  // SWIG_CSharpSetPendingException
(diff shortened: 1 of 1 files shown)


ojwb:
I'm no expert on the C# backend, but earlier in the C# chapter it says:

> Note that if SWIG detects you have used `SWIG_CSharpSetPendingException()` or `SWIG_CSharpSetPendingExceptionArgument()` without setting the `canthrow` attribute you will get a warning message similar to
>
> ```
example.i:21: Warning 845: Unmanaged code contains a call to a SWIG_CSharpSetPendingException
method and C# code does not handle pending exceptions via the canthrow attribute.
```
>
> Actually it will issue this warning for any function beginning with `SWIG_CSharpSetPendingException`.

So the reference to warning 845 you're suggesting removing seems relevant to me.

wsfulton:
As @ojwb points out, the original looks good to me.

PR #536, opened 2015-09-30
(C#) Bug Fix: Rename CTYPE to CSTYPE

ezralanglois:
Symptom: Using the current typemaps would not work for byte arrays. 

Turns out, they were misnamed.
They should have been named by the C# type and not the C-type.


ezralanglois:
Looks like fixing this broke the string typemaps and some int/long array typemap.

A python unit test failed, but that is probably unrelated.

ezralanglois:
These are not very platform independent:

CSHARP_ARRAYS_FIXED(long, int)
CSHARP_ARRAYS_FIXED(unsigned long, uint)
CSHARP_ARRAYS_FIXED(long long, long)
CSHARP_ARRAYS_FIXED(unsigned long long, ulong)

wsfulton:
I'm afraid this is not right. Typemaps are specified using C/C++ types not C# types. Am closing as there is no description nor test showing what the actual problem is that the patch is meant to fix.

ezralanglois:
I see. Sorry for the static. I guess making it a CTYPE is a convention I should have followed. It really does not seem to matter whether you use the CSTYPE or CTYPE as long as you are consistent.

I will add a small unit test example of what I was trying to accomplish and then rework and resubmit the pull request.

PR #542, opened 2015-10-09
Fix Go example in Go.html

Sajmani:
Missing "func" keyword.


ojwb:
Thanks - sorry this has sat for a while.

PR #544, opened 2015-10-20
%ignore should imply %nodirector.

LindleyF:
My team recently discovered that using the %ignoreall, %rename pattern (http://www.swig.org/Doc3.0/SWIGDocumentation.html#SWIG_chosen_unignore) does not properly ignore director methods.

This change attempts to fix this. There may be a better way, this was just the first thing that seemed to work.

### DIFF
--- a/Source/Modules/lang.cxx
+++ b/Source/Modules/lang.cxx
@@ -1949,7 +1949,8 @@ int Language::unrollVirtualMethods(Node *n, Node *parent, List *vm, int default_
       Node *m = Getattr(item, "methodNode");
       /* retrieve the director features */
       int mdir = GetFlag(m, "feature:director");
-      int mndir = GetFlag(m, "feature:nodirector");
+      int mndir = GetFlag(m, "feature:nodirector") ||
+                  GetFlag(m, "feature:ignore");
       /* 'nodirector' has precedence over 'director' */
       int dir = (mdir || mndir) ? (mdir && !mndir) : 1;
       /* check if the method was found only in a base class */
@@ -1958,7 +1959,8 @@ int Language::unrollVirtualMethods(Node *n, Node *parent, List *vm, int default_
 	Node *c = Copy(m);
 	Setattr(c, "parentNode", n);
 	int cdir = GetFlag(c, "feature:director");
-	int cndir = GetFlag(c, "feature:nodirector");
+	int cndir = GetFlag(c, "feature:nodirector") ||
+                    GetFlag(c, "feature:ignore");
 	dir = (cdir || cndir) ? (cdir && !cndir) : dir;
 	Delete(c);
       }
@@ -2095,7 +2097,8 @@ int Language::classDirectorMethods(Node *n) {
     Node *item = Getitem(vtable, i);
     String *method = Getattr(item, "methodNode");
     String *fqdname = Getattr(item, "fqdname");
-    if (GetFlag(method, "feature:nodirector"))
+    if (GetFlag(method, "feature:nodirector") ||
+        GetFlag(method, "feature:ignore"))
       continue;
 
     String *wrn = Getattr(method, "feature:warnfilter");
@@ -2455,7 +2458,8 @@ int Language::classDeclaration(Node *n) {
   if (!ImportMode) {
     if (directorsEnabled()) {
       int ndir = GetFlag(n, "feature:director");
-      int nndir = GetFlag(n, "feature:nodirector");
+      int nndir = GetFlag(n, "feature:nodirector") ||
+                  GetFlag(n, "feature:ignore");
       /* 'nodirector' has precedence over 'director' */
       dir = (ndir || nndir) ? (ndir && !nndir) : 0;
     }


wsfulton:
From swig-user mailing list...
On 21 October 2015 at 19:17, Lindley French wrote:
> My team recently found that using the "ignore-everything, then unignore
> specific things" pattern does not work properly for directors. Director
> methods are instantiated even if they are not explicitly unignored. This is
> a problem if the method use types that are not buildable by SWIG.
>
> http://www.swig.org/Doc3.0/SWIGDocumentation.html#SWIG_chosen_unignore
>
> I have a possible fix out in https://github.com/swig/swig/pull/544, but it's
> choking on the director_ignore test. I understand why it's choking; it's
> clearly invalid to ignore a pure-virtual method on a class that you want to
> act as a director.
>
> What I don't understand is what the director_ignore test is trying to prove.
> Why is it trying to %ignore pure-virtual methods at all? I don't understand
> the motivation behind the test as-written. Anyone have any insights?

This test has been built up and expanded over quite a few commits. I dare say, with a bit of effort you could trace some of the changes back to bug reports on the old SF system or emails sent to the mailing lists, or I could have just made up the tests as I fixed things. I'm not sure I have any great insight, other than it might be handy for a user to effectively remove a function from an abstract base class for use from a chosen target language. A default behaviour can be provided by providing an implementation in a derived class and then wrapping that and ignoring the method.

Presumably you have a problem that the C++ director class has to have an implementation for all virtual methods in order for the resulting C++ to compile? You'll note that a runtime error is generated for all languages if the derived class does not override the method, even strongly typed ones like Java. I guess for your case, given all methods are %ignore'd, you still need to generate C++ code where the director class can be instantiated and that the C++ code compiles - that seems reasonable to me.

The patch currently removes %ignore tests, these should be restored to ensure there are no regressions.

William

LindleyF:
Effectively, the change here is to assume that %ignore implies %nodirector.

A runtime error makes sense when you don't know what methods the overriding class is declaring until introspection at runtime. It doesn't make sense at all to not even give the overriding class the *opportunity* to override a pure-virtual method, though, because there would be no way to create a non-abstract class. This certainly should cause a compile error. That's why I removed the %ignores on pure-virtual methods in the test.

Now, the issue that's bothering me is that it *might* make some amount of sense to allow ignoring a method for Java-to-C++ calls but not for C++-to-Java director calls. This is what we lose by assuming that %ignore implies %nodirector. I can't think of a case where this is useful offhand, but that doesn't mean one can't exist. I think it's more intuitive for %ignore to ignore things in both directions, though.

wsfulton:
I don't understand what is not being properly ignored. I've tried some simple examples, and everything looks fine. Let's go back to basics. What is the motivating example for this change? Please include details of what is not being ignored, including target language and add it to the patch as a test.

wsfulton:
Also, is the proposal for %ignore to imply %nodirector for both classes as well as methods?

LindleyF:
Build the director_ignore java test case and take a look at the outputs.

director_ignore.i uses %ignore on PublicPureVirtualMethod1, among other things, which is a member of DIgnores.

There is no PublicPureVirtualMethod1 mentioned in director_ignoreJNI.java (either outgoing native method or incoming SwigDirector method), and there is no PublicPureVirtualMethod1 in DIgnores.java. However, there *is* a SwigDirector_DIgnores::PublicPureVirtualMethod1() in director_ignore_wrap.cxx which does nothing except throw an exception.

In other words, the current scheme allows us to %ignore our way into creating a situation where it's impossible to define a non-abstract Java subclass of DIgnores, but not detect that until runtime. I propose that it is better to cause a compile error if one tries to %ignore a pure-virtual method of a class using the director feature.

One might argue that the current approach allows a Java user to implement only part of an interface's functionality and just "trust" that the other part won't be used, but that strikes me as poor design that SWIG shouldn't be encouraging.

This is not the only aspect of the problem, though. I originally noticed this because there were actually compile errors occurring from SwigDirector C++ methods that we didn't expect to be generated at all due to the ignore-all %rename "". (Not pure virtual methods in this case though.) I'll see if I can find a repro of that, but I don't have one yet.

LindleyF:
Ping?

wsfulton:
Hi @Lindley, sorry this one fell off my radar. Coming back to this, it isn't clear what the right behaviour ought to be. If this patch is used, what do you propose a user should do and wants to %ignore a pure virtual method? From what I can see that works and the code compiles without the patch. With the patch, how would one avoid the compile error? Normally SWIG tries very hard to avoid compile errors.

greg-hellings:
I see this is a zombie thread at this point, but I've arrived here through chasing my own compiler warnings.

In the C++ library I'm wrapping I have a deprecated method which I have marked as %ignore, and that works as intended and does not generate issues with the main method itself. However, the class that method is a member of has a director. That director is still generating an implementation for the deprecated method despite the method being ignored.

I can't seem to suppress the director method for that deprecated method, generating deprecation warnings when I compile the generated binding file.

It would be nice for my ignore to also suppress the director method so I can avoid the compiler warnings!

PR #546, opened 2015-10-21
R enum support2015

richardbeare:
Apologies for taking so long over this, but I've finally managed to scrape something together for you to look at.

Thus far I've only implemented the most general option - using C calls to determine the underlying enum value. Enums in the R interface are represented via variables in special "hidden" environments, and the C calls are used only when setting up these variables. There are no runtime calls to the C code after the variables have been initialised, so I don't see very much benefit in pursuing the more specialised approaches that the other languages have, such as generating R code or R constants to provide initialisation values.

The use of C calls to initialise enums requires that the compiled code is available when testing - sourcing no longer works, so I've added runme files for the relevant tests. At present these only confirm that the code compiles and initialisation completes - I don't check correctness yet.

Hopefully this is a start. I hope I have managed to stick to name mangling rules etc.


richardbeare:
I've had to change strategy slightly after testing with SimpleITK. The code using enumerations generates class names, that need to match the enumeration name, using an established function called getRClassName. I've adopted this in the generation of enumeration names, but have attempted to use the recommended Swig_wrapper functions when generating C binding names. The SimpleITK test suite seems to pass, as do the swig r tests.

blowekamp:
It would be great if this patch can be worked into SWIG. This is needed for our SimpleITK project to get R  binding stable.

wsfulton:
Some comments:

Most of the new runtime tests don't seem to do much, or have I missed something?
R::getCurrentScopeName doesn't seem to be used.
I don't follow what you mean by 'The use of C calls to initialise enums requires that the compiled code is available when testing' ?
With regard to the name mangling rules, there is an inconsistency that needs fixing before this can be merged to master. C++ namespace names should NOT by default be mangled into the enum names or anywhere for that matter as the default is to ignore namespace names. The %nspace feature should turn this on. Please use the Swig_name_xxxx family of functions for the implementation (reference implementation is java or c#).
'Leading _ required in enum names not in namespaces' is also inconsistent with the rest of SWIG, why is this?

richardbeare:
Just a quick reply - the runtime tests didn't do anything before - simply sourced the generated R code, checking for syntax. Sourcing would fail now as initialization of the R side structures representing the enums requires calls to the compiled library, whereas previously it used literals/constants. This is what I mean by "requires the compiled code available when testing". I've implemented the simplest and most general option that you mentioned ages ago - something that always works, but requires a call to C library to initialise. Bear in mind that the R side of things was totally broken for all but the simplest enum situations prior to this work.

Regarding the namespaces - I did attempt to follow the java/c# rules, but ran into problems. The issue is to do with R side code using mangled names to define classes that are in turn used to implement a range of dispatching and function overloading options. Obeying the mangling rules breaks a lot of existing code.

My aim with this branch was to get the enums working in a general way. I appreciate that consistent name mangling is a good thing, but I think it needs to be done in a few stages because of the way class names are used on the R side. I'm thinking about the issue now - it may be nicer and neater to generate code for R reference classes and thus avoid some of the issues associated with the way the R binding implements its own special class structures.

Summary - way more R enums work than they used to. Name mangling isn't correct, but it wasn't before, and incorrect mangling (like leading _ ) is extensively used in parts of the R interface that this patch doesn't touch - it just copies them to make stuff work. From the SimpleITK point of view it is a big improvement.

Personally, I'd like to get this in as an incremental improvement so I can work on the other issues, which look like they'd take a major amount of work to even start - e.g. I have no idea where to even start including a %nspace feature. Correct name mangling needs some extensive analysis to deal with how existing code does it "wrong"

richardbeare:
Regarding R::getCurrentCopeName - definitely isn't being used at present. It is possible that I tested it before discovering that I couldn't use the correct name mangling without more extensive changes to the R code generators.

richardbeare:
Further notes on the naming - The R module generates enumeration names like this (when used in SimpleITK) '_itk__simple__PixelIDValueEnum'. These are the names that break the naming conventions, however they were like that before my patches. I have attempted to stick to naming conventions on the C side. The problem with fixing the R side is the extensive use of the incorrectly constructed names as attributes for functions:

attr(`Image___seteq__`, 'returnType') = '_p_itk__simple__Image'
attr(`Image___seteq__`, "inputTypes") = c('_p_itk__simple__Image', '_p_itk__simple__Image')
class(`Image___seteq__`) = c("SWIGFunction", class('Image___seteq__'))

wsfulton:
Just to note here that the tests are not passing. Also the outcome of some offline communicattions with @richardbeare and @joequant are that they will collaborate on fixing the R module naming conventions.

richardbeare:
Superseded by PR #1328.

PR #549, opened 2015-10-28
Add checks for new fields in PyTypeObject and PyNumberMethods for Python version 3.4 and 3.5

ahnolds:
The [tp_finalize field of PyTypeObjects](https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_finalize) is new in Python v3.4
The [nb_matrix_multiply and nb_inplace_matrix_multiply fields of PyNumberMethods](https://docs.python.org/3/c-api/typeobj.html#number-object-structures) are new in Python v3.5

The existing code causes gcc/g++ to emit warnings if -Wmissing-field-initializers is enabled when building the swig-generated files - the patch ensures that the fields are properly defined

### DIFF
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -435,6 +435,9 @@ SwigPyStaticVar_Type(void) {
 #if PY_VERSION_HEX >= 0x02060000
       0,                                        /* tp_version */
 #endif
+#if PY_VERSION_HEX >= 0x03040000
+      0,                                        /* tp_finalize */
+#endif
 #ifdef COUNT_ALLOCS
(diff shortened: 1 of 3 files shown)


ahnolds:
The AppVeyor failure is from a failure to clone the repository, not a real error, but I don't see a way for me to re-run the check without adding another commit.

wsfulton:
I started the build again for you on appveyor.

ahnolds:
Much better, thanks!

ahnolds:
Just pinging - the checks all passed

wsfulton:
Thanks for this. I've also added in a few other similar corrections for -builtin and a few other missing initializers.

Issue #534, opened 2015-09-28
[Scilab] Argument pointer type checking missing


khumba:
Hi folks,

I'm working on extending an existing set of SWIG bindings for the CGAL library to Scilab, and I'm hitting an issue with overloaded functions: https://github.com/khumba/cgal-swig-bindings/issues/4

The generated C++ wrappers use `SwigScilabPtrToObject` on incoming arguments to get pointers, but `SwigScilabPtrToObject` (defined in `scirun.swg`) fails to do any type checking, so whenever there is an overloaded function taking multiple pointer types, the check for the first one generated always succeeds, regardless of the actual object type.  In the bug report above I thought the `Line_2(Point_2, Direction_2)` constructor was mistakenly being called, but it actually calling the (still incorrect) `Line_2(Point_2, Vector_2)` constructor, which happens to *work* because the memory layout is the same :(.

What's needed to fix this?

(edits for clarification)

khumba:
I'm happy to help out with this.  @smarchetto, do you know what's involved?

khumba:
Fixed by @smarchetto between 75510a1 and 88e2ce8; thanks!

PR #552, opened 2015-11-02
[WIP] Make Scilab runtime support keep track of pointer types

khumba:
Hello, here's a preliminary "works for me" fix for #534, although I have a couple of questions because I'm not exactly sure what the code is doing:

- I don't know what to do with the `newmemory` value that `SWIG_TypeCast` returns in `SwigScilabPtrToObject`.

- Does `SwigScilabPtrFromObject` need to look at its `flags` parameter for anything?  It's not currently.

- I still need to write unit tests.

(CC @smarchetto)

### DIFF
--- a/Lib/scilab/scirun.swg
+++ b/Lib/scilab/scirun.swg
@@ -99,6 +99,16 @@ SWIG_Scilab_SetOutput(void *pvApiCtx, SwigSciObject output) {
 }
 
 
+/* Typed object wrappers, modelled after SwigPyObject */
+
+struct SwigSciWrapperObj {
+  void *ptr;
+  swig_type_info *ty;
+
+  SwigSciWrapperObj(void *ptr, swig_type_info *ty) : ptr(ptr), ty(ty) {}
+};
+
+
 /* Pointer conversion functions */
 
 SWIGINTERN int
@@ -106,6 +116,7 @@ SwigScilabPtrToObject(void *pvApiCtx, int iVar, void **pObjValue, swig_type_info
   SciErr sciErr;
   int iType = 0;
   int *piAddrVar = NULL;
+  void *sobj;
 
   sciErr = getVarAddressFromPosition(pvApiCtx, iVar, &piAddrVar);
   if (sciErr.iErr) {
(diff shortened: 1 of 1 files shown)


smarchetto:
Hello, 

Sorry I missed your messages !
Tracking type information in the runtime is one of the things I have been wanting for Swig Scilab for a while, I am enthousiastic about your proposal and your smart fix about pointer type .

About your questions:
- I am not sure what this 'newmemory' is. It seems to be used by the converter function during cast. I think we don't care since we do not use a converter function in our case.
- as explained in the SWIG development documentation (Doc/Devel/runtime.txt from the root of your swig folder) the flags can be used for several purposes in memory management. I think we do not care about this also since pointer lifecycle is always the responsability of the end user.

There is a problem with your fix, however. It is written in C++ while SWIG wrapping code requires C. I rewrote your fix in C, and I am testing it. As you said, it needs some specific tests and larger tests.

khumba:
Hi Simon, it's great to hear from you, and thanks for the interest!

After I posted this issue, I had some concerns about whether this approach was leaking wrapper objects, since the patch creates a new wrapper object every time an object is passed to Scilab, but only deletes a (single) wrapper when Scilab deletes the underlying object.  For instance, I have a global variable holding a CGAL object, `const Origin ORIGIN`, and my Scilab process consumes an additional 30MiB of memory every time I run:

    for i = 1:1000000; p = ORIGIN_get(); end

I [posted](https://sourceforge.net/p/swig/mailman/swig-devel/thread/CANGqftC_tyLyaPFx9i8SKM%2BvzYZ4rMLfxhAdY20X0soxTHSG_g%40mail.gmail.com/#msg34738908) about this on swig-devel with a slightly updated patch (not that that matters now :) and also asked about `newmemory`, for which William provided a good answer (not necessary for me right now; in our use case we're not using `shared_ptr`).

It seems like Scilab needs to manage the lifetime of the wrappers as well, but I wasn't able to find a way to do that.  Maybe with Scilab 6?

smarchetto:
Hi Bryan,

I've choosen and implemented a slightly different appproach to solve memory leak problems. 

Instead of using a C structure pointer `SwigSciWrapperObj`, to encapsulate the pointer type information, we use a Scilab [TList](https://help.scilab.org/docs/5.3.2/en_US/tlist.html). Unlike pointers, `TList` lifecycles are managed by Scilab.

The `TList` contains three fields: 
- object/pointer type name (Scilab string)
- object/pointer type information (Scilab pointer)
- object/pointer value (Scilab pointer)

The big advantage of a TList, is that it exposes a new type in Scilab, which you can manipulate, one example is the built-in `typeof` function, but you can also do overloading....

An example of using, on a ClassA type
```
--> a = new_ClassA()
 a  =
            a(1)
ClassA * 
            a(2)

            a(3)
--> typeinfo(a)
ans  = 
  ClassA*
```

(The way the `TList` is printed could be changed, to hide the pointer fields for example).

This done, I have fixed the Swig type pointer checking for Scilab, now the overloading also works properly for object/pointers. 
I have done several tests, including your initial problem, overloading in constructor. Everything is OK : test suite, no memory leaks.

It just needs now a documentation update, and maybe some configuration through SWIG `%feature`.
But I would like to push now the fix to master, so that you can test. 
Oh, as you are the initiator of all this, I would like to mention your name somewhere, if you don't mind.

khumba:
This is great!  It also works for me for khumba/cgal-swig-bindings#4 and the loop above.  I haven't done more thorough testing than that at the moment; I'm currently getting ready for a couple months of travelling.  And +1 for being able to see the type of the pointer.  Thanks a lot Simon, this will save us a lot of time renaming things.

> Oh, as you are the initiator of all this, I would like to mention your name somewhere, if you don't mind.

If you like :).

> Note: in the case of Scilab, for now the SWIG runtime needs to be initialized manually (with the provided function `<module>_Init`). I have to find a way to automatize this.

I found that out after finding all the type structures were null :).  For now we're calling this manually in a custom loader script (that's also performing some name shortening, which we'd eventually like to drop when we move to Scilab 6, but SWIG's not quite ready for that yet).  I assume there's some problem with calling init from `loader.sce`...?

Anyway, I think I can close this now.  Thanks again.

Lib/scilab/scirun.swg
```diff
@@ -118,26 +129,44 @@ SwigScilabPtrToObject(void *pvApiCtx, int iVar, void **pObjValue, swig_type_info
...
     return SWIG_ERROR;
   }
 
+  if (sobj) {
+      SwigSciWrapperObj *wrapper = static_cast<SwigSciWrapperObj*>(sobj);
+      swig_cast_info *cast = SWIG_TypeCheck(wrapper->ty->name, descriptor);
+      if (!cast) {
+          return SWIG_ERROR;
+      }
+      int newmemory = 0;
```
ojwb: This mixes declarations and code, which isn't allowed by C90 and is causing a test failure: https://travis-ci.org/swig/swig/jobs/114753648

Just moving this line before the `if` is probably enough.

Issue #559, opened 2015-11-14
SwigPyObject_dealloc asserts when iterating over a generator


zjturner:
**I'm using a Python 3.5 Debug interpreter on Windows**

Python 3 has a lot of debug assertions to ensure that exceptions don't get silently swallowed by API calls.  One of these is surfacing in a way that I believe to be a SWIG bug, but I can't be 100% sure.

When I import my SWIG generated module (called `lldb` in this example) and write something like this:

````python
obj = lldb.SBModule()
for symbol in obj.symbol_in_section_iter(lldb.SBSection()):
    s = str(symbol)
````

where `SBModule` and `SBSection` are SWIG wrapped C++ classes, and `symbol_in_section_iter` is implemented as follows on the Python class that SWIG generates:

````python
def symbol_in_section_iter(self, section):
    """Given a module and its contained section, returns an iterator on the
    symbols within the section."""
    for sym in self:
        if in_range(sym, section):
            yield sym
````

When it goes into SwigPyObject_dealloc, it executes these lines:

````python
/* we need to create a temporary object to carry the destroy operation */
PyObject *tmp = SwigPyObject_New(sobj->ptr, ty, 0);
res = SWIG_Python_CallFunctor(destroy, tmp);
````

If you look in the Python source of `SWIG_Python_CallFunctor` in Python 3.5, file `Objects/abstract.c` line `2135`, you will see this:

````c++
PyObject *
PyObject_Call(PyObject *func, PyObject *arg, PyObject *kw)
{
    ternaryfunc call;
    PyObject *result;

    /* PyObject_Call() must not be called with an exception set,
       because it may clear it (directly or indirectly) and so the
       caller looses its exception */
    assert(!PyErr_Occurred());
````

This assertion is firing because an exception is set.  The exception that's occuring is a `StopIteration` exception, presumably because the object is being de-allocated in the same context that the generator is being destroyed.  I don't know, I'm a little confused myself, but this is the best I can figure out.  

Any Python experts around that can help me figure out if there's a workaround?  Is this a real bug in SWIG or something I'm doing wrong?

zjturner:
Update: I can fix this by replacing the code in `SwigPyObject_dealloc` with the following:
````c++
PyObject *ex_val = nullptr, *ex_type = nullptr, *ex_tb = nullptr;  // add this line
PyErr_Fetch(&ex_val, &ex_type, &ex_tb);   // add this line
res = SWIG_Python_CallFunctor(destroy, tmp);
[...]

PR #560, opened 2015-11-14
Python - Save and restore exception state before calling destroy.

zjturner:
PyObject_CallFunction has the potential to silently drop the active
exception.  In cases where the user just finished iterating a
generator, StopIteration will be active.  Most of the time this
is fine because destroy() won't raise an exception.  On Python 3
however, and with a debug interpreter, you will get an assertion
failure inside of Python.  And in the worst case scenario, if
destroy() does throw an exception, the intepreter probably won't
be able to correctly detect the end of the iteration.

This attempts to fix Issue #559.

I also fixed some inconsistent whitespace issues with this patch
(mixed tabs and spaces).

### DIFF
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -537,14 +537,21 @@ SwigPyObject_dealloc(PyObject *v)
       /* destroy is always a VARARGS method */
       PyObject *res;
       if (data->delargs) {
-	/* we need to create a temporary object to carry the destroy operation */
-	PyObject *tmp = SwigPyObject_New(sobj->ptr, ty, 0);
-	res = SWIG_Python_CallFunctor(destroy, tmp);
-	Py_DECREF(tmp);
+        /* we need to create a temporary object to carry the destroy operation */
+        PyObject *tmp = SwigPyObject_New(sobj->ptr, ty, 0);
+        /* PyObject_CallFunction() has the potential to silently drop the active
+           active exception.  In cases where we just finished iterating over a
+           generator StopIteration will be active right now, and this needs to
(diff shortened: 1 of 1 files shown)


zjturner:
Unfortunately I use primarily Windows and I'm not prepared to set up a Cygwin or MSYS environment to build.  But this change seems trivial at first glance and I tested it by pasting the patched version into my generated code, so hopefully this isn't necessary.  If there's any other steps I can take let me know.

This is my first patch against SWIG, so let me know if I've missed or overlooked any necessary steps.

zjturner:
I'm not sure why the AppVeyor build failed, but look at the details do not appear to be related to my patch.  Is there any way to retry it without submitting a new pull request?

ojwb:
> I couldn't find a clear description of the coding style on the SWIG website.

See: http://www.swig.org/Doc3.0/Extending.html#Extending_coding_style_guidelines

Read strictly, that suggests that everything should be tab-filled (so you'll never see 8 spaces together), but in practice that's not universally true in the current codebase.  But at least failing to tab-fill doesn't make things look screwy for other developers, whereas it's pretty catastrophic if we don't all agree on how wide a tab character displays as.

> Is there anyone who can look at the substance of the patch and comment on the correctness?

I don't know the Python C API well enough to really say, but it looks plausible to me.

Not sure if we currently have any Python people who are active in SWIG development.  If it doesn't break the testsuite, and nobody speaks up against it given a reasonable opportunity to, I'd argue for applying it.  Bonus points for adding a regression testcase to the testsuite (i.e. one which fails without the patch and passes with).

wsfulton:
#573 builds on this patch and has been merged into master.

PR #572, opened 2015-12-15
Fixing PyLong->long conversion

ahnolds:
Fixing PyLong->long conversion so that Python3 ints that don't fit in a C long are properly identified and not mistakenly treated as PyInt.
Prior to this patch, because PyInt_Check/PyInt_AsLong are #defined to be PyLong_Check/PyLong_AsLong in Python3, arbitrary sized integers were incorrectly identified and processed as fixed width integers.  This caused assorted issues with large integers being incorrectly converted to -1 (the error return code from PyLong_AsLong) and errors being raised unnecessarily and potentially at confusing times.

Also adding a unittest for this issue.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -5,6 +5,25 @@ See the RELEASENOTES file for a summary of changes in each release.
 Version 3.0.8 (in progress)
 ===========================
 
+2015-12-23: ahnolds
+            [Python] Fixes for conversion of signed and unsigned integer types:
+
+            No longer check for PyInt objects in Python3. Because PyInt_Check
+            and friends are #defined to the corresponding PyLong methods, this
+            had caused errors in Python3 where values greater than what could be
+            stored in a long were incorrectly interpreted as the value -1 with
+            the Python error indicator set to OverflowError. This applies to
+            both the conversions PyLong->long and PyLong->double.
+
+            Conversion from PyLong to long, unsigned long, long long, and
+            unsigned long long now raise OverflowError instead of TypeError in
+            both Python2 and Python3 for PyLong values outside the range
+            expressible by the corresponding C type. This matches the existing
+            behavior for other integral types (signed and unsigned ints, shorts,
+            and chars), as well as the conversion for PyInt to all numeric
+            types. This also indirectly applies to the size_t and ptrdiff_t
+            types, which depend on the conversions for unsigned long and long.
+
 2015-12-19: wsfulton
             [Python] Python 2 Unicode UTF-8 strings can be used as inputs to char * or
             std::string types if the generated C/C++ code has SWIG_PYTHON_2_UNICODE defined.
--- a/Examples/test-suite/primitive_types.i
+++ b/Examples/test-suite/primitive_types.i
@@ -365,6 +365,7 @@ macro(size_t,             pfx, sizet)
 %define ovr_decl(type, pfx, name)
   virtual int pfx##_##val(type x) { return 1; }
   virtual int pfx##_##ref(const type& x) { return 1; }
+  virtual const char* pfx##_##str(type x) { return "name"; }
 %enddef
 
 
--- a/Examples/test-suite/python/primitive_types_runme.py
+++ b/Examples/test-suite/python/primitive_types_runme.py
@@ -1,3 +1,5 @@
+import ctypes
+import sys
 from primitive_types import *
 
 var_init()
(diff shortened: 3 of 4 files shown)


wsfulton:
Thanks for submitting the change and providing tests is much appreciated. However, the test is adding an example, rather than using the test-suite. The examples are meant to be interesting for users, where I feel the test for this change is more suited to the test-suite. Could you replace the example with a test-suite test? This patch is a quite a fundamental change so I would only feel comfortable accepting it with some better tests verifying inputs and outputs as well as range checking. I suggest you just add in some runtime tests to Examples/test-suite/primitive_types_runme.py. Please also include the values sys.maxint and sys.maxint + 1 as the test inputs and outputs to functions/variables using long, unsigned long, int, unsigned int, short and unsigned short. The Examples/tests-suite/primitive_types.i testcase is hard to follow in all honesty as it has a pile of macros generating test functions and variables. If you run the following, which passes '-E' to swig when running the testcase, you will see the output from the preprocessor and the expanded macros so that you can see all the functions names that can be called, such as ```long wrp_vval_long(long x)```:

```
~/swig/Examples/test-suite/python$ env SWIG_FEATURES=-E make primitive_types.cpptest 
```
Overloading is an important area to test too, so call the ```ovr_val``` functions too.

ahnolds:
I realized after I pushed that I had originally made an example and not a test - whoops!  Now that I've had time, try a version with real tests.

My ability to actually test ovr_val/ovr_ref is limited, since I don't know which version it's calling since the return value is always 1.  That said, I added calls in the unit tests with a wide range of inputs, including both ULONG_MAX (2 * sys.maxint + 1) and LONG_MIN (- sys.maxint - 1), which guarantees that it can at least call both the unsigned long and signed long versions.

Let me know if there are any other tests you think are needed.

wsfulton:
The travis tests failed for no good reason but are okay now after I restarted the one failure. However, the Appveyor tests show some problems on Windows.

Feel free to modify or add some new tests to the .i file for checking the overloading.

ahnolds:
The issues were mostly related to `long long` values - as a Linux developer, I'm not used to platform where `long long != long`.  Hopefully I've now resolved everything to work with Windows weird LLP64 model...

I also added an `ovr_str` function in the .i file to return the type name of the chosen overload to make testing that cleaner.

ahnolds:
Just a note: the failure-allowed Travis build with `SWIG_FEATURES=-O` fails this test I added.  The
`-fastdispatch` flag causes it to raise `TypeError` rather than `NotImplementedError` if it fails all the checks since it falls into the last case by default.  The only way I can think of to code around that is to add something like this to the end of checkOverload:

    except TypeError:
        import os
        if '-O' not in os.environ['SWIG_FEATURES'] or val != limit:
            raise RuntimeError, "bad " + name + " typemap"

But that seems sufficiently kludgy that I'm not going to add it unless others think I should

wsfulton:
I would just accept either exception being thrown. Add the additional exception handler but skip the check for -O and just add a comment saying -O or -fastdispatch will cause a TypeError instead of a NotImplementedError.

wsfulton:
This is a very important core area of the Python primitive types that has changed and so it needs some good user documentation as to exactly what has got better. Can you put this into the CHANGES.current file and mention when the OverflowError is thrown in python 2&3 compared to before and exactly which C types have changed for Python3 - looks like just long and unsigned long to me.

Could you also please fix the whitespace/tabs of the lines you've added to pyrimtypes.swg to match the surrounding lines. Otherwise I'd like to merge this patch in asap for the swig-3.0.8 release.

ahnolds:
Whitespace fixed, changelog updated, and everything rebased into 3 nice clean commits.

wsfulton:
@ahnolds, could you please rework the tests associated with this pull requests that I had to disable in 26d0b7f5b93bd26c0768d350061f25ee914b4fa9 ?

ahnolds:
I'm honestly not sure how easy to do that is. I could probably do all the same calculations from the wrapped C code or just return INT_MAX/LONG_MAX/PTRDIFF_MAX/etc, but the whole point of writing the unit tests the way I did was that it gets a check on the max/min values completely independent of Swig.

My counter-argument would be that if we're going to drop support for Python pre-2.6 starting in 3.1.0 anyway, it's not really worth the work to revamp these unit tests to add support for 9+ year old versions of Python in the one remaining Swig release that supports them at all.  After all, sys.maxsize has been available since 2.6b1 (June 18, 2008), and ctypes since 2.5a1 (April 5, 2006).

wsfulton:
I understand, that's okay. BTW, I havn't suggested we drop support for 2.4 and 2.5 altogether and I don't think it is too onerous to keep this support. I don't believe there are any large enough specific features introduced in 2.4 and 2.5 to reduce the complexity of the implementation. When we get around to the python tidy up for swig-3.1, exactly which swig options to drop can be discussed in finer detail. I just want to state here that I don't think it is okay to drop 2.4 support without a thorough analysis of what exactly this means.

ojwb:
> as a Linux developer, I'm not used to platform where `long long != long`

They're actually only the same size for 64-bit Linux architectures, as https://wiki.debian.org/ArchitectureSpecificsMemo notes:

> All Debian architectures have:
[...]
> sizeof(long) == sizeof(void *)
> sizeof(long long) == 8

PR #573, opened 2015-12-15
Python3.5 destructor exception fix

coleb:
Further improvements to the #560 pull request. 

Uses PyErr_WriteUnraisable to write error messages to sys.stderr when ```__del__``` raises a Python exception that can't be handled. This is an unlikely situation, but the standard python approach. 

Also adds a test case where this destructor problem is seen with unnamed temporary variables. 

Still need test cases for StopIteration type of problems to be sure this fixes those.

### DIFF
--- a/Examples/test-suite/python/Makefile.in
+++ b/Examples/test-suite/python/Makefile.in
@@ -58,6 +58,7 @@ CPP_TEST_CASES += \
 	primitive_types \
 	python_abstractbase \
 	python_append \
+	python_destructor_exception \
 	python_director \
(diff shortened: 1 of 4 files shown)


wsfulton:
Using PyErr_WriteUnraisable, as this patch does, seems to be the right way to handle potential, albeit unlikely, exceptions thrown during destruction.  I've merged this in and improved the testcase added here to replicate the particular StopIteration problem that was reported in #559 and #560. I also had to add a similar fix for -builtin.

Issue #568, opened 2015-12-03
stdint.i typedefs


andy-slac:
Hi, I'm having an issue with the code generated for one of the C++ methods that accepts a vector of uint64_t integers. I think this is due to inconsistency in typedefs for uint64_t in SWIG and gcc (this is happening with gcc 4.8 on Ubuntu 14.04 with swig3.0 installed from repo). Here is is the simplest test case that reproduces my problem:

    # test.i
    %module test
    %{
    #include <vector>
    #include <stdint.h>
    void testMethod(std::vector<uint64_t> const& v) {}
    %}
    
    %include <std_vector.i>
    %include <stdint.i>
    
    %template(VectorUint64) std::vector<uint64_t>;
    
    void testMethod(std::vector<uint64_t> const& v);

And here is the error that I see:

    % swig3.0 -v
    Must specify an input file. Use -help for available options.
    root@ubuntu1404:~/tmp# swig3.0 -version

    SWIG Version 3.0.2

    Compiled with g++ [x86_64-unknown-linux-gnu]

    Configured options: +pcre

    Please see http://www.swig.org for reporting bugs and further information

    % swig3.0 -python -c++ test.i
    %  g++ -I/usr/include/python2.7 -c -o test_wrap.o test_wrap.cxx
    test_wrap.cxx: In function ‘std::vector<long long unsigned int, std::allocator<long long unsigned int> >* std_vector_Sl_uint64_t_Sg____getslice__(std::vector<long unsigned int>*, std::vector<long long unsigned int, std::allocator<long long unsigned int> >::difference_type, std::vector<long long unsigned int, std::allocator<long long unsigned int> >::difference_type)’:
    test_wrap.cxx:4671:42: error: cannot convert ‘std::vector<long unsigned int>*’ to ‘std::vector<long long unsigned int, std::allocator<long long unsigned int> >*’ in return
           return swig::getslice(self, i, j, 1);
                                              ^
    test_wrap.cxx: In function ‘std::vector<long long unsigned int, std::allocator<long long unsigned int> >* std_vector_Sl_uint64_t_Sg____getitem____SWIG_0(std::vector<long unsigned int>*, PySliceObject*)’:
    test_wrap.cxx:4691:47: error: cannot convert ‘std::vector<long unsigned int>*’ to ‘std::vector<long long unsigned int, std::allocator<long long unsigned int> >*’ in return
           return swig::getslice(self, id, jd, step);
                                                   ^
    ..... a lot more errors

[...]

PR #575, opened 2015-12-22
Win64 ptrdiff t

ahnolds:
Adds support for ptrdiff_t on 64-bit Windows platforms where `sizeof(ptrdiff_t) == 64` and `sizeof(long) == 32`.  Previously this caused problems because in Python slice notation given the index [NUM:], the second index is set to PTRDIFF_MAX, which is greater than LONG_MAX on LLP64 platforms like 64-bit Windows.

This (should) now pass the previously failure-allowed Appveyor build for Python 2 on x64 Windows.

### DIFF
--- a/Examples/test-suite/primitive_types.i
+++ b/Examples/test-suite/primitive_types.i
@@ -368,6 +368,17 @@ macro(size_t,             pfx, sizet)
   virtual const char* pfx##_##str(type x) { return "name"; }
 %enddef
 
+/* checking size_t and ptrdiff_t typemaps */
+%include "stdint.i"
+%inline {
+  size_t    get_size_min()    { return 0; }
+  size_t    get_size_max()    { return SIZE_MAX; }
+  ptrdiff_t get_ptrdiff_min() { return PTRDIFF_MIN; }
+  ptrdiff_t get_ptrdiff_max() { return PTRDIFF_MAX; }
+
+  size_t    size_echo   (size_t val)    { return val; }
+  ptrdiff_t ptrdiff_echo(ptrdiff_t val) { return val; }
+}
 
 %inline {
   struct Foo
--- a/Examples/test-suite/primitive_types.i
+++ b/Examples/test-suite/primitive_types.i
@@ -369,6 +369,10 @@ macro(size_t,             pfx, sizet)
 %enddef
 
 /* checking size_t and ptrdiff_t typemaps */
+%begin %{
+// Must be defined before Python.h is included, since this may indirectly include stdint.h
+#define __STDC_LIMIT_MACROS
+%}
 %include "stdint.i"
 %inline {
   size_t    get_size_min()    { return 0; }
--- a/Examples/test-suite/python/primitive_types_runme.py
+++ b/Examples/test-suite/python/primitive_types_runme.py
@@ -573,3 +573,34 @@ if val_double(sys.maxint + 1) != float(sys.maxint + 1):
     raise RuntimeError, "bad double typemap"
 if val_double(-sys.maxint - 2) != float(-sys.maxint - 2):
     raise RuntimeError, "bad double typemap"
+
+
+# Check the minimum and maximum values that fit in ptrdiff_t and size_t
+def checkType(name, maxfunc, maxval, minfunc, minval, echofunc):
+    if maxfunc() != maxval:
+        raise RuntimeError, "bad " + name + " typemap"
+    if minfunc() != minval:
+        raise RuntimeError, "bad " + name + " typemap"
+    if echofunc(maxval) != maxval:
+        raise RuntimeError, "bad " + name + " typemap"
+    if echofunc(minval) != minval:
+        raise RuntimeError, "bad " + name + " typemap"
+    error = 0
+    try:
+        echofunc(maxval + 1)
+        error = 1
+    except OverflowError:
+        pass
+    if error == 1:
+        raise RuntimeError, "bad " + name + " typemap"
+    try:
+        echofunc(minval - 1)
+        error = 1
+    except OverflowError:
+        pass
+    if error == 1:
+        raise RuntimeError, "bad " + name + " typemap"
+
+# sys.maxsize is the largest value supported by Py_ssize_t, which should be the same as ptrdiff_t
+checkType("ptrdiff_t", get_ptrdiff_max, sys.maxsize,           get_ptrdiff_min, -(sys.maxsize + 1), ptrdiff_echo)
+checkType("size_t",    get_size_max,    (2 * sys.maxsize) + 1, get_size_min,    0,                  size_echo)
--- a/Lib/javascript/jsc/javascriptprimtypes.swg
+++ b/Lib/javascript/jsc/javascriptprimtypes.swg
@@ -98,18 +98,21 @@ SWIG_AsVal_dec(unsigned long)(JSValueRef obj, unsigned long *val)
 
 %fragment(SWIG_From_frag(long long),"header",
     fragment=SWIG_From_frag(long),
-    fragment="<limits.h>") {
+    fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERNINLINE JSValueRef
 SWIG_From_dec(long long)(long long value)
 {
   return JSValueMakeNumber(context, value);
 }
+%#endif
 }
 
 %fragment(SWIG_AsVal_frag(long long),"header",
     fragment=SWIG_AsVal_frag(long),
     fragment="SWIG_CanCastAsInteger",
-    fragment="<limits.h>") {
+    fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERN int
 SWIG_AsVal_dec(long long)(JSValueRef obj, long long* val)
 {
@@ -120,6 +123,7 @@ SWIG_AsVal_dec(long long)(JSValueRef obj, long long* val)
 
   return SWIG_OK;
 }
+%#endif
 }
 
 /* unsigned long long */
@@ -127,19 +131,22 @@ SWIG_AsVal_dec(long long)(JSValueRef obj, long long* val)
 
 %fragment(SWIG_From_frag(unsigned long long),"header",
 	  fragment=SWIG_From_frag(long long),
-	  fragment="<limits.h>") {
+	  fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERN JSValueRef
 SWIG_From_dec(unsigned long long)(unsigned long long value)
 {
   return (value > LONG_MAX) ?
     JSValueMakeNumber(context, value) : JSValueMakeNumber(context, %numeric_cast(value,long));
 }
+%#endif
 }
 
 %fragment(SWIG_AsVal_frag(unsigned long long),"header",
     fragment=SWIG_AsVal_frag(unsigned long),
     fragment="SWIG_CanCastAsInteger",
-    fragment="<limits.h>") {
+   fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERN int
 SWIG_AsVal_dec(unsigned long long)(JSValueRef obj, unsigned long long *val)
 {
@@ -158,6 +165,7 @@ SWIG_AsVal_dec(unsigned long long)(JSValueRef obj, unsigned long long *val)
 
   return SWIG_OK;
 }
+%#endif
 }
 
 /* double */
--- a/Lib/javascript/v8/javascriptprimtypes.swg
+++ b/Lib/javascript/v8/javascriptprimtypes.swg
@@ -112,18 +112,21 @@ int SWIG_AsVal_dec(unsigned long)(v8::Handle<v8::Value> obj, unsigned long *val)
 
 %fragment(SWIG_From_frag(long long),"header",
     fragment=SWIG_From_frag(long),
-    fragment="<limits.h>") {
+    fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERNINLINE
 v8::Handle<v8::Value> SWIG_From_dec(long long)(long long value)
 {
   return SWIGV8_NUMBER_NEW(value);
 }
+%#endif
 }
 
 %fragment(SWIG_AsVal_frag(long long),"header",
     fragment=SWIG_AsVal_frag(long),
     fragment="SWIG_CanCastAsInteger",
-    fragment="<limits.h>") {
+    fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERN
 int SWIG_AsVal_dec(long long)(v8::Handle<v8::Value> obj, long long* val)
 {
@@ -134,6 +137,7 @@ int SWIG_AsVal_dec(long long)(v8::Handle<v8::Value> obj, long long* val)
 
   return SWIG_OK;
 }
+%#endif
 }
 
 /* unsigned long long */
@@ -141,19 +145,22 @@ int SWIG_AsVal_dec(long long)(v8::Handle<v8::Value> obj, long long* val)
 
 %fragment(SWIG_From_frag(unsigned long long),"header",
     fragment=SWIG_From_frag(long long),
-    fragment="<limits.h>") {
+    fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERNINLINE
 v8::Handle<v8::Value> SWIG_From_dec(unsigned long long)(unsigned long long value)
 {
   return (value > LONG_MAX) ?
     SWIGV8_INTEGER_NEW_UNS(value) : SWIGV8_INTEGER_NEW(%numeric_cast(value,long));
 }
+%#endif
 }
 
 %fragment(SWIG_AsVal_frag(unsigned long long),"header",
     fragment=SWIG_AsVal_frag(unsigned long),
     fragment="SWIG_CanCastAsInteger",
-    fragment="<limits.h>") {
+    fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERN
 int SWIG_AsVal_dec(unsigned long long)(v8::Handle<v8::Value> obj, unsigned long long *val)
 {
@@ -171,6 +178,7 @@ int SWIG_AsVal_dec(unsigned long long)(v8::Handle<v8::Value> obj, unsigned long 
 
   return SWIG_OK;
 }
+%#endif
 }
 
 /* double */
--- a/Lib/octave/octprimtypes.swg
+++ b/Lib/octave/octprimtypes.swg
@@ -97,15 +97,20 @@ SWIG_AsVal_dec(bool)(const octave_value& ov, bool *val)
 
 // long long
 
-%fragment(SWIG_From_frag(long long),"header") {
+%fragment(SWIG_From_frag(long long),"header",
+	  fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
   SWIGINTERNINLINE octave_value SWIG_From_dec(long long)  (long long value)
     {    
       return octave_int64(value);
     }
+%#endif
 }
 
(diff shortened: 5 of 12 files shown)


wsfulton:
While I like this patch as it fixes the win64 bit problem which I have not yet come up with a good approach of doing, it is unfortunately not portable. 'long long' is not in the C++98 standard and we need to generate code that works with this standard. The patch introduces a dependency for many STL types on 'long long'. I can't think of any tweaks right now to solve this, can you? I'm wondering if something in the python header files can help.

ahnolds:
Well the issue actually applies beyond Python, it's just that the only place large values of ptrdiff_t appear (at least in what's tested on 64-bit Windows) are in Python slices without a specified end.  The lack of endpoint causes python to use sys.maxsize (aka max ssize_t) as the endpoint, which is greater than the maximum value of a long.

The best solution that comes to mind is preprocessor based, something like:

    #if defined(PTRDIFF_IS_LONG)
    # define SWIG_PTRDIFF_TYPE long
    #elif defined(PTRDIFF_IS_LONG_LONG)
    # define SWIG_PTRDIFF_TYPE long long
    #else
    # error "Unknown ptrdiff_t type"
    #endif

And then use `SWIG_PTRDIFF_TYPE` everywhere that `long` appears now in the ptrdiff_t fragments. Incidentally, the same should probably be done for size_t, since it also is 64-bits vs a 32-bit long on 64-bit Windows.

As for actually generating PTRDIFF_IS_LONG/PTRDIFF_IS_LONG_LONG, I would suggest that you could probably determine the size of ptrdiff_t/size_t during configure, add the relevant entry to swigconfig.h, and then have the actual swig executable write it at the top of the generated wrapper files.  This _should_ work, but (a) I'm not an expert with autotools stuff so it would be not-trivial for me to do this and (b) I have no idea if this would fit with the approach you want for issues like this.

As a side note, I question the need to have strict C++98 compliance with a 2015 (realistically 2016) release of a mature software tool.  I doubt many users are upgrading to the very latest swig but still using a 15+ year old compiler, although I certainly could be wrong:)

ahnolds:
Just a guess on a configure.ac that might be right/at least work:

    AC_CHECK_TYPES([long long])
    AC_CHECK_SIZEOF([ptrdiff_t])
    AC_CHECK_SIZEOF([long])
    AC_CHECK_SIZEOF([long long])

Then your definition for SWIG_PTRDIFF_TYPE could be as follows:

    #if SIZEOF_PTRDIFF_T == SIZEOF_LONG
    # define SWIG_PTRDIFF_TYPE long
    #elif defined(HAVE_LONG_LONG) && (SIZEOF_PTRDIFF_T == SIZEOF_LONG_LONG)
    # define SWIG_PTRDIFF_TYPE long long
    #else
    # error "Unknown ptrdiff_t type"
    #endif

ojwb:
> As a side note, I question the need to have strict C++98 compliance with a 2015 (realistically 2016) release of a mature software tool. I doubt many users are upgrading to the very latest swig but still using a 15+ year old compiler, although I certainly could be wrong:)

The next full C++ standard iteration after C++98 was C++11, so even going by standard release date it's actually less than 5 years and not 15+.  But it's taken a few years for the compiler vendors to fully implement C++11 - e.g. GCC 5 claims to fully implement it, and that was only released 8 months ago.

There's also the difference between a C++11-capable compiler and one which accepts C++11 by default - GCC 5.3 (released this month) still defaults to C++98 mode and you need to specify `-std=c++11` or similar to enable C++11.

OTOH, `long long` was widely implemented as an extension by all the major C++ compilers long ago (for example, I know Xapian has assumed `long long` exists for a long time without any reported portability issues).

> As for actually generating SWIG_PTRDIFF_TYPE, I would suggest that you could probably determine the size of ptrdiff_t/size_t when running configure, add the relevant entries to swigconfig.h, and then have the actual swig executable write the computed #define for SWIG_PTRDIFF_TYPE at the top of the generated wrapper files. This should work, but (a) I'm not an expert with autotools stuff so it would be not-trivial for me to do this and (b) I have no idea if this would fit with the approach you want for issues like this.

SWIG aims to support generating the wrappers on one machine and building them elsewhere, so configure-time tests for things like type sizes aren't a suitable approach.

ahnolds:
> SWIG aims to support generating the wrappers on one machine and building them elsewhere, so configure-time tests for things like type sizes aren't a suitable approach.

That may make this a lost cause, but here are some possibilities, ranked from best to worst (I think):

1) `PTRDIFF_MAX` is defined in stdint.h as of C99.  If that's long enough ago to get your approval, you could write the entire following macro comparison into the wrapper:

    #if PTRDIFF_MAX == LONG_MAX
    # define SWIG_PTRDIFF_TYPE long
    #elif PTRDIFF_MAX == LLONG_MAX
    # define SWIG_PTRDIFF_TYPE long
    #else
    # error "Unknown ptrdiff_t type"
    #endif

You could add other comparisons (INT_MAX/SHORT_MAX/SCHAR_MAX) if there exist platforms like that, which seems unlikely.  This has the advantage of loudly warning you if you try and compile on a platform with a weird ptrdiff_t type.

2) GCC defines `__PTRDIFF_TYPE__`, and you could use _WIN64 to flag the one known `ptrdiff_t == long long` case and assume `ptrdiff_t = long` everywhere else, but that's not realistic and it could fail silently during compilation and only freak out at runtime (like the current version).

3) You could say if `LLONG_MAX > LONG_MAX`, use `ptrdiff_t == long long`.  This seems questionable since it basically assumes `sizeof(ptrdiff_t) == sizeof(long long)` (if long long is bigger, then we should just use long; if long long is smaller, then we're still hosed if we use it).  Again, not necessarily a realistic assumption going forward, and the error isn't apparent until runtime.

wsfulton:
Related to #568.

I hate to be a downer, but 

1) PTRDIFF_MAX requires c99 and stdint.h which isn't widely available (Visual Studio 2010) was the first time VC++ supplied stdint.h.

2) I think you are saying we need 'long long' which isn't guaranteed with _WIN64 (but __w64 is, but is now deprecated and is removed in more recent VC++).

3) Assumes LLONG_MAX is present which requires C99. Possibly we can detect the existence of this macro to generate 'long long' code as the way forward. If the macro is not available, assume there is no 'long long' and then the bug remains as 'long long' and SWIG_AsVal(long long) cannot be used. If someone wants the fix, a compiler upgrade is required.

ahnolds:
My turn to be a downer: On Win32, sizeof(long) == sizeof(ptrdiff_t) == 4, but sizeof(long long) == 8.  So just detecting that long long is available isn't sufficient.  This basically rules out 3) as far as I can tell.

I think that leaves only one option for solving the underlying problem, namely using a long chain of macros to guess at the right type for ptrdiff_t e.g.

    #ifndef SWIG_PTRDIFF_TYPE
    # if defined __PTRDIFF_TYPE__
    #  define SWIG_PTRDIFF_TYPE __PTRDIFF_TYPE__
    # elif defined PTRDIFF_MAX
    #  if PTRDIFF_MAX == LONG_MAX
    #   define SWIG_PTRDIFF_TYPE long
    #  elif PTRDIFF_MAX == LLONG_MAX
    #   define SWIG_PTRDIFF_TYPE long long
    #  endif
    # elif defined _WIN64
    #  define SWIG_PTRDIFF_TYPE long long
    # endif
    #endif
    #ifndef SWIG_PTRDIFF_TYPE
    # warning "Unable to reliably determine SWIG_PTRDIFF_TYPE, assuming long"
    # define SWIG_PTRDIFF_TYPE long
    #endif

Alternatively, we could give up on a general solution to ptrdiff_t problems and handle the specific failure from Python.  I think the easiest thing here would be to use PyLong_AsSsize_t and PyLong_FromSsize_t to handle ptrdiff_t.  This _should_ safely resolve the problem since a missing slice endpoint is set to sys.maxsize, which is defined as the maximum value of a Py_ssize_t type.  I can't imagine a platform where `sizeof(size_t) == sizeof(ssize_t) != sizeof(ptrdiff_t)`, so most likely this would be safe (although I don't think the standard guarantees it).

wsfulton:
I don't follow your ruling out of 3). The patch below is roughly how I'd implement it for Python (needs expanding to other languages). The patch is on top of this pull request and basically uses your original idea and only uses ```long long``` if it is available by detecting the existence of LLONG_MAX.

```patch
diff --git a/Lib/python/pyprimtypes.swg b/Lib/python/pyprimtypes.swg
index 30bb64f..a53927c 100644
--- a/Lib/python/pyprimtypes.swg
+++ b/Lib/python/pyprimtypes.swg
@@ -186,21 +186,30 @@ SWIG_AsVal_dec(unsigned long)(PyObject *obj, unsigned long *val)
 
 /* long long */
 
+%fragment("SWIG_LongLongAvailable","header", fragment="<limits.h>") {
+%#if defined(LLONG_MAX) && !defined(SWIG_LONG_LONG_AVAILABLE)
+%#define SWIG_LONG_LONG_AVAILABLE
+%#endif
+}
+
 %fragment(SWIG_From_frag(long long),"header",
 	  fragment=SWIG_From_frag(long),
-	  fragment="<limits.h>") {
+	  fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERNINLINE PyObject* 
 SWIG_From_dec(long long)(long long value)
 {
   return ((value < LONG_MIN) || (value > LONG_MAX)) ?
     PyLong_FromLongLong(value) : PyLong_FromLong(%numeric_cast(value,long)); 
 }
+%#endif
 }
 
 %fragment(SWIG_AsVal_frag(long long),"header",
 	  fragment=SWIG_AsVal_frag(long),
 	  fragment="SWIG_CanCastAsInteger",
-	  fragment="<limits.h>") {
+	  fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERN int
 SWIG_AsVal_dec(long long)(PyObject *obj, long long *val)
 {
@@ -236,25 +245,29 @@ SWIG_AsVal_dec(long long)(PyObject *obj, long long *val)
 %#endif
   return res;
 }
+%#endif
 }
 
 /* unsigned long long */
 
 %fragment(SWIG_From_frag(unsigned long long),"header",
 	  fragment=SWIG_From_frag(long long),
-	  fragment="<limits.h>") {
+	  fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERNINLINE PyObject* 
 SWIG_From_dec(unsigned long long)(unsigned long long value)
 {
   return (value > LONG_MAX) ?
     PyLong_FromUnsignedLongLong(value) : PyLong_FromLong(%numeric_cast(value,long)); 
 }
+%#endif
 }
 
 %fragment(SWIG_AsVal_frag(unsigned long long),"header",
 	  fragment=SWIG_AsVal_frag(unsigned long),
 	  fragment="SWIG_CanCastAsInteger",
-	  fragment="<limits.h>") {
+	  fragment="SWIG_LongLongAvailable") {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
 SWIGINTERN int
 SWIG_AsVal_dec(unsigned long long)(PyObject *obj, unsigned long long *val)
 {
@@ -289,6 +302,7 @@ SWIG_AsVal_dec(unsigned long long)(PyObject *obj, unsigned long long *val)
 %#endif
   return res;
 }
+%#endif
 }
 
 /* double */
diff --git a/Lib/typemaps/primtypes.swg b/Lib/typemaps/primtypes.swg
index c17d504..cfb73b3 100644
--- a/Lib/typemaps/primtypes.swg
+++ b/Lib/typemaps/primtypes.swg
@@ -190,19 +190,23 @@ SWIG_From_dec(ptrdiff_t)(ptrdiff_t value)
 SWIGINTERNINLINE int
 SWIG_AsVal_dec(ptrdiff_t)(SWIG_Object obj, ptrdiff_t *val)
 {
+%#ifdef SWIG_LONG_LONG_AVAILABLE
   int res = SWIG_TypeError;
-  if (sizeof(ptrdiff_t) <= sizeof(long))
-  {
+  if (sizeof(ptrdiff_t) <= sizeof(long)) {
     long v;
     res = SWIG_AsVal(long)(obj, val ? &v : 0);
     if (SWIG_IsOK(res) && val) *val = %numeric_cast(v, ptrdiff_t);
-  }
-  else if (sizeof(ptrdiff_t) <= sizeof(long long))
-  {
+  } else if (sizeof(ptrdiff_t) <= sizeof(long long)) {
     long long v;
     res = SWIG_AsVal(long long)(obj, val ? &v : 0);
     if (SWIG_IsOK(res) && val) *val = %numeric_cast(v, ptrdiff_t);
   }
+%#else
+  /* long long can't be used - will fail for large values if sizeof(ptr_diff_t) > sizeof(long) */
+  long v;
+  int res = SWIG_AsVal(long)(obj, val ? &v : 0);
+  if (SWIG_IsOK(res) && val) *val = %numeric_cast(v, ptrdiff_t);
+%#endif
   return res;
 }
 }
```

ahnolds:
Oh yes, you are completely right, I was just being obtuse and thinking about my original 3) which actually checked the value of LLONG_MAX, not yours which just checks that it is defined.

I'll try to knock out the changes for all the languages over the weekend.  I'll also add the corresponding checks for size_t and the from fragments for ptrdiff_t/size_t which would have the same problem.  Maybe I'll even be ambitious and add some explicit test cases for this somewhere.

ahnolds:
Any reason not to put the definition of SWIG_LONG_LONG_AVAILABLE into the existing <limits.h> fragment?  It seems cleaner to me to put it there, but I don't know if you try to keep those fragments to strictly things that appear/should appear in the associated header file.  It seems like that #define should either go there or in the long long section of primtypes.swg to avoid putting the same thing in each language's primtypes file.

wsfulton:
Looks good. Could you also add a simple run time test - Python at the  minimum, plus any other languages you fancy? How about adding a C function to the primitive_types.i test that returns SIZE_MAX and use that value for testing size_t and your changes? If it fails on a system that does not have stdint.h, that's okay, likewise if long long is not available. Similarly a test for PTRDIFF_MAX/PTRDIFF_MIN.

Also I suggest a cosmetic change:
```diff
+  } else /* hopefully sizeof(ptrdiff_t) <= sizeof(long long) */ {
```
to:
```diff
+  } else {
+    /* assume sizeof(ptrdiff_t) <= sizeof(long long) */
```

ahnolds:
I'm not entirely sure why this is the case, but it looks like SIZE_MAX/PTRDIFF_MAX/PTRDIFF_MIN are not by default defined when including stdint.h on:
* Travis Linux with g++ 4.6.3
* my Windows box using MinGW g++ 4.8.1

But they are on:
* Appveyor
* Travis OSX (clang 6.0)
* Travis Linux with g++ 5.2.1
* my Linux box (g++ 4.8.4 if I recall correctly)

Hence all the failures for the unit test I just added.

I'm not coming up with a great fix offhand - on my Windows box adding

    %begin %{
    #define __STDC_LIMIT_MACROS
    %}

to primitive_types.i fixes the problem since then the macros get defined when stdint.h is included farther down.  Obviously that's not really portable (and I can't tell without trying if it would even work on Travis), but that may be ok for a unit test.  Thoughts on this suggestion or a better workaround?

wsfulton:
We compile with -std=c++98 which is based on c89 and stdint.h was only introduced in c99. This file is available more recently for c++98 though, however, the c99 standard also states:
```
  C++ implementations should define these macros only when _ _STDC_LIMIT_MACROS is defined
before <stdint.h> is included.
```
So I suggest you modify the patch to define ```__STDC_LIMIT_MACROS``` as this is in the standard and so should be portable.  I'd add it just above the stdint.h include, rather than use %begin.

I ran simple c++ tests quickly on Ubuntu Precise (version used on Travis) and SIZE_MAX is  available for g++ 4.6, 4.8, 4.9 and 5.1 even with -std=c++98. The system's version of stdint.h on Ubuntu Trusty has dropped the requirement for ```__STDC_LIMIT_MACROS```, I guess it has something to do with the c++11 standard saying this macros does not need to be defined for cstdint!

ahnolds:
At least on my Windows box, stdint.h is indirectly included by Python.h (Python.h->unicodeobject.h->wchar.h->stdint.h), so I went with %begin to make sure `__STDC_LIMIT_MACROS` was defined before stdint.h got included that way.

wsfulton:
I see, yes, that is then the right thing to do. Thanks for working on this patch, getting Appveyor truly green is a nice step forward.

Yensan:
Hey, all of you in this issue, I need you help. Discussion is hot, BUT for me... I just can't compile to AMD64 !!!
swigwin-3.0.12, 
Python 3.6.3 (v3.6.3:2c5fed8, Oct  3 2017, 18:11:49) [MSC v.1900 64 bit (AMD64)] on win32,
Visual C++ Build Tools

> (.env3) D:\py_c\swig>C:\Python3664\python.exe setup.py build_ext --inplace
> running build_ext
> building 'sample' extension
> C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MT -IC:\Python3664\include -IC:\Python3664\include "-IC:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MS
> VC\14.10.25017\ATLMFC\include" "-IC:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\include" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Win
> dows Kits\10\include\10.0.15063.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\um" "-IC:\Program Files (x86)\Windows Kits\10\includ
> e\10.0.15063.0\winrt" /Tcsample_wrap.c /Fobuild\temp.win-amd64-3.6\Release\sample_wrap.obj
> sample_wrap.c
> C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MT -IC:\Python3664\include -IC:\Python3664\include "-IC:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MS
> VC\14.10.25017\ATLMFC\include" "-IC:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\include" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Win
> dows Kits\10\include\10.0.15063.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.15063.0\um" "-IC:\Program Files (x86)\Windows Kits\10\includ
> e\10.0.15063.0\winrt" /Tcsample.c /Fobuild\temp.win-amd64-3.6\Release\sample.obj
> sample.c
> sample.c: warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss
> C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\link.exe /nologo /INCREMENTAL:NO /LTCG /nodefaultlib:libucrt.lib ucrt.lib /DLL /MANIFEST:EMBED,ID=2 /MANIFESTUAC:NO /LIBPATH:C:\Python3664\libs /LIBPATH:C:\Python3664\
> PCbuild\amd64 "/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\ATLMFC\lib\x64" "/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.10.25017\
> lib\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\lib\um\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.15063.0\ucrt\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.15063.0\um
> \x64" /EXPORT:PyInit_sample build\temp.win-amd64-3.6\Release\sample_wrap.obj build\temp.win-amd64-3.6\Release\sample.obj /OUT:D:\py_c\swig\sample.cp36-win_amd64.pyd /IMPLIB:build\temp.win-amd64-3.6\Releas
> e\sample.cp36-win_amd64.lib /MACHINE:X64
> build\temp.win-amd64-3.6\Release\sample_wrap.obj : fatal error LNK1112: module machine type 'X86' conflicts with target machine type 'x64'
> error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\bin\\link.exe' failed with exit status 1112

PR #580, opened 2016-01-08
Fix test failure on PPC{64} where the char is unsigned by default

ismail:
This fixes a test failure on PPC, PPC64 and PPC64LE.


ahnolds:
I suspect this is a spurious failure from Travis, @wsfulton want to restart the test?

wsfulton:
I've restarted, all looks good now.

PR #581, opened 2016-01-09
Ruby fix unbalanced braces causing issue with the YARD parser

aurelj:


wsfulton:
I've put in a slightly different fix as I don't like ```#ifdef 0```. This should solve the parser problem but shout if it doesn't.

aurelj:
Yes it solves the parser issue as well and is indeed nicer.
Thanks !

PR #582, opened 2016-01-11
Ruby add support for docstring option in %module()

aurelj:
This was already documented but not actually implemented.


aurelj:
I've added some basic test and fixed the minor memory leak (BTW: the exact same leak is present in the Python code).

I'm not exactly sure what you mean by being robust to the `*/` sequence. Do you mean swig should error out if detecting this or should it try to escape it with something like `*\/` ?
Anyway, somebody writing some invalid interface containing `*/` in the docstring will just get an invalid warper source that won't compile, so the compiler will point out the issue, anyway. So I don't really see this as an issue, but if you think I should improve on this, tell me what you expect exactly.

ojwb:
I've pushed a fix for the equivalent leak in Python - thanks for pointing that out.

The testcase would be better under `Examples/test-suite` really.

Here's an example of a valid interface file containing a docstring which would cause the problem:

```
%module(docstring="This module parses // and /* ... */ comments") commentparser
```

Escaping as you suggest would probably work, assuming that the tools which process these docstrings handle that.  If not, maybe replace `*/` as `*""/`, since AIUI Ruby concatenates adjacent string literals?

I suspect `"` and `\` in docstrings are probably problematic for Ruby, and should also be `\`-escaped.

(Looks like there's an equivalent bug in Python's handling - e.g.:

```
%module(docstring="This module parses \"\"\" quoted strings, as well as // and /
* ... */ comments") commentparse
```

Results in a syntax error in the generated `.py` wrapper:

```
"""This module parses """ quoted strings, as well as // and /* ... */ comments"""
```

And `\` also ought to be escaped there too - I'll sort out a fix for that...)

aurelj:
> The testcase would be better under `Examples/test-suite` really.

Well, there is already a testcase for module docstring in `Examples/test-suite/autodoc.i` so I don't think adding another one would do any good.

> Escaping as you suggest would probably work, assuming that the tools which process these docstrings handle that. If not, maybe replace `*/` as `*""/`, since AIUI Ruby concatenates adjacent string literals?

SWIG does not generate any Ruby code. The whole Ruby extension module is contained in the C/C++ generated wrapper, including the documentation, and the Ruby documentation tools (RDoc/YARD) are able to extract the documentation from this C/C++ source code.
So using `*""/` inside a C comment is not really a nice option.
And anyway, whatever escaping we choose, it will end up in the final generated documentation, so it will mess-up what the user expected. So now I think that we should simply consider `*/` as invalid in the docstring and just let the C/C++ compiler complain about it if someone dare to try it.

> I suspect `"` and `\` in docstrings are probably problematic for Ruby, and should also be `\`-escaped.

As this end up only in C/C++ code and never in any Ruby code, this is not an issue.

wsfulton:
Yes I think we can live with the parsing limitations. Can you remove the doc example though and make it a testcase to test the docstring is working? Also could you please document 'docstring' in Ruby.html instead of in the index.html file.

aurelj:
> Yes I think we can live with the parsing limitations.

OK.

> Can you remove the doc example though

Sure. Just skip the corresponding patch.

> and make it a testcase to test the docstring is working?

Not sure how you want me to do this...
As stated, there is already a testcase in `Examples/test-suite/autodoc.i`.
It just doesn't verify that the module docstring actually ends up in the
generated C++ file, but I have no idea how to do such a test in a clean way
with current test-suite.

> Also could you please document 'docstring' in Ruby.html instead of in
> the index.html file.

It is already documented in Ruby.html since a long time (even though it
was not actually implemented !).

wsfulton:
Regarding the test for this, what don't you no how to do? Is it how the test-suite works or how one would write some Ruby code to check for documentation? I can help with the test-suite, but surely there is a way to programmatically check for documentation in a module? The test just need to access the documentation comment.

aurelj:
The documentation is never part of the compiled ruby module. To make use of the documentation, the ruby way is to use rdoc (or yard) to parse the module's C++ source code and convert it to various formats (html, etc...). Doing this in the test would add a build dependency on rdoc and we would still be left having to parse the generated documentation to check if it contains what we expect.
It could look like this:
```
rdoc -o Examples/test-suite/ruby/autodoc Examples/test-suite/ruby/autodoc_wrap.cxx
grep -q '<p>hello\.</p>' Examples/test-suite/ruby/autodoc/Autodoc.html
```

Another simpler option would be to grep the generated C++ file looking for the expected comment string.
It would be as simple as:
```
grep -q 'hello\.' Examples/test-suite/ruby/autodoc_wrap.cxx
```
I think this would enough. What do you think ?

Anyway, whatever you think is best, I have no idea how such a check would be integrated in the test-suite.

wsfulton:
I see, thanks for the clear explanation. I don't intend to add this complexity to the test-suite, so we can go without the test in this case. 
```
	Printf(f_header, "/*\n  Document-module: %s\n\n%s\n*/\n", module, mod_docstring);
```
I don't think SWIG should impose the heading 'Document-module:' ...etc unless this is some Ruby documentation standard, so I'll change it to be completely free-form and up to the user as to what goes in to their documentation.

wsfulton:
Mmm, rdoc ignores 'Document-module:..." and doesn't seem to matter if it is the comment or not. So is this some sort of special keyword and is it needed by other doc tools? I couldn't find anything on this doing a quick search.

aurelj:
Yes, the `Document-module:...` is some [Ruby documentation standard](http://docs.seattlerb.org/rdoc/RDoc/Parser/C.html). Without it, the Autodoc.html generated by the previously described rdoc command does not contain the `hello.` string in the `module Autodoc` section.

wsfulton:
Great reference, thanks. Strange that this was documented in Ruby.html yet it was not implemented!

PR #596, opened 2016-01-23
Passing unit tests for Go on OSX

ahnolds:
This patches two unrelated bugs that were causing the Go unit tests to fail on OSX:

* Apparently, the linker on OSX requires that symbols be defined at link time, even if they are marked with the weak/weak_import attribute. This is different from the behavior of ld on Linux, which just marks the symbols as weak in the binary. I've found two possible solutions to this problem:
  1. Explicitly tell the linker that it's ok for the symbol in question to be undefined with the `-U` linker flag. This is the approach I took in this patch since it's much simpler, although it does result in the symbol being marked as undefined in the binary rather than weak. It looks like the undefined symbol is still resolved at runtime though (possibly with a slight performance penalty?), so I don't think this matters.
  2. Write and compile a dummy .dylib that exports the symbol in question, and weak link against it with the `-weak_library` linker flag. This is more complicated since you need to create a dummy library to link against. Some sources also suggest you need to also use the `-flat_namespace` linker command to allow the symbol to be loaded from a library with a different name at runtime.
* Calling a C/C++ function creates a swigargs struct intended for passing arguments and getting the return value. However, if the function takes no arguments and returns void, the resulting struct is empty. Empty structs are not allowed in C, so clang warns about empty structs in `extern "C"` blocks when -Wall is active. The solution here is to only create the swigargs struct if it isn't empty, since if it's empty it's also not used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -162,9 +162,6 @@ matrix:
     - compiler: gcc
       os: linux
       env: SWIGLANG=python SWIG_FEATURES=-O
-    - compiler: clang
-      os: osx
-      env: SWIGLANG=go
 before_install:
   - date -u
   - uname -a
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -5,6 +5,15 @@ See the RELEASENOTES file for a summary of changes in each release.
 Version 3.0.9 (in progress)
 ===========================
 
+2016-01-23: ahnolds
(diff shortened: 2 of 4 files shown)


wsfulton:
Very nice to see go on OSX working.
@ianlancetaylor - okay to merge?

ianlancetaylor:
Yes, looks good to me.  Thanks!

PR #597, opened 2016-01-23
Escaped char values in enums

wsfulton:
[C#, Java, PHP] Fix generated code parsing enum values using char escape sequences
when these values appear in the C#/Java/PHP code such as:
```
  enum X { x1 = '\n', x2 = '\1' };
```
Except C# does not support the octal escape sequences, so %csconstvalue
is still required to fix.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -5,6 +5,15 @@ See the RELEASENOTES file for a summary of changes in each release.
 Version 3.0.9 (in progress)
 ===========================
 
+2016-01-23: wsfulton
+	    [C#, Java, PHP] Fix generated code parsing enum values using char escape sequences
+            when these values appear in the C#/Java/PHP code such as:
+
+              enum X { x1 = '\n', x2 = '\1' };
+
+            Except C# does not support the octal escape sequences, so %csconstvalue
+            is still required to fix.
+
 2016-01-12: olly
 	    [Javascript] Look for "nodejs" as well as "node", as it's packaged
 	    as the former on Debian.
--- a/Examples/test-suite/chartest.i
+++ b/Examples/test-suite/chartest.i
@@ -12,4 +12,23 @@ char GetUnprintableChar() {
   return 0x7F;
 }
 
+static const char globchar0 = '\0';
+static const char globchar1 = '\1';
+static const char globchar2 = '\n';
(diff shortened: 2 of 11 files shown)


wsfulton:
This didn't work out as well as I'd hoped. Travis clang fails with: 
```
char_constant_wrap.c:328:19: error: illegal character encoding in character literal [-Werror,-Winvalid-source-encoding]
  result = (char)('<E1>');
                  ^
```
due to generated code changing from:
```c
  result = (char)('\341');
```
to
```c
  result = (char)('á');
```
when wrapping this (latin1 encoded):
```
#define SPECIALCHAR 'á'
```
I thought it might be better to generate the char exactly as it is in the input file (same encoding), but this opens up a can of file encoding worms, so probably need to revert it. I suspect that SWIG has previously encoded characters such as this using octal escaping to avoid this kind of problem, so probably it is best to restore this behaviour.

Regarding the code that will get generated as java/c#/php code, probably we should take a similar approach and use an escape sequence for non-ascii characters that works with that particular language, so '\341' would work in Java and PHP, but '\xE1' is needed in C#.

ojwb:
`'\341'` won't work for PHP - it'd need to be `"\341"`:

```
$ php <<'END'
<?php
$a = '\341';
print $a;
?>
END

\341
```

vs:

```
php <<'END'
<?php
$a = "\341";
echo $a;
?>
END

�
```

vadz:
FWIW Perl is the same as PHP (or, rather, vice versa...): escape sequences are only supported inside double, but not single, quotes.

ojwb:
I focused on PHP as that was mentioned in the original description.  If this change actually affects all the backend languages, we ought to consider each one rather than just hoping that any C/C++ character constant will be valid.

vadz:
Sorry, I didn't read this carefully enough and thought that the plan was to make this change for all the languages, please disregard the Perl comment above if it's not planned to change anything for it.

wsfulton:
I abandoned this approach and made language specific changes instead in  ac495d5c6635043569d555877480516b72641b81. These improved C#, D and Java. For PHP there was just one fix to accept '\0' and the values are still obtained from the C layer although this could be changed to be like Java and generate these values as PHP code.

PR #599, opened 2016-01-26
Passing unit tests for -classic python

ahnolds:
Support the full Python test suite in -classic mode.  The following are substantive changes:
* Convert long/unsigned long/long long/unsigned long long to PyInt rather than PyLong when possible. Certain python functions like len() require a PyInt when operating on old-style classes.
* Add support for static methods in classic mode, including support for pythonappend, pythonprepend, and docstrings.
* Removing the use of \_\_swig_getmethods\_\_ for static member methods since they will always be found by the standard argument lookup.
* Fix a bug where the wrong type of exception was caught when checking for new-style class support

The following cosmetic changes were made:
* Don't print a banner claiming compatibility with new-style classes when running in -classic mode

The following changes were made to the Python examples and test-suite:
* Support checking names of old-style classes in the import_packages example
* Add a check calling a static method with pythonappend/pythonprepend (previously the pythonappend/pythonprepend blocks were dropped for static methods in -classic mode)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -160,10 +160,6 @@ matrix:
     # Not quite working yet
     - compiler: gcc
       os: linux
-      env: SWIGLANG=python SWIG_FEATURES=-classic
-    # Not quite working yet
-    - compiler: gcc
-      os: linux
       env: SWIGLANG=python SWIG_FEATURES=-O
     # Not quite working yet
     - compiler: gcc
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -8,6 +8,18 @@ Version 3.0.9 (in progress)
 2016-01-27: steeve
             [Go] Ensure structs are properly packed between gc and GCC/clang.
 
+2016-01-25: ahnolds
+            [Python] Support the full Python test suite in -classic mode
+            * Convert long/unsigned long/long long/unsigned long long to PyInt
+              rather than PyLong when possible. Certain python functions like
+              len() require a PyInt when operating on old-style classes.
+            * Add support for static methods in classic mode, including support
+              for pythonappend, pythonprepend, and docstrings.
+            * Removing the use of __swig_getmethods__ for static member methods
+              since they will always be found by the standard argument lookup
+            * Fix a bug where the wrong type of exception was caught when
+              checking for new-style class support
+
 2016-01-23: ahnolds
             [Go] Enable support for the Go test-suite on OSX:
             * The linker on OSX requires that all symbols (even weak symbols)
--- a/Examples/python/import_packages/same_modnames1/runme.py
+++ b/Examples/python/import_packages/same_modnames1/runme.py
@@ -8,6 +8,12 @@ import pkg2.foo
 print "  Finished importing pkg2.foo"
 
 var2 = pkg2.foo.Pkg2_Foo()
-if str(type(var2)).find("'pkg2.foo.Pkg2_Foo'") == -1:
-    raise RuntimeError("failed type checking: " + str(type(var2)))
+
+classname = str(type(var2))
+# Check for an old-style class if swig was run in -classic mode
+if classname == "<type 'instance'>":
+    classname = str(var2.__class__)
+
+if classname.find("pkg2.foo.Pkg2_Foo") == -1:
+    raise RuntimeError("failed type checking: " + classname)
 print "  Successfully created object pkg2.foo.Pkg2_Foo"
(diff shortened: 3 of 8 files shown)


vadz:
This is obviously very useful for classic mode support, but I still wonder why don't we just remove classic mode completely instead...

ahnolds:
I have no objection to that plan - I kind of assumed from the comments on #232 that _someone_ was using classic mode as of about a year ago, but I don't know if that's still the case.  I seriously doubt anyone is using a version of Python before 2.2 when new-style classes were introduced, and if they are, I can't imagine they're updating to the latest version of SWIG on a regular basis.

ojwb:
I'm not entirely clear what `-classic` does - it doesn't seem to be documented aside from in `-help` (it's mentioned once in the Python chapter of the manual, where it notes that `-py3` disables it).  Looking at the backend code, it seems it's needed to support Python < 2.0.

We also have `-modern` which requires Python >= 2.2, and is somewhat better documented.

Looking at the option processing code, it seems that `-classic` and `-modern` are mutually exclusive (each turns the other off), but the default is neither on, whatever that means.  There's also `-nomodern` which undoes the effects of `-modern` (but is different to `-classic`).

I strongly feel we need to streamline the rather excessive number of options which the Python backend has accumulated, many of which aren't clearly documented, and most of which only seem to be documented by old entries in `CHANGES`.  It seems that any non-trivial Python change ends up breaking some other combination to the one the patch submitter is using, but there's no way we can sanely test all the possible combinations.

vadz:
We've discussed this here before (but somehow I can't find the issue where we did it right now) and even back then everybody seemed to agree that we could limit support to Python 2.6+ and by now I think even 2.7+ would be acceptable (but then I don't think we gain much from dropping 2.6 which is still used in some RedHat installations...). So I would definitely drop both `-classic` and `-modern` (and, ideally, also make `-builtin` default and remove it too, because why should we have two ways to generate the bindings when the "no builtin" doesn't seem to have any advantages?).

ojwb:
> why should we have two ways to generate the bindings when the "no builtin" doesn't seem to have any advantages?

According to the docs, there are limitations of `-builtin`:

http://www.swig.org/Doc3.0/Python.html#Python_builtin_limitations

The first few version-related ones aren't very relevant now, but some of the later ones seem like they might be reasons to not want to force `-builtin`.

ahnolds:
I'll also note that in `-builtin` mode, there are some compiler warnings that are generated (I compile with quite strict flags). In particular I saw `-Wunused-variable` for some instances of `PyObject *swig_obj[1]`, and some cases where `NULL` was being returned from methods that are supposed to return `int`. I'd really rather not have the default/only Python wrapper output generate compiler warnings if at all possible.

wsfulton:
@vadz, the discussion you are looking for is probably the recent one on the swig-devel mailing list http://article.gmane.org/gmane.comp.programming.swig.devel/23772. I can't see -builtin being the default while we have limitations with it. Compiler warnings using -builtin can surely be fixed though.

I don't mind accepting this patch to fix -classic for swig-3.0.x series, -classic will get dropped (along with the myriad of other options supporting older versions) in swig-3.1. I don't want this patch to cause any regressions, so we should look carefully at the change to return int instead of long in - 8df71912a7ea7e661c5ed67e4d5be0ad257b968e. If we don't have any tests for testing unsigned long/long around values of LONG_MAX and LONG_MIN I think we should add them in. SWIG_From_dec(unsigned long) and others will now return a different type depending on the size of the numeric value whereas it consistently returned a long. I wonder how problematic this might be?

Getting he test-suite working for -O though is a higher priority as that is probably what the default will be for swig-3.1.

ahnolds:
The tests added in 2f8a7b82 cover converting long around LONG_MAX and LONG_MIN.  I don't think they handle unsigned long/unsigned long long/long long in that range though.

ojwb:
Cool that we're going to raise the minimum Python 2 bar to a sensible position in 3.1.0.  I'd suggest >= Python 2.6, as that reflects the reality of what's actually in use now.

As a minimum, >= 2.5 would eliminate most of the version special casing, and the mean the only real restriction on `-builtin` would be not being able to raise the wrapped types as python exceptions (and I think that probably works if the class is marked with `%feature("exceptionclass")` or equivalent, which the Python docs don't currently mention).

I've cleaned up some lingering traces of Python 1.x I noticed while trying to work out what `-classic` actually does.

wsfulton:
Unfortunately we have not been testing all the versions of Python we though we were on Travis for a few months. Mostly we have just tested 2.7 and 3.2. I've just fixed this problem. Would you mind rebasing your changes on to latest master and resubmitting the pull request to run the tests again?

ahnolds:
Tests continue to pass after rebasing onto the latest master

PR #601, opened 2016-01-28
Python support for Bytes/Unicode distinction

ahnolds:
This patch is somewhat the opposite of edd36b28, in that it adds support for treating Python 3 Bytes objects as C/C++ char\*/std::string.  In particular, it adds support for overloaded C++ methods taking std::string and std::wstring.

Defining `SWIG_PYTHON_STRICT_BYTE_CHAR` changes std::string and char\* conversions to only accept and return Python 2 String and Python 3 Bytes objects.  By default, Python 3 Bytes are not accepted, and Python 3 (unicode) Strings are accepted and converted to UTF-8 before being converted to char\*.  Also by default, std::string and char\* are returned as Python 3 (unicode) Strings - this definition causes them to be returned as Python 3 Bytes instead.

Defining `SWIG_PYTHON_STRICT_UNICODE_WCHAR` changes std::wstring and wchar\_t\* conversions to only accept and return Python 2 Unicode and Python 3 (unicode) String objects.  By default, Python 2 String objects are also accepted and converted to unicode before being converted to wchar\_t\*.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -5,6 +5,11 @@ See the RELEASENOTES file for a summary of changes in each release.
 Version 3.0.9 (in progress)
 ===========================
 
+2016-01-27: ahnolds
+            [Python] Added support for differentiating between Python Bytes
+            and Unicode objects using by defining SWIG_PYTHON_STRICT_BYTE_CHAR
+            and SWIG_PYTHON_STRICT_UNICODE_WCHAR.
+
 2016-01-27: steeve
             [Go] Ensure structs are properly packed between gc and GCC/clang.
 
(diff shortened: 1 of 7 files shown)


wsfulton:
Looks okay to me. Unfortunately the Travis tests havn't been testing all version of Python properly, so the tests havn't actually passed. Can you fix your new testcase please, eg for python 2.5, I get:
```
checking python testcase python_strict_unicode (with run test)
  File "python_strict_unicode_runme.py", line 3
    test_bytes  = b'hello \x01world\x99'
```                                       ^
I suggest fixing these problems then rebasing on top of master to pick up the fixes I committed today that run correctly run different versions of python.

ahnolds:
Done - unit tests now pass on all versions.  I just fixed them to avoid the b prefix for string literals in Python < 2.6

Issue #604, opened 2016-02-03
csharp: duplicate symbols in generated C++ code from multiple SWIG modules


pjohangustavsson:
Yes, this is in the context where classes sharing names belong to different namespaces are wrapped. At least offer the option to include the namespace identifier in the C++ wrapper code function names. P/Invoke does not enforce any naming conventions the way JNI does, and the Java module of SWIG does not suffer from this issue.

wsfulton:
SWIG does not support symbols with the same name in different namespaces, this is all documented. You can use the %nspace feature to solve this problem. If this isn't working, then please provide some specifics.

pjohangustavsson:
%nspace is a feature that affects how C# or Java code is generated, not C++ wrapper code. Another feature or %module setting might be an option? I believe many would desire something like that.

wsfulton:
%nspace does affect the c++ code generated by mangling in the C++ namespace into the C PInvoke functions. I still don't know what this issue is about as you are not providing any specifics.

pjohangustavsson:
Either way, using %nspace while -namespace is used for a module is not a good idea when the SWIG interface file is intended to represent a namespace that sits at global scope. 

Assume a namespace Logging. You have a logging.i file you pass to SWIG with -namespace Logging. In the original C++ code, Logging contains a class Context that is troublesome. I do %nspace Logging::Context in logging.i. In C# this becomes Logging.Logging.Context. If I remove the -namespace argument, %nspace works but I am left to enable that feature for all classes, else the logging.i file becomes troublesome to %import in other modules. 

Maybe the problem is this: the -namespace argument isn't working for C# the same as the -package argument works for Java. In other words, the -namespace argument do not mangle the names of the PInvoke wrapper functions, while -package do. Is this an issue or by design?

pjohangustavsson:
I've made a patch which checks if -namespace is set and formats the wrapper function names from a formatted string (using Printf) which I then pass to SWIG_name_register, as follows:

String mangled = NewStringf("");
Printf(mangled, "CSharp_%s_%%f", namespc);
SWIG_name_register("wrapper", mangled);
Delete(mangled);

Works like a charm, but will also put wrappers of symbols found at global scope in that namespace as well, but requires less effort by far to adjust than the other way around. Please introduce this patch. I will request a pull shortly. 

PS I think it's ludicrous that you're not using cmake. Took me less than one hour to setup a rough cmake script and successfully build swig on Windows with VS2015.

[...]

PR #605, opened 2016-02-04
csharp: give -namespace meaning to the generated c++ wrapper code

pjohangustavsson:
When people use the -namespace option, they often intend for the input to swig to wrap a particular namespace in a single module. This option should also have an effect on how C++ wrapper function names are mangled, and not just the C# proxy code.

This fix solves many issues related to duplicate symbols when linking the source from multiple swig modules into one shared library. When -c++ and -namespace are used together, global symbols that might be placed in the namespace specified with the -namespace argument is a rare occurrence and simple to fix.

### DIFF
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -415,11 +415,19 @@ public:
     }
 
     Printf(f_runtime, "\n");
+    if (namespce) {
+        String *wrapper_name = NewStringf("");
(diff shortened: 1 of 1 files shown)


wsfulton:
Thanks, I've fixed the incorrect spacing in this patch and removed the unnecessary whitespace changes and committed in 1458326f4c93de3f3413e4963ff047dc6823e9d8.

Closes #604.

FObermaier:
Can this please depend upon `SWIG2_CSHARP` not being set?

PR #612, opened 2016-02-12
Make SWIG modules compatable with imports from namespace packages

mromberg:
This patch uses importlib.load_module and the SWIG shadow modules __name__ attribute to
locate and load the C module _foo.so rather than the __file__ attribute.  This makes SWIG modules compatable with python loaders that do not place modules in the same filesystem path (or even on a filesystem at all).  With this patch the SWIG shadow module can be placed anywhere a python importer can find it and the C module (shared object file) can be placed on a filesystem where the operating system can load it.

  The patch uses importlib for python versions 2.7 and newer.  All older python implementations are left unchanged.  I'm not sure when python added the importlib module so it is possible this could be used for earlier implementations as well.

  I've added an example which demonstrates and tests this with namespace packages and the zipimporter.   I'm not 100% sure I did the Makefile for the example/tests the correct way.  It seems to work.  But if it needs changes let me know.


ahnolds:
Per [the docs](https://docs.python.org/2/library/importlib.html), importlib was indeed added in 2.7 (and a quick look at the history for CPython confirms this).

wsfulton:
@mromberg I think this is okay to merge in. For some reason the Travis tests did not run for your last commit and given the nature of the changes, I'd like to see them all pass before merging. If you complete the minor improvements below and resubmit the pull request, then the Travis tests should be run for a final check...

* Add a bit of info about this into the Python.html docs in the "Python Packages" section (add in a *compatibility note* for when the Python features (frozen/namespace packages) were introduced (that will be swig-3.0.9 if done very soon)
* Fix spelling off seperate -> separate.
* Check for python-3.5.1 or later for the parts of the test that require this as a minimum version.

Thanks for the improvement.

Source/Modules/python.cxx
```diff
@@ -854,7 +863,14 @@ class PYTHON:public Language {
       Printv(f_shadow, "del version_info\n", NULL);
 
       if (builtin) {
-	Printf(f_shadow, "from %s import *\n", module);
+        /*
+         * Python3 removes relative imports.  So 'from _foo import *'
+         * will only work for non-package modules.
+         */
+        Printv(f_shadow, "if __name__.rpartition('.')[0] != '':\n", NULL);
+        Printf(f_shadow, tab4 "from .%s import *\n", module);
```
ptomulik: Maybe:
```c++
Printv(f_shadow, tab4 "from %s%s import *\n", (py3 ? "." : ""), module )
```
for py2 backward compatibility?

mromberg:
The existing code in the master branch inserts this into the shadow wrapper if -builtin is enabled:

from _foo import *

  Near as I can tell, the only reason this ever worked at all is that the existing import helper function generated by swig always placed the _foo.so module into the *global* namespace.  It was never part of a package at all.  So, "from _foo import *" worked.  At least that is my understanding of the current behavior (could be wrong).

  My changes correctly place foo.py and _foo.so into the package based on the __name__ attribute.  Exacly why "from _foo import *" needs to be done with -builtin, I'm not sure.  But it no longer works with python3 if foo.py and _foo.so are correctly placed into the package.

  So, this change replaces 'from _foo import *' (for builtin) with:
if __name__.rpartition('.')[0] != '':
    from ._foo import *
else:
    from _foo import *
 
  It is my intention that this code (and my changes to the import helper) should work no matter what package foo is located in.  It should also work even if foo is moved or completly removed from a package.

  There could be a more elegant way to do this.  But swig supports so many historical versions of python it is hard to keep track of how python imports things for each one.

mromberg:
Also, I apologize for the current state of this pull request looking like a giant mess.  I wanted to combine my branch with the latest commits on master so I could run the tests on my local machine.  It looks like rebaseing was not the correct way to do this as the pull request now shows signs of that rebase.

  A git diff on my branch relative to master still shows just my changes.  If desired, I can cancel this pull request and make another if that would look cleaner.

Source/Modules/python.cxx
mromberg: >>>>> " " == Pawe� Tomulik <notifications@github.com> writes:

    >> @@ -854,7 +863,14 @@ class PYTHON:public Language {
    >> Printv(f_shadow, "del version_info\n", NULL);
    >> 
    >> if (builtin) { - Printf(f_shadow, "from %s import *\n",
    >> module); + /* + * Python3 removes relative imports.  So 'from
    >> _foo import *' + * will only work for non-package modules.  +
    >> */ + Printv(f_shadow, "if __name__.rpartition('.')[0] !=
    >> '':\n", NULL); + Printf(f_shadow, tab4 "from .%s import *\n",
    >> module);

     > Maybe: ```c++ Printv(f_shadow, tab4 "from %s%s import *\n",
     > (py3 ? "." : ""), module ) ``` for py2 backward compatibility?

  Thanks for the tip on where to locate this!   At first glance this
looks like a more straight forward solution than the one I committed.
The down side to it is that it will not work if the module is built
for python3 and not placed into a package.  The shadow file will still
have 'from ._foo import *' in it and this will fail because _foo.so is
in the global namespace.

Mike

ptomulik: Actually, my comment was for the one line only. In fact, I thought about something like this (to have context):
```c++
Printv(f_shadow, "if __name__.rpartition('.')[0] != '':\n", NULL);
Printv(f_shadow, tab4 "from %s%s import *\n", (py3 ? "." : ""), module );
Printv(f_shadow, "else:\n", NULL);
Printf(f_shadow, tab4 "from %s import *\n", module);
```

BTW, the module and package names are already known at "swigging" time. Then it's not necessary to generate the whole conditional python code to be executed at runtime. I think, the following should be enough (instead of the four lines above):

```c++
Printv(f_shadow, "from %s%s import *\n", (py3 && package && *Char(package)) ? "." : "", module );
```


You may also take a look at ``importDirective()`` and ``((abs|rel)_)?import_directive_string()`` functions.

ptomulik:
Rebasing against master should be enough (``git checkout master && git pull upstream master`` then ``git checkout implpkg && git rebase master``). After rebase, ``git push --force origin implpkg`` is necessary.

wsfulton:
Thanks, this a useful enhancement. I've merged it in d01efd82e12cc7fc1c04a76e7caad255ad0578ad after removing the whitespace changes in python.cxx.

PR #613, opened 2016-02-13
swiglabels.swg: added check if __GNUC__ is defined

pjohangustavsson:
It makes sense to check if the preprocessor definition `__GNUC__` is defined before using it as a value.


ojwb:
Thanks

PR #617, opened 2016-02-18
Implement the operator overload test suite for Python.

ejulien:
The suite fails on Python 3 (with and without -builtin) due to the following issues:

- Operator / is mapped to the div slot which is now truediv in Python3 (see PR #136 for a WIP fix).
- Operators +=, -=, *= and probably others assign the method return type (void) to the object they are applied on. A workaround is to return *this from the affected operators.

Also included in this patch is support for the PY3BUILTIN flag to run the test suite with the -builtin SWIG option (`make PY3=1 PY3BUILTIN=1 operator_overload.cpptest` from `swig/Examples/test-suite/python`).

### DIFF
--- a/Examples/Makefile.in
+++ b/Examples/Makefile.in
@@ -292,11 +292,7 @@ PYTHON_SO     = @PYTHON_SO@
 ifeq (,$(PY3))
   SWIGPYTHON = $(SWIG) -python
 else
-  ifeq (,$(PY3BUILTIN))
-    SWIGPYTHON = $(SWIG) -python -py3
-  else
-    SWIGPYTHON = $(SWIG) -python -py3 -builtin
-  endif
+  SWIGPYTHON = $(SWIG) -python -py3
 endif
 
 # ----------------------------------------------------------------
--- a/Examples/Makefile.in
+++ b/Examples/Makefile.in
@@ -292,7 +292,11 @@ PYTHON_SO     = @PYTHON_SO@
 ifeq (,$(PY3))
   SWIGPYTHON = $(SWIG) -python
 else
-  SWIGPYTHON = $(SWIG) -python -py3
+  ifeq (,$(PY3BUILTIN))
+    SWIGPYTHON = $(SWIG) -python -py3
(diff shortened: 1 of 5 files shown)


ejulien:
From the CI results the `/` operator is broken:
* On Python 2 when `-builtin` is used.
* On Python 3 regardless of the `-builtin` flag.

ejulien:
@wsfulton I think the last two commits form an acceptable fix for the division issue in Python 2.x and 3.x with and without -builtin used. I'm pretty confident in the modifications I made but knowledgeable feedback would be nice.

I am trying to tackle the last issue for the operator_overload test suite to pass. More precisely the `+=` operator corrupts the object it is applied to because the Python VM executes `a+=b` as `a=a+b` so `a` is assigned the return value of `void operator+=(...`), so `None` in that case.

I almost have a solution working but I don't think my approach is very nice. I'll polish it up and open a new PR so we can discuss it without polluting this one.

ejulien:
I could not get my initial plan working to solve the issue with +=. It looked easy enough but I don't think I can manage a fix in the time span I have left for this task.

For what its worth my idea was to:
* Flag with a new feature the affected operator nodes (very easy to do in `pyopers.swg`).
* Catch that feature flag in the `functionWrapper` method of `python.cxx` and fix the return type there.

I got stuck because the out typemap expects a variable named `result` to carry the out value and it is obviously not setup. I have no idea how to properly create let alone assign `*this` to that variable.

As a fallback I modified the `operator_overload.i` interface to `return *this` from operators such as `+=`, that should let the complete suite pass.

This modification affects all target languages, if this is a problem I'll get rid of the failing operator tests so that the division fix can be merged. Note that the failing tests I'm talking about are not run at all for Python even in the most current SWIG version as far as I can tell.

wsfulton:
This looks okay to me, thanks. I'd rather not have introduced the SWIGPYTHON_PY3 macro in the SWIG preprocessor, but I can't think of any solution. I'll merge it in shortly.

PR #628, opened 2016-03-07
[PHP] Fix #627

### LINKED ISSUE #627: [PHP] Wrong code generation for attribute "feature:director:except"
When use [Exception unrolling](http://www.swig.org/Doc3.0/SWIGDocumentation.html#Php_nn3_4) feature:
```
%feature("director:except") {
    if ($error == FAILURE) {
        throw Swig::DirectorMethodException();
    }
}
```
generated code is not correct, i.e. when add this feature to the `php/callback` example, then get:
```cpp
void SwigDirector_Callback::run() {
  ...
  int error;
  ...
  call_user_function(EG(function_table), (zval**)&swig_self, &funcname, result, 0, args TSRMLS_CC);
  {
    if (error == FAILURE) {
      throw Swig::DirectorMethodException();
    }
  }
  ...
}
```
Local variable **error** is not initialized and not defined later, which leads to the unpredictable behavior.
[This patch](https://github.com/swig/swig/files/161113/php_feature_director_except_fix.txt) fix this issue and generate next code:
```cpp
void SwigDirector_Callback::run() {
  ...
  int error;
  ...
  error = call_user_function(EG(function_table), (zval**)&swig_self, &funcname, result, 0, args TSRMLS_CC);
  {
    if (error == FAILURE) {
      throw Swig::DirectorMethodException();
    }
  }
  ...
}
```
Now local variable `error` contains result of the function  `call_user_function()`.

SergeySeroshtan:
Fix issue https://github.com/swig/swig/issues/627

### DIFF
--- a/Source/Modules/php.cxx
+++ b/Source/Modules/php.cxx
@@ -1,5 +1,5 @@
 /* -----------------------------------------------------------------------------
- * This file is part of SWIG, which is licensed as a whole under version 3 
+ * This file is part of SWIG, which is licensed as a whole under version 3
  * (or any later version) of the GNU General Public License. Some additional
  * terms also apply to certain portions of SWIG. The full details of the SWIG
  * license and copyrights can be found in the LICENSE and COPYRIGHT files
@@ -2530,7 +2530,7 @@ done:
 
     Printf(w->code, "TSRMLS_FETCH_FROM_CTX(swig_zts_ctx);\n");
 
-    /* declare method return value 
+    /* declare method return value
      * if the return value is a reference or const reference, a specialized typemap must
      * handle it, including declaration of c_result ($result).
      */
@@ -2622,6 +2622,7 @@ done:
       }
 
       /* exception handling */
+      bool error_appended = false;
       tm = Swig_typemap_lookup("director:except", n, Swig_cresult_name(), 0);
       if (!tm) {
 	tm = Getattr(n, "feature:director:except");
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for the patch.

However, the conditions under which `error` is set by your patch don't seem to match the conditions under which `int error;` is emitted higher up.  If there's a typemap but it doesn't use `$error`, it looks like your patch would result in code which fails to compile.  We probably need to do something like setting a flag when `$error` is used in the typemap and use that to trigger both.

It would also be good to have a regression testcase covering this, to ensure it's actually fixed and stays fixed.

ojwb:
@SergeySeroshtan In #627 you commented:
> I'll try to add the regression test as soon as possible.

But that was in late March, so I guess it's slipped off your radar.  Any chance you could sort out a regression test to go with this fix?

ojwb:
I took a look at patching `Examples/test-suite/director_exception.i` to provide a regression test.  I couldn't managed to actually produce one, I think because `error` happens to get initialised to a value other than `-1` (which is what FAILURE is).

However, this made me realise that the manual is actually wrong:

> By default, the directors ignore exceptions that occur during method calls that are resolved in PHP.

The exception is not in fact ignored in this case - by default the PHP interpreter knows there was an exception and we return PHP `NULL` from the function.  Assuming the `directorout` typemap handles that gracefully (the default ones should) then once control returns back to PHP once more the exception is still ready to go.

So I think the exception unrolling feature isn't really needed in PHP.

@SergeySeroshtan What led you to use this feature?

ojwb:
Since we have this feature, it ought to generate working code, so I added a regression test and have merged this.  Leaving open as we should fix the manual to not mislead.

ojwb:
Docs improved, closing.

Issue #563, opened 2015-11-18
[python] _swigconstant helpers needlessly generated for primitive types


yarikoptic:
Please pardon my ignorance in swig since I haven't actively used it in the past years.  While preparing new release of PyMVPA I have ran into failed builds due to the fact that constants we had defined in svmc.i interface (https://github.com/PyMVPA/PyMVPA/blob/master/mvpa2/clfs/libsvmc/svmc.i#L142)
```C
enum { C_SVC, NU_SVC, ONE_CLASS, EPSILON_SVR, NU_SVR }; /* svm_type */
enum { LINEAR, POLY, RBF, SIGMOID, PRECOMPUTED };	/* kernel_type */
```

used to build Python bindings acquired _swigconstant suffixes all of a sudden.  with 3.0.4 it is fine, with 3.0.5 we get already _swigregister (introduced in rel-3.0.3-5-gc21e242) which then later got renamed in  rel-3.0.3-6-gcfaf2f9 into _swigconstant.

I guess it all was intentional and for common good, but I thought to double-check and may be seek some guru advice if there is possibly a cmdline (or other) option to revert to original behavior, possibly making it not too complicated to support multiple versions of swig across various platforms.  Thank you in advance!

yarikoptic:
FWIW for now I have worked around by not using swig3.0 and relying on swig2.0 in Debian so the problem is not that "pressing"

ojwb:
Can you provide a small, self-contained example and explain exactly what the problem with the code generated by the latest version is?  Please also include the SWIG command line to process the file.

yarikoptic:
here is not quite a minimal but quite verbatim and complete as working for me on a Debian testing/sid:

```
git clone https://github.com/PyMVPA/PyMVPA  ; cd PyMVPA

swig -python -I/usr/lib/python2.7/dist-packages/numpy/core/include -I3rd/libsvm -c++ -I/usr/lib/python2.7/dist-packages/numpy/core/include -I3rd/libsvm -o mvpa2/clfs/libsvmc/svmc_wrap.cpp -outdir mvpa2/clfs/libsvmc mvpa2/clfs/libsvmc/svmc.i

g++ -I/usr/lib/python2.7/dist-packages/numpy/core/include -I3rd/libsvm -I/usr/lib/python2.7/dist-packages/numpy/core/include -I/usr/include/python2.7 -c mvpa2/clfs/libsvmc/svmc_wrap.cpp -fPIC

[...]

Issue #344, opened 2015-02-23
M2Crypto bindings no longer functional with swig 3.0.5


DimStar77:
See also the M2Crypto issue 
https://github.com/martinpaljak/M2Crypto/issues/60
as well as https://bugzilla.suse.com/show_bug.cgi?id=917759

With Swig 3.0.5, the bindings for openssl created using M2Crypto are no longer functional:

<pre>
$ python
Python 2.7.6 (default, Sep  9 2014, 15:04:36) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import M2Crypto
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Volumes/Kristofer/ExternalUsers/kyounger/Projects/tioga-flask-mini/venv/lib/python2.7/site-packages/M2Crypto/__init__.py", line 24, in <module>
    import ASN1
  File "/Volumes/Kristofer/ExternalUsers/kyounger/Projects/tioga-flask-mini/venv/lib/python2.7/site-packages/M2Crypto/ASN1.py", line 12, in <module>
    import BIO
  File "/Volumes/Kristofer/ExternalUsers/kyounger/Projects/tioga-flask-mini/venv/lib/python2.7/site-packages/M2Crypto/BIO.py", line 221, in <module>
    class CipherStream(BIO):
  File "/Volumes/Kristofer/ExternalUsers/kyounger/Projects/tioga-flask-mini/venv/lib/python2.7/site-packages/M2Crypto/BIO.py", line 227, in CipherStream
    SALT_LEN = m2.PKCS5_SALT_LEN
AttributeError: 'module' object has no attribute 'PKCS5_SALT_LEN'
</pre>

I attempted to bisect swig between 3.0.4 (working) and 3.0.5 and ended up on commit
commit c21e2423a02939d2a4bc02bdfaa5c3bed3be95b2
Author: Paweł Tomulik <ptomulik@meil.pw.edu.pl>
Date:   Sat Oct 25 22:56:41 2014 +0200

    make %constant directive to work with structs/classes

as being the first bad commit.

DimStar77:
Some more digging:
it appears as if all %constant (and also #define, they seem to be treated equally) ARE exported, BUT:

%module _m2crypto
%constant int PKCS5_SALT_LEN=8;

[...]

PR #631, opened 2016-03-09
[python] prevent _swigconstant boilerplate for non-class types

ptomulik:
In response to  #344, #563.

This patch reverts the old way of generating code for ``%constant``s if the constant's type is not a class. It shall reduce the code-bloat for primitive constants (``int``, ``enum`` and so on). I hope it fixes [regression](https://github.com/swig/swig/issues/563#issuecomment-193542011) mentioned in #563.

Note, that cases such as:

```c++
struct T { /* ... */  };
T v1, v2, v3, v4, v5;
//...
%constant T& t1 = &v1;
%constant T const& t2 = &v2;
%constant T* t3 = &v3;
%constant T const* t4 = &v4;
%constant T const* const& t5 = &v5;
```
will be handled as class objects (i.e. ``_swigconstant`` will be generated) -- pointers, references, qualifiers and template arguments get stripped before determining whether the constant's type is a class or not.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -3458,11 +3458,15 @@ public:
     String *name = Getattr(n, "name");
     String *iname = Getattr(n, "sym:name");
     SwigType *type = Getattr(n, "type");
+    SwigType *basetype = SwigType_base(type);
     String *rawval = Getattr(n, "rawval");
     String *value = rawval ? rawval : Getattr(n, "value");
     String *tm;
     int have_tm = 0;
     int have_builtin_symname = 0;
+    int is_classobj = SwigType_isclass(basetype);
+
+    Delete(basetype);
 
     if (!addSymbol(iname, n))
       return SWIG_ERROR;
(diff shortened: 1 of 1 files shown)


wsfulton:
The unnecessary _swigconstant helpers are nearly all gone. I compared the test-suite to just before #250 was pulled and the li_std_stream.i and li_std_wstream.i testcases are the only ones generating _swigconstant helpers compared to before #250. These are for std::endl, std::ends, std::flush in Lib/std/std_iostream.i. I suspect SwigType_isclass() is failing in some way, but is probably a separate bug.

wsfulton:
Merged in c1d200ba7da53c50dbb682adb2319994e27500c9

ptomulik:
Perhaps std:endl and siblings are somehow treated as constants by swig (in C++ they're global variables AFAIR). Then they're constant objects of class type, so they will generate the _swigconstant helpers, and it should be expected I think.

ptomulik:
I was wrong, std::endl is a function (pointer to a function?).

ptomulik:
It may be ``SwigType_base`` (or rather the fact that I have used it). I suspect that for function and function pointer types it returns the return type of function, which in our case is a class type (std::ostream for std::end(...)).

ojwb:
I've tested this with [xapian](https://xapian.org/)'s bindings, and all the `_swigconstant` functions are all gone again there.  Thanks for the fix.

ptomulik:
further fixes to ``%constant``'s are included in #641. You can test it as well.

ojwb:
The further fixes sound good, but I don't have any function pointer constants to check the wrapping of.

PR #633, opened 2016-03-11
Suppress unmessage at package load

sbpcs59:
The generated R 'packagename.R' file defines a 'print' method that overrides the base package
print, so 'ExternalReferences' can be printed. This is all just fine, but because a 'generic' function
is not defined for base::print(), an unnecessary informational message is printed at package load, so for example for the 'redland' R package:

```
Creating a generic function for ‘print’ from package ‘base’ in package 'redland'
```

In the interest of reducing unneeded messages at package load, I've introduced this PR to
simply suppress this message from being printed.

### DIFF
--- a/Lib/r/r.swg
+++ b/Lib/r/r.swg
@@ -264,8 +264,8 @@ x
 setAs('ExternalReference', 'character',
 function(from) {if (!is.null(from$"__str__")) from$"__str__"()})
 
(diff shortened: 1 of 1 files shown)


wsfulton:
@joequant, can you okay this patch please? It seems to be related to fe39ef5faee7c66a87053096c41ab1bbace9c26a.

joequant:
Yeah.  I'm OK.

PR #634, opened 2016-03-11
Add missing UNIX build step

masonicboom:


wsfulton:
This isn't correct as autogen.sh should NOT be run from the installation tarball. Developers installing from Github do need to run autogen.sh though but there are full details at http://swig.org/svn.html. The above instructions are meant to be for users using the tarball, so any changes to this page should clarify all this and possibly point to the developer info on svn.html page.

masonicboom:
Thanks for the correction. For posterity, here was the path that led to this PR:

1. Desire to build SWIG from source.
2. Go to https://github.com/swig/swig.
3. "Installation" section of README says read Doc/Manual/Preface.html#Preface_installation.
4. That doc page does not mention running autogen.sh, consequently, can't build SWIG.

PR #576, opened 2015-12-24
Gsoc2012 objc

maxep:
Hi,

This PR merges master into gsoc2012-objc. 
It fixes a memory leak in objc.cxx and the objc configuration.
Examples pass on my osx but not on travis.
Test-suite is still broken for objc...

Regards,

Maxime

PR #639, opened 2016-03-23
[python] fix function shadow code

ptomulik:
Consider this example:
``` cpp
%module example

%{
int foo() {}
%}

int foo();
```

If I run swig as follows
```
SWIG -python -o example_wrap.cc example.i
```
the shadow code generated to ``example.py`` for this function is
``` python
def foo()
    return _test.foo()
foo = _test.foo
```

I believe, it should be just
``` python
foo = _test.foo
```

This patch fixes it, such that the unnecessary boilerplate is not generated.


Source/Modules/python.cxx
```diff
@@ -2376,21 +2376,22 @@ class PYTHON:public Language {
...
-    Printv(f_dest, "\ndef ", name, "(", parms, ")", returnTypeAnnotation(n), ":\n", NIL);
-    if (have_docstring(n))
-      Printv(f_dest, tab4, docstring(n, AUTODOC_FUNC, tab4), "\n", NIL);
-    if (have_pythonprepend(n))
-      Printv(f_dest, indent_pythoncode(pythonprepend(n), tab4, Getfile(n), Getline(n), "%pythonprepend or %feature(\"pythonprepend\")"), "\n", NIL);
-    if (have_pythonappend(n)) {
-      Printv(f_dest, tab4 "val = ", funcCall(name, callParms), "\n", NIL);
-      Printv(f_dest, indent_pythoncode(pythonappend(n), tab4, Getfile(n), Getline(n), "%pythonappend or %feature(\"pythonappend\")"), "\n", NIL);
-      Printv(f_dest, tab4 "return val\n", NIL);
+    String *retAnnotation = returnTypeAnnotation(n);
```
vadz: What's the purpose of this change? This variable is used only once...

vadz:
I was also very puzzled by the current behaviour when I discovered it for the first time as well, but I assumed that there had to be some reason for it. I admit I have no idea what could it be, but it seems hard to imagine that someone did this accidentally...

Source/Modules/python.cxx
ptomulik: You're right. This is something that I've accidentally left-over from my previous versions of the patch.

ptomulik:
The commit: 8a6874e introduced this exact condition. It was a part of #576 which contains 250+ commits. The original condition in form:
```cpp
if (Getattr(n, "feature:python:callback") || !have_addtofunc(n)) {
      /* If there is no addtofunc directive then just assign from the extension module (for speed up) */
      Printv(f_dest, name, " = ", module, ".", name, "\n", NIL);
    } 
```
was introduced together with ``emitFunctionShadowHelper()`` in 2824b0c, which is another huge commit (looks like some sort of import from SVN).

ptomulik:
"then just assign from the extension module" (note the "just") in the comment may suggest, that this assignment was intended to be an alternative for the function code generation.

wsfulton:
How is #576 relevant to the original introduction of this code? I don't see how it is relevant to code in master. 

While this patch seems to make sense, I don't really know what odd ramifications it may have so I will not accept this for swig-3.0.x. I'll use this patch for swig-3.1 which is going to drastically change and optimise the python layer.

ptomulik:
From what I've seen, one of last changes to ``emitFunctionShadowHelper()`` was introduced by #576, see [this line of diff](https://github.com/swig/swig/pull/576/commits/8a6874e6337843fb6584cf298bb0b526776ec437#diff-3cc91cccc4a3a707448c1f5993015108L2392). However, it doesn't seem to be a crucial change to generator's logics. The generation of doubled code seems to be there from the beginning, i.e. since 2824b0c.

ojwb:
That wasn't really part of #576 - probably master got merged to the branch before the branch was merged, or something like that.  Unhelpfully github just shows it as "wsfulton and others added some commits".

mromberg:
Here is a quote from the swig python documentation concerning the low level wrappers:

"The fact that the class has been wrapped by a real Python class offers certain advantages. For instance, you can attach new Python methods to the class and you can even inherit from it (something not supported by Python built-in types until Python 2.2)."

  And a built-in or low level function is not exactly identical to one that is wrapped as swig is doing it by default.   One instance where this differs I bumped into recently.  The __doc__ attribute is immutable with a built-in but not so with a "normal" python function.   I suspect many other things such the following would also be affected:

https://www.python.org/dev/peps/pep-0232/

  Overall, I like swig doing things like this (eliminating the middle man) with the -O switch.  I think this change should be rolled into the -O switch for 3.1.  

  And I also guess that swig should at least document the fact that not all code will work 100% exactly the same way with -O as without it for all possible edge cases a user may have.  If this is not a popular position then I think swig should either optimize all the time or never optimize.  Because when the output generated by swig is python code it is very hard to guarantee identical behavior without literally only creating the code in exactly one way.

wsfulton:
Let's use the `-O` switch idea. After all the command line simplifications, see #1340, `-O` comprises the individual options `-fastdispatch -fastproxy -fvirtual`. `-fastproxy` is now explained in the updated docs in git, see https://htmlpreview.github.io/?https://github.com/swig/swig/blob/master/Doc/Manual/Python.html#Python_fastproxy. Note that `-olddefs` adds in the 2nd faster call to the wrapped global method. Consider wrapping a member method:
```swig
%inline %{
struct X {
  int foo() {}
};
%}
```
The default generated output

wsfulton:
Oops, I posted that unfinished, so edited it to remove some junk.

All I wanted to say next is that we should change the code generation for wrapped global C/C++ functions to be consistent to C++ class wrappers. So for the foo example given in this issue, the default output should be:

```python
def foo():
    return _example.foo()
```
Using `-fastproxy` would result in:
```python
foo = _example.foo
```
Using `-fastproxy -olddefs` would result in what is currently output:
```python
def foo():
    return _example.foo()
foo = _example.foo
```
I'll implement this unless there are any objections.

PR #641, opened 2016-03-23
fix handling of function pointers by %constant directive

ptomulik:
With this fix, ``%constant`` directive correctly handles function pointers. I believe, this fixes the issue with ``std::endl`` etc. mentioned by @wsfulton  [here](https://github.com/swig/swig/pull/631#issuecomment-198248842) in PR #631.


wsfulton:
Thanks @ptomulik, that has indeed fixed std::endl etc mentioned in #631.

PR #645, opened 2016-04-05
R character vector support

richardbeare:
Existing support for transparent conversion between R string vectors and c++ ```std::vector<std::string>``` was incomplete. std::strings were imported, but incorrect tags remained. This patch corrects
the types and adds support for exporting to ```std::vector<std::string>```. There is also a simple
run test that compares a c++ reverse of strings (not string contents) to the R equivalent.


wsfulton:
Ignore the Appveyor Java failure (I've restarted the test so it might pass later today). 

@joequant one for you, looks okay to me.

Issue #646, opened 2016-04-06
Java integer types do not work correctly with -DSWIGWORDSIZE64


johnwason:
On most systems, `int64_t` is the same as `long long`, and  `uint64_t` is the same as `unsigned long long`.  On gcc Linux x86_64 however, `int64_t` becomes `long` and `uint64_t` becomes `unsigned long`.  To handle this the switch `-DSWIGWORDSIZE64` is used.  While Python and C# handle this switch correctly, Java does not.  It tries to typedef `int64_t` to a java `int` when `-DSWIGWORDSIZE64` is specified because it sees the `long`/`unsigned long` types. It should be converting `int64_t` to a java `long` and `uint64_t` to `BigInteger` in all cases. It should detect that `-DSWIGWORDSIZE64` is defined and translate accordingly.

Mizux:
My 2 cents currently I use this workaround
```swig
%include "typemaps.i"
...
#if defined(SWIGJAVA)
#if defined(SWIGWORDSIZE64)
%define PRIMITIVE_TYPEMAP(NEW_TYPE, TYPE)
%clear NEW_TYPE;
%clear NEW_TYPE *;
%clear NEW_TYPE &;
%clear const NEW_TYPE &;
%apply TYPE { NEW_TYPE };
%apply TYPE * { NEW_TYPE * };
%apply TYPE & { NEW_TYPE & };
%apply const TYPE & { const NEW_TYPE & };
%enddef // PRIMITIVE_TYPEMAP
PRIMITIVE_TYPEMAP(long int, long long);
PRIMITIVE_TYPEMAP(unsigned long int, long long);
#undef PRIMITIVE_TYPEMAP
#endif // defined(SWIGWORDSIZE64)
#endif // defined(SWIGJAVA)
```
ref: https://github.com/google/or-tools/blob/stable/cmake/doc/swig.md

PR #649, opened 2016-04-06
Fix for Issue #646 Java with -DSWIGWORD64

johnwason:
This patch will fix the issues with word size #646 on Linux 64-bit that requires the use of -DSWIGWORD64 .  This has only been tested against my software which may not be an exhaustive test. Further testing should be done before accepting this patch.  Note that users may also want to add:

```
#if defined(SWIGWORDSIZE64)
%apply unsigned int { size_t };
%apply const unsigned int & { const size_t & };
#endif
```

after including any STL structures to maintain the same Java size_t type for both architectures. Otherwise `size_t` will be `long` on 32-bit and `BigInteger` on 64-bit.  Note that Java `long` is 64-bit anyway so the loss of precision should be negligible.

### DIFF
--- a/Lib/java/arrays_java.i
+++ b/Lib/java/arrays_java.i
@@ -102,9 +102,13 @@ JAVA_ARRAYS_DECL(short, jshort, Short, Short)         /* short[] */
 JAVA_ARRAYS_DECL(unsigned short, jint, Int, Ushort)   /* unsigned short[] */
 JAVA_ARRAYS_DECL(int, jint, Int, Int)                 /* int[] */
 JAVA_ARRAYS_DECL(unsigned int, jlong, Long, Uint)     /* unsigned int[] */
+#if defined(SWIGWORDSIZE64)
+JAVA_ARRAYS_DECL(long, jlong, Long, Long)        /* long long[] */
+#else
 JAVA_ARRAYS_DECL(long, jint, Int, Long)               /* long[] */
 JAVA_ARRAYS_DECL(unsigned long, jlong, Long, Ulong)   /* unsigned long[] */
 JAVA_ARRAYS_DECL(jlong, jlong, Long, Longlong)        /* long long[] */
+#endif
 JAVA_ARRAYS_DECL(float, jfloat, Float, Float)         /* float[] */
 JAVA_ARRAYS_DECL(double, jdouble, Double, Double)     /* double[] */
 
@@ -126,9 +130,13 @@ JAVA_ARRAYS_IMPL(short, jshort, Short, Short)         /* short[] */
 JAVA_ARRAYS_IMPL(unsigned short, jint, Int, Ushort)   /* unsigned short[] */
 JAVA_ARRAYS_IMPL(int, jint, Int, Int)                 /* int[] */
 JAVA_ARRAYS_IMPL(unsigned int, jlong, Long, Uint)     /* unsigned int[] */
+#if defined(SWIGWORDSIZE64)
+JAVA_ARRAYS_IMPL(long, jlong, Long, Long)        /* long long[] */
+#else
 JAVA_ARRAYS_IMPL(long, jint, Int, Long)               /* long[] */
 JAVA_ARRAYS_IMPL(unsigned long, jlong, Long, Ulong)   /* unsigned long[] */
 JAVA_ARRAYS_IMPL(jlong, jlong, Long, Longlong)        /* long long[] */
+#endif
 JAVA_ARRAYS_IMPL(float, jfloat, Float, Float)         /* float[] */
 JAVA_ARRAYS_IMPL(double, jdouble, Double, Double)     /* double[] */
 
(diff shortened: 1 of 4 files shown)


johnwason:
Is there a plan to merge this pull request and #650? I'd rather not maintain a forked version of SWIG. Without these patches, 64-bit gcc platforms do not work correctly and cause pointer corruption.

kdunee:
@johnwason These changes are really helpful, thanks.
@ojwb Could this please be considered for merging or is it already fixed in 4.0?

ojwb:
@kdunee I only tagged the ticket so people can navigate the open issues more easily.  I don't really know a lot about the Java backend so probably I can't usefully review such a patch (though I'll note that we try to generate portable code which isn't tied to aspects of a particular environment, and if I follow the patch it seems to make the generated code non-portable between environments where `sizeof(long)` is different).

The original reporter also indicated this still needs further testing.

kdunee:
@ojwb Thanks for replying. Actually the patch makes it so the Java code generated on x86-64 Windows and Linux is identical. Without it I'm getting different method signatures all over the place, like in [this diff](https://github.com/SimpleITK/SimpleITK/files/1303577/simpleitk.txt) from generating [SimpleITK](https://github.com/SimpleITK/SimpleITK) wrapping. In particular, some Java types on Linux, where `long` is 64-bit are in fact _smaller_ than on Windows. I believe it's also related to the last paragraph of [this mailing list post](http://swig.10945.n7.nabble.com/uint64-t-with-std-vector-and-java-portability-td13901.html).

It is true though that this needs thorough testing, particularly as I have concerns with proposed translation of `size_t` to `unsigned int`.

johnwason:
@kdunee The use of `unsigned int` for `size_t` was to maintain compatibility between platforms. If `size_t` is allowed to change between 32 and 64 bit integers, the wrappers will become incompatible depending no the platform currently in use. If necessary another switch can be added to allow for size_t to vary if necessary. I typically use `stdint.i` to prevent the variable sized integer problem.

johnwason:
@ojwb What needs to be done to have this pull request accepted? Without this patch and #650 the generators don't produce the correct sized types on Linux x86_64.

ojwb:
I think you'll need to wait for @wsfulton to have time to review.

m-adamson:
I use SWIG to generate Java wrappers for Android apps and this is becoming an issue for that too.  The Google Play Store has announced that all apps will need to be 64-bit by August 2019 so it would be great to have this pull request integrated to allow rebuilding my apps to be compliant.
[Android Police - native app 64-bit support](http://www.androidpolice.com/2017/12/19/play-store-require-new-updated-apps-target-recent-api-levels-distribute-native-apps-64-bit-support/)

johnwason:
Will this pull request be accepted at some point?

m-adamson:
Would it be possible to get this patch integrated into the upcoming SWIG 4.0?
It'd really help with the migration of Android apps to 64-bit ([Android Developer Blog - Get your Apps Ready For 64-Bit](https://android-developers.googleblog.com/2019/01/get-your-apps-ready-for-64-bit.html))
Thanks!

wsfulton:
I've added the swig-4.0 milestone, so will consider it for inclusion.

johnwason:
Please include both pull requests #649 and #650 in the milestone. C# has the same problem.

wsfulton:
Hi @johnwason, I'm looking at this to see if it can be included in the release.

Why are changes like the following:
```diff
+#if defined(SWIGWORDSIZE64)
+%typemap(directorin, descriptor="J") long        "$input = (jlong) $1;"
+#else
 %typemap(directorin, descriptor="I") long             "$input = (jint) $1;"
 %typemap(directorin, descriptor="J") unsigned long    "$input = (jlong) $1;"
 %typemap(directorin, descriptor="J") long long        "$input = (jlong) $1;"
+#endif
```
made so that when SWIGWORDSIZE64 is defined, it completely removes the typemaps for `unsigned long` and `long long` ? It'll break any code using these fundamental types.

johnwason:
@wsfulton unsigned long, long long, and unsigned long long are too large to fit in the Java "long" type and are mapped to BigInteger in the other modified files.

wsfulton:
So for example, where exactly are the directorin `long long` and `const long long&` typemaps that have been removed from java.swg (the file that provides all the default typemaps)?

johnwason:
I may have missed a few typemaps for long long and unsigned long long. I don't use those types as I never go beyond uint64_t in my software.

wsfulton:
Implementing this correctly and consistently across all languages is a fair amount of work, so am deferring it until after swig 4.0, probably until 4.1. Anyone needing this now can override the modified typemaps in a separate file and %include it at the top of the main interface file. This might be how the final implementation should work too.

StefanBattmer:
Just out of curiosity: @wsfulton: With SWIG 4.1.1 being out this didn't make it into SWIG mainline so far, did it? We also have some trouble with 64-bit integer wrapping from C++ to Java and right now work with SWIG 4.0.1 and these patches. This makes migrating to a newer SWIG version quite cumbersome. Is this issue still on the agenda for upcoming SWIG releases?

ojwb:
I see @wsfulton intended to consider this for 4.1 but it wasn't on the milestone so got missed.  Setting milestone for 4.2 - I don't know what the planned timescale for that is so maybe it won't make it, but at least it won't just be forgotten about.

johnwason:
@ojwb also add #650 which is the same fix for C#

ojwb:
I'll leave that for wsfulton to decide - the Java and C# target languages are both outside my area of expertise so it's not really my call whether he wants to prioritise it for a particular release.

The difference with this PR is his comment above implies he intended to prioritise this and it seems to have been overlooked because it wasn't actually on a milestone.

PR #650, opened 2016-04-06
Fixed SWIGWORDSIZE64 for csharp

johnwason:
This is a similar fix to  #646 for C#.  It turns out C# had the same problem as Java with gcc 64-bit. This patch needs to be fully tested.

### DIFF
--- a/Lib/csharp/arrays_csharp.i
+++ b/Lib/csharp/arrays_csharp.i
@@ -94,13 +94,15 @@ CSHARP_ARRAYS(short, short)
 CSHARP_ARRAYS(unsigned short, ushort)
 CSHARP_ARRAYS(int, int)
 CSHARP_ARRAYS(unsigned int, uint)
-// FIXME - on Unix 64 bit, long is 8 bytes but is 4 bytes on Windows 64 bit.
-//         How can this be handled sensibly?
-//         See e.g. http://www.xml.com/ldd/chapter/book/ch10.html
+#if defined(SWIGWORDSIZE64)
+CSHARP_ARRAYS(long, long)
+CSHARP_ARRAYS(unsigned long, ulong)
+#else
 CSHARP_ARRAYS(long, int)
 CSHARP_ARRAYS(unsigned long, uint)
 CSHARP_ARRAYS(long long, long)
 CSHARP_ARRAYS(unsigned long long, ulong)
+#endif
 CSHARP_ARRAYS(float, float)
 CSHARP_ARRAYS(double, double)
 CSHARP_ARRAYS(bool, bool)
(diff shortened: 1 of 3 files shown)


koliyo:
This should be merged to master

This is still (2020) a problem when using `int64_t` on linux, which is defined as 
```c++
# if __WORDSIZE == 64
typedef long int		int64_t;
# else
```

in `stdint.h`, so it becomes `int` in C# bindings.

Ie, 8 bytes in native code library, and 4 bytes in generated C# bindings.

Lib/csharp/arrays_csharp.i
```diff
@@ -94,13 +94,15 @@ CSHARP_ARRAYS(short, short)
...
-//         How can this be handled sensibly?
-//         See e.g. http://www.xml.com/ldd/chapter/book/ch10.html
+#if defined(SWIGWORDSIZE64)
+CSHARP_ARRAYS(long, long)
+CSHARP_ARRAYS(unsigned long, ulong)
+#else
 CSHARP_ARRAYS(long, int)
 CSHARP_ARRAYS(unsigned long, uint)
 CSHARP_ARRAYS(long long, long)
 CSHARP_ARRAYS(unsigned long long, ulong)
```
vadz, line 104: I'm not sure if it's correct to exclude typemaps for `[unsigned] long long` completely when `SWIGWORDSIZE64` is defined. Of course, it has the same size as `long` in this case, but we should still have typemaps for it, shouldn't we?

vadz (commented):
FWIW I agree that this should be merged, but I wonder if we could just define `CSHARP_[U]LONG` once (as either `[u]int` or `[u]long`, depending on `SWIGWORDSIZE64`) and then use them instead of having all these `#ifdef SWIGWORDSIZE64` checks? It seems like this would result in less code duplication -- and would also leave `[unsigned] long long` alone, as I think should be done.

Lib/csharp/arrays_csharp.i
Mizux: +1
note: In 32bits we already have a "map collision" between `int` and `long` (mapped to .Net `int`).
So in 64bits having `long` and `long long` mapped to .Net `long` look good to me.

johnwason: Are we sure long long is 64 bits wide? I was only using stdint.h so I didn't worry about that type.

Mizux: According to my tests on 64bits OS `sizeof(long long) == 8`
ref: https://github.com/google/or-tools/blob/stable/cmake/doc/swig.md

wsfulton:
Superceded by #2379.

PR #651, opened 2016-04-08
[ruby] With %newobject, the wrong error message of an overloaded function is created

tamuratak:
Hi,
When creating a c++ wrapper for ruby, with %newobject, the wrong error message of an overloaded function is created. For example, 

````
%module sample

%newobject f;
void f();
int* f(int);
````
then 

````
$ swig -c++ -ruby sample.i
````
will create sample_wrap.cxx in which

````
  Ruby_Format_OverloadedError( argc, 1, "f.new", 
    "    f.new()\n"
    "    f.new(int)\n");
````
".new" is appended to the function names.

I hope this patch would help your development.
Regards.

### DIFF
--- a/Source/Modules/ruby.cxx
+++ b/Source/Modules/ruby.cxx
@@ -2110,13 +2110,11 @@ public:
     // Generate prototype list, go to first node
     Node *sibl = n;
 
-    String* type = SwigType_str(Getattr(sibl,"type"),NULL);
-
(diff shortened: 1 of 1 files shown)


wsfulton:
The patch introduces a minor memory leak by removing the ```type``` variable. I've fixed this and will merge. Thanks.

PR #654, opened 2016-04-11
Refactor naming formats code to reduce duplication

vadz:
While working on something else I wanted to add a new function to `naming.c` and realized that I had to create a 5th copy of an existing function, so, before doing this in my own branch, I decided to refactor the code to avoid having 4 copies in the first place.

These trivial changes don't (or at least shouldn't) change anything in the program behaviour but they do remove a bunch of duplicated code.


ojwb:
Looks sensible to me (I didn't compare the code before and after in detail though).

wsfulton:
Thanks, looks good to me too.

PR #658, opened 2016-04-14
Fixed clang memset warning in Python

johnwason:
A warning is created by the `memset` function in `traits_as<>::as` in Python when compiling with clang.  Adding a  `void*` cast fixes this warning.


johnwason:
Sorry, this request pulled in some other updates.  I will create a new one with just this change.

PR #659, opened 2016-04-14
Fixed clang memset warning in Python

johnwason:
A warning is created by the `memset` function in `traits_as<>::as` in Python when compiling with clang.  Adding a  `void*` cast fixes this warning.

(This is a replacement for pull request #658 that had too many commits)

### DIFF
--- a/Lib/python/pystdcommon.swg
+++ b/Lib/python/pystdcommon.swg
@@ -139,7 +139,7 @@ namespace swig {
 	  %type_error(swig::type_name<Type>());
 	}
 	if (throw_error) throw std::invalid_argument("bad type");
-	memset(v_def,0,sizeof(Type));
+	memset((void*)v_def,0,sizeof(Type));
 	return *v_def;
       }
     }


vadz:
I wonder what is the exact warning given because I just don't understand how can explicitly casting any pointer to `void*` change anything when this conversion is done implicitly anyhow?

johnwason:
It was warning about overwriting the vtable of a class with virtual members. In this case it isn't an issue since it is zeroing about a block of memory that was just allocated.

johnwason:
It looks like there might also be a memory leak in this function if an exception is throw.  v_ptr is never freed.

ojwb:
There's no leak, it's a `static` variable.

johnwason:
I see.  There may be a risk of causing an error by returning an uninitialized class by returning a zeroed memory region without calling the constructor.  If the calling function tries to call a virtual function it will result in a segfault.  Under what conditions would this function not throw an exception? My code doesn't seem to use traits_as.

ojwb:
The code looks suspect to me too - I think the clang warning probably isn't something we want to just suppress without checking if this code is actually OK.

The template maze seems to only lead to:

```
  template <class Type>
  inline Type as(PyObject *obj, bool te = false) {
    return traits_as<Type, typename traits<Type>::category>::as(obj, te);
  }
```

But I don't see anywhere which actually calls (or generates a call) to that which doesn't explicitly pass `true` for `te` (which means "throw a C++ exception on error" it seems).  So the dodgy code-path never appears to actually be used.  None of this is documented anywhere I can see (not even in `CHANGES`), so it's probably not used by user typemaps.

johnwason:
I don't see any conditions that would make sense returning an uninitialized memory segment.  It would probably make more sense to remove the bool parameter and have this function always throw.  I will close out this pull request since it isn't the correct solution.  Is this function ever used?  I can't find a place that "as" is called.

ojwb:
The only place I could find it used is here:

```
Lib/python/pycontainer.swg:     return swig::as<T>(item, true);
```

For python that is - octave and ruby have similar code (and ruby looks like it might actually use the suspect code path).

johnwason:
It looks like this is designed to mirror the way that the pointer error is handled.  In this case, the Python error string is set and a null pointer is returned.  The zeroed out memory is returned in lieu of a null pointer. Because the Python error is set, I don't think that the memory segment is ever used.  There is however the danger of a copy constructor/destructor being called at some point and crashing the program if a zeroed out region is returned.

PR #661, opened 2016-04-18
Allow failing tests

vadz:
As discussed in [this thread](http://thread.gmane.org/gmane.comp.programming.swig.devel/23813), this can be useful while working on a new module.

The second commit is useful to monitor the progress of the work.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -670,6 +670,11 @@ MULTI_CPP_TEST_CASES += \
 wallkw.cpptest: SWIGOPT += -Wallkw
 preproc_include.ctest: SWIGOPT += -includeall
 
+# Allow modules to define temporarily failing tests.
+C_TEST_CASES := $(filter-out $(FAILING_C_TESTS),$(C_TEST_CASES))
+CPP_TEST_CASES := $(filter-out $(FAILING_CPP_TESTS),$(CPP_TEST_CASES))
+MULTI_CPP_TEST_CASES := $(filter-out $(FAILING_MULTI_CPP_TESTS),$(MULTI_CPP_TEST_CASES))
+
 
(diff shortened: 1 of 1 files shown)


Examples/test-suite/common.mk
```diff
@@ -700,6 +705,7 @@ all: $(NOT_BROKEN_TEST_CASES) $(BROKEN_TEST_CASES)
 broken: $(BROKEN_TEST_CASES)
 
 check: $(NOT_BROKEN_TEST_CASES)
+	@echo $(shell echo $^|wc -w) tests passed
```
ojwb: You can easily avoid the overhead of running external processes here (and avoid relying on `wc` being available):

```
@echo $(words $^) tests passed
```

Personally what I've done for this case is to run `make check-foo-test-suite > check-foo.new 2>&1` each time, and keep the output for the currently committed version in `check-foo.current`.  Then to see what difference a change I'm working on makes, I can `diff check-foo.current check-foo.new`.  The advantage of this approach is you get to see when tests move from failing to passing due to a change, while the above change seems to just prevent a blacklist of tests from running at all.

I guess the advantage of this is that automated builds can run the testsuite and understand if the outcome is a regression.

vadz: I somehow completely forgot about `$(words)` :-( The worst is that I was sure that there was a function to do this in GNU make and spent some time looking for it, but failed to find it. I'll update this, of course, thanks!

As for redirecting make output, I also do this, of course, but I prefer to just be able do `make check-xxx-test-suite` and check there is no error at the end, this is much less error-prone.

wsfulton:
Let's merge but also show the language in the new message for when using make -j:
```
	echo $(words $^) $(LANGUAGE) tests passed
```

PR #663, opened 2016-04-19
R vector refs rebase

richardbeare:
The last patch for std::vectors left out some typemaps for references, resulting in some incorrect typechecks in generated code. I've added them in this patch.

I've also removed a redundant trace statement.


wsfulton:
You shouldn't need both const and non-const of the same type. Please change:
```patch
+%typemap("rtype") std::vector<float>, std::vector<float> const, std::vector<float> const& = std::vector<double>;
```
to
+%typemap("rtype") std::vector<float>, std::vector<float> const& = std::vector<double>;
```
etc and test again.

richardbeare:
Thanks - I've removed duplications with const, and made the supported types consistent.

PR #665, opened 2016-04-20
Add check-failing target to check that failing tests still do fail

vadz:
This is useful to remove the tests which pass after the latest fixes from the
list of the failing tests.

---
I realized that I was running the shell one-liner doing the equivalent of this all the time, so I decided to add support for checking whether the failing tests still do fail to the makefile itself, this is quite convenient, because quite often several tests start to pass after fixing just one bug (which, IMHO, is a problem with the test suite itself, the tests should be more orthogonal, but I'm not about to propose changing this...).

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -713,6 +713,13 @@ check-cpp: $(CPP_TEST_CASES:=.cpptest)
 
 check-cpp11: $(CPP11_TEST_CASES:=.cpptest)
 
+check-failing-test = \
+	$(MAKE) -s $1.$2 >/dev/null 2>/dev/null && echo "Failing test $t passed."
+
+check-failing:
+	+-$(foreach t,$(FAILING_C_TESTS),$(call check-failing-test,$t,ctest);)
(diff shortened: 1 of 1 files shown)


Examples/test-suite/common.mk
```diff
@@ -713,6 +713,13 @@ check-cpp: $(CPP_TEST_CASES:=.cpptest)
 
 check-cpp11: $(CPP11_TEST_CASES:=.cpptest)
 
+check-failing-test = \
+	$(MAKE) -s $1.$2 >/dev/null 2>/dev/null && echo "Failing test $t passed."
```
ojwb: Shouldn't that be `"Failing test $1 passed."`?

vadz: Oops, you're right, it would be more hygienic to use `$1` here. `$t` happens to work too but it's a leftover before I extracted this command into its own variable.

I'll update it if there are no objections.

ojwb:
Updating it would be better - then it's robust to being used in other places which might not happen to have set `$t` to be the same as `$1`.

PR #668, opened 2016-04-27
Remove gcc warnings when using -Wconversion.  

dnwillia-ansys:
Right side of expressions are signed char.  Could not find a way to suppress the warnings using compiler pragma's.


dnwillia-ansys:
The travis-ci build failure does not seem to be due to these changes.
https://travis-ci.org/swig/swig/jobs/126127902

wsfulton:
I've re-run the failed travis-ci build which has now passed. Thanks for the patch.

dnwillia-ansys:
No problem.

dnwillia-ansys:
Glad this made it back.  @brgirgis

PR #670, opened 2016-04-28
Take long, not int, in std::vector Java wrappers methods

vadz:
The type of the argument of get() and set() was inconsistent with the return
type of size() (and, incidentally, that of C++ at() method) and writing

```
vec.get(vec.size() - 1)
```

unexpectedly didn't compile because of a type mismatch.

Fix this by taking size_t in get()/set() too.

---

As the changelog says, this is an incompatible change, but it's really annoying that you need a cast to compile the above currently.

The same should arguably be done in `java/std_array.i` too...

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -5,6 +5,12 @@ See the RELEASENOTES file for a summary of changes in each release.
 Version 3.0.9 (in progress)
 ===========================
 
+2016-04-28: vadz
+	    [Java] Take long, not int, in std::vector wrappers get() and set() methods.
+	    This allows to write vec.get(vec.size()-1) without casts.
+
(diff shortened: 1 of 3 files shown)


wsfulton:
Yes, this needs fixing, but not in a patch release, so I've marked it for swig-3.1. There is a similar patch at https://sourceforge.net/p/swig/patches/196/. 

I've always wanted to improve the std_vector wrappers and so if we are going to change the api, we should do it just once in a new version. The main thing I'd like is to be able to iterate over the elements. There is also this patch to consider: https://sourceforge.net/p/swig/patches/278/ and

wsfulton:
Superceded by #842

PR #674, opened 2016-05-07
Java std::list as an AbstractCollection 

fflexo:
Most of the wrapping of containers in Java doesn't implement any interfaces that make them easy or intuitive for a Java developer to use. Fairly frequently people ask about making them extend java.util.List. That's clearly wrong since each item would have to be stored twice. But AbstractSequentialList does make a lot of sense.

To support both primitive and non-primitive types in Java generics we need to pretend we're autoboxing the primitives somewhere at the point of crossing the language boundaries just like Java would. To support that there's a second file added that includes a new typemap, which makes this automatic and reusable.

So there are two new files to this:
- added std_list.i implemenatation that extends Java's AbstractSequentialList
- added autobox.i that provides supporting typemaps for generics in containers

(This work came out of http://stackoverflow.com/q/36552136/168175)


wsfulton:
Instead of using a generic
```
 +%javamethodmodifiers "private";
```
can you target each method explicitly, eg:
```
 +%javamethodmodifiers  "private" std::list::begin;
```
This looks like jdk-1.5 is required as a minimum and the rest of the generated code works with jdk-1.2 and later This is just an observation but I think it is okay.

I need a test case in the test-suite to accept this patch. It'll need to call all the methods for say a struct, enum and a primitive type. Should be quite easy to do, there is the li_std_list.i testcase that could be used as a start.

fflexo:
Hopefully on the right lines here:
- I've edited li_std_list.i in Examples/test-suite to expose a bit more functionality, this would impact any other languages using that test case (oddly it's commented out in common.mk though)
- li_std_list is now in Examples/test-suite/java/Makefile.in so it gets run on make check etc.
- This process threw up a few bugs that I've fixed along the way

If that looks like a sensible approach (I've never touched the SWIG test cases before) I'll add a bit more depth to the li_std_list_runme.java part and then hopefully it'll be good to go.

wsfulton:
Yes, this all looks it is moving in the right direction.

I think li_std_list.i is commented out purely because it does not have many languages that have provided std::list wrappers or is an oversight.

Your return type for std::list::size does not match that in the C++ standard. There'll be corner case compilation problems unless you fix this. This is an unfortunate mismatch between Java and C++ (unsigned vs signed). I think we should ```size_t``` in the %extend methods too and use a jstype typemap explicitly for std::list for using ```int``` on the Java side. Also, it'll be worth looking at #670.

Your use of the autobox typemap is novel... it is a made up typemap not used other than for a lookup table in the template. Kind of neat, but maybe confusing that it isn't a conventional typemap, so probably worth documenting its intended usage in autobox.i.

The non-const CTYPE& should be dropped from the autobox typemap as the SWIG conventions are only to wrap pass by value and pass by const ref primitive types as a primitive type on the target language side. 
```
%typemap(autobox) CTYPE, const CTYPE&, CTYPE& "JTYPE"
```

I see that there are lots of ```SWIGTYPE_p_std__listT_XXXXX__iterator``` type wrapper classes. I'd like to see these wrapped as a proper proxy, maybe even as an inner Java class or some equivalent in the Java world. The existence of type wrapper classes are usually an indicator of a badly wrapped interface and I make my build systems fail if any of them appear. Hence, the desire for a proper iterator proxy class of some sorts. 

If you aren't familiar with Lib/csharp/std_vector.i you may want to have a look at the techniques in there, although I think you've used a lot of them already. I envisage a new std_vector.i improvements using similar ideas from here and clearly they can also be used in std_list.i for Java.

Can you please rebase this on to master for your next pull request so that the Travis and Appveyor tests run (they won't run in its current conflict state).

fflexo:
To elaborate on some of my thoughts around this:

I'd planned to submit std::vector in a similar way after this. It reuses the autobox typemap I added for exactly the same reasons, but because of the fact that there's an existing std_vector.i for Java it's slightly harder to meet the requirements of AbstractList without breaking existing code bases as well. (Mostly around that painful long/size_type/jint problem).

(I've also got std::unique_ptr, std::weak_ptr and std::function code that I think can work with a few tweaks and some test cases as library functions for other languages besides just Java. My plan was to check these off one-by-one with the simpler parts like std::list for Java only first).

The rest of the comments are fairly straightforward for me to adapt to, so I'll pick them up over the next week or so.

vadz:
I've just found this *after* submitting #842 addressing the same problem and in a very (conceptually) similar way for `std::vector<>`. I even define `jboxtype` typemap very similar to your `autobox` one. FWIW I think it's better to have this leading `j` to make it clear that it's Java-specific, but this is a minor detail, of course...

wsfulton:
Thanks @fflex, I've merged all this in now and made some enhancements, see d04eb8874251f029a50699736dc2889f777c526e..df899cfef1dc2efdb6c23f79586e2059ab0bf97e.

wsfulton:
For the record, these were alternative proposed implementations for std::list but were either out of date or not as complete as this pull request:
https://sourceforge.net/p/swig/patches/134/
https://sourceforge.net/p/swig/patches/144/
https://sourceforge.net/p/swig/patches/259/

Issue #676, opened 2016-05-15
[Go] Whether capitalize C++ class name or not leads to different go code


dontpanic92:
Hi, 

I found that the class name in C++ code will affect generated go code when it is a parameter type of overloaded functions. For example,

```
%inline %{

class ClassA {};

class classB {};

int Test1(ClassA* a);
int Test1(int i);

int Test2(classB* a);
int Test2(int i);

%}
```

`ClassA` and `classB` are only different in the name. The generated dispatching functions are,

```
func Test1(a ...interface{}) int {
    argc := len(a)
    if argc == 1 {
        if _, ok := a[0].(ClassA); !ok { // Note here
            goto check_1
        }
        return Test1__SWIG_0(a[0].(ClassA))
    }
check_1:
    if argc == 1 {
        return Test1__SWIG_1(a[0].(int))
    }
    panic("No match for overloaded function call")
}

func Test2(a ...interface{}) int {
    argc := len(a)
    if argc == 1 {
        if _, ok := a[0].(SwigcptrClassB); !ok { // And here
            goto check_1
        }
        return Test2__SWIG_0(a[0].(ClassB))
    }
check_1:
    if argc == 1 {
        return Test2__SWIG_1(a[0].(int))
    }
    panic("No match for overloaded function call")
}

```

As a result any derived classes of `classB` cannot passed into `Test2`.

dontpanic92:
I have examined swig source code and found the problem:.In [`goOverloadType`](https://github.com/swig/swig/blob/master/Source/Modules/go.cxx#L6513), swig will check whether the type is defined:

```
String *goOverloadType(Node *n, SwigType *type) {
    SwigType *ty = SwigType_typedef_resolve_all(type);

    //..... irrelavant code omitted

    if (Getattr(undefined_types, ty) && !Getattr(defined_types, ty)) {
      return goWrapperType(n, type, true);
    }

    return goType(n, type);
  }
```
For `classB`, `ty` here is `classB`, the class name in C++. And in [`classHandler`](https://github.com/swig/swig/blob/master/Source/Modules/go.cxx#L2872):

```
// irrelavant code is omitted
virtual int classHandler(Node *n) {
    String *name = Getattr(n, "sym:name");
    String *go_name = exportedName(name);
    Setattr(defined_types, go_name, go_name);
}
```

What added into `defind_types` is `ClassB`, which is the type in Go rather than in C++.

So I think code in `goOverloadType` might be
```
[...]

PR #677, opened 2016-05-15
[Go] Fix overloaded functions with noncapitalized class as parameter type

dontpanic92:
#676

### DIFF
--- a/Source/Modules/go.cxx
+++ b/Source/Modules/go.cxx
@@ -6525,8 +6525,10 @@ private:
 	break;
       }
     }
-
-    if (Getattr(undefined_types, ty) && !Getattr(defined_types, ty)) {
+    
+    String* go_type = goType(n, ty);
+    
+    if (Getattr(undefined_types, go_type) && !Getattr(defined_types, go_type)) {
       return goWrapperType(n, type, true);
     }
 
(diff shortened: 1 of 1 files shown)


ojwb:
@ianlancetaylor Could you review this when you get a chance?  It looks like a plausible fix to me, but I'm not very familiar with Go or the SWIG Go backend.

Source/Modules/go.cxx
```diff
@@ -6525,8 +6525,10 @@ class GO:public Language {
 	break;
       }
     }
-
-    if (Getattr(undefined_types, ty) && !Getattr(defined_types, ty)) {
+    
+    String* go_type = goType(n, ty);
+    
+    if (Getattr(undefined_types, go_type) && !Getattr(defined_types, go_type)) {
```
ianlancetaylor: As far as I can see, the index into `undefined_types` should be `ty` and the index into `defined_types` should be `go_type`.

Also you should `Delete(go_type)` somewhere.

ianlancetaylor:
@ojwb Thanks for the pointer, I missed this.

ojwb:
Sorry, it seems your final update got missed.

I've added a CHANGES,current entry and a testcase based on the reproducer in #676 and merged this.

PR #681, opened 2016-05-18
fix for nested ignored types

wkalinin:


wsfulton:
Looks okay to me. Could you pop in the testcase with the patch please?

wkalinin:
added test

wsfulton:
@wkalinin did you see the test failed?

wkalinin:
Sorry, took it verbatim from Vadim's patch and forgot to look at the results.

wkalinin:
The test still fails under 'go', but I have no idea why. PHP and Python Travis fails seem to be unrelated to this case.

wkalinin:
Found the cause of 'go' fail -  the name of the module in the test should be the same as the file name.
Now the only platform failing is Octave 3.8, but it fails not on this test but before the tests start.

wsfulton:
Travis fails sometimes with network errors like this. I've restarted this Octave test.

PR #688, opened 2016-05-25
Fix infinite recursion in Python tests makefile

vadz:
$(srcdir)/%_runme.py could match the rule for ./%_runme.py in which it
appeared as a dependency when srcdir started with a period.

It is not clear why this didn't happen before and doesn't happen for all the
files even now, yet the problem does happen with both GNU make 3.81 and 4.0
and this workaround fixes it in both cases.

---

Somehow, running `make check-python-test-suite` with the latest master now results in errors due to an apparent recursion in the makefile. The exact error message is best seen when running make directly in the `Examples/test-suite/python` directory, see the block at the end (I put it there because it is so long).

This is crazy because it didn't happen before and yet I can't find any recent commit introducing the problem and now I see it even when trying very old commits which definitely worked for me before. It clearly has something to do with building outside of the source directory and using relative paths to it, i.e. I run `../../../src/3rdparty/swig/configure` (which, I think, it still required for C#/Java tests at least) as I do, but I don't understand how can it happen nor why didn't it happen before. But I already spent almost an hour on this and I don't think it makes sense to continue, hopefully the workaround can be applied even if things are not totally clear as it should be harmless and it does fix the problem for me under two different machines.

Here is the error message, which can also be seen by directly making `./argcargvtest_runme.py`. On this machine (Debian Jessie, GNU make 4.0) it seems to happen for all the files. On the other machine (Debian Wheezy, GNU make 3.81) it only happens for some of them which is even less understandable :-(

```
% make argcargvtest.cpptest
if [ -f ../../../../../../src/3rdparty/swig/Examples/test-suite/python/argcargvtest_runme.py ]; then make ./argcargvtest_runme.py; fi
make[1]: Entering directory '/home/zeitlin/build/swig/master/Examples/test-suite/python'
make[1]: stat: ../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/argcargvtest_runme.py: File name too long
make[1]: *** No rule to make target '../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/argcargvtest_runme.py', needed by '../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/../../../../../../src/3rdparty/swig/Examples/test-suite/python/argcargvtest_runme.py'.  Stop.
make[1]: Leaving directory '/home/zeitlin/build/swig/master/Examples/test-suite/python'
Makefile:103: recipe for target 'argcargvtest.cpptest' failed
make: *** [argcargvtest.cpptest] Error 2
```

### DIFF
--- a/Examples/test-suite/python/Makefile.in
+++ b/Examples/test-suite/python/Makefile.in
@@ -155,7 +155,7 @@ convert_testcase = \
 	  $(MAKE) $(SCRIPTDIR)/$(py_runme); \
 	fi
 
-$(SCRIPTDIR)/$(SCRIPTPREFIX)%$(SCRIPTSUFFIX): $(srcdir)/$(SCRIPTPREFIX)%$(PY2SCRIPTSUFFIX)
+$(SCRIPTDIR)/$(SCRIPTPREFIX)%$(SCRIPTSUFFIX): $(abspath $(srcdir)/$(SCRIPTPREFIX)%$(PY2SCRIPTSUFFIX))
 	test x$< = x$@ || cp $< $@ || exit 1
 	test x$(PY3) = x || $(PY2TO3) -w $@ >/dev/null 2>&1 || exit 1
 


ojwb:
autoconf provides `abs_srcdir`, which is probably cleaner (it would need `abs_srcdir = @abs_srcdir@` adding to this `Makefile.in` too).

But either approach will defeat the attempt to avoid copying a file to itself when `srcdir` is relative (and we're testing Python3) won't it?

```
test x$< = x$@ || cp $< $@ || exit 1
```

Though that code seems like it's already rather brittle (and indeed this whole mechanism does...)

ojwb:
Ah no, I see - in the Python 3 case the suffix is different, so `$<` and `$@` can't be equal.  And in the Python2 case, the equality case is handled by the make conditional wrapping this, so again `$<` and `$@` can't be equal.  The `|| exit 1` seems to have no obvious purpose (other than to change any non-zero exit status to be 1, which doesn't seem useful), so it looks like that first command could just be `cp $< $@`!

vadz:
Yes, I don't understand the `|| exit 1` part neither. But my goal was just to fix the build, and AFAIR I tested both Python 2 and 3, so at least my change doesn't break anything and IMHO it would be nice to start by applying it.

OTOH I definitely wouldn't object to simplifying this, it was 2.5 months ago but I still remember the fun I had tracking down the logic of what was happening here...

ojwb:
I've no objection to the patch (though using autoconf's `abs_srcdir` seems better to me) - I already realised my initial worry was unfounded (as I noted in the follow-up comment).

vadz:
@wsfulton Can I apply this myself or should I wait for you to have a look this? TIA!

vadz:
@wsfulton It's really, really annoying to have to apply this trivial fix every time when running Python test suite. Could you please consider applying it or let me know if I could do it?

wsfulton:
If it didn't use absolute paths, I'd accept it as absolute paths are mostly avoided in the build system for various reasons and I have no intention of starting now. There is one rogue language (Javascript) which does though and needs bringing into line if I could find the time. Can you find a relative path solution? If it only happens for some testcases maybe it is a Make bug, perhaps try run valgrind over make? If you can't find another solution, I'll try recreate it, but I've certainly never seen this.

ojwb:
@vadz How are you configuring your tree?  (You can see the command line used near the top of `config.log`)

I wondered if the issue was doing an out of tree build with `./` on the start of a relative path to configure, but I tried that and it worked fine for me.

Looking at your output, it looks to me like you're doing something like:

```
../../../src/3rdparty/swig/configure
```

But I also tried that myself and it worked fine too.  Tested on Debian unstable.

Also, are any symlinks involved?

vadz:
I configure it using relative paths, i.e. I run `../../mirrors/swig/configure` in `~/build/swig`.

The only unusual thing I see is that my `~/build` directory is actually a mount point, but I have no idea how could it break anything.

In any case, using `$(abspath)` here really doesn't seem like a big deal, so could we just do it? What possible negative consequences could it have?

wsfulton:
As mentioned, I would like a relative path solution. I **really** dislike the verbosity of absolute paths and won't accept absolute paths, certainly not until the problem is reproducible and fully understood. Sorry. Have you tried to workaround it or debug it?

ojwb:
@vadz As a workaround, can you just make the path you use to `configure` absolute?  If you're scripting this, then you could probably just use:

```
"`realpath ../../mirrors/swig/configure`"
```

vadz:
@wsfulton I spent *tons* of time trying to fix it, but completely unsuccessfully and I just don't see any way around it. Even if it is a bug in make, I don't have much hope of fixing it and it's still not a practical solution.

I must say I'm completely stumped by your dislike of absolute paths in this context. Using the absolute path doesn't change anything at all, what verbosity are you speaking about? It's just the way the path appears in the makefile, what possible harm could this change ever do? Have you actually looked at the proposed change? What concrete drawback of making it do you see?

@ojwb I'll have to test, but I think using absolute path to configure still breaks C# unit test suite. Of course, I can have 2 build directories and run C# test suite in one and Python in the other but it's such a PITA and all for what? Avoiding a single extra `$(abspath)`... Compared to tons of other ugly and questionable constructs in SWIG makefiles it's a bit too much irony for me to stomach.

ojwb:
> using absolute path to configure still breaks C# unit test suite

So it does, due to stuff like:

```
top_srcdir   = ../@top_srcdir@
top_builddir = ../@top_builddir@
```

and

```
SRCDIR       = ../$(srcdir)/
```

and so on.

Perhaps if @wsfulton isn't willing to accept this patch he should fix the csharp test suite to work when the user specifies an absolute path to configure (which is something that really ought to work).  Then at least there's an easy way for you (and anyone else affected) to work around whatever the underlying problem is here...

> Compared to tons of other ugly and questionable constructs in SWIG makefiles it's a bit too much irony for me to stomach.

I don't really buy this as an argument - if applied consistently then everything just gets worse and worse, and any attempt at clean-up that doesn't achieve perfection will be largely futile.

wsfulton:
I fixed the recursion problem by using terminal rules / double colon rules. So now the pattern rule is exactly the same but with a double colon:
```make
$(SCRIPTDIR)/$(SCRIPTPREFIX)%$(SCRIPTSUFFIX):: $(srcdir)/$(SCRIPTPREFIX)%$(PY2SCRIPTSUFFIX)
```
It prevents the right hand side from looking for further rules as the prerequisite actually exists.  The 2nd situation in the Stackoverflow question [What are double-colon rules in a Makefile for?](https://stackoverflow.com/a/33466599/1492764) is useful to understand. So are these sections in the Make manual: [Double-Colon Rules](https://www.gnu.org/software/make/manual/make.html#Double_002dColon), [Pattern Rule Examples](https://www.gnu.org/software/make/manual/make.html#Pattern-Examples).

The Makefile is a bit different to when this patch was first created as I just fixed an out of source test not running, see 4930fab5a068ecc8a48ce4b6dc73828dea207377.

Issue #691, opened 2016-05-31
[Python] cannot import swig module in package using swig >= 3.0.9


m7thon:
I have a SWIG python module named `module` that lives in a python package `pkg`. When swigging with swig 3.0.9, the module fails to load properly (using python 3.4).

Here is a minimal example:

```
mkdir pkg
echo "%module module" > module.i
echo "from . import module" > pkg/__init__.py
swig -python -builtin -outdir pkg module.i
gcc -shared -o pkg/_module.so module_wrap.c -I/opt/local/Library/Frameworks/Python.framework/Versions/3.4/include/python3.4m -L/opt/local/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/config-3.4m -lpython3.4m
python -c 'import pkg'
```
gives:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/mthon/dev/swigtests/pkg/__init__.py", line 1, in <module>
    from . import module
  File "/Users/mthon/dev/swigtests/pkg/module.py", line 42, in <module>
    from _module import *
ImportError: No module named '_module'
```

Without the `-builtin` option the above example seems to work, <strike>but `_module` nevertheless fails to import.</strike>

EDIT: Actually, this may indeed only be a problem when using `-builtin`, but I am unsure. mromberg mentions [here](https://github.com/swig/swig/pull/612#issuecomment-205490592) that he anticipated this failing and wants to use this code snippet
```
if __name_.rpartition('.')[0] != '':
    from ._foo import *
else:
    from _foo import *
```
This is essentially the code that fails, however, in my above example the wrapper code that is generated lacks a `.`:
```
if __name__.rpartition('.')[0] != '':
    from _module import *
else:
    from _module import *
```
Adding that '.' manually seems to do the trick. Don't know why that didn't make it into the pull request.

wsfulton:
@mromberg, can you look at this please? Looks like this regression is due to #612.

mromberg:
I have a branch in my repo which now has four new tests:

https://github.com/mromberg/swig/tree/impfix

  The new tests are Examples:
https://github.com/mromberg/swig/tree/impfix/Examples/python/import_packages/split_wrapper

  Right now two of the tests fail.  These are the cases where the swig generated module are split and 1/2 is in a python package and the other half is global.  I think these two failing tests cover the known regressions in swig with the implicit namespace package support.  I will submit a pull request on my branch once I've updated swig to make all tests pass.

[...]

PR #693, opened 2016-06-01
[Python] fix issue 691 (relative C extension module import when using `-builtin`)

m7thon:
It turns out #691 has a simple fix, so I wrote a PR for it.

@mromberg, please have a quick look.

The point is that when using `-builtin`, we want to import `*` form the C extension module `_module`. This should be done using a relative import if the C extension is a submodule of a package. For this, the syntax

    from ._module import *

must be used for Python 3, and may be used for Python >= 2.7. Otherwise,

    from _module import *

may be used.

Notes:

* The syntax `from ._module import *` was used only when invoking swig with the `-py3` option. This is not correct, since even without this option, the wrapper code should be compatible under python 3. The `-py3` option may only break backwards compatibility.

* The old code to include C extension modules, which is currently used for python < 2.7, actually imports the extension module as a global module (well, sometimes it does, sometimes it doesn't...). In either case, the syntax

        from _module import *

    will work, and therefore, for this case this syntax should be used.

Tl;dr
---
The case distinction between python 2 and python 3 syntax for relative imports should be handled in the python code, not in the C++ code relying on the `-py3` swig option.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -859,21 +859,19 @@ public:
       Printv(f_shadow, "else:\n", NULL);
       Printf(f_shadow, tab4 "import %s\n", module);
 
+      if (builtin) {
+        /* The C extension module may have been imported as a package submodule above.
+         * In that case, we need to use a relative import here, too. */
+        Printv(f_shadow, "if __name__.rpartition('.')[0] != '' and version_info >= (2, 7, 0):\n", NULL);
+        Printf(f_shadow, tab4 "from .%s import *\n", module);
+        Printv(f_shadow, "else:\n", NULL);
+        Printf(f_shadow, tab4 "from %s import *\n", module);
+      }
+
       /* Delete the version_info symbol since we don't use it elsewhere in the
        * module. */
       Printv(f_shadow, "del version_info\n", NULL);
 
-      if (builtin) {
-        /*
-         * Python3 removes relative imports.  So 'from _foo import *'
-         * will only work for non-package modules.
-         */
-        Printv(f_shadow, "if __name__.rpartition('.')[0] != '':\n", NULL);
-        Printf(f_shadow, tab4 "from %s%s import *\n", (py3 ? "." : ""),
-          module);
-        Printv(f_shadow, "else:\n", NULL);
-        Printf(f_shadow, tab4 "from %s import *\n", module);
-      }
       if (modern || !classic) {
 	Printv(f_shadow, "try:\n", tab4, "_swig_property = property\n", "except NameError:\n", tab4, "pass  # Python < 2.2 doesn't have 'property'.\n\n", NULL);
       }


mromberg:
>>>>> " " == Michael Thon <notifications@github.com> writes:

     > It turns out #691 has a simple fix, so I wrote a PR for it.
     > @mromberg, please have a quick look.

     > The point is that when using `-builtin`, we want to import `*`
     > form the C extension module `_module`. This should be done
     > using a relative import if the C extension is a submodule of a
     > package. For this, the syntax

     >     from ._module import *

     > must be used for Python 3, and may be used for Python >=
     > 2.7. Otherwise,

     >     from _module import *

     > may be used.

  The '.' will show up if the -py3 switch is given to swig.   Right
now this is the wrapper code when builtin is used:

Printv(f_shadow, "if __name__.rpartition('.')[0] != '':\n", NULL);
Printf(f_shadow, tab4 "from %s%s import *\n", (py3 ? "." : ""),
    module);
Printv(f_shadow, "else:\n", NULL);
Printf(f_shadow, tab4 "from %s import *\n", module);

  So, the '.' is added if the -py3 switch is supplied to swig. So, for
modules that are in a package and we don't know if it is python2 or
python3 the syntax is:

from _foo import *

  I'll admit this is broken for python3.  With the -py3 switch it
changes to:

from ._foo import *

  Which works for python2.7+  but does not work for python 2.6 and
earlier (where I think the '.' then becomes a problem again).  The
older swig code "worked" because _foo.so was really located in the
global namespace.  It was not in the same package as foo.py.

  But now foo.py will correctly load _foo.so from the same package
foo.py is in (it is the wrapper code above the builtin).  But the
wrapper code for loading all the members of _foo is problematic.

  If swig needs to support python2 (I'm not sure how far back) up to
python3 with just the vanilla -python switch then the builtin section
of wrapper code is gonna probably need an import_helper.  It will have
to determine at run-time exactly what kind of python it is dealing
with and if the foo.py module is in a package or not.

Mike

m7thon:
I think you misunderstand what the `-py3` option is for. This is to enable new features that are not backward compatible. Swigging without `-py3` *must* be compatible with python 3 even *without* the `-py3` switch. This is important, for instance, when shipping pre-generated wrapper code as part of a C++ extension that needs to be compatible with python >= 2.

>  I'll admit this is broken for python3.  With the -py3 switch it changes to:
>
>    from ._foo import *
>
> Which works for python2.7+  but does not work for python 2.6 and earlier.

Exactly. So if python >= 2.7 (and we are in a package), use the new relative import syntax, otherwise the old one. But this case distinction must be done in python code, not in C++ relying on `-py3`. It's doable with a simple `if version_info >= (2, 7, 0)`.

wsfulton:
A note on the options and versions of Python...

```-py3``` generates code that requires python 3 and higher. When used with ```-builtin```, it happens to mostly work with python 2.7 because not much of the generated code is python code where the syntax can be quite different. 

> Swigging without -py3 must be compatible with python 3 even without the -py3 switch. 

@m7thon, this isn't correct, you can't expect code to work in Python 3 unless you use ```-py3```. It doesn't work, so I don't understand how you think this is the case.

From the docs http://swig.org/Doc3.0/Python.html#Python_builtin_limitations, ```-builtin``` is fully supported from Python 2.5 and later. Without ```-bultin```, we test for Python 2.4, but currently 2.2 and later is expected to work. swig-3.1 will increase the minimum version supported but not swig-3.0.x. 

Unless ```-py3``` is provided, the code must work for all supported python 2 versions. I agree the relative import syntax ```from ._module import *``` should be generated for the versions of python that support it (2.7 and later). One other thought, is there a need for relative import syntax to work if ```from __future__ import absolute_import``` is added in 2.5 or 2.6 with:

%pythonbegin %{
from __future__ import absolute_import
%}

wsfulton:
Can we have a test added to the patch that shows the problem is fixed and thus prevent future regressions? Probably the best place is under Examples/python/import_packages as the test-suite doesn't allow much flexibility in working with packages and this is where @mromberg has all the new relative import tests.

wsfulton:
Does this patch fix the subversion problem too, mentioned on the swig-devel mailing list? It isn't clear to me.

mromberg:
>>>>> " " == William S Fulton <notifications@github.com> writes:

     > Can we have a test added to the patch that shows the problem is
     > fixed and thus prevent future regressions? Probably the best
     > place is under Examples/python/import_packages as the
     > test-suite doesn't allow much flexibility in working with
     > packages and this is where @mromberg has all the new relative
     > import tests.

  I was going to spend tomorrow creating tests for what I think are
the two regressions in python imports swig generates with the
namespace package imports.  Unless someone else wants to or already
has.  I think the two cases are:

  1)  If a _foo.so module is not found in the package foo.py is in (or
      even in the package swig was told it was suppose to be in, then
      try and load it from the global namespace.

  2)  when -builtin is used the section of wrapper code needs to work

if __name__.rpartition('.')[0] != '':
    # if foo.py is in a package based on __name__
    if version_info >= (2, 7, 0):
        try:
            from ._foo import *  # python3 compatible import
        except ImportError:
            from _foo import *   # fallback to global namespace
    else
        from _foo import *       # fallback to no package at all
else:
    from _foo import *           # foo.py was not in a package

mromberg:
The SVN bindings are broken for the tests in the SVN tree only.  And I don't think this patch addresses that issue.  As mentioned above there seem to be two issues.

  The first is fixed by this and was a misunderstanding on my part of the purpose of -py3.   Python 3 needs the import to be "from ._module import *".

  The second issue which affects the SVN binding tests only (apparently the real world install works fine) is that swig used to fall back to importing all _foo.so modules from the global namespace.  Even if the shadow file (foo.py) was in a package.   The SVN tests for the python bindings are using this.  So, to be backwards compatable, every import of a _foo.so needs to fall back to trying the import from the global namespace.  

  For example if 'from ._foo import *' fails when using python3 the wrapper code should still try 'from _foo import *'.   It is sorta wrong (or totally wrong) but it is what swig used to do.  The SVN test cases are built around it.

  I think this needs to be done as part of this patch as well (code is supplied above).

m7thon:
@wsfulton:
> you can't expect code to work in Python 3 unless you use -py3. It doesn't work, so I don't understand how you think this is the case.

Well, I got this from the documentation under [Python 3 support](http://www.swig.org/Doc3.0/SWIGDocumentation.html#Python_python3support):

> SWIG is able to support Python 3.0. The wrapper code generated by SWIG can be compiled with both Python 2.x or 3.0. Further more, by passing the `-py3` command line option to SWIG, wrapper code with some Python 3 specific features can be generated (see below subsections for details of these features). The `-py3` option also disables some incompatible features for Python 3...[list of specific features]

I thought that as long as one is not using one of the incompatible features, the code generated by swig should be fine under python 2 and 3. If `-py3` is really required for python 3 compatibility, then this should be made very clear in the documentation. This will cause some problems, though:

1.) Currently, I thought one can include pre-generate wrapper code in the distribution of a C python extension. This then only needs to be compiled and installed locally. I haven't checked recently, but I am convinced there are many projects doing just this. So you are saying that this is not possible, and one needs to generate two versions of wrapper code, one for python 2 and one for python 3?

2.) Python setuptools support for swig (which is limited), does not pass the `-py3` flag to swig when building an extension for python 3. So you are saying that this is a bug? I do understand that it is trivial to insert the `-py3` flag oneself in the `setup.py` script.

---

@mromberg:

Ok, can you please fix this and add a test case? Sorry for "budging in" with this PR. I'll leave you to it. Thanks.

ojwb:
FWIW, I'm also surprised to read in the docs that code generated without `-py3` the code is meant to work with Python3 - this also contrary to how I understood this to work.  That whole paragraph in the docs is unchanged from the initial merge Haoyu Bai's branch in 3d8ddfc44, so it does indeed seem this was Haoyu's original plan.  I'm not sure at what point that stopped being the case - probably when the people maintaining the code failed to remember this was the plan...

We should certainly either fix the code to match the docs, or the docs to match the code.  And if `-py3` is to be required, someone should fix `setuptools`.

I don't have a strong opinion on which to do, though fixing the docs sounds like less work than making sure everything works with Python3 without `-py3`.

FWIW, for Xapian we just generate separate sets of wrapper files for Python 2 and Python 3, and build the appropriate one, handling it much like we would for two different languages supported by SWIG.  I think we'd need to do this even if `-py3` wasn't needed, as we have different custom typemaps for each case.

m7thon:
I'll close this, since Mike is working on this in PR #694.

I opened a new issue #695 summarizing the discussion points about `-py3` and python2 and python3 compatibility.

PR #694, opened 2016-06-03
New tests and fix for regressions in 3.0.9

mromberg:
These changes add new tests to swig for the cases where a swig python module is split between multiple packages.  There are 8 cases tested:

  1)  foo.py is in a package and _foo.so is in the same package.
   2) foo.py is in a package and _foo.so is in no package at all.
   3) foo.py and _foo.so are in the same package and -builtin
   4) foo.py is in a package and _foo.so is in no package and -builtin
     
  All tests pass on my machine.  But I'd like to see what the full test suite has to say...

EDIT:  remove the nopy3 tests.


mromberg:
Well, on my machine (fedora 23 x86_64) the following works:
make
make  check-python-examples
make clean
make
make check-python-examples PY3=1

  It all runs clean with no errors.   I have no idea what is going on with the travis thing.  It is not providing enough information to guess what is going wrong.  And I can find no documentation on how a travis build works.  I have no more ideas on how to move forward with making travis work.

wsfulton:
> 2) The same as #1 but -py3 is not used even for python3.
I don't see how you are testing python3 without -py3 in the patch. There is no need to do this btw.

Regarding Travis, it is going wrong running this command, where it doesn't matter what ```$CONFIGOPTS``` is set to (you should be able to replicate with it being empty, besides it goes wrong before re-running configure):
```
$ make check-maintainer-clean && ../../configure $CONFIGOPTS
```
You can see this if you scroll up a bit from where it goes wrong:
```
cleaning Examples/python/import_packages
/bin/sh: 1: cd: can't cd to pkg1
make[7]: *** [clean] Error 2
/bin/sh: 1: cd: can't cd to pkg1
make[7]: *** [clean] Error 2
/bin/sh: 1: cd: can't cd to pkg1
make[7]: *** [clean] Error 2
/bin/sh: 1: cd: can't cd to pkg1
make[7]: *** [clean] Error 2
make[6]: *** [clean] Error 2
make[5]: *** [clean] Error 2
make[4]: *** [import_packages.actionexample] Error 2
```

wsfulton:
btw, Travis is more or less running the bash code in .travis.yml in the ```before_install``` and ```script``` sections.

mromberg:
> I don't see how you are testing python3 without -py3 in the patch. There is no need to do this btw.

  From the Makefile in builtin_nopy3/pkg1/
build:
	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' \
        SWIGPYTHON='$$(SWIG) -python' \

  I'm setting SWIGPYTHON back to remove the -py3 that was placed there in $(TOP)/Makefile:

# SWIG option for Python
ifeq (,$(PY3))
  SWIGPYTHON = $(SWIG) -python
else
  SWIGPYTHON = $(SWIG) -python -py3
endif

  The changes in my branch no longer have any dependency on the -py3 switch.  So, it may not be needed for that.  But apparently there are users out there who do build code for python3 and do not use the -py3 switch.  So, unless it is decided that this is not allowed, then it is a good idea to test for this case.

mromberg:
> btw, Travis is more or less running the bash code in .travis.yml in the before_install and script sections.

  Thanks.  I'll look at that.  I'm guessing some sort of out of tree build is going on with the travis tests.  They work fine in the source tree.   It is very difficult to debug what is broken when one can't see much about what went wrong or the current state of the files on that machine.  I'll examine the yml shell code and see if I can recreate a similar build.

  If it is the case that an out of tree build is going on, then do all of the python source files need to get copied from some location (the source tree) to somewhere else (build tree)?   I'm sorta guessing that something like this is going on.  But it is not obvious (at least to me) what needs to be done to ensure python files that exist in git also exist wherever the tests are actually being run.

mromberg:
> set to (you should be able to replicate with it being empty, besides it goes wrong before re-running configure):
> 
> $ make check-maintainer-clean && ../../configure $CONFIGOPTS

This runs clean on my machine.  Near the end this appears:
skipping maintainer-clean check (in-source-tree build)

  Thus I think that travis is doing an out of source-tree build.  But I have no idea how to recreate that.

vadz:
@mromberg Just run the commands from `.travis.yml` that Travis CI runs, i.e.

      ./autogen.sh && mkdir -p build/build && cd build/build && ../../configure
      make

The only thing you need to know about out of source tree builds (which I personally strongly recommend anyhow, Travis or not) is that you must specify a relative path to configure, absolute path won't work due to the way SWIG makefiles are written.

mromberg:
@vadz  Thank you so much!  I am now able to reproduce the problem on my build machine.  Which compared to waiting for the cloud thing to finish is like going from the stone age to the 21st century!

  It is as I suspected.  I guess one must manually copy files out of the source tree into the build tree to actually run the tests.

mromberg:
OK, the tests now pass in every environment they need to run on.  Done tweaking (unless there are suggestions).

wsfulton:
Thanks @mromberg. There is just one thing. The split_modules examples duplicate tests, one for 'vanilla' and one for 'builtin'. Can you just provide one of these to simplify things. The reason is we can test the examples for -builtin by running ```make check-python-examples SWIG_FEATURES=-builtin``` and this is what Travis does for some of the python tests. Can you apply these whitespace fixes too please: https://gist.github.com/wsfulton/9bdfb32485f95e6ecd20a1e141d88641.

mromberg:
>>>>> " " == William S Fulton <notifications@github.com> writes:

     > Thanks @mromberg. There is just one thing. The split_modules
     > examples duplicate tests, one for 'vanilla' and one for
     > 'builtin'. Can you just provide one of these to simplify
     > things. The reason is we can test the examples for -builtin by
     > running ```make check-python-examples SWIG_FEATURES=-builtin```
     > and this is what Travis does for some of the python tests. Can
     > you apply these whitespace fixes too please:
     > https://gist.github.com/wsfulton/9bdfb32485f95e6ecd20a1e141d88641.

  Sure.   Now that I know -builtin is covered by one of the
SWIG_FEATURES permutations it makes sense to remove the redundant
tests.

  On a somewhat related note, I attempted to build the vanilla SVN 1.9.4
with my branch of swig.  The vanilla SVN sources check for a version
of swig older than 3.0.0.  Bypassing the check in the SVN configure
script reveals that the swig interface files do indeed not work at all
with newer swig versions.

  So, redhat/fedora (and probably others) must have a separate set of
patches to make the swig bindings work for SVN (and possibly other
packages).

Mike

mromberg:
Looks like travis failed on scilab because some java thing took more than 10 minutes to report its version number.   Everything else passed.   If you want, I can add another harmless commit to kick the tests off again.   Maybe there is a better way to restart the tests?

vadz:
I've restarted the failing build.

wsfulton:
Now I'm a bit lost. There are meant to be two scenarios fixed. Firstly could you give a final detailed summary of these two scenarios and I'll put it in the top level README file for the release notes. 

Secondly, there are two new tests, vanilla and vanilla_split. I can get vanilla_split to go wrong using the old SWIG executable (and this seems to replicate the svn problem), but vanilla always works, no matter what combination of -builtin and -py3 I use. What is this test doing that was broken in 3.0.9?

mromberg:
The remaining scenario is/was the builtin_nopy3 case.  That is:

  - Both foo.py and _foo.so are in the same package.

   - Swig was *not* given the -py3 switch

  - Python3 is used to import the module and run the test.

  You don't see this test/case fail because you aked for the tests where -py3 was not used for python3 to be removed.  So, swig never tests this case since you said the use of python3 without -py3 does not work and the docs are wrong.  :)

  If you do the above things with what went out in 3.0.9 then it breaks because of the dependency of the C portion of swig on the -py3 switch.   It used the state of the -py3 flag (for python2 backwards compatibility) to do either 'from ._foo import *' or 'from _foo import *'.   But swig never tested the case where python3 is used but -py3 was not used to build the wrapper.   Swig still does not do this.

  If you would like I could add the tests I called "nopy3" back in.  They were the ones that never added -py3 to SWIGPYTHON when PY3 was set.

mromberg:
> Secondly, there are two new tests, vanilla and vanilla_split. I can get vanilla_split to go wrong using the old SWIG executable (and this seems to replicate the svn problem), but vanilla always works, no matter what combination of -builtin and -py3 I use. What is this test doing that was broken in 3.0.9?

  When I created the tests, I was thinking of covering all the permutations of how swig would treat the _foo.so part of a wrapped module.  The idea was to collect all these cases in one spot so one could see how all the bases were being covered by the tests.   That way in the future if yet another wrinkle is added to this puzzle the next developer could see where to add a test for it.

  There were originally 8 tests because there are thee wrinkles.  Each with a binary condition:
  1)  The pair of modules either are placed into the same package or split.

   2)  The -builtin switch is used or not (makes the total tests now 4).  This permutation is covered by SWIG_FEATURES setting the flag.  So, it is not explicitly there in the code.

   3) Python3 is used with and without giving swig the -py3 switch.  Which doubles the tests to 8.   These cases were removed by a request above.

   It is probably true that the vanilla test is already covered by other swig tests.   And it is/was there to try and make what tests were covered more obvious.  But I'll remove it too if you'd like me to.

mromberg:
Thinking about this a bit, I wonder if part of this fix should be in the documentation too.  There are two points I can think of that could use clarification in the swig python documentation:

  1)  The whole bit about swig generating code that allows for foo.py to be in a package yet the C portion (_foo.so) being a global module.   I don't think this is ever specifically pointed out of being a feature of swig.  In particular, I can see one way of thinking about this as if foo.i was told it was to be in the package "bar" then both of the modules bar.foo and bar._foo are the only way to do it.  Yet swig has always allowed bar.foo and _foo to exist and work.   But I don't recall ever seeing this documented.  It clearly is a feature that folks are currently using.  And should probably be documented clearly somewhere.

  2)  The documentation about when one should use the -py3 switch.  Right now the documentation can very easily be read to mean that this switch turns on python3 only features.  But is not required for one to generate a wrapper that works with python3.   If this is incorrect then it should be fixed in the documentation.

  I'll volunteer to take a stab at a documentation update.  But feel it may be better done by someone with a bit more swig street cred :).

mromberg:
> Now I'm a bit lost. There are meant to be two scenarios fixed.

  In the interest of clarifying things (I hope) there are two scenarios fixed in these changes.  They are:

  1)  The code in 3.0.9 did not fall back to 'import _foo' if 'import bar._foo' failed (assuming bar.foo was the main module).   Every place _foo is imported now first tries it from the package where foo was found and if that fails tries _foo as a global module.

  2)  The separate block of python code that injected code to pull in the attributes from _foo when -builtin is used made use of the -py3 switch to either do 'from ._foo import *' or "from _foo import *".  This block of code no longer does this and instead checks the python version at runtime to switch between the two syntaxes.

wsfulton:
Let's leave the vanilla example in, although this one didn't break, it can't do any harm.

Expanded docs will be welcome, I'm quite happy to edit them if need be, but I think you've got a good handle on the import complications which seems overly complex in Python to me. Let's keep the discussion about -py3 requirement for Python3 in #695.

mromberg:
OK.  I'll take a pass at updating the documentation to cover the two points I mentioned above.  It may take me a day or so (I've got other commitments this evening).   I'm sorry again about all this.

  And yes. Python imports do seem more complicated than one would think they ought to be.  In the end swig should produce code that is more compatible with various package layouts by using the __name__ of the module instead of it's __file__.  Using __file__ is not compatible with namespace packages or custom python loaders.  But now it should be.  There was just the unexpected land mine of swig importing the C extension not using a package.   Now that will be covered and soon documented.

PR #698, opened 2016-06-06
Fix broken tests for python3 -relativeimport and no -py3

mromberg:
Converts the test for python3 support from a swigtime check of py3 to a runtime check of the python version number.  The exact same import syntax is still used.  But now the test for python3 is done at the time the module is imported.

  This allows the .py file to work (with the -relativeimport switch) with both python2 and python3.

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -5680,8 +5680,9 @@ class M2(pkg2.mod3.M3): pass
 <p>By default, SWIG would generate <tt>mod2.py</tt> proxy file with
 <tt>import</tt> directive as in point 1. This can be changed with the
 <tt>-relativeimport</tt> command line option. The <tt>-relativeimport</tt> instructs
-SWIG to organize imports as in point 2 (for Python 2.x) or as in point 4 (for
-Python 3, that is when the -py3 command line option is enabled). In short, if you have
+SWIG to organize imports as in point 2 (for Python < 2.7.0) or as in point 4
+for Python 2.7.0 and newer. This is a check done at the time the module is
+imported. In short, if you have
 <tt>mod2.i</tt> and <tt>mod3.i</tt> as above, then without
 <tt>-relativeimport</tt> SWIG will write</p>
 
@@ -5695,21 +5696,15 @@ import pkg1.pkg2.mod3
 write</p>
 
 <div class="targetlang">
-<pre>
-import pkg2.mod3
+  <pre>
+import sys
+if sys.version_info >= (2, 7, 0):
+    from . import pkg2
+    import pkg1.pkg2.mod3
+else:
+    import pkg2.mod3
 </pre>
 </div>
-
-<p>if <tt>-py3</tt> is not used, or</p>
-
-<div class="targetlang">
-<pre>
-from . import pkg2
-import pkg1.pkg2.mod3
-</pre>
-</div>
-
-<p>when <tt>-py3</tt> is used.</p>
 
 <p>You should avoid using relative imports and use absolute ones whenever
 possible. There are some cases, however, when relative imports may be
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -1251,13 +1251,14 @@ public:
       Printf(out, "import %s%s%s%s\n", apkg, *Char(apkg) ? "." : "", pfx, mod);
       Delete(apkg);
     } else {
-      if (py3) {
-        if (py3_rlen1)
-	  Printf(out, "from . import %.*s\n", py3_rlen1, rpkg);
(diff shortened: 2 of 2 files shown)


wsfulton:
Thanks. I made a few additions:
- Added print statements to all import_packages runme files to easier track the examples (use ```make -s``` in the Examples/python/import_packages directory, see dd40a25349dadc4011040474334c1fb7be3abecf
- Change import of all of sys - see 81adedd7dd3433d600a584337dcff2335dc12e9b. I'm not entirely happy with this as it adds ```version_info``` into the code and this name might be one that a user is using. How would one reduce the scope of this, other than putting the code into a 'if True: ...' block?

mromberg:
> Change import of all of sys - see 81adedd. I'm not entirely happy with this as it adds version_info into the code and this name might be one that a user is using. How would one reduce the scope of this, other than putting the code into a 'if True: ...' block?

  Yea.  I thought about just pulling in version_into from sys and calling it _swig_version_info  like so:

from sys import version_info as _swig_version_info

  But then it occurred to me that I am not a python historian and I don't know if this 'as' syntax works for all pythons.  If someone else knows if this works for all then that would be better

wsfulton:
I was worried about that too. From what I can see 'import x as y' requires python 2 and later, but given this is a stable branch of SWIG, I don't want to risk anything introducing this. Answering my own question, probably best to pop this code into a helper function instead.

m7thon:
I didn't see this pull request. Sorry. I'll repeat my comment here. You should definitely not add the `version_info` symbol into the module namespace.

> `from sys import version_info as _swig_version_info` is valid for all versions of python 2 [[ref](https://docs.python.org/release/2.0/ref/import.html)]

> You can put the above code snippet for `_swig_absolute_import`, or, if you don't feel like fixing that, then instead `from sys import version_info as _swig_python_version_info` between [these](https://github.com/swig/swig/blob/1495313e53c26f6e0394e0c7f0411fc27f1054ce/Source/Modules/python.cxx#L1050-L1051) lines.

> To put the `del _swig_python_version` or `del _swig_absolute_import` statement at the end of the file, print it [here](https://github.com/swig/swig/blob/1495313e53c26f6e0394e0c7f0411fc27f1054ce/Source/Modules/python.cxx#L1054).

> However, please do **not** use either of `import sys`, or `from sys import version_info`, since it will overwrite any previously defined symbol called "sys" or "version_info" (I can give you a test case that breaks, if you like).

m7thon:
I have a test [future_absolute_import](https://github.com/m7thon/swig/tree/relativeimports-improvements/Examples/python/import_packages/future_absolute_import) added to `import_packages` to show:

* That using `from __future__ import absolute_import` can be tested (and skipped for Python < 2.5). You just need to have `runme.py` import a module that has this first.
* That `-relativeimports` combined with `%pythonbegin %{ from __future__ import absolute_import %}` currently still does not work with Python 2.5 and 2.6 (it is currently fixed by this PR for 2.7).

To be honest, I have no idea why anyone would want to add this to their interface file, but it is explicitely mentioned in the docs under [36.11.3 Enforcing absolute import semantics](http://www.swig.org/Doc3.0/SWIGDocumentation.html#Python_absimport).

wsfulton:
Of course, don't test the package if python < 2.5! We can use this, thanks.

BTW, don't check in the runme3.py file, it gets generated by ```2to3```.

wsfulton:
```
Testing future_absolute_import - explicit relativeimports for python >= 2.5
  File "runme.py", line 8, in <module>
    import pkg
  File "/home/william/swig/github/swig/Examples/python/import_packages/future_absolute_import/pkg/__init__.py", line 1, in <module>
    from .bar import Pkg_Bar
  File "/home/william/swig/github/swig/Examples/python/import_packages/future_absolute_import/pkg/bar.py", line 108, in <module>
    import foo
ImportError: No module named foo
```
This is for python 2.7. I think you need py2 and py3 packages like the relativeimport1 test to deal with the differences between python 2 and python 3. Can you fix, rebase and create a pull request for this please? Then all the combinations on Travis will be tested.

wsfulton:
We could remove the repeated ```from sys import version_info as _swig_python_version_info``` but I'm running out of time atm. I'd like to make the release candidate this eve, so unless one of you put in a pull request to make further improvements, I think we should go with what we have for 3.0.10.

m7thon:
@wsfulton: Yes, go ahead and release, I currently don't se any serious issues.

I'll keep tinkering towards Option 2 above, i.e., get the default argument conversions working for Python 2 and 3, and maybe fix the imports when `from __future__ import absolute_imports` is used, including making some tests. But I wont rush this.

I see these things are tricky, and the current code is actually quite buggy for corner cases not covered by the tests.

PR #699, opened 2016-06-07
[Python] octal default values for python 2 *and* 3

m7thon:
This allows handling of octal default argument values in a Python version independent way by writing an octal value, say '0644' as `int('644',8)`, instead of `0644` for Python 2 or `0o644` for Python 3.

Also, the newly allocated `NewStringWithSize` is now deleted, which was previously not done (this was a memory leak, right?).

The function `int(x, radix)` is indeed available since Python 2.0, see the 
[Python 2.0 docs](https://docs.python.org/release/2.0/lib/built-in-funcs.html#l2h-167).

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -2127,14 +2127,15 @@ public:
 		// This must have been a hex number, we can use it directly in Python,
 		// so nothing to do here.
 	      } else {
-		// This must have been an octal number, we have to change its prefix
-		// to be "0o" in Python 3 only (and as long as we still support Python
-		// 2.5, this can't be done unconditionally).
-		if (py3) {
-		  if (end - s > 1) {
-		    result = NewString("0o");
-		    Append(result, NewStringWithSize(s + 1, (int)(end - s - 1)));
-		  }
+		// This must have been an octal number. We will let python convert it
+		// to an int using base 8, since we would have to change its prefix
+		// to be "0o" in Python 3 only.
+		if (end - s > 1) {
+		  result = NewString("int('");
+		  String *octal_string = NewStringWithSize(s + 1, (int)(end - s - 1));
+		  Append(result, octal_string);
+		  Append(result, "', 8)");
+		  Delete(octal_string);
 		}
 	      }
 	    }


wsfulton:
Nice idea, but I can't see this working for wrapping values > 32 bit on Python 2 on 32 bit operating systems where sys.maxint is 0xFFFFFFFF (if I recall correctly - can't check right now). This would happen when wrapping for example:

void f(unsigned long long big=040000000000);

m7thon:
Good point, but the docs for `int(x, radix)` for Python 2.0 state:

> ... the argument may be a plain or long integer or a floating point number...

I tested with Python 2.5 on a 32-bit machine, and indeed
```python
>>> import sys
>>> sys.maxint()
2147483647
>>> int('77777777777', 8)
8589934591L
>>> type(int('77777777777', 8))
<type 'long'>
```

In other words, Python does not overflow when converting integer literals to integers and will propagate to long (which is arbitrary length?) automatically.

m7thon:
I did some Python historical work, and edited (well rewrote) my above comment. Please have a look.

Basically, this PR does not change the situation for octal values > 32 bit. This didn't and doesn't work for Python < 2.3, but did and does still work for Python >= 2.3. It is independent of the system architecture.

m7thon:
Oh no, the old `convertValue` code never actually worked for negative octals in Python 3 (with or without the `-py3` flag), so this PR extends this failure to Python 2.

I guess I need to fix the old bugs first... And perhaps extend the test-case to cover negative octal literals.

Sigh.

mromberg:
> This didn't and doesn't work for Python < 2.3, but did and does still work for Python >= 2.3

  Does swig even support pythons that old?   I think the tests are only being run on python 2.4 - 3.5.  So, maybe this is not a problem.

mromberg:
> writing an octal value, say '0644' as int('644',8), instead of 0644 for Python 2 or 0o644 for Python 3.

  Nice!  I forgot about this.  Why not just do int('644', 8) for both python2 and python3?

m7thon:
According to the SWIG docs, Python >= 2.0 should in principle work. But yes, if larger than 32-bit octagonal or hexagonal literals cause problems on Python < 2.3, well, then so be it.

But the issue I would like to fix is with negative octal literals, e.g. -01 (who in their right mind would use such a thing, I wonder...). It seems someone just patched in a fix for Python 3 without considering all the possible cases that can go wrong. Also, C++11 allows long hexadecimal literals such as `0xFFL` (I think), and this I think currently breaks on Python 2 and 3. 

In general I want to have a think abut what the logic of the `convertValue()` function is/should be. What I think the function should do is:

* Literals that can safely be written as Python literals should be used/converted (this does not have to cover all cases)
* Cases that can lead to problems should be handled by C default arguments (currently, problematic cases can slip through to Python)
 
> Why not just do int('644', 8) for both python2 and python3

Erm, that is exactly what I propose.

wsfulton:
I'm inclined to make as few changes as possible in swig-3.0.x particularly given the minimum versions we aim to support. For 3.1, we havn't decided what minimum version to support, but it'll probably be somewhere between 2.5-2.7 - see newly opened issue #701. So I suggest your new patch targets python-2.5 minimum for now and give it the milestone of swig-3.1 (there are a few others already).

PR #704, opened 2016-06-08
java directorargout + java/typemaps.i fixes

andrey-starodubtsev:
- directorargout didn't work - it used string "jresult" instead of
  argument name
- java/typemaps.i didn't work with directors
- test for using java/typemaps.i in directors added


wsfulton:
Thanks for adding the leak fixes and improved director support. The test case makes it good comprehensive patch. Can you revert the change in javahead.swg and adjust accordingly and submit an updated patch? Note that you can use {} braces with `noblock=1` to pass the code through SWIG's preprocessor and not generate {} braces, see http://swig.org/Doc3.0/Preprocessor.html#Preprocessor_typemap_delimiters, thereby avoiding having the JCALLx macros in the generated code. I don't want the JCALLx macros in the generated code.

Source/Modules/java.cxx
```diff
@@ -4387,7 +4387,7 @@ class JAVA:public Language {
       for (p = l; p;) {
 	if ((tm = Getattr(p, "tmap:directorargout"))) {
 	  addThrows(n, "tmap:directorargout", p);
-	  Replaceall(tm, "$result", "jresult");
+	  Replaceall(tm, "$result", Getattr(p, "name"));
```
wsfulton: It looks like a similar change is required in plenty of the other languages, like csharp.cxx.

Examples/test-suite/java_director_typemaps.i
```diff
@@ -0,0 +1,265 @@
...
+  Quux() {}
+  virtual ~Quux() {}
+
+  virtual void director_method_bool_output(
+    bool& boolarg_output,
+
+    signed char& signed_chararg_output,
+    unsigned char& unsigned_chararg_output,
+
+    short& shortarg_output,
```
wsfulton: Can you also add a test here where there is no parameter name in the wrapped virtual function and make sure that works? Something odd happened when I tried, the type seems to change.

wsfulton: I suggest `makeParameterName()` is used to create the name.

wsfulton:
Ping @andrey-starodubtsev, did you see the changes I requested? I can extend the changes to the other languages if you polish up the Java side.

andrey-starodubtsev:
@wsfulton, sorry, it took a while to get back to this work. I tried to fix all issues you mentioned.

wsfulton:
Thanks, merged for the swig-3.0.12 release.

PR #705, opened 2016-06-08
check-android-examples  related fixes

andrey-starodubtsev:
- android artefacts added to .gitignore
- build.xml's are updated by latest `android` tool
- -fexceptions added to CFLAGS to make android tests pass
- local.properties removed from source control (they contain local paths
  and created/updated automatically during `android update project`)


wsfulton:
The Android docs need to be updated in line with these changes. Can you please have a look at the Doc/Manual/Android.html file? You'll need to mirror your Android.mk file changes in here at the minimum before I merge in this patch. If there is anything else in these docs that needs updating, please enhance your patch with the updates.

It would also be immensely helpful if you were able to make an additional pull request to enhance our Travis builds to run the Android examples. I understand Travis now supports Android and if it can just run ```make check-android-examples``` after whatever 'sudo apt-get install' packages are needed to make this work now, it'll enhance the SWIG testing.

ojwb:
@andrey-starodubtsev Did you see William's comment above?

andrey-starodubtsev:
@wsfulton, I can check Doc/Manual/Android.html, but English is not my mother tongue, so all changes I make will require to be proofread.

Also, in this PR I just tried to restore android examples functionality, but modern android applications use `gradle` instead of `ant` - I can add additional examples, but I think it would be better to do this in separate PR.

About travis: that's right, travis supports Android, but it does it in a specific way: according https://docs.travis-ci.com/user/languages/android, one has to set `language: android` to set up travis android build. Alas, travis CI has no support of multiple languages (https://github.com/travis-ci/travis-ci/issues/4090). I can change language, but it can potentially break matrix build for desktops (I can try this in a separate branch anyway). On the other hand, it's possible to create android automated builds based on bare 'language: cpp' installing all necessary android SDKs on fly each time travis build is triggered - this approach was used as standard until travis added `language: android`. But this means that we have to download and install android SDK + android NDK + at least one platform each build (>3Gb). Yet another way is using separate CI for android builds, like Circle CI or Greenhouse CI.
Which approach do you think would be better for swig?

wsfulton:
No problem about proof-reading, I'm more than happy to edit the English.

The android 'language' on Travis just provides a machine with the Android ndk and whatever else is already required pre-installed. Like testing other SWIG languages, we have to find out how these VMs are set up and just replicate it by hand. I suspect we can find this out by asking the Travis folk or finding the appropriate scripts they run to set up a machine. For the other SWIG languages, this is easy to guess as it is just a 'sudo apt-get install' of one or two packages, but I don't know what the best way to automate android installation now is. 

But in answer to your questions, we a small addition to our .travis.yml which uses 'language:cpp' and modify Tools/travis-linux-install.sh to install the appropriate packages.

PR #707, opened 2016-06-10
[Python] improved default argument handling

m7thon:
The current SWIG handling of default arguments for python has several issues:
- Tangled code.
- Incorrect handling of several cases:
  1. Negative octals. Currently not handled correctly when `-py3` given (unusual case, but incorrect).
  2. Arguments of type "octal + something" (e.g. `0640 | 04`). Currently drops everything after the first octal. Nasty!
  3. Bool arguments "0 + something" (e.g. `0 | 1`) are always "False" (unusual case, but incorrect).
- Test cases test only for correct _syntax_, not for correctness of the converted value.
- Different handling of octal literals for Python 2 and Python 3 based on `-py3` switch.

This PR addresses all of the above issues. Here are some comments:
- The relevant conversion code is refactored and should be much more clear now.
- Test cases including some new tests are (slightly) improved (but still only test for corner cases)
- Identical code generated for Python 2 and Python 3. This now writes octals such as `0644` as `int('0644', 8)` (available in Python >= 2.0, propagates to long integer as of 2.3).
- There have always been issues with integer literals exceeding 32 bit on Python <= 2.2. These issues remain.
- Tests now pass with Python 2 and Python 3.

Note that this is a much improved version of the previous PR #699.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -2013,6 +2013,66 @@ public:
   }
 
   /* ------------------------------------------------------------
+   * convertIntegerValue()
+   *
+   * Check if string v is an integer and can be represented in
+   * Python. If so, return an appropriate Python representation,
+   * otherwise (or if we are unsure), return NIL.
+   * ------------------------------------------------------------ */
+  String *convertIntegerValue(String *v, SwigType *resolved_type) {
+    const char *const s = Char(v);
+    char *end;
+    String *result = NIL;
+
+    // Check if this is an integer number in any base.
+    long value = strtol(s, &end, 0);
+    if (errno == ERANGE || end == s)
+      return NIL;
+    if (*end != '\0') {
+      // If there is a suffix after the number, we can safely ignore any
+      // combination of "l" and "u", but not anything else.
+      for (char *p = end; *p != '\0'; ++p) {
+        switch (*p) {
+          case 'l':
+          case 'L':
+          case 'u':
+          case 'U':
+            break;
+          default:
+            return NIL;
+        }
+      }
+    }
+    // So now we are certain that we are indeed dealing with an integer
+    // that has a representation as long given by value.
+
+    if (Cmp(resolved_type, "bool") == 0)
+      // Allow integers as the default value for a bool parameter.
+      return NewString(value ? "True" : "False");
+
+    if (value == 0)
+      return NewString(SwigType_ispointer(resolved_type) ? "None" : "0");
+
+    // v may still be octal or hexadecimal:
+    const char *p = s;
+    if (*p == '+' || *p == '-')
+      ++p;
+    if (*p == '0' && *(p+1) != 'x' && *(p+1) != 'X') {
+      // This must have been an octal number. This is the only case we
+      // cannot use in Python directly, since Python 2 and 3 use non-
+      // compatible representations.
+      result = NewString(*s == '-' ? "int('-" : "int('");
+      String *octal_string = NewStringWithSize(p, (int) (end - p));
+      Append(result, octal_string);
+      Append(result, "', 8)");
+      Delete(octal_string);
+      return result;
+    }
+    result = *end == '\0' ? v : NewStringWithSize(s, (int) (end - s));
+    return result;
+  }
+
+  /* ------------------------------------------------------------
    * convertDoubleValue()
    *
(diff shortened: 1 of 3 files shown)


Source/Modules/python.cxx
```diff
@@ -2060,120 +2119,34 @@ class PYTHON:public Language {
...
-    }
-
-    // Check if this is a floating point number (notice that it wasn't
-    // necessarily parsed as a long above, consider e.g. ".123").
-    if (!fail && !result) {
+    result = convertIntegerValue(v, resolved_type);
+    if (!result) {
       result = convertDoubleValue(v);
       if (!result) {
-	if (Strcmp(v, "true") == 0 || Strcmp(v, "TRUE") == 0)
```
mromberg: I realize this commit is recreating the logic (to some extent) of what was already there.  But I wonder about the wisdom of converting "TRUE" and "FALSE" into python's True and False.  

  In C++ "true" and "false" are reserved keywords.  So, doing this for them makes sense.  But "TRUE" and "FALSE" are only possible pre-processor tokens.  And it is possible that they may not be defined in a manner consistent with True and False.

mromberg:
I wonder how well convertIntegerValue() is covered by the existing tests?   It is nice to see more tests with this pull request.  It could be that convertIntegerValue() is used for more than just default arguments.   It would be nice to have tests for all branches inside it.  But I'm not sure how difficult this would be to do.  Just thoughts...

Source/Modules/python.cxx
m7thon: Yes, I fully agree. Actually, someone may define FALSE as -1 and use this as a default value for an int, so this could cause a real problem (and it will never do any good).

I'll add a commit to throw those tokens out.

m7thon:
> It could be that convertIntegerValue() is used for more than just default arguments.

No, I checked. That is all it is used for.

> I wonder how well convertIntegerValue() is covered by the existing tests? ... It would be nice to have tests for all branches inside it.

Yes, that would be nice. As I say, the tests are improved, but are far from being meaningful unit tests.

mromberg:
> It could be that convertIntegerValue() is used for more than just default arguments.
> 
> No, I checked. That is all it is used for.

  convertValue() is not declared static.  So, it is (at least in theory) possible to call it from other places other than just python.cxx.  Might be worth sticking a static on there even in your local build tree and make sure no other spots use it.

  If it is only being used for default args then that is nice.  There is at least an indirect way to test it. :)

m7thon:
The `convert...Value()` functions are member functions of `PYTHON` (which inherits from `Language`). Sure they could be made `static` (since they do not depend on the state of the `PYTHON` object), or `private`, or become non-member functions and declared `static` (to be available only in the local compilation unit).

But what's the point? Why should it not be allowed to reuse these functions for some new purpose if someone wants to do so?

mromberg:
> The convert...Value() functions are member functions of PYTHON

  Whoops.  You are right.   I was not used to seeing all those inline functions declared like that in a cxx file.   Disregard my comment.

ojwb:
There are tedious conflicts in `Source/Modules/python.cxx` thanks to a mass whitespace change @wsfulton did since this PR was submitted, so I've squashed it into one commit to make it simpler to reapply.

In pre-C++11 code, technically handling `nullptr` as `NULL` has similar issues to assuming `FALSE` is `0`, but I don't think we have a way to know if we're parsing for C++11 or earlier, and I think the benefits of assuming `nullptr` has its C++11 meaning outweigh the theoretical problems.

Will push once I see the testsuite passing locally.

m7thon:
@ojwb Do you want me to rebase this PR on the current master? I am unsure.

ojwb:
@m7thon Sorry, I meant I'd push once the tests passed, but it seems I got distracted.  Sorry about that.

Have re-rebased and am currently running the tests...

ojwb:
The merged branch failed on a new part of test case `default_args`, because a default parameter such as `unsigned foo = -1U` is converted to constant `-1` in Python which is then rejected because `-1` is out of range for an unsigned.

I've fixed that simply by just handling negative unsigned constants in C/C++ instead:

2425c8d6d896dfb4eacb1b5877a5fa6274fde8d8

More sophisticated would be to convert such constants correctly, but I wanted to get this branch merged - enhancements can be done as a follow-on if someone wants to.

Issue #695, opened 2016-06-03
[Python] Clarify meaning of `-py3` and python2 vs. python3 compatibility


m7thon:
Currently, according to the SWIG docs:
- `swig -python` will generate wrappers that should work with python2 _and_ python3, and
- the `-py3` flag can be specified to enable python3 specific features which might break backwards compatibility with python2 (this clearly was the original intention of the `-py3` flag).
- The swig python examples (in the docs) make little/no mention of the `-py3` flag.

However, the SWIG code assumes in several places that **the `-py3` flag must be given to enable python3 compatibility**, i.e., that SWIG cannot generate wrapper code that is simultaneously compatible with python2 and python3. Furthermore, all SWIG tests are run with the `-py3` flag when generating wrapper code for python3.

This is a discrepancy asking for trouble, and should be resolved either by **(1) clarifying that python3 requires `-py3`** in the documentation (and possibly merging the `-python` and `-py3` flags), or by **(2) making `swig -python` work for python2 _and_ python3** by fixing the codebase. I list some things worth considering that I came across for anyone interested in tackling this 

**EDIT: This is now resolved via Option 2 below.**
## Option 1: Python3 really does require the `-py3` flag
- Obviously, the documentation under "Python 3 Compatibility" needs to be updated. But also, the typical use cases throughout the documentation should show the use of `-py3` (perhaps by default?) to make sure people are aware of this flag.
- The logic of having `-python` as a language specifier that really means `-python2`, and having `-py3` as an extra flag that really means `-python3`, is unintuitive and unsymmetrical. It would make more sense to use either language specifiers `-python2` and `-python3`, or one language specifier `-python` and flags `-py2` and `-py3` to select the language version, perhaps defaulting to `-py2`. The typical usage of `swig -python -py2` could be shown throughout the documentation to emphasized the point that `-python -py2` wrapped code is not guaranteed to work under Python 3.
- The minor downside of this is that any project that wishes to include pre-generated python wrapper code and wishes to support python2 and python3 is forced to handle both cases separately.
- Python distutils and setuptools support swig-wrapped extensions by automatically swigging and building them. They don't ever specify the `-py3` flag, so this needs to be fixed.
## Option 2: `-python` means Python 2 _and_ 3 compatibility

There are only three code-paths that currently use the `-py3` flag for Python 3 compatibility. These are:
[...]

PR #709, opened 2016-06-12
[Python] improved wrapping of division operators

m7thon:
This changes the way the division operators `operator /` and `operator /=` are wrapped in the case that `-builtin` is _not_ used (for `-builtin` everything stays the same, except that the operators are renamed to `__truediv__` and `__itruediv__`, but the same slots `nb_divide` and `nb_inplace_divide` are used).

The division operators are now always wrapped as `__truediv__` and `__itruediv__`, with alias methods
`__div__` and `__idiv__` generated for compatibility with Python 2. The same technique was already used for the wrapping of `operator bool` as `__nonzero__` and `__bool__`.

This allows correct wrapping for Python 2 and 3 **independent of the `-py3` flag**, and fixes these operators for Python 2 when using `from __future__ import division` (this was broken).

### DIFF
--- a/Lib/python/pyopers.swg
+++ b/Lib/python/pyopers.swg
@@ -103,11 +103,6 @@
 %pybinoperator(__neg__,      *::operator-(),		unaryfunc, nb_negative);
 %pybinoperator(__neg__,      *::operator-() const,	unaryfunc, nb_negative);
 %pybinoperator(__mul__,      *::operator*,		binaryfunc, nb_multiply);
-#if defined(SWIGPYTHON_PY3)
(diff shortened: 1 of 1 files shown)


wsfulton:
This is good. @m7thon, would you mind adding a runtime test for checking this as a new patch?

As discussed in #695, it removes one the last of the problems for python 2 if -py3 is used. The `SWIGPYTHON_PY3` SWIG preprocessor macro which was only recently introduced to fix python 3 division bugs, could also be removed and help ensure py2 and py3 incompatibilities don't creep in again.

PR #711, opened 2016-06-14
-python -O Implicittest "fix"

mromberg:
In the interest of getting all the tests working with -python and -O I started with the first that failed.   And I "sorta" fixed it here.   Let me first describe what I found.

  This test is failing on the case where an object (CCC) has two constructors.  One is native and the other added with %extend.  Like so:

class BBB {
  public:
    BBB(const B &) {}
};

class CCC {
  public:
    CCC(const BBB &bbb);
%}

// CCC(const BBB &) was being called instead of this constructor (independent of being added via %extend)
%extend CCC {
  CCC(const B& b) {
  }

  The test that fails calls a CCC constructor with a new B like so:

ccc = CCC(B())

  And then checks to see if the extend constructor is called.   When -O is used it is not.

  Here is my understanding of why it fails.  First the swig python module creates a prioritized dispatch order for each of the overloaded functions.  Note that this order is *not* dependent on the type of the object of the parameter at runtime.  It is based on the types of the parameters for each overloaded function called.  The above test works through a combination of blind luck and the fact that SWIG_ConvertPtr() is called with no flags set as it's last argument.  The test "works" without -O.  But it only does because of the current order swig sees the overloaded methods and the current priority scheme.

  When the -O flag is set then Swig_ConvertPtr() morphs from a function that just does an "is a" type check into one that also checks for known casts.  It changes becaue the SWIG_POINTER_IMPLICIT_CONV flag is now given to it.  Now the fixed chain of the dispatch table will stop at the first overloaded function that is given not just an argument that "is a" B but any argument that can be cast in any way (by calling other ctors) to a B.  This causes the first ctor CCC(BBB &bbb) to be called since BBB has a ctor that takes a B and SWIG_ConvertPtr() "gets 'er done".

  The only way to fix this that I can think of (there may be a way I'm not thinking of) is to save part of the parse tree for every set of overloaded functions and lookup the correct overloaded function at runtime based on the type of the arguments at runtime.  One would have to abandon the use of Swig_overload_dispatch() from overload.cxx and instead use a custom dispatch method for python just to do this.   And it seems it would be pretty messy.

  I chose to not explore this route since swig is not a full C++ compiler and python is not a strongly typed language and does not support overloaded methods.  Attempting to glue these two together is going to cause problems.  The swig documentation mentions several times that overloaded methods are going to cause problems.

  So, I see a few options to "fixing" this:

  1) Remove the test that is breaking.  Or change it to use a %rename
   2) Add runtime support as mentioned above.
   3) Tweak swig slightly to make the test continue to work.

  These two commits are in the third category.  I found that if I enabled -castmode then the swig runtime code still kept the fixed dispatch "table" from overload.cxx.  But added a maximum cast depth check.  This additional code causes swig to prefer the CCC(B&) constructor and the test works even with -O enabled.   One could even argue that -castmode should always be on for this test anyway since that is what this test is doing.  And I think -castmode makes the overloading more closely (but not exactly) match how C++ does things.  All in all if you use overloaded functions I think it should always be on.

  I added two commits.  One switches -castmode on in the Makefile for just the test in question.  The second commit enabled -castmode whenever -O is enabled.  I did the second commit because I think the idea is to make -O -python work with all tests and then make -O the default for swig-3.1.  Either one or both of the commits fix the test failure.

  Anyway.  Sorry to be long winded.  Dispatching overloaded methods in python is a bit on the complicated side.

### DIFF
--- a/Examples/test-suite/implicittest.i
+++ b/Examples/test-suite/implicittest.i
@@ -1,4 +1,7 @@
-%module(naturalvar="1") implicittest
+%module(naturalvar="1", castmode="1") implicittest
+// castmode is enabled because %implicitconv is enabled by -O
+// and %implicitconv highly recommends castmode.  This test will
+// fail with -O and no castmode.
 
 %implicitconv;
 
--- a/Examples/test-suite/python/Makefile.in
+++ b/Examples/test-suite/python/Makefile.in
@@ -96,7 +96,7 @@ LIBS         = -L.
 VALGRIND_OPT += --suppressions=pythonswig.supp
 
 # Custom tests - tests with additional commandline options
-implicittest.cpptest: SWIGOPT += -castmode
+# none!
 
 # Rules for the different types of tests
 %.cpptest:
--- a/Examples/test-suite/python/Makefile.in
+++ b/Examples/test-suite/python/Makefile.in
@@ -96,7 +96,7 @@ LIBS         = -L.
 VALGRIND_OPT += --suppressions=pythonswig.supp
 
 # Custom tests - tests with additional commandline options
-# none!
+implicittest.cpptest: SWIGOPT += -castmode
 
 # Rules for the different types of tests
 %.cpptest:
--- a/Lib/python/pyprimtypes.swg
+++ b/Lib/python/pyprimtypes.swg
@@ -223,6 +223,8 @@ SWIG_AsVal_dec(long long)(PyObject *obj, long long *val)
     const double mant_min = -mant_max;
     double d;
     res = SWIG_AsVal(double)(obj,&d);
+    if (SWIG_IsOK(res) && !SWIG_CanCastAsInteger(&d, mant_min, mant_max))
+      return SWIG_OverflowError;
     if (SWIG_IsOK(res) && SWIG_CanCastAsInteger(&d, mant_min, mant_max)) {
       if (val) *val = (long long)(d);
       return SWIG_AddCast(res);
@@ -280,6 +282,8 @@ SWIG_AsVal_dec(unsigned long long)(PyObject *obj, unsigned long long *val)
     const double mant_max = 1LL << DBL_MANT_DIG;
     double d;
     res = SWIG_AsVal(double)(obj,&d);
+    if (SWIG_IsOK(res) && !SWIG_CanCastAsInteger(&d, 0, mant_max))
+      return SWIG_OverflowError;
     if (SWIG_IsOK(res) && SWIG_CanCastAsInteger(&d, 0, mant_max)) {
       if (val) *val = (unsigned long long)(d);
       return SWIG_AddCast(res);
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -161,7 +161,7 @@ static const char *usage3 = "\
      -safecstrings   - Use safer (but slower) C string mapping, generating copies from Python -> C/C++\n\
      -threads        - Add thread support for all the interface\n\
      -O              - Enable the following optimization options: \n\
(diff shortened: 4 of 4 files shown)


mromberg:
Turns out castmode was also broken with the overflow tests in primitive_types.  I think I fixed that too.

wsfulton:
The notes on %implicitconv in the CHANGES file on 12/27/2005 state:
```
...
it is highly recommended that you use
the new 'castmode' when mixing implicit conversion and
overloading.
```
and are in line with your conclusions. Hence, I'll accept adding -castmode to the test. However, I'd prefer that %module(castmode="1") approach to adding -castmode in the Makefile including some comments as to why. For SWIG 3.1, we could warn/error if castmode is not enabled with implicitconv.

Looks like the primitive_types testcase started failing due to the patch that add -castmode to -O. This will probably break someone's wrappers in swig-3.0.x, so I'll reject this commit. I think the pull request should be changed as follows:

7fe274d - change to turn it on in %module as mentioned above
ad5f68c - remove
7cbbc86 - accept as a fix for when the test-suite is run with SWIG_FEATURES=-castmode

mromberg:
> Looks like the primitive_types testcase started failing due to the patch that add -castmode to -O.

  Yes.  Cast mode always failed to throw an overflow error on long long and unsigned long long.  I have a commit in this pull request that fixes castmode.

mromberg:
> However, I'd prefer that %module(castmode="1") approach to adding -castmode in the Makefile including some comments as to why.

  Sure I can do this.  What about the patch which adds -castmode to -O?   It seems that implicitconv can break pretty hard without -castmode.  And if a user just does -O and -O does not also enabe -castmode it is kinda non-obvious that the .i file should also be tweaked.   I like the notion of -O also adding -castmode and then maybe with swig-3.1 we can throw a warning.

wsfulton:
I really want the absolute minimum change to 3.0.x as there have been too many 'improvements' that have broken previously working code. Changing what -O does should go into 3.1. So as I said, remove ad5f68c. Maybe we can add -castmode to -O for 3.1 (which will probably the default option for 3.1). However, I'm not convinced -castmode should be turned on by default in 3.1. Discuss in the 3.1 development wiki or issues. Right now the advice is for users to use -castmode for implicitconv.

ahnolds:
I __strongly__ disagree with the analysis that the current dispatch system relies on luck.  The implementation of Swig_overload_dispatch first checks for exact matches, and only then checks for implicit casts, hence the two passes through the list of overloaded functions, first with `flags = 0` and then with `flags = SWIG_POINTER_IMPLICIT_CONV`.  The problem (as I see it, at least), is that when @wsfulton addressed this issue in b58dabced9e7424e6107bfd40963806f3b831a19, he only did so for the normal (i.e. not `-fastdispatch` not `-castmode`) case.

See #730 for a pull request that addresses this issue in a (I claim) better way.

wsfulton:
I've opted for patch #730 over this patch as it fixes the problem rather than working around it. However, 7cbbc86 from this patch which fixes -castmode and integer conversion is a separate improvement and I've cherry-picked it out and applied it. @ahnolds, it extends your improvements merged in ee867a61432f210f228a7fe113ddd444fabdde1a for -castmode in case you want to cast your eye over it.

ahnolds:
7cbbc86 looks fine to me.  And for what it's worth, I'd be against making castmode a default in 3.1, since silently converting between ints and floats seems undesirable to me.  Compare the [-Wconversion](https://gcc.gnu.org/wiki/NewWconversion) flag in GCC, which explicitly warns about such implicit casts.  I'm happy to discuss in more depth elsewhere.

PR #717, opened 2016-06-16
[Java] Fix compilation problems with cMemOwnDerived

schnelle:
This pull request fixes the compilation problems with the cMemOwned vs cMemOwnDerived problems when using boost shared pointers and directors.

I tried to keep the original idea of introducing a new variable in the derived class for that purpose, but avoided compilation problems by simply keeping the old name. This leverages from the need to introduce new mapping rules for derived classes. Java takes care of selecting the correct variable. So far, I did not observer any problems with it. All tests are passed.

Moreover, Java complains about missing calls to super in the finalize. This has been fixed by calling super.finalize and adding an exception that is defined by Object.

### DIFF
--- a/Lib/java/boost_shared_ptr.i
+++ b/Lib/java/boost_shared_ptr.i
@@ -161,11 +161,11 @@
 // Derived proxy classes
 %typemap(javabody_derived) TYPE %{
   private transient long swigCPtr;
-  private boolean swigCMemOwnDerived;
+  private transient boolean swigCMemOwn;
 
   PTRCTOR_VISIBILITY $javaclassname(long cPtr, boolean cMemoryOwn) {
     super($imclassname.$javaclazznameSWIGSmartPtrUpcast(cPtr), true);
-    swigCMemOwnDerived = cMemoryOwn;
+    swigCMemOwn = cMemoryOwn;
     swigCPtr = cPtr;
   }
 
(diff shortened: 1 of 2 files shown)


schnelle:
Trying to rebuild since it failed because of an authentication request for jsnode

wsfulton:
I've restarted the failing Travis tests so they should pass shortly. The patch will need a testcase before acceptance. There is a shared_ptr testcase already using directors in Examples/test-suite/director_smartptr.i. and runtime test Examples/test-suite/java/director_smartptr_runme.java. Please add in whatever problem this patch fixes into this testcase.

Can you adapt the patch to keep the original names, that is, have different variable names for base and derived classes because this ensures a pile of subtle problems are not masked. But, currently, without a testcase, I'm not really sure what you are trying to solve and why you had to change make them the same.

ojwb:
@schnelle Did you see William's comment above?

ojwb:
@schnelle Ping?  Please see the questions William asked above.

Without some feedback as to exactly what problem this is trying to solve this ticket seems like it's going to remain stalled indefinitely (in which case we might as well just close it IMO - having a ticket open forever doesn't seem useful).

schnelle:
The fix was mainly about changing the name which caused a lot of problems for me. Main issues where unresolved references to swigCMemOwn which was unknown in derviced classes. On the other side, I would be interested to learn which problems occur if the name for base and derived classes are the same.

wsfulton:
`swigCMemOwn` is purposefully renamed in the shared_ptr wrappers as the memory handling is quite different and a compiler error is much better easier to track down than subtle memory leak bugs when typemaps aren't quite right and also memory handling bugs due to a mix of non shared_ptr and shared_ptr wrappers in an inheritance hierarchy.

As the main thrust of this patch is base on the renaming, I suggest we close it. Regarding the "adapted finalize to java recommendations" commit, this will need doing uniformly for all finalizers. I suggest a new patch and references to this recommendation.

wsfulton:
No follow up - closing.

PR #718, opened 2016-06-16
Fix for -python -O %feature nondynamic

mromberg:
This code uses metaclasses.  It did before my fix.  But the old code only worked for python2 meta classes.    The old code also did not pay any attention to the python properties switched on with the -O switch.

  This works with python2 and python3.  It may not work for python2 older than 2.7.  If that is the case, then someone who's brain can grapple with metaclasses and properties better than mine can take a shot at it.

  Otherwise, this fixes the tests to move forward for swig-3.1 in supporting python 2.7+

Mike

### DIFF
--- a/Examples/test-suite/python/python_nondynamic_runme.py
+++ b/Examples/test-suite/python/python_nondynamic_runme.py
@@ -4,6 +4,11 @@ aa = python_nondynamic.A()
 
 aa.a = 1
 aa.b = 2
+
+# Sanity check to see if all the funky setattr methods make it into C.
+assert(python_nondynamic._python_nondynamic.A_a_get(aa) == 1)
+assert(python_nondynamic._python_nondynamic.A_b_get(aa) == 2)
+
 try:
     aa.c = 2
     err = 0
@@ -23,6 +28,11 @@ class B(python_nondynamic.A):
     pass
 
 bb = B()
+bb.a = 4
+bb.b = 5
+assert(bb.a == 4)
+assert(bb.b == 5)
+
 
 try:
     bb.c = 3
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -910,25 +910,60 @@ public:
          defined in this module (try both Python 3 and Python 2 names) */
       Printv(f_shadow, "try:\n", tab4, "import builtins as __builtin__\n", "except ImportError:\n", tab4, "import __builtin__\n", NULL);
 
+      Printv(f_shadow, "\n", NULL);
+      Printv(f_shadow, "class _swig_nondynamic_meta(type):\n", NULL);
+      Printv(f_shadow, tab4, "def __init__(cls, name, bases, dct):\n", NULL);
+      Printv(f_shadow, tab8, "# Adds a __setattr__ method to a class.\n", NULL);
+      Printv(f_shadow, tab8,
+        "super(_swig_nondynamic_meta, cls).__init__(name, bases, dct)\n", NULL);
+      Printv(f_shadow, tab8,
+        "cls.__setattr__ = cls.__class__.get_setattr(cls)\n\n", NULL);
+      Printv(f_shadow, tab4, "@classmethod\n", NULL);
+      Printv(f_shadow, tab4, "def get_setattr(mcs, cls):\n", NULL);
+      Printv(f_shadow, tab8,     "def setter(self, name, value):\n", NULL);
+      Printv(f_shadow, tab8, tab4,
+        "return _swig_setattr_nondynamic(self, cls, name, value)\n", NULL);
+      Printv(f_shadow, tab8, "return setter\n\n", NULL);
+
+      Printv(f_shadow, "def _swig_with_metaclass(meta, *bases):\n", NULL);
+      Printv(f_shadow, tab4, "class metaclass(meta):\n", NULL);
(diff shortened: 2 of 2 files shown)


m7thon:
This unfortunately still fails for the following uses (as mentined in #719):

1. Module import fails for Python 2.7 when using `-modern -fastproxy` (e.g., `-O`) for: 

```
%module nd
%pythonnondynamic NonDynamic;
%inline %{
  class NonDynamic {
  public:
    NonDynamic() : m(1) {}
    int m;
    bool broken() { return false; }
  };
%}
```

2. Class extension of nondynamic classes fails with no SWIG flags on Python 2 and 3:

Same SWIG interface as above, and following python code:
```
import nd
class NDE(nd.NonDynamic):
    n = 0
    def __init__(self, n, m):
        self.n = n
        NonDynamic.__init__(self)

nde = NDE(1, 2)
```
gives
`AttributeError: You cannot add attributes to <__main__.NDE;  >`

I understand the SWIG test-suite currently checks to make sure that exactly this use actually **fails**, and you are trying to "fix" that, but this is rather crazy. At least [this entry in an old version of the Python cookbook](https://books.google.de/books?id=1Shx_VXS6ioC&pg=PT275&lpg=PT275&dq=python+restricting+dynamism&source=bl&ots=BD723F_9I1&sig=p4oStCXGhY7DQiPR-HXFv7K53_4&hl=de&sa=X&ved=0ahUKEwid_qbG6K7NAhXGORQKHT7gAnEQ6AEIHjAA#v=onepage&q=python%20restricting%20dynamism&f=false), which seems somewhat in line with the original idea of this "feature", mentions this case as an *example use*.

On another note, it seems, this "nondynamic" crazyness has been removed from the current versions of the Python cookbook altogether, and for a good reason. It simply does not make much sense.

m7thon:
Sorry, I tested your wrong branch. This PR does indeed seem to "fix" the broken case I mention under (1). The problem with case (2), however, remains.

m7thon:
Another note:

I suggest the code to support "nondynamic" Python classes should *only* be printed to the python shadow module if the "nondynamic" feature is actually being used.

The point is that anyone giving SWIG a first try for python wrapping, and having a peek at the generated wrapper file, will be completely stumped by this wizardry and might figure that SWIG wrapping for Python is just insanely complex and hacky.

mromberg:
> The problem with case (2), however, remains.

  Believe it or not, this _is_ working as designed.   There is/was even a specific test for this already in the swig test suite.  Slamming the door on derived classes adding attributes to the derived class itself seems to be the whole purpose of using the meta class.  

  The metaclass sticks to anything and everything that inherits from it like stink on  a pig and enforces the same constraints (no additional attributes).  So, if a class has the nondynamic flag set on it then that is "it" for attributes for itself and all classes that extend it.  The set of attributes from the original class are the only attributes any of them will ever have :).

  I'm not defending this.  I'm describing how it works.   I too think the feature should be removed.  However removing it will be a bit tricky as there are bits of it all over the python swig module.  Heck, I think normal operation of attribute set/get all pass through _swig_setattr_nondynamic.  Extraction will be somewhat like brain surgery in that we will have to be careful not to kill the patient in the process.

  So, it does make sense to me to get the whole test suite working with the -O flag (including this feature) and then pulling it (and other) features out.   Then one might be able to catch regressions from feature extraction.

m7thon:
"Fixing" this feature first, and then removing it altogether is an ok way to proceed. I don't really like the newly added complexity, though, but ok, if it will be removed again soon, I guess it's ok. However, you could fix this trivially by simply removing the nondynamic testcase from the test-suite...

> However removing it will be a bit tricky as there are bits of it all over the python swig module. Heck, I think normal operation of attribute set/get all pass through _swig_setattr_nondynamic. Extraction will be somewhat like brain surgery in that we will have to be careful not to kill the patient in the process.

What a mess... 😦

wsfulton:
The runtime tests should not be using assert as the code within the assert statement may be completely ignored making the testcase useless. I see a few have crept into the tests, I'll fix those.

mromberg:
Sorry about the asserts.  I was unaware that -O disabled them.  I will avoid their use from now on.

wsfulton:
I thought I'd have a look as I think we should understand it before thinking about blindly removing it. I'm posting what I've found to date. 

The testcase mentions:
https://github.com/swig/swig/blob/f8ffc04215b1be9ade91d418c0e0cf309e1809a2/Examples/test-suite/python_nondynamic.i#L28-L33
so refers to an ActiveState recipe to freeze Python classes.

This test was added in 74c523675dd6d7a836a914b1d496e55a8f7d06f5 for swig-1.3.32 and did indeed work with -modern back then. I ran git bisect using -modern and found 0cdf8658c72f21fb734122a08bb55308cc100e12 was the first bad commit. This commit was added between swig-2.0.3 and swig-2.0.4. The log https://github.com/swig/swig/commits/0cdf865 shows a couple of nondynamic changes shortly before a merge and was added by @szager during the development of builtin classes. 76e7d01c4ecf2ca0d616f0343ea91d6386ee8ab7 mentions a mailing list discussion and these are the relevant posting that I found:
https://sourceforge.net/p/swig/mailman/message/27149081/
https://sourceforge.net/p/swig/mailman/message/26787772/
So from what I can make out, %pythonnondynamic broke for -modern during the development of builtin classes and/or during an attempt to improve %pythonnondynamic for derived classes.

Issue #710, opened 2016-06-14
go.cxx: Use of variable after 'free'


avalluri:
In function oneClassDirectorMethod() of go.cxx, The 'Wrapper *dummy' has used after it's been freed(DelWrapper).

```
Wrapper *dummy = NewWrapper();
emit_attach_parmmaps(parms, dummy);
DelWrapper(dummy);

Swig_typemap_attach_parms("godirectorin", parms, w);
Swig_typemap_attach_parms("goin", parms, dummy);
Swig_typemap_attach_parms("goargout", parms, dummy);
```

wsfulton:
@avalluri could you put a fix for this as well as all your other recent issues into a pull request?

avalluri:
@wsfulton , Sure I work on that.

Issue #716, opened 2016-06-15
[ccache-swig] Possible file descriptor leaks


avalluri:
In function `stats_update_size()` in `stats.c` opened file descriptor `fd(=safe_open(stats_file)` not closed while return.
```
 /* open safely to try to prevent symlink races */
  fd = safe_open(stats_file);

  /* still can't get it? don't bother ... */
  if (fd == -1) return;

  memset(counters, 0, sizeof(counters));

  if (lock_fd(fd) != 0) return;
```

In function `create_cachedirtag()` in `utils.c` open file descriptor `f` leaks on `fwrite()` failure case:
```
f = fopen(filename, "w");
  if (!f) goto error;
  if (fwrite(CACHEDIR_TAG, sizeof(CACHEDIR_TAG)-1, 1, f) != 1) {
    goto error;
  }
  if (fclose(f)) goto error;
success:
  free(filename);
  return 0;
error:
  free(filename);
  return 1;
```

In function `copy_fd()` in `utils.c` file descriptor `gz_in` leaked:
```
void copy_fd(int fd_in, int fd_out) {
	char buf[10240];
	int n;
	gzFile gz_in;

	gz_in = gzdopen(dup(fd_in), "rb");

	if (!gz_in) {
		fatal("Failed to copy fd");
	}

	while ((n = gzread(gz_in, buf, sizeof(buf))) > 0) {
		if (write(fd_out, buf, n) != n) {
			fatal("Failed to copy fd");
		}
	}
}
```
In function `unify_hash()` in `unify.c` open file descriptor `fd` leaks in error case.
```
fd = open(fname, O_RDONLY|O_BINARY);
  if (fd == -1 || fstat(fd, &st) != 0) {
    cc_log("Failed to open preprocessor output %s\n", fname);
    stats_update(STATS_PREPROCESSOR);
    return -1;
  }

  /* we use mmap() to make it easy to handle arbitrarily long
           lines in preprocessor output. I have seen lines of over
           100k in length, so this is well worth it */
  map = mmap(NULL, st.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
  if (map == (char *)-1) {
    cc_log("Failed to mmap %s\n", fname);
    stats_update(STATS_PREPROCESSOR);
    return -1;
  }
  close(fd);
```

Issue #715, opened 2016-06-15
[ccache-swig] Possible memory leaks


avalluri:
Found below memory leaks when run static code analysis:

- In function `failed()` in ccache.c memory returned from `find_executable()` has not been freed.
```
 if ((e=getenv("CCACHE_PREFIX"))) {
    char *p = find_executable(e, MYNAME);
    if (!p) {
      cc_log("could not find executable (%s)\n", e);
      perror(e);
      exit(1);
    }
    args_add_prefix(orig_args, p);
  }
```
* In function `process_args()` memory allocated for `default_depfile_name` not freed:
```
if (!dependency_filename_specified) {
      char *default_depfile_name = x_strdup(output_file);
      char *p = strrchr(default_depfile_name, '.');

      if (p) {
        if (strlen(p) < 2) {
          cc_log("badly formed dependency file %s\n", output_file);
          stats_update(STATS_ARGS);
          failed();
          return;
        }
        *p = 0;
      }
      else  {
        int len = p - default_depfile_name;

        p = x_malloc(len + 3);
        strncpy(default_depfile_name, p, len - 1);
        free(default_depfile_name);
        default_depfile_name = p;
      }

      strcat(default_depfile_name, ".d");
      args_add(stripped_args, "-MF");
      args_add(stripped_args, default_depfile_name);
    }
```
* In function `find_compiler()` memory allocated for `char *base` has not been freed.
```
 if ((path=getenv("CCACHE_CC"))) {
    base = x_strdup(path);
  }

  orig_args->argv[0] = find_executable(base, MYNAME);

  /* can't find the compiler! */
  if (!orig_args->argv[0]) {
    stats_update(STATS_COMPILER);
    cc_log("could not find compiler (%s)\n", base);
    perror(base);
    exit(1);
  }
```

* Similarly, in function `find_executable()` in `execute.c` the memory allocated for `char* path` has not been freed.
```
  ....
  path = x_strdup(path);

  /* search the path looking for the first compiler of the right name
     that isn't us */
  for (tok=strtok(path,":"); tok; tok = strtok(NULL, ":")) {
    char *fname;
    x_asprintf(&fname, "%s/%s", tok, name);
    /* look for a normal executable file */
       .....
      }

      /* found it! */
      free(path);
      return fname;
    }
    free(fname);
  }

  return NULL;
```
* In function `find_hash()` in `ccache.c` memory leaked returned by `str_basename()`
  ```
  if (st.st_nlink > 1) {
[...]

Issue #713, opened 2016-06-15
[include.c] Possible file descripotr leak 


avalluri:
In function `Swig_insert_file()`
```
int Swig_insert_file(const_String_or_char_ptr filename, File *outfile) {
  char buffer[4096];
  int nbytes;
  FILE *f = Swig_include_open(filename);

  if (!f)
    return -1;
  while ((nbytes = Read(f, buffer, 4096)) > 0) {
    Write(outfile, buffer, nbytes);
  }

  return 0;
}
```

file opened by `Swig_include_open()` was not closed before return.

PR #720, opened 2016-06-17
Fix memory leaks found in static analysis

avalluri:
Fixes issues #710, #712 , #713, #715, #716


wsfulton:
Thanks for the fixes. Out of interest, which static analysis tool did you use?

PR #721, opened 2016-06-17
Fix -python -O autodoc

mromberg:
The -O option is replacing class methods with instancemethod objects.  Which is great for performance.  But the doc strings attached to the functions do not match what the tests expect.  So, this patch detects when these methods have been switched like this and skips the tests.

  One could of course attempt to adjust the tests to match strings for the swig generated instancemethods.  But then each tests would now have three expected results depending upon what kind of callable is being used:

  1) Normal shadow methods.
  2) "Normal" builtin methods
  3) Instancemethod objects added with -O

  The tests already handle cases 1 and 2.  I thought it best not to continue the madness and just skip the tests.  After swig-3.1, I hope that python just uses one kind of callable for the wrappers (optimized instancemethods seem like a good choice).  Then there is just one type of string to test for autodoc.

### DIFF
--- a/Examples/test-suite/autodoc.i
+++ b/Examples/test-suite/autodoc.i
@@ -139,5 +139,11 @@ bool is_python_builtin() { return true; }
 #else
 bool is_python_builtin() { return false; }
 #endif
+
+#ifdef SWIGPYTHON_FASTPROXY
+bool is_python_fastproxy() { return true; }
+#else
+bool is_python_fastproxy() { return false; }
+#endif
 %}
   
--- a/Examples/test-suite/python/autodoc_runme.py
+++ b/Examples/test-suite/python/autodoc_runme.py
@@ -1,5 +1,5 @@
 from autodoc import *
-import sys, types
+import sys
 
 
 def check(got, expected, expected_builtin=None, skip=False):
@@ -15,17 +15,6 @@ def check(got, expected, expected_builtin=None, skip=False):
 def is_new_style_class(cls):
     return hasattr(cls, "__class__")
 
-def is_builtin_method(fn):
-    # Instance method enabled by -O overwrite the function
-    # with an PyInstanceMethod.  This type is not exported
-    # to python.  This check is for "normal" instance
-    # methods (as yes dodgy using im_func).  We will
-    # use this test to skip checks for instance methods.
-    if hasattr(fn, "__func__"):
-        return type(fn.im_func) == types.BuiltinMethodType
-    return True
-
-
 if not is_new_style_class(A):
     # Missing static methods make this hard to test... skip if -classic is
     # used!
@@ -38,12 +27,10 @@ check(A.__doc__, "Proxy of C++ A class.", "::A")
 check(A.funk.__doc__, "just a string.")
 check(A.func0.__doc__,
       "func0(self, arg2, hello) -> int",
-      "func0(arg2, hello) -> int",
-      skip=is_builtin_method(A.func0))
+      "func0(arg2, hello) -> int")
 check(A.func1.__doc__,
       "func1(A self, short arg2, Tuple hello) -> int",
-      "func1(short arg2, Tuple hello) -> int",
-      skip=is_builtin_method(A.func1))
+      "func1(short arg2, Tuple hello) -> int")
 check(A.func2.__doc__,
       "\n"
       "        func2(self, arg2, hello) -> int\n"
@@ -62,8 +49,8 @@ check(A.func2.__doc__,
       "arg2: short\n"
       "hello: int tuple[2]\n"
       "\n"
-      "",
-      skip=is_builtin_method(A.func2))
+      ""
+      )
 check(A.func3.__doc__,
       "\n"
       "        func3(A self, short arg2, Tuple hello) -> int\n"
@@ -82,8 +69,8 @@ check(A.func3.__doc__,
       "arg2: short\n"
       "hello: int tuple[2]\n"
       "\n"
-      "",
-      skip=is_builtin_method(A.func2))
+      ""
+      )
 
 check(A.func0default.__doc__,
       "\n"
@@ -93,8 +80,8 @@ check(A.func0default.__doc__,
       "\n"
       "func0default(e, arg3, hello, f=2) -> int\n"
       "func0default(e, arg3, hello) -> int\n"
-      "",
-      skip=is_builtin_method(A.func0default))
+      ""
+      )
 check(A.func1default.__doc__,
       "\n"
       "        func1default(A self, A e, short arg3, Tuple hello, double f=2) -> int\n"
@@ -103,8 +90,8 @@ check(A.func1default.__doc__,
       "\n"
       "func1default(A e, short arg3, Tuple hello, double f=2) -> int\n"
       "func1default(A e, short arg3, Tuple hello) -> int\n"
-      "",
-      skip=is_builtin_method(A.func1default))
+      ""
+      )
 check(A.func2default.__doc__,
       "\n"
       "        func2default(self, e, arg3, hello, f=2) -> int\n"
@@ -143,8 +130,8 @@ check(A.func2default.__doc__,
       "arg3: short\n"
       "hello: int tuple[2]\n"
       "\n"
-      "",
-      skip=is_builtin_method(A.func2default))
+      ""
+      )
 check(A.func3default.__doc__,
       "\n"
       "        func3default(A self, A e, short arg3, Tuple hello, double f=2) -> int\n"
@@ -183,8 +170,8 @@ check(A.func3default.__doc__,
       "arg3: short\n"
       "hello: int tuple[2]\n"
       "\n"
-      "",
-      skip=is_builtin_method(A.func3default))
+      ""
+      )
 
 check(A.func0static.__doc__,
       "\n"
@@ -352,8 +339,7 @@ check(F.__init__.__doc__,
 
 check(B.funk.__doc__,
       "funk(B self, int c, int d) -> int",
-      "funk(int c, int d) -> int",
-      skip=is_builtin_method(B.funk))
+      "funk(int c, int d) -> int")
 check(funk.__doc__, "funk(A e, short arg2, int c, int d) -> int")
 check(funkdefaults.__doc__,
       "\n"
--- a/Examples/test-suite/python/autodoc_runme.py
+++ b/Examples/test-suite/python/autodoc_runme.py
@@ -3,7 +3,9 @@ import sys
 
 
 def check(got, expected, expected_builtin=None, skip=False):
-    if not skip:
+    # The fastproxy feature uses instancemethod objects that
+    # have new autodoc strings not covered in these tests.
+    if not skip and (not is_python_fastproxy()):
         expect = expected
         if is_python_builtin() and expected_builtin != None:
             expect = expected_builtin
--- a/Examples/test-suite/python/autodoc_runme.py
+++ b/Examples/test-suite/python/autodoc_runme.py
@@ -1,5 +1,5 @@
 from autodoc import *
-import sys
+import sys, types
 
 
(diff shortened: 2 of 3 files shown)


m7thon:
I have not looked at this, but have a -- maybe stupid -- question:

The way a method is wrapped should not affect how the method is used in Python, right? So, shouldn't the docstring be exactly the same in all three cases?

mromberg:
One would think the docs should be the same.  But this is harder to do than one would first think.  What makes it tricky is two points.

  First the doc strings can be attached at two points.  Either on the python methods on the shadow class.  This is done as as "normal" doc string.  But swig also wraps C functions.  And when those functions are exposed to python a doc string is supplied.  So, there are two methods used to generate doc strings (I guess I've not looked closely but the evidence is there for their existance).  One makes doc strings for the pure python methods and the other makes doc string for the C/builtin methods.

  My first guess at tackling this failing test was to try simply copying the "python" doc string from the shadow class method to the builtin method.  This leads to the second problem.  Python itself makes docstrings immutable (or nondynamic to bring up a sore topic :).   So, one can't modify the doc string of the C method after a static char * has been passed to python.

  I'm also think there is a need for the two types of doc strings to support free standing type functions in the swig wrappers (no shadow type module wrapper).   So, doc strings for these type methods reflect what the underlying C call looks like and not the more python looking wrapped method.

  Again, reducing the sheer number of ways swig can generate code for python wrappers would go a long long way in simplifying things like autodoc.  Features like autodoc and pretty much any other feature have way to many combinations of wrapper code to have to deal with.  

  Get code working with one combination of -python flags.  Then add one more flag and boom!  Broken again.  If the test suite were run with every permutation of the python flags I bet it may take years for one run (possible exageration :).

Mike

mromberg:
> The way a method is wrapped should not affect how the method is used in Python, right?

  To specifically answer your question, no.  It can varry.

class X {
    static void foo(void);
};

  Can be used from python as:

A.foo(self)

  or

A_foo(self)

  The __doc__ differs between the two methods becaus the actual name of the function is different (just to name one thing).

  The optimizer replaces the call to A.foo with A_foo.  Which I assume bypasses dict lookups for the method.

m7thon:
Ah, thanks, now I understand *what* is happening. But it still is an issue, isn't it? The *correct* behavior is that `module.A.foo?` should always give the same autodoc function signature, regardless which of the type 1) through 3) it is, no? Of course, this is a rather minor issue...

One way to solve this would be to always assign the same docstring to `_module.A_foo` and `module.A.foo` to begin with. I wonder why that is not done.

Is it possible to move parts of tests that are not really fixed to a `currently_broken` folder instead of just disabling them altogether? Then the Travis CI will run, but it is clear that there are still issues.

mromberg:
I'm not really sure exactly why each method type is documented in a different way.  I suspect it is due to 36.4.1 Proxy classes:

http://www.swig.org/Doc3.0/Python.html#Python_nn28

  So, there really are two different callables generated by swig.  The differ at least in name.  And the low level wrapper can take a "self" pointer for class methods where the proxy method does not.  So, they may differ by the number of arguments too.  So, what name should they use if they share the exact same doc string?   They still need to be called with different names.  Self I think may not confuse anyone.  So, that is not a problem.

  I agree that this should be fixed in a better manner.  But the cancer is already there.  The existing autodoc tests were already skipping over built-in type one methods or provided a second doc string to use for the check.  So, fixing just the test for -O would require a third string to compare in the test or finding some way at runtime to disable the test.  This is what I did.

  I also agree that the "right way" is to have one type of autodoc string that is consistently used.  I think doing this makes more sense in swig-3.1 where (with luck and work) the possible python output of swig is reduced considerably.   Then one doc string would be a lot more manageable.   Trying to do this now with all the differing method types seems like it would be more work.  Especially if some of those methods are made to go away.

m7thon:
I don't think the three different wrapping options can go, since they offer distinct features. If you want to customize methods using e.g., `%pythonprepend`, or rewrite the wrapper with `%feature("shadow")`, you need to fall back to the "Normal" proxy methods, and these need to call the C wrapper method.

I guess one could disallow the instancemethods, which call the C method directly, but its a rather simple optimization.

Or do you want to get rid of builtin, which is elegant and fast?

vadz:
I'm all but completely resigned to spend the rest of my life in merging doxygen branch code with master anyhow but, just as a reminder, this will result in more conflicts with the changes there... It would be really great if autodoc-related stuff could be left untouched until doxygen branch is merged.

mromberg:
How does the doxygen branch deal with this issue of the three different python signatures?

vadz:
It doesn't deal with it in any way, it tries (and hopefully succeeds) to preserve exactly the same behaviour as in the mainline. But it does add handling of Doxygen comments to the same code.

FWIW I do think that autodoc code needs to be changed. I just don't want it to be changed _now_.

mromberg:
> FWIW I do think that autodoc code needs to be changed. I just don't want it to be changed now.

  Just to clarify, this patch only modifies the autodoc tests to check for a case that is not currently considered.  It simply checks the function type and skips tests for method types the tests never did handle.  In no way is anything else related to autodoc touched.

vadz:
Yes, this test is also modified in the doxygen branch however.

wsfulton:
Agreed autodoc needs to be improved in/after doxygen branch is merged. Hence, skipping these tests for now is okay. @vadz, sorry that this will result in one more merge conflict, I don't think it'll be too hard to resolve, maybe @mromberg can help as he authored the changes? For now I'm keen to get -O green in preparation for all the swig-3.1 changes.

wsfulton:
This test is skipping more than I think you intended. I added this patch to it:
```patch
diff --git a/Examples/test-suite/python/autodoc_runme.py b/Examples/test-suite/python/autodoc_runme.py
index a46fb60..8c660dc 100644
--- a/Examples/test-suite/python/autodoc_runme.py
+++ b/Examples/test-suite/python/autodoc_runme.py
@@ -15,7 +15,7 @@ def check(got, expected, expected_builtin=None, skip=False):
 def is_new_style_class(cls):
     return hasattr(cls, "__class__")
 
-def is_builtin_method(fn):
+def is_builtin_method_impl(fn):
     # Instance method enabled by -O overwrite the function
     # with an PyInstanceMethod.  This type is not exported
     # to python.  This check is for "normal" instance
@@ -25,6 +25,11 @@ def is_builtin_method(fn):
         return type(fn.im_func) == types.BuiltinMethodType
     return True
 
+def is_builtin_method(fn):
+    is_builtin = is_builtin_method_impl(fn)
+    if is_builtin:
+        print("Changed to skip: {}".format(fn))
+    return is_builtin
 
 if not is_new_style_class(A):
     # Missing static methods make this hard to test... skip if -classic is
```
The 'Changed to skip' messages appear when setting PY3=1 or SWIG_FEATURES=-builtin. I'd expect to see this message only when setting SWIG_FEATURES=-O.

I'm also confused why the method is called ```is_builtin_method``` when the comments say it detects when -O is being used not when -builtin is used. The doc comment for ```is_builtin_method``` needs to be clearer, I don't really understand it. 

If we are going to use the 'skip' as a named parameter, let's do it consistently in the test and perhaps change the line 'skip=True' to 'fully_skip=True' and make appropriate changes.

This is a really annoying test, but let's not make it skip what it already tests.

mromberg:
> I'm also confused why the method is called is_builtin_method when the comments say it detects when -O is being used not when -builtin is used.

  Yea.  This is tricky.  The -O switch enables the use of these things:

https://docs.python.org/3/c-api/method.html

  These are wrapper type different from the "normal" SWIG -builtin methods.  And autodoc spits out different output for them.  Detecting that this has happened after the fact is difficult because I know of no way to access any of the SWIG flags that were used to generate the code after SWIG has run.  In addition the above docs mention the following.

  Note that it says:  "Python instance method type. It is not exposed to Python programs.". 

  It is very possible that my test for this new type of python object also picks up the "normal" SWIG -builtin types.  There is no good way to tell them apart from python itself.  But autodoc saw them differently and created different docs.  I'll look into it some more.

  But unless SWIG saves things like flags and features somewhere the scripting language can use I'm not sure any runtime test can accurately tell the difference between these two other than comparing the string generated by autodoc.

  I'll clean up the names you suggest.

m7thon:
> I'm not sure any runtime test can accurately tell the difference between these two

Can't you simply look at the string `A.foo.__repr__()`? This is:

* `"<method 'foo' of 'module.A' objects>"` when using `-builtin`
* `"... A_foo ..."` when using `-O` without `-builtin`
* `"... A.foo ..."` for the "normal" case

The exact representation differs a little depending on the Python version, but I believe you can use this for differentiating the cases.

mromberg:
__repr__ is a novel idea!   However the bit about it differing with python versions bothers me.   I think I may have to deal with this the exact same way -builtin itself does and scatter some crumbs in the forest on the way to grandma's house.

  There is a function already injected into the shadow wrapper called is_python_builtin().  I'm assuming that the -builtin code itself does this in order to deal with this exact issue.  It is kinda/very ugly but the alternatives are even worse.  

  I could have the code that generates instance method objects (when -O is used) do this too.  Another function/variable in the shadow wrapper for use by this test only called something like is_instance_method_object().   I don't exactly like it, but is_python_builtin() is already there for the same purpose (making the autodoc test work).

  On this subject it would be nice if there was some way to store flags such as this somewhere in the generated wrappers.  It could be even stashed away somewhere in the _foo.so module in a data structure like a dict with a name along the lines of _swig_options.  There swig could store things like what command line options, environment variables and other things that were used to generate the wrapper.  Then hacks such as this would not need to be exposed (much).

mromberg:
I've reverted the commit that attempted to determine the method type at runtime.  I don't think it is easy or even possible from the non C side of python.

  This new commit skips autodoc tests when -fastproxy is enabled.   It is very similar to how -builtin is doing things.  And it touches only a few lines of the autodoc test code.

wsfulton:
SWIGPYTHON_BUILTIN is defined in the preprocessor because it is needed for -builtin. I'm not keen on adding ad-hoc preprocessor definitions just for testing. On the other hand, I wouldn't mind having some sort of unpublished way to find obtain the command line options SWIG was invoked with for testing purposes. However, I've found what I think is a simple approach to detecting -fastproxy, just by looking at the differences in the output of ```dir()``` when -fastproxy is on or off. I propose this patch instead:

```patch
diff --git a/Examples/test-suite/python/autodoc_runme.py b/Examples/test-suite/python/autodoc_runme.py
index ce0aae0..af04c8c 100644
--- a/Examples/test-suite/python/autodoc_runme.py
+++ b/Examples/test-suite/python/autodoc_runme.py
@@ -15,11 +15,19 @@ def check(got, expected, expected_builtin=None, skip=False):
 def is_new_style_class(cls):
     return hasattr(cls, "__class__")
 
+def is_fastproxy(module):
+    return "new_instancemethod" in module
+
 if not is_new_style_class(A):
     # Missing static methods make this hard to test... skip if -classic is
     # used!
     sys.exit(0)
 
+if is_fastproxy(dir()):
+    # Detect when -fastproxy is specified and skip test as it changes the function names making it
+    # hard to test... skip until the number of options are reduced in SWIG-3.1 and autodoc is improve
+    sys.exit(0)
+
 # skip builtin check - the autodoc is missing, but it probably should not be
 skip = True
```
Any comments or improvements?

m7thon:
@wsfulton: That will do nicely :-). Generally, I think any symbol like `new_instancemethod` that gets added to the module namespace by SWIG should be named `_swig_...` though.

wsfulton:
Yes, it definitely should be prefixed with swig. A patch earmarked for swig-3.1 for this will be nice.

m7thon:
Ok, I'll just make a PR. I'll rename any SWIG symbols added to `_swig_...` while I'm at it.

wsfulton:
pep7 suggests an underscore prefix just for internal implementation attributes. I suspect you'll need a mix of using underscore and not underscore. I don't think you should rename public attributes like thisown though.

PR #728, opened 2016-06-23
Fixing python docstring handling for -fastproxy

ahnolds:
Following up on #721 for some time after the doxygen branch is merged.  This standardizes the docstrings generated in the -fastproxy case (and mostly in the -builtin case as well).  The required changes were:
- Saving the unmodified `sym:name` and `params` attributes before they are modified and checking for the original versions when building documentation
- New `memberfunction` flag to help distinguish between methods and functions when generating documentation in the C files
- Standardized indentation


vadz:
This is another PR which will result in non-trivial conflicts with the doxygen branch...

wsfulton:
Hi @ahnolds. This patch does not actually test -fastproxy because the patch does not prevent the entire test being skipped when run with -fastproxy. This is still in the testcase:
https://github.com/swig/swig/blob/d832d666040cd8d899d4d3b5080f9c412eb98e59/Examples/test-suite/python/autodoc_runme.py#L26-L30

I can see that there are improvements which will work with -builtin so the patch is good in that respect.

Now that the Doxygen branch has been merged, I'd like to incorporate this patch. I also re-implemented -fastproxy, see #1310. It doesn't seem to have changed the autodoc behaviour in master. 

@ahnolds, would you be able to update the patch?

Just as an observation, I noticed that autodoc combined with -fastproxy emits two definitions of each method, eg in the test case, the generated code was:
```python
class A(_object):
    ...
    def funk(self, a):
        """just a string."""
        return _autodoc.A_funk(self, a)
    ...
A.funk = new_instancemethod(_autodoc.A_funk, None, A)
```
but after #1310 is:
```python
class A(_object):
    ...
    def funk(self, a):
        r"""just a string."""
        return _autodoc.A_funk(self, a)
    ...
    funk = _swig_new_instance_method(_autodoc.A_funk)
    ...
```
I'm not sure this is the optimal way of adding documentation to a method for this 'fast' optimization.

ahnolds:
Rebased onto master with that autodoc check removed - I'm just going to through this onto CI and see what is broken at the moment since I haven't looked at this in ~2 years.

If stuff passes I can maybe take a look at the double definitions later.

ahnolds:
Looks like this version is still going to only be partially correct (although it may pass CI) - static function docstrings are still are slightly off in the generated C code. More tweaks will follow.

@wsfulton would you object to just removing the python wrapper functions entirely if `-fastproxy` is enabled (except for if `%pythonappend` or similar are used)? The documentation will then just be from the C code (so the source is less readable, but the documentation is present for users, and we already know `-fastrproxy` makes the python code much less readable).

ahnolds:
Should pass and be clean now - I would consider this ready for merging (which of course will curse the CI to fail :stuck_out_tongue: : )

ahnolds:
CI passed, good to merge from my perspective

wsfulton:
I checked the differences in the output from the python test-suite after running:
```
make clean-python-test-suite partialcheck-python-test-suite
```
and can see some changes which look like bugs that have been introduced. Here are a some:

```
diff -Naur /home/william/swig/partialgoodts/python/cpp11_default_delete.py Examples/test-suite/python/cpp11_default_delete.py
--- /home/william/swig/partialgoodts/python/cpp11_default_delete.py	2018-12-20 08:47:11.527697403 +0000
+++ Examples/test-suite/python/cpp11_default_delete.py	2018-12-20 08:50:51.799300558 +0000
@@ -159,7 +159,7 @@
     __repr__ = _swig_repr
 
     def __init__(self, arg2):
-        _cpp11_default_delete.sometype_swiginit(self, _cpp11_default_delete.new_sometype(arg2))
+        _cpp11_default_delete.sometype_swiginit(self, _cpp11_default_delete.new_sometype(arg1))

diff -Naur /home/william/swig/partialgoodts/python/cpp11_null_pointer_constant_wrap.cxx Examples/test-suite/python/cpp11_null_pointer_constant_wrap.cxx
--- /home/william/swig/partialgoodts/python/cpp11_null_pointer_constant_wrap.cxx	2018-12-20 08:47:11.527697403 +0000
+++ Examples/test-suite/python/cpp11_null_pointer_constant_wrap.cxx	2018-12-20 08:50:52.347299536 +0000
@@ -2927,7 +2927,7 @@
 	 { "new_A", _wrap_new_A, METH_NOARGS, NULL},
 	 { "A__myA_set", _wrap_A__myA_set, METH_VARARGS, NULL},
 	 { "A__myA_get", _wrap_A__myA_get, METH_O, NULL},
-	 { "A_NullPtrMethod", _wrap_A_NullPtrMethod, METH_VARARGS, "A_NullPtrMethod(A self, double * ptr=None)"},
+	 { "A_NullPtrMethod", _wrap_A_NullPtrMethod, METH_VARARGS, "NullPtrMethod(A self, double * ptr=None)"},
``` 
Additional method below definition, not sure if this should be there or not, probably not.
```
diff -Naur /home/william/swig/partialgoodts/python/default_args.py Examples/test-suite/python/default_args.py
--- /home/william/swig/partialgoodts/python/default_args.py	2018-12-20 08:47:11.467697507 +0000
+++ Examples/test-suite/python/default_args.py	2018-12-20 08:50:38.283325723 +0000
@@ -506,6 +506,7 @@
 def slightly_off_square(*args):
     r"""slightly_off_square(def17) -> int"""
     return _default_args.slightly_off_square(*args)
+slightly_off_square = _default_args.slightly_off_square

diff -Naur /home/william/swig/partialgoodts/python/doxygen_alias.py Examples/test-suite/python/doxygen_alias.py
--- /home/william/swig/partialgoodts/python/doxygen_alias.py	2018-12-20 08:47:11.515697424 +0000
+++ Examples/test-suite/python/doxygen_alias.py	2018-12-20 08:50:49.003305772 +0000
@@ -111,5 +111,6 @@
     :return: A new object which may be None.
     """
     return _doxygen_alias.make_something()
+make_something = _doxygen_alias.make_something

```
This is an unwanted whitespace change:
```
diff -Naur /home/william/swig/partialgoodts/python/cpp11_null_pointer_constant.py Examples/test-suite/python/cpp11_null_pointer_constant.py
--- /home/william/swig/partialgoodts/python/cpp11_null_pointer_constant.py	2018-12-20 08:47:11.527697403 +0000
+++ Examples/test-suite/python/cpp11_null_pointer_constant.py	2018-12-20 08:50:52.347299536 +0000
@@ -101,7 +101,6 @@
     def NullPtrMethod(self, *args):
         r"""NullPtrMethod(A self, double * ptr=None)"""
         return _cpp11_null_pointer_constant.A_NullPtrMethod(self, *args)
-
     __swig_destroy__ = _cpp11_null_pointer_constant.delete_A
```
Please address these regressions.

wsfulton:
Note that global functions are also now included with `-fastproxy`, see #639 and I'd expect the patch to also have some effect on the docstrings in global function wrappers.

wsfulton:
Hi @ahnolds. I've got this earmarked for SWIG 4.0.0 due out soon (maybe within a month). Will you be able to address these issues by then or should I target a later version?

ahnolds:
I'll see if I can resolve these shortly, depends on my schedule

ahnolds:
1.1. Fixed - anonymous argument numbering is a pain to get right
1.2. Not a bug - if anything, the new name in the docstring is more accurate without the leading "classname_"
2. Fixed - got this for free by merging in the changes for #639
3. Won't fix here - this is a symptom of an already existing bug (see li_std_map.py for an example). While it would maybe be possible to fix, it would require a much more widespread set of whitespace changes. The change in behavior comes from docstrings no longer being considered as add-to-function items, so rather than getting the single newline at [line 4560](https://github.com/swig/swig/blob/master/Source/Modules/python.cxx#L4560), we get the double newline at [line 4574](https://github.com/swig/swig/blob/master/Source/Modules/python.cxx#L4574).seem

Let me know if that seems reasonable to you @wsfulton

ahnolds:
I did also work up some changes that standardize the spacing more rigorously, and I'm happy to submit a separate pull request for that. I'm not really sure that part is 100% ready for primetime though, so I'd rather just get this merged and handle that separately since that part is entirely cosmetic and this part is functional.

wsfulton:
It isn't mentioned in the issue or commit message anywhere, but the patch also better documents member variables that are wrapped, via property(doc="..."). This should be added to the CHANGES.current file.

I've been checking the effects of this patch using the default commandline options, that is, without -fastproxy. There are few good changes. I do see regressions in the generated _wrap.cxx files though in the low-level module, where functions can be called in a flattened API. It is the only API when using -noproxy. The documentation on this API is now quite inconsistent in not showing self and getting the documented function name wrong.
```patch
--- python/autodoc_wrap.cxx	2019-01-16 08:48:25.167675377 +0000
+++ Examples/test-suite/python/autodoc_wrap.cxx	2019-01-16 08:59:32.512645045 +0000

-		"A_variable_d_set(A self, int variable_d)\n"
+		"A_variable_d_set(int variable_d)\n"

-		"A_variable_d_get(A self) -> int\n"
+		"A_variable_d_get() -> int\n"

-	 { "B_funk", _wrap_B_funk, METH_VARARGS, "B_funk(B self, int c, int d) -> int"},
-	 { "delete_B", _wrap_delete_B, METH_O, "delete_B(B self)"},
+	 { "B_funk", _wrap_B_funk, METH_VARARGS, "funk(B self, int c, int d) -> int"},
+	 { "delete_B", _wrap_delete_B, METH_O, "delete_B()"},

-	 { "TInteger_inout", _wrap_TInteger_inout, METH_VARARGS, "TInteger_inout(TInteger self, TInteger t) -> TInteger"},
+	 { "TInteger_inout", _wrap_TInteger_inout, METH_VARARGS, "inout(TInteger self, TInteger t) -> TInteger"},
```
> 1.2. Not a bug - if anything, the new name in the docstring is more accurate without the leading "classname_"

Similarly, this is not correct as this is for the low-level module api.

All these different variations are tricky, I know!

ahnolds:
Didn't even know about the `-noproxy` option, that's new since I last worked on swig! In an effort to cover all my bases this time, here are the combinations of flags I'm planning to test with/compare to their equivalents from trunk:

- nothing
- `-builtin`
- `-fastproxy`
- `-fastproxy -olddefs`
- `-noproxy`

Can you think of any other flags/combinations I should test with? And do you think it makes sense to add a build for either of the last two to Travis (I think the others are already effectively there)?

wsfulton:
The `-olddefs` option is not very important, is very simple and merely additionally generates the proxy function definitions generated by the default option when `-fastproxy` is used. So testing of `-olddefs` is basically covered by the default option. 

The `-noproxy` option was introduced in swig-1.3.14 (Aug 2002). Before then no proxy definitions were generated by default and the `-proxy`, a.k.a. `-shadow`, option needed to be used to get the current output. We can't test `-noproxy` easily because most of the runtime tests in the test-suite require proxy classes to be generated. Basically, all that `-noproxy` does is not generate the .py file containing the proxy code. You can effectively check that `-noproxy` is okay by using the default options and making sure there are no regressions in all the generated `_wrap.cxx` files in the test-suite. You can programmatically check this by adding in runtime tests to check the documentation on the low-level module when using the default options.

Regarding the remaining options, they are tested on Travis (`-fastproxy` is included as part of `-O`).

ahnolds:
Having looked at this more, I'm now kind of stuck. In the `-fastproxy` case, the functions for the low-level API (e.g. A_NullPtrMethod) are just re-used for the normal "shadow" API. That means the docstring will be the same for both the low-level function and the shadow method. I see a few options, none of which I love:

- Accept that the docstring name will be wrong on the shadow method in the fastproxy case
- Accept that the docstring name will be wrong on the low-level function in all non-builtin cases
- Have two entries in the C module function table with the same underlying function, just two different docstrings (e.g. A_NullPtrMethod with the "A_NullPtrMethod()" docstring and _A_NulllPtrMethod with the "NullPtrMethod()" docstring) and have the shadow class use the relevant one as its method. This gets all the docstrings right but feels kind of kludgy.

Any thoughts, or other approaches I haven't come up with?

wsfulton:
This is not easy. I'd be inclined to have consistency between -builtin/-fastproxy and the default. That means that the low-level API will not have the correct names. However, when using -noproxy, autodoc must still give the correct names. Can you give an example output of the full output using help(C.m) and help(C_m) where C is a wrapped class and m is a wrapped method if we go with what I suggest? Also compare the output when autodoc is off.

wsfulton:
Note that I verified that the additions in this patch also fixes autodoc for properties/member variables - #1367.

ahnolds:
Might have found a way around this - I think I can copy the Python object representing the function (I think it should be a `PyCFunctionObject`) at runtime and modify the docstring for the method at the C level before calling `PyInstanceMethod_New` (or py2 equivalent). I'll have to pass the docstring in to `SWIG_PyInstanceMethod_New` as an extra argument, but that seems like a small price to pay since `-fastproxy` already results in the generated python looking funky.

I'll give it a shot later and report back.

ahnolds:
Making progress, mostly there - I still have a few things to iron out for static functions and variables, but hopefully that's it. I should get it done sometime this weekend.

ahnolds:
This should do it baring unforseen CI failures. I will note that the last set of changes make it harder to manually diff the full set of generated code, so I wrote [a script to "clean"](https://gist.github.com/ahnolds/77ffe86778c03d500d70e2910be02b3c) the new files to match the old files as much as possible, so I would suggest using that if you again want to diff the new and old versions.

I have, by the way, benchmarked this with all combinations of the followinlg:

- Python 2.7 and 3.6
- `-fastproxy`, `-builtin`, and default flags
- static methods, member methods without counting object instantiation time, and member methods with counting object instantiation time

In all cases, it performs basically identically to the existing code in terms of runtime.

I think at least theoretically there should probably be a tiny amount of extra time required to load the module in the default flags and `-fastproxy` cases, since there is now a call to the new `getProxyDoc` to find the correct high-level docstring and a call to `PyCFunction_NewEx` to create an extra Python C-API function object for member functions and static functions and the time to construct. That said, the total time for loading three such functions in a couple manual trials appears to be about 10 microseconds, which is tiny compared to the module load time as a whole, so I think it's ok even if for some crazy reason someone decides to load a swig-generated module in a tight inner loop.

ahnolds:
Whoops this is what I get for being overly confident in my code :man_shrugging:

ahnolds:
Just FYI this is now all good on my end and CI is passing - let me know if there's anything else outstanding I need to address.

wsfulton:
I'll commit the merge of this shortly. However, there are a couple of minor issues that need addressing. 

1. The native_directive.i test was added recently, and with this patch -fastproxy is always on for %native functions.
2. When using -noproxy, there is some type information missing, here's an example of a diff in autodoc_wrap.cxx:
```diff
-"A_func3(A self, short arg3, Tuple hello) -> int\n"
+"A_func3( self, short arg2, Tuple hello) -> int\n"
```
3. The docstring for `__dict__` when using -builtin has changed:
```diff
abstract_access_wrap.cxx:
-    { (char *)"__dict__", (getter)SwigPyBuiltin_FunpackGetterClosure, (setter)0, (char *)"A.__dict__", (void *)&A___dict___getset }
+    { (char *)"__dict__", (getter)SwigPyBuiltin_FunpackGetterClosure, (setter)0, (char *)"", (void *)&A___dict___getset }

autodoc_wrap.cxx:
-    { (char *)"__dict__", (getter)SwigPyBuiltin_FunpackGetterClosure, (setter)0, (char *)"A.__dict__", (void *)&A___dict___getset }
+    { (char *)"__dict__", (getter)SwigPyBuiltin_FunpackGetterClosure, (setter)0, (char *)"A : p.A", (void *)&A___dict___getset }
```
@ahnolds, could you fix these in a separate patch please? You were going to do a line spacing tidy up or have I recalled that incorrectly?

wsfulton:
Oh yes, I meant to mention that one of the checks I used was to use pydoc, eg in the Examples/test-suite/python directory:
```
make -s autodoc.cpptest && pydoc autodoc
```
to show all the documentation for the autodoc test.  Piping the output into files and comparing the output using -fastproxy, -builtin and the default option was interesting. The main problem with -builtin is missing docstrings for `__init__`. Any suggestions for fixing this?

ahnolds:
I'll look into the above. Would you prefer separate pull requests for the spacing tidy up and the straggling autodoc stuff, or all in one is fine?

And if I recall correctly from when I last looked at the `__init__` thing, I don't think there is a way to explicitly document the initializer for a builtin type, but I'll take another look.

ahnolds:
I can't reproduce the %native issue. If I run `make -s native_directive.cpptest`, the output ends with
```python
def CountAlphas(instring):
    return _native_directive.CountAlphas(instring)
```

Is that not what I expect to see in the normal (non `-fastproxy`) case? And are %native functions even relevant for python? I don't see any mention of it in the docs.

wsfulton:
An all in one patch is fine in this case (but with separate commits).

Actually, I'm not entirely sure how %native ought to work for Python given the lack of docs and examples. I think I'd need to go way back in time to experiment to see what used to work. Right now `-fastproxy` has no effect if a %native directive is used, eg:
```swig
%native(CountAlphaCharacters) PyObject *_wrap_CountAlphaCharacters(PyObject *self, PyObject *args);
```
But on second thought that is probably okay. Anyway, just leave it as it is for now, as the output in the generated .py file is effectively the same as in swig-3.0.12 with or without `-fastproxy`.

PR #730, opened 2016-06-23
Alternative fix for the implicittest problems

ahnolds:
As I commented in #711, the real problem is that the fix @wsfulton made for handling implicit conversions in b58dabced9e7424e6107bfd40963806f3b831a19 only applied to the regular (non `-fastdispatch` non `-castmode`) case.  This patch extends those changes to the other dispatch cases.

The result is that in all overload cases where there are multiple possibilities with the same number of arguments, the dispatch function will first check for exact (aka non implicit) matches, and then subsequently check for implicit casting matches.  This was already happening in the normal dispatch situation, and in the `-fastdispatch` case two passes through the candidates were happening, just with `SWIG_POINTER_IMPLICIT_CONV` always set.  After this patch, it is not set on the first pass, and then set on the second pass.


wsfulton:
Thanks @ahnolds for spotting and fixing these shortcomings of my original patch.

Issue #733, opened 2016-06-27
[C#] Wrong inheritance modifier since SWIG 3.0.9: CS0114


Cpt-Kork:
> SWIG 3.0.9: Improved C++ multiple inheritance support for Java/C# wrappers

Changes introduced with SWIG 3.0.9 appear to break inheritance support when working with modules, 3.0.8 works just fine.

C++: Project **B** and its SWIG interface file:

``` c++
// module: B
// file: /path/to/B/Base.h
class Base {
public:
    virtual void Calculate();
};
```

```
// module: B
// file: /path/to/B/swig.i
%module B
%include "/path/to/B/Base.h"
```

C++: Project **C** and its SWIG interface file:

``` c++
// module: C
// file: /path/to/C/Class.h
class Class : Base {
public:
//    virtual void Calculate(); // not overridden here
};
```

```
// module: C
// file: /path/to/C/swig.i
%module C
%import " /path/to/B/swig.i"
%include "/path/to/C/Class.h"
```

C++: Project **D** and its SWIG interface file:

``` c++
// module: D
// file: /path/to/D/Derived.h
class Derived : Class {
public:
    virtual void Calculate(); // but here
};
```

```
// module: D
// file: /path/to/D/swig.i
%module D
// %import " /path/to/B/swig.i" // <- does not help
%import " /path/to/C/swig.i"
%include "/path/to/D/Derived.h"
```

C# generated files:

``` c#
// module B
public class Base : IDisposable {
    public virtual void Calculate() { /*...*/ }
}
```

``` c#
// module C
public class Class : Base, IDisposable {
//    public virtual void Calculate() { /*...*/ } // <- no Calculate, since not in C++
}
```

Broken inheritance modifier since 3.0.9:

``` c#
// module D
public class Derived : Class, IDisposable {
    public virtual void Calculate() { /*...*/ } // Since SWIG 3.0.9: CS0114: 'Derived.Calculate()' hides inherited member 'Base.Calculate()'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword.
}
```

Whereas, everything's fine in 3.0.8:

``` c#
// module D
public class Derived : Class, IDisposable {
    public override void Calculate() { /*...*/ } // SWIG 3.0.8: OK
}
```

Cpt-Kork:
[...]

PR #734, opened 2016-06-27
#733 - wrong "override" calculation in import mode

wkalinin:
"override" calculation in allocate.cxx should take into account interfaces.
"direct_base" attribute has no other use now, so it may be removed completely (it had some intended usage, but I can't remember what exactly)

### DIFF
--- a/Examples/test-suite/imports_a.h
+++ b/Examples/test-suite/imports_a.h
@@ -18,4 +18,12 @@ class A {
   virtual MemberEnum member_virtual_test(MemberEnum e) { return e; }
   virtual GlobalEnum global_virtual_test(GlobalEnum e) { return global_test(e); }
 };
+
+/* This class overrides nothing. Inherited classes should see base functions.
+*/
+class A_Intermediate : public A { 
+ public:
+  A_Intermediate(){}
+  ~A_Intermediate(){}
+};
 #endif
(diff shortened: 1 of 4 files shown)


wsfulton:
Could you please put in a testcase for this?

wkalinin:
Looks like there are zero tests of import mode in the test suite.

wsfulton:
I can see these:
```
~/swig/github/swig/Examples/test-suite (master=)$ grep "%import" -l *
clientdata_prop_b.i
go_subdir_import_a.i
import_nomodule.i
imports_a.i
imports_b.i
import_stl_b.i
mod_b.i
multi_import_a.i
packageoption_c.i
template_typedef_import.i
```
Note that most of these are run using ```make xx.multicpptest``` instead of ```make xx.cpptest```. Maybe you mean some other sort of import?

wkalinin:
Oh, indeed there are. Somehow my search for "#import" gave me no results, probably I misspelled it.
As I understand, test case should be a C# runtime one? I'll try to make one taking some other test as a model, but it will probably take some iterations, because I was unable to set up C# test suite working

wsfulton:
C# should work out the box if you install the appropriate packages:
```
sudo apt-get install mono-devel #ubuntu/debian
zypper install mono-core #suse
sudo dnf install mono-core #fedora
```
SWIG has ```%import``` and C# does not have a ```#import```.

Do you have any emails/other info on the problem, I presume it is is related to the new %interface support we added recently.

wkalinin:
I referenced the issue in the commit comment, it is: https://github.com/swig/swig/issues/733

PR #736, opened 2016-06-30
Added support to $typemap for looking up attribute values.

johnb003:
For example in you might have:
 %typemap(imtype, out="IntPtr") ctype "HandleRef";
and you can now access the out attribute with $typemap(imtype:out, ctype).

### DIFF
--- a/Source/Swig/typemap.c
+++ b/Source/Swig/typemap.c
@@ -1967,8 +1967,9 @@ static void replace_embedded_typemap(String *s, ParmList *parm_sublist, Wrapper 
 	    String *parm = Getitem(l, i);
 	    char *eq = strchr(Char(parm), '=');
 	    char *c = Char(parm);
-	    if (eq && (eq - c > 0)) {
-	      String *name = NewStringWithSize(c, (int)(eq - c));
+	    int size = (int)(eq - c);
+	    if (eq && size > 0) {
+	      String *name = NewStringWithSize(c, size);
 	      String *value = NewString(eq + 1);
 	      Insert(name, 0, "$");
 	      Setattr(vars, name, value);
@@ -1986,8 +1987,9 @@ static void replace_embedded_typemap(String *s, ParmList *parm_sublist, Wrapper 
 	{
 	  char *eq = strchr(Char(tmap_method), ':');
 	  char *c = Char(tmap_method);
-	  if (eq && (eq - c > 0)) {
-	    tmap_method = NewStringWithSize(c, (int)(eq - c));
+	  int size = (int)(eq - c);
+	  if (eq && size > 0) {
+	    tmap_method = NewStringWithSize(c, size);
 	    /* tmap_method memory is owned by a list, but since we're overriding it,
 	     * keep track of this so we can clean up. */
 	    tmap_method_with_attr = tmap_method;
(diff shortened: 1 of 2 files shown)


johnb003:
Do I need to do something to get people to look at this?

wsfulton:
This looks okay to me. It just needs a test for it to be accepted. I suggest enhancing the two main tests that already test ```$typemap```, that is, special_variable_macros.i and special_variable_attributes.i and include appropriate changes to the Java and C# runme files:
```
~/swig/github/swig/Examples/test-suite (master=)$ find . -name "special_variable*_runme.*"
./go/special_variable_macros_runme.go
./csharp/special_variable_macros_runme.cs
./csharp/special_variable_attributes_runme.cs
./d/special_variable_macros_runme.2.d
./d/special_variable_macros_runme.1.d
./python/special_variable_macros_runme.py
./java/special_variable_macros_runme.java
./java/special_variables_runme.java
```

wsfulton:
Also a mention in the docs in Typemap.html would be helpful.

Source/Swig/typemap.c
```diff
@@ -1967,8 +1967,9 @@ static void replace_embedded_typemap(String *s, ParmList *parm_sublist, Wrapper
 	    String *parm = Getitem(l, i);
 	    char *eq = strchr(Char(parm), '=');
 	    char *c = Char(parm);
-	    if (eq && (eq - c > 0)) {
-	      String *name = NewStringWithSize(c, (int)(eq - c));
+	    int size = (int)(eq - c);
+	    if (eq && size > 0) {
+	      String *name = NewStringWithSize(c, size);
```
ojwb: This change seems to introduce undefined behaviour in the case where `eq == NULL`.

johnb003: I'm not seeing the problem, can you elaborate?
size is now computed with an "undefined" eq, but it's not a crash, and eq is checked for null before comparing size and protects all of the remaining code in the block.

vadz: You can't use null pointer in pointer arithmetics, this is undefined behaviour whatever you do or don't do with the result later. Pointer arithmetics is only defined for pointers inside the same array.

ojwb:
Indeed - and if a program invokes undefined behaviour, anything could happen:

http://c-faq.com/ansi/undef.html

wsfulton:
@johnb003 do you still need this patch applied?

johnb003:
Yes. Sorry it fell off of my radar. I'll make the suggested changes.

wsfulton:
ping @johnb003 - the pending swig-4.0 release would be a good time to add this patch if you can make those changes.

ojwb:
ping @johnb003 again - this missed 4.0 but could still make 4.1.  Needs testcases and the undefined behaviour issue fixing.

ojwb:
> ping @johnb003 again - this missed 4.0 but could still make 4.1. Needs testcases and the undefined behaviour issue fixing.

Make that **docs**, testcases and the undefined behaviour issue fixing.

johnb003:
I no longer have any interest.
Finish it however you like or abandon it.

Regards,
John

ojwb:
Thanks for letting us know.

I'm going to close this then - it's a new feature rather than a bug fix, it's an interesting idea but I don't currently see any compelling use cases for it, and frankly we're drowning in open tickets.

However if someone actually wants to pick this up and complete it, please don't be put off by it getting closed.

PR #738, opened 2016-07-01
In preprocessor, replace multiple macro arguments "at once" via placeholders

### LINKED ISSUE #737: Macro misexpansion when replacement contains another macro parameter
Consider the following, rather simple, code:
`#define TAG( a, b ) a | b
enum Bug { BUG = TAG( 'b', 'x' ) };`
(extracted from freetype headers)

Should work fine, but fails with the cryptic "Error: Empty character constant".

What happens? If you look at [Source/Preprocessor/cpp.c:928](https://github.com/swig/swig/blob/36bbeb478dfcecc7739284cd454ada8b4cb96333/Source/Preprocessor/cpp.c#L928), you'll note that parameter expansion occurs one at a time.

The correct behavior by C standard would be to form a list of tokens, then replace each token that matches a macro parameter with its replacement string.

FeepingCreature:
This avoids an issue where argument replacement with a string that contains the name of another argument can cause arguments to be replaced in text that wasn't meant to contain macro arguments.

Example:
    #define TAG( a, b ) a | b
    enum Bug { BUG = TAG( 'b', 'x' ) };

Ie. `a | b` becomes `'b' | b`, then `''x'' | 'x'`. Desired was `'b' | 'x'`.

The patch instead goes `\006 0 \007 | \006 1 \007`, then replaces `\006 0 \007` with `'b'` and `\006 1 \007` with `'x'`.

Fixes #737.

### DIFF
--- a/Examples/test-suite/preproc_macro_arg_order.i
+++ b/Examples/test-suite/preproc_macro_arg_order.i
@@ -1,0 +1,2 @@
+#define TAG( a, b ) a | b
+enum Bug { BUG = TAG( 'b', 'x' ) };
--- a/Source/Preprocessor/cpp.c
+++ b/Source/Preprocessor/cpp.c
@@ -828,13 +828,25 @@ static String *expand_macro(String *name, List *args, String *line_file) {
   temp = NewStringEmpty();
   tempa = NewStringEmpty();
   if (args && margs) {
+    /* instead of replacing the args one by one, build a list of markers
+     * of the form \006[num]\007, then replace them all at the end
+     * this solves issues where the expansion of one arg contains another arg.
+     */
+    String **list_markers;
+    String **list_repargs;
     l = Len(margs);
+    list_markers = malloc(sizeof(String*) * l);
+    list_repargs = malloc(sizeof(String*) * l);
     for (i = 0; i < l; i++) {
       DOH *arg, *aname;
-      String *reparg;
+      String *reparg, *amarker;
       arg = Getitem(args, i);	/* Get an argument value */
       reparg = Preprocessor_replace(arg);
       aname = Getitem(margs, i);	/* Get macro argument name */
+      amarker = NewStringEmpty();
+      Printf(amarker, "\006%i\007", i);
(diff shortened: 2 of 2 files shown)


wsfulton:
I suspect this is okay. @ojwb would you mind taking a close look as you are probably most familiar with this area.

FeepingCreature:
Ah, do note that the testcase I added is kind of cosmetic at the moment because I had no idea how to actually add testcases to be run.

ojwb:
Not sure I'm really the most familiar with this code, but have taken a look.

While the patch addresses the provided testcase, this slightly modified version is still mishandled:

```
$ cat test.i
#define TAG( a, b ) a | b | 'b'
enum Bug { BUG = TAG( 'A', 'B' ) };
$ swig -E test.i|grep BUG
enum Bug { BUG = 'A' | 'B' | ''B'' };
```

Same output with or without the changes in this PR - a real C preprocessor doesn't expand the `'b'`:

```
$ cpp test.i|grep BUG
enum Bug { BUG = 'A' | 'B' | 'b' };
```

Also this is mishandled (using SWIG patched as per this PR):

```
$ cat test.i
#define TAG( a, b ) #a[b]
int BUG = TAG( b, 0 );
$ swig -E test.i|grep BUG
int BUG = "0"[0];
```

C preprocessor gives:

```
$ cpp -E test.i|grep BUG
int BUG = "b"[0];
```

From what I can see, the current expansion of preprocessor macros in SWIG is fundamentally flawed - it expands macro arguments by a simple text substitution (it checks that it only replaces a whole identifier, but doesn't consider quotes) which just isn't how the C preprocessor actually works.

I haven't checked the C or C++ standards for the preprocessor spec, but it looks to me like the preprocessor needs to do a basic tokenisation to identify what to expand, and then expand tokens found in that stream, which would get all these cases correct, and shouldn't require these new numbered placeholders (nor I suspect the existing placeholders).

ojwb:
The approach seems the wrong way forwards to me and there's been no progress here in over a year, so closing.  #737 remains for the issue itself.

Issue #656, opened 2016-04-12
Non-template base class treated incorrectly if in different namespace and has same name as derived template class


fschlimb:
This issue involves templates, namespaces and inheritance. The issue is about inheriting from a base-class with the same name but in in a different namespace. Additionally, the derived class is a template class while the parent class is not. 

I could boil it down to the following very simple reproducer (using 2 modules to get around the naming issue with flattened namespaces):

File ta.i:

```
%module ta

namespace NA
{
    struct Parameter {};
}
// eof ta.i
```

File tt.i:

```
%module tt
%import(module="ta") "ta.i"
namespace NB
{
    template <typename T>
        struct Parameter : public NA::Parameter {};
    %template(Pi) Parameter<int>;
}
// eof tt.i
```

SWIG complains about 

```
tt.i:1: Warning 401: Nothing known about base class 'NA::Parameter< int >'. Ignored.
tt.i:1: Warning 401: Maybe you forgot to instantiate 'NA::Parameter< int >' using %template.
```

The resulting Python wrapper defines “Parameter” in tt.py but it doesn’t inherit from ta.Parameter, which it should do.

Note1: SWIG incorrectly thinks NA::Parameter is a template, but in fact it is not!
Note2: Strangely SWIG reports the error in line 1.

Observation1: Renaming NA::Parameter (to “Para”, say) makes SWIG happy and the resulting code is correct.
Observation2: If NB::Parameter is not a template, all works fine, too.
Observation3: deriving from “::NA::Parameter” issues an even stranger message: 
`tt.i:1: Warning 401: Nothing known about base class ''. Ignored.”`

I'd be willing to invest some time in fixing it myself but would need some hints as I am not familiar with the SWIG code.

Any idea and/or help?

wsfulton:
I verified that this is fixed by fa00622614d35716b429b12aa9dc1bde8e4b586d, see #1484.

PR #742, opened 2016-07-07
do not resolve full template declarations

fschlimb:
The symbol resolution is only needed for (partial) template specializations. This is because they can have template parameters and can be defined in a different scope/namespace.

However, a normal (full) template declaration can never have its home anywhere else than in the current scope/namespace. For such normal (full), first-time template declarations the resolution will incorrectly find classes if such exist with the same name but in different namespace.

I tested this with my own very complex, templated code which is full of namespaces. Unfortunately I have no small reproducer.

### DIFF
--- a/Examples/test-suite/duplicate_class_name_in_ns.i
+++ b/Examples/test-suite/duplicate_class_name_in_ns.i
@@ -1,0 +1,21 @@
+%module a;
+
+namespace A
+{
+    class Foo {
+    public:
+        Foo(){};
+    };
+}
+
+namespace B
+{
+    template<typename T, typename U>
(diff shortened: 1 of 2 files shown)


wsfulton:
I'm afraid that we need a cutdown testcase to accept this and understand it fully. I'd have thought you could distill this down to something minimal given you have code that this fixes. I suggest you run ```swig -E``` if your code is complex and edit it down to something understandable.

fschlimb:
Ok, I added a reproducer. SWIG will issue a warning. This is related to #656.

fschlimb:
covered by #1115

PR #744, opened 2016-07-10
Added support for guile's native pointer type

briancaine:
I believe the guile exit currently represents swig pointers as [tagged smobs](https://www.gnu.org/software/guile/manual/html_node/Smobs.html). Guile also has a [native pointer type](https://www.gnu.org/software/guile/manual/html_node/Void-Pointers-and-Byte-Access.html#Void-Pointers-and-Byte-Access) that works with strings, numeric arrays and the low-level FFI.

The guile exit only accepts swig pointers right now. This patch changes it to accept native pointers as well.

### DIFF
--- a/Doc/Manual/Guile.html
+++ b/Doc/Manual/Guile.html
@@ -30,6 +30,7 @@
 <li><a href="#Guile_nn14">Smobs</a>
 <li><a href="#Guile_nn15">Garbage Collection</a>
 </ul>
+<li><a href="#Guile_nn12">Native Guile pointers</a>
 <li><a href="#Guile_nn16">Exception Handling</a>
 <li><a href="#Guile_nn17">Procedure documentation</a>
 <li><a href="#Guile_nn18">Procedures with setters</a>
@@ -453,13 +454,14 @@ is exactly like described in <a href="Customization.html#Customization_ownership
 Object ownership and %newobject</a> in the SWIG manual.  All typemaps use an $owner var, and
 the guile module replaces $owner with 0 or 1 depending on feature:new.</p>
 
-<H2><a name="Guile_nn12">24.7 Native Guile pointers</a></H2>
+<H2><a name="Guile_nn12">24.8 Native Guile pointers</a></H2>
+
 
 <p>
 In addition to SWIG smob pointers, <a href="https://www.gnu.org/software/guile/manual/html_node/Foreign-Pointers.html">Guile's native pointer type</a> are accepted as arguments to wrapped SWIG functions. This can be useful for passing <a href="https://www.gnu.org/software/guile/manual/html_node/Void-Pointers-and-Byte-Access.html#">pointers to bytevector data</a> to wrapped functions.
 </p>
 
-<H2><a name="Guile_nn16">24.8 Exception Handling</a></H2>
+<H2><a name="Guile_nn16">24.9 Exception Handling</a></H2>
 
 
 <p>
@@ -485,7 +487,7 @@ mapping:
 The default when not specified here is to use "swig-error".
 See Lib/exception.i for details.
 
-<H2><a name="Guile_nn17">24.9 Procedure documentation</a></H2>
+<H2><a name="Guile_nn17">24.10 Procedure documentation</a></H2>
 
 
 <p>If invoked with the command-line option <code>-procdoc
@@ -520,7 +522,7 @@ like this:
 typemap argument <code>doc</code>. See <code>Lib/guile/typemaps.i</code> for
 details.
 
-<H2><a name="Guile_nn18">24.10 Procedures with setters</a></H2>
+<H2><a name="Guile_nn18">24.11 Procedures with setters</a></H2>
 
 
 <p>For global variables, SWIG creates a single wrapper procedure
@@ -548,7 +550,7 @@ struct members, the procedures <code>(<var>struct</var>-<var>member</var>-get
 pointer)</code> and <code>(<var>struct-member</var>-set pointer
 value)</code> are <em>not</em> generated.
 
-<H2><a name="Guile_nn19">24.11 GOOPS Proxy Classes</a></H2>
+<H2><a name="Guile_nn19">24.12 GOOPS Proxy Classes</a></H2>
 
 
 <p>SWIG can also generate classes and generic functions for use with
@@ -694,7 +696,7 @@ Notice that &lt;Foo&gt; is used before it is defined.  The fix is to just put th
 <code>%import "foo.h"</code> before the <code>%inline</code> block.
 </p>
 
-<H3><a name="Guile_nn20">24.11.1 Naming Issues</a></H3>
+<H3><a name="Guile_nn20">24.12.1 Naming Issues</a></H3>
 
 
 <p>As you can see in the example above, there are potential naming conflicts.  The default exported
@@ -731,7 +733,7 @@ guile-modules.  For example,</p>
 (use-modules ((Test) #:renamer (symbol-prefix-proc 'goops:)))
 </pre></div>
 
-<H3><a name="Guile_nn21">24.11.2 Linking</a></H3>
+<H3><a name="Guile_nn21">24.12.2 Linking</a></H3>
 
 
 <p>The guile-modules generated above all need to be linked together.  GOOPS support requires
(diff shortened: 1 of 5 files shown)


Lib/guile/guile_scm_run.swg
```diff
@@ -176,6 +176,9 @@ SWIG_Guile_ConvertPtr(SCM s, void **result, swig_type_info *type, int flags)
   if (SCM_NULLP(smob)) {
     *result = NULL;
     return SWIG_OK;
+  } if (SCM_POINTER_P(s)) {
```
ojwb: I'd suggest an `else` before the `if` here.

While it doesn't actually make any difference as the code stands (because there preceding block ends with a `return`), having `} if` seems a trap for future developers if that `return` is removed, and `else if` would be consistent with the code below what you've inserted.

ojwb:
I don't think there's currently an active maintainer for guile as a target language in SWIG, so you'll need to excuse some probably dumb questions.

This change doesn't seem to change the code SWIG generates to set pointers using this native pointer type, only to accept pointers already in that type.  So how is that actually useful by itself?  It seems people would have to convert C/C++ pointers to guile behind SWIG's back, which seems an odd use case.

Lib/guile/guile_scm_run.swg
briancaine: ​​Ah, missed that. Should I fix it and create a new pull request?

briancaine:
With the guile pointer type, I can pass bytevectors to SWIG generated code like this:

```
(use-modules (srfi srfi-4) (system foreign))

(define some-u32-data (u32vector 3 4 5 6))
(some-swig-function (bytevector->pointer some-u32-data))
```

ojwb:
> With the guile pointer type, I can pass bytevectors to SWIG generated code like this:

Ah I see, makes sense.  It would be good to have a testcase for this feature then, and a note in the Guile chapter of the manual explaining that it is supported.

[missing else]
> Ah, missed that. Should I fix it and create a new pull request?

You can just push more commits to the same branch and they'll get added to the PR (github even supports "Squash and merge" via the web UI now).

briancaine:
Added the else, documentation and test.

Doc/Manual/Guile.html
```diff
@@ -30,6 +30,7 @@
 <li><a href="#Guile_nn14">Smobs</a>
 <li><a href="#Guile_nn15">Garbage Collection</a>
 </ul>
+<li><a href="#Guile_nn12">Native Guile pointers</a>
```
ojwb: There's already a `#Guile_nn12` in this file, so adding a second isn't right.

I'm not sure what we normally do when inserting a section - renumbering the existing anchors seems unhelpful, as it'll break external bookmarks.  @wsfulton What's the preferred way to handle this?

Doc/Manual/Guile.html
```diff
@@ -454,14 +453,13 @@
 Object ownership and %newobject</a> in the SWIG manual.  All typemaps use an $owner var, and
 the guile module replaces $owner with 0 or 1 depending on feature:new.</p>
 
-<H2><a name="Guile_nn12">24.8 Native Guile pointers</a></H2>
-
+<H2><a name="Guile_nn12">24.7 Native Guile pointers</a></H2>
```
ojwb: There's now two sections number `24.7`.

briancaine: Yeah... I saw that and the other section issues you noted. The documentation stuff (indices, section numbers) is maintained partially by scripts. I just figured it was generated elsewhere.

But of course, now that I think about it, that wouldn't really explain why the section numbers are up-to-date in the repository.

I regenerated the section numbers and everything and committed them.

Sorry about that.

briancaine:
Updated the changes file

Doc/Manual/Guile.html
ojwb: @wsfulton Ping re anchor numbering when inserting a section - how's this usually handled?

briancaine:
Heh, so, I gave this another stab and I think I finally got the anchors working properly.

So, if you have a section like this:
```
<H2><a>24.8 Native Guile pointers</a></H2>
```

Doing `make docs` in the base directory will generate TOCs in Doc/Manual/Guile.html and Doc/Manual/Contents.html.

It'll also turn that section line above into:
```
<H2><a name="Guile_nn16">24.8 Native Guile pointers</a></H2>
```

But the script isn't shy about reusing existing section anchor names. So if you add a section, to regenerate the TOCs properly, you have to edit every section after the new section to the original nameless anchor tag. Then when it goes through and inserts the `name="Guile_nnXX"` attributes, it'll number them in order properly.

I think this issue is fixed.

(Sorry about this mess.)

ojwb:
Thanks for the updates. If the renumbering of anchors is scripted, then that must be what we usually do in this case, so looks good to me.

ojwb:
Squashed and pushed as 391bb79cbaa914db2dbbe2edec5c2af1d76bed86, thanks.

wsfulton:
This patch has completely broken guile-1.8 support as SCM_POINTER_VALUE and SCM_POINTER_P do not seem to be available. 
```
[  190s] checking Examples/guile/constants
[  190s] example_wrap.c: In function 'SWIG_Guile_ConvertPtr':
[  190s] example_wrap.c:896: warning: assignment makes pointer from integer without a cast
[  190s] /tmp/cc7SJpHj.o: In function `SWIG_Guile_ConvertPtr':
[  190s] example_wrap.c:(.text+0xded): undefined reference to `SCM_POINTER_P'
[  190s] example_wrap.c:(.text+0xdfc): undefined reference to `SCM_POINTER_VALUE'
[  190s] collect2: ld returned 1 exit status
```
Full log for our SLES testing can be seen at https://build.opensuse.org/public/build/home:kwk:swig/SLE_11_SP3/i586/swig-raw/_log.

@briancaine, can you provide macros for older versions of guile that provide equivalent functionality? If not, then this enhancement should be compiled in for versions that don't support it and the test also skipped if not supported.

wsfulton:
Oh, and if you can look at this urgently, it would be appreciate as I had planned to release swig-3.0.12 today.

wsfulton:
The argout_runme.scm test you added also needs to be fixed to only run when this support is available. You could do a runtime check on the Guile version (if you can find out which version this was support was added into) or maybe you know of some other way? Maybe add a C function in the test to return true if the macros are defined.

One other thing, how about instead using
```
#if SCM_MAJOR_VERSION >= 2
```
or whatever version is appropriate in case these macros turn into real functions at some point in the future.

briancaine:
It doesn't seem like there's an analog to pointer objects in 1.8. So as you can see, I #ifdef'd out the code. I can change that #ifdef to the version one you propose tomorrow.

I rewrote the automated test to check for the relevant function (bytevector->pointer), and if the function doesn't exist, it just returns as a successful test.

I couldn't find a way to determine the Guile version from within Guile.

wsfulton:
Thanks for the quick response, I've committed a4ec4df5c993ed939db1b878aa82e6cd24b0b117 567f944129cd6213d6d98bc01bc5ab2998a2ea7b 892f3592cc81dccb8ffe5be7cfce16a4b70a340b

gjanssens:
This commit has negative side effects in gnucash (one consumer of the guile bindings in swig).
If the guile bindings are generated with guile 2.0, they no longer work with guile 1.8 as of this PR. This used to work before.

gjanssens:
By the  way the way to test for guile 2 from within guile is:

    (cond-expand
      (guile-2
        (your-guile-2-only-code-goes-here)
      (else)
        (your-guile-1-only-code-goes-here)
    )

gjanssens:
I can't reach the swig website right now, so I'll ask: is there a way to delay the expansion of #ifdef SCM_MAJOR_VERSION > 2 ? It seems to be interpreted now while generating the bindings files, so the generated swig-foo.c file doesn't have the conditional inclusion anymore.

The solution out of this would be to retain the conditional inclusion (#ifdef/#endif) in the generated swig-foo.c file. That way the the choice of inclusion can be delayed until the bindings files are compiled as part of the project that uses them.

ojwb:
```
%#ifdef SCM_MAJOR_VERSION > 2
```

The docs are in git in the source tree though (look in `Docs/Manual`).

gjanssens:
Thanks. I have dug a bit deeper and it turns out I reacted too fast. Swig 3.0.11 behaves as I described, but I see now this has been fixed already in swig 3.0.12.

My apologies for the noise.

PR #750, opened 2016-07-17
Use python-config if available in configure

vadz:
There is no need to check for Python prefix, headers and library ourselves if
we can just ask python-config to do it for us. This won't work for the systems
with very old Python versions, so still keep the existing Python detection
code for them, but use python-config if it's available as it's simpler and
more reliable.

As python-config doesn't return the library directory separately, remove PYLIB
(and PY3LIB, for consistency) and the corresponding make variables and use
just PYLINK which contains both the directory and the library itself instead.

---

Hopefully we can get rid of the manual Python detection code in the future.

### DIFF
--- a/Examples/Makefile.in
+++ b/Examples/Makefile.in
@@ -309,11 +309,9 @@ PYTHON_FLAGS =
 # Make sure these locate your Python installation
 ifeq (,$(PY3))
   PYTHON_INCLUDE= $(DEFS) @PYINCLUDE@
-  PYTHON_LIB    = @PYLIB@
   PYTHON        = @PYTHON@ $(PYTHON_FLAGS)
 else
   PYTHON_INCLUDE= $(DEFS) @PY3INCLUDE@
-  PYTHON_LIB    = @PY3LIB@
   PYTHON        = @PYTHON3@ $(PYTHON_FLAGS)
 endif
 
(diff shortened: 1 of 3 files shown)


vadz:
Sorry for the noise with the updates (I made a couple of errors because I didn't realize that we could be running under Cygwin but using Windows Python and Travis also finished with spurious errors a few times, so I had to relaunch it), but this seems to work now and I think it's worth merging even if it doesn't change anything much, other than speeding up configure a tiny bit and making its messages a bit more clear, because it simplifies configure logic a little and it really doesn't hurt, seeing how complicated it is now.

wsfulton:
This is a sensible move into the modern world! I'd like to test it in a pure Cygwin environment before accepting. Cygwin uses python-configure and it ought to work in this environment too, but from what I can see, python-configure is explicitly ignored on Cygwin and this is wrong. I think we ought to just change the way we configure pure windows python in the Cygwin env on Travis by removing:
```
- set PATH=C:\Python%VER%%LANG_PLATFORM%;%PATH%
```
and instead passing ```--with-python$(VER)=C:\Python%VER%%LANG_PLATFORM%``` to configure. The philosophy should be to use pure Cygwin (python/gcc) etc, unless otherwise explicitly configured (either by mucking about with the paths or configure options). Could you modify the patch to try this out?

A few days ago Appveyor made some changes that have nearly halved the build times down to about 100 mins. I think now would be a good time to extend our usage of Appveyor to add in pure Cygwin tests for something like Python and Ruby. Unless someone else volunteers, I'll add it to my TODO list.  I also have a Cygwin env set up so I can test the pull request, but don't have access for a another week as I'm away on hols.

vadz:
I am leaving myself soon and probably won't be able to do it until then...

I wonder if we shouldn't turn off `python-config` detection if `--with-python=whatever` is explicitly given. IMO there are two cases:

1. Automatic detection: we use Python and `python-config` that we find. This should work out of the box on all Unixish systems, including Cygwin.
1. Manual specification of everything: the user explicitly gives us Python and (a) also compile and link flags for it (this is, of course, not implemented now), or (b) we can discover those ourselves using the existing logic.

2(b) is only really useful under Windows (with Windows Python), but this is probably still a reason enough to keep it, unfortunately, even though it complicates things a lot. For Unix, 2(a) is possibly more useful, e.g. if you have some custom-compiled Python (requiring some extra compilation options). But (1) should undoubtedly be the most common use case under Unix.

vadz:
Revisiting my old PRs, I think this one should really be merged in, other improvements could be done later, of course, but applying it right now would be an improvement right now and doesn't seem to have any drawbacks.

ojwb:
It's a shame this never got merged, but given this is only about Python2 (we already support `python-config` for Python3) I wonder if we're at the point where it's better to just leave Python2 things alone until Python2 support is dropped?

The `PY3LIB`/`PY3LINK` clean-up is perhaps still worth doing as it makes the `python-config` and non-`python-config` cases more similar.

ojwb:
> [vadz reacted with confused emoji]

I'm not sure what your confusion here is, but I can try to restate more clearly.

I'm just suggesting that we're probably at the point where Python2 support is effectively in "maintenance mode".  While it would have been nice to have merged this when Python2 was still an active thing, this patch doesn't fix a bug as such, but carries (as almost every change does) some risk of introducing problems.

vadz:
> > [vadz reacted with confused emoji]
> 
> I'm not sure what your confusion here is,

My confusion is due to not recognizing the emojis properly, sorry (for confusion?), I thought it was a "sad" emoji and this referred to the "never got merged" part, which I also regret, but, as with many other patches, can't do anything about (and gave up any hope of ever being able to).

> but I can try to restate more clearly.

I do agree with everything you wrote, sorry again for trying to express myself in an unfamiliar language, I'll try to stick to the words next time.

ojwb:
Ah, I see - emoji considered harmful!

I certainly agree it's sad.  I'll apply the `PY3LIB`/`PY3LINK` clean-up and close then.

ojwb:
Actually, having looked at doing that I think I'm going to leave it alone.

We don't currently use either the `--libs` or `--ldflags` options to `python3-config` but instead dig out these values by other means so we in fact still have the split here whether or not we're using `python3-config`.  I suspect we don't **need** the split as it looks like we probably always use them together, but the machinery here is quite hard to follow and I don't know the history of most of it.

Cleaning up this probing code is probably worthwhile at some point (indeed I think similar looking at the configure probes for most of the target languages) but me trying to tease this out of someone else's larger change seems too likely to go wrong.

PR #756, opened 2016-07-30
Classes, Values, Pointers, References etc. with Examples

ng420:
Adds features to hhvm modules


ojwb:
I merged up to 6e8ba0b to the gsoc2016-hhvm branch in the swig/swig repo earlier today, as that was the latest that built with the hhvm version I'm using.

ojwb:
OK, merged.

Let's start a fresh PR when there are further changes ready to merge.

PR #763, opened 2016-08-01
Adds JavaScript Native Directive

TekuConcept:
In the Utah State RoboSub Team's Poseidon project, JavaScript uses socket callbacks. I want to replace sockets with swig wrapper objects to remove some of the unnecessary overhead and complexity in communicating with hardware.

While the current version of swig does not support callbacks, Node.h however does. Something like this can then be added by the user to their wrapper files:

```
#ifndef SWIG
v8::Persistent<v8::Function> r_call;
void SaveCallback(const v8::FunctionCallbackInfo<v8::Value>& args) {
    v8::Isolate* isolate = args.GetIsolate();
    if (args[0]->IsFunction()) {
        v8::Local<v8::Function> func = v8::Local<v8::Function>::Cast(args[0]);
        v8::Function * ptr = *func;
        r_call.Reset(isolate, func);
    }
    else {
        r_call.Reset();
    }
}
#endif
```

Unfortunately the current version of swig also doesn't support the `%native` directive for JavaScript so programmers need to hand-type the line `SWIGV8_AddMemberFunction(_exports_<module>_class, "<custom>", <custom>);` every time swig (re)generates the new <module>_wrap.cxx.

I played around with Python's native directive, treating swig like a black box to see what the command does and it appears it just appends the desired single line to the end of the member functions.
Therefore the code in this pull request is based on the code used for python's native implementation.

Other changes are simply conforming to code convention. For some reason cloning master broke many lines with indentation and attempts to revert the undesired changes failed.

### DIFF
--- a/Source/Modules/javascript.cxx
+++ b/Source/Modules/javascript.cxx
@@ -105,7 +105,7 @@ private:
 /**
  * JSEmitter represents an abstraction of javascript code generators
  * for different javascript engines.
- **/
+ */
 class JSEmitter {
 
 protected:
@@ -121,15 +121,15 @@ protected:
 
 public:
 
-   enum JSEngine {
-     JavascriptCore,
-     V8,
-     NodeJS
-   };
+  enum JSEngine {
+    JavascriptCore,
+    V8,
+    NodeJS
+  };
 
-   JSEmitter(JSEngine engine);
+  JSEmitter(JSEngine engine);
 
-   virtual ~ JSEmitter();
+  virtual ~ JSEmitter();
 
   /**
    * Opens output files and temporary output DOHs.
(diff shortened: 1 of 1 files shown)


wsfulton:
Could you remove all the whitespace changes from the patch, squash, rebase and resubmit the pull request? Whitespace changes in a separate pull request are okay if they don't make any code changes. 

Also, this part of the documentation needs amending:
https://github.com/swig/swig/blob/054a85c5463cb3ceee61e83ac4d5dcbca975b149/Doc/Manual/Javascript.html#L166

I've just added a new testcase for %native testing as we didn't have one. It's called native_directive, added in 50a80e6ee35fb0a45c17bf63afbbe450c9783174. Please add a native_directive_runme.js to show it working. As you had Python %native working and I don't know of any documentation for Python, would you also mind demonstrating how you got Python to work by adding in a runtime Python file for this testcase? That'll be a useful reference.

wsfulton:
Superceded by #1535.

PR #770, opened 2016-08-21
Add anon namespace.

LindleyF:
Wrap SwigDirector_* classes generated for Java in an anonymous namespace.

When two C++ classes live in different namespaces, and are mapped to Java classes in different packages, but otherwise have the same name, previously the SwigDirector_* classes would collide. By wrapping them in an anonymous namespace, we allow them to peacefully coexist so long as they're in different swig invocations.

### DIFF
--- a/Source/Modules/java.cxx
+++ b/Source/Modules/java.cxx
@@ -4638,6 +4638,7 @@ public:
 
     directorDeclaration(n);
 
+    Printf(f_directors_h, "namespace {\n");
     Printf(f_directors_h, "%s {\n", Getattr(n, "director:decl"));
     Printf(f_directors_h, "\npublic:\n");
     Printf(f_directors_h, "    void swig_connect_director(JNIEnv *jenv, jobject jself, jclass jcls, bool swig_mem_own, bool weak_global);\n");
@@ -4788,7 +4789,8 @@ public:
       Printf(f_directors_h, "    }\n");
     }
 
-    Printf(f_directors_h, "};\n\n");
+    Printf(f_directors_h, "};\n");
+    Printf(f_directors_h, "}  // namespace\n\n");
     Printf(w->code, "}\n");
     Printf(w->code, "}\n");
 


LindleyF:
The failure appears unrelated to this change.

wsfulton:
The director classes are generated into a header file so that they can be used from other code in a different translation unit. Putting them into an anonymous namespace will break code that is using the director classes from a different translation unit.

The ```nspace``` feature is designed to fix problems for classes with the same name in multiple namespaces. Can you try this please? We even have a simple director and %nspace test in director_nspace_director_name_collision.i. Hopefully this works okay for you, even though your usage is slightly different.

LindleyF:
The nspace feature appears to modify the Java package as well as tweaking the Director somehow. Modifying the Java package is undesirable.

Emitting the director class into the same namespace as the original C++ class, or one derived from it, would solve the problem as well. Is that what nspace does? Any reason it couldn't be triggered independently of the Java package adjustments?

ojwb:
With `%nspace` the director class for `TopLevel::A::Foo` is named `SwigDirector_TopLevel_A_Foo` instead of `SwigDirector_Foo`.

The only reason not to always use the `%nspace` naming that I can see is compatibility with what is currently generated.  I'm not sure if that matters, as I'm not sure what the scenarios are where you'd want to use the generated header file in a different translation unit.  If we're going to change this, 4.1.0 would be a good time to make sure a change.

_Aside: This naming doesn't seem totally ideal either.  `TopLevel::A::Foo` will collide with `TopLevel::A_Foo`, and also `namespace_::A` -> `namespace__A` which is a reserved identifier due to the double `_`.  The only way I can really see to avoid these issues would be essentially escape any existing `_` in the identifiers, kind of like how JNI does.  This does make for nasty looking identifiers in the common case where there isn't currently a problem, unless the escaping scheme is more complex but then it's harder to understand._

ojwb:
I don't think this is actually specific to Java as the naming of this class is essentially the same for any target language which supports directors, though looking at the code some backends appear to handle this themselves rather using `Language::directorClassName()` in `lang.cxx`.

ojwb:
("this" being the underlying issue - the patch here is specific to Java, but doesn't seem to be the right approach, and I think we'd want to address this for all the affected target languages.)

ojwb:
> With `%nspace` the director class for `TopLevel::A::Foo` is named `SwigDirector_TopLevel_A_Foo` instead of `SwigDirector_Foo`.
> 
> The only reason not to always use the `%nspace` naming that I can see is compatibility with what is currently generated. I'm not sure if that matters, as I'm not sure what the scenarios are where you'd want to use the generated header file in a different translation unit. If we're going to change this, 4.1.0 would be a good time to make sure a change.

It'd have to be 4.2.0 now...

While the patch to use an anonymous namespace isn't suitable, there's definitely an issue here as loading two modules which wrap the same C++ class name but in different namespaces will use the same C++ director class name (without `%nspace`), so loading them together can result in colliding symbols.

I had a quick look - there's `directorClassName()` in `Source/Modules/lang.cxx` which builds this class name.  We can't just use the `%nspace` code path always since that takes the namespace from the `sym:nspace` attribute which is only set in the `%nspace` case, but we could determine it from other attributes of the class node (e.g. `classtype` and `name` both have the C++ class name including the namespace).

It also looks like many of the target languages currently duplicate knowledge about how to build a director class name, as does `Source/Swig/cwrap.c`.  Ideally we'd have this encapsulated in a single place.  I wouldn't be surprised if there are currently bugs with `%nspace` and directors due to disagreements about the director class name.

ysx1993:
> > With `%nspace` the director class for `TopLevel::A::Foo` is named `SwigDirector_TopLevel_A_Foo` instead of `SwigDirector_Foo`.
> > The only reason not to always use the `%nspace` naming that I can see is compatibility with what is currently generated. I'm not sure if that matters, as I'm not sure what the scenarios are where you'd want to use the generated header file in a different translation unit. If we're going to change this, 4.1.0 would be a good time to make sure a change.
> 
> It'd have to be 4.2.0 now...
> 
> While the patch to use an anonymous namespace isn't suitable, there's definitely an issue here as loading two modules which wrap the same C++ class name but in different namespaces will use the same C++ director class name (without `%nspace`), so loading them together can result in colliding symbols.
> 
> I had a quick look - there's `directorClassName()` in `Source/Modules/lang.cxx` which builds this class name. We can't just use the `%nspace` code path always since that takes the namespace from the `sym:nspace` attribute which is only set in the `%nspace` case, but we could determine it from other attributes of the class node (e.g. `classtype` and `name` both have the C++ class name including the namespace).
> 
> It also looks like many of the target languages currently duplicate knowledge about how to build a director class name, as does `Source/Swig/cwrap.c`. Ideally we'd have this encapsulated in a single place. I wouldn't be surprised if there are currently bugs with `%nspace` and directors due to disagreements about the director class name.

Hello, has this problem been solved?

ojwb:
> Hello, has this problem been solved?

There isn't a simple reproducer here to check with, but I'm fairly sure it hasn't.

PR #771, opened 2016-08-23
Make Python builtin types hashable by default

wsfulton:
Default hash is the underlying C/C++ pointer.

This matches up with testing for equivalence (Py_EQ in SwigPyObject_richcompare) which compares the pointers. Usually a == b implies hash(a) == hash(b) and this is now true.

Changes also fixes corrects SWIGPY_HASHFUNC_CLOSURE to use the new Py_hash_t type introduced in Python 3.2.


wsfulton:
Merging in given no objections

PR #777, opened 2016-08-31
Update xml.cxx to return typenames

ahmed-usman:
Typename information wasn't being written to the XML file. Added a condition at line 149 to handle templateparms differently to return typenames.

### DIFF
--- a/Source/Modules/xml.cxx
+++ b/Source/Modules/xml.cxx
@@ -146,7 +146,10 @@ public:
 	Xml_print_parmlist(Getattr(obj, k));
       } else if (Cmp(k, "catchlist") == 0) {
 	Xml_print_parmlist(Getattr(obj, k), "catchlist");
-      } else {
+      } else if (Cmp(k, "templateparms") == 0) {
+      	Xml_print_parmlist(Getattr(obj, k), "templateparms");
+      }
+      else {
 	DOH *o;
 	print_indent(0);
 	if (DohIsString(Getattr(obj, k))) {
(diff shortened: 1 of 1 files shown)


ahmed-usman:
The failure appears unrelated to this change.

Source/Modules/xml.cxx
```diff
@@ -146,7 +146,10 @@ class XML:public Language {
 	Xml_print_parmlist(Getattr(obj, k));
       } else if (Cmp(k, "catchlist") == 0) {
 	Xml_print_parmlist(Getattr(obj, k), "catchlist");
-      } else {
+      } else if (Cmp(k, "templateparms") == 0) {
```
vadz: Maybe unify this with the `if` above, especially if there could be more such cases in the future? I.e. just do

      } else if (Cmp(k, "catchlist") == 0 || Cmp(k, "templateparms") == 0) {
	Xml_print_parmlist(Getattr(obj, k), k);
      } else {

?

vadz:
This looks obviously correct to me, so IMO it should be merged.

Source/Modules/xml.cxx
ahmed-usman: I had to write it like this:

`Xml_print_parmlist(Getattr(obj, k), Char(k));`

print_parmlist takes a char * as the second argument so k needed to be converted before it could be passed. Thank you for the combining suggestion.

ojwb:
Code change looks good to me too.  But it fixes a user-visible bug, so ought to have a corresponding entry in `CHANGES.current`.

PR #781, opened 2016-09-08
Fix enum and #define error when value contains char in compound expression

jiulongw:
Problem: When enum value contains compound expression with a char
constant, the quotes around char constant is missing in the generated
expression. Example:

enum media_type {
  YUY2 = ((('2' << 24) | ('Y' << 16)) | ('U' << 8)) | 'Y'
};

The generated C# enum becomes:

public enum media_type {
  YUY2 = (((2 << 24)|(Y << 16))|(U << 8))|Y
}

While the correct representation (after this fix) should be:

public enum media_type {
  YUY2 = ((('2' << 24)|('Y' << 16))|('U' << 8))|'Y'
}

Cause: the exprcompound promotes the expression type from char to int
and uses $1.val in the generated expression. However $1.val does not
contain the quotes. Since the type is promoted to int, there's no way to
know there's char component in the compound expression.

Solution: in exprcomound, use $1.rawval if $1.type is T_CHAR or T_WCHAR.
The rawval contains quotes which yield correct expression.

### DIFF
--- a/Examples/test-suite/csharp/enum_thorough_typesafe_runme.cs
+++ b/Examples/test-suite/csharp/enum_thorough_typesafe_runme.cs
@@ -413,6 +413,8 @@ public class runme {
       if (enum_thorough_typesafe.differentTypesTest(DifferentTypes.typeboolfalse).swigValue != 0) throw new Exception("differentTypes 4 failed");
       if (enum_thorough_typesafe.differentTypesTest(DifferentTypes.typechar).swigValue != (int)'C') throw new Exception("differentTypes 5 failed");
       if (enum_thorough_typesafe.differentTypesTest(DifferentTypes.typedefaultint).swigValue != (int)'D') throw new Exception("differentTypes 6 failed");
+      if (enum_thorough_typesafe.differentTypesTest(DifferentTypes.typecharcompound).swigValue != (int)'A' + 1) throw new Exception("differentTypes 7 failed");
+      if (enum_thorough_typesafe.differentTypesTest(DifferentTypes.typecharcompound2).swigValue != (int)'B' << 2) throw new Exception("differentTypes 8 failed");
 
       int global_enum = enum_thorough_typesafe.global_typeint;
       if (enum_thorough_typesafe.globalDifferentTypesTest(global_enum) != 10) throw new Exception("global differentTypes 1 failed");
@@ -426,6 +428,10 @@ public class runme {
       if (enum_thorough_typesafe.globalDifferentTypesTest(global_enum) != 'C') throw new Exception("global differentTypes 5 failed");
       global_enum = enum_thorough_typesafe.global_typedefaultint;
       if (enum_thorough_typesafe.globalDifferentTypesTest(global_enum) != 'D') throw new Exception("global differentTypes 6 failed");
+      global_enum = enum_thorough_typesafe.global_typecharcompound;
+      if (enum_thorough_typesafe.globalDifferentTypesTest(global_enum) != (int)'A' + 1) throw new Exception("global differentTypes 7 failed");
+      global_enum = enum_thorough_typesafe.global_typecharcompound2;
+      if (enum_thorough_typesafe.globalDifferentTypesTest(global_enum) != (int)'B' << 2) throw new Exception("global differentTypes 8 failed");
     }
   }
 }
--- a/Examples/test-suite/csharp/enum_thorough_typesafe_runme.cs
+++ b/Examples/test-suite/csharp/enum_thorough_typesafe_runme.cs
@@ -405,6 +405,10 @@ public class runme {
       if (enum_thorough_typesafe.repeatTest(repeat.llast).swigValue != 3) throw new Exception("repeatTest 5 failed");
       if (enum_thorough_typesafe.repeatTest(repeat.end).swigValue != 3) throw new Exception("repeatTest 6 failed");
     }
+    {
+      if (enum_thorough_typesafe.enumWithMacroTest(enumWithMacro.ABCD).swigValue != (('A' << 24) | ('B' << 16) | ('C' << 8) | 'D')) throw new Exception("enumWithMacroTest 1 failed");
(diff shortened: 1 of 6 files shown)


jiulongw:
Anyone know how to re-run appveyor job?  One of the builds failed due to git clone error...

```
Build started
git clone -q https://github.com/swig/swig.git C:\projects\swig
fatal: early EOF
fatal: The remote end hung up unexpectedly
fatal: index-pack failed
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
Command exited with code 128
```

wsfulton:
I've restarted Appveyor.

The patch requires a test. Can you please add in your testcase to Examples/test-suite/enum_thorough.i. A runtime test in these files would also be a bonus as they are runtime tests for code in enum_thorough.i:
```
$ find Examples/test-suite/ -name "enum_thorough_*_runme.*"
Examples/test-suite/csharp/enum_thorough_simple_runme.cs
Examples/test-suite/csharp/enum_thorough_typesafe_runme.cs
Examples/test-suite/java/enum_thorough_typeunsafe_runme.java
Examples/test-suite/java/enum_thorough_proper_runme.java
Examples/test-suite/java/enum_thorough_simple_runme.java
```

jiulongw:
Thanks @wsfulton!  Added test cases related to this patch and verified runtime tests passed.

jiulongw:
Ready to merge.  @wsfulton could you help to review / merge this PR?  Added new test cases and passed all checks.

ojwb:
We've had a very similar issue to this one reported almost a decade ago, and I even attempted a fix but it caused other issues so I had to back it out - I think this is the mail thread about it:

https://sourceforge.net/p/swig/mailman/message/6450308/

Unhelpfully SF renumbered all the tickets for the new trackers, but it seems you can search for #1474463 to find the renumbered ticket, which brings up this issue:

https://sourceforge.net/p/swig/bugs/668/

So that case was for a `#define` not a `enum`:

```
#define H_SUPPRESS_SCALING_MAGIC (('s'<<24) | ('u'<<16) | ('p'<<8) | ('p')) 
```

It'd be good to check that the fix handles this variant, and add it to the regression test.

ojwb:
https://sourceforge.net/p/swig/bugs/1368/ has another case which I think boils down to the same issue as this one - here's the testcase from there, reduced significantly in size:

```
#define PACK(C1,C2,C3,C4) ((C1<<24)|(C2<<16)|(C3<<8)|C4)

typedef enum ids {
   ABCD = PACK('A','B','C','D'),
   ABCD2 = ABCD
} ids;
```

jiulongw:
Thanks @ojwb for the cases and history of this issue.

This patch fixes the second issue (where enums value is a macro expansion).

However for the first case, the generated wrapper code doesn't compile with or without this patch.

For example:
```
#define XYZ ('a' << 23)
```

Without this patch:

```
result = (int)((a << 23));
```

With this patch:

```
result = (int)((\'a\' << 23));
```

Looks like there's extra escape here.

I will try to fix it without breaking existing test cases.  But I don't think this is a blocking issue for this patch since it doesn't break working cases.

jiulongw:
Hi @ojwb @wsfulton I added couple more test cases per @ojwb's suggestion.  Looks like all tests passed.  Could you guys help to merge this change?

For the issue below I see couple of FIXMEs in existing test cases.  I have rough idea about the fixes but prefer to send a separate PR since it is a different issue (not enum)

```
#define H_SUPPRESS_SCALING_MAGIC (('s'<<24) | ('u'<<16) | ('p'<<8) | ('p'))
```

ojwb:
Thanks for the updates.

I'm a bit nervous that the quoting issue might actually manifest as a problem in a case we don't have test coverage for currently.  Perhaps I'm being overly cautious due to my earlier failed attempt at fixing this.  While it doesn't seem we usefully recorded what the exact issue was, my memory is it was due to incorrect quoting in some cases.

This case currently works and is broken by the patch:

```
%module foo
%inline %{
#define FOO ('\0'+7)
%}
```

At least for PHP, that currently gives:

```
SWIG_LONG_CONSTANT(FOO, (int)(+7));
```

And with the patch gives:

```
SWIG_LONG_CONSTANT(FOO, (int)(\'\\0\'+7));
```

But that currently works mainly by luck (because `'\0'` has value 0, and currently gets handled as an empty string) - this isn't really the class of issues I'm worried about.

I wouldn't want to block merging this patch - it does seem like an improvement over the current situation - but I'd like to have a bit of time to make sure we aren't just fixing some cases at the expense of others.

jiulongw:
@ojwb I agree.  I'm not familiar with the merge process but how does it help without merging the patch?  Do we have beta release with unmerged patches, or are you going to do more test on this patch? 😄

I have a tentative patch that fixes the `#define` issue.  However I worry that might break more cases.  So I want to make it a different PR.

jiulongw:
Ping... any updates? :-)  @ojwb

wsfulton:
We'll only merge patches that we are confident will not regress master, so I suggest you add additional commits to this patch request until we think the combined set of changes is okay.

jiulongw:
@ojwb @wsfulton Looks like CI has some problems.  I checked all failures, they are all environment setting related.  Could you guys take a look?  (Sorry I don't have full test environment locally so need to rely on CI to run all test cases).

e.g:
/Users/travis/build.sh: line 45: gcc-4.2: command not found
/Users/travis/build.sh: line 45: shell_session_update: command not found

wsfulton:
The continuous integration errors are just the runs on OSX. The HEAD of master has fixed these now, but I think they can be safely ignored anyway.

wsfulton:
One observation, for this code in the string_constants.i testcase:
```c
#define ZS1 "\0"
#define ES1 ""
```
the Go wrappers generate
```c
  result = (char *)"";
```
previously it was:
```c
  result = (char *)"\0";
```
for both ZS1 and ES1. The generated string is now the same as Python, but I don't see this as a problem.

@ojwb I think this is ready to merge, unless you have any more comments.

ojwb:
@wsfulton: I have some lingering nervousness about it, but at this point I think only because of my previous bad experience trying to address this - the patches here seem a much more successful attempt than mine.  Since you're happy with it, please go ahead.

@jiulongw: Thanks for all your work on this.

jiulongw:
Thanks guys! :)

ojwb:
Also, this clearly fixes a number of cases which people have hit over the years and it doesn't regress our testsuite (which is significantly more thorough than it was a decade ago).

Issue #240, opened 2014-10-14
C++ exceptions cause 'segmentation fault: 11' in PHP wrapper.


SergeySeroshtan:
Reproduced on SWIG version 3.0.2.
###### Steps to reproduce:
 1. Write value-object C++ class with copy constructor.
 1. Write class with method that conforms to conditions described below.
 1. Write SWIG interface file with standard exception handling code.
 1. Write PHP code that invokes method which throws exception.
 1. Run PHP code.

###### Wrapped method conditions:
- wrapped method should return object by value;
- returned object should contain copy constructor;
- wrapped method should throw exception before resulting object will be returned.

###### Reason:
Default SWIG macro function 'SWIG_exception' defined for PHP as:
```
#define SWIG_exception(code, msg) zend_throw_exception(NULL, (char*)msg, code TSRMLS_CC)
```
**does not interrupt method's execution flow**, which leads to **dereferencing of NULL object** that returned as result.

###### Solution:
Redefine SWIG macro function 'SWIG_exception'  for PHP as:
```
#ifdef SWIGPHP
%{
#undef SWIG_exception
#define SWIG_exception(code, msg) { zend_throw_exception(NULL, (char*)msg, code TSRMLS_CC); return; }
%}
#endif
```

###### Example: modification of the SWIG example located at 'Examples/php/class'
example.h
```
/* File : example.h */

class Value {
public:
  Value(int a, int b);
  Value(const Value& other);
private:
  int a_;
  int b_;
};

class Shape {
public:
  Shape() {
    nshapes++;
  }
  virtual ~Shape() {
    nshapes--;
  }
  double  x, y;
  void    move(double dx, double dy);
  virtual double area() = 0;
  virtual double perimeter() = 0;
  virtual Value throwException();
  static  int nshapes;
};

class Circle : public Shape {
private:
  double radius;
public:
  Circle(double r) : radius(r) { }
  virtual double area();
  virtual double perimeter();
};

class Square : public Shape {
private:
  double width;
public:
  Square(double w) : width(w) { }
  virtual double area();
  virtual double perimeter();
};
```
example.cxx
```
/* File : example.cxx */

#include "example.h"
#define M_PI 3.14159265358979323846

#include <stdexcept>

Value::Value(int a, int b) : a_(a), b_(b) {}
Value::Value(const Value& other) : a_(other.a_), b_(other.b_) {}

/* Move the shape to a new location */
void Shape::move(double dx, double dy) {
  x += dx;
  y += dy;
}

[...]

PR #785, opened 2016-09-09
Arrange to return after calling zend_throw_exception()

ojwb:
Should fix github issue #240.


ojwb:
I've created a new testcase from the reproducer from #240 and squashed that onto the commit, rebased, and force pushed - CI can test it for more languages than I easily can.

ojwb:
There were some test failures on OS X due to `M_PI` being redefined by the testcase - I've changed that to `PI`, added a `CHANGES.current` entry, squashed and pushed as 26bbc96d58b81293a2357ba3ba7c4419d87417dd.

PR #787, opened 2016-09-10
Fix one casting warning on Linux

brgirgis:
This change removes one casting warning on Linux using g++ (SUSE Linux) 6.2.1 20160830 [gcc-6-branch revision 239856]


ojwb:
What's the warning message this eliminates?

brgirgis:
warning: conversion to ‘double’ from ‘long int’ may alter its value [-Wconversion]
     if (val) *val = PyInt_AsLong(obj);
                     ~~~~~~~~~~~~^~~~~

ojwb:
Merged - thanks for the patch (though I'm surprised there aren't rather more warnings with `-Wconversion`).

brgirgis:
No problem :) There were more but @dnwillia-ansys  and I fixed them in a previous PR.

PR #792, opened 2016-09-17
[Javascript] Fix SWIG_exception() macro

### LINKED ISSUE #789: Documentation about C++ exceptions may leads to errors
Hello,

Just a little message about how C++ exceptions handling is described in the documentation especially when targeting JavaScript : the documentation is explaining us to use something like this : 
```
%exception {
  try {                                                     
    $action
  } catch (const std::runtime_error &e) {
    SWIG_exception(SWIG_RuntimeError, e.what());
  } catch (...) {
    SWIG_exception(SWIG_RuntimeError, "unknown exception");
  } 
} 
``` 

It's working for most of the situations but when an error is thrown in the constructor of an object, it makes the app crashing (by calling the destructor while the object wasn't created). I found a solution by adding a `goto fail;`after the calls to `SWIG_exception`which seems to work actually.

Maybe the documentation needs an update to deal with this situation :wink:.

Thanks for making this nice tool which saved a lot of my time, I hope this helps =).

ojwb:
Fix SWIG_exception() macro to return from the current function.
Fixes #789, reported by Julien Dutriaux.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -6,6 +6,10 @@ Version 3.0.11 (in progress)
 ============================
 
 2016-09-16: olly
+            [Javascript] Fix SWIG_exception() macro to return from the current
+            function.  Fixes #789, reported by Julien Dutriaux.
+
+2016-09-16: olly
             [PHP] Fix SWIG_exception() macro to return from the current function.
             Fixes #240, reported by Sergey Seroshtan.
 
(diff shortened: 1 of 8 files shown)


ojwb:
CI failure seems to be unrelated, as I didn't change anything related to Python:

```
checking python testcase director_thread (with run test)
This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
make[1]: *** [Makefile:108: director_thread.cpptest] Error 3
```

wsfulton:
The director_thread example sometimes fails, probably there is an intermittent bug to be fixed in it.

PR #793, opened 2016-09-17
Add std::bad_cast to std_except.i

### LINKED ISSUE #783: swig's std_except.i is missing the std::bad_cast exception
This exception occurs for example when the target of a `dynamic_cast` is a reference instead of a pointer (e.g. `dynamic_cast<Bar &>(*this)`).

The fix would be to add
```
  %std_exception_map(bad_cast, SWIG_TypeError); // or maybe SWIG_RuntimeError?
```
to std_except.i

q-p:
This exception occurs when dynamic_cast<T&> fails.

Fixes #783.


wsfulton:
@q-p did you see the build errors and request to extend this idea?

q-p:
I did, and thought I fixed them with the additional commit (dd7a5a6), or was there something else?

wsfulton:
Sorry, I missed that somehow. All looks good, thanks.

Issue #795, opened 2016-09-21
[Go] Wrong argument names for inherited functions that has more than 8 parameters


dontpanic92:
Hi,

When a function is inherited from the base class and it has more than 8 parameters, the generated C++ wrapper code will have wrong arguments' names. For example,

```
// test2.i
%module test2
%inline %{

class A {
public:
void a(int a1, int a2, int a3, int a4, int a5, int a6, int a7, int a8, int a9, int a10, int a11, int a12)
{
    return 0;
}

void b(int a1, int a2, int a3, int a4, int a5, int a6, int a7, int a8) {
    return 0;
}
};

class B : public A{
public:
void c();
};

%}

```

The generated C++ wrapper code for `B::a` will be

```
void _wrap_B_a_test2_ad4149af75286dd4(B *_swig_go_0, intgo _swig_go_1, intgo _swig_go_2, intgo _swig_go_3, intgo _swig_go_4, intgo _swig_go_5, intgo _swig_go_6, intgo _swig_go_7, intgo _swig_go_8, intgo _swig_go_9, intgo _swig_go_10, intgo _swig_go_11, intgo _swig_go_12) {
  B *arg1 = (B *) 0 ;
  int arg2 ;
  int arg3 ;
  int arg4 ;
  int arg5 ;
  int arg6 ;
  int arg7 ;
  int arg8 ;
  int arg9 ;
  int arg10 ;
  int arg11 ;
  int arg12 ;
  int arg13 ;
  
  arg1 = *(B **)&_swig_go_0; 
  arg2 = (int)_swig_go_1; 
  arg3 = (int)_swig_go_2; 
  arg4 = (int)_swig_go_3; 
  arg5 = (int)_swig_go_4; 
  arg6 = (int)_swig_go_5; 
  arg7 = (int)_swig_go_6; 
  arg8 = (int)_swig_go_7; 
  arg9 = (int)_swig_go_8; 
  arg10 = (int)_swig_go_9; 
  arg11 = (int)_swig_go_10; 
  arg12 = (int)_swig_go_11; 
  arg13 = (int)_swig_go_12; 
  
  A *swig_b0 = (A *)arg1;
  (swig_b0)->a(arg2,arg3,arg4,arg5,arg6,arg7,arg8,arg9,swig_b00,swig_b01,swig_b02,swig_b03);
  
}
```

Note that `swig_b0X` are not defined and should be `arg1X`. 

[...]

PR #797, opened 2016-09-22
[Go] Fix argument names in inherited functions #795

dontpanic92:
This commit fixes a wrong argument name replacement in the inherited functions as well as a memory leak.


dontpanic92:
@ojwb Sorry that I didn't get the point at that time. Is this a proper way to add a testcase?

ojwb:
That look about right, except it's relevant for all languages, not just Go.

The CI failures look entirely unrelated.

ojwb:
Thanks.  CI failure is unrelated, so merging.

PR #801, opened 2016-09-26
fix warnings from compiling generated python wrappers

teythoon:
I'm trying to reduce the number of warnings that are emitted when compiling the SWIG-generated wrapper code.

### DIFF
--- a/Lib/cpointer.i
+++ b/Lib/cpointer.i
@@ -114,7 +114,7 @@ static NAME * frompointer(TYPE *t) {
 
 %define %pointer_functions(TYPE,NAME)
 %{
-static TYPE *new_##NAME() { %}
+static TYPE *new_##NAME(void) { %}
 #ifdef __cplusplus
 %{  return new TYPE(); %}
 #else
@@ -149,7 +149,7 @@ static TYPE NAME ##_value(TYPE *obj) {
 }
 %}
 
-TYPE *new_##NAME();
+TYPE *new_##NAME(void);
 TYPE *copy_##NAME(TYPE value);
 void  delete_##NAME(TYPE *obj);
 void  NAME##_assign(TYPE *obj, TYPE value);
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1337,7 +1337,7 @@ SWIG_Python_ConvertPacked(PyObject *obj, void *ptr, size_t sz, swig_type_info *t
 */
 
 SWIGRUNTIME PyObject* 
-SWIG_Python_NewShadowInstance(SwigPyClientData *data, PyObject *swig_this)
+SWIG_Python_NewShadowInstance(SwigPyClientData *data, PyObject *swig_this_value)
 {
 #if (PY_VERSION_HEX >= 0x02020000)
   PyObject *inst = 0;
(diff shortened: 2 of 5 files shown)


ojwb:
For a warning fix, it's useful to know which compiler you see the warning with and what the warning is.  It's not uncommon for warning fixes to trigger different warnings elsewhere, and knowing exactly what warning the original fix was for is really useful at that point.

teythoon:
Yes, sorry, I thought it was kind of clear what kind of warnings were fixed by each individual commit, but better be explicit.

We are using gcc 6.1.1 from Debian testing.

The first patch fixes warnings of the form:

```
gpgme_wrap.c:4046:23: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
   static gpgme_ctx_t *new_gpgme_ctx_t_p() {
```

The second fixes a bunch of warnings like this:

```
gpgme_wrap.c:4549:53: warning: unused parameter ‘self’ [-Wunused-parameter]
 SWIGINTERN PyObject *_wrap_gpgme_err_make(PyObject *self, PyObject *args) {
```

(This is *really* noisy, 346 instances of that warning per build.)

Finally, the last patch fixes:

```
gpgme_wrap.c:2447:65: warning: declaration of ‘swig_this’ shadows a global declaration [-Wshadow]
 SWIG_Python_NewShadowInstance(SwigPyClientData *data, PyObject *swig_this)
                                                                 ^~~~~~~~~
gpgme_wrap.c:2188:18: note: shadowed declaration is here
 static PyObject *swig_this = NULL;
                  ^~~~~~~~~
gpgme_wrap.c: In function ‘SWIG_Python_SetSwigThis’:
gpgme_wrap.c:2522:51: warning: declaration of ‘swig_this’ shadows a global declaration [-Wshadow]
 SWIG_Python_SetSwigThis(PyObject *inst, PyObject *swig_this)
                                                   ^~~~~~~~~
gpgme_wrap.c:2188:18: note: shadowed declaration is here
 static PyObject *swig_this = NULL;
                  ^~~~~~~~~
```

Do you want me to add these information to the commit messages and create another pull request?

FWIW, I don't think the ci failure is related to my changes.

ojwb:
> Do you want me to add these information to the commit messages and create another pull request?
Be good to add, but feel free to force-push to the same branch which will update this PR.
> FWIW, I don't think the ci failure is related to my changes
Yeah, that guile failure seems to be a travis issue.

wsfulton:
The `-Wshadow` and `-Wstrict-prototypes` patches look okay. These should be added to Tools/testflags.py for **all** languages, not just Python, that is, to c_common and cxx_common in this script. This will ensure we don't regress these warnings going forwards. Possibly some other languages will fail as a result of this... @teythoon would you be willing to fix those too if Travis testing shows some failures?

Regarding `-Wunused-parameter` warnings, these are a real nuisance. From what I can make out, you are trying to squash these warnings for `-builtin` where `self` is sometimes used in the wrapper function and sometimes not. I think the most portable approach is to add `(void)self;` into the wrapper method where this might happen as SWIGUNUSED only works for gcc. The patch results in some code generating `PyObject *SWIGUNUSEDPARM(self) SWIGUNUSED` sometimes which is not good. How about generating `PyObject *self` and adding in `(void)self;`.

wsfulton:
Sorry, I can't accept this patch while it has broken nearly every test on Travis

teythoon:
Sure thing.  I just enabled the flags as you suggested to see how bad it is.  I'll continue to work on this eventually.

ojwb:
I've peeled off and merged the first commit as I could reproduce those warnings locally so that's clearly still useful.

I'm going to rebase the rest onto a new branch and see how things look in CI.

ojwb:
The `swig_this` change is no longer necessary - commit 344e9fefa8930b45b4595c0cda052ed9b2af5b40 renamed the global that was being shsdowed (which seems a better fix actually):

```
-static PyObject *swig_this = NULL;
+static PyObject *Swig_This_global = NULL;
```

So my branch will actually just be adding the warning switches.

ojwb:
Branch pushed: https://github.com/swig/swig/tree/enable-more-warnings-in-ci

CI running: https://github.com/swig/swig/actions/runs/1788490061

I'll look at the results and merge either/both option if they're actually clean.  I suspect they won't be but we can at least see if we're anywhere close.

@teythoon Thanks for the patch and sorry it stalled so long.  If you do want to work more on this, please do.

I'm going to close this now as the proposed changes are all merged or otherwise addressed.

ojwb:
The testsuite has a lot of code where a class member is initialised by a ctor parameter of the same name which warns with `-Wshadow`, for example:

```
smart_pointer_const.i:   Bar(Foo *f) : f(f) { }
```

There's also quite a lot from the other warning too, so we can't easily turn on either in CI right now.

teythoon:
Happy the patches were still useful :)

I'm afraid I have moved on and am no longer using SWIG in any of my current projects, so I won't continue to work on this.

Issue #807, opened 2016-10-04
SWIG fails handling template class with same name as template class in parent namespace


fschlimb:
A template class in a nested namespace (a::b) is not treated correctly if a template class exists with the same name in a parent namespace (a):

```
%module b;

namespace a {
    template< typename T >
    class Batch {
    };
}

namespace a {
    namespace b {
        template< typename T >
        class Batch {
        };
    }
}
```

```
> ./preinst-swig -python -c++ grrr/b.i
grrr/b.i:16: Warning 302: Identifier 'Batch' redefined (ignored),
grrr/b.i:7: Warning 302: previous definition of 'Batch'.
grrr/b.i:16: Warning 302: Identifier 'Batch' redefined (ignored),
grrr/b.i:7: Warning 302: previous definition of 'Batch'.
```

Similar but probably not related to #656. Still, a fix to this might also fix #656 (for which I have a fix which does not fix this one).

Note, this construct works fine if
- the template classes have different names
- the classes are non-template classes
- the parent class is not in a parent namespace

I tried to debug this but I am getting lost. When the child class is handled in `Swig_symbol_add` (symbol.c ) the current scope (e.g. `current`) is incorrect, it points to the parent class.

Any idea, thoughts, hints?

fschlimb:
This also fails if the previous declaration comes from a using directive.

PR #810, opened 2016-10-06
SWIG fails handling template class with same name as template class in parent namespace

fschlimb:
resolve_create_node_scope treated template classes with same name which it found in a different namespace as a "weak" symbol of the same def. This made it create spurious namespaces and incorrectly reported multiply defined classes.

This is not correct, it can only act so if in the same namespace.
We are now adding a new node in such cases.

fixing issue #807


fschlimb:
I guess I need some help.

If I am not mistaken, the symtabs indicate "using directives" (e.g. "using namespace X") with the "inherit" key, the "using declaration" (e.g. "using std::pair") with "uname". That's at least what I infer from the failing example. If that's right, functions like ```Swig_symbol_clookup_no_inherit``` use a swapped/inverse terminology in their comments.

Anyway, in ```parser.y:resolve_create_node_scope``` we need to differentiate between a template instantiation and a potentially new template declaration. The test case I added is the latter, but in li_std_pair_using.i it's the former. I do not know how to differentiate.

I need to know the difference, since a new declaration must always create a new node (unless it exists already, but that's an error anyway). However, a specialization/instantiation must not, it must instead use a existing declaration if found. All I have is the name (string), so I cannot decide if I have to create a new node or not.

Any suggestion/idea/help would be appreciated.

fschlimb:
Anyone? @ojwb? @wsfulton?

ojwb:
If it's not obvious from looking at the node in each case, I guess you'll have to add a new flag when such nodes are created so you can tell later.

fschlimb:
Ok, I can investigate this option.
Any hint on where/what "later" could be?

ojwb:
Rereading you want to know in the parser so what I said is probably not useful.  I don't know how you'd tell those apart.

Examples/test-suite/same_template_name.i
```diff
@@ -0,0 +1,15 @@
...
+
+}
+
+namespace ns__a {
+    namespace ns__b {
+
+        template< typename T > class TName : public TName<T> {};
+
+    }
+}
```
wsfulton, line 15: This is not valid C++. I used godbolt.org to check and I get compiler errors with up to date compiler versions: gcc-12.2 and  clang-15.0. Interestingly, icc-2021 and msvc-19.3 did not complain with the code above, but an attempt to instantiate the template with the following then resulted in errors:
```C++
void tester() {
  ns__a::ns__b::TName<int> tn;
}
```
Latest SWIG parses it now without error, but instantiating it with %template does lead to incorrect results. I think that is par for the course.

wsfulton:
Closing as invalid.

PR #813, opened 2016-10-10
Templ def cache

fschlimb:
The cache for default template types was too aggressive. Using the same template on types with the same name but in different namespaces resulted in wrong lookups (`Swig_symbol_template_deftype`).
Using qualified scope fixes this.

Here's a reproducer:

amodel.i

```
%module amodel;

namespace ns__a {
    namespace iface1 {
        class Model {};
        typedef d::d<Model> ModelPtr;
    }
}
```

bmodel.i

```
%module bmodel;

%import "amodel.i"

namespace ns__b {
    namespace iface1 {
        class Model {};
        typedef d::d<Model> ModelPtr;
    }
 }
```

`-debug-typedef` incorrectly produced:

```
Type scope 'ns__b::iface1' (0x7fabc55eae10)
                               ModelPtr -> d::d<(ns__a::iface1::Model)>
                                   Model -> Model
```

(It uses `Model` from `ns__a`, not `ns__b`).


wsfulton:
Can you please add your testcase into the test-suite please in your pull requests?
Modifying type lookup is somewhat fragile and if you aren't running the test-suite on a few different mainstream target languages before submitting your requests, then I suggest you please do so that at least a couple of the target language test-suites pass on Travis!

fschlimb:
Is this sufficient?

wsfulton:
This is a simple but great fix. Using an artificial scope '..' is somewhat unusual though. A more conventional approach used in the code base would be to have a separate hash for each of scopes before the '..', that is, create a new hash for each scope from your additional calls to Swig_symbol_qualifiedscopename. Could you make this small tweak, otherwise it looks good.

The `-debug-typedef` showed it wasn't working, but I couldn't get your testcase to fail, so I've modified it slightly to end up with: 
```swig
%inline %{
    namespace d {
        template< typename T > class d {};
    }
%}

%ignore ns__a::iface1::Model;

%inline %{
namespace ns__a {
    namespace iface1 {
        class Model {};
        typedef d::d<Model> ModelPtr;
    }
    using iface1::ModelPtr;
}
%}

%inline %{
namespace ns__b {
    namespace iface1 {
        class Model {
        public:
            ns__a::ModelPtr foo() { return ns__a::ModelPtr(); };
        };
        typedef d::d<Model> ModelPtr;
        ns__a::ModelPtr get_mp_a() { return ns__a::ModelPtr(); }
        ModelPtr get_mp_b() { return ModelPtr(); }
    }
 }
%}
%template(AModelPtr) d::d<ns__a::iface1::Model>;
%template(BModelPtr) d::d<ns__b::iface1::Model>;
```
The `get_mp_b` returned `AModelPtr` instead of `BModelPtr` before your fix, so is easy to detect if working.

fschlimb:
Thanks! Sure I'll change the test (or will you push your changes?).

Not sure I really understand your suggestion.
Are you suggesting using a 2-level hashing? One hash mapping each added scope (e.g. which I put before '..') to other (e.g. the old) hashes?
If that's what you suggest then I don't see the benefit we'd trade against adding a second hash-lookup and the additional memory overhead. The caching mechanism is meant to be a performance optimization...

wsfulton:
Please add the testcase modifications into the pull request.

Yes, a 2 level hash. Contrary to your opinion, I suspect this will be faster and use less memory because the strings used for the hashes will be shorter. Any difference is likely to be irrelevant though in the real world, but without some actual timings on a real code base, both our opinions will remain conjecture!

wsfulton:
Thanks, I've merged this in now and also fixed the testcase so it checks the correct type is returned by `get_mp_b`.

wsfulton:
The merge commit showing the changes is: 4c1152efcdcdaf1db4047ca1898a8905ede6677c

fschlimb:
Sorry, I couldn't find time to fix this last myself.
Thanks a lot for finalizing this!
Stay tuned for more PRs...

PR #818, opened 2016-10-15
Add `defpackage`, define slots while `defclass`

muyinliu:
1. Add `defpackage` for module(like `allegrocl.cxx` do).
2. Define slots while `defclass` to prevent `When attempting to read the slot's value (slot-value), the slot xxx is missing`.
3. Fix bug while generate `(cl:defmethod initialize-instance :after ((obj %s) &key%s)` in `emit_initialize_instance`, there is no need and should NOT declare parameter-type to a parameter declared after `&key`, just declare parameter itself would be fine.
4. Fix some indentation.

### DIFF
--- a/Source/Modules/cffi.cxx
+++ b/Source/Modules/cffi.cxx
@@ -34,6 +34,7 @@ CFFI Options (available with -cffi)\n\
 class CFFI:public Language {
 public:
   String *f_cl;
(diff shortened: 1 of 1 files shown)


muyinliu:
Sorry to say I pass some ERROR code. I have to fix it and pull another request.

PR #819, opened 2016-10-16
Fix for CFFI: Add `defpackage` for module, fix `defmethod` parameter type

muyinliu:
1. Add `defpackage` for module(like `allegrocl.cxx` do).
2. Fix bug while generate `(cl:defmethod initialize-instance :after ((obj %s) &key%s)` in `emit_initialize_instance`, there is no need and should NOT declare parameter-type to a parameter declared after `&key`, just declare parameter itself would be fine.
3. Fix some indentation.

### DIFF
--- a/Source/Modules/cffi.cxx
+++ b/Source/Modules/cffi.cxx
@@ -178,6 +178,9 @@ int CFFI::top(Node *n) {
 
   Swig_banner_target_lang(f_lisp, ";;;");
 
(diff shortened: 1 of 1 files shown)


wsfulton:
@muyinliu would you be able to help get the SWIG test-suite working for CFFI ? While there is no known way of making it work for CFFI, we can't accept patches as there is no way of verifying patches as none of the SWIG developers are familiar with CFFI.

wsfulton:
Closing all open CFFI issues as CFFI has been disabled, please see #1447.

Issue #722, opened 2016-06-17
Python index operators with -builtin not possible


SimplyKnownAsG:
I've spent a lot of time trying to figure out how to get the C++ `operator[]` to be correctly wrapped for Python with `-builtin`. My definition of correctly is that on the python side, I should get the equivalent of `__getitem__` and `__setitem__`. I think I understand why this is difficult, but I'd like to make sure there isn't some way this could be done.

I believe some reasons that `operator[]` is not / cannot be defined in `pyopers.swg` are:
- In C++ the `operator[]` method should return a reference so that it can also support assignment operations.
- In Python this could take [**four** difference slots](https://docs.python.org/2/c-api/typeobj.html):

```
%feature("python:slot", "sq_item", functype="ssizeargfunc") *::operator[];
%feature("python:slot", "sq_ass_item", functype="ssizeargfunc") *::operator[];
%feature("python:slot", "mp_subscript", functype="binaryfunc") *::operator[];
%feature("python:slot", "mp_ass_subscript", functype="objobjargproc") *::operator[];
```

I have a simple array implementation that uses `operator[]`, `-builtin` with

```
%rename("__getitem__") *::operator[];
%feature("python:slot", "sq_item", functype="ssizeargfunc") *::operator[];
```

defined.

When the return type is a reference (i.e. `double&`) Python says the type is `<Swig Object of type 'double *' at 0x0000000003E2DA08>`. When the return type is a value (i.e. `double`), Python gives the value. Both of these results make sense, but it is a bit of a pain that I cannot seem to get this to work.

Are there swig `%feature`, `%rename` operations that can be used to get `operator[]` to work as expected? I haven't fully looked at what is done for `std::vector` and `std::map`s yet, but I imagine the answer is out there!

mromberg:
Supporting operator[] in 100% C++ is not even a straight forward exercise for the reason you mentioned above.  Often a proxy class is needed just to deal with this in C++

https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Temporary_Proxy

  All this messy code would be involved before swig would even take a peek at the C++.  Then swig would have to try to create another layer of proxies (meta proxy?) for the scripting language to use.

  If it were me, I'd try really hard (using things like %extend if needed) to wrap the C++ in a way to not expose this type of interface to the scripting language.

m7thon:
Here is what I used once to get exactly this working for a simple C++ `Sequence` class using `-builtin`:

```c++
class Sequence {
...
};
[...]

PR #826, opened 2016-10-27
Added description of the operator[] caveats

SimplyKnownAsG:
#722
I found the `operator[]` to be particularly confusing as to why it wasn't automatically wrapped, probably due to my inexperience with C++. In trying to figure out how to get it to work, I saw a few questions on SO, and now that I have at least some sort of understanding, I think all that is needed is a documentation clarification.

Undoubtedly, I've missed some sort of guideline on how to do a pull request, please let me know...

### DIFF
--- a/Doc/Manual/SWIGPlus.html
+++ b/Doc/Manual/SWIGPlus.html
@@ -2766,6 +2766,16 @@ have to handle it like a normal function. For example:
 are ignored as well as conversion operators.
 </p></li>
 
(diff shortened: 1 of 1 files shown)


wsfulton:
Thanks for the improvement. I've edited it slightly and dropped the reference to the wiki page as I think this is quite obscure. Hopefully that's clear with you too and feel free to provide further edits.

PR #830, opened 2016-10-31
Fixes #230: shared_ptr breaks with C++ class inheritance by making swigCMemOwn private

### LINKED ISSUE #230: %shared_ptr breaks with C++ class inheritance by making swigCMemOwn private
see:
https://github.com/swig/swig/blob/master/Lib/java/boost_shared_ptr.i#L149

I noticed it by having a class B inheriting from A.
As soon as I made
%shared_ptr(A)
the java code breaks with:

Description Resource    Path    Location    Type
The field DllVersion.swigCMemOwn is not visible DllVersionUEye.java /uEyeSwig/src/ch/imagic/imagesource/idsueye line 29 Java Problem

haseebq:
Fixes this issue: https://github.com/swig/swig/issues/230

### DIFF
--- a/Lib/java/boost_shared_ptr.i
+++ b/Lib/java/boost_shared_ptr.i
@@ -146,7 +146,7 @@
 // Base proxy classes
 %typemap(javabody) TYPE %{
   private transient long swigCPtr;
-  private transient boolean swigCMemOwn;
+  protected transient boolean swigCMemOwn;
 
   PTRCTOR_VISIBILITY $javaclassname(long cPtr, boolean cMemoryOwn) {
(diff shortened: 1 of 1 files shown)


ojwb:
Your commit message is misleading - it seems to suggest you're making the member `private`, but actually you're changing it from `private` to `protected`.

Also, could you add the testcase from #230 to the testsuite so we have a regression test for this?

wsfulton:
4bf607589f821092f45742ef85b58b4818b60ab4 fixes this properly. Note that the `swigCMemOwn`/`swigCMemOwnDerived` must be set in all classes up the inheritance hierarchy.

PR #831, opened 2016-11-02
Prevent thread name being reset while using Java directors

andrey-starodubtsev:
Android JVM, Dalvik, resets thread name during `AttachCurrentThread`/`AttachCurrentThreadAsDaemon` calls (see similar issue in Chrominium bugtracker: https://bugs.chromium.org/p/chromium/issues/detail?id=384603).
This patch tries to remedy that using one of several approaches to get current thread name.


wsfulton:
Can you modify one of the examples/testcases to show it working and I can help incorporate it into a proper testcase and see what happens on Sun's JVM on Windows/Linux/Solaris? Sounds like this is just an Android JVM problem, and if so, perhaps you can simplify all this with one Android solution, maybe even put it all inside a `#if defined(__ANDROID__)` block? If not, I suggest the new code goes into a private helper function in JNIEnvWrapper with some comments as to why this is all necessary. 

Instead of using perror, can you just set a NULL thread name and use std::cout inside a `#if defined(DEBUG_DIRECTOR_THREAD)` block for debugging, similar to other code in this area which uses DEBUG_DIRECTOR_OWNED or DEBUG_DIRECTOR_EXCEPTION.

wsfulton:
ping @andrey-starodubtsev , can you address my comments?

wsfulton:
Please re-open when comments have been addressed.

Issue #772, opened 2016-08-25
(JS/Node/V8) C++ exceptions are not propagated when functions have default arguments


gpean:
Let's say we have
`%catches(std::runtime_error, std::exception, ...);`

Then the function calls will be enclosed in try/catch, which is good:
```
  try {
    (arg1)->foo();
  }
  catch(std::runtime_error &_e) {
    SWIG_exception_fail(SWIG_RuntimeError, (&_e)->what());
  }
  catch(std::exception &_e) {
    SWIG_exception_fail(SWIG_SystemError, (&_e)->what());
  }
  catch(...) {
    SWIG_exception_fail(SWIG_RuntimeError,"unknown exception");
  }
```

But if we have a function with default arguments, another wrapping functions will be generated which will silence the exception and prevent it from propagating to JS:

```
  OverloadErrorHandler errorHandler;
  
  if(args.Length() == 1) {
...
  }
  
  if(args.Length() == 0) {
...
  }
  
// !!!! exception info in errorHandler is lost :-( instead we have always the message below

  SWIG_exception_fail(SWIG_ERROR, "Illegal arguments for function foo");
```

The solution is this:

```
if (!errorHandler.err.IsEmpty())
  SWIGV8_THROW_EXCEPTION(errorHandler.err);
else
  SWIG_exception_fail(SWIG_ERROR, "Illegal arguments for function foo");
```

Right now I have added this to my .i to workaround that:

```
%fragment("js_function_dispatcher", "templates")
%{
static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args) {
    SWIGV8_HANDLESCOPE();
    
    v8::Handle<v8::Value> jsresult;
    OverloadErrorHandler errorHandler;
    $jscode
    
    //<ADDED BY GPEAN>
    if (!errorHandler.err.IsEmpty())
        SWIGV8_THROW_EXCEPTION(errorHandler.err);
    else
    //</ADDED BY GPEAN>
    SWIG_exception_fail(SWIG_ERROR, "Illegal arguments for function $jsname");
    
    goto fail;
    fail:
    SWIGV8_RETURN(SWIGV8_UNDEFINED());
}
%}
%fragment ("js_ctor_dispatcher", "templates")
%{
static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args) {
    SWIGV8_HANDLESCOPE();
    
    OverloadErrorHandler errorHandler;
    v8::Handle<v8::Value> self;
    
[...]

PR #832, opened 2016-11-03
Error thrown from overloaded constructor now throws the specified error.

ben51degrees:
When an error was thrown from an overloaded constructor, SWIGV8_ESCAPE
would not get called, which is correct. However, the program would then
carry on and get the error message overridden by a general error message
"Illegal arguments for construction of...". Adding an "else goto fail"
skips this line and returns the error that was thrown.

### DIFF
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -106,11 +106,15 @@ fail:
     self = $jswrapper(args, errorHandler);
     if(errorHandler.err.IsEmpty()) {
       SWIGV8_ESCAPE(self);
+    } else {
+      goto fail;
     }
 #else
     $jswrapper(args, errorHandler);
     if(errorHandler.err.IsEmpty()) {
       return;
+    } else {
+      goto fail;
     }
 #endif
   }


wsfulton:
We are missing a javascript maintainer atm and there isn't much chance of this being accepted unless the patch includes a runtime test in the test-suite showing the fix.

ben51degrees:
@wsfulton I will add some tests to this branch when I find some free time. From the travis tests it looks like this has had an effect on existing contructor tests, so I will take a look at this too before updating the request.

wsfulton:
Possibly related to #520 and #772.

ojwb:
I took a look at the proposed change here.  The first hunk of the patch is no longer relevant as we've dropped support for the older V8 versions that was for and removed that alternative code.

The remaining change looks wrong to me - e.g. consider the `overload` example - with current git master without this patch:

```
cd Examples/javascript/overload
make
```

Then look at `example_wrap.cxx`.

```
static SwigV8ReturnValue _wrap___wrap_f(const SwigV8Arguments &args) {
  SWIGV8_HANDLESCOPE();
  
  SWIGV8_VALUE jsresult;
  OverloadErrorHandler errorHandler;
  
  
  if(args.Length() >= 0 && args.Length() <= 0) {
    errorHandler.err.Clear();
    _wrap_f__SWIG_0(args, errorHandler);
    if(errorHandler.err.IsEmpty()) {
      return;
    }
  }
  
  
  if(args.Length() >= 1 && args.Length() <= 1) {
    errorHandler.err.Clear();
    _wrap_f__SWIG_1(args, errorHandler);
    if(errorHandler.err.IsEmpty()) {
      return;
    }
  }
  
  
  if(args.Length() >= 2 && args.Length() <= 2) {
    errorHandler.err.Clear();
    _wrap_f__SWIG_2(args, errorHandler);
    if(errorHandler.err.IsEmpty()) {
      return;
    }
  }
  
  
  if(args.Length() >= 1 && args.Length() <= 1) {
    errorHandler.err.Clear();
    _wrap_f__SWIG_3(args, errorHandler);
    if(errorHandler.err.IsEmpty()) {
      return;
    }
  }
  
  
  if(args.Length() >= 1 && args.Length() <= 1) {
    errorHandler.err.Clear();
    _wrap_f__SWIG_4(args, errorHandler);
    if(errorHandler.err.IsEmpty()) {
      return;
    }
  }
  
  
  if(args.Length() >= 1 && args.Length() <= 1) {
    errorHandler.err.Clear();
    _wrap_f__SWIG_5(args, errorHandler);
    if(errorHandler.err.IsEmpty()) {
      return;
    }
  }
  
  
  SWIG_exception_fail(SWIG_ERROR, "Illegal arguments for function f.");
  
  goto fail;
fail:
  SWIGV8_RETURN(SWIGV8_UNDEFINED());
}
```

If we `goto fail` on an exception then we won't try other overloads which take the number of parameters passed.

Normally SWIG uses typecheck typemaps to decide which overload to call rather than just checking the number of parameters and trying each overload which takes that number of parameters, but it seems we're missing most typecheck typemaps for Javascript currently, so that would need fixing first.  Once that's done this change is probably OK.

ojwb:
Seems we're still missing most `typecheck` typemaps for Javascript - there are only 7 judging by `git grep typecheck Lib/javascript`, all for pointer-related types.

ben51degrees:
Closing, as SWIG has progressed since this PR was opened 10 years ago, and this is no longer relevant.

PR #833, opened 2016-11-04
Better output from errors in Python classes when using directors

SidJ4t1qbit:
Directors.

### DIFF
--- a/Lib/python/director.swg
+++ b/Lib/python/director.swg
@@ -174,6 +174,7 @@ namespace Swig {
       }
       if (!PyErr_Occurred()) {
         PyErr_SetString(error, what());
+        PyErr_PrintEx(0);
       }
       SWIG_PYTHON_THREAD_END_BLOCK;
     }


wsfulton:
The interpreter ought to print the error message. Is this not happening? Can you please provide a description and use case for what you think the problem is?

wsfulton:
Did you see the python tests have broken...
```
checking python testcase director_exception (with run test)
TypeError: SWIG director type mismatch in output value of type 'std::string'
  File "director_exception_runme.py", line 52, in <module>
    b.pong()
  File "/home/travis/build/swig/swig/build/build/Examples/test-suite/python/director_exception.py", line 121, in pong
    return _director_exception.Foo_pong(self, val)
  File "/home/travis/build/swig/swig/build/build/Examples/test-suite/python/director_exception.py", line 121, in pong
    return _director_exception.Foo_pong(self, val)
SystemError: error return without exception set
```

wsfulton:
Duplicate of #836 not intended for SWIG master.

PR #834, opened 2016-11-14
get/set must be looked for at begin of name, not end

bartoli:
To find if methods are accessors, R module looks for get/set at the end of method names for some reason. It should look for them at the beginning?

### DIFF
--- a/Source/Modules/r.cxx
+++ b/Source/Modules/r.cxx
@@ -984,8 +984,7 @@ int R::OutputClassMemberTable(Hash *tb, File *out) {
     
     String *className = Getitem(el, 0);
     char *ptr = Char(key);
-    ptr = &ptr[Len(key) - 3];
-    int isSet = strcmp(ptr, "set") == 0;
+    int isSet = Strncmp(ptr, "set", 3) == 0;
     
     //        OutputArrayMethod(className, el, out);        
     OutputMemberReferenceMethod(className, isSet, el, out);
@@ -1036,13 +1035,12 @@ int R::OutputMemberReferenceMethod(String *className, int isSet,
     
     String *dup = Getitem(el, j + 1);
     char *ptr = Char(dup);
-    ptr = &ptr[Len(dup) - 3];
     
-    if (!strcmp(ptr, "get"))
+    if (!Strncmp(ptr, "get", 3))
       varaccessor++;
 
     String *pitem;
-    if (!Strcmp(item, "operator ()")) {
+    if (!Strncmp(item, "operator ()")) {
       pitem = NewString("call");
     } else if (!Strcmp(item, "operator ->")) {
       pitem = NewString("deref");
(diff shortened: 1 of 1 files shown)


ojwb:
"2,788 additions, 2,791 deletions not shown because the diff is too large"!

Did you reformat the whole file?  Please don't do that, it makes it so much harder to review changes, and for people looking back in the history in the future.

bartoli:
I guess i have somehow converted all the newlines at my first commit.. I'll also fix a warning about a variable that is now unused since i removed some code at the same time
(I'm still a git* beginner, i'm gonna have to figure out how to resubmit new code to the same PR in a clean way)

ojwb:
@bartoli For a case like this you can sensibly just rewrite history locally and then `git push --force` to replace the branch for the PR.  Or start again with a fresh PR if that's simpler.

If it's just your editor rewriting line endings and/or messing with indentation, then for the "rewrite history" part you can probably just generate a patch ignoring whitespace changes, reset the branch, then apply that patch and commit it:

```
git diff -b HEAD^^ > cleaner.patch
git reset --hard HEAD^^
git apply cleaner.patch
git diff
# and check the new diff looks sane
make && make check-r-test-suite
# do any further testing that's appropriate
git add -u
git commit
git push --force
```

If the code has been reformatted between lines, that won't be resolved by this.

richardbeare:
See #839 - I think the check for methods ending in "set" are intended for automatically generated code that provide access to class/structure variables, not user supplied class methods.

ojwb:
Assuming that's right, it would probably be cleaner to set a flag on the node for such functions rather than try to detect from the name (or use an existing flag if there is one).  That would avoid confusing a normally wrapped function with one of these generated functions.

richardbeare:
Eventually I think that would be the way to go. There's the decorator facility that might be right for that. However there is a lot of work to be done. #839 is designed to make the existing architecture more specific - currently we have the situation where methods with names like GetOffset are wrapped incorrectly because of the word ending. The patch doesn't eliminate the possibility, but makes it less likely.

There is a heap of work that needs to be done on the R module, and decisions based on names are part of what should change. I imagine that it might fix a lot of potential problems and improve flexibility.

bartoli:
Then i guess this PR can be closed/rejected, if the proper work will be done in #839?

ojwb:
@richardbeare "There's the decorator facility that might be right for that" - I meant a flag on the node in SWIG's parse tree, not at the R level.  I suspect there's actually a suitable one already - e.g. if `membervariableHandler:sym:name` is set then we're wrapping a member variable.

@bartoli Indeed - seems this check is meant to be for symbols SWIG itself generates, not getters/setters in the API being wrapped.

richardbeare:
Yep - I thought the swig documentation used the term "decorators" for adding extra bits to the internal tree. Definitely needs to be there, not the R level.

PR #839, opened 2016-11-23
Corrected generation of "set" methods for R.

richardbeare:
The function "OutputClassMemberTable" tags methods generated
for a class for R wrapper code supporting syntax like:

WrappedClass$VarName <- value

The comments in the function indicate that it is looking for
method names ending in "_set", but in fact it was searching for
methods ending in just "set", This was resulting in class methods
ending in set (names like GetOffset, SetOffset etc) being placed
in the wrong accessor functions and thus not being available in
the normal way.

There is probably a more appropriate way to tag the _set functions
when they are being created, which is a future step. This
patch makes the behaviour conform to the original intention.

### DIFF
--- a/Source/Modules/r.cxx
+++ b/Source/Modules/r.cxx
@@ -989,9 +989,10 @@ int R::OutputClassMemberTable(Hash *tb, File *out) {
     int klen = Len(key);
     int isSet = 0;
     if (klen > 4) {
-      ptr = &ptr[Len(key) - 4];
+      ptr = &ptr[klen - 4];
       isSet = strcmp(ptr, "_set") == 0;
     }
+
     //        OutputArrayMethod(className, el, out);        
     OutputMemberReferenceMethod(className, isSet, el, out);
     
@@ -1278,7 +1279,9 @@ void R::addAccessor(String *memberName, Wrapper *wrapper, String *name,
   if(isSet < 0) {
     int n = Len(name);
     char *ptr = Char(name);
-    isSet = Strcmp(NewString(&ptr[n-3]), "set") == 0;
+    if (n>4) {
+      isSet = Strcmp(NewString(&ptr[n-4]), "_set") == 0;
+    }
   }
   
   List *l = isSet ? class_member_set_functions : class_member_functions;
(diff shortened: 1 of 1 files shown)


blowekamp:
Is there any way to print a warning if an existing C++ method matches this name?

ojwb:
This seems bound to conflict with #834.

richardbeare:
It will definitely conflict - I think #834 is wrong - for the following reasons:

Check out the test results for preproc.i. It contains code like this:

``` c++
typedef struct EmbeddedDefines {
  int dummy;
#define  EMBEDDED_DEFINE 44
#ifdef SWIG
%constant EMBEDDED_SWIG_CONSTANT = 55;
#endif
} EmbeddedDefines;

```

The preproc.R file contains functions called 
``` r
EmbeddedDefines_dummy_set
EmbeddedDefines_dummy_get
```
which are called from 
``` r
setMethod('$<-', '_p_EmbeddedDefines', function(x, name, value)

{
  accessorFuns = list('dummy' = EmbeddedDefines_dummy_set);
  ;        idx = pmatch(name, names(accessorFuns));
  if(is.na(idx)) 
  return(callNextMethod(x, name, value));
  f = accessorFuns[[idx]];
  f(x, value);
  x;
}

```

This method allows code like this:
``` r
G <- EmbeddedDefines()
G$dummy <- 3
```

The trailing "set" search definitely shouldn't be changed to a leading "set" search, without breaking all this stuff. I don't think the idea is to use names to recognise method purposes in the user supplied code.

wsfulton:
@joequant i'd like to release swig-3.0.11 in the next week or so. Can you please resolve this one before then?

joequant:
I'll  take a look at it again this weekend.  I tried a general solution,
but the quick fix seems to be just make sure that special symbols are
something that's unlikely to be a duplicate.

wsfulton:
Hi @joequant, I think we are just about ready for the release. Is this change nearly ready?

blowekamp:
We would very much appreciate a solution to this issue for the next release. We have many transform classes in SimpleITK which are missing "SetOffset" methods in R. Thank you for your effort!

@richardbeare

wsfulton:
Thanks @joequant for sorting out.

PR #842, opened 2016-11-28
Make std::vector<> wrappers conform to List interface in Java

vadz:
There are two commits here, but the first one was just a refactoring and the second one contains the real changes that allow using `for (T obj: vec) ...` in Java for SWIG-generated vector wrappers.

git log:

--------------------------------------------------------------------------------
Make std::vector<> wrappers conform to List interface in Java

Derive the class wrapping std::vector<> in Java from java.util.AbstractList<>.
This makes it possible to use it with various algorithms working with Java
collections and, maybe even more importantly, makes it possible to iterate
over the wrapped vectors using for-each loops.

This commit is based on the original patch by Volker Diels-Grabsch from
https://sourceforge.net/p/swig/patches/278/

--------------------------------------------------------------------------------
Add helper macro to avoid duplication in Java vector typemaps

No real changes yet, just use a C#-like macro in Java std::vector<> typemaps
too to avoid having to repeat almost exactly the same class declaration for
the general case and for the bool specialization.

### DIFF
--- a/Lib/java/std_vector.i
+++ b/Lib/java/std_vector.i
@@ -9,13 +9,11 @@
 #include <stdexcept>
 %}
 
-namespace std {
-    
-    template<class T> class vector {
-      public:
+%define SWIG_STD_VECTOR_MINIMUM_INTERNAL(CTYPE, CREF_TYPE)
+    public:
         typedef size_t size_type;
-        typedef T value_type;
-        typedef const value_type& const_reference;
+        typedef CTYPE value_type;
+        typedef CREF_TYPE const_reference;
         vector();
         vector(size_type n);
         size_type size() const;
@@ -27,7 +25,7 @@ namespace std {
         %rename(add) push_back;
         void push_back(const value_type& x);
         %extend {
-            const_reference get(int i) throw (std::out_of_range) {
+            CREF_TYPE get(int i) throw (std::out_of_range) {
                 int size = int(self->size());
                 if (i>=0 && i<size)
                     return (*self)[i];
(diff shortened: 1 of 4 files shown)


Lib/java/std_vector.i
```diff
@@ -10,38 +10,124 @@
...
     public:
         typedef size_t size_type;
         typedef CTYPE value_type;
         typedef CREF_TYPE const_reference;
         vector();
         vector(size_type n);
-        size_type size() const;
+        // Note that it returns int, and not size_type, for compatibility with
+        // the signature of size() in Java interface.
+        int size() const;
```
ojwb: Shouldn't that really be `jint`, since that's the C type corresponding to Java `int`?

vadz: Hmm, I initially decided to use `int` for `size()` return value because it seemed that Java couldn't support vectors with more than `2^31` elements anyhow, but I've just found #674 and it's probably a good idea to throw an exception if the C++ vector is larger than that as that PR does...

wsfulton:
#670 is the patch earmarked for swig-3.1 in the wiki https://github.com/swig/swig/wiki/SWIG-3.1-Development. Let's keep the wiki up to date with the appropriate patches we should use for 3.1.

vadz:
Sorry, why should this wait for 3.1 too? It's backwards compatible, AFAICS, so what's the problem with applying it right now? I can fix the problem with `int` overflow in Java without touching #670.

vadz:
FYI: I've added a commit addressing the issue with C++ vectors with more than `2^31` elements.

The other commit just fixes a bug in the initial refactoring which resulted in Travis failure when using libc++ (the Appveyor failure of the initial CI build was spurious).

This would need to be rebase on the latest master to resolve the trivial conflict in the changelog, but other than that I think it could be merged right now, why not?

vadz:
I had to rebase this on master to avoid the conflict mentioned above which prevented the CI builds from running.

vadz:
Just FYI: AppVeyor build timed out, i.e. it's not a real failure.

To reiterate: I think this could (and should) be merged into the current master as it's a pretty big improvement for the users of the Java API.

wsfulton:
This patch is not compatible with pre-1.5 versions of the j2sdk which are currently okay and supported. Supporting 1.5 and later only is a reasonable documented change in swig-3.1 but not 3.0.x. I'm also keen we ramp up the 3.1 development so that we can get these sorts of changes released.

I think using jint is probably the best approach in the C++ layer.

`size_as_int` is probably overkill as in practice you aren't going to be able to use any containers that are greater in size than the limit of 2^31-1 due to memory restrictions (possibly vector<char> though is possible). We also cast size_t in other places to long, so I'd say this is a much bigger and a more real problem that should be fixed more generally if the desire is there to solve this problem. A typemap fragment should then be used to implement it.

Issue #615, opened 2016-02-16
Test overload_numeric failed against GCC 6


jplesnik:
I tried to rebuild swig 3.0.8 against GCC 6, but the python testcase overload_numeric  failed. I got following error:

```
checking python testcase overload_numeric (with run test)
Traceback (most recent call last):
  File "overload_numeric_runme.py", line 37, in <module>
    check(nums.over(float("inf")), "float")
  File "overload_numeric_runme.py", line 11, in check
    raise RuntimeError("got: " + got + " expected: " + expected)
RuntimeError: got: double expected: float
Makefile:103: recipe for target 'overload_numeric.cpptest' failed
make[1]: *** [overload_numeric.cpptest] Error 1
```

GCC 6 is part of the Fedora Rawhide.

The test passed with GCC 5.3.1.

wsfulton:
which version of Python?

ojwb:
Works for me on Debian unstable x86-64 with:

```
g++-6 (Debian 6-20160117-1) 6.0.0 20160117 (experimental) [trunk revision 232481]
Python 2.7.11
```

I wonder if it's an optimisation issue - @jplesnik: can you try:

```
make -C Examples/test-suite/python overload_numeric.cpptest CXXFLAGS='-fno-strict-aliasing -fwrapv'
```

Python says we should use these flags:

```
$ python-config --cflags
-I/usr/include/python2.7 -I/usr/include/x86_64-linux-gnu/python2.7  -fno-strict-aliasing -Wdate-time -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security  -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes
```

A newer GCC version might well optimise more based on alias analysis or integer overflow analysis.  We should probably be asking python what flags to compile the C/C++ code with in configure.

jplesnik:
Version of Python version is
```
Python 2.7.11
```

The version of GCC is
```
gcc version 6.0.0 20160212 (Red Hat 6.0.0-0.11) (GCC) 
```

@ojwb: I tried the command, but it failed 

```
$ make -C Examples/test-suite/python overload_numeric.cpptest CXXFLAGS='-fno-strict-aliasing -fwrapv'
make: Entering directory '/root/swig/Examples/test-suite/python'
[...]

PR #788, opened 2016-09-10
Add missing #include <cmath> for std::isfinite

AMDmi3:
Using `std::isfinite()` requires `#include <cmath>`, which doesn't exist in surrounding code. This results, for instance, in build failure of [ZNC](http://znc.in/) on FreeBSD 11.x (clang 3.8, libc++):

```
modperl/ZNC.cpp:2573:9: error: no member named 'isfinite' in namespace 'std'; did you mean simply 'isfinite'?
    if (SWIG_Float_Overflow_Check(v)) {
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
modperl/ZNC.cpp:2561:73: note: expanded from macro 'SWIG_Float_Overflow_Check'
# define SWIG_Float_Overflow_Check(X) ((X < -FLT_MAX || X > FLT_MAX) && SWIG_isfinite(X))
                                                                        ^~~~~~~~~~~~~~~~
modperl/ZNC.cpp:2549:29: note: expanded from macro 'SWIG_isfinite'
#  define SWIG_isfinite(X) (std::isfinite(X))
                            ^~~~~
/usr/include/c++/v1/math.h:380:1: note: 'isfinite' declared here
isfinite(_A1 __lcpp_x) _NOEXCEPT
^
2 warnings and 1 error generated.
```

full log: http://beefy1.nyi.freebsd.org/data/110i386-quarterly/421620/logs/errors/znc-1.6.3.log

PR #849, opened 2016-12-16
Attempt to fix isfinite() checks to work with all C++11 compilers

ojwb:
Fixes https://github.com/swig/swig/issues/615 and
https://github.com/swig/swig/pull/788


wsfulton:
i like it,  very neat solution. There is one potential problem I can see though. If the std namespace is not defined (which it may not be with just the math.h header include),  then the `using namespace std;` will give a compiler error. This can be solved by adding in `namespace std {}` into the fragment, but I suspect that SWIG unfortunately, includes enough STL headers that std is defined somewhere, so may not be a problem in practice, even with with a minimal example.

Also, I think you should add SWIGRUNTIME to SWIG_isfinite_func to avoid linking problems when using multiple modules and %import.

ojwb:
Good point about `namespace std { }`.  For the second point, might be better to just make it inline, since it's just a forwarding wrapper so definitely wants to be inlined.  Will address later this weekend (or feel free to push a commit to the branch).

ojwb:
> If the std namespace is not defined (which it may not be with just the `math.h` header include), then the `using namespace std;` will give a compiler error.

FWIW, this does not actually seem to result in a compiler error (tested with GCC and clang), though clang gives a warning which we want to avoid:

```
$ echo 'using namespace std;'|clang++-4.0 -Wall -W -x c++ -c -o /dev/null -
<stdin>:1:17: warning: using directive refers to implicitly-defined namespace
      'std'
using namespace std;
                ^
1 warning generated.
```

wsfulton:
Let's go with this then. Indentation is conventionally 2 not 4 spaces though.

PR #852, opened 2016-12-17
Fixed potential null pointer dereference in Python binding

ezralanglois:
There is no check as to whether args is null. 

Right below this line, a check is performed.
This check is also performed on line 2932.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -2542,7 +2542,7 @@ public:
     if (!fastunpack) {
       Wrapper_add_local(f, "ii", "Py_ssize_t ii");
       if (maxargs - (add_self ? 1 : 0) > 0)
-	Append(f->code, "if (!PyTuple_Check(args)) SWIG_fail;\n");
+	Append(f->code, "if (args && !PyTuple_Check(args)) SWIG_fail;\n");
       Append(f->code, "argc = args ? PyObject_Length(args) : 0;\n");
       if (add_self)
 	Append(f->code, "argv[0] = self;\n");


ezralanglois:
I am digging into why the director_thread_runme.py test segfaulted as the result of this change.

wsfulton:
The patch looks reasonable, but what command line options did you use to cause a null pointer dereference?

The director_thread test sometimes fails. Clearly there is some sort of undefined behaviour in this test, I'm not sure what it is, but it would be great to know what if you do take a look.

ezralanglois:
This code was written out with the default options when using the Python language.

wsfulton:
Okay, thanks. This problem has not appeared in our test-suite. Could you provide a snippet of code causing the problem?

ezralanglois:
I will try. I found this originally by running coverity over my own code, which uses SWIG. Coverity pointed out the issue, and then I looked over what SWIG was doing and saw that in most cases, it was checking whether args was null before attempting to dereference it.

ojwb:
It'd definitely be good to have a regression test for something like this, as otherwise there's a real risk of regression when the mess of current python options is cleaned up for 3.1.

Did you actually hit a segfault here in practice though?

It could be the static checker just spotted that there was a `NULL` check in the code following this, and inferred that `args` could be `NULL` from that, in which case the lack of a `NULL` check prior to that is rather suspicious.  But the reality could be that the later `NULL` check is unnecessary.

The documented cases where Python can pass `NULL` to us are with `METH_NOARGS` or `METH_OLDARGS`:

https://docs.python.org/2/c-api/structures.html#c.PyMethodDef

AFAICS (`git grep METH_OLDARGS` finds no references at least) the latter is never generated by SWIG:

SWIG can generate code using `METH_NOARGS` (e.g. with `-fastunpack -modernargs`).  However, all the option combinations which I can see which produce this involve `-fastunpack` and the code this PR patches is only active when `-fastunpack` is **NOT** in use.

If my analysis is correct, I'd suggest we generate `assert(args);` for cases where we know it shouldn't be `NULL`, and then try to clean up the unnecessary `NULL` checks rather than adding yet more of them.

ezralanglois:
I agree. This case is already being tested by compactdefaultargs_runme.py and it does not get a null pointer. 

The problem was the extraneous null checks. I was (along with the static analysis tool) confused by those. I will close this out as not a problem.

wsfulton:
SWIG can use Coverity for free, but because it is not run automatically in Travis, it needs a manual invocation and hence is hardly ever run. I'd quite like to see it set up as part of automated builds, say on a once a week basis if anyone feels inspired to do this. There were just a handful of potential problems from what I recall when it was last run. The Travis configuration is set up on the coverity-scan branch: https://github.com/swig/swig/blob/coverity-scan/.travis.yml.

ezralanglois:
According to the travis-ci docs,  cron jobs are supported
https://docs.travis-ci.com/user/cron-jobs/

I imagine what you suggested could be done with a script that merges master into the coverity scan branch, or, alternatively, resets the coverity scan branch to master/HEAD.

Then the cron job can be setup in the travis-ci settings.

I am still a novice with coverity, but I believe there is a significant limitation on who can view the defects.

PR #853, opened 2016-12-18
Renamed all operators beginning with underscore.

ivannp:
Hello,

In R, identifier cannot start with an underscore. I renamed the operators.

FYI: This is my first time looking into swig, so be extra careful with the pull request - I am not sure how to verify the fix further.

### DIFF
--- a/Examples/r/class/example.cxx
+++ b/Examples/r/class/example.cxx
@@ -19,8 +19,35 @@ double Circle::perimeter() {
   return 2*M_PI*radius;
 }
 
+Circle::Circle(double xx, double yy, double rr)
+  : radius(rr)
+{
+  x = xx;
+  y = yy;
+}
+
+bool Circle::operator==(const Circle & other)
+{
+  return x == other.x && y == other.y && radius == other.radius;
+}
+
+bool Circle::operator!=(const Circle & other)
+{
+  return !operator==(other);
+}
+
 double Square::area() {
   return width*width;
+}
+
+bool Square::operator==(const Square & other)
+{
+  return x == other.x && y == other.y && width == other.width;
+}
+
+bool Square::operator!=(const Square & other)
+{
+  return !operator==(other);
 }
 
 double Square::perimeter() {
--- a/Examples/r/class/example.h
+++ b/Examples/r/class/example.h
@@ -20,8 +20,12 @@ private:
   double radius;
 public:
   Circle(double r) : radius(r) { }
+  Circle(double xx, double yy, double rr);
   virtual double area();
   virtual double perimeter();
+
+  bool operator==(const Circle & other);
+  bool operator!=(const Circle & other);
 };
 
 class Square : public Shape {
@@ -31,4 +35,7 @@ public:
   Square(double w) : width(w) { }
   virtual double area();
   virtual double perimeter();
+
+  bool operator==(const Square & other);
+  bool operator!=(const Square & other);
 };
--- a/Examples/r/class/runme.R
+++ b/Examples/r/class/runme.R
@@ -31,6 +31,14 @@ print("Here is their current position:")
 sprintf("    Circle = (%f, %f)", circle$x,circle$y)
 sprintf("    Square = (%f, %f)", square$x,square$y)
 
+c2 <- Circle(1, 5, 15)
+c1 <- circle
+c1$move(1, 5)
+print(Circle_MemberEq(c1, c2))
+print(Circle_MemberNotEq(c1, c2))
+
+print(c1$MemberEq(c2))
+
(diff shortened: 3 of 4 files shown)


wsfulton:
This isn't the right approach. If leading underscores are problematic in R then, yes,
the renames that were adding in underscores need fixing. The renames with
underscores are originally from Python where the names used have special meaning
in Python, for example `__getitem__` is called when indexing into arrays such as
`x[0]`. The %rename philisophy in SWIG is to only rename operators if there is
a special name with the equivalent meaning in the target language. If there
isn't then there should be no rename. So for R, do the new names in this patch
have special meaning or convention? I guess not, in which case they should all be
removed and a user is then responsible for providing a %rename using a name of their
preference as is done in the other target languages.

ivannp:
I am new to SWIG, but IMO it's better to have some defaults rather than leaving everything for the user to boot? It is overwhelming, especially when documentation is succinct (R is not as rich as Python).

wsfulton:
I'm afraid that SWIG's approach is to do a lot for you, but not everything as sometimes there is no one correct answer. It is a semi-automated tool and some things where it isn't obvious require you to write some code. It will probably take you 5 minutes to put these renames in your own interface file, so it isn't a hard ask, SWIG even gives you the appropriate warnings to tell you what operator overloads need a %rename.

I'll revert the patch shortly and remove the %rename statements that need fixing.

Patches to the documentation will be much appreciated in this case where you think it could be improved.

ivannp:
I still don't get it why defaults are bad, but apparently it doesn't matter.

wsfulton:
Ah sorry if my explanation is not clear, it does matter you understand the philisophy. Not sure how much more I can explain it other than adding in consistency amongst languages is important and the names are provided only where they make sense as explained.

wsfulton:
Sorry if the explanation wasn't all that clear, hopefully this helps... What seems to be helpful isn't always. Adding in a default name could easily clash with other members in the class or override base class members leading to surprising and obscure bugs, so the SWIG philosophy is to make the user do a little bit of extra work at the cost of a little inconvenience. Consistency amongst all the languages is a high priority and getting consistent additional names that will keep all users happy in all languages with regard to naming conventions is all but impossible, so it is left to the user to provide the names unless there is a correct/conventional name for the operator overload in a target language.

joequant:
FYI, the reason I accepted the patch is that it turns out that starting a
function name in R is not allowed, and causes all sorts of odd problems,
since what you end up with is a function that works some of the time, but
not others.

https://stat.ethz.ch/pipermail/r-help/2006-June/106869.html

The fact that the SWIG modules for R had functions with underscores at the
beginning was an oversight.  While there may be good reasons to not rename
the operators to something else, since it causes a rename to something that
isn't language acceptable, something needs to be done.

Thoughts?

joequant:
Actually the way that it *should* work is to rewrite the overloads so that
the R operators get overloaded

http://stackoverflow.com/questions/5322546/operation-overloading-in-r

This is what in fact happens with python and the other modules.  __plus__
is a special name in python so that "+" works.

Ivan, can you experiment with R to see if you can get the c++ operator "+"
to map to the R operator "+".  That's the right thing to do.

ojwb:
> something needs to be done

It already was - this commit removed the renaming to the underscored names: 94e7482e48186c51fe7591a60eaa09f36616e0f8

ivannp:
@joequant 

I don't think this is possible with the current design for R bindings in SWIG. See [issue 854](https://github.com/swig/swig/issues/854), which I created.

First, with the new lack of MemberPlus, what happens is Circle::operator+ is mapped to Plus (the global +). Not a big deal, just a lot more to explain to the user after [commit 94e7482](https://github.com/swig/swig/commit/94e7482e48186c51fe7591a60eaa09f36616e0f8).

The rest of the problem is the class approach taken by swig. S4 requires the operator '+' to be overloaded using setMethod:

    setMethod('+', signature(e1='_p_Circle', e2='_p_Circle'), function(e1, e2) e1$MemberPlus(e2))

But setMethod is not used currently, except for delete and a few other pre-defined things. I think SWIG should move to S4 dispatching for all class methods. There are slightly more details in the above-mentioned issue.

wsfulton:
The remaining operators with pre-chosen names in ropers.swg should also be removed as they have no special meaning in R. I contemplated doing this as part of 94e7482 but as these were not causing problems like the leading underscores and may already be used by R users, I decided not to break backwards compatibility in a point release. I suggest we remove them in swig-3.1.x if the proper R operator overloading is not added in by then. Clearly this area needs some work done to fit in with how SWIG conventions should be implemented and even better, proper R operator overloads using the ideas of @ivannp.

PR #860, opened 2016-12-29
Remove inheritance from std::unary_function and std::binary_function,

ajrheading1:
they are deprecated in C++11 and already removed in Visual C++ '15'
running with /std:c++latest

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3145.html

### DIFF
--- a/Lib/octave/octcontainer.swg
+++ b/Lib/octave/octcontainer.swg
@@ -68,7 +68,7 @@ namespace swig {
 
 namespace std {
   template <>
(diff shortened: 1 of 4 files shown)


ojwb:
CI failure is (as usual) just travis being flaky.

wsfulton:
I thought we could add in the typedefs (argument_type, first_argument_type etc) that would be needed given the removed inheritance, however, we don't seem to use these typedefs, so have accepted the patch as is. Thanks.

PR #861, opened 2016-12-31
Fix some typos in docs and examples and make the code look nicer.

sunoru:
I was reading the manual for adding support for new languages, and noticed there were some small typos, so I fixed them in passing. Mainly in "Doc/Manual/Extending.html" and the files under "Examples/python".


wsfulton:
Thanks, I made a couple of small corrections and merged to master.

PR #864, opened 2017-01-10
Redundant typedef fix

### LINKED ISSUE #838: Incorrect generated code when tag name and typedef name are identical.
Consider the following interface file:
```
%module Broken

%inline {
typedef struct Test Test;
void function(Test* test);
struct Test {
      int foo;
};
}
```

This produces the following module class:
```
public class Broken {
  public static void function(SWIGTYPE_p_Test test) {
    BrokenJNI.function(SWIGTYPE_p_Test.getCPtr(test));
  }
}
```

It seems to be assuming that the Test struct only has a forward declaration, despite providing an implementation later. The Test class is generated in Java, but `function` doesn't make use of it.

I'm happy to take a look at fixing this bug, if anyone has any pointers of where to look in the SWIG codebase?

davedissian:
This pull request fixes #838 

I determined that `Language::classLookup` seems to fail when there exists a `cdecl` in the symbol table where the name is identical to another struct/class in the symbol table. As SWIG doesn't have a tag namespace, a typedef such as `typedef struct Foo Foo` can simply be ignored at the parsing stage, which prevents the issue all together.

Added a test case which will attempt to compile some Java code and see if the proxy class is set up correctly.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,12 +7,16 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 3.0.12 (in progress)
 ============================
 
+2017-01-10: davedissian
+        Issue #838 - A typedef where the "name" matches the "type" no longer
+        causes symbol lookups to fail when generating code.
+
 2016-12-31: ajrheading1
-	    Issue #860 - Remove use of std::unary_function and std::binary_function
+        Issue #860 - Remove use of std::unary_function and std::binary_function
             which is deprecated in C++11.
 
 2016-12-30: olly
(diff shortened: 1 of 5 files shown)


wsfulton:
I'm not comfortable with this approach. It removes the typedef from the parse tree which would be rather surprising for things like using `-xml`. It feels like something more clever should be done when adding the symbol into the symbol table or maybe fixing typedef lookup (Swig_typedef_resolve_all). Maybe some inspiration can be obtained from forward class references which also use the same name as a subsequent full class definition.

davedissian:
Hmmm, fair point. I'll take a further look in the coming days.

wsfulton:
Please re-open when you have a new patch.

Issue #865, opened 2017-01-10
Javascript array template not fully applied


nshmyrev:
The declaration part of the fragment of arrays_javascript.i is not fully applied:

```
%typemap(in, fragment="SWIG_JSCGetIntProperty") int[], int[ANY]
    (int length = 0, v8::Local<v8::Array> array, v8::Local<v8::Value> jsvalue, int i = 0, int res = 0, $*1_ltype temp) {
  if ($input->IsArray())
```

This is a code it generates:

```
static SwigV8ReturnValue _wrap_AudioIndex_lookup(const SwigV8Arguments &args) {
  SWIGV8_HANDLESCOPE();

  v8::Handle<v8::Value> jsresult;
  AudioIndex *arg1 = (AudioIndex *) 0 ;
  unsigned int *arg2 = (unsigned int *) 0 ;
  int arg3 ;
  void *argp1 = 0 ;
  int res1 = 0 ;
  int val3 ;
  int ecode3 = 0 ;
  LookupResult *result = 0 ;
    
  if(args.Length() != 2) SWIG_exception_fail(SWIG_ERROR, "Illegal number of arguments for _wrap_AudioIndex_lookup.");
    
  res1 = SWIG_ConvertPtr(args.Holder(), &argp1,SWIGTYPE_p_AudioIndex, 0 |  0 );
  if (!SWIG_IsOK(res1)) {
    SWIG_exception_fail(SWIG_ArgError(res1), "in method '" "AudioIndex_lookup" "', argument " "1"" of type '" "AudioIndex *""'");
  }
  arg1 = (AudioIndex *)(argp1);
  {
    if (args[0]->IsArray())
    {
      // Convert into Array
      array = v8::Local<v8::Array>::Cast(args[0]);
  
      length = arg2_dim0;
  
```

So a lot of errors in compilation like

```
asign_wrap_js.cpp: In function 'SwigV8ReturnValue _wrap_AudioIndex_lookup(const SwigV8Arguments&)':
asign_wrap_js.cpp:2350:7: error: 'array' was not declared in this scope
       array = v8::Local<v8::Array>::Cast(args[0]);
       ^
asign_wrap_js.cpp:2352:7: error: 'length' was not declared in this scope
       length = arg2_dim0;
       ^
asign_wrap_js.cpp:2352:16: error: 'arg2_dim0' was not declared in this scope
       length = arg2_dim0;
                ^
asign_wrap_js.cpp:2357:12: error: 'i' was not declared in this scope
[...]

PR #872, opened 2017-01-12
[Javascript] Proper array typemaps in Javascript

nshmyrev:
Fixes issue https://github.com/swig/swig/issues/865


Lib/javascript/jsc/arrays_javascript.i
```diff
@@ -21,103 +21,70 @@
...
+    JSObjectRef array = JSValueToObject(context, $input, NULL);
 
-    length = $1_dim0;
+    int length = ANYLENGTH;
 
     $1  = ($*1_ltype *)malloc(sizeof($*1_ltype) * length);
 
     // Get each element from array
-    for (i = 0; i < length; i++)
+    for (int i = 0; i < length; i++)
```
ojwb: William's keen to keep C output compatible with C90, which means you can't declare loop variables like this.

Lib/javascript/jsc/arrays_javascript.i
```diff
@@ -21,103 +21,70 @@
...
         SWIG_exception_fail(SWIG_ERROR, "Failed to convert $input to double");
       }
       arg$argnum[i] = temp;
     }
-
   }
   else
   {
-    SWIG_exception_fail(SWIG_ERROR, "$input is not JSObjectRef");
+    SWIG_exception_fail(SWIG_ERROR, "$input is not an JSObjectRef");
```
ojwb: "an"?

Examples/javascript/array/runme.js
```diff
@@ -0,0 +1,7 @@
+var example = require("example");
+
+var arr = [1, 2, 3, 4, 5];
+
+console.log(example.sum1(arr, arr.length));
+console.log(example.sum2(arr));
+console.log(example.data3)
```
ojwb: This doesn't fail if the results are incorrect, it just writes it to the console and without any context there's basically no chance anyone will notice if they're wrong.

nshmyrev: All other examples in javascript do that as well. It checks that things compile with node, it will fail if not compiled.

ojwb: Non-automated tests are much less useful, as we won't catch regressions right away (and probably not until somebody finds their SWIG-generated wrappers broke).

If you don't want to make code under `Examples/javascript` self-check its results, then please put this in the main test-suite under `Examples/test-suite` instead.

nshmyrev: Ok, I'll check in coming days.

wsfulton:
Heads up, I'm modifying this patch to work with modern node version so I can merge it in.

PR #875, opened 2017-01-13
Octave: add support for version 4.2

kwwette:
This pull request add support for Octave version 4.2 to SWIG. The Travis build has been upgraded to trusty, and tests against versions 3.8 (the system version), 4.0, and 4.2 (from PPA kwwette/octaves). C++11 is now required to build the Octave examples and test suite. There are some potential incompatibilities due to Octave dropping support for << and >> operators.

### DIFF
--- a/Lib/octave/octruntime.swg
+++ b/Lib/octave/octruntime.swg
@@ -1,7 +1,93 @@
 %insert(runtime) %{
+
 #include <cstdlib>
 #include <iostream>
+
 #include <octave/oct.h>
+#include <octave/version.h>
+
+// Macro for enabling features which require Octave version >= major.minor.patch
+// - Use (OCTAVE_PATCH_VERSION + 0) to handle both '<digit>' (released) and '<digit>+' (in development) patch numbers
+#define SWIG_OCTAVE_PREREQ(major, minor, patch) \
+  ( (OCTAVE_MAJOR_VERSION<<16) + (OCTAVE_MINOR_VERSION<<8) + (OCTAVE_PATCH_VERSION + 0) >= ((major)<<16) + ((minor)<<8) + (patch) )
+
+// Reconstruct Octave major, minor, and patch versions for releases prior to 3.8.1
+#if !defined(OCTAVE_MAJOR_VERSION)
+
+# if !defined(OCTAVE_API_VERSION_NUMBER)
+
+// Hack to distinguish between Octave 3.8.0, which removed OCTAVE_API_VERSION_NUMBER but did not yet
+// introduce OCTAVE_MAJOR_VERSION, and Octave <= 3.2, which did not define OCTAVE_API_VERSION_NUMBER
+#  include <octave/ov.h>
+#  if defined(octave_ov_h)
+#   define OCTAVE_MAJOR_VERSION 3
+#   define OCTAVE_MINOR_VERSION 8
+#   define OCTAVE_PATCH_VERSION 0
+#  else
+
+// Hack to distinguish between Octave 3.2 and earlier versions, before OCTAVE_API_VERSION_NUMBER existed
+#   define ComplexLU __ignore
+#   include <octave/CmplxLU.h>
+#   undef ComplexLU
+#   if defined(octave_Complex_LU_h)
+
+// We know only that this version is prior to Octave 3.2, i.e. OCTAVE_API_VERSION_NUMBER < 37
+#    define OCTAVE_MAJOR_VERSION 3
+#    define OCTAVE_MINOR_VERSION 1
+#    define OCTAVE_PATCH_VERSION 99
+
+#   else
+
+// OCTAVE_API_VERSION_NUMBER == 37
+#    define OCTAVE_MAJOR_VERSION 3
+#    define OCTAVE_MINOR_VERSION 2
+#    define OCTAVE_PATCH_VERSION 0
+
+#   endif // defined(octave_Complex_LU_h)
+
+#  endif // defined(octave_ov_h)
+
+// Correlation between Octave API and version numbers extracted from Octave's
+// ChangeLogs; version is the *earliest* released Octave with that API number
+# elif OCTAVE_API_VERSION_NUMBER >= 48
+#  define OCTAVE_MAJOR_VERSION 3
+#  define OCTAVE_MINOR_VERSION 6
+#  define OCTAVE_PATCH_VERSION 0
+
+# elif OCTAVE_API_VERSION_NUMBER >= 45
+#  define OCTAVE_MAJOR_VERSION 3
+#  define OCTAVE_MINOR_VERSION 4
+#  define OCTAVE_PATCH_VERSION 1
+
+# elif OCTAVE_API_VERSION_NUMBER >= 42
+#  define OCTAVE_MAJOR_VERSION 3
+#  define OCTAVE_MINOR_VERSION 3
+#  define OCTAVE_PATCH_VERSION 54
+
+# elif OCTAVE_API_VERSION_NUMBER >= 41
+#  define OCTAVE_MAJOR_VERSION 3
+#  define OCTAVE_MINOR_VERSION 3
+#  define OCTAVE_PATCH_VERSION 53
+
+# elif OCTAVE_API_VERSION_NUMBER >= 40
+#  define OCTAVE_MAJOR_VERSION 3
+#  define OCTAVE_MINOR_VERSION 3
+#  define OCTAVE_PATCH_VERSION 52
+
+# elif OCTAVE_API_VERSION_NUMBER >= 39
+#  define OCTAVE_MAJOR_VERSION 3
+#  define OCTAVE_MINOR_VERSION 3
+#  define OCTAVE_PATCH_VERSION 51
+
+# else // OCTAVE_API_VERSION_NUMBER == 38
+#  define OCTAVE_MAJOR_VERSION 3
+#  define OCTAVE_MINOR_VERSION 3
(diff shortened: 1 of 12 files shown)


besser82:
@kwwette Unfortunately Perl5 tests are failing, if built in a row with your changes for Octave 4.2…  :(

```
checking perl5 testcase cpp11_raw_string_literals
./../cpp11_raw_string_literals.i:20: Error: Unable to find 'std_wstring.i'
make[2]: *** [../../../Examples/Makefile:257: perl5_cpp] Error 1
make[1]: *** [Makefile:36: cpp11_raw_string_literals.cpptest] Error 2
```

Looks like it works fine, if one copies over `std_wstring.i` from the ocatave to perl5-dir…

besser82:
Same goes for php

```
checking php testcase cpp11_raw_string_literals
./../cpp11_raw_string_literals.i:20: Error: Unable to find 'std_wstring.i'
make[2]: *** [../../../Examples/Makefile:1191: php_cpp] Error 1
make[1]: *** [Makefile:38: cpp11_raw_string_literals.cpptest] Error 2
```

besser82:
And R, Lua, Go…  Possibly some others, too…

kwwette:
Thanks @besser82 - I guess this is because Octave now turns on C++11, but C++11 support hasn't been added for all SWIG languages yet.

It's probably better if SWIG turns on C++11 just for Octave, so that this doesn't disturb other languages. I'll have a look at doing this and then update the pull request.

kwwette:
@besser82 - this should now be fixed in the current pull request. By adding C++11 support for Octave I was accidentally enabling the C++11 tests in the test suite, but this should no longer happen, i.e. the C++11 tests are only enabled by configuring with `--enable-cpp11-testing`.

@wsfulton - once these changes are merged, would it be possible to roll a new SWIG release? While I appreciate that the last release was very recent, it would be very useful to have a released SWIG with Octave 4.2 support, so that it could be packaged on the same timescale as Octave 4.2 itself. In particular, Octave 4.2 is already on MacPorts [1], so the project I work on which uses SWIG cannot (easily) support Octave on Macs until the next SWIG release.

[1] https://www.macports.org/ports.php?by=name&substr=octave

besser82:
@kwwette Thank you!  Now it works.  =)

wsfulton:
Is C++11 really needed for Octave 4.2? I don't see this as a requirement in the
4.2 release notes at https://www.gnu.org/software/octave/NEWS-4.2.html.

Which version of autoconf is required to support: AX_CXX_COMPILE_STDCXX_11?  Our configure.ac has a minimum version requirement set to 2.58. If Octave 4.2 really requires C++11, is it absolutely necessary to dynamically test this? Can you not just code in a version detection and use that to determine whether or not a c++11 compiler is required?

I don't mind releasing this when it has been merged to master. I'll email the swig-devel list with advance warning.

ojwb:
@wsfulton `AX_CXX_COMPILE_STDCXX_11` isn't provided by autoconf at all (as the `AX_` prefix indicates - it would be `AC_` if it were from autoconf).  It's already in our source tree in `Tools/config/ax_cxx_compile_stdcxx_11.m4` and that macro is already used you configure with `--enable-cpp11-testing`.

besser82:
Anyways…  If there is need for a C++11-compiler when compiling a generated wrapper, there should be compile-time detection in the swig-generated code, too.

```
#if __cplusplus < 201103L
#    error This Wrapper for Octave 4.2+ can only be compiled with a compiler that supports C++11
#endif
```

wsfulton:
I still don't understand why a C++11 compiler is needed. What has caused this requirement? SWIG's usual approach is definitely not to force an upgrade to C++11 on users and I'd be somewhat surprised if the Octave API requires a minimum of C++11.

besser82:
@wsfulton From the [Octave-devel mailinglist](http://octave.1599824.n4.nabble.com/Will-version-4-2-require-C-11-td4677055.html):

```
> 5/20/16 
> 
> jwe, 
> 
> I notice that the 'auto' keyword is now used in 5 places in liboctave: 
> 
> system/lo-sysdep.h:const auto octave_getcwd = octave::sys::getcwd; 
> system/lo-sysdep.h:const auto octave_chdir = octave::sys::chdir; 
> system/lo-sysdep.h:const auto octave_popen2 = octave::sys::popen2; 
> util/oct-glob.h:const auto octave_fnmatch = octave::sys::fnmatch; 
> util/oct-glob.h:const auto octave_glob = octave::sys::glob; 
> 
> That keyword is a C++11 feature.  Are we going to move to require that 
> feature level for the development branch?

I still think that it makes sense to use some restraint with C++11 
features, but I'm OK with using auto for things like this. 

OTOH, using auto to make aliases won't work for overloaded function 
names.  In those cases I just used inline wrapper functions.  If there 
is some problem with auto then we could always just use inline wrappers 
for these cases as well. 

We are already using unique_ptr unconditionally and I haven't seen a lot 
of complaints about that.  So my guess is that C++11 is not causing a 
lot of trouble. 

jwe 
```

So Octave 4.2 requires some C++11 features…

Lib/octave/octruntime.swg
```diff
@@ -211,10 +318,11 @@ DEFUN_DLD( SWIG_name, args, nargout, SWIG_name_usage ) {
 
   // workaround to prevent octave seg-faulting on exit: set Octave exit function
   // octave_exit to _Exit, which exits immediately without trying to cleanup memory.
-  // definitely affects version 3.2.*, not sure about 3.3.*, seems to be fixed in
-  // version 3.4.* and above. can be turned off with macro definition.
+  // definitely affecteds version 3.2.*, not sure about 3.3.*, seems to be fixed in
```
wsfulton: Small typo here

wsfulton:
@besser82, thanks for clarification.
@ojwb thanks for reminding me, I should have remembered, I even introduced the original use of that macro in SWIG. http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=history;f=m4/ax_compare_version.m4 has quite a few improvements to our version, so I'll look at updating it.
@kwwette Please merge to master.

Lib/octave/octruntime.swg
```diff
@@ -1,16 +1,110 @@
...
+#if SWIG_OCTAVE_PREREQ(4,2,0)
+#include <octave/interpreter.h>
+#else
 #include <octave/toplev.h>
+#endif
 #include <octave/unwind-prot.h>
+#if SWIG_OCTAVE_PREREQ(4,2,0)
+#include <octave/call-stack.h>
+#endif
+
```
besser82: Somewhere at this point there should be a compile-time check for C++11, when generated code for Octave 4.2+ is compiled.

```
#if SWIG_OCTAVE_PREREQ(4,2,0)
#  if __cplusplus < 201103L
#    error A compiler supporting C++11 is required to compile extensions for Octave 4.2.0 or later.
#  endif
#endif
```

ojwb:
There's a difference between needing C++11 to build Octave and needing C++11 to build code against the Octave API.  The former doesn't imply the latter (though of course both may be true).  The quote from octave-devel is really only evidence of the former.

Checking `#if __cplusplus < 201103L` may exclude some compilers which lack complete C++11 support but are actually entirely capable of compiling the code - e.g. `g++-4.6 -std=c++0x` defines `__cplusplus` to `1` but has pretty good C++11 support (https://gcc.gnu.org/gcc-4.6/cxx0x_status.html) - e.g. `auto` and `unique_ptr` were mentioned above and are both supported.

wsfulton:
Agreed with @ojwb, adding the `__cplusplus` check will be unnecessarily restrictive. Given the real restriction to C++11 features is due to using the Octave headers, I feel compiler suitability checks ought to be in the Octave headers. Besides, anyone using SWIG needs to understand C++ and compiler errors and what they mean, so there is little value added trying to say in different words that the C++ compiler isn't suitable. Documenting in Octave.html though in addition to the CHANGES.current file would be helpful though.

Lib/octave/octruntime.swg
kwwette: Fixed, thanks.

kwwette:
@wsfulton - as pointed out by others, the C++11 requirement is because the Octave API now uses some C++11 features.

However, I've now realised that SWIG should not have to check for C++11 itself. Octave supplies a helper program `mkoctfile` for building Octave extensions, and this program should already supply whatever flags are required to build against the Octave API. `./configure` already uses `mkoctfile` to extract the required `CXXFLAGS`, `LDFLAGS`, etc., but it wasn't picking up any `-std=c++11` flag. It turned out this flag was part of the compiler command used by `mkoctfile`, which is given by `mkoctfile -p CXX`, so I've modified `./configure` to parse this output for any `-std=*` flags.

In short, there is no longer a C++11 `./configure` check for Octave 4.2.

I've updated `CHANGES.current` to point out that the Octave API now uses some C++11 features, and remind users to either use `mkoctfile` to ensure they get the right C++ compiler/options, or otherwise make sure they parse the output of `mkoctfile -p CXX`. The Octave documentation already uses `mkoctfile` in the tutorial on building an Octave extension.

I'll wait for the CI tests to pass for this pull request, then merge it to master.

Lib/octave/octruntime.swg
kwwette: As others have pointed out, this is probably too restrictive. Also see my comment [below](https://github.com/swig/swig/pull/875#issuecomment-274294445) - if users use the `mkoctfile` program supplied by Octave to compile extensions, they shouldn't need to worry about what kind of C++ compiler Octave requires.

PR #878, opened 2017-01-20
Improve advice for finding the Ruby headers include directories

vadz:
Don't advise using "$:", which is for Ruby models, and doesn't include/map to
the directories containing the headers. Advise using rbconfig module instead.

The new instructions are known to work out of the box both under Linux and
macOS.

[skip ci]

### DIFF
--- a/Doc/Manual/Ruby.html
+++ b/Doc/Manual/Ruby.html
@@ -192,28 +192,13 @@ to compile this file and link it with the rest of your program. </p>
 
 
 <p> In order to compile the wrapper code, the compiler needs the <tt>ruby.h</tt>
-header file. This file is usually contained in a directory such as </p>
-
-<div class="code shell diagram">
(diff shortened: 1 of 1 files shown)


wsfulton:
Yes, thanks, this is better, I fixed a typo. We use `ruby -rrbconfig` instead of `ruby -e 'require rbconfig'`. I don't know which is more standard Ruby, but probably doesn't matter as they amount to the same thing!

wsfulton:
Merged in 93c1d531f55778edd23efae2b5d544cb03fbfdec

vadz:
Yes, `ruby -rrbconfig` is arguably better because more clear/shorter/doesn't require an extra pair of nested quotes. I didn't use it simply because I didn't know about it (as about almost everything else in Ruby... which, thanks to SWIG, doesn't prevent me from providing API wrappers in it!).

wsfulton:
Okay, I just switched it to `-rrbconfig`.

PR #879, opened 2017-01-20
CFFI enhancements

gitfoxi:
This isn't a real pull request, but a chance for others to comment on the general shape of this large patch while I work towards making the tests pass.


wsfulton:
I would move the README.cffi information into a new Doc/Manual/CFFI.html as a starting point for user documentation. It may contain more developer information to begin with, but that can morph into user documentation as features get added/tested.

The Examples ought to be clones of what you see in other languages for consistency, so I would keep the class and simple examples, scrap the rest and if you feel up to it, port the other common tests: enum constants funcptr template pointer reference variables value. The removed tests can probably be replicated by converting tests from another language that you are familiar with. Common tests are really helpful not only for users but us developers who aren't familiar with all of the other languages, but can pick up some basics just by using good translations of the common examples. 

The changes to cffi,cxx unfortunately are quite inconsistent with anything else in SWIG (use of STL instead of DOH, code conventions/style), sigh. It honestly is much easier to crib bits of code from other target languages as maintaining a target language with different coding idioms is not so easy. But for now, just getting the code running with the latest code base will suffice.

wsfulton:
Closing all open CFFI issues as CFFI has been disabled, please see #1447.

PR #883, opened 2017-01-26
fix Lib/std/std unordered containers

tamuratak:
This patch fixes Lib/std/std unordered containers.

1. unordered containers do not have reverse_iterator. so we have introduced SWIG_NO_REVERSE_ITERATOR_METHODS flag.

2. fix the arguments of templates.

3. remove declarations of constructors which do not exist.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -545,6 +545,7 @@ CPP11_TEST_CASES = \
 	cpp11_explicit_conversion_operators \
 	cpp11_final_override \
 	cpp11_function_objects \
+	cpp11_hash_tables \
 	cpp11_inheriting_constructors \
 	cpp11_initializer_list \
 	cpp11_initializer_list_extend \
@@ -572,7 +573,6 @@ CPP11_TEST_CASES = \
 
 # Broken C++11 test cases.
 CPP11_TEST_BROKEN = \
-#	cpp11_hash_tables \           # not fully implemented yet
 #	cpp11_variadic_templates \    # Broken for some languages (such as Java)
 #	cpp11_reference_wrapper \     # No typemaps
 
--- a/Examples/test-suite/cpp11_hash_tables.i
+++ b/Examples/test-suite/cpp11_hash_tables.i
@@ -4,19 +4,27 @@
 
 %inline %{
 #include <set>
-//#include <map>
+#include <map>
 #include <unordered_set>
-//#include <unordered_map>
+#include <unordered_map>
 %}
 
 %include "std_set.i"
-//%include "std_map.i"
+%include "std_multiset.i"
+%include "std_map.i"
+%include "std_multimap.i"
 %include "std_unordered_set.i"
-//%include "std_unordered_map.i"
+%include "std_unordered_multiset.i"
+%include "std_unordered_map.i"
+%include "std_unordered_multimap.i"
 %template (SetInt) std::set<int>;
-//%template (MapIntInt) std::map<int, int>;
+%template (MultiSetInt) std::multiset<int>;
+%template (MapIntInt) std::map<int, int>;
+%template (MultiMapIntInt) std::multimap<int, int>;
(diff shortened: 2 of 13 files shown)


wsfulton:
Firstly we need some tests to pass before accepting the patch. There is a half written set of tests for this that were disabled, please apply the following patch:
```diff
diff --git a/Examples/test-suite/common.mk b/Examples/test-suite/common.mk
index 8d97b4b..1d35e0f 100644
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -545,6 +545,7 @@ CPP11_TEST_CASES = \
        cpp11_explicit_conversion_operators \
        cpp11_final_override \
        cpp11_function_objects \
+       cpp11_hash_tables \
        cpp11_inheriting_constructors \
        cpp11_initializer_list \
        cpp11_initializer_list_extend \
@@ -572,7 +573,6 @@ CPP11_TEST_CASES = \
 
 # Broken C++11 test cases.
 CPP11_TEST_BROKEN = \
-#      cpp11_hash_tables \           # not fully implemented yet
 #      cpp11_variadic_templates \    # Broken for some languages (such as Java)
 #      cpp11_reference_wrapper \     # No typemaps
```
This shows some failures which need addressing. We need some comprehensive runtime tests. if you can write some decent tests in your language of choice (I presume Ruby), I can convert them to Python.

Unfortunately, you can't use the `SWIG_NO_REVERSE_ITERATOR_METHODS` because reverse iterators are needed by other container types. You'll notice that your approach breaks usage of the other containers when an unordered container .i file is used in addition to another STL container .i file.

To facilitate the fixes you need to do, I've committed 65689af8486184560754f3e7fe8a4a702ad43802. You'll need to remove the use of `SWIG_NO_REVERSE_ITERATOR_METHODS`  and replace usage of `std_container_methods` with `std_container_methods_without_reverse_iterators` in the unordered*.i files. Please rebase when you update the pull request.

Finally, thanks, this will be a much appreciated addition to SWIG.

tamuratak:
Thanks for your response. I have added some tests. 

The following test has passed.
````
 make check-ruby-test-suite CPP_TEST_CASES=cpp11_hash_tables C_TEST_CASES="" MULTI_CPP_TEST_CASES=""
````
I will add more tests if needed.

Regards.

tamuratak:
I have written a patch to fix std unordered container codes for python although I don't know much about Python C API. Please check https://github.com/tamuratak/swig/compare/fix_std_unordred...tamuratak:fix_python_unordered_2

I made %swig_sequence_iterator in  Lib/python/pycontainer.swg to take an additional argument, a function making iterators. I also added forward iterator classes. I hope this would help your development.

Regards.

ojwb:
> Please check tamuratak@fix_std_unordred...tamuratak:fix_python_unordered_2

It's much better to submit proposed code changes as a PR, as then they are automatically built and tested.

Adding an extra argument to `%swig_sequence_iterator` would break any external uses of it - probably better to rename it and then add a simple `%swig_sequence_iterator` definition which just forwards to the new version, passing `swig::make_output_iterator` for the new argument.

tamuratak:
> Adding an extra argument to %swig_sequence_iterator would break any external uses of it

Thanks. I missed that point. I fixed the branch. 

Because the branch depends on this PR #883, I want to wait until  this PR is merged. After that, I will submit a new PR of the branch.

wsfulton:
Thanks, I've merged this in now. I look forward to the Python patch.

I've put the cpp11_hash_tables test into the Ruby makefile until support covers the other languages that we test C++11 support (currently Java, C#, Ruby, Python)

tamuratak:
Thanks. I will submit a new PR.

wsfulton:
@tamuratak if you get around to doing this (it'd be much appreciated), note that I committed b56d7c0539e398ec74d4e354f36beaaa2c758514 which fixes a few problems for the test to run when using the -builtin option. Python 3 and -builtin still needs some attention.

PR #885, opened 2017-01-31
Update ocaml documentation

guillaumeguerin:
Replaced python references by ocaml ones.

### DIFF
--- a/Examples/ocaml/simple/index.html
+++ b/Examples/ocaml/simple/index.html
@@ -84,7 +84,7 @@ let g = _gcd '(x,y) as int
 </pre>
 </blockquote>
 
-<li>C global variables are accessed through a special variable called '_Foo'. For example:
+<li>C global variable Foo is wrapped as _Foo in ocaml. For example:
 <blockquote>
 <pre>
 let _ = Printf.printf "Foo = %f\n" (_Foo '() as float)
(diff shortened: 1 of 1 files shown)


Examples/ocaml/simple/index.html
```diff
@@ -57,37 +57,37 @@
...
+let g = _gcd '(x,y) as int
 </pre>
 </blockquote>
 
-<li>C global variables are accessed through a special variable called 'cvar'. For example:
+<li>C global variables are accessed through a special variable called '_Foo'. For example:
 <blockquote>
 <pre>
-a = example.cvar.Foo
+let _ = Printf.printf "Foo = %f\n" (_Foo '() as float)
```
ojwb: The "through a special variable called" part doesn't seem the right wording for ocaml - in python `Bar` maps to `example.cvar.Bar`.  I'm not familiar with the ocaml backend, but in ocaml `Bar` presumably isn't accessed through `_Foo` - probably this should say something like "C global variable `Foo` is wrapped as `_Foo` in ocaml"?

guillaumeguerin: Yes that seems more correct, I'll edit this too.

ojwb:
Thanks, squashed and merged.

PR #887, opened 2017-02-05
Replace ICollection with IEnumerable<T>

jcsharp:
Adding a constructor that accepts` IEnumerable<T>` avoids the boxing and unboxing overhead of the original constructor, when the type parameter is a value type. This also allows passing `IList<T>`, which does not implement `ICollection` (ironically the generated type implements `IList<T>`).

Kept the original constructor for backward compatibility, but replaced `ICollection` with `IEnumerable` for added flexibility.


wsfulton:
Nice improvement and maintains backwards compatibility. I've added some tests to show this and added extra tests which failed before this patch was applied - 281eec7d18b5129175a266b3fcdb6218cb0d9090.

PR #895, opened 2017-02-11
Fix python support for unordered containers

tamuratak:
Hi, this PR fixes python support for unordered containers.
`make  check-python-test-suite` has passed. But tests for unordered containers are not included, which must be added when merging.

Regards.

### DIFF
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -654,7 +654,7 @@ namespace swig
 }
 }
 
-%define %swig_sequence_iterator(Sequence...)
+%define %swig_sequence_iterator(Make_output_iterator_func,Sequence...)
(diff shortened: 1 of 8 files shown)


wsfulton:
Thanks, I've merged this in. I tried to add in the the 4 new standalone tests `cpp11_std_unordered_map, cpp11_std_unordered_multimap, cpp11_std_unordered_multiset, cpp11_std_unordered_set`, but they led to compile errors, so more work to be done there.

I did however port the Ruby `cpp11_hash_tables` runtime test  in 7ad8ac92c0d6b9034303dedd1c2d0319dbd549d8 to check the implementation. There may be a problem obtaining the values for a given key for both `std::multimap` and `std::unordered_multimap` wrappers using indexing `[]` notation. Python's defaultdict doesn't quite work the same as the wrappers, but does return the list of values given this kind of indexing.

PR #899, opened 2017-02-14
[Coverity] fix issue reported for SWIG_Python_ConvertFunctionPtr

srepmub:
Fix Coverity issue reported for SWIG_Python_ConvertFunctionPtr:

"Execution cannot reach this statement: *ptr = vptr;"

Because if 'ty' is null, then desc becomes null and we return with
SWIG_ERROR. So 'ty' cannot be null at 'if (ty)'.


ojwb:
Results in lots of:

```
example_wrap.c:2441:5: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
```

SWIG aims to generate C90-compliant code (unless features from newer C standards are used) so `swig_cast_info *tc` will need to be declared at the top of the function, or else a block left where the
removed `if` was.

srepmub:
apologies, updated the PR.

PR #900, opened 2017-02-14
fixes for smart-pointers, namespaces and package names

fschlimb:
- white space cleanup (following SWIG conventions)
- making smart-pointer classes derive from parent smart-pointers
- constructors now returning smart pointers (if class declared as such)
- fixing smart-pointer and NAMESPACE support
- allowing R package names with containing dosts ('.')
- allowing multiple get accessors in R

### DIFF
--- a/Source/Modules/r.cxx
+++ b/Source/Modules/r.cxx
@@ -1,5 +1,5 @@
 /* -----------------------------------------------------------------------------
- * This file is part of SWIG, which is licensed as a whole under version 3 
+ * This file is part of SWIG, which is licensed as a whole under version 3
  * (or any later version) of the GNU General Public License. Some additional
  * terms also apply to certain portions of SWIG. The full details of the SWIG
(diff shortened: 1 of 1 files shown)


wsfulton:
@fschlimb, you've managed to break something fundamental. See the Travis CI build where there are lots of error message like:
```
cat: ../../../../../Examples/test-suite/r/r_double_delete_runme.Rout: No such file or directory
```
@joequant Could you review this please? BTW, you can see the changes in the browser without the whitespace changes by adding ?w=1 after the URL, such as https://github.com/swig/swig/pull/900/files?w=1.

fschlimb:
Yes, I know what it is. I will try to find the time to get this fixed.

joequant:
Can you refactor the patch in order to separate this into three or four
different commits?

There seem to be three or four logically separate fixes, and I'd appreciate
it if each fix was in a separate commit.

Thanks

PR #901, opened 2017-02-14
[Coverity] fix issue reported for wrapper argument checking

srepmub:
Fix Coverity issue reported for wrapper argument checking:

"Null-checking args suggests that it may be null, but it has already
been dereferenced on all paths leading to the check."

So 'args' is null checked, but after dereferencing it with
PyTuple_Check(args).

If we assume that args may be null (else why check for it), then we
need to avoid dereferencing null.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -2541,9 +2541,14 @@ public:
 
     if (!fastunpack) {
       Wrapper_add_local(f, "ii", "Py_ssize_t ii");
-      if (maxargs - (add_self ? 1 : 0) > 0)
-	Append(f->code, "if (!PyTuple_Check(args)) SWIG_fail;\n");
-      Append(f->code, "argc = args ? PyObject_Length(args) : 0;\n");
+
+      if (maxargs - (add_self ? 1 : 0) > 0) {
+        Append(f->code, "if (!PyTuple_Check(args)) SWIG_fail;\n");
+        Append(f->code, "argc = PyObject_Length(args);\n");
+      } else {
+        Append(f->code, "argc = args ? PyObject_Length(args) : 0;\n");
+      }
+
       if (add_self)
 	Append(f->code, "argv[0] = self;\n");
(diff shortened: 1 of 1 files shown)


ojwb:
It's not clear to me if `args` can actually be `NULL` here, or if we ought to remove the `NULL` check from the next line:

```
Append(f->code, "argc = args ? PyObject_Length(args) : 0;\n");
```

It may depend on the calling convention in use - Python has a bewildering variety of them (https://docs.python.org/2/c-api/structures.html#c.PyMethodDef) though it's not entirely clear to me which take this code path.

I guess the proposed patch is at least a safe change, even if it may result in unnecessary checks.

srepmub:
glancing at the surrounding code:

      if (maxargs - (add_self ? 1 : 0) > 0)
>-------Append(f->code, "if (args && !PyTuple_Check(args)) SWIG_fail;\n");
      Append(f->code, "argc = args ? PyObject_Length(args) : 0;\n");

it appears that the PyTuple_Check is not added for METH_O, which seems to be the only case where args can be NULL (looking at the possibilities in void add_method() in python.cxx, SWIG can only emit METH_NOARGS, METH_O, METH_VARARGS and METH_KEYWORDS). so then the check would indeed be unnecessary..

srepmub:
so then the obvious change would be:

'''if (maxargs - (add_self ? 1 : 0) > 0)
    Append(f->code, "if (!PyTuple_Check(args)) SWIG_fail;\n");
    Append(f->code, "argc = PyObject_Length(args);\n");
else
    Append(f->code, "argc = args ? PyObject_Length(args) : 0;\n");
'''

if you agree, I'm happy to create a pull-request of course..

srepmub:
hmm so ruby testing fails..

srepmub:
it seems far fetched that this change could break the ruby tests, right.. could the tests be run again perhaps, or something?

wsfulton:
Was just glitch on Travis. I've rerun the failing test and now it is all green. I'll merge shortly.

PR #906, opened 2017-02-18
Fix ruby support for std::wstring

tamuratak:
Hi, this PR fixes ruby support for std::wstring, which is completely broken now. 

I am not sure how to treat wchar_t as portable. I understand portability has high priority for SWIG. This PR may not satisfy you at that point.

At least, this PR clarifies how to convert the encoding of string in Ruby C API.  `make check-ruby-test-suite` has passed.

Regards.

### DIFF
--- a/Examples/test-suite/ruby/Makefile.in
+++ b/Examples/test-suite/ruby/Makefile.in
@@ -19,6 +19,7 @@ CPP_TEST_CASES = \
 	li_std_queue \
 	li_std_set \
 	li_std_stack \
+	li_std_wstring \
 	primitive_types \
 	ruby_keywords \
 	ruby_minherit_shared_ptr \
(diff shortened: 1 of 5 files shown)


ojwb:
Travis had a moment and failed to fetch packages for two of the ruby CI jobs (and some non-ruby ones). I've restarted the Ruby ones as they actually matter here.

I think in practical terms `sizeof(wchar_t)` is going to be 2 or 4, but checking `WCHAR_MAX` with the preprocessor would mean any system with wider `wchar_t` would fail at compile time rather than giving a runtime error, which seems better.

ojwb:
OK, Ruby CI jobs all passed now.

tamuratak:
Thanks for your comment. I have switched to using WCHAR_MAX.

tamuratak:
It turned out that I have misunderstood some points. So I draw back this PR at this time. I will submit a new PR later.

Sorry.

PR #908, opened 2017-02-19
[Issue 905] Add recognition for `throw (.+) final|override` in parser.y

assambar:
PROBLEM:
 There is a small ommission in parser.y, which will lead
 to syntax errors in cases when non-empty throw declaration is
 followed by `override`, `final` or both. E.g. in cases like:

       void finalOverriden() throw(std::exception) final override;

SOLUTION:
 - Add a `THROW LPAREN parms RPAREN virt_specifier_seq` to
   exception_specification in Source/CParse/parser.y
 - Add several methods in test-suite/cpp11_final_override.i
   to verify the fix works.

### DIFF
--- a/Examples/test-suite/cpp11_final_override.i
+++ b/Examples/test-suite/cpp11_final_override.i
@@ -16,6 +16,8 @@ struct Base {
   virtual void finaloverride2() {}
   virtual void finaloverride3() {}
   virtual void finaloverride4() const {}
+  virtual void finaloverride5() {}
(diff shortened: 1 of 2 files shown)


vadz:
I guess there is no real harm in supporting this, but this is rather strange as `final` and `override` are C++11 additions, while `throw()` with parameters is deprecated in C++11, so I really don't see why would anybody use them together.

assambar:
@vadz I've got just one word - legacy

PR #911, opened 2017-02-20
Fix ruby support for std::wstring

tamuratak:
Hi, I have rewritten PR #906.  I have added comprehensive tests also.

### DIFF
--- a/Lib/ruby/rubywstrings.swg
+++ b/Lib/ruby/rubywstrings.swg
@@ -8,11 +8,28 @@
 SWIGINTERN int
 SWIG_AsWCharPtrAndSize(VALUE obj, wchar_t **cptr, size_t *psize, int *alloc)
 {
-  VALUE tmp = rb_str_conv_enc(obj, rb_enc_get(obj),
-                                   rb_to_encoding(rb_str_new_cstr( SWIG_RUBY_WSTRING_ENCODING )) );
-  *alloc = SWIG_NEWOBJ;
 
-  return SWIG_AsCharPtrAndSize(tmp, (char**)cptr, psize, alloc);
+  if (TYPE(obj) == T_STRING) {
+    VALUE tmp = rb_str_conv_enc(obj, rb_enc_get(obj),
+                                rb_to_encoding(rb_str_new_cstr( SWIG_RUBY_WSTRING_ENCODING )) );
+    wchar_t* cstr = (wchar_t*) StringValuePtr(tmp);
+    size_t   size = RSTRING_LEN(tmp) / sizeof(wchar_t) + 1;
+
+    if( RSTRING_LEN(tmp) % sizeof(wchar_t) != 0 ) {
+        rb_raise(rb_eRuntimeError,
+                 "The length of the byte sequence of converted string is not a multiplier of sizeof(wchar_t). Invalid byte sequence is given. Or invalid SWIG_RUBY_WSTRING_ENCODING is given when compiling this binding.");
+    }
+    if (cptr && alloc)  {
+      *alloc = SWIG_NEWOBJ;
+      *cptr = %new_array(size, wchar_t);
+      memmove(*cptr, cstr, RSTRING_LEN(tmp));
+    }
+    if(psize) *psize = size;
+
+    return SWIG_OK;
+  } else {
+    return SWIG_TypeError;
+  }
 }
 }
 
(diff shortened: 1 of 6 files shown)


Lib/ruby/rubywstrings.swg
```diff
@@ -1,71 +1,55 @@
...
-//       swig_type_info* pwchar_descriptor = SWIG_pwchar_descriptor();
-//       return pwchar_descriptor ? 
-// 	SWIG_NewPointerObj(%const_cast(carray,wchar_t *), pwchar_descriptor, 0) : Qnil;
-//     } else {
-//       return SWIG_Unicode_FromWideChar(carray, %numeric_cast(size,int));
-//     }
-//   } else {
-//     return Qnil;
-//   }
+  if (carray && size <= LONG_MAX) {
```
ojwb: I think you want `size <= LONG_MAX / sizeof(wchar_t)` here to make sure the size calculation below won't overflow.

Lib/ruby/rubywstrings.swg
```diff
@@ -1,71 +1,55 @@
...
-//   }
-//   return SWIG_TypeError;
+
+  if (TYPE(obj) == T_STRING) {
+    VALUE tmp = rb_str_conv_enc(obj, rb_enc_get(obj),
+                                rb_to_encoding(rb_str_new_cstr( SWIG_RUBY_WSTRING_ENCODING )) );
+    wchar_t* cstr = (wchar_t*) StringValuePtr(tmp);
+    size_t   size = RSTRING_LEN(tmp) / sizeof(wchar_t) + 1;
+
+    if( RSTRING_LEN(tmp) % sizeof(wchar_t) != 0 ) {
```
ojwb: Your patch has a mixture of `if (` and `if(`.  Sadly the SWIG code base doesn't consistently use either, but it's particularly unhelpful to mix them within a single function like this.  It seems `if (` dominates, so I'd go for that consistently.

tamuratak:
Thanks for your comments. I have edited codes following it.

Regards.

tamuratak:
Due to the failure of tests, I have switched not using static variables that are still mysterious for me. Hmm.

Lib/ruby/rubywstrings.swg
```diff
@@ -8,23 +8,23 @@
 SWIGINTERN int
 SWIG_AsWCharPtrAndSize(VALUE obj, wchar_t **cptr, size_t *psize, int *alloc)
 {
+  static rb_encoding* wstr_enc = rb_to_encoding(rb_str_new_cstr( SWIG_RUBY_WSTRING_ENCODING ));
```
ojwb: That may work in C++ (though could have issues if multiple threads can execute this concurrently), but in C a `static` variable needs a constant initialiser.  The usual idiom if to make the static initialiser `NULL` and then to check at runtime, something like: `if (wstr_enc == NULL) wstr_enc = rb_to_encoding(rb_str_new_cstr( SWIG_RUBY_WSTRING_ENCODING ));`

But again that's potentially racy when multi-threading.

So perhaps it's best to find a way to get the encoding more efficiently.

For UTF-8, it looks like `rb_utf8_encoding()` might be a better option (it returns `rb_encoding *`).

For UTF-16/UTF-32 perhaps `rb_enc_find_index()` (takes `const char *`) then `rb_enc_from_index()` to get the `rb_encoding *`?

But I'm not really familiar with these APIs, so maybe these are still too slow.

tamuratak:
Thanks. Static variables are still mysterious for me. I have moved the initializing into %init{} block, in which codes would not be executed concurrently by multiple threads.

ojwb:
Probably better to only set those encodings up if this wstring stuff is used, so we don't impose the extra start-up overhead for everyone - e.g. put that `%init` code into a fragment and have the new wstring typemaps pull in that fragment:

http://www.swig.org/Doc3.0/SWIGDocumentation.html#Typemaps_fragments

So something like:

```
%fragment("SWIG_rb_enc_find", "init") {
  swig_ruby_wstring_encoding = rb_enc_find( SWIG_RUBY_WSTRING_ENCODING );
  swig_ruby_internal_encoding = rb_enc_find( SWIG_RUBY_INTERNAL_ENCODING );
}
```

And then change:

```
%fragment("SWIG_AsWCharPtrAndSize","header",fragment="<wchar.h>",fragment="SWIG_pwchar_descriptor",fragment="SWIG_AsCharPtrAndSize") {
```

to specify this new fragment:

```
%fragment("SWIG_AsWCharPtrAndSize","header",fragment="<wchar.h>",fragment="SWIG_pwchar_descriptor",fragment="SWIG_AsCharPtrAndSize",fragment="SWIG_rb_enc_find") {
```

And similarly for any other typemaps which requires these encodings to be set up.

tamuratak:
Hi, I have patched that way. Thanks.

wsfulton:
@ojwb, you okay with this now? Looks okay to me.

wsfulton:
@ojwb, I'll merge this in unless you have any objections?

ojwb:
@wsfulton If you're happy with it, go for it.

PR #916, opened 2017-02-25
[ruby] add simple assertion functions for tests.

tamuratak:
Hi, using the current assertion functions in Examples/test-suite/ruby/swig_assert.rb is honestly painful. They require arguments to be given as strings. It also makes test codes less readable. This PR adds simple assertion functions that fail when they should.

Regards.


wsfulton:
I have merged this in but have changed the names and added in a ```swig_``` prefix . I'll make the same change in your other patch requests as I go through them.

PR #928, opened 2017-02-28
[Coverity] fix issue reported for setslice (pycontainer.swg)

srepmub:
Fix Coverity issue reported for setslice (pycontainer.swg):

"CID 11151 (#3-1 of 3): Using invalid iterator (INVALIDATE_ITERATOR)18.
increment_iterator: Incrementing iterator it though it is already past
the end of its container."

Coverity does not understand 'replace_count', so warns that we may go
past self->end() (or self->rend() I guess). Because after each slice
step, it depends on the correctness of 'replace_count' if we iterate
further (which we probably won't, but still).

### DIFF
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -351,7 +351,7 @@ namespace swig {
         typename Sequence::const_iterator isit = is.begin();
         typename Sequence::iterator it = self->begin();
         std::advance(it,ii);
-        for (size_t rc=0; rc<replacecount; ++rc) {
+        for (size_t rc=0; rc<replacecount && it != self->end(); ++rc) {
           *it++ = *isit++;
           for (Py_ssize_t c=0; c<(step-1) && it != self->end(); ++c)
             it++;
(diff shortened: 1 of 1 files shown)


ojwb:
CI failed, but won't show me the logs so I can't tell if it a genuine issue with the patch.

jakecobb:
@ojwb The appveyor results do show a failing test:

```
checking python testcase li_std_map_member (with run test)
/bin/sh: line 1:  3752 Segmentation fault      env LD_LIBRARY_PATH=.:$LD_LIBRARY_PATH PYTHONPATH=.:.:$PYTHONPATH python li_std_containers_int_runme.py
make[1]: *** [Makefile:111: li_std_containers_int.cpptest] Error 139
```

ojwb:
OK, well that really needs fixing before we can merge.

srepmub:
argh, forgot to run the tests before pushing, sorry. will fix shortly.

srepmub:
to compensate, the code should be faster now as well. could even be faster by moving the 0-check outside, but I don't that's worth the extra code.

srepmub:
I'm having trouble seeing what is wrong with the current patch.. the tests work fine for me, it should be standard C++ right, and I don't see how the change might segfault other tests..

srepmub:
reverting to a more obvious "fix". 

it would still be nice to optimize this code, to not scan 1 million items when only changing 3.. maybe I will have another go later.

wsfulton:
I think this is a false positive by Coverity, but if you are using Coverity it is nice to have a clean sheet so I'll merge this as the tests I put in around setslice are very comprehensive and the change looks harmless to me.

PR #929, opened 2017-02-28
%pythonappend docs and memory management example

jakecobb:
Document 'val' variable for the %pythonappend directive.
Add memory management example equivalent to Java and C#
docs using %pythonappend.

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -5319,10 +5319,18 @@ wheel size: 135019664
 What has happened here is the garbage collector has collected the <tt>Bike</tt> instance as it doesn't think it is needed any more.
 The proxy instance, <tt>wheel</tt>, contains a reference to memory that was deleted when the <tt>Bike</tt> instance was collected.
 In order to prevent the garbage collector from collecting the <tt>Bike</tt> instance, a reference to the <tt>Bike</tt> must
-be added to the <tt>wheel</tt> instance. You can do this by adding the reference when the <tt>getWheel()</tt> method
-is called using the typemap-like <tt>%pythonappend</tt> directive (see <a href="#Python_nn42">36.6.2 Adding additional Python code</a>):
+be added to the <tt>wheel</tt> instance. 
 </p>
-
+    
+<p>
+You can do this by adding the reference when the <tt>getWheel()</tt> method
+is called using one of two approaches:
+</p>
+    
+<p>
+The easier, but less optimized, way is to use the typemap-like <tt>%pythonappend</tt> directive 
+(see <a href="#Python_nn42">36.6.2 Adding additional Python code</a>):
+</p>
 
 <div class="code">
 <pre>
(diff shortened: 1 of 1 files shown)


jakecobb:
Not sure what's wrong with the Travis-CI, but this is a documentation-only change that could not affect code tests.

ojwb:
Amazon S3 broke: https://www.traviscistatus.com/incidents/hmwq9yy5dh9d

jakecobb:
Any reason this hasn't been accepted?  Changes needed?

ojwb:
Just lack of time to review and a backlog of patches I think.  Sorry about that.

It looks reasonable to me.

I'm not entirely sure about documenting `val` - it looks to me like it's being created for this purpose, but I can't see any existing documentation mentioning `val` (not even a mention in `CHANGES`) and generally such things are abstracted into a `$foo` placeholder (for example, `$self` in `%extend`) which gives us more flexibility with implementation.

@wsfulton Thoughts?

wsfulton:
Sorry @jakecobb, I didn't get a chance to look beforehand. I think this will be a useful addition to the docs. However, I'd rather document a solution that does the same thing but in C code and thus works for the `-builtin` option too. This is the code needed:
```swig
%fragment("extra_reference", "header") {

static PyObject *extra_reference() {
  static PyObject *extra_reference_string = NULL;
  if (!extra_reference_string)
    extra_reference_string = SWIG_Python_str_FromChar("_extra_reference");
  return extra_reference_string;
}

}

%extend Wheel {
%typemap(ret, fragment="extra_reference") Wheel& getWheel %{
  // A reference to the parent class is added to ensure the underlying C++
  // object is not deleted while the item is in use
  PyObject_SetAttr($result, extra_reference(), $self);
%}
/* FYI: Alternative approach, but is possibly harder to understand, so suggest above
%typemap(out, fragment="extra_reference") Wheel& getWheel %{
  $typemap(out, Wheel &)
  // A reference to the parent class is added to ensure the underlying C++
  // object is not deleted while the item is in use
  PyObject_SetAttr($result, extra_reference(), $self);
%}
*/
}

%inline %{
#include <iostream>
struct Wheel {
  int size;
  Wheel(int sz) : size(sz) {}
  ~Wheel() { std::cout << "~Wheel" << std::endl; }
};

class Bike {
  Wheel wheel;
public:
  Bike(int val) : wheel(val) {}
  Wheel& getWheel() { return wheel; }
};
%}
```

jakecobb:
@ojwb  I'm not competent enough in SWIG internals to make it use some `$return` style substitution, but considering `val` is there I think it's reasonable to document it since you otherwise can't do anything with the return value in `%pythonappend` (without playing games with `locals()` or similar hacks).

@wsfulton How about documenting both methods?  When using proxy classes the `%pythonappend` method is easiest to understand IMO (don't need to understand CPython API calls or the SWIG CPython API helpers).  For conciseness and flexibility I would also want to drop the fragment portion so understanding the example doesn't depend on knowing that feature, do we need to decrement the PyObject reference after use in that case?

%extend Bike {
%typemap(ret, fragment="extra_reference") Wheel& getWheel %{
  // A reference to the parent class is added to ensure the underlying C++
  // object is not deleted while the item is in use
  PyObject* bike_reference = SWIG_Python_str_FromChar("_bike");
  PyObject_SetAttr($result, bike_reference, $self);
  Py_DECREF(bike_reference);
%}
}

Let me know your thoughts on this and then I can update the pull request.

wsfulton:
Sorry, I didn't reply earlier, but yes, add the code I suggested after what you have done so both approaches are documented. The fragment might be harder to understand but I'd like it put in as it provides an important optimization.

jakecobb:
Moved this to PR 1227 against current master, closing this one.  Comments are addressed in that PR.

Issue #773, opened 2016-08-28
C++/Python with shared_ptr lists: Assertion `own' failed.


nschloe:
In a SWIGged C++/Python application, I'd like to create a function that accepts a `std::vector`/Python list of abstract base class objects, e.g.,

``` c++
#ifndef MYTEST_HPP
#define MYTEST_HPP

#include <iostream>
#include <vector>
#include <memory>

class MyBaseClass {
  public:
  virtual void eval() const = 0;
};

class MyTestClass: public MyBaseClass {
  public:
  virtual void eval() const {
    std::cout << "MyTestClass::eval()" << std::endl;
  }
};

void
mytest(const std::vector<std::shared_ptr<MyBaseClass>> & a) {
  std::cout << "length: " << a.size() << std::endl;
}

#endif // MYTEST_HPP
```

Note that we're using `std::shared_ptr`s here. Using this with

``` swig
%module mytest

%{
#define SWIG_FILE_WITH_INIT
#include "mytest.hpp"
%}

%include <std_shared_ptr.i>
%shared_ptr(MyBaseClass);
%shared_ptr(MyTestClass);

%include <std_vector.i>
namespace std {
%template(List) vector<std::shared_ptr<MyBaseClass>>;
}

%include "mytest.hpp"
```

does compile, but fails at runtime. From

``` python
import mytest
a = mytest.MyTestClass()
mytest.mytest([a])
```

I'm getting

```
python: /tmp/l/build/mytestPYTHON_wrap.cxx:2385: int SWIG_Python_ConvertPtrAndOwn(PyObject*, void**, swig_type_info*, int, int*): Assertion `own' failed.
Aborted (core dumped)
```

[This works with SWIG 3.0.2](https://stackoverflow.com/questions/39163006/vector-list-of-abstract-base-class-objects-in-swig-c-python), but fails at least since 3.0.8.

Note that with

``` python
import mytest
a = mytest.MyTestClass()
l = mytest.List()
l.append(a)
mytest.mytest(l)
```

works.

---

_Note:_ The corresponding Gist is [here](https://gist.github.com/nschloe/c6d1dcb0837031183da0da5423ea23f6).

nschloe:
The same happens with [the latest SWIG from master](https://launchpad.net/~nschloe/+archive/ubuntu/swig-nightly/).

wsfulton:
[...]

Issue #731, opened 2016-06-24
shared_ptr Derived -> Base conversion triggers memory ownership assert


q-p:
In the attached repo, I attempt to create a `vector<shared_ptr<Base> >` and I pass in a `shared_ptr<Derived>`. This leads to an assertion failure in `Assertion failed: (own), function SWIG_Python_ConvertPtrAndOwn, file Node_wrap.cxx, line 2385.`

Is there any way to make this work correctly?

`Node.hpp`

``` c++
#ifndef NODE_HPP_3712EF25
#define NODE_HPP_3712EF25

#include <memory>
#include <iostream>
#include <vector>

struct Base
{
  virtual void Print() const { std::cout << "Base" << std::endl; }
  static std::shared_ptr<Base> Make() { return std::make_shared<Base>(); }
};

struct Derived : Base
{
  void Print() const override { std::cout << "Derived" << std::endl; }
  static std::shared_ptr<Derived> Make() { return std::make_shared<Derived>(); }
};

#endif /* end of include guard: NODE_HPP_3712EF25 */
```

`Node.i`

``` swig
%module Node

// includes to compile the wrapper code
%{
#include "Node.hpp"
%}

// notify shared_ptrs
%include <std_shared_ptr.i>
%shared_ptr(Base)
%shared_ptr(Derived)

// load typemaps
%include <std_vector.i>

// and create wrappers
%include "Node.hpp"

%template(BaseVector) std::vector<std::shared_ptr<Base> >;
```

`test.py`

``` python
#!/usr/bin/env python

from Node import *

l = [Base.Make(), Derived.Make()]
vec = BaseVector(l)
```

`Makefile`

``` Makefile
# it is important that NDEBUG is *not* defined, as otherwise the assert is disabled
_%.so : %_wrap.cxx %.py
    clang++ -std=c++11 -UNDEBUG -g -fno-common -fno-strict-aliasing -fwrapv -dynamic -I/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/include/python2.7 -L/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config -lpython2.7 -ldl -shared $< -o $@

%_wrap.cxx %.py : %.i
    swig -c++ -python -fvirtual -shadow -modern $<

test : Node.py _Node.so test.py
    python test.py

clean :
    rm -rf *.pyc Node_wrap.cxx Node.py _*.so*
```

[...]

PR #930, opened 2017-03-01
[ruby] upcasting shared_ptr within std containers

tamuratak:
An issue about std_shared_ptr.i mentioned in #731 and #773 also exists for Ruby. This PR fixes it. 

When upcasting `Derived*` to `Base*`, it is sufficient that `Base*` is allocated on stack, which is done by allocating `Type *p` on stack in the current `traits_asval::asval` for example. 

But when upcasting from `std::shared_ptr<Derived>` to `std::shared_ptr<Base>`,  `std::shared_ptr<Base>` needs to be allocated somewhere. In other words, when `Type = std::shared_ptr<Base>`,  we need `Type p`,  instead of `Type *p`, on stack. In the current  `Lib/ruby/rubystdcommon.swg`, `traits_asptr::asptr` does not treat the latter case properly.

In this PR, by using template specialization, a new `traits_asptr::asptr` that treats the above cases properly is defined.  The same solution might be applied to python.

Notice that in new `traits_asptr::asptr`, we call `**val = *p`.  In other words,  `**val` is assumed to be already allocated when `traits_asptr::asptr` called.  New `traits_asval::asval`s properly allocate it. I am not sure this change is acceptable.  Anyway I think this PR is a good place to start fixing this issue.

Regards.

### DIFF
--- a/Examples/test-suite/cpp11_shared_ptr_upcast.i
+++ b/Examples/test-suite/cpp11_shared_ptr_upcast.i
@@ -9,9 +9,9 @@
 #include <vector>
 %}
 
-%include std_vector.i
-%include std_map.i
-%include std_shared_ptr.i
+%include <std_vector.i>
+%include <std_map.i>
+%include <std_shared_ptr.i>
 
 %{
 
--- a/Examples/test-suite/cpp11_shared_ptr_upcast.i
+++ b/Examples/test-suite/cpp11_shared_ptr_upcast.i
@@ -1,0 +1,97 @@
+%module cpp11_shared_ptr_upcast
+
+%{
+#include <set>
+#include <map>
+#include <memory>
+#include <unordered_set>
+#include <unordered_map>
+#include <vector>
+%}
+
+%include std_vector.i
+%include std_map.i
+%include std_shared_ptr.i
+
+%{
+
+class Base {
+public:
+  Base() : m(-1) {}
+  Base(int i) : m(i) {}
+  int get_m() { return m; }
+  int m;
+};
+
+class Derived : public Base {
+public:
+  Derived() : n(-2) {}
+  Derived(int i) : n(i) {}
+  int get_n() { return n; }
+  int n;
+};
+
+ typedef std::shared_ptr<Base>    BasePtr;
+ typedef std::shared_ptr<Derived> DerivedPtr;
+
+ int derived_num(DerivedPtr v) {
+   return (*v).get_n();
+ }
+
+ int derived_num(std::vector<DerivedPtr> v) {
+   return (*v[0]).get_n();
+ }
+
+ int derived_num(std::map<int, DerivedPtr> v) {
+   return (*v[0]).get_n();
+ }
+
+ int base_num(BasePtr v) {
+   return (*v).get_m();
+ }
+
+ int base_num(std::vector<BasePtr > v) {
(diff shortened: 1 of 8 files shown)


wsfulton:
@tamuratak ... kudos for getting this to work and figuring out what to do! 

The only things I'd like to change are:
1. The from specialization would also be needed (when returning a shared_ptr from a function).
2. The way the template specializations are included. While your SWIG_RUBYSTDCOMMON macro might work to ensure the template specializations are after the primary template, I think we can do better than this. I've had real trouble getting the order right in other developments when using the std templates. I think we should solve this by using a forward template declaration in a fragment. Let's add:
```swig
%fragment("StdTraitsForwardDeclaration","header")
{
namespace swig {
  template <class Type> struct traits_asptr;
  template <class Type> struct traits_asval;
}
}
```
and modify ```StdTraits``` to depend on this too, so it becomes:
```swig
%fragment("StdTraits","header",fragment="StdTraitsForwardDeclaration",fragment="StdTraitsCommon")
...
```
Then the std::shared_ptr specializations you created can be put into a fragment:
```swig
%fragment("StdSharedPtrTraits","header",fragment="StdTraitsForwardDeclaration")
{
namespace swig {
  template <class Type> struct traits_asptr<std::shared_ptr<Type> > { ... }
  template <class Type> struct traits_asval<std::shared_ptr<Type> > { ... }
}
}
%fragment("StdSharedPtrTraits") // force the fragment or add a dependency to it into the shared_ptr typemaps.
```
The order of ```%include <std_vector.i>``` or ```%include <std_shared_ptr.i>``` won't then matter as the primary forward template declarations will always appear before any template specializations. It also won't matter that the primary template **definition** may or may not appear after the specializations as the primary template **declaration** will always be first. If just ```std_shared_ptr.i``` is included without any stl interface files, it won't matter that the specialization code is included, it'll compile fine and be ignored.

Examples/test-suite/cpp11_shared_ptr_upcast.i
```diff
@@ -0,0 +1,97 @@
...
+#include <map>
+#include <memory>
+#include <unordered_set>
+#include <unordered_map>
+#include <vector>
+%}
+
+%include std_vector.i
+%include std_map.i
+%include std_shared_ptr.i
```
wsfulton: Please use <>, eg ```%include <std_shared_ptr.i>```

wsfulton:
I see you have covered the case of returning a ```vector<shared_ptr>``` in #937 and added a traits_from in there. Changes in either pull request will probably break the other, so I suggest combining them and merging #937 into this one.

tamuratak:
Hi, 
I have merged #937 to this PR.

I also have added `rubystdcommon_forward.swg` included by `rubystdcommon.swg` and `std_shared_ptr.i`.  The fragment "StdTraitsForwardDeclaration" is needed if just `std_shared_ptr.i` is included.

Regards.

wsfulton:
Thanks for this. I've made some additional changes in final merge a692e5de93be37a9295b6c6609e58f0698a2d41d. Some of the template specializations looked to be incomplete, so I added those in and added testcases too. These were mostly around containers of pointers of ```shared_ptr```. Perhaps not a common use case, but I feel it completed the specializations. I also added in support for empty ```shared_ptr``` / the equivalent to a null pointer.

PR #932, opened 2017-03-01
Updated Python documentation with info on -threads option(s)

### LINKED ISSUE #927: Python documentation does not mention that "-threads" option
Nowhere in the SWIG Python documentation does it mention that "-threads" option for enabling multithreaded support by automatically disabling the GIL when wrapped C/C++ code is entered.  This option is absolutely critical for many common use cases, including mine.  Hence, it should be mentioned in the documentation.

If you run the following at the command line:
```bash
swig -python -help
```

You will see the following option present among the Python Options
```bash
Python Options (available with -python)
    -threads        - Add thread support for all the interface
```

I found that this "-threads" option was essential for expected multithreaded behavior when using SWIG-wrapped C++ code in a multithreaded Python application.

Oh, and thanks for making such a great tool which makes it incredibly easy to wrap C/C++ code!

tleonhardt:
My changes are all at the very top in the TOC and at the very bottom in a new section on multithreaded Python SWIG applications.

I used an editor which automatically strips trailing whitespace at the ends of lines, so there is are a bunch trivial whitespaces changes that weren't intentional, but shouldn't hurt.

I've never contributed to SWIG before, so if I didn't follow some sort of style guide, please be kind.  I didn't see a CONTRIBUTING.md file to guide me on what to do or not do, so I winged it.

This closes #927

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -133,6 +133,10 @@
 <li><a href="#Python_nn77">Byte string output conversion</a>
 <li><a href="#Python_2_unicode">Python 2 Unicode</a>
 </ul>
+<li><a href="#Python_multithreaded">Support for Multithreaded Applications</a>
+<ul>
+<li><a href="#Python_thread_UI">UI for Enabling Multithreading Support</a>
+</ul>
 </ul>
 </div>
 <!-- INDEX -->
@@ -162,7 +166,7 @@ Basics</a>" chapter.
 
 <p>
 To build Python extension modules, SWIG uses a layered approach in which
-parts of the extension module are defined in C and other parts are 
+parts of the extension module are defined in C and other parts are
 defined in Python.  The C layer contains low-level wrappers whereas Python code
 is used to define high-level features.
 </p>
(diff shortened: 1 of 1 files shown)


wsfulton:
Thanks @tleonhardt , this is a useful addition to the docs. I'll merge shortly and have made a couple of style changes to fit in with the rest of the documentation. Just one question, I see the profiletest.i you reference in the test-suite. Do you have the runtime tests to obtain the numbers as it is a bit vague what the numbers mean without a runtime test.

tleonhardt:
@wsfulton The profiling data I added was taken from the entry in the CHANGES file by mmatus on 12/06/2005.  @ojwb helpfully pointed that out as the only extant documentation related to the **-threads** option and I borrowed from it liberally, mostly translating it from markdown to HTML.

ojwb:
I just noticed these profiling benchmarks in the manual and wondered what Python version they were for and whether they were still relevant.  Trying to run `make -C Examples/test-suite/python profiletest.cpptest` reveals it's not currently compatible with Python 3, which made me suspect they were from Python 2 and made me doubt they're relevant still.  Tracking down this PR, I see they're close to 18 years old!

I'll see if I can work out how Marcelo generated the different numbers and update them.  Fixing the minor Python 3 incompatibility, the test takes about 1 second (but with a large variance) and I don't see a built-in way to run it with other options.

ojwb:
Increasing the number of iterations 20-fold from 50000 to 1000000 I can time the cases without threads and with threads easily:

```
make -C Examples/test-suite/python profiletest.cpptest RUNTOOL=time
make -C Examples/test-suite/python profiletest.cpptest RUNTOOL=time SWIGOPT=-threads
```

The timing is still quite noisy (I'll try timing in the python code and see if that's better) but suggests that with Python 3.11.7 the overhead of `-threads` is more like 40% than the 60% the manual suggests.

The other two cases seem to need features to be specified in the interface file - I'll hook something up to allow that.

Issue #586, opened 2016-01-12
Ruby crash on member function returning std::vector<std::shared_ptr<const Foo>>


tjensen:
Using SWIG version 3.0.8 with Ruby, I'm observing crashes when calling member functions that return vectors of shared_ptrs, where the shared_ptrs point to **const** objects.

Here are the hpp and cpp files I used to isolate the issue:

``` cpp
// Blob.hpp
#include <memory>
#include <vector>

struct Foo
{
    typedef std::shared_ptr<Foo> Ptr;
    typedef std::shared_ptr<const Foo> ConstPtr;
};

struct Bar
{
    std::vector<Foo::Ptr> foos() const;
    std::vector<Foo::ConstPtr> const_foos() const;
};
```

``` cpp
// Blob.cpp
#include <memory>
#include <vector>

#include "Blob.hpp"

std::vector<Foo::Ptr> Bar::foos() const
{
    std::vector<Foo::Ptr> result;
    for ( auto i = 0; i < 100; ++i )
    {
        result.push_back( std::make_shared<Foo>() );
    }
    return result;
}

std::vector<Foo::ConstPtr> Bar::const_foos() const
{
    std::vector<Foo::ConstPtr> result;
    for ( auto i = 0; i < 100; ++i )
    {
        result.push_back( std::make_shared<Foo>() );
    }
    return result;
}
```

Here is the SWIG .i file used to generate the Ruby wrappers:

```
%module Blob

%include <std_shared_ptr.i>
%include <std_vector.i>

%{
#include "Blob.hpp"
%}

%shared_ptr(Foo);

%template (FooVector) std::vector<std::shared_ptr<Foo>>;
%template (FooConstVector) std::vector<std::shared_ptr<const Foo>>;

%include "Blob.hpp"
```

And, finally, here is a small Ruby script for reproducing the crash:

``` ruby
bar = Blob::Bar.new

foos = bar.foos  # OK
cfoos = bar.const_foos  # crash!
```

Only the `Bar::const_foos` member function, which returns `std::vector<std::shared_ptr<const Foo>>`, crashes. The `Bar::foos` member function, which returns `std::vector<std::shared_ptr<Foo>>`, does not have any problems.

I know SWIG does not attempt to preserve const-correctness (I can live with that), but I suspect crashing is not the desired behavior. For now, I've managed to work around the issue on the affected classes, using `%extend` and `%ignore`, by replacing the functions with wrappers that return vectors containing non-const objects.

[...]

PR #935, opened 2017-03-02
[ruby] avoid crashing due to std::shared_ptr of const Type.

tamuratak:
By adding a null check, we can avoid crashing as like mentioned in #586.
Or should we use assert here? 

Regards.


wsfulton:
I've added an assert as something has gone horribly wrong if type is passed in as NULL.

Issue #753, opened 2016-07-28
Problems with vector of shared_ptr<const X> in Python


Richard-Gmac:
I'm using SWIG to add a Python wrapper to a large existing C++ library. The library contains many std::shared_ptr to a const type and these are used extensively in containers. There seems to be a problem mapping vectors of such pointers into Python. I'm using the following SWIG file:

```
%module py

%include <std_shared_ptr.i>
%include <std_vector.i>

%{
#include <memory>
#include <vector>
using std::shared_ptr;
using std::make_shared;

struct DoubleValue {
        double value;
        DoubleValue();
        DoubleValue(double v);
};

typedef shared_ptr<DoubleValue> DoubleValuePtr;
typedef shared_ptr<const DoubleValue> ConstDoubleValuePtr;

double sumIVP(const std::vector<DoubleValuePtr>& is);
double sumConstIVP(const std::vector<ConstDoubleValuePtr>& is);
std::vector<DoubleValuePtr> toDV(const std::vector<double>& vs);
std::vector<ConstDoubleValuePtr> toConstDV(const std::vector<double>& vs);
%}

%shared_ptr(DoubleValue)

struct DoubleValue {
        double value;
        DoubleValue();
        DoubleValue(double v);
};

%typedef shared_ptr<DoubleValue> DoubleValuePtr;
%typedef shared_ptr<const DoubleValue> ConstDoubleValuePtr;

%template(DoubleVector) std::vector<double>;
%template(DoubleValuePtrVector) std::vector<DoubleValuePtr>;
%template(ConstDoubleValuePtrVector) std::vector<ConstDoubleValuePtr>;

%inline %{

DoubleValue::DoubleValue() : value(0){}

DoubleValue::DoubleValue(double v)
    : value(v)
{
}

double sumDV(const std::vector<DoubleValuePtr>& is)
{
    double s = 0;
    for(auto i : is)
    {
        s += i->value;
    }
    return s;
}

double sumConstDV(const std::vector<ConstDoubleValuePtr>& is)
{
    double s = 0;
    for(auto i : is)
    {
        s += i->value;
    }
    return s;
}

std::vector<DoubleValuePtr> toDV(const std::vector<double>& vs)
{
    std::vector<DoubleValuePtr> vec;
    for(auto x : vs)
    {
        vec.push_back(std::make_shared<DoubleValue>(x));
    }
    return vec;
}

std::vector<ConstDoubleValuePtr> toConstDV(const std::vector<double>& vs)
{
[...]

PR #937, opened 2017-03-03
[ruby] call swig::from with casting shared_ptr<const Type> to shared_ptr<Type>

tamuratak:
Crashing due to `std::shared_ptr<const Type>` is reported on #586 for Ruby, on #753 for Python. This PR fixes it for Ruby. The same solution might be applied to Python.

The root cause of this issue is the lacking of `%template` treating constness properly, which turns into  `SWIG_TypeQuery("std::shared_ptr<const Type>") == NULL`. 

In this PR, by using template specialization for `swig::from`, we remove the const qualifier of `std::shared_ptr<const Type>`  with `std::const_pointer_cast`, then call `swig::from` for the cast value.

This PR may conflict with #930 when merging. I will rebase and edit if needed.

Regards.

### DIFF
--- a/Examples/test-suite/cpp11_shared_ptr_const.i
+++ b/Examples/test-suite/cpp11_shared_ptr_const.i
@@ -1,0 +1,47 @@
+%module cpp11_shared_ptr_const
+
+%{
+
+#include <memory>
+#include <vector>
+
+class Foo
+{
+public:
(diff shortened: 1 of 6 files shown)


tamuratak:
This PR is merged into #930. So I close this PR.

PR #939, opened 2017-03-06
[ruby] must not do a null check for VALUE obj

tamuratak:
Doing a null check for `VALUE obj` is inappropriate. `VALUE obj` can be `0x0 == Qfalse`, which is a valid Ruby object. 

In the current master, calling `func([false])` in Ruby for `func(const std::vector<bool>&)` always causes TypeError. This PR fixes it.

Regards.


wsfulton:
Thanks @tamuratak, I'll merge this in.

tamuratak:
Thank you too. 

I am happy to rebase my other pull requests if needed.

PR #942, opened 2017-03-09
[Java] Call super in finalize override.

kennethshackleton:
Zealous static analysers complain about the missing call to `super.finalize()`. We address their complaint by changing the generated `finalize` override to call super.


wsfulton:
I need convincing that this change is correct, so am rejecting and closing this until otherwise convinced. The change leads to unnecessary additional code. There is no need to call `super.finalize()` due to the existing `super.delete()` call. If you want to make a trade-off between code efficiency and keeping your static analyzer happy, you are of course able to provide your own typemaps in your interface file containing the code in the patch.

PR #943, opened 2017-03-11
Fix dependency of SwigPySequence_Base fragment

vadz:
This fragment needs to depend on StdTraits fragment as it specializes the
template struct traits<> defined there and it could happen that the struct was
not declared before being specialized, resulting in compilation errors in the
generated code.

It doesn't seem to need to depend on stddef.h inclusion, however, so replace
it with the correct dependency instead of keeping both.

---

Unfortunately I can't provide a simple test case for the bug fixed by this change as it only arises when using multiple modules with `%template` declarations in both of them, but I hope this still can be applied as it seems to be "obviously correct" and just makes `SwigPySequence_Base` fragment consistent with the other fragments defined in the same file.

### DIFF
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -102,7 +102,7 @@ namespace swig {
 } 
 }
 
-%fragment("SwigPySequence_Base","header",fragment="<stddef.h>")
+%fragment("SwigPySequence_Base","header",fragment="StdTraits")
 {
 %#include <functional>
 


vadz:
Updated PR after rebasing it on the master version with the fixed `member_funcptr_galore` test.

wsfulton:
Sorry for the inconvenience @vadz, I think this will need rebasing again as Go is implemented quite differently and I've had to commit another change to get the full test-suite passing

vadz:
OK, done, thanks for the heads up.

vadz:
@wsfulton Any comments on/chance of applying this trivial patch now that the CI builds all pass? TIA!

wsfulton:
The change makes sense, but I couldn't quite see why the dependency on including stddef.h was dropped. Probably because the functional c++ header depends on std::size_t which probably includes stddef.h in some way. Note that the code in this fragment depends on ::size_t not std::size_t.

wsfulton:
If my comments are clear, ::size_t is defined in stddef.h.

vadz:
Originally I thought `stddef.h` would be included anyhow because it's a dependency of `SwigPyIterator_T` fragment used by `SwigPySequence_Cont`, but now I'm not sure any more. The fact that everything still compiles without it does seem to indicate that it's unnecessary, but there is no harm in adding it back too, if you prefer...

Actually, I think the simplest, and hence best, thing to do would be to always include `stddef.h` unconditionally from `pyinit.swg` (currently it's only done in `SWIGPYTHON_BUILTIN` case), this has negligible cost and would simplify things a little.

wsfulton:
Okay, I'll sort it out shortly...

PR #947, opened 2017-03-17
Add support for std::complex<> to C#

vadz:
Create Lib/csharp/complex.i with simplest possible typemaps mapping
std::complex<> to System.Numerics.Complex.

---

I'm submitting this because I absolutely need to export functions taking/returning `std::complex<>` in our code anyhow, and I think it could be useful to others (see e.g. this [old SO question](http://stackoverflow.com/q/18623406/15275)). However I'm not too happy about these typemaps, I'd really like to avoid exporting `Complex_T` to C# at all and just, somehow, pass the 2 doubles from C# to C++ code directly. Unfortunately I don't see how to do it, there is no `Int128` type in .NET that could be used to pack the two doubles together and I don't have any other ideas about how to do it, any suggestions of a better approach would be welcome.

But, again, in the meanwhile this is clearly better than nothing.


vadz:
I've now also updated configure to default to `mcs` instead of `gmcs`, so that we don't need to explicitly use `-sdk:4` option (which is not supported by Microsoft `csc`) for this test.

Of course, this does still mean that the test will fail if it runs with an older version of Mono, but I think we should be able to count on .NET 4 for running the tests nowadays (it still isn't required for the client code, as long as it doesn't use `<complex.i>`, of course).

vadz:
Argh, sorry, it turns out `csc` doesn't append `.dll` extension automatically, unlike Mono compiler, hence another update to add it.

vadz:
NB: CI builds pass now, Guile failure in Travis is completely unrelated to the changes here.

vadz:
I've added another commit, for complex-valued properties support, and had to rebase on the latest master to give CI a chance to test it because of a conflict in `CHANGES.current` (I wonder if it's possible to avoid these conflicts, e.g. by creating the change log before the release from some specially marked git commits instead of having to maintain and merge it manually...).

wsfulton:
I made some improvements to avoid generating a normal SWIG proxy class. @vadz, could you take a look at https://github.com/wsfulton/swig/commits/vadz-csharp-complex and let me know what you think?

PR #951, opened 2017-03-21
[Python] Apply docstring contents (if present) to builtin attributes

baldurk:
Hi! This is a local change I made to fix an issue I was seeing on my own generated interfaces. I use `-builtin` for generation, and I wasn't getting docstrings for member variables in structs.

It seems like this works OK for the non `-builtin` path as the docstrings are written out while the parse tree is being iterated, but for the `-builtin` route the members are deferred in `PyGetSetDef` entries and I didn't see a way to get the node later to get the docstring at the right time. This commit stores the docstring in the `PyGetSetDef` hash and retrieves it later.

I'm still new to SWIG, so apologies if I've made some error preparing the change, I tried to at least get things like indentation and style matching the existing code as much as possible. Let me know any changes I should make to get this PR in shape, or if I'm going off in completely the wrong direction with my fix.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -3311,6 +3311,22 @@ public:
       Setattr(h, "getter", "SwigPyObject_get___dict__");
     }
 
+    /* for builtin types, store the docstring (if it exists) in the PyGetSetDef entry. */
+    if (builtin && have_docstring(n)) {
+      String *memname = Getattr(n, "membervariableHandler:sym:name");
+      if (!memname)
+	memname = iname;
+      Hash *h = Getattr(builtin_getset, memname);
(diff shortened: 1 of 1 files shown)


baldurk:
From further experimentation and testing - the autodoc string from `AUTODOC_FUNC` doesn't really make much sense for the attribute getters/setters really.

I don't see a reliable way to suppress the autodoc for a given call to `cdocstring` though. Perhaps a new `AUTODOC_NONE` enum in `autodoc_t`? That way if there's no docstring provided, we fall back to the original `"Foo.bar"` string.

wsfulton:
docstrings for member variables were fixed by #728. Please try the latest and re-open if you spot anything wrong.

PR #953, opened 2017-03-24
Typo in Python.html

jschueller:


jschueller:
interesting doc, thanks @tleonhardt

PR #959, opened 2017-03-31
[Lua] Fix compiler warnings

jleveque:


wsfulton:
Which version of Lua and what are the compiler warnings? I'm just wondering if the change for lua_pushinteger will work with all versions of Lua or if this api has changed?

jleveque:
Lua version is 5.3.0.

Warnings are from Visual Studio 2015, as follows:

`warning C4244: 'argument': conversion from 'lua_Number' to 'lua_Integer', possible loss of data`
`warning C4244: 'initializing': conversion from 'long' to 'char', possible loss of data`

Issue #804, opened 2016-10-01
Support new/current v8 api


hmic:
Using the windows precompiled swig-3.0.10 I'm seeing multiple errors like
"error C2039: 'WeakCallbackData': is not a member of 'v8'"
or
"error C2660: 'v8::PersistentBasev8::Object::SetWeak': function does not take 2 arguments"
and 
"error C2039: 'GetHiddenValue': is not a member of 'v8::Object'"
when compiling a "swig -c++ -javascript -node" generated wrapper.
Looks like the v8 implementation support of swig stops somewhere before 5.0, because these functions got deprecated and removed back then...

Is nobody using the node and v8 support or am I doing something wrong here?

If so/not, would somebody be interested in upgrading/implementing support for (up to) the current v8 5.3 api? I'm considering me quite novice in C++, but I'd try to help if guided...

ojwb:
We don't seem to have a very active maintainer for SWIG's Javascript support, but people are clearly at least trying to use it, as we're getting issues and PRs filed:

https://github.com/swig/swig/labels/Javascript

Generally clean patches with regression tests get applied, patches with unclear intent or implementation and/or without any test coverage don't because maintainers without Javascript expertise don't feel comfortable that merging them does more good than harm.

hmic:
What about dropping the custom v8 versions support code and use/require NaN instead? This would provice a much broader support for current and upcoming v8 and node.js versions...

https://github.com/nodejs/nan

ojwb:
I know next to nothing about all this, but if someone's prepared to step in as SWIG's javascript backend maintainer and thinks that's the way to go, I've no objection personally.

ojwb:
From #825:
> Swig wrapper code makes use of WeakCallbackData which has been deprecated and is now removed in node v7.0.0. I think we have to move to WeakCallbackInfo instead, but I didn't look into it further.

SlickWish:
I used the branch here : https://github.com/arfoll/swig/tree/node7 as a guide and created my own patch. I haven't tested it extensively, but it seems to work. A long term solution would be to switch NaN.

Unfortunately, I also did the shared_ptr patch for 2014 within the same branch fixing the conflicts to bring it to 3.0.12 and the changes are mixed together.

arfoll:
@SlickWish you say you have your own patch based on my branch/fork? Would you mind submitting it to the fork as a PR so I could take a look?

SlickWish:
I will try to have it up soon.

[...]

PR #962, opened 2017-04-07
Add support for newer v8 used in Node 7.x

tomleavy:
Provides a solution for #804


tomleavy:
#968 works better. closing this one

Issue #837, opened 2016-11-15
Crash when using shared_ptr and directors


bpiwowar:
The bug happens when a method handled with director returns a null object (None in the case of Python), and when the object is handled by a shared_ptr.

The bug can be reproduced with the following files

- `test.i`
```swig
%module(directors="1") test

%{
#include "test.h"
%}

%include "std_shared_ptr.i";
%shared_ptr(Created);
%feature("director") Creator;

%include "test.h"
```
- `crash.py`
```python
import test

class Creator(test.Creator):
    def __init__(self, return_none):
        test.Creator.__init__(self)
        self.return_none = return_none
        
    def create(self):
        return None if self.return_none else test.Created()
        
print("No crash")
test.crash(Creator(False))

print("Crash")
test.crash(Creator(True))
```
- test.h
```c++
#include <memory>

struct Created {};

struct Creator {
	virtual std::shared_ptr<Created> create() = 0;
	virtual ~Creator() {}
};

void crash(Creator &creator) {
	std::shared_ptr<Created> ptr = creator.create();
}
```

This is due to the line in the swig generated file:
``` c++
c_result = *(reinterpret_cast< std::shared_ptr< Created > * >(swig_argp));
```
where `swig_argp` is a null pointer

pjohangustavsson:
Directors + std_shared_ptr at the same time in SWIG/Python is currently not stable. It is possible to work around using a good old fashioned reference counting system which works at least as well as smart pointers in garbage collected languages such as Python or C#.

ojwb:
> Directors + std_shared_ptr at the same time in SWIG/Python is currently not stable.

@pjohangustavsson What is your basis for this assertion?  I don't see any statements to this effect in the relevant documentation, nor any bug reports from you regarding such instability.

Anyway, it looks like a very similar issue was reported some time ago in #59, but seemingly never investigated.  The testcase there doesn't seem to involve directors.

bpiwowar:
The only work-around I found was to store each created object in a set (so objects are kept until the end of the execution, which is OK in my case). I looked a bit at the SWIG code, and I am unsure on how to modify the code to correct this. I guess the director related code should be aware of the fact that the pointer is a smart pointer.

PR #964, opened 2017-04-11
[Ruby] fix crashing when using shared_ptr.i and directors.

tamuratak:
Crashing when using shared_ptr and directors is reported in #837 for Python. The same issue exists for Ruby. This PR fix it for Ruby. 

The cause of this issue is that `%typemap(directorout)`  treating null shared_ptr properly  is not defined in `boost_shared_ptr.i`. `%typemap(directorin)` also has to be defined in `boost_shared_ptr.i` for the same reason.

This PR defines appropriate typemaps. `#error` directives are also added in the cases appropriate typemaps can not be defined.

### DIFF
--- a/Examples/test-suite/cpp11_shared_ptr_director.i
+++ b/Examples/test-suite/cpp11_shared_ptr_director.i
@@ -5,28 +5,54 @@
 %}
 
 %include "std_shared_ptr.i";
-%shared_ptr(Created);
-%feature("director") Creator;
+%shared_ptr(C);
+%feature("director") Base;
 
 %inline %{
-struct Created {
-  Created() {};
-  int get_m() { return 1; };
+struct C {
+  C() : m(1) {};
+  C(int n) : m(n) {};
+  int get_m() { return m; };
+  int m;
 };
 
-struct Creator {
-  Creator() {};
-  virtual std::shared_ptr<Created> create() = 0;
-  virtual ~Creator() {}
+struct Base {
+  Base() {};
+  virtual std::shared_ptr<C> ret_c_shared_ptr() = 0;
+  virtual C ret_c_by_value() = 0;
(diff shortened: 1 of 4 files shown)


Lib/ruby/boost_shared_ptr.i
```diff
@@ -24,7 +24,9 @@
 
 // Typemap customisations...
 
+//
```
wsfulton: I'd like the boost_shared_ptr.i typemaps to continue to be easily diff'd, so please remove the whitespace and comment changes unless they are replicated in all of the boost_shared_ptr.i files.

Examples/test-suite/cpp11_shared_ptr_director.i
```diff
@@ -0,0 +1,90 @@
+%module(directors="1") "cpp11_shared_ptr_director"
+
+%{
+#include <memory>
+%}
+
+%include "std_shared_ptr.i";
```
wsfulton: Please test using C++98 and boost_shared_ptr.i instead of std_shared_ptr.i as std::shared_ptr cannot currently be tested for all the languages that support shared_ptr in boost as these languages are not C++11 ready yet. Extending the typemaps to all the other languages which have boost_shared_ptr.i would be my ultimate goal, so better to set it up ready for the additions to the other target languages; replicating the typemap changes in all the other target languages won't be too hard.

Lib/ruby/boost_shared_ptr.i
tamuratak: I get it. I will edit this PR in that way. 

Regards.

tamuratak:
I leave the name of the test file unchanged because I can not get the rule of naming test files,  especially when to add the `li_` prefix.

Regards.

PR #968, opened 2017-04-13
Add Node 7.x aka V8 5.2+ support

p2k:
* Use WeakCallbackInfo instead of WeakCallbackData
* Use GetPrivate instead of GetHiddenValue
* Adopted new signature for SetWeak to support destructor calling
* SetAccessor deprecation fixed
* Proper version checks where applicable

This pull request is similar to the one from @tomleavy (i.e. @arfoll) but with less "hack" (no offence). It also fixes more warnings and resolves the apparent memory leak from not calling destructors. Credits go to the mentioned users for their inspiration.

### DIFF
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -133,9 +133,12 @@ static void $jswrapper(v8::Isolate *isolate, v8::Persistent<v8::Value> object, v
   SWIGV8_Proxy *proxy = static_cast<SWIGV8_Proxy *>(parameter);
 #elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < SWIGV8_SETWEAK_VERSION)
 static void $jswrapper(v8::Isolate *isolate, v8::Persistent<v8::Object> *object, SWIGV8_Proxy *proxy) {
-#else
+#elif (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
 static void $jswrapper(const v8::WeakCallbackData<v8::Object, SWIGV8_Proxy> &data) {
   v8::Local<v8::Object> object = data.GetValue();
+  SWIGV8_Proxy *proxy = data.GetParameter();
+#else
+  static void $jswrapper(const v8::WeakCallbackInfo<SWIGV8_Proxy> &data) {
   SWIGV8_Proxy *proxy = data.GetParameter();
 #endif
 
@@ -147,7 +150,9 @@ static void $jswrapper(const v8::WeakCallbackData<v8::Object, SWIGV8_Proxy> &dat
   }
   delete proxy;
 
+#if (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
   object.Clear();
+#endif
   
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031710)
   object.Dispose();
@@ -155,7 +160,7 @@ static void $jswrapper(const v8::WeakCallbackData<v8::Object, SWIGV8_Proxy> &dat
   object.Dispose(isolate);
 #elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032100)
   object->Dispose(isolate);
-#else
+#elif (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
   object->Dispose();
 #endif
 }
@@ -177,9 +182,12 @@ static void $jswrapper(v8::Isolate *isolate, v8::Persistent<v8::Value> object, v
   SWIGV8_Proxy *proxy = static_cast<SWIGV8_Proxy *>(parameter);
 #elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < SWIGV8_SETWEAK_VERSION)
 static void $jswrapper(v8::Isolate *isolate, v8::Persistent< v8::Object> *object, SWIGV8_Proxy *proxy) {
-#else
+#elif (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
 static void $jswrapper(const v8::WeakCallbackData<v8::Object, SWIGV8_Proxy> &data) {
   v8::Local<v8::Object> object = data.GetValue();
+  SWIGV8_Proxy *proxy = data.GetParameter();
+#else
+static void $jswrapper(const v8::WeakCallbackInfo<SWIGV8_Proxy> &data) {
   SWIGV8_Proxy *proxy = data.GetParameter();
 #endif
 
@@ -197,7 +205,7 @@ static void $jswrapper(const v8::WeakCallbackData<v8::Object, SWIGV8_Proxy> &dat
   object->Dispose(isolate);
 #elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < SWIGV8_SETWEAK_VERSION)
   object->Dispose();
-#else
+#elif (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
   object.Clear();
 #endif
 }
@@ -211,7 +219,11 @@ static void $jswrapper(const v8::WeakCallbackData<v8::Object, SWIGV8_Proxy> &dat
  * ----------------------------------------------------------------------------- */
 %fragment("js_getter", "templates")
 %{
+#if (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
 static SwigV8ReturnValue $jswrapper(v8::Local<v8::String> property, const SwigV8PropertyCallbackInfo &info) {
+#else
+static SwigV8ReturnValue $jswrapper(v8::Local<v8::Name> property, const SwigV8PropertyCallbackInfo &info) {
+#endif
   SWIGV8_HANDLESCOPE();
   
   v8::Handle<v8::Value> jsresult;
@@ -233,8 +245,11 @@ fail:
  * ----------------------------------------------------------------------------- */
 %fragment("js_setter", "templates")
 %{
-static void $jswrapper(v8::Local<v8::String> property, v8::Local<v8::Value> value,
-  const SwigV8PropertyCallbackInfoVoid &info) {
+#if (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
+static void $jswrapper(v8::Local<v8::String> property, v8::Local<v8::Value> value, const SwigV8PropertyCallbackInfoVoid &info) {
+#else
+static void $jswrapper(v8::Local<v8::Name> property, v8::Local<v8::Value> value, const SwigV8PropertyCallbackInfoVoid &info) {
+#endif
   SWIGV8_HANDLESCOPE();
   
   $jslocals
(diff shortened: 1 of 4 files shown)


p2k:
Provides a more complete solution for #804

arfoll:
@p2k no offence taken - my commit was labelled a hack ;-) This is really cool, I'll test it and see if it works for my usecase next week.

tomleavy:
@p2k thanks for taking over and getting this done right. No offense taken. I'm certainly not an expert with v8, I'm just excited to get my wrapper to work properly.

p2k:
I've had hours and days of fun with the oh so great V8 and I'm more than happy to have finally found SWIG which ends my glue code nightmares all at once. As soon as you have multiple projects that require Node wrappers (or any language) it gets obvious that it's easier to keep a code generator maintained than fixing everything over and over again.

aberaud:
Thanks, we used this PR for our project and it builds and works fine.

It would be great it that was merged soon, because the SWIG Node.js generator, as it currently ships from many sources, is broken and can't be used to build Node.js modules with distributed Node.js / V8 versions.

I think this is a nice and proper fix, and even it there was some issue, it would still be better than not being able to build at all. Version checking macros ensure that the behavior remains the same with older V8 versions. Having this in master would also give it more visibility so it would be tested by more people before an eventual release.

mislavn:
This PR also fixes our problem we had with node 7.x.

tomleavy:
+1

SergeySeroshtan:
@ojwb, could you please tell when this PR will be merged to the master?

ojwb:
The patch looks sane to me, but I know next to nothing about nodejs, and ideally I'd like to see CI coverage for this so I have confidence that it works.  That'd also help ensure it doesn't get accidentally broken by a future patch being applied.

As best I can make out, travis is currently testing a really old version:

```
Node.js: v0.10.36
node-gyp: v3.6.0
```

(At least I assume 0.10.36 < 7.x!)

That build appear to be on Ubuntu precise still - adding `dist: trusty` would probably give us a newer version, but not sure if it'd be new enough.

halton:
You could use nvm to manage to install different node versions for CI.

yegorich:
Successfully tested in Buildroot (swig 3.0.12 and Node.js 8.5).

wsfulton:
Is there anyone with C/C++ and Javascript experience who would like to step forward and maintain the Javascript module? Without a maintainer, it is hard to know if the various Javascript improvements like this one can't be added.

murillo128:
Hi! current support for node 7 & 8 is completely broken, so why not apply this patch and see what happens? I understand that it is far from ideal, but it won't work worst than now.

kopiro:
+1

ojwb:
@murillo128 I think we really need CI coverage for the newer node versions.  Otherwise "what happens" will be that someone will propose a patch that accidentally breaks support for them, and we'll merge it and not spot the breakage until some some time later when it's harder to deal with.

Ideally someone would step up to actually maintain the node backend, but without someone in that role CI coverage is particularly important.

murillo128:
@ojwb I agree, but in your worst case we will be at the same place as we are currently..  it can't get worse than now that it is broken.. ;)

wsfulton:
'Fixing' things more often than not introduces a regression and keeping the current code base working correctly is of highest priority as it affects the whole perception of how stable SWIG is. Without someone putting themselves forward to help maintain this backend, I'd rather keep the status quo as in my experience fixing regressions can be really time consuming for someone who isn't familiar with the backend. If anyone here really wants these fixes, please please put yourself forward and provide the rather minimal commitment to help out as your Javascript experience will be invaluable in moving things forwards rather than keeping the status quo. SWIG is an entirely volunteer project. It is very powerful and if you have gained anything from using it, then you ***should*** feel obliged to offer something back to the community, especially if you 'have an itch to scratch'.

murillo128:
Could we avoid generalities and focus on this patch? 

Let's recap:
 - Support for node 7 and 8 is broken, completely
 - All changes in the patch are encapsulated in an 
```
 +#elif (V8_MAJOR_VERSION-0) < 5 || (V8_MAJOR_VERSION == 5 && V8_MINOR_VERSION < 2)
    old code
 +#else
 +  new code
  #endif
``` 

So, this patch can't introduce regressions for older versions, and is tested by a few people to (including me) to fix support for node 7 and 8.

Again, I agree with the general view that you are proposing, but with regards to this particular patch, IMHO it should be safe and easy to merge it.

p2k:
Just to let you know, I'm about 80% towards stepping up and going for that job of maintaining the Node support including unit tests etc. I am held back by some deadlines in my company after which we'll schedule things better in 2018. So, figuratively speaking, you can expect the pain to go away with proper treatment in roughly one month. I will also reconsider the current API development of Node and have another look at NaN (see #971 if you like to add comments).

aberaud:
Hi,
Just a kind reminder that SWIG' master branch have been completely broken with NodeJS for a while and that every user **needs** this patch to get started with SWIG and NodeJS.

This PR has now been tested by quite a few people and it definitely fixes the build. Its design makes it very unlikely to introduce any regression (none where reported over almost a year). Do you think it might be merged soon ?

Many thanks

ojwb:
> Do you think it might get merged soon, regardless of the maintainer situation ?

Do you think someone might set up CI coverage for newer node versions soon?

That's the blocker to merging this, as least for me.  If there's CI coverage for newer node versions and it passes then we can be confident this works as intended, and more importantly we can also be confident that future patch submissions aren't going to break node support.  It's great if quite a few people have tested this by hand, but that really doesn't help us at all going forwards.

aberaud:
@ojwb I understand that there should ideally be CI coverage/maintainers/unit tests for everything, but the point is node support _is already broken_ and have been for a while.
So independently from the CI and maintainer situation (which should indeed be improved), merging this PR would at least allow users to make use of SWIG.

ojwb:
And yet the CI build for `SWIGLANG=javascript ENGINE=node` is still passing so I can't tell it's broken.

I understand the point you're making, but if we don't insist on CI coverage then we mostly won't have CI coverage, and we'll gradually return to the situation SWIG used to be in where about half the target languages didn't actually work properly most of the time, and fixing breakage was often hard because the change that broke things is no longer fresh in anyone's mind.  I really shouldn't have to explain why CI is so desirable to have.

Adding CI coverage with a recent node should not be a long or difficult job for somebody who knows node - there's even an existing CI build using an older node to start from.  This really doesn't seem an unreasonable requirement to me.

p2k:
Okay, I'll have a look at the CI stuff too. I had to learn a few things about Travis first in another project, also my personal workload became somewhat heavy again over the last two months. I believe I'm ready now and can write some proper tests; I make sure to add one that tests the memory leak as well which is fixed by the PR.

p2k:
@ojwb About this stepping up to maintain Node: I'd like to go for it. Where can I sign up for this? What would be my duties? I'm fine to discuss this at another place than this ticket, just point me where to start getting in touch. Cheers.

wsfulton:
@p2k, thanks for stepping up. I suggest sending an email to the swig-devel mailing list mentioning what you can do straight away and get some feedback on what folk think might need doing.  Some of the older javascript maintainers might still be monitoring the list. The responsibilities are to get involved as much or little as you are able to / would like to. The basics are though to review/comment on anything that is Node related. There are numerous patches in Github which need pulling in. The simple pre-conditions for accepting patches are to make sure the current Travis tests continue to work and that there is a test covering each commit (either new or existing) and within reason. If there isn't one, then write one yourself or request it from the patch's author. Let's discuss more on the mailing list.

ojwb:
@p2k Indeed, thanks.

Note that @nshmyrev has opened #1208 for sorting out travis, so you should probably coordinate to avoid duplicated work there.

We try to tag issues and PRs which are target language specific, so you can see all the open issues for a particular target language (or at least those that have been tagged):

https://github.com/swig/swig/labels/Javascript

There's only a "Javascript" tag at present, but in practice these are almost all about node.

p2k:
Can do. I also did see the PR and I'm going to put further conversation about the topic there.

"Javascript" basically covers the v8 engine which is used in Node.js, Electron and Safari plugins. I have worked with all three of these, so I see no problems there, should a specific issue arise. Also I'll simply be honest if I don't know something and can't do the research myself.

So I'll basically monitor the Javascript tag and any tasks that are assigned to me or where I get mentioned. Github sends me an email and depending on timezones/vacation/weekend I can probably react within 24h. That's what I can offer. Details follow on the mailing list tomorrow.

ojwb:
The SWIG manual says that SWIG's Javascript backend also supports "JavascriptCore, the Javascript engine used by Safari/Webkit" which seems to be an independent Javascript implementation to v8.  It seems to be hooked up in the CI.

Do Safari plugins really use a different Javascript engine to Safari itself?  That seems surprising (and I can't find evidence of it online).

p2k:
I'm going to double check this asap. v8 is courtesy of Google, but open source; it might still not be a part of WebKit/Safari in the near future.

PS: A project in my company recently escalated, which forces me to delay my application on the SWIG mailing list. I should be ready in one or two days.

ojwb:
AIUI, Safari's Javascript engine is also FOSS but derives from KDE's KJS not v8.

But really having someone actively maintaining only the v8-based Javascript targets in SWIG would be a big improvement, and the CI should at least ensure that JSC support doesn't get accidentally broken by v8-related changes.

notedit:
hello  any process about this PR.  NODE 9 can not build.

udp:
Seems odd that the official reason this isn't being merged is the lack of tests ("it might not work") when the bindings swig currently generates are useless for recent node versions ("it definitely does not work").  Forgive me if I'm wrong, but this looks more like the maintainers don't use nodejs and therefore don't care if this gets merged or not.

Why not at least merge this as a separate "node7+" target so that people can use it without having to merge a patch and compile swig from source?

p2k:
It's also stuck, because the guy that wants to become maintainer (and does use nodejs and does care) is still under arrest by high work load in his company.

murillo128:
This is simply a nonsense.

SWIG does not support neither current nodejs version (v9) nor LTS (v8). Claiming that this patch breaks compatibility with **unsupported** node version is the worst excuse ever.

If you have a node<v7 project it is expected that things stop working for you, swig or any other dependency.

ojwb:
All we're asking for is to have CI coverage for the new node versions that this patch aims to support in order to demonstrate that the patch works, and to ensure that this support isn't broken by future patches.

I'm frankly amazed at the amount of time and energy people seem prepared to put in to arguing for us to merge this patch without having any CI coverage - far more in total now than it would have taken for someone to just sort the necessary CI changes.

murillo128:
FWIW node 10.0.0 it's going or tomorrow..

https://github.com/nodejs/Release/issues/328

So swig will not support node 7,8,9 and 10.

Regarding CI , is there any support for node right now? What CI tests do you expect to be added and what should they cover? How can the CI changes be tested easily?

murillo128:
Node 0.10.36 is from 2015... 

The nodejs packaged in ubuntu is unusable. The best way of installing specific node versions is using nvm (which do not require root):

https://github.com/creationix/nvm

```
wget -qO- https://raw.githubusercontent.com/creationix/nvm/v0.33.10/install.sh | bash
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" # This loads nvm
```

Then installing & using an specific node version is: 
```
nvm install 6
nvm install 7
nvm install 8
nvm install 9
nvm install 10
nvm use 6
```
So the question is, is that enough?

RebelSyntax:
so, for someone who just looked at swig and node-gyp for the first time today and is wrapping a rather complex shared object, is there an easy way to fix this on my local box?  I am fairly certain that I compiled the latest distribution.  I'd love to jump in and see what I can do with travis and ci coverage, cause I planned on experimenting with that technology here as well.  But, alas, I must find a way to consume this shared object in my electron app as a top priority.  really don't want to go the FFI route based on what I have read and the time I spent making the choice between the two.  Thanks!

yatli:
@RebelSyntax just copy the updated swg files to your swig installation.

sancelot:
I don't really understand what is happening but I don't manage to use it with 8.11.2.

colinxsmith:
node 6 was messed up because there's a bug #elifif instead of #elif in the 4'th file to be patched.

wsfulton:
#1290 from @furylynx contains builds on this patch to add CI tests on Travis. Please take a look as I intend to merge it in soon.

wsfulton:
Merged via #1290. Please check using SWIG master.

PR #969, opened 2017-04-13
Added check to prevent crash on illegal constructor call

### LINKED ISSUE #626: [node.js] addons crash with calling constructor without new
Hello,
I swig a simple class like this:

``` cpp
class Base {
public:
        Base(const int& y, const int& m, const int& d)
                : year(y), month(m), day(d)
        {
        }
        int year;
        int month;
        int day;
};
```

and wrap.i like this:

```
%{
#include "base.h"
%}
%include "base.h"
```

when I call the constructor of swiged Base, I got this **CRASH** install of **Exception**

``` js
> var BASE = require('.');
undefined
> var a = BASE.Base(1, 2, 3);
```

``` cpp
FATAL ERROR: v8::Object::SetAlignedPointerInInternalField() Internal field out of bounds
```

Is this a bug? Swig version is 3.0.7

p2k:
Constructors not called as part of object instantiation (using "new" or via inheritance) will not crash the VM anymore. Fixes #626.

### DIFF
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -10,9 +10,10 @@
 %fragment("js_ctor", "templates") %{
 static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args) {
   SWIGV8_HANDLESCOPE();
-  
+
   v8::Handle<v8::Object> self = args.Holder();
   $jslocals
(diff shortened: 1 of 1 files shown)


ojwb:
CI failures are from appveyor and all for other languages, so clearly unrelated.

Is it feasible to add a testcase to show this fix works?  I'm not familiar with nodejs or v8, and would be more comfortable applying this if it was clearly definitely fixing the case it's supposed to.

p2k:
Yes, why not. Will add a test as soon as I have some free time.

Issue #360, opened 2015-03-18
Feature Request: Specify the version string in PHP generated code


dhait:
PHP language module emits a hardcoded "NO_VERSION_YET" for version identifier in the zend_module_entry structure.
See Source/Modules/php.cxx at about line 537:

```
Printf(s_init, "    NO_VERSION_YET,\n");
```

It would be nice to be able to specify a version string at swig time, so that (for example)

```
echo "The version is" . phpversion('MyModule') . '\n'
```

gives a useful version (such as a SemVer).

ojwb:
A patch to address this would be great.

PR #970, opened 2017-04-14
Addressing issue #360, adding pragma version directive for php5 and php7

nihal95:
Example - %pragma(php) version="XX.XX"

### DIFF
--- a/Doc/Manual/Php.html
+++ b/Doc/Manual/Php.html
@@ -821,6 +821,15 @@ echo "example.php execution\n";
 </pre></div>
 
 <p>
+The <b>version</b> pragma can be used to add version to generated PHP extension module. The version is inserted in the zend_module_entry block.
+</p>
+
+<div class="code"><pre>
+%module example
+%pragma(php) version="1.5"
+</pre></div>
+
+<p>
 The <b>include</b> pragma is a short cut to add include statements to
 the example.php file.
 </p>
--- a/Examples/php/pragmas/runme.php
+++ b/Examples/php/pragmas/runme.php
@@ -2,4 +2,5 @@
 
 require "example.php";
 
+echo "Version - " . ((new ReflectionExtension('example'))->getVersion());
 ?>
--- a/Examples/test-suite/php/php_pragma_runme.php
+++ b/Examples/test-suite/php/php_pragma_runme.php
@@ -1,0 +1,12 @@
+<?php
+
+require "tests.php";
+require "php_pragma.php";
+
+
+check::equal('1.5',(new ReflectionExtension('php_pragma'))->getVersion(),"1.5==version(php_pragma)");
+
+check::done();
+
+?>
+
--- a/Source/Modules/php.cxx
+++ b/Source/Modules/php.cxx
@@ -96,6 +96,7 @@ static String *all_cs_entry;
 static String *pragma_incl;
 static String *pragma_code;
 static String *pragma_phpinfo;
+static String *pragma_version;
 static String *s_oowrappers;
 static String *s_fakeoowrappers;
 static String *s_phpclasses;
@@ -359,6 +360,7 @@ public:
     /* sub-sections of the php file */
     pragma_code = NewStringEmpty();
     pragma_incl = NewStringEmpty();
+    pragma_version = NULL;
 
     /* Initialize the rest of the module */
 
@@ -515,7 +517,11 @@ public:
       } else {
 	Printf(s_init, "    NULL, /* No MINFO code */\n");
       }
-      Printf(s_init, "    NO_VERSION_YET,\n");
+      if (Len(pragma_version) > 0) {
+	Printf(s_init, "    \"%s\",\n", pragma_version);
+      } else {
+	Printf(s_init, "    NO_VERSION_YET,\n");
+      }
       Printf(s_init, "    STANDARD_MODULE_PROPERTIES\n");
       Printf(s_init, "};\n");
       Printf(s_init, "zend_module_entry* SWIG_module_entry = &%s_module_entry;\n\n", module);
@@ -2006,6 +2012,10 @@ done:
 	} else if (Strcmp(type, "phpinfo") == 0) {
 	  if (value) {
 	    Printf(pragma_phpinfo, "%s\n", value);
+	  }
+	} else if (Strcmp(type, "version") == 0) {
+	  if (value) {
+	    pragma_version = value;
 	  }
 	} else {
 	  Swig_warning(WARN_PHP_UNKNOWN_PRAGMA, input_file, line_number, "Unrecognized pragma <%s>.\n", type);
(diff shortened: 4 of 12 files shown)


Source/Modules/php.cxx
```diff
@@ -2007,6 +2013,10 @@ class PHP : public Language {
 	  if (value) {
 	    Printf(pragma_phpinfo, "%s\n", value);
 	  }
+	} else if (Strcmp(type, "version") == 0) {
+	  if (value) {
+	    Printf(pragma_version, "%s", value);
```
ojwb: This appends `value` to any existing contents of `pragma_version`, which is a bad way to handle multiple occurrences of this pragma.

I'd suggest initialising `pragma_version` to `NULL` by default, and then you can probably just assign `value` to it here (`value` points to a string in the parse tree which should stay valid while we need it).

Source/Modules/php5.cxx
```diff
@@ -528,7 +530,14 @@ class PHP5 : public Language {
...
     Printf(s_entry, "static zend_function_entry %s_functions[] = {\n", module);
 
+    /* Emit all of the code */
+    Language::top(n);
+
+    SwigPHP_emit_resource_registrations();
+
     /* start the init section */
+    String * s_init_old = s_init;
+    s_init = NewString("/* init section */\n");
```
ojwb: These refactoring changes are not obviously related to the change you say you're making, so without any explanation it looks like you've accidentally included some unrelated change in the commit.  I can guess roughly why you might need this, but only because this is essentially the same change I had to make while adding PHP7 support - someone unfamiliar with that work wouldn't know that.

Probably the best way to address this is to perform this refactor first in a separate commit, then add the new feature in a second commit (the PR would include both commits).

Source/Modules/php5.cxx
```diff
@@ -542,7 +551,13 @@ class PHP5 : public Language {
     Printf(s_init, "    PHP_RINIT(%s),\n", module);
     Printf(s_init, "    PHP_RSHUTDOWN(%s),\n", module);
     Printf(s_init, "    PHP_MINFO(%s),\n", module);
-    Printf(s_init, "    NO_VERSION_YET,\n");
+
+    if(Len(pragma_version) > 0) {
```
ojwb: SWIG coding style has a space between `if` and `(`.

ojwb:
A new feature like this should also be documented (I think you mentioned adding that already), and have test coverage, so we're confident it actually works (both when merged and in the future as the code evolves).

Looks like you can get the version of an extension in PHP easily:

http://php.net/manual/en/reflectionextension.getversion.php

Source/Modules/php5.cxx
nihal95: Hey, I've applied all the changes, do take a look again. And I've also added the Documentation and Example code as a separate commit.

nihal95:
Hey olly, can u check the pr and let me know. Thanks

Source/Modules/php.cxx
ojwb: You've not done what I suggested here - did you try it and find it didn't work?

If we do it your way, then ideally we don't leak memory if this pragma is used more than once.  Also, using `Copy()` would be a clearer way to make a copy.

Doc/Manual/Php.html
```diff
@@ -821,6 +821,15 @@
 </pre></div>
 
 <p>
+The <b>version</b> pragma can be used to add version to PHP module in the generated "example.php". The version is inserted in the zend_module_entry block.
```
ojwb: That's not quite right - it adds a version to the generated **PHP extension** (not the generated PHP wrapper script).

Examples/php/pragmas/runme.php
```diff
@@ -2,4 +2,5 @@
 
 require "example.php";
 
+echo "Version - " . ((new ReflectionExtension('example'))->getVersion());
 ?>
```
ojwb: Good, except this is an example, not part of the testsuite.  If someone in the future makes a change which breaks your new feature or it doesn't work correctly with a new version of PHP we won't get a test failure.  Running this example will print the wrong thing, but not flag an error.

We really want an automated test which will fail if the version reported via reflection doesn't match that set with the pragma - that should be done via the automated testsuite under `Examples/test-suite`.  There's support for language-specific tests there (and you'll need to get familiar with that testsuite for your gsoc project).

Having an automated feature test also gives a patch reviewer more confidence your implementation works as it should - CI runs the testsuite automatically on each PR for most languages (including PHP 5 and 7).

ojwb: Did you get a chance to sort out the loose ends?  This is pretty close to being ready to merge.

nihal95: Hi. I'll be taking it up from now. My exams just finished. Will tie up loose ends and finish this.

nihal95:
Hi olly, can u check this and let me know. Thanks

Source/Modules/php.cxx
nihal95: Hey, You suggested initialising pragma_version to NULL by default and then assign the value to it here. Probably did you mean pragma_version pointing to where value is pointing to, since both are string pointers. Instead of allocating new resource of a String to pragma_version pointer. 
` pragma_version = value `

ojwb: Yes, exactly.  That string is in the parse tree, so should remain valid for as long as we need it to.

Examples/test-suite/php/php_pragma_runme.php
```diff
@@ -0,0 +1,11 @@
...
+
+require "tests.php";
+require "php_pragma.php";
+
+
+check::equal('1.5',(new ReflectionExtension('php_pragma'))->getVersion(),"1.5==version(php_pragma)");
+
+check::done();
+
+?>
```
ojwb: The red symbols github shows in the diff here mean there's no newline at the end of this file (if you hold the mouse pointer over them it'll even tell you).

It's a good idea to make sure text files have a newline at the end.  Some tools don't cope well if it's missing, and some editors automatically add one if there isn't one creating an unrelated whitespace change when someone using such an editor tries to update the file.

As a general rule of thumb, any differences in a source file you can't see are bad - so the last line should have a newline, there shouldn't be trailing whitespace at the end of any line, you shouldn't indent some lines with just spaces and use tabs in others (most code bases adopt a rule for how indents should be done), and there shouldn't be trailing blank lines at the end of the file.  There are cases where such whitespace can change the meaning of a file (e.g. `\` at the end of a line in a shell script or makefile is a line continuation, but if there's a space after it, it escapes that space instead!)

Examples/php/pragmas/runme.php
ojwb: Thanks for sorting out the tests - aside from the "missing newline at end of file" nit, those look good.

Source/Modules/php.cxx
nihal95: So shall I change the commit with this change?

nihal95:
I've sorted out the issues. Thanks

Examples/test-suite/php/php_pragma_runme.php
nihal95: I've taken care of it. Thanks. And will keep this in mind

Issue #59, opened 2013-06-06
std::vector< boost::shared_ptr< X > > in python: iteration and indexing produce different results


dstndstn:
Hi,

I am finding that when I put an empty (ie, pointing to NULL) boost::shared_ptr into a std::vector, and then look at it in Python, the iteration and indexing results are different.  That is,

```
X = myswigged.getListOfStuff()
for i,x in enumerate(X):
    print '  iteration:', x
    print '  indexing:', X[i]
```

produce different results; indexing notices that the empty (null) boost::shared_ptr is null and returns None; the iteration does not, producing a seemingly-valid boost::shared_ptr that will explode upon use.

Minimal test case:   http://broiler.astrometry.net/~dstn/temp/test.tgz

```
wget http://broiler.astrometry.net/~dstn/temp/test.tgz
tar xzf test.tgz
python setup.py build_ext --inplace
python runtest.py
```

I have verified that this occurs in swig 2.0.10 and git master.

thanks,
dustin

ojwb:
Still present in git master.

Here's an all-in-one version of the testcase:

```
%module testcase

%include "std_vector.i"
%include "boost_shared_ptr.i"

%inline %{
#include <vector>
#include <boost/shared_ptr.hpp>

class MyThing {
public:
MyThing(int x) : _x(x) {}
private:
	int _x;
};

std::vector<boost::shared_ptr<MyThing> > getListOfMyThings();

long getAddressOf(boost::shared_ptr<MyThing> x);
%}

%{
std::vector<boost::shared_ptr<MyThing> > getListOfMyThings() {
	// return value:
	std::vector<boost::shared_ptr<MyThing> > lst;
	// push non-empty
	lst.push_back(boost::shared_ptr<MyThing>(new MyThing(42)));
	// push empty
	lst.push_back(boost::shared_ptr<MyThing>());
	return lst;
}

long getAddressOf(boost::shared_ptr<MyThing> x) {
	return (long)x.get();
}
%}

%shared_ptr(MyThing);
%template(MyThingPtr) boost::shared_ptr<MyThing>;
%template(MyThingPtrList) std::vector<boost::shared_ptr<MyThing> >;
```

Save as `testcase.i`, process and run with:

```
swig -c++ -python testcase.i
g++ `python-config --includes --cflags` -shared -fPIC -o _testcase.so -Wall -W -O2 testcase_wrap.cxx `python-config --libs`
python runtest.py
```

where `runtest.py` is:

```
import testcase

[...]

PR #972, opened 2017-04-19
[ruby] null shared_ptr in std containers.

tamuratak:
Hi,
in #59, it is reported that null shared_ptr in std containers is not properly treated for Python. The same issue exists for Ruby. This PR fixes it for Ruby.

This PR depends on the PR #930, which I hope will be merged soon.  You can see the [diff](https://github.com/tamuratak/swig/compare/shared_ptr_derived_2...tamuratak:fix_ruby_null_shared_ptr) between #930 and this PR.

The way of fixing this issue is similar to the one in #930. We define `traits_from_ptr<std::shared_ptr<Type>>::from` by using template specialization. `traits_asptr<std::shared_ptr<Type>>::asptr` defined in #930 also needs a patch.

Regards.

### DIFF
--- a/Examples/test-suite/cpp11_shared_ptr_nullptr_in_containers.i
+++ b/Examples/test-suite/cpp11_shared_ptr_nullptr_in_containers.i
@@ -3,6 +3,8 @@
 %{
 #include <memory>
 #include <vector>
+
+class C;
 %}
 
 %include <std_shared_ptr.i>
@@ -38,7 +40,7 @@ public:
   }
 
   bool is_last_null(const std::vector<std::shared_ptr<C> >& v) {
-    if( *v.end() ) {
+    if( v.back() ) {
       return false;
     } else {
       return true;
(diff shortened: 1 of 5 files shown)


wsfulton:
I think this is superceded by some fixes I put in 83a389d3fbccbd588f3a649c4a736feeae617341. Please check and close if okay now.

tamuratak:
I will check it soon.

Regards.

tamuratak:
Hi,
this PR still seems necessary.  

For example in the test of this PR,  although `std::vector<std::shared_ptr<C>>` returned by `a = ret_vec_c_shared_ptr()` includes a null shared_ptr as the second element, `c = a[1]` returns a class C object, not `nil`, without this PR. 

Then calling `c.get_m()` will crash. 

We need to do a null check for `std::shared_ptr<Type>` somewhere when converting it into `VALUE`.

Regards.

Issue #526, opened 2015-09-22
C++11 noexcept should propagate to director classes


VojtechFried:
Tested with python.
Say I have an interface class IFoo:

``` c++
class IFoo 
{
public:
    virtual void bar() noexcept = 0;
};
```

Swig generates a director class:

``` c++
class SwigDirector_IFoo : public Strs::DocPlatform::IFoo, public Swig::Director {
public:
    SwigDirector_IFoo(PyObject *self);
    virtual void bar();
...
};
```

The compiler then complains, that "overriding virtual function has less restrictive exception specification than base class virtual member function". Swig should put C++11 noexcept to bar method of SwigDirector_IFoo.

wsfulton:
Patch #975 fixed this.

PR #975, opened 2017-04-25
Issue #526 Fix propagate noexcept to director classes

yag00:
propagate noexcept to director classes for python and java binding

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -543,6 +543,7 @@ CPP11_TEST_CASES += \
 	cpp11_default_delete \
 	cpp11_delegating_constructors \
 	cpp11_director_enums \
+	cpp11_directors \
 	cpp11_explicit_conversion_operators \
 	cpp11_final_override \
 	cpp11_function_objects \
--- a/Examples/test-suite/cpp11_directors.i
+++ b/Examples/test-suite/cpp11_directors.i
@@ -1,0 +1,20 @@
+%module(directors="1") cpp11_directors
+%feature("director");
+
+%{
+
+class Foo {
+  public:
+    virtual ~Foo() noexcept {}
+    virtual int ping() noexcept = 0;
+    virtual int pong() noexcept = 0;
+};
+
+%}
+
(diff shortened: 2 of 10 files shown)


wsfulton:
Patch looks okay, it just needs a testcase. If you are feeling brave, please add in the same modifications to the other director supported languages... the same testcase will test them all very easily.

yag00:
@wsfulton yes I can add the patch for other language. I didn't do it yet because I had no way to test it. How should I implement the testcase?

wsfulton:
See http://swig.org/Doc3.0/Extending.html#Extending_test_suite. Add a new testcase, which I suggest you call cpp11_directors.i in Examples/test-suite and add it into CPP11_TEST_CASES in Examples/test-suite/common.mk as it requires c++11. Then run it using `make cpp11_directors.cpptest` in each of the language subdirectories such as Examples/test-suite/java. The C++11 languages it needs to work with are java, csharp, python, ruby, octave as those are the languages we test under C++11. php, d, go, perl5 are also director enabled languages, but the tests for these might not work as their support for C++11 is variable, but probably it'll be okay for this simple C++11 feature once you've added the support.

yag00:
Ok I add support for other languages and a test-case. 
Some trouble that I did not anticipate. Generated c++ code throw a swig exception for java and csharp.
 - For csharp I just replaced the throw by an assert for noexcept case only (9dcf101)
 - For java I removed the throw for noexcept case only  (e13eaaa)
Let me know If you want to handle this differently.

wsfulton:
The other languages also throw DirectorPureVirtualException c++ exceptions, but the generated code does it via helper functions. I've implemented the same for D, Java, C#. This foxes the C++ compiler to not issue a warning. Not exactly a fix, but directors are designed to throw exceptions in some cases, so this will have to do for now. Java can use the `director:except` typemap to customize the generated code to do something else and this is covered in the Java docs, but support for this typemap still needs to be extended to other target languages. 

#349 might be of interest as it attempts to remove exceptions from directors.

Issue #974, opened 2017-04-25
Bug handling comments


redbrain:
I just found recently a bug in the parser when handling comments:

```c
#ifndef __BUG_H__
#define __BUG_H__


#define TCX_PACKED(d) d __attribute__ ((__packed__))


TCX_PACKED (typedef struct tcxMessageTestImpl
{
    int mHeader; /**< comment */
}) tcxMessageTest;


TCX_PACKED (typedef struct tcxMessageBugImpl
{
    int mBid; /**< Bid price and size, check PresentMap if available in message */
}) tcxMessageBug;


#endif //__BUG_H__
```

My interface:

```
%module TcxBug

%{
#include "bug.h"
%}

%define __attribute__(x)
%enddef

%import "stdint.i";
%include "bug.h"
```

Notice in the struct with the long comment it uses a comma. Removing comma's within the comments everything is handled. Tested against master and is still there. I am trying to look into the parser at the moment to see if i can make a fix.

redbrain:
Forgot to show the error message:

```
$ swig -Wall -python -outcurrentdir interface.i
bug.h:14: Error: Macro 'TCX_PACKED' expects 1 argument
```

PR #976, opened 2017-04-25
Fix bug with comments inline in macros

redbrain:
Addresses #974

### DIFF
--- a/Source/Preprocessor/cpp.c
+++ b/Source/Preprocessor/cpp.c
@@ -607,6 +607,23 @@ static List *find_args(String *s, int ismacro, String *macro_name) {
 	skip_tochar(s, '\'', str);
 	c = Getc(s);
 	continue;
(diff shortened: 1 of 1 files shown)


Source/Preprocessor/cpp.c
```diff
@@ -607,6 +607,23 @@ static List *find_args(String *s, int ismacro, String *macro_name) {
 	skip_tochar(s, '\'', str);
 	c = Getc(s);
 	continue;
+      } else if (c == '/') {
+          /* Ensure comments are ignored by eating up the characters */
+          c = Getc(s);
+          if (c == '*') {
+              while (1) {
```
wsfulton: This loop may never exit. Check for ```(c != EOF)``` instead.

redbrain: Thanks for the review i will fix that.

PR #977, opened 2017-04-26
[ruby] director should not use nesting counter

tamuratak:
Hi,
director implementation for ruby now uses a counter of nesting.
When ruby level exception occurs during calling a callback function and the counter > 0, rb_raise will be called. Then c++ objects on stack at the moment will  not be destroyed properly because exception of ruby is implemented by using setjmp and longjmp.

If we execute the attached test without this PR, c++ objects call_ret_m (2 to 11) puts on stack will not be destroyed.

```
1 call_ret_m
1 d.ret_m   # (1-b)
2 call_ret_m
2 d.ret_m
....
10 call_ret_m
10 d.ret_m
11 call_ret_m
11 d.ret_m  # < rb_raise here (longjmp to 1-b)
```

This PR fixes this issue. I can not think of cases the nesting counter.  If they exist, we need another fix. Any idea for such a case?

Regards.


wsfulton:
The original depth count code seems to be from this commit 60ded6c0c2791b80577d7fb97c2a0a277858b7fc. Perhaps it was put in for the case when an exception is thrown from code that has gone in and out of Ruby and C++ code multiple times via director methods? Unfortunately there wasn't an associated test with the commit, but I think we should try and replicate a testcase for the original problem before ripping out this code.

Lib/ruby/director.swg
```diff
@@ -115,7 +115,7 @@ namespace Swig {
     VALUE swig_error;
     std::string swig_msg;
   protected:
-    DirectorException(VALUE error) : swig_error(error) {
+    DirectorException(VALUE error, int s = 0) : swig_error(error), status(s) {
     }
 
     DirectorException(VALUE error, const char *hdr, const char *msg ="") : swig_error(error), swig_msg(hdr) {
```
wsfulton: This constructor does not initialize `status`.

Lib/ruby/director.swg
```diff
@@ -151,6 +151,8 @@ namespace Swig {
     const char *what() const throw() {
       return swig_msg.c_str();
     }
+
+    int status;
```
wsfulton: This public member variable should be private with accessors added. Also, if I rerun the test-suite making this private it shows that it doesn't get used, so am wondering what is the purpose storing it?

tamuratak: Hi,
calling `rb_jump_tag(status)` is a standard way to re-raise an exception raised while`rb_protect` called. Please see [reference](https://docs.ruby-lang.org/en/2.4.0/extension_rdoc.html#label-Control+Structure).  I am not sure calling `rb_raise(error)` or `rb_exc_raise(error)` always works. For that reason, I have added the public member variable.

wsfulton: Let's not store status unless we know it is needed. It can be added in if someone requires it, then we'll have a valid use case too.

wsfulton:
Ping @tamuratak . Please address the review comments in order for the patch to be considered any further.

PR #984, opened 2017-04-29
[ruby] check whether object is of RTypedData.

tamuratak:
Hi,
`struct RTypedData` has been introduced from ruby-1.9.3, which is not used in the current SWIG, in which `struct RData` is used. 

However `TYPE(obj) == T_DATA` holds for the both `struct RData` objects and `struct RTypedData` objects. So we must check which an object is even if `TYPE(obj) == T_DATA` holds for the object.

Without this PR, overloading in the attached test does not work. This PR fixes that.

Regards.

### DIFF
--- a/Examples/test-suite/ruby/Makefile.in
+++ b/Examples/test-suite/ruby/Makefile.in
@@ -25,6 +25,7 @@ CPP_TEST_CASES = \
 	ruby_keywords \
 	ruby_minherit_shared_ptr \
 	ruby_naming \
+	ruby_rdata \
 	ruby_track_objects \
 	ruby_track_objects_directors \
 	std_containers \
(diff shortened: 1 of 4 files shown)


wsfulton:
Presumably `RTYPEDDATA_P` was introduced in 1.9.3? The docs say 1.8 is known to work. We could change that minimum to 1.9, but if it is easy to keep 1.8 working that would be helpful. There are some backwards compatible macros in https://github.com/swig/swig/blob/master/Lib/ruby/rubyhead.swg#L50. If you don't have 1.8.x handy, I can try something out, just let me know.

wsfulton:
I added in support for Ruby < 1.9.3 in 76bf7c04a3c77f419e94661e744cf02579394116.

wsfulton:
Thanks, merged, see ff9e9f9f98e195764a1f7ec0c34d8c311a7fd6f8.

PR #985, opened 2017-05-02
Swig for Delphi

lynxnake:
This is an initial implementation of SWIG for Delphi.


wsfulton:
Hi @lynxnake 

Firstly, thanks for the contribution and the docs which I will be reading next.

I see you have added the examples and test-suite. The most important next step is to set up Travis so we can these working. Can you get Travis working before you make any further changes? You need to know what `apt-get install xxxx commands` are needed to install delphi on an Ubuntu Precise machine. If you need help with Travis set up let me know... my first question will be the answer to installing delphi. I suggest you get Travis working in your own FMXExpress:master branch before submitting any more pull requests. I'd remove all the other languages from your .travis.yml file while setting up Travis otherwise you need to wait about 5 hours for all the other Travis builds to complete.

wsfulton:
Sorry, I assumed delphi was available on Linux too. Does the generated code also work with Object Pascal? If so, Travis can be still be used for testing. Appveyor will need to be set up to test Delphi proper on Windows. We currently use Appveyor for testing Windows and C#, Java, Python.

lynxnake:
Hi @wsfulton,

At first, I must say I am neither the original developer of the Delphi extension nor the "copyright owner". It has been brought to life mostly by the generous efforts of @FMXExpress, with the help of couple more developers. I just the one to whom these developers handed over the project.

About your note on auto-builds and tests: In order to "fix" the build broken here I already set up the Appveyor build (to the extent I understood). Look https://ci.appveyor.com/project/lynxnake/swig-ojrr4. However, as you can see, in my personal repo Appveyor is building everything ok, so I didn't get why the build broke here.

About the Delphi - to be exact, it's actually had just became available on Linux with the latest 10.2 release. However, you are right, this extension was developed without Linux.

Ok, I will set up Appveyor to build FMXExpress:master.

As I am not the original developer, and he is not easily accessible, perhaps you will be able to hint/advise me - is there a way to make all the examples at once? I see there is a check.list file, but no any Makefiles in Examples/delphi

It seems to be all for now,
Thank you for attention,
--
Serhii

lynxnake:
Ok, I set up Appveyor for my personal fork: https://ci.appveyor.com/project/lynxnake/swig-ojrr4 Also removed the other languages. Say please if you need anything more.

lynxnake:
Seems that the will be no free Delphi for the Linux: https://www.embarcadero.com/docs/rad-studio-feature-matrix.pdf
The only free (as in 'free beer') edition of it is Delphi Starter, but it supports only Win32.
So, if Linux Delphi is a must, I would try Free Pascal. As far as I see there is no need in IDE thus in Lazarus. Also, I would not say that the Delphi code generated is very different from Free Pascal

lynxnake:
... So, for the FPC Ubuntu's `apt-get` command is `sudo apt-get install fp-compiler`
However, I'm unsure our code will start work with fpc as it is now. So, if we will go with FPC, say, please, I will try to adapt the code to it.

lynxnake:
...Last but not least - my English is far from perfect, so feel free to ask any question

wsfulton:
Your English is fine for me! You seem to have done the right thing and got Appveyor builds to run. It'll probably take quite a bit of tweaking to get it to work. You probably want to get it building on a local computer first with the same environment (mingw), then finish setting it up on Appveyor.

Linux Delphi or Linux Free Pascal is not necessarily a 'must have'. Any additional platforms that build are a bonus, so is entirely up to you or anyone else interested to put effort into getting that working. What is a 'must have', is continuous integration working on any suitable platform. Your choice to use Appveyor and the free Delphi version seems to the most appropriate as I suspect most users will be using Delphi's original platform - Windows.

wsfulton:
If you could, it would be rather good to use the Delphi C/C++ compiler instead of gcc (or in addition to gcc) for any C/C++ compilation (including compiling SWIG source). 

Regarding Delphi for Linux, you could approach Embarcadero for a Linux license, they may want to support efforts to fix the SWIG Delphi module on Linux as they are pushing their suite to be cross-platform??

lynxnake:
About Appveyor on the local PC - is it possible without paying Appveyor?

wsfulton:
@lynxnake, I don't think that setting up Appveyor on your own pc is a good idea, presumably you want this just to test a setup locally instead of using the real Appveyor boxes? If so, all that Appveyor does is the usual build instructions that you can run locally, so a `./autogen.sh && ./configure && make -s -j2 && make check-test-suite` etc. It does set up some paths etc and you need to install the same software locally. Alternatively, you can rdp into a box for interactive exploration if you need to, see https://www.appveyor.com/docs/how-to/rdp-to-build-worker/.

ojwb:
`fpc -Mdelphi` is supposed to be fairly compatible, though I don't know if that's also true of the C <-> Pascal interface.

wsfulton:
Hi @lynxnake. This seems to have gone quiet. We have a clearer instructions for adding a new target language into SWIG now, please see http://www.swig.org/Doc4.0/Extending.html#Extending_prerequisites. Getting it working on Travis (or Appveyor) is critical, so please let us know how that is progressing and then we can have a good look at merging it in. If you can get something runing on Linux it would be easier for me to help.

BTW, Appveyor and Travis are not doing much more than running './autogen.sh && ./configure && make check', so you don't get bogged down by these systems. If this works locally, I can probably help setup Appveyor/Travis.

ojwb:
Note we've switched from travis to github actions now, but otherwise wsfulton's comment should still be still valid.

@lynxnake Are you still interested in progressing this?  It looks like it was only really missing the CI setup, but it's approaching 5 years since your last response.

ojwb:
This still needs work (to merge it needs to fulfil the requirements list for a new "experimental" target language: https://www.swig.org/Doc4.0/Extending.html#Extending_experimental_status), but it's been over 5 years since the last comment from the submitter and nearly 5 months since I pinged them, so I'm going to close this.

If somebody is interested in pushing this forwards we can reopen.

chris9672:
There seems to be a fundamental misunderstanding about Delphi. Embarcadero never wanted to port Delphi to Linux.
The Delphi IDE is only available on Windows. Delphi uses cross-compilers to compile software for Windows, macOS, iOS, Android and Linux but the compilers can only be executed under Windows, but generate code for the corresponding platforms. Embarcadero also never wanted to port Delphi to Linux. Borland once released a Delphi version for Linux called Kylix, but it was never successful.

Issue #952, opened 2017-03-23
assignment of element of std::vector in python fails


ROTA-tor:
If I use a std::vector (with c++11 enabled). The swig file looks like this:
```
%include <std_shared_ptr.i>
%shared_ptr(std::vector<std::string>)
%template(StringVector) std::vector<std::string>;
```
I  wrap this to python. In python I cannot assign to the vector. The following example fails
```
>>> import test
>>> str = test.StringVector()
>>> str.append('foo')     
>>> str[0] = "bar"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/$USER/path/test.py", line 678, in __setitem__
    return _test.StringVector___setitem__(self, *args)
NotImplementedError: Wrong number or type of arguments for overloaded function 'StringVector___setitem__'.
  Possible C/C++ prototypes are:
    std::vector< std::string >::__setitem__(PySliceObject *,std::vector< std::string,std::allocator< std::string > > const &)
    std::vector< std::string >::__setitem__(PySliceObject *)
    std::vector< std::string >::__setitem__(std::vector< std::string >::difference_type,std::vector< std::string >::value_type const &)
```
This problem exists also for a vector of a self implemented class (e.g. `std::vector<MyClass>`)

I can fix this problem in the generated text.cxx. I have to apply the following change (please look for the comments):
```
SWIGINTERN PyObject *_wrap_StringVector___setitem__(PyObject *self, PyObject *args) {
  Py_ssize_t argc;
  PyObject *argv[4] = {
    0
  };
  Py_ssize_t ii;

  if (!PyTuple_Check(args)) SWIG_fail;
  argc = args ? PyObject_Length(args) : 0;
  for (ii = 0; (ii < 3) && (ii < argc); ii++) {
    argv[ii] = PyTuple_GET_ITEM(args,ii);
  }
  if (argc == 2) {
    int _v;
    int res = swig::asptr(argv[0], (std::vector< std::string,std::allocator< std::string > >**)(0));
    _v = SWIG_CheckState(res);
    if (_v) {
      {
        _v = PySlice_Check(argv[1]);
      }
      if (_v) {
        return _wrap_StringVector___setitem____SWIG_1(self, args);
      }
    }
  }
  if (argc == 3) {
    int _v;
    int res = swig::asptr(argv[0], (std::vector< std::string,std::allocator< std::string > >**)(0));
    _v = SWIG_CheckState(res);
    if (_v) {
      {
[...]

PR #986, opened 2017-05-04
[ruby] workaround for shared_ptr.i  and shared_ptr of std containers.

tamuratak:
Hi,
in #952, it is reported that std_shared_ptr.i  can not properly treat std shared pointers of std containers for python. This issue also exists for ruby. This PR fixes it for ruby.

The cause of this issue is that typemaps defined in std_shared_ptr.i are overridden by typemaps called in std/std_vector.i, etc through `%typemap_traits_ptr` which are always called after ones in std_shared_ptr.i. 

I have no idea how to correct the order. So this PR is a workaround. Although I doubt  this is ready for merged, I submit this PR to clarify this issue.

Regards.

### DIFF
--- a/Examples/test-suite/cpp11_shared_ptr_of_std_containers.i
+++ b/Examples/test-suite/cpp11_shared_ptr_of_std_containers.i
@@ -1,0 +1,13 @@
+%module cpp11_shared_ptr_of_std_containers
+
+%{
+#include <memory>
+%}
+
+%include <std_string.i>
+%include <std_vector.i>
+%include <std_shared_ptr.i>
+
+%shared_ptr(std::vector<std::string>);
+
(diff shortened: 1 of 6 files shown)


ojwb:
The new `cpp11_shared_ptr_of_std_containers` testcase here fails if run with SWIG from git master, so it seems this hasn't been addressed in the meantime.

It does sound like it would be better to fix the order of the typemaps as the submitter suggests.

We should really enable this new testcase for other languages too.

ojwb:
The new `cpp11_shared_ptr_of_std_containers` testcase here fails if run with SWIG from git master, so it seems this hasn't been addressed in the meantime.

It does sound like it would be better to fix the order of the typemaps as the submitter suggests.

We should really enable this new testcase for other languages too.

PR #991, opened 2017-05-19
[Python] Fix E731: do not assign a lambda expression

jschueller:
https://www.python.org/dev/peps/pep-0008/#programming-recommendations

Also more consistent with the few lines that follow where def is used too.


wsfulton:
The Travis failures are due to a new PEP failure introduced by the patch, for example:
```
callback.py:130:5: E704 multiple statements on one line (def)
```

jschueller:
One test failed due to a connection error, could it be restarted ?

Issue #994, opened 2017-05-30
[Scilab] Undefined symbol error when loading in Scilab 6


smarchetto:
Built the following example: 

```
%module classes

%inline %{

class Point {
public:
  int x, y;
  Point(int _x, int _y) : x(_x), y(_y) {}
  double distance(const Point& rhs) {
    return sqrt(pow(x-rhs.x, 2) + pow(y-rhs.y, 2));
  }
  void set(int _x, int _y) {
    x=_x;
    y=_y;
  }
};

%}
```

with the following commands:

```
$ swig -scilab -c++ classes.i
$ g++ -fPIC -c -I/usr/local/include/scilab  classes_wrap.cxx
$ g++ -shared classes_wrap.o -o libclasses.so 
````

Loading it in Scilab 6 provokes the following error:

```
--> exec loader.sce
--> addinter(fullfile(libclasses_path, 'libclasses' + getdynlibext()), 'libclasses', list_functions);
Link failed for dynamic library '/home/itsme/swigSamples/swig/classes/libclasses.so'.
An error occurred: /home/itsme/swigSamples/swig/classes/libclasses.so: undefined symbol: _Z17addCStackFunctionPKwPFiPcPvES0_
Link failed for dynamic library '/home/itsme/swigSamples/swig/classes/libclasses.so'.
An error occurred: /home/itsme/swigSamples/swig/classes/libclasses.so: undefined symbol: _Z17addCStackFunctionPKwPFiPcPvES0_
at line    27 of executed file /home/itsme/swigSamples/swig/classes/loader.sce
```

ojwb:
So can this be closed now?

PR #995, opened 2017-05-30
fix Scilab linking issue

davidcl:
Add the missing `extern "C"` when wrapping C++ codes


davidcl:
should fix issue #994

Issue #955, opened 2017-03-27
callback failed in Examples/go/callback/gocallback.go


NETkiddy:
The callback just failed as described in [issue#600](https://github.com/swig/swig/issues/600)
To fix this:
line 39 in gocallback.go:  

Change this:
```
func (p *goCallback) Run() {
	fmt.Println("GoCallback.Run")
}
```
To this:
```
func (p *overwrittenMethodsOnCallback) Run() {
	fmt.Println("GoCallback.Run")
}
```
Then the callback worked.

ojwb:
Please create a pull request for your proposed change - then it'll be automatically tested by the CI system.

Issue #600, opened 2016-01-28
Can't call to goCallback -> Run() in Example/go/callback


DavidLiuXh:
I try to Example/go/callback, but can't call to goCallback -> Run()

PR #1000, opened 2017-06-15
Fix Go callback example (swig/swig#600 and swig/swig#955)

fabrice102:
Fix Go callback example.
See swig/swig#600 and swig/swig#955.

### DIFF
--- a/Examples/go/callback/gocallback.go
+++ b/Examples/go/callback/gocallback.go
@@ -36,6 +36,6 @@ func DeleteGoCallback(p GoCallback) {
 	p.deleteCallback()
 }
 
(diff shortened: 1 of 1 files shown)


ojwb:
C# CI failure is presumably travis being flaky, but I've restarted that build.

ojwb:
And the output diff resulting from the patch is:

```
@@ -56,7 +56,7 @@
 
 Adding and calling a Go callback
 --------------------------------
-Callback::run()
+GoCallback.Run
 Callback::~Callback()
 
 Go exit
```

Which looks like an improvement, so merging.  Thanks.

PR #1011, opened 2017-06-23
Change to PHP7 Examples

nihal95:
A pre-requisite to converting "flat" function space to object oriented.

### DIFF
--- a/Examples/php/enum/runme.php
+++ b/Examples/php/enum/runme.php
@@ -11,22 +11,22 @@ print "    BLUE   =" . BLUE;
 print "    GREEN  =" . GREEN;
 
 print "\n*** Foo::speed ***";
-print "    Foo_IMPULSE   =" . Foo_IMPULSE;
-print "    Foo_WARP      =" . Foo_WARP;
-print "    Foo_LUDICROUS =" . Foo_LUDICROUS;
+print "    Foo::IMPULSE   =" . Foo::IMPULSE;
+print "    Foo::WARP      =" . Foo::WARP;
+print "    Foo::LUDICROUS =" . Foo::LUDICROUS;
 
 print "\nTesting use of enums with functions\n";
 
-enum_test(RED, Foo_IMPULSE);
-enum_test(BLUE, Foo_WARP);
-enum_test(GREEN, Foo_LUDICROUS);
+enum_test(RED, Foo::IMPULSE);
+enum_test(BLUE, Foo::WARP);
+enum_test(GREEN, Foo::LUDICROUS);
 enum_test(1234,5678);
 
 print "\nTesting use of enum with class method\n";
-$f = new_Foo();
+$f = new Foo();
 
-Foo_enum_test($f,Foo_IMPULSE);
-Foo_enum_test($f,Foo_WARP);
-Foo_enum_test($f,Foo_LUDICROUS);
+$f->enum_test(Foo::IMPULSE);
+$f->enum_test(Foo::WARP);
+$f->enum_test(Foo::LUDICROUS);
 
 ?>
--- a/Examples/php/value/runme.php
+++ b/Examples/php/value/runme.php
@@ -3,15 +3,15 @@
 	require "example.php";
 
 
-	$v = new_vector();
-        vector_x_set($v,1.0);
-        vector_y_set($v,2.0);
-        vector_z_set($v,3.0);
+	$v = new Vector();
+        $v->x = 1.0;
+        $v->y = 2.0;
+        $v->z = 3.0;
 
-	$w = new_vector();
-        vector_x_set($w,10.0);
-        vector_y_set($w,11.0);
-        vector_z_set($w,12.0);
+	$w = new Vector();
+        $w->x = 10.0;
+        $w->y = 11.0;
+        $w->z = 12.0;
 
 	echo "I just created the following vector\n";
 	vector_print($v);
@@ -25,7 +25,7 @@
 
 	echo "\nNow I'm going to add the vectors together\n";
 
-        $r = new_vector();
+        $r = new Vector();
 	vector_add($v, $w, $r);
 
 	vector_print($r);
--- a/Examples/php/enum/Makefile
+++ b/Examples/php/enum/Makefile
@@ -5,7 +5,7 @@ CXXSRCS    = example.cxx
 TARGET     = example
 INTERFACE  = example.i
 LIBS       =
-SWIGOPT    = -noproxy
+SWIGOPT    =
 
 check: build
 	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' php_run
--- a/Examples/php/value/Makefile
+++ b/Examples/php/value/Makefile
@@ -5,7 +5,7 @@ SRCS       = example.c
 TARGET     = example
 INTERFACE  = example.i
 LIBS       =
-SWIGOPT    = -noproxy
+SWIGOPT    =
 
 check: build
 	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' php_run


Examples/php/callback/runme.php
```diff
@@ -2,8 +2,6 @@
 
 # This file illustrates the cross language polymorphism using directors.
 
-require("example.php");
-
```
ojwb: For now at least, I think we need to keep generating (and requiring/including) the PHP wrapper script.  It will probably just be an empty script (or perhaps only include the magic to try to load the extension, but that's increasingly not useful as all the mechanisms it can use are disabled by default except in the CLI version of PHP).  But by keeping the script we can have the same code working with currently generated wrappers and with your new style wrappers.

Examples/php/class/runme.php
```diff
@@ -14,7 +12,7 @@
 
 # ----- Access a static member -----
 
-print "\nA total of " . Shape::nshapes() . " shapes were created\n";
+print "\nA total of " . Shape::nshapes_get() . " shapes were created\n";
```
ojwb: The `_get()` accessors are part of the "noproxy" flat wrapping, so this seems a backward change.

Examples/php/cpointer/runme.php
```diff
@@ -1,40 +1,38 @@
...
+	delete_intp($b);
+	delete_intp($c);
 
 	# Now try the typemap library
 	# This should be much easier. Now how it is no longer
 	# necessary to manufacture pointers.
 
 	print "Trying the typemap library\n";
-	$r = example::sub(37,42);
+	$r = sub(37,42);
```
ojwb: At the moment, the `delete_intp` is the flat wrapper layer and the `example::delete_intp` is the class wrapper around that.  If we were starting from scratch it probably would make more sense to wrap non-class function C/C++ function `intp_value` as a non-class PHP function, rather than placing it in this `example` class, especially now PHP has added namespaces, though I think we probably need to have an option to generate the `example` class for compatibility.  With current wrappers, I'm not sure if you can just globally replace `example::x` with `x`, or if there will be issues when objects are passed and/or returned, since they are handled as resources by this layer.

Examples/php/enum/runme.php
```diff
@@ -23,10 +21,10 @@
...
 print "\nTesting use of enum with class method\n";
-$f = new_Foo();
+$f = new Foo();
 
-Foo_enum_test($f,Foo_IMPULSE);
-Foo_enum_test($f,Foo_WARP);
-Foo_enum_test($f,Foo_LUDICROUS);
+$f->enum_test(Foo_IMPULSE);
+$f->enum_test(Foo_WARP);
+$f->enum_test(Foo_LUDICROUS);
```
ojwb: So currently this example explicitly builds with `-noproxy`, so this change will stop it working.

I think we probably should use the class wrappers instead for all the examples, so this change seems OK.  But it ought to also change to use `Foo::IMPULSE`, etc as well since those are the class wrapped constants.

Examples/php/value/runme.php
```diff
@@ -1,17 +1,15 @@
 <?php
 
-	require "example.php";
 
+	$v = new Vector();
+        $v->x_set(1.0);
+        $v->y_set(2.0);
+        $v->z_set(3.0);
```
ojwb: This example also currently builds with `-noproxy`, but aside from that I think this is good.

Examples/php/value/runme.php
```diff
@@ -25,7 +23,8 @@
 
 	echo "\nNow I'm going to add the vectors together\n";
 
-        $r = new_vector();
+    $r = new Vector();
+
```
ojwb: You've changed the indentation and added a blank line - please avoid whitespace changes mixed in with functional changes.  If the whitespace changes are needed, make them a separate commit.  Here they don't seem needed though.

Examples/php/variables/runme.php
```diff
@@ -26,10 +25,10 @@
 	echo "pt	= ".pt_get(), point_print(pt_get()) , "\n";
 
 	/* Try to set the values of some global variables */
-$a = "42.14";
+    $a = "42.14";
 
 	ivar_set($a);
-echo "a = $a\n";
+    echo "a = $a\n";
```
ojwb: These two lines could do with reindenting (though more than you've reindented them) but that shouldn't be mixed in with functional changes.

Examples/php/callback/runme.php
nihal95: Alright. I'll keep the require line and change accordingly.

Examples/php/class/runme.php
nihal95: In the PHP wrapper, the static nshapes() fn does both the work of calling nshapes_get() and nshapes_set(). So one less indirections here since we are nearly scraping what's generated at .php

Examples/php/cpointer/runme.php
nihal95: With the current wrappers `example::x` cannot be called with `x` . Since the functions were all non-class functions. I've wrapped such that, they can be accessed as `x` and that there's no need for class example. But I didn't get what you mentioned about `example` class for compatibility. And I'll do look at namespaces. I've got a reference here of namespaces, [HERE](https://stackoverflow.com/questions/24687502/can-i-put-php-extension-classes-functions-etc-in-a-namespace)

Examples/php/enum/runme.php
nihal95: Noted. Will work on it.

Examples/php/cpointer/runme.php
nihal95: This seemed to work. `example::x` to `x`

Examples/php/value/runme.php
```diff
@@ -3,15 +3,15 @@
...
-        vector_z_set($w,12.0);
+	$v = new Vector();
+        $v->__set("x",1.0);
+        $v->__set("y",2.0);
+        $v->__set("z",3.0);
+
+	$w = new Vector();
+        $w->__set("x",10.0);
+        $w->__set("y",11.0);
+        $w->__set("z",12.0);
```
nihal95: There isn't a `$v->x_set()` in the current wrappers. So using the magic methods instead.

Examples/php/cpointer/runme.php
ojwb: It may work here, but `example::sub()` is returning an integer.  As I said above, the problems will be with objects, as in that case the flat functions (i.e. `x`) return a PHP resource wrapping the C++ object pointer while the class functions defined in the PHP wrapper (i.e. `example::x`) wrap a PHP object around the resource and return that instead.

> I didn't get what you mentioned about example class for compatibility.

`example::sub()` is the documented way this gets wrapped unless `-noproxy` is used.  If we stop wrapping things like that, then SWIG will suddenly generate different wrappers and all existing code relying on things like `example::sub()` working will stop working.  And that would annoy both users of SWIG and end users of wrappers generated with SWIG.

For example, users of Xapian's SWIG-generated PHP bindings are told to write `Xapian::DB_CREATE_OR_OPEN`, and a newer SWIG version which suddenly stopped generating a wrapper like that would be a problem.  Perhaps `DB_CREATE_OR_OPEN` can be safely used instead, but users of Xapian's PHP bindings will rightly be annoyed that they have to go and rewrite their scripts.  So with the new style of wrapping you're working on, I think we need to provide a way to still have this wrapping generated.

It seems I need to stress this again: we need to generate wrappers that are compatible with what's currently generated to the greatest extent possible.  If we don't, then it's not going to be reasonable to merge this work to git master, and it'll just sit on a branch.  There are already too many SWIG GSoC projects from previous years bitrotting on branches with decreasing hope of ever being merged as master naturally evolves and incompatibilities with the branch increase.  If we're going to approach project this with a disregard for compatibility, this will just be another dead project at the end of the summer and we might as well not bother spending the time working on it.

Examples/php/value/runme.php
ojwb: This should be `$v->x = 1.0;` etc.

Examples/php/value/runme.php
```diff
@@ -3,15 +3,15 @@
...
 
 	echo "I just created the following vector\n";
 	vector_print($v);
```
ojwb: This should be `example::vector_print($v);`

ojwb:
The first commit is OK but the second needs more work, so I've cherry-picked the first to master.

Feel free to `git rebase master` (which should automatically spot the first commit can just be dropped from the branch) then `git push --force` to update the PR (in general, don't force-push branches which other people may be reviewing or working with without agreement).

PR #1014, opened 2017-06-25
Object Structure Approach - PHP7

nihal95:
Moving from "flat" function approach to object approach using Zend C API.

### DIFF
--- a/Source/Modules/php.cxx
+++ b/Source/Modules/php.cxx
@@ -858,6 +858,8 @@ public:
     int overloaded = 0;
     String *overname = 0;
     String *modes = NULL;
+    bool static_setter = false;
+    bool static_getter = false;
 
     if (constructor) {
       modes = NewString("ZEND_ACC_PUBLIC | ZEND_ACC_CTOR");
@@ -875,9 +877,20 @@ public:
       wname = (String*) ptr;
     }
     else if (wrapperType == staticmembervar) {
-      char *ptr = Char(iname);
-      ptr+= strlen(Char(iname)) - 4 - strlen(strrchr(GetChar(n, "name"),':') + 1);
+      // Shape::nshapes -> nshapes
+      char *ptr = Char(strrchr(GetChar(n, "name"),':'));
+      ptr+= 1;
       wname = (String*) ptr;
+
+      /* We get called twice for getter and setter methods. But to maintain
+         compatibility, Shape::nshapes() is being used for both setter and 
+         getter methods. So using static_setter and static_getter variables
+         to generate half of the code each time.
+       */
+      if (Cmp(strrchr(GetChar(n, "sym:name"),'_'),"_set") == 0)
+        static_setter = true;
+      else
+        static_getter = true;
     }
     else if (wrapperType == staticmemberfn) {
       char *ptr = Char(iname);
@@ -911,17 +924,19 @@ public:
     String *outarg = NewStringEmpty();
     String *cleanup = NewStringEmpty();
 
-    if (Cmp(class_name,NULL) != 0)
-      Printv(f->def, "PHP_METHOD(", class_name, ",", wname,") {\n", NIL);
-    else
-      Printv(f->def, "PHP_FUNCTION(", wname,") {\n", NIL);
+    if (!static_getter) {
+      if (Cmp(class_name,NULL) != 0)
+        Printv(f->def, "PHP_METHOD(", class_name, ",", wname,") {\n", NIL);
+      else
+        Printv(f->def, "PHP_FUNCTION(", wname,") {\n", NIL);
+    }
 
     emit_parameter_variables(l, f);
     /* Attach standard typemaps */
 
     emit_attach_parmmaps(l, f);
     // Not issued for overloaded functions.
-    if (!overloaded) {
+    if (!overloaded && !static_getter) {
       create_command(class_name, wname, n, modes);
     }
 
@@ -965,6 +980,12 @@ public:
       Printf(f->code, "if(arg_count<%d || arg_count>%d ||\n", num_required, num_arguments);
       Printf(f->code, "   zend_get_parameters_array_ex(arg_count,args)!=SUCCESS)\n");
       Printf(f->code, "\tWRONG_PARAM_COUNT;\n\n");
+    } else if (static_setter || static_getter) {
+      if (num_arguments == 0) {
+        Printf(f->code, "if(ZEND_NUM_ARGS() == 0) {\n");
+      } else {
+        Printf(f->code, "if(ZEND_NUM_ARGS() == %d && zend_get_parameters_array_ex(%d, args) == SUCCESS) {\n", num_arguments, num_arguments);
+      }
     } else {
       if (num_arguments == 0) {
 	Printf(f->code, "if(ZEND_NUM_ARGS() != 0) {\n");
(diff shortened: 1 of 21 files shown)


Examples/php/class/runme.php
```diff
@@ -14,7 +14,7 @@
 
 # ----- Access a static member -----
 
-print "\nA total of " . Shape::nshapes() . " shapes were created\n";
+print "\nA total of " . Shape::nshapes_get() . " shapes were created\n";
```
ojwb: Ideally the new-style wrapping should provide the same interface in PHP, and we can avoid such changes to the examples, etc.

Examples/php/cpointer/runme.php
```diff
@@ -6,35 +6,35 @@
...
-	$a = example::new_intp();
-	$b = example::new_intp();
-	$c = example::new_intp();
-	example::intp_assign($a,37);
-	example::intp_assign($b,42);
+	$a = new_intp();
+	$b = new_intp();
+	$c = new_intp();
+	intp_assign($a,37);
+	intp_assign($b,42);
```
ojwb: Similarly here - as we discussed on IRC, let's keep the `example` wrapper class in the generated `.php` wrapper to handle this case.

Examples/php/enum/Makefile
```diff
@@ -5,7 +5,7 @@ CXXSRCS    = example.cxx
 TARGET     = example
 INTERFACE  = example.i
 LIBS       =
-SWIGOPT    = -noproxy
+SWIGOPT    =
```
ojwb: If you rebase your branch onto current master, these `-noproxy` changes and the `Foo_IMPULSE` ones below should disappear automatically.  It's OK to do that and `git push --force` to update this PR.

Source/Modules/php.cxx
```diff
@@ -133,6 +137,18 @@ extern "C" {
...
+ * Used to mainly distingush between standard datatype pointers and typedef pointers.
+ */
+static bool SWIG_is_valid_class(String *Type_class) {
+  Iterator c_iterator;
+  for (c_iterator = First(classes); c_iterator.item; c_iterator = Next(c_iterator)) {
+    if (Cmp(c_iterator.item,Type_class) == 0)
+      return true;
+  }
+  return false;
+}
```
ojwb: I'm not really following the code where this is used, but I'm pretty certain this isn't the right approach to any problem you're needing to solve.  SWIG should be able to tell you if a type is a class or not - probably you want `SwigType_isclass(const SwigType *t)`.

Source/Modules/php.cxx
```diff
@@ -441,6 +457,9 @@ class PHP : public Language {
     Printf(s_header, "}\n");
     Printf(s_header, "#endif\n\n");
 
+    Printf(s_header,"#ifdef __GNUG__\n#define SWIG_remove(zv) delete zv\n");
+    Printf(s_header,"#else\n#define SWIG_remove(zv) free(zv)\n#endif\n\n");
```
ojwb: I'm guessing you want to test `#ifdef __cplusplus` here - making it conditional on a particular compiler doesn't seem right.

Source/Modules/php.cxx
```diff
@@ -646,16 +665,19 @@ class PHP : public Language {
...
     if (directorsEnabled()) {
       Dump(f_directors, f_begin);
     }
     Printv(f_begin, s_vdecl, s_wrappers, NIL);
-    Printv(f_begin, all_cs_entry, "\n\n", s_arginfo, "\n\n", s_entry,
+    Printv(f_begin, s_arginfo, "\n\n", all_cs_entry, "\n\n", s_entry,
 	" SWIG_ZEND_NAMED_FE(swig_", module, "_alter_newobject,_wrap_swig_", module, "_alter_newobject,NULL)\n"
 	" SWIG_ZEND_NAMED_FE(swig_", module, "_get_newobject,_wrap_swig_", module, "_get_newobject,NULL)\n"
-	" ZEND_FE_END\n};\n\n", NIL);
+	" { NULL, NULL, NULL }\n};\n\n", NIL);
```
ojwb: `ZEND_FE_END` is (a) more self-explanatory (b) more correct (there are more than 3 entries in the `struct`) and (c) more future proof should the entries in the struct change in the future.

nihal95:
This pr is not yet up to date. I have few more commits on local which takes care of the compatibility. And I'm currently working on magic methods. Will update it after its done.

Source/Modules/php.cxx
nihal95: This is because, the pointer in struct needs to be freed. And free is used in C and delete in C++. delete calls the destructor.

Source/Modules/php.cxx
nihal95: I'll use ZEND_FE_END. The thing was ZEND_FE_END in module functions, that is non-class. And ZEND_ME_END in class functions I think. I had got some errors using FE_END in class functions. So kept it on NULL. Will change it

nihal95:
I've updated the pr. Take a look at it.

Source/Modules/php.cxx
ojwb: The correct test for "is this a C++ compiler?" is `#ifdef __cplusplus`.

You are instead testing: "is this GNU C++?"

Source/Modules/php.cxx
```diff
@@ -441,6 +475,15 @@ class PHP : public Language {
...
     Printf(s_header, "#endif\n\n");
 
+    Printf(s_header,"#ifdef __GNUG__\n#define SWIG_remove(zv) delete zv\n");
+    Printf(s_header,"#else\n#define SWIG_remove(zv) free(zv)\n#endif\n\n");
+
+    Printf(s_header, "#define CALL_METHOD(name, retval, thisptr)                  \
+                      call_user_function(EG(function_table),thisptr,&name,retval,0,NULL TSRMLS_CC);\n\n");
+
+    Printf(s_header, "#define CALL_METHOD_PARAM_1(name, retval, thisptr, param)                  \
+                      call_user_function(EG(function_table),thisptr,&name,retval,1,&param TSRMLS_CC);\n\n");
```
ojwb: `TSRMLS_CC` isn't needed with PHP7.  None of the `TSRM` stuff is.  They're still present, but defined to "no op" values so that existing code using them still builds.  No point using them in new code.

Source/Modules/php.cxx
```diff
@@ -714,7 +763,10 @@ class PHP : public Language {
 
     String * s = cs_entry;
     if (!s) s = s_entry;
-    Printf(s, " SWIG_ZEND_NAMED_FE(%(lower)s,%s,swig_arginfo_%s)\n", cname, iname, arginfo_code);
+    if (Cmp(cname,NULL) != 0)
```
ojwb: `Cmp()` against `NULL` here is a strange test (also used just above).  To test if `cname` is `NULL`, you want `if (!cname)` or `if (cname != NULL)` or similar.

Source/Modules/php.cxx
```diff
@@ -133,6 +140,33 @@ extern "C" {
...
+              Cmp(type,"double") == 0 ||
+               Cmp(type,"long double") == 0)
+      return_value = NewString("double");
+    else if (Cmp(type,"bool") == 0)
+      return_value = NewString("bool");
+    else if (Cmp(type,"char") == 0 ||
+              Cmp(type,"String") == 0)
+      return_value = NewString("string");
+  return return_value;
+}
```
ojwb: There shouldn't be a hard-coded mapping like this.  Constants are handled via typemaps (see `Lib/php/const.i`).

Source/Modules/php.cxx
nihal95: Ah Yes yes. sure. sorry

Source/Modules/php.cxx
nihal95: Hey Olly. I'm not exactly handling the constants here. This is used to declare properties in minit section. I just wanted the conversion of standard datatypes available in C++ to respective datatypes supported in zval. This is used in zend_declare properties. All integers types are covered by long, decimal types by double, and bool, string in the zval. So I had to hard-code this rundown conversion.

nihal95: I have looked at `Lib/php/const.i` but that's all about declaring constants.

Source/Modules/php.cxx
```diff
@@ -1050,11 +1429,13 @@ class PHP : public Language {
     DelWrapper(f);
     f = NULL;
 
+    if (wrapperType == membervar && Strstr(wname,"_get") != NULL)
```
ojwb: This checks for `_get` anywhere in the name, not just at the end.  Better to re-use the existing check (search `php.cxx` for `"_get"`).

ojwb:
I've now created branch gsoc2017-php7-classes-via-c-api and merged your branch to it.

There are some issues on the branch though - I think perhaps it's best for us to sit down at some point soon and work through them.

Source/Modules/php.cxx
ojwb: This still needs changing.

nihal95: I had changed it. Will make a commit if not. Thanks

PR #1015, opened 2017-06-27
Fix OUTPUT Typemap not having return statement bug in PHP wrapper.

nihal95:
The PHP wrapper for OUTPUT Typemap was missing return statement

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -8,6 +8,10 @@ Version 4.0.0 (in progress)
 ===========================
 
 2017-06-27: nihaln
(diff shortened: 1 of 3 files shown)


ojwb:
Looks good, but the same fix should probably be made for PHP5 too as it looks to have the exact same issue:

```
$ make -s -C Examples/php5/cpointer
Testing the pointer library
	a = Resource id #5
	b = Resource id #6
	c = Resource id #7
	37 + 42 = 79
Trying the typemap library
	37 - 42 = 
```

nihal95:
Done

ojwb:
Thanks.

PR #1017, opened 2017-07-02
fix: memory leak, clang memset warning and undefined behavior

mromberg:
The clang++ compiler warns that the memset() involved with this code points at an object with a vtable pointer that may be overwritten.   One can silence this warning with a cast.   But I think the warning is a sign of deeper problems.

1.  For starters the malloced memory is leaked.   It is not a huge leak as it happens once.   But it is leaked (unless something really strange is going on to free it).
2. The function in question returns Type v_def by **value**.  So, copy/move constructors and operators as well as the destructor for Type may also be called.   Odds are very good that not every possible class that Type can be will do anything reasonable when occupying a block of memory that is all zeros.   For example, the destructor could examine a boolean member and if it is false start doing some action.

   This patch fixes (I think) all of these problems.   The memory leak is fixed by creating a static Type object to return.   Since the objects constructor was called all the other member functions it has are safe.

  The only thing I'm unsure of is why the code in master had a comment about not needing the default constructor.   There may be some reason for this comment.   But you can't really do this memset "trick" unless Type is a POD.

Mike

### DIFF
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -434,8 +434,8 @@ namespace swig
     {
       swig::SwigVar_PyObject item = PySequence_GetItem(_seq, _index);
       try {
-	return swig::as<T>(item, true);
-      } catch (std::exception& e) {
+	return swig::as<T>(item);
+      } catch (const std::invalid_argument& e) {
 	char msg[1024];
 	sprintf(msg, "in sequence element %d ", (int)_index);
 	if (!PyErr_Occurred()) {
--- a/Lib/python/pystdcommon.swg
+++ b/Lib/python/pystdcommon.swg
@@ -107,14 +107,14 @@ namespace swig {
 
   template <class Type> 
   struct traits_as<Type, value_category> {
-    static Type as(PyObject *obj, bool throw_error) {
+    static Type as(PyObject *obj) {
       Type v;
       int res = asval(obj, &v);
       if (!obj || !SWIG_IsOK(res)) {
 	if (!PyErr_Occurred()) {
 	  ::%type_error(swig::type_name<Type>());
 	}
-	if (throw_error) throw std::invalid_argument("bad type");
+	throw std::invalid_argument("bad type");
       }
       return v;
     }
@@ -122,7 +122,7 @@ namespace swig {
 
   template <class Type> 
   struct traits_as<Type, pointer_category> {
-    static Type as(PyObject *obj, bool /* throw_error */) {
+    static Type as(PyObject *obj) {
       Type *v = 0;      
       int res = (obj ? traits_asptr<Type>::asptr(obj, &v) : SWIG_ERROR);
       if (SWIG_IsOK(res) && v) {
@@ -144,7 +144,7 @@ namespace swig {
 
   template <class Type> 
   struct traits_as<Type*, pointer_category> {
-    static Type* as(PyObject *obj, bool throw_error) {
+    static Type* as(PyObject *obj) {
       Type *v = 0;      
       int res = (obj ? traits_asptr<Type>::asptr(obj, &v) : SWIG_ERROR);
       if (SWIG_IsOK(res)) {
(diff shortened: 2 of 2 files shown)


mromberg:
So, some tests blew up.  For example this:
```
class Foo
  {
    int x;
    Foo(int i) : x(i) {
    }
  };
```

  You can put into a std container.   And it has no default constructor.   This slight variant of Foo is what I'm attempting to fix:

```
struct Foo {
    int x;
    int flag;
   Foo(int xv) { x = xv; flag = 1; }
   ~Foo() { if (!flag) startWorldWarIII(); }
};
```

   The code swig is now creating will set everything to zero, kill any vtable pointers and ~Foo() will start world war three.

 I'm assuming swig needs to generate C++ code that is only C++98 or something?   If it could use C++11 or 14 then there are some ways to tell if a template argument is a POD or not.   I'll think about it some more.   It would be nice to not have to rewrite all those templates in the swig std module.

ojwb:
I think the comment is really saying "don't assume the type is default constructable".

The generated code should ideally be C++98 unless the input is C++11 (or later).

This was reported before in #659.  I dug into the template maze at the time and as far as I could make out this branch is never used by anything in SWIG, and it's not documented anywhere so it seems unlikely it's used from user typemaps.  Looks like the reporter closed that issue and it fell off my radar, but if we can confirm this code path is never used, we should remove it.

mromberg:
I've tried to follow the usage in there as well.   The thing is that the undefined behavior is occurring in a branch of the function which is dealing with an error.   If the boolean is set, it will throw an exception and there is no problem (since no bogus object is created).

  I followed the object I wrapped back and it seems that the "as()" method was called with the boolean set and exceptions caught.   I thought about just removing the boolean and having this code always throw.   I think this is the only right thing to do.   

  But it is difficult to try and find all the locations where this method may be used.   And it is placed into wrapper modules.  So, in theory some user code could be using it.

  If the undefined behavior (returning a memset "constructed" object) were replaced with always throwing an exception then we could see if any existing tests fail.  They probably would due to an uncaught exception.   If all existing swig tests pass would this be "good enough"?

  Tracking down where it is used is rough (like you know) because it is an awesome mix of overloaded functions and templates all with the name "as".

  Of course clang could be silenced with a simple cast.  But if that code were ever to run it would not be a good thing.   I think it needs to go.

ojwb:
I think it's best to kill it.  I doubt users are using this in their typemaps as it isn't documented anywhere.  But even if they are, invoking undefined behaviour like this doesn't seem helpful to anyone.

We could keep the `bool` parameter and just ignore it, but I think it's better to give a compile-time error so that the calling code can be checked to see what's actually wanted.  It isn't hard to stop passing the extra parameter, and that will work with existing releases of SWIG too.

mromberg:
I removed the boolean from both struct traits_as<Type*,...>::as(), struct traits_as<Type,...>::as() and the more generic Type as().   The generic as().   They all now always throw.

  The generic as() took the boolean as a default argument (false).   To check the spots the test suite used this I made it a std::string (for my own testing).   The only spot as() was used was in operator T () const [pycontainer.swg].   This bit of code was using a try catch block.

  So, these changes should cover everything the existing swig tests cover.   There very well may be some template code using an as with a boolean or a use of as() which does not expect a throw.  If this code exists, I could not find it.   More importantly there is no swig test case that covers it.

ojwb:
Thanks.

wsfulton:
Please, let's replicate the fix across all languages...
```
~/swig/github/swig/Lib $ grep -r -i "uninitialized return value" .
./r/rstdcommon.swg:	// Uninitialized return value, no Type() constructor required.
./octave/octstdcommon.swg:	// Uninitialized return value, no Type() constructor required.
./ruby/rubystdcommon.swg:	// Uninitialized return value, no Type() constructor required.
./python/pystdcommon.swg:	// Uninitialized return value, no Type() constructor required.
./scilab/scistdcommon.swg:	// Uninitialized return value, no Type() constructor required.
```

wsfulton:
I see ruby is addressed in #1018, but that still leaves R, octave and scilab to be addressed.

mromberg:
If you would like, I could take a stab at the other languages.   I just did python and ruby because those are the only ones I have any experience with.  And I kind think at least one of those other ones has current tests that are failing.   But maybe that won't matter.  The fix should be pretty much the same.

  Which does bring up an interesting subject.   The fix is the same because most of these modules were just cut/paste jobs from whatever language got the stl support first.   It would be nice to move any common STL support into one place rather than replicating it across each scripting language.  This sort of thing would probably best be done in a X.Y type branch rather than a X.Y.Z branch.

wsfulton:
@mromberg, thanks for doing this in #1027

PR #1018, opened 2017-07-03
as() memset() fix for ruby

mromberg:
Since the templates are fresh in my brain, here is a similar fix for ruby as the one for python.

### DIFF
--- a/Lib/ruby/rubycontainer.swg
+++ b/Lib/ruby/rubycontainer.swg
@@ -183,8 +183,8 @@ namespace swig
     {
       VALUE item = rb_ary_entry(_seq, _index );
       try {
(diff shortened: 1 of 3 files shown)


mromberg:
Anyone know the majick incantation to make the cpp11 test cases run?   I'm using gcc version 6 and they don't by default.   Looking at the travis thing I see a CPP14=1.   Setting that does nothing for me.   What decides to run these or not?

wsfulton:
```
./configure --enable-cpp11-testing
```
will turn on the c++11 testcases. Change the compiler and flags used to configurewith CXXFLAGS=-std=c++14 and CXX=g++14, see configure --help.

ojwb:
Thanks.

Issue #948, opened 2017-03-18
Using C++11 raw string literals in docstrings leads to delimiters in output


baldurk:
If I modify the `Examples/python/docstrings/example.i` like so:

```cpp
/* %feature("docstring") has to come before the declaration of the method to
 * SWIG. */
%feature("docstring") Foo::bar R"FOO(No comment
as a raw string
on multiple lines)FOO"
```

Then the output looks something like:

```cpp
static PyMethodDef SwigMethods[] = {
	 { (char *)"SWIG_PyInstanceMethod_New", (PyCFunction)SWIG_PyInstanceMethod_New, METH_O, NULL},
	 { (char *)"Foo_bar", _wrap_Foo_bar, METH_VARARGS, (char *)"\n"
		"FOO(No comment\n"
		"as a raw string\n"
		"on multiple lines)FOO\"\n"
		""},
```

You can see that part of the delimiter on either side (including the trailing `"`) is present in the output. Even if there's no custom `FOO` then it still includes the `()`s and the trailing `"`.

I went spelunking and I found a bit of code that seems like it's weird. In `scanner.c` when it returns a plain string, it trims off the ending `"`:

```cpp
      if (!str_delimiter) { /* Ordinary string: "value" */
	if (c == '\"') {
	  Delitem(s->text, DOH_END);
	  return SWIG_TOKEN_STRING;
```

but when it returns a raw string literal, it doesn't trim anything:

```cpp
	  if (Strcmp( str_delimiter, end_delimiter )==0) {
	    Delete( end_delimiter ); /* Correct end delimiter )XXXX" occured */
	    Delete( str_delimiter );
	    str_delimiter = 0;
	    return SWIG_TOKEN_STRING;
```

I tried adding some simple calls to remove the appropriate amount from start and end, and it seems to work in my simple test case. The `cpp11_raw_string_literals` test still passes (I think, I had trouble running the tests properly so I ended up running it manually).

So hopefully this fix is along the right lines but maybe it needs to be tidied up or done in a better place, as I have _absolutely no idea_ what I'm doing in the swig codebase :smile:. The patch is here:

```patch
diff --git a/Source/Swig/scanner.c b/Source/Swig/scanner.c
index f62ddda..4901ff1 100644
--- a/Source/Swig/scanner.c
+++ b/Source/Swig/scanner.c
@@ -749,6 +749,9 @@ static int look(Scanner *s) {
          }

          if (Strcmp( str_delimiter, end_delimiter )==0) {
[...]

PR #1019, opened 2017-07-06
Experimental fix for delimiter leakage in raw strings

baldurk:
As discussed in more detail in #948, this is a fix I've been using locally to avoid raw strings accidentally including the delimiter.


wsfulton:
Thanks, this patch looks good to me. I'm adding a bunch of regression tests and will merge to master shortly.

Also fixes #1273.

PR #1027, opened 2017-07-20
stdcommon as() memset() removal for octave, R and scilab

mromberg:
The octave and R test-suite passes on my machine.   The scilab c++11 tests failed before my patches and continue to fail with them.

  In the scilab case, it looks like the failing tests are related to strongly typed enums (and possibly other things).   The container related tests do seem to work.


ojwb:
Thanks.

(I'd agree that all this cut and pasted code isn't great.)

PR #1028, opened 2017-07-23
[WIP] Kotlin/Native support

thestr4ng3r:
Work in progress support for Kotlin/Native: https://github.com/JetBrains/kotlin-native
Kotlin/Native contains a tool to generate bindings for simple C functions, but no C++, which is why using SWIG would make sense here.

Feedback is appreciated!


thestr4ng3r:
Update: class and simple examples are already working.

wsfulton:
Hi. I think I may be missing something, but I don't see the point of adding a Kotlin language module when Kotlin is 100% compatible with Java and SWIG already has a fully fledged mature Java language module. Surely you just wrap your C/C++ code with the Java module and called the Java wrappers from Kotlin?

thestr4ng3r:
Kotlin/Native is actually completely independent of Kotlin for JVM. They share the same language and compiler frontend, but Kotlin/Native is then compiled natively using LLVM and won't normally use anything Java-related.Interoperability with C functions is already provided for it, so I think it would be logical to use swig to wrap C++.

wsfulton:
I see. So you are basically saying that Kotlin/Native does not support JNI.

eskatos:
Kotlin Native compiles to native objects (libs & executables) for linux, windows, mac, ios etc.. it doesn't involve any JVM. So yes, JNI isn't in the picture.

@thestr4ng3r do you plan to put more effort into this?

thestr4ng3r:
Not really, I don't have much time for it at the moment. But feel free to continue with it if you want.

ojwb:
Given this is still WIP, the submitter says above they don't plan to work on it further, and nobody has stepped up to take over in 2.5 years I'm going to close this.

If somebody is interested in getting it to the state of being merged we can reopen.  You need to be aiming to fulfil the requirements list for a new "experimental" target language: https://www.swig.org/Doc4.0/Extending.html#Extending_experimental_status

PR #1034, opened 2017-07-26
Add directorin typemap for Python shared_ptr

### LINKED ISSUE #1029: Python: Incorrect handling of null shared_ptr in director, should be converted to None
I found that when I override a C++ class using a director, if an overridden virtual function is called that takes a shared_ptr, it incorrectly gets an invalid SWIG proxy object instead of None.  Unfortunately this means it's impossible to test for null arguments, and the code crashes if it tries to do anything. Example,

test.i:
```c++
%module(directors="1") test
%include std_shared_ptr.i
%shared_ptr(IntValue)

%{
#include <memory>
#include <iostream>

class IntValue { public: int value; };

class Test
{
public:
  Test() {}
  virtual ~Test() {}

  virtual void test(std::shared_ptr<IntValue> i) {
    if (i)
      std::cout << "Default impl " << i->value << std::endl;
    else
      std::cout << "Default impl i==None" << std::endl;
  }

  virtual void testnull() {
    std::cout << "Default testnull" << std::endl;
    test(std::shared_ptr<IntValue>());
  }
};

%}

%feature("director") Test;

class IntValue { public: int value; };

class Test
{
public:
  Test();
  virtual ~Test();
  virtual void test(std::shared_ptr<IntValue> i);
  virtual void testnull();
};
```

testtest.py:
```python

import test

class OverTest(test.Test):
    def test(self, i):
        print('Override Test, i =', i)
        if i:
            print('i is truthy')
        else:
            print('i is falsy')
        print('value:', i.value)

j = test.IntValue()
j.value = 10

t = OverTest()
t.test(j)
t.testnull()
```

buildrun.sh:
```shell
#!/bin/bash
swig -c++ -python -py3 test.i
g++ -c -o test_wrap.o -fPIC $(pkg-config --cflags python3) test_wrap.cxx
g++ -o _test.so -shared -fPIC test_wrap.o $(pkg-config --libs python3)
python3 testtest.py
```

output:
```
$ ./build.sh 
Override Test, i = <test.IntValue; proxy of <Swig Object of type 'std::shared_ptr< IntValue > *' at 0x7fdc990cef00> >
i is truthy
value: 10
Default testnull
Override Test, i = <test.IntValue; proxy of <Swig Object of type 'std::shared_ptr< IntValue > *' at 0x7fdc97a96f60> >
i is truthy
./build.sh: line 5: 13440 Segmentation fault      python3 testtest.py
```

For the call to `t.testnull()`, the output should have said `i is falsy`, and ideally, `i==None`, but instead it points as an invalid object.

radarsat1:
Fixed #1029 for me, comments welcome.  Not sure if it's appropriate to put directorin typemaps in this file, nor am I sure how to do the equivalent directorout or if it is necessary, so I would consider it a separate issue.

### DIFF
--- a/Examples/test-suite/li_boost_shared_ptr_director.i
+++ b/Examples/test-suite/li_boost_shared_ptr_director.i
@@ -21,6 +21,9 @@ struct Base {
   virtual boost::shared_ptr<C> ret_c_shared_ptr() = 0;
   virtual C ret_c_by_value() = 0;
   virtual int take_c_by_value(C c) = 0;
+  virtual int take_c_by_ref(C& c) = 0;
+  virtual int take_c_by_pointer(C* c) = 0;
+  virtual int take_c_by_pointer_ref(C*& c) = 0;
   virtual int take_c_shared_ptr_by_value(boost::shared_ptr<C> c) = 0;
   virtual int take_c_shared_ptr_by_ref(boost::shared_ptr<C>& c) = 0;
   virtual int take_c_shared_ptr_by_pointer(boost::shared_ptr<C>* c) = 0;
@@ -47,8 +50,34 @@ int call_take_c_by_value(Base* b) {
   return b->take_c_by_value(c);
 }
 
+int call_take_c_by_ref(Base* b) {
+  C c(6);
+  return b->take_c_by_ref(c);
+}
+
+int call_take_c_by_pointer(Base* b) {
+  C c(7);
+  return b->take_c_by_pointer(&c);
+}
+
+int call_take_c_by_pointer_ref(Base* b) {
+  C c(8);
+  C* p(&c);
+  return b->take_c_by_pointer_ref(p);
+}
+
+int call_take_c_by_pointer_with_null(Base* b) {
+  C* p = NULL;
+  return b->take_c_by_pointer(p);
+}
+
+int call_take_c_by_pointer_ref_with_null(Base* b) {
+  C* p = NULL;
+  return b->take_c_by_pointer_ref(p);
+}
+
 int call_take_c_shared_ptr_by_value(Base* b) {
-  boost::shared_ptr<C> ptr(new C(6));
+  boost::shared_ptr<C> ptr(new C(9));
   return b->take_c_shared_ptr_by_value(ptr);
 }
 
@@ -58,7 +87,7 @@ int call_take_c_shared_ptr_by_value_with_null(Base* b) {
 }
 
 int call_take_c_shared_ptr_by_ref(Base* b) {
-  boost::shared_ptr<C> ptr(new C(7));
+  boost::shared_ptr<C> ptr(new C(10));
   return b->take_c_shared_ptr_by_ref(ptr);
 }
 
@@ -68,7 +97,7 @@ int call_take_c_shared_ptr_by_ref_with_null(Base* b) {
 }
 
 int call_take_c_shared_ptr_by_pointer(Base* b) {
-  boost::shared_ptr<C> ptr(new C(8));
+  boost::shared_ptr<C> ptr(new C(11));
   return b->take_c_shared_ptr_by_pointer(&ptr);
 }
 
@@ -78,7 +107,7 @@ int call_take_c_shared_ptr_by_pointer_with_null(Base* b) {
 }
 
 int call_take_c_shared_ptr_by_pointer_ref(Base* b) {
-  boost::shared_ptr<C> *ptr = new boost::shared_ptr<C>(new C(9));
+  boost::shared_ptr<C> *ptr = new boost::shared_ptr<C>(new C(12));
   return b->take_c_shared_ptr_by_pointer_ref(ptr);
 }
 
--- a/Examples/test-suite/python/li_boost_shared_ptr_director_runme.py
+++ b/Examples/test-suite/python/li_boost_shared_ptr_director_runme.py
@@ -1,0 +1,80 @@
+from li_boost_shared_ptr_director import *
+
+class Derived(Base):
+    def __init__(self, flag):
+        self.return_none = flag
+        Base.__init__(self)
+
+    def ret_c_shared_ptr(self):
+        if self.return_none:
+            return None
+        else:
+            return C()
+
+    def ret_c_by_value(self):
+        return C()
+
+    def take_c_by_value(self,c):
+        return c.get_m()
+
+    def take_c_by_ref(self,c):
+        return c.get_m()
+
+    def take_c_by_pointer(self,c):
+        if c:
+            return c.get_m()
+        else:
+            return -2
+
+    def take_c_by_pointer_ref(self,c):
+        if c:
+            return c.get_m()
+        else:
+            return -3
+
+    def take_c_shared_ptr_by_value(self,c):
+        if c:
+            return c.get_m()
+        else:
+            return -4
+
+    def take_c_shared_ptr_by_ref(self,c):
+        if c:
+            return c.get_m()
+        else:
+            return -5
+
+    def take_c_shared_ptr_by_pointer(self,c):
+        if c:
+            return c.get_m()
+        else:
+            return -6
+
+    def take_c_shared_ptr_by_pointer_ref(self,c):
+        if c:
+            return c.get_m()
+        else:
+            return -7
+
+a = Derived(False)
+b = Derived(True)
+
+assert call_ret_c_shared_ptr(a) ==  1
+assert call_ret_c_shared_ptr(b) == -1
+assert call_ret_c_by_value(a)   ==  1
+
+assert call_take_c_by_value(a)                  == 5
+assert call_take_c_by_ref(a)                    == 6
+assert call_take_c_by_pointer(a)                == 7
+assert call_take_c_by_pointer_ref(a)            == 8
+assert call_take_c_shared_ptr_by_value(a)       == 9
+assert call_take_c_shared_ptr_by_ref(a)         == 10
+assert call_take_c_shared_ptr_by_pointer(a)     == 11
+assert call_take_c_shared_ptr_by_pointer_ref(a) == 12
+
+assert call_take_c_by_pointer_with_null(a)                == -2
+assert call_take_c_by_pointer_ref_with_null(a)            == -3
+assert call_take_c_shared_ptr_by_value_with_null(a)       == -4
+assert call_take_c_shared_ptr_by_ref_with_null(a)         == -5
+assert call_take_c_shared_ptr_by_pointer_with_null(a)     == -6
+assert call_take_c_shared_ptr_by_pointer_ref_with_null(a) == -7
--- a/Examples/test-suite/ruby/li_boost_shared_ptr_director_runme.rb
+++ b/Examples/test-suite/ruby/li_boost_shared_ptr_director_runme.rb
@@ -25,7 +25,11 @@ class Derived < Base
     c.get_m
   end
 
-  def take_c_shared_ptr_by_value(c)
+  def take_c_by_ref(c)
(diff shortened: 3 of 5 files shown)


ojwb:
The explicit check for a `NULL` pointer in the new typemaps shouldn't be needed - if you follow how `  SWIG_NewPointerObj(NULL, x, y)` is handled:

```
#define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerOb
j(ptr, type, flags)
```

and

```
SWIGRUNTIME PyObject *
SWIG_Python_NewPointerObj(void *ptr, swig_type_info *type, int flags) {
  if (!ptr) {
    return SWIG_Py_Void();
  } else {
```

and

```
SWIGRUNTIMEINLINE PyObject *
SWIG_Py_Void(void)
{
  PyObject *none = Py_None;
  Py_INCREF(none);
  return none;
}
```

So you should get `None` out for `NULL` in automatically.  It also looks like your version fails to increment the ref count of `Py_None`, which seems to be required - see: https://docs.python.org/3/c-api/none.html

Looks plausible otherwise, though it'd be good to extend the testsuite to cover this case.

radarsat1:
Okay, I'll prefer SWIG_Py_Void, that makes sense.  (Sorry lots of little things like that, hard to notice..)

To be clear, my understanding of the issue is that in the definition,

~~~
SWIGRUNTIME PyObject *
SWIG_Python_NewPointerObj(void *ptr, swig_type_info *type, int flags) {
  if (!ptr) {
...
~~~

Here, ptr is a pointer to a shared_ptr, which itself contains `NULL`, so maybe it's not necessary to check that `ptr!=NULL`, but it is necessary to check that `*ptr != nullptr`, which can't be done after casting it to `void*`.  The point is that in the case that shared_ptr(0), ptr is non-NULL, but points to a NULL pointer, which is not detectable by Python code, so then the dereferenced object causes a segfault.

So I can't actually get rid of the `($1 && *$1)` stuff, since that must be checked before it is cast to `void*`.

I'll update the code to use SWIG_Py_Void and see if I can turn this into a test case as suggested.

ojwb:
Oh I see - then yes, you probably do need to check that up front.

wsfulton:
For runtime testing, please convert Examples/test-suite/ruby/li_boost_shared_ptr_director_runme.rb to python and add as Examples/test-suite/li_boost_shared_ptr_director_runme.py. See #964.

radarsat1:
Thanks to @wsfulton for the suggestion.  I updated the branch (warning, force push) to basically copy the ruby solution.  I also added implementations for take-by-ref, take-by-pointer, take-by-pointer-ref, which I split off into a different commit in case it causes problems -- since it adds new abstract functions it might break the tests for other languages.

radarsat1:
I see that the travis job failed due to adding some tests that broke in Ruby.  I don't know much about Ruby but I think I can probably manage to fix that.

wsfulton:
@radarsat1 I'd rather this patch didn't bit rot, are you able to add in the missing Ruby typemaps to fix the Ruby tests?

radarsat1:
Hopefully better luck now.

radarsat1:
Can someone help please? There is still a Travis error but now it is unrelated to Ruby.  In fact it is when "SWIGLANG=" is the empty string, and only on Mac, and there does not appear to be an actual error in the job log.  Any help appreciated.

The travis error above unhelpfully says "The Travis CI build could not complete due to ..." without completing the explanation. Clicking on Details does not seem to have the rest of the phrase.  I suspect a timeout.

wsfulton:
The Travis infrastructure is having a bad week, I've restarted this failure (which is not related to your change), so it should go green soon. The patch looks okay after a quick look. I'll examine it in more detail in a few days and merge. Thanks.

radarsat1:
Excellent, thanks!  I'll keep my eye on it in case there are any new errors.

wsfulton:
The li_boost_shared_ptr.i files were nearly identical between ruby/python. There are some subtle differences now. You can see them doing a visual diff between them. I'd like to easily 'diffable' to help maintain them, so could you put the new typemaps in the same place in each file? The other subtle diffs should be easily seen then. One is the out typemap for TYPE *CONST& was fixed for python but not ruby... not sure why your runtime tests didn't pick this up.

Similar changes could be added to the other languagess with shared_ptr typemaps. That will be useful in a separate patch. If you are interested in contributing this please say.

radarsat1:
I see only one file named `li_boost_shared_ptr.i`, in `Examples/test-suite`, so I assume you mean `Lib/ruby /boost_shared_ptr.i` and `Lib/python/boost_shared_ptr.i`

I think I've changed all the important differences now.  I'm happy to squash this PR if you prefer.

> One is the out typemap for TYPE *CONST& was fixed for python but not ruby... not sure why your runtime tests didn't pick this up.

I had not yet found a test that breaks on this.  I think it is fixed now.

> Similar changes could be added to the other languagess with shared_ptr typemaps. That will be useful in a separate patch. If you are interested in contributing this please say.

I wish I had the time but I don't have much experience with the FFIs of many other languages.  Maybe I can do it for Java.  For the rest, I feel I'd be shooting in the dark a bit, and I don't really have time for it, as my project is purely C++/Python.

radarsat1:
I updated the branch to be rebased on the current master.  The original branch can be found at https://github.com/radarsat1/swig/tree/fix-null-shared_ptr-directors2

I have run all ruby and python tests, so hopefully everything is fixed.  I have one question that I am not sure about.  Even before my patch, some places in `boost_shared_ptr.i` use `SWIG_NO_NULL_DELETER_0` and some places use `SWIG_NO_NULL_DELETER_$owner`.  I was not sure exactly when one or the other was correct.  In my directorin impls, I used  `SWIG_NO_NULL_DELETER_0`.  Can you confirm with it is correct, or should be  `SWIG_NO_NULL_DELETER_$owner`?

radarsat1:
Similar, in some calls to `SWIG_NewPointerObj` the last argument is `SWIG_POINTER_OWN` and in others it is `SWIG_POINTER_OWN | %newpointer_flags`.

I generally copied the option from the existing directorin instances in each file, which includes `%newpointer_flags`, but it would be nice if you could double-check correctness.

Thanks in advance!

ojwb:
`SWIG_NO_NULL_DELETER_0` is used when a new object is created to return, because then `$owner` (which tells you whether the original object is owned or not) is not relevant, e.g.:

```
%typemap(varout, fragment="SWIG_null_deleter_python") CONST TYPE * {
  SWIG_SHARED_PTR_QNAMESPACE::shared_ptr< CONST TYPE > *smartresult = $1 ? new S
WIG_SHARED_PTR_QNAMESPACE::shared_ptr< CONST TYPE >($1 SWIG_NO_NULL_DELETER_0) :
 0;
  %set_varoutput(SWIG_NewPointerObj(%as_voidptr(smartresult), $descriptor(SWIG_S
HARED_PTR_QNAMESPACE::shared_ptr< TYPE > *), SWIG_POINTER_OWN));
}
```

If you follow the macro chain, you'll end up at code such as this:

```
%fragment("SWIG_null_deleter", "header") {
struct SWIG_null_deleter {
  void operator() (void const *) const {
  }
};
%#define SWIG_NO_NULL_DELETER_0 , SWIG_null_deleter()
%#define SWIG_NO_NULL_DELETER_1
}
```

So in short, if `$owner` is `0` or when `SWIG_NO_NULL_DELETER_0` is hard-coded, the smart pointer's deleter is set to do nothing; when `$owner` is `1`, the default deleter is used.

Not sure about `%newpointer_flags` - it looks like it's actually always zero, except for Perl where it's defined to `$shadow` (see `Lib/typemaps/swigtypemaps.swg` and `Lib/perl5/perltypemaps.swg`, but sadly there's nothing really explaining this in either place).

radarsat1:
> So in short, if $owner is 0 or when SWIG_NO_NULL_DELETER_0 is hard-coded, the smart pointer's deleter is set to do nothing; when $owner is 1, the default deleter is used.

Got it.  But I guess the question is, what is correct for directors passing objects by shared_ptr?  I am afraid hard-coding 0 could result in dangling pointers.  In this light I suppose for shared_ptr that allowing $owner=1 would be correct.  That is, new shared_ptr objects should be fully created and destroyed by the director mechanism.  Only in the case that raw pointers are converted to shared_ptr should $owner=0 be hard-coded.  Is that correct?  In that case all the shared_ptr directorin typemaps should use _$owner.

radarsat1:
Okay, looks like I managed to confuse myself.  The directorin typemaps use the `shared_ptr<const shared_ptr&>` constructor, which doesn't even take a deleter argument, where as the existing, non-shared_ptr directorin typemaps use the `shared_ptr<Y>(Y*, deleter>` constructor, and therefore need to specify SWIG_NO_NULL_DELETER.  So it doesn't apply in the case that we are instantiating new shared_ptrs based on existing shared_ptrs.  The commit should be okay.

However, in looking at this I realized the shared_ptr-by-ref case could be simpler.  Currently it points at the ref object passed to the function but if that ref goes out of scope and the calling function holds onto it, there could be problems.  I'll change it to the approach in the shared_ptr-by-value case, where it creates a new, owned shared_ptr object.  Same for the other cases in fact (shared_ptr-by-pointer, shared_ptr-by-pointer-ref).

ojwb:
You need to use `$owner` when putting an existing pointer into a `shared_ptr` (because in this case the pointer should eventually be deleted by the `shared_ptr` if and only if we own it).

I'm not entirely sure what the hardcoded `0` case is for now I think about it more.  The case where we make a copy of the pointed-to object using `new` should use the default deleter regardless of `$owner` (because we definitely own a copy we make).  The code I quoted is actually allocating a new **shared_ptr** with `new`, not copying the pointed to object.

And it generally isn't useful to copy a shared pointer but change the deleter to be used for the pointed to object - it would be unusual for there to be more than one right way to delete a particular object.

radarsat1:
Yes, agreed on all points.  I realized after writing that question that much of the code was either (a) making a new shared_ptr to an existing pointer, in which case specifying the deleter is necessary, and `DELETER_0` should be hard-coded, or (b) not copying the shared_ptr but passing in a pointer to the existing shared_ptr on the stack.  I realized (a) was only for non-shared_ptr cases, and (b) was wrong in all shared_ptr cases, since if the called function keeps the pointer object* and then the shared_ptr goes off the stack, it could lead to segfaults.  Better to make a new, owned shared_ptr, so I changed the code correspondingly.  I think the patch is ok now.

\* caveat: unless the language-side code is sufficiently smart to make a proper smart shared_ptr copy with reference increase?  I think this is not the case since it is passed in via `%as_voidptr(smartresult)`, which tells me the language doesn't know how to make a shared_ptr copy.  Correct me if I'm wrong, but my understanding is that the language will keep its own ref-count of the "newed" shared_ptr; in that case always using `new shared_ptr<>` is correct.

radarsat1:
Can this be merged?

ojwb:
I'm not familiar enough with the shared_ptr stuff to feel confident to OK it, so you'll need to wait for William to get to it.

radarsat1:
@ojwb @wsfulton just a friendly 2-weeks *bump*

> I'd rather this patch didn't bit rot

wsfulton:
I've actually merged this locally and am filling in some holes in the implementation. I've extended it to Octave/R/Scilab and am now doing Java and C#. I'll push it to master when I've completed it (in a few days time I guess).

radarsat1:
Ah excellent, thanks for the update!

wsfulton:
See 2d095b2fe3815b665d6a663258819fba0d5471e7 and 65020645aa8752ef3c18bd80b297f528672418ff for a bunch of director shared_ptr related changes added after this patch.

PR #1035, opened 2017-07-26
Prevent writeback of a const char* array through a director when usin…

gtbX:
…g the byte[] %typemap.

### DIFF
--- a/Lib/java/java.swg
+++ b/Lib/java/java.swg
@@ -1394,6 +1394,7 @@ SWIG_PROXY_CONSTRUCTOR(true, true, SWIGTYPE)
 }
 %typemap(javadirectorin, descriptor="[B") (const char *STRING, size_t LENGTH) "$jniinput"
 %apply (const char *STRING, size_t LENGTH) { (char *STRING, size_t LENGTH) }
+/* Enable write-back for non-const version */
 %typemap(argout)  (char *STRING, size_t LENGTH) {
   if ($input) JCALL3(ReleaseByteArrayElements, jenv, $input, (jbyte *)$1, 0);
 }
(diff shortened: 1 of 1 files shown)


wsfulton:
Can you please elaborate on what this change is trying to achieve and include a use case/example. Thanks.

gtbX:
Sure.

We have a C++ class like below that we're trying to extend in Java:

    class Foo {
    public:
      void write(const void* data, size_t len) = 0;
    };
    %apply (char *STRING, size_t LENGTH) { (const void *data, size_t len) };

However, this fails when passing actual const data, since the default typemap will try to write data back into the *STRING argument:

    static const uint32_t bar = 0xFFFF0000;
    Foo *foo = getFooInstanceFromJava();
    foo->write(&bar, sizeof(bar)); // fail, can't write to write-protected memory

This patch adds a const char* typemap that doesn't have the write-back. ie., the %apply above would change to

    %apply (const char *STRING, size_t LENGTH) {const void *data, size_t len) };

wsfulton:
I see, your patch is basically making sure the 'directorargout' typemap is not used ensuring `GetByteArrayRegion` is not called, hence the input byte[] array remains unchanged.

I also think the argout typemap for `const char *` should make sure the Java byte[] array is not modified by using `JNI_ABORT` instead of 0:
```
%typemap(argout)  (const char *STRING, size_t LENGTH) {
  if ($input) JCALL3(ReleaseByteArrayElements, jenv, $input, (jbyte *)$1, JNI_ABORT);
}
```
and leave the non-const version as is:
```
%typemap(argout)  (char *STRING, size_t LENGTH) {
  if ($input) JCALL3(ReleaseByteArrayElements, jenv, $input, (jbyte *)$1, 0);
}
```

ojwb:
> I also think the argout typemap for `const char *` should make sure the Java byte[] array is not modified by using `JNI_ABORT` instead of 0

That seems a good idea, though this only ensures that the Java array isn't modified if the buffer was copied - if the buffer was pinned then we got a C pointer to the actual contents of the Java array, and so any modifications will have been made directly to that already.

PR #1046, opened 2017-08-01
New support of php7 wrapper via zend classes api

nihal95:


Lib/php/phprun.swg
```diff
@@ -258,46 +260,9 @@ SWIG_SetZval( zval *zv, int object, int class_obj ,void *ptr, swig_type_info *ty
...
-  if (!zval_ce || !lookup_ce)
-    return 0;
-
-  if (strcmp(zend_class_name->val,ty_name)) {
-    if (is_derived_class(zval_ce,lookup_ce))
-      return_value = 1;
-  }
-  else
-    return_value = 1;
+#define Z_FETCH_OBJ_P(zv) php_fetch_object(Z_OBJ_P(zv))
```
ojwb: We shouldn't define our own macros with names like those used by PHP's API headers.  If someone reading the code sees a use of `Z_FETCH_OBJ_P` they'll go and look at the PHP API to try to find what it does.  And if a future version of PHP adds a macro called `Z_FETCH_OBJ_P` it'll clash.  Name it something qualified with "SWIG" instead, e.g. `SWIG_Z_FETCH_OBJ_P`.

Lib/php/phprun.swg
```diff
@@ -258,46 +260,9 @@ SWIG_SetZval( zval *zv, int object, int class_obj ,void *ptr, swig_type_info *ty
...
 
-  return return_value;
+static inline
+swig_object_wrapper * php_fetch_object(zend_object *obj) {
```
ojwb: `php_fetch_object` ought to be qualified with a `swig_` prefix too.

This function is only used by the macro above, so why not have the function take a `zval *` parameter instead, do the `Z_OBJ_P` conversion inside the function, and then the macro isn't needed at all.

Source/Modules/php.cxx
```diff
@@ -2573,51 +2660,21 @@ class PHP : public Language {
...
-      Printf(s_header, "  %s_object_handlers.free_obj = %s_free_storage;\n",symname,symname);
-      Printf(s_header, "  %s_object_handlers.dtor_obj = %s_destroy_object;\n",symname,symname);
-      Printf(s_header, "  obj->std.handlers = &%s_object_handlers;\n  obj->newobject = 1;\n  return &obj->std;\n}\n\n\n",symname);
+      if (className != symname)
+        class_name = symname;
+      else
+        class_name = className;
 
       if (Len(classes) != 0)
         Printf(all_cs_entry, " { NULL, NULL, NULL }\n};\n\n");
```
ojwb: I believe this should be `ZEND_FE_END`.

Source/Modules/php.cxx
```diff
@@ -502,7 +554,11 @@ class PHP : public Language {
     /* Emit all of the code */
     Language::top(n);
 
+    if (Len(classes) > 0)
+      Printf(all_cs_entry, " { NULL, NULL, NULL }\n};\n\n");
```
ojwb: I believe this should be `ZEND_FE_END`.

ojwb:
You've rewritten at least one commit which was already merged to the gsoc2017-php7-classes-via-c-api branch.

Don't do that - it means I can't just merge your branch but have to deal with the conflicts, and it makes a mess of the history of the merged branch.

There's no real need to keep rewritten the tip of the branch either, and I'd suggest you don't, but you really shouldn't rewrite older commits, especially ones which have been merged.

I'm going to push the merge, but the comments above do need addressing.

ojwb:
FYI, this is what the history now looks like:

```
*   25397f0a (HEAD -> gsoc2017-php7-classes-via-c-api) Merge remote-tracking branch 'nihaln/changed_code' into gsoc2017-php7-classes-via-c-api
|\  
| * f6acfc2b (nihaln/changed_code) SWIG Director Support with Class Structure. Refactor Code to support rename.
| * 2e5f0fac Add class method check in Test Cases.
| * 6b8aae18 Refactor Code.
| * 4c559199 Refactor Code to use swig_object_wrapper for wrapping classes.
* | b5c408dd (origin/gsoc2017-php7-classes-via-c-api) Refactor Code to use swig_object_wrapper for wrapping classes.
|/  
* 05c25a71 Remove globals check in the php test-suite.
* b38e6d95 Remove SWIG_Get_Type function usage in class constants.
```

Source/Modules/php.cxx
nihal95: I believe I missed this. Will do

nihal95:
Yeah. This error is due to the discrepancy between the __wrap_delete of resource and the free_storage method. Will resolve it.

Lib/php/factory.i
```diff
@@ -95,7 +95,15 @@ if (!dcast) {
   Type *dobj = dynamic_cast<Type *>($1);
   if (dobj) {
     dcast = 1;
-    SWIG_SetPointerZval(return_value, SWIG_as_voidptr(dobj),$descriptor(Type *), $owner);
+    zend_object *std = NULL;
+    if ($newobj) {
+      zend_class_entry *ce = zend_lookup_class(zend_string_init("Type", sizeof("Type")-1, 0));
```
ojwb: This isn't quite right - in the example in the comment at the top, `Type` will be `Foo::Point` (and then `Foo::Circle`), and that's not going match a PHP class name.

Lib/php/php.swg
```diff
@@ -119,30 +119,40 @@
 
 %typemap(in) SWIGTYPE &
 %{
-	if (SWIG_ConvertPtr(&$input, (void **) &$1, $1_descriptor, 0) < 0 || $1 == NULL)
+  if (zend_lookup_class(zend_string_init("$lower_param",sizeof("$lower_param")-1,0))) {
+    $1 = $obj_value;
+  }
+  else {
+	  if (SWIG_ConvertPtr(&$input, (void **) &$1, $1_descriptor, 0) < 0 || $1 == NULL)
       SWIG_PHP_Error(E_ERROR, "Type error in argument $argnum of $symname. Expected $1_descriptor");
```
ojwb: Crazy indentation of the `if` above.

Source/Modules/php.cxx
```diff
@@ -1270,11 +1298,23 @@ class PHP : public Language {
...
+        ptr+= strlen(Char(iname)) - 4 - strlen(Char(name));
+        wname = (String*) ptr;
+      }
+      else {
+        wname = iname;
+      }
+    }
     else if (wrapperType == staticmembervar) {
       // Shape::nshapes -> nshapes
       char *ptr = Char(strrchr(GetChar(n, "name"),':'));
```
ojwb: `Char(x)` is a SWIG thing to get you a `char *` pointer for a `DOH` type.

But `strrchr()` is a standard C function which returns a `char*`, so at best applying `Char()` to the returned value is useless here.

However, I don't think you should be parsing the C++ type name here, as that will ignore any `%rename` in effect.  I think you probably want something like `char * ptr = Char(Getattr(n, "staticmembervariableHandler:sym:name"));` here.

Source/Modules/php.cxx
```diff
@@ -1288,8 +1328,11 @@ class PHP : public Language {
        */
       if (Cmp(strrchr(GetChar(n, "sym:name"),'_'),"_set") == 0)
         static_setter = true;
-      else
-        static_getter = true;
+      else if (Cmp(strrchr(GetChar(n, "sym:name"),'_'),"_get") == 0) {
```
ojwb: I assume you're trying to test if the name ends in "_set" (or "_get") above?

The code doesn't actually do that though - it'll match for names containing "_set" or "_get" anywhere in them, which means that `num_settlers_get` be misidentified as a setter.

I'd suggesting adapting this existing code which does a very similar check correctly:

```
      const char *p = Char(iname);
      if (strlen(p) > 4) {
        p += strlen(p) - 4;
        String *varname = Getattr(n, "membervariableHandler:sym:name");
        if (strcmp(p, "_get") == 0) {
        } else if (strcmp(p, "_set") == 0) {
          Setattr(shadow_set_vars, varname, iname);
        }
      }
```

Source/Modules/php.cxx
```diff
@@ -2712,8 +2752,7 @@ class PHP : public Language {
 	while (base.item && GetFlag(base.item, "feature:ignore")) {
 	  base = Next(base);
 	}
-        Printf(s_oinit, "%s_ce = zend_register_internal_class_ex(&%s_internal_ce, %s_ce);\n", class_name , class_name, Getattr(base.item, "sym:name"));
-
+  baseClassExtend = Getattr(base.item, "sym:name");
```
ojwb: Line misindented.

Source/Modules/php.cxx
```diff
@@ -3038,17 +3118,25 @@ class PHP : public Language {
...
     Wrapper *f = NewWrapper();
+
     Printf(f->def, "/* This function is designed to be called by the zend list destructors */\n");
     Printf(f->def, "/* to typecast and do the actual destruction */\n");
     Printf(f->def, "static void %s(zend_resource *res, const char *type_name) {\n", destructorname);
 
-    Printf(f->code, "if(zend_lookup_class(zend_string_init(\"%s\",sizeof(\"%s\")-1,0))) {\n", name, name);
-    Printf(f->code, "return;\n}\n\n");
+    Printf(f->def, "\n\nif(zend_lookup_class(zend_string_init(\"%s\",sizeof(\"%s\")-1,0))) {\n", intermediate_name, intermediate_name);
+    Printf(f->def, "return;\n}\n");
```
ojwb: What's this trying to check?

Looking at the generated code, it seems to be checking "is this type a wrapped class", which we should be able to statically determine that at wrapper generation time, rather than needing a run-time check.

It looks to me like if the type isn't a wrapped class, then the generated code which follows isn't going to compile anyway, but maybe I'm missing something...

ojwb:
I've rebased and merged to the gsoc2017-php7-classes-via-c-api branch, but please address the points I've raised above for the next batch of changes.

PR #1056, opened 2017-08-04
R fixes 2

fschlimb:
* replaces #900
* white space cleanup (following SWIG conventions)
* making smart-pointer classes derive from parent smart-pointers
* constructors now returning smart pointers (if class declared as such)
* fixing smart-pointer and NAMESPACE support
* allowing R package names containing dots ('.')
* allowing multiple get accessors in R

I expanded this to several commits as requests by @joequant.

@joequant @wsfulton #900 fails because I changed the following option:
      -dll <name>      - Name of the DLL (without the .dll or .so suffix).
from
 * Default is the module name.

to
 * Default is the package name.

This makes more sense to me. If you agree, I can adjust the tests so that they pass.

### DIFF
--- a/Source/Modules/r.cxx
+++ b/Source/Modules/r.cxx
@@ -12,6 +12,7 @@
  * ----------------------------------------------------------------------------- */
 
 #include "swigmod.h"
+#include "cparse.h"
 
 static const double DEFAULT_NUMBER = .0000123456712312312323;
 
@@ -908,7 +909,7 @@ int R::DumpCode(Node *n) {
     writeListByLine(namespaceFunctions, ns);
     Printf(ns, ")\n");
     Printf(ns, "\nexportMethods(\n");
-    writeListByLine(namespaceFunctions, ns, 1);
+    writeListByLine(namespaceMethods, ns, 1);
     Printf(ns, ")\n");
     Delete(ns);
     Delete(s_namespace);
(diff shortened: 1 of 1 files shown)


joequant:
Give me a few days and I'll take a look.

joequant:
The commits look reasonable.  I'll check them in over the weekend.

joequant:
Also what I'd like to do is to check in all of the patches except for the
module name one.  I'd like the tests to be fixed and checked in as part of
the same patch.

fschlimb:
Yes, I can fix the examples if you agree to the change to the argument "-dll"

fschlimb:
Ok, I removed the changes which made the test fail:
- R does not allow packages to have '_' in the name (for doing "R CMD build" and alike). This will not be checked with this patch. Getting this right requires significant changes to the test suite and I currently have no time address (what seems to have caused the failures is that the test-suite by-passes the "regular" package creation).
- the overload changes also broke some tests. The fixes were partial anyway. I will come back later with this, maybe.
- no change in behavior of -dll

All tests pass now.
@joequant ok?

joequant:
Committed changes

theartful:
I'm wondering why the change to make smart pointers inherit from parent smart pointers.
This causes diamond problem kind of situation.
For example: if we have
```
A inherits from B inherits from C
```
we get
```
(smart A)  inhertis from (A)               inherits from (B) inherits from (C) 
                 inherits from (smart B)     inherits from (B) inherits from (C)
                                                          inherits from (smart C) inherits from C
```
This caused problems when byte compiling (to create a package) a codebase I work on with big hierarchies while using smart pointers everywhere.
After I removed the extra inheritance from the parent smart pointer, byte compilation worked again.

PR #1071, opened 2017-08-17
fix bug: raise NameError when compiled with cython

Huarong:
After compile the python wrapper with Cython, it will throw and exception when running.

    NameError: name '__file__' is not defined

which was raise by this piece of code:
    
    def swig_import_helper():
         from os.path import dirname
         import imp
         fp = None
         try:
             fp, pathname, description = imp.find_module(
                 '_grnn_qt', [dirname(__file__)])
         except ImportError:
             import _grnn_qt
             return _grnn_qt
         if fp is not None:
             try:
                 _mod = imp.load_module('_grnn_qt', fp, pathname, description)
             finally:
                 fp.close()
             return _mod

After doing some [search](https://stackoverflow.com/questions/16771894/python-nameerror-global-name-file-is-not-defined), I found it is because `__file__` does not exist after being complied.

This piece of code will not catch `NameError` when `__file__` is not defined. **I change it to catch every exceptions**. So it will import from the same directory.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -850,7 +850,7 @@ public:
       Printf(default_import_code, tab8 "mname = '.'.join((pkg, '%s')).lstrip('.')\n", module);
       Printv(default_import_code, tab8, "try:\n", NULL);
       Printv(default_import_code, tab8, tab4, "return importlib.import_module(mname)\n", NULL);
-      Printv(default_import_code, tab8, "except:\n", NULL);
+      Printv(default_import_code, tab8, "except (ImportError, NameError):\n", NULL);
       Printf(default_import_code, tab8 tab4 "return importlib.import_module('%s')\n", module);
       Printf(default_import_code, tab4 "%s = swig_import_helper()\n", module);
       Printv(default_import_code, tab4, "del swig_import_helper\n", NULL);
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -850,7 +850,7 @@ public:
       Printf(default_import_code, tab8 "mname = '.'.join((pkg, '%s')).lstrip('.')\n", module);
       Printv(default_import_code, tab8, "try:\n", NULL);
       Printv(default_import_code, tab8, tab4, "return importlib.import_module(mname)\n", NULL);
-      Printv(default_import_code, tab8, "except ImportError:\n", NULL);
+      Printv(default_import_code, tab8, "except:\n", NULL);
       Printf(default_import_code, tab8 tab4 "return importlib.import_module('%s')\n", module);
       Printf(default_import_code, tab4 "%s = swig_import_helper()\n", module);
       Printv(default_import_code, tab4, "del swig_import_helper\n", NULL);


ojwb:
Thanks for the patch.

However silently swallowing unexpected exceptions seems unhelpful as it's likely to lead to more confusing error messages - the error will be for a failure trying to import from the current directory instead of the original error we hit.  So it seems better to just add `NameError` as an exception type to catch here, i.e.:

```
except (ImportError, NameError):
```

Also, this really deserves a mention in `CHANGES.current`.

Huarong:
@ojwb You are right. I update the patch.

ojwb:
Thanks for the update.

The CI failure was due to a timeout which I doubt is related to this change, but I've restarted the failed job so we can be sure.

wsfulton:
I'm not sure about this. I can't replicate your problem with cython, it works fine with the 'simple' examples that ships with SWIG. I added setup.py as follows:
```python
from distutils.core import setup
from Cython.Build import cythonize

setup(
  name = 'Hello world app',
  ext_modules = cythonize("example.pyx"),
)
```
then ran the following in the Examples/python/simple directory and everything is okay without your patch:
```
~/swig/github/swig/Examples/python/simple $ rm -rf build && git checkout example.c && make clean build && mv example.py example.pyx && python setup.py build_ext --inplace && python runme.py
make -f ../../Makefile SRCDIR='' TARGET='example' python_clean
make[1]: Entering directory '/home/william/swig/github/swig/Examples/python/simple'
rm -rf __pycache__
rm -f *_wrap* *~ .~* mypython *.pyc
rm -f core 
rm -f *.o *.so *.so
rm -f example.py
if test -f runme.py; then rm -f runme3.py runme3.py.bak; fi
case "x" in x|x./);; *) rm -f runme.py;; esac
make[1]: Leaving directory '/home/william/swig/github/swig/Examples/python/simple'
make -f ../../Makefile SRCDIR='' SRCS='example.c' \
TARGET='example' INTERFACE='example.i' python
make[1]: Entering directory '/home/william/swig/github/swig/Examples/python/simple'
env SWIG_LIB=../../../Lib  ../../../swig -python   -o example_wrap.c example.i
gcc -c -fpic -I.  example_wrap.c example.c   -I/usr/include/python2.7 -I/usr/lib/python2.7/config
gcc -shared   example.o  example_wrap.o   -o _example.so
make[1]: Leaving directory '/home/william/swig/github/swig/Examples/python/simple'
Compiling example.pyx because it changed.
[1/1] Cythonizing example.pyx
running build_ext
building 'example' extension
creating build
creating build/temp.linux-x86_64-2.7
x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fno-strict-aliasing -Wdate-time -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security -fPIC -I/usr/include/python2.7 -c example.c -o build/temp.linux-x86_64-2.7/example.o
x86_64-linux-gnu-gcc -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions -Wl,-Bsymbolic-functions -Wl,-z,relro -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -Wdate-time -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security -Wl,-Bsymbolic-functions -Wl,-z,relro -Wdate-time -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security build/temp.linux-x86_64-2.7/example.o -o /home/william/swig/github/swig/Examples/python/simple/example.so
The gcd of 42 and 105 is 21
Foo =  3.0
Foo =  3.1415926
```
There is no NameError, what are you doing differently?

ojwb:
@wsfulton FWIW I didn't attempt to reproduce the issue myself as I'm not familiar with cython, I just checked the patch assuming the reported problem existed.  If we don't get any feedback, I've no problem with it being reverted.

ojwb:
> I've no problem with it being reverted.

Oh, I didn't actually apply it yet.

@wsfulton Looking at the code for `swig_import_helper()` quoted in the original report, this is specific to the Python 2.6 code path (you seem to be using Python 2.7, which uses different code which doesn't look at `__file__`).

However this PR changes the Python 2.7 code path, so it seems unlikely it actually fixes the problem reported.

@Huarong Can you fix the patch to change the correct `except` (you want the one about 11 lines later) and verify it actually fixes your situation (i.e. testing with the same Python version you get a failure without the patch and a pass with)?

Huarong:
@wsfulton Your command raise error:

```
~/open/swig/Examples/python/simple on git:master x [14:18:23]
$ rm -rf build && git checkout example.c && make clean build && mv example.py example.pyx && python setup.py build_ext --inplace && python runme.py
make -f ../../Makefile SRCDIR='' TARGET='example' python_clean
make[1]: Cd directory “/home/huohuarong/open/swig/Examples/python/simple”
make[1]: ../../Makefile: No such file or directory
make[1]: *** No rule to create target“../../Makefile”。 Stop。
make[1]: Leaving directory“/home/huohuarong/open/swig/Examples/python/simple”
make: *** [clean] error 2
```

@ojwb this is not specific to the Python 2.6 code, but for >= 2.6.
```
from sys import version_info
if version_info >= (2, 6, 0):

     def swig_import_helper():
         from os.path import dirname
         import imp
         fp = None
```

And I have already verified it. Everything worked fine when I added `NameError`

wsfulton:
@Huarong, you need to run configure in the top level directory to get the missing Makefile, see the top level README file. If you can't/won't/don't run it, it would be useful to instead see your own equivalent script so that we can reproduce. As @ojwb said, the if statement checking for python >=2.7.0 means that the block of code checking for python >= 2.6.0 will only run for 2.6.x code only and not >= 2.7.0. Your patch modifies the >= 2.7.0 code path. Please provide more definitive details on the version of Python this fixes and how you it can be reproduced.

ojwb:
> this is not specific to the Python 2.6 code, but for >= 2.6.

The code you quote isn't what SWIG currently generates - e.g. after running the testsuite `Examples/test-suite/python/cpp11_static_assert.py` contains:

```
from sys import version_info as _swig_python_version_info
if _swig_python_version_info >= (2, 7, 0):
    def swig_import_helper():
        import importlib
        pkg = __name__.rpartition('.')[0]
        mname = '.'.join((pkg, '_cpp11_static_assert')).lstrip('.')
        try:
            return importlib.import_module(mname)
        except ImportError:
            return importlib.import_module('_cpp11_static_assert')
    _cpp11_static_assert = swig_import_helper()
    del swig_import_helper
elif _swig_python_version_info >= (2, 6, 0):
    def swig_import_helper():
        from os.path import dirname
        import imp
        fp = None
        try:
            fp, pathname, description = imp.find_module('_cpp11_static_assert', 
[dirname(__file__)])
        except ImportError:
            import _cpp11_static_assert
            return _cpp11_static_assert
        try:
            _mod = imp.load_module('_cpp11_static_assert', fp, pathname, descrip
tion)
        finally:
            if fp is not None:
                fp.close()
        return _mod
    _cpp11_static_assert = swig_import_helper()
    del swig_import_helper
else:
    import _cpp11_static_assert
del _swig_python_version_info
```

This has `elif _swig_python_version_info >= (2, 6, 0):` not the `if` you show, and the >= 2.6.0 case is only checked if >= 2.7.0 is not true, so as I said this code is only used for 2.6.x.

It looks like your problem is you're trying to use some old SWIG version, or perhaps an old branch, and the bug you're trying to fix has already been addressed there.

wsfulton:
Seems to be a non-issue. Closing. Please re-open if you can provide the additional requested details.

Issue #644, opened 2016-04-01
Swig generation for R fails if function prototype uses the keyword  "result"


purplefishies:
If you have a function definition as follows

###  /* foo.h */
#ifndef FOO_H
float *foo( int *result );
#endif

and 

### /* foo.c */
#include  "foo.h"
float *foo( int *res )
{
    static float value = 100;
    float *retval = &value;
    int result = 43;
    printf("Going to return a result\n");
    *res = 42;
    printf("Other result: %d\n", result );
    returl retval;
}

And then run Swig to generate a wrapper, I get a wrapper that has the following entry

SWIGEXPORT SEXP
R_swig_foo ( SEXP result)
{
**float *result = 0 ;**   /* This will cause a compiler error */
 int *arg1 = (int *) 0 ;
  unsigned int r_nprotect = 0;
  SEXP r_ans = R_NilValue ;
  VMAXTYPE r_vmax = vmaxget() ;
...

What you get is that result is used twice.  This is different than if you run python generators where the name is always called   ***SWIGUNUSEDPARM(self)*  and  **args**

## How to reproduce 

Using these files
[RSwigFail.zip](https://github.com/swig/swig/files/200154/RSwigFail.zip)

```bash
unzip RSwigFail.zip
cd RSwigFail
swig -r 
R CMD SHLIB -o AIOUSB.so Foo_wrap.c foo.c
```

ojwb:
@joequant Did you see this report?  It seems there's a fixed name `result` which should probably be `swig_result` or similar, or else the parameter name shouldn't just use the name from the function being wrapped.  The latter would avoid similar issues with `arg1`, etc, but perhaps there's a reason to want to use the same parameter name here.

joequant:
Yup.  Trying to come up with a general fix, but I might just hack it by
using a internal name like _swig_result.  Will fix this weekend.

ojwb:
I've looked at this a bit, and I think actually the answer is to not use the C/C++ parameter names for the wrapper function parameter names.  There are many potential collisions, not just `result`.  For example `arg1`, `arg2`, etc (which could be `swig_` prefixed) but also collisions with the R API which can't be avoided except by renaming the parameter - consider `Examples/r/simple` with this simple change - a totally valid C API to be wrapping:

```
--- a/Examples/r/simple/example.i
+++ b/Examples/r/simple/example.i
@@ -2,6 +2,6 @@
 %module example
 
 %inline %{
-extern int    gcd(int x, int y);
[...]

PR #1077, opened 2017-08-25
Safer default for cresult_variable_name

ajbouh:
If a wrapped method has a parameter called `result`, compilation of the resulting C++ unit will fail.

A safer default name for `cresult_variable_name` is `"_swig_result"` as it makes a collision with a parameter name much less likely.

A full fix would be a unique and deterministically chosen name. In the meantime, this workaround is simple and I believe it's strictly better than the current value.

It's also spiritually equivalent to the workaround described here: https://github.com/swig/swig/issues/644

### DIFF
--- a/Examples/python/libffi/example.i
+++ b/Examples/python/libffi/example.i
@@ -56,7 +56,7 @@
   }
   if (ffi_prep_cif(&cif, FFI_DEFAULT_ABI, vc+3,
                    &ffi_type_uint, types) == FFI_OK) {
-    ffi_call(&cif, (void (*)()) execlp, &result, values);
+    ffi_call(&cif, (void (*)()) execlp, &_swig_result, values);
   } else {
     free(types);
     free(values);
@@ -155,7 +155,7 @@ int execlp(const char *path, const char *arg1, ...);
   }
   if (ffi_prep_cif(&cif, FFI_DEFAULT_ABI, vc+1,
                    &ffi_type_uint, types) == FFI_OK) {
-    ffi_call(&cif, (void (*)()) printf, &result, values);
+    ffi_call(&cif, (void (*)()) printf, &_swig_result, values);
   } else {
     free(types);
     free(values);
--- a/Examples/test-suite/errors/swig_typemap_warn.stderr
+++ b/Examples/test-suite/errors/swig_typemap_warn.stderr
@@ -1,6 +1,6 @@
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int abc (arg1) - argnum: 1 &1_ltype: int * descriptor: SWIGTYPE_int
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int arg3 (arg3) - argnum: 3 &1_ltype: int * descriptor: SWIGTYPE_int
-swig_typemap_warn.i:7: Warning 1001: Test warning for 'out' typemap for double mmm (result) - name: mmm symname: mmm &1_ltype: double * descriptor: SWIGTYPE_double
+swig_typemap_warn.i:7: Warning 1001: Test warning for 'out' typemap for double mmm (_swig_result) - name: mmm symname: mmm &1_ltype: double * descriptor: SWIGTYPE_double
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int abc (arg1) - argnum: 1 &1_ltype: int * descriptor: SWIGTYPE_int
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int arg3 (arg3) - argnum: 3 &1_ltype: int * descriptor: SWIGTYPE_int
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int abc (arg1) - argnum: 1 &1_ltype: int * descriptor: SWIGTYPE_int
--- a/Examples/test-suite/java/special_variables_runme.java
+++ b/Examples/test-suite/java/special_variables_runme.java
@@ -15,13 +15,13 @@ public class special_variables_runme {
   public static void main(String argv[]) 
   {
     verify(special_variables.ExceptionVars(1.0, 2.0),
-           "result = Space::exceptionvars(arg1,arg2);  Space::exceptionvars  ExceptionVars   Java_special_1variables_special_1variablesJNI_ExceptionVars  ");
+           "_swig_result = Space::exceptionvars(arg1,arg2);  Space::exceptionvars  ExceptionVars   Java_special_1variables_special_1variablesJNI_ExceptionVars  ");
 
     verify(special_variables.overloadedmethod(),
-           "result = Space::overloadedmethod();  Space::overloadedmethod  overloadedmethod  __SWIG_1 Java_special_1variables_special_1variablesJNI_overloadedmethod_1_1SWIG_11  ");
+           "_swig_result = Space::overloadedmethod();  Space::overloadedmethod  overloadedmethod  __SWIG_1 Java_special_1variables_special_1variablesJNI_overloadedmethod_1_1SWIG_11  ");
 
     verify(special_variables.overloadedmethod(10.0),
-          "result = Space::overloadedmethod(arg1);  Space::overloadedmethod  overloadedmethod  __SWIG_0 Java_special_1variables_special_1variablesJNI_overloadedmethod_1_1SWIG_10  ");
+          "_swig_result = Space::overloadedmethod(arg1);  Space::overloadedmethod  overloadedmethod  __SWIG_0 Java_special_1variables_special_1variablesJNI_overloadedmethod_1_1SWIG_10  ");
 
     ABC a = new ABC(0, 0.0);
     verify(special_variables.getDeclaration(), "SpaceNamespace::ABC::ABC(int,double) SpaceNamespace::ABC::ABC(int,double)");
--- a/Examples/test-suite/python_varargs_typemap.i
+++ b/Examples/test-suite/python_varargs_typemap.i
@@ -39,7 +39,7 @@
 
 %feature("action") testfunc {
   char **vargs = (char **) arg3;
-  result = testfunc(arg1, arg2, vargs[0], vargs[1], vargs[2], vargs[3], vargs[4],
+  _swig_result = testfunc(arg1, arg2, vargs[0], vargs[1], vargs[2], vargs[3], vargs[4],
                     vargs[5], vargs[6], vargs[7], vargs[8], vargs[9], NULL);
 }
 
--- a/Examples/test-suite/rename.h
+++ b/Examples/test-suite/rename.h
@@ -35,7 +35,7 @@ namespace Space {
 %exception Space::ABC::operator ABC %{
 #if defined(__clang__)
   // Workaround for: warning: conversion function converting 'Space::ABC' to itself will never be used
-  result = *arg1;
+  _swig_result = *arg1;
 #else
   $action
 #endif
--- a/Examples/test-suite/rename4.i
+++ b/Examples/test-suite/rename4.i
@@ -81,7 +81,7 @@ namespace Space {
 %exception Space::ABC::operator ABC %{
 #if defined(__clang__)
   // Workaround for: warning: conversion function converting 'Space::ABC' to itself will never be used
-  result = *arg1;
+  _swig_result = *arg1;
 #else
   $action
 #endif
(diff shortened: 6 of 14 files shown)


ojwb:
Seems the testsuite also needs updating to match this change (see all the failures in the CI builds).

Also, it looks like this actually fixes #644 to me - I had assumed this name was R-specific when I commented there.

ajbouh:
There are a lot of places that result was used in tests (instead of $result). I'm not an expert in all these code generators or tests, so it may take me some time to find them all.

@ojwb, can you provide some support in debugging these failures?

wsfulton:
Mangling all symbol names with a `swig_` prefix arguably should be done throughout the generated code base, but it isn't the approach that was taken when SWIG was first written. Note that `swig_result` not `_swig_result` would be in line with the normal SWIG mangling. However, I am going to reject this patch as it will break too much user code, where it is possible to override what gets assigned to `result` in the "except", "action" and other SWIG features. The fact that the tests fail should ring alarm bells that this change is too disruptive.

The convention for the generated C++ code is to **not** use the names of the parameter variables. This is adhered to by most of the target languages that I'm aware of except R. The correct approach is for R to not use 'result' as the parameter name in the generated C code, but to use s_arg1 for the 1st parameter, s_arg2 for the 2nd parameter and so on. I suggest these names as this is what is already used if a user does not provide a parameter name in the function declaration. This will be a fairly easy change in the R layer for #644.

ajbouh:
I experienced this problem in the director implementation for C++, not R.

It sounds like the C++ code that SWIG generates doesn't properly follow the convention you've described?

Also, based on my efforts to get the tests to pass, it seems that many of the generators (and tests) don't actually respect Swig_cresult_name().

ajbouh:
And it seems that I can't finish debugging which tests are still broken because you've closed my PR.

ajbouh:
Your comment regarding users of SWIG assigning directly to `result` implies that there shouldn't be a `Swig_cresult_name_set()` method in cwrap.c, as any use of it would render a build of SWIG that can't pass tests and won't work with user code.

Based on my research, I think that tests and users alike should be using the $result special variable, which respects `Swig_cresult_name()`. Most of the tests and codegen logic already use this value, only a few do not.

I'm happy to use "swig_result" instead of "_swig_result", and to update as many tests as I can to use $result.

But like I said, as of now SWIG creates C++ code that won't compile if you use the director feature for Go and one of the virtual methods has a parameter named result.

wsfulton:
@ajbouh, I missed the Go problem in this area, is it in another issue? This patch was to change the name of the result variable for all languages. Arguably $result should be used by users, but when I tried, it doesn't expand in the action or except feature. As outlined I'm not going to break a huge existing code base. I'd rather help you work around this corner case of one parameter having the name 'result', or fix it in Go/R whichever language is of concern to you. I'm not overly familiar with Go, but will see if I can reproduce your problem.

Sorry if the closing of this one no longer shows the link to Travis for your debugging of the failing tests. The link you need is https://travis-ci.org/swig/swig/builds/269448879. Using Travis repeatedly with multiple updates for debugging is discouraged as our test-suite is very heavyweight and it stops Travis from being used for other users. You're better off running Travis in a targeted manner using your own Travis+Github account. Or even better (although possibly harder to set up initially) is to run the tests locally, see https://github.com/swig/swig/wiki/Getting-Started if you go down this route.

The generators ought to be using `Swig_cresult_name()`. A separate patch to fix this would be most welcome. Your observation of `Swig_cresult_name_set()` is entirely valid, however, it isn't used anywhere. From git blame, I can see that I added it in commit 24133bacd7ba5a670607a7ce5e7ac9800c99ae1e in a previous cleanup to remove the hard coding of 'result' to make this more flexible, but no-one has ever used it.

wsfulton:
Okay, this is a problem specific to directors and only in the additional director generated method. The convention for these methods is to use the user provided name for some historic reason instead of arg1, arg2 and result. In this case, there is far less opportunity for users to change the code with special features and the name can be changed from result. In Go the extra variables should be mangled as swig_arg1, swig_arg2, swig_result. The swig_arg1, swig_arg2 mangling does already happen. The patch below fixes it for Go. We should add a test using result as a variable name in a director method and make similar fixes to other languages. @ajbouh, if you'd like to do this for all the director languages, I can re-open the ticket if you rebase your patch to get rid the commits you have already put in this patch as the commits are just noise. Otherwise, a new pull request referring to this one will be fine.

```patch
diff --git a/Source/Modules/go.cxx b/Source/Modules/go.cxx
index 9160304..dcec81f 100644
--- a/Source/Modules/go.cxx
+++ b/Source/Modules/go.cxx
@@ -5116,6 +5116,7 @@ private:
 
     ParmList *parms = Getattr(n, "wrap:parms");
     SwigType *result = Getattr(n, "type");
+    String *result_name = NewStringf("swig_%s", Swig_cresult_name());
 
     String *callback_wname = Swig_name_wrapper(callback_name);
     Append(callback_wname, unique_id);
@@ -5172,11 +5173,9 @@ private:
       }
       if (SwigType_type(result) != T_VOID) {
 	Printv(w->code, "    long : 0;\n", NULL);
-	String *rname = NewString(Swig_cresult_name());
-	String *cg = gcCTypeForGoValue(n, result, rname);
+	String *cg = gcCTypeForGoValue(n, result, result_name);
 	Printv(w->code, "    ", cg, ";\n", NULL);
 	Delete(cg);
-	Delete(rname);
       }
 
       Printv(w->code, "  } SWIGSTRUCTPACKED swig_a;\n", NULL);
@@ -5254,11 +5253,9 @@ private:
       Printv(f_gc_wrappers, "}\n\n", NULL);
     } else {
       if (SwigType_type(result) != T_VOID) {
-	String *r = NewString(Swig_cresult_name());
-	String *tm = gccgoCTypeForGoValue(n, result, r);
-	Wrapper_add_local(w, r, tm);
+	String *tm = gccgoCTypeForGoValue(n, result, result_name);
+	Wrapper_add_local(w, result_name, tm);
 	Delete(tm);
-	Delete(r);
       }
 
       String *args = NewString("");
@@ -5322,7 +5319,7 @@ private:
 	  Swig_warning(WARN_TYPEMAP_DIRECTOROUT_UNDEF, input_file, line_number,
 		       "Unable to use type %s as director method result\n", SwigType_str(result, 0));
 	} else {
-	  Replaceall(tm, "$input", Swig_cresult_name());
+	  Replaceall(tm, "$input", result_name);
 	  Replaceall(tm, "$result", "c_result");
 	  Printv(w->code, "  ", tm, "\n", NULL);
 	  String *retstr = SwigType_rcaststr(result, "c_result");
@@ -5334,6 +5331,7 @@ private:
       }
     }
 
+    Delete(result_name);
     Delete(callback_wname);
   }
 
@@ -5346,6 +5344,7 @@ private:
   void makeCgoDirectorMethodWrapper(Node *n, Wrapper *w, String *callback_name) {
     ParmList *parms = Getattr(n, "wrap:parms");
     SwigType *result = Getattr(n, "type");
+    String *result_name = NewStringf("swig_%s", Swig_cresult_name());
 
     Printv(f_c_directors, "extern \"C\" ", NULL);
 
@@ -5378,11 +5377,9 @@ private:
     Printv(f_c_directors, ";\n", NULL);
 
     if (SwigType_type(result) != T_VOID) {
-      String *r = NewString(Swig_cresult_name());
-      String *tm = gcCTypeForGoValue(n, result, r);
-      Wrapper_add_local(w, r, tm);
+      String *tm = gcCTypeForGoValue(n, result, result_name);
+      Wrapper_add_local(w, result_name, tm);
       Delete(tm);
-      Delete(r);
     }
 
     String *args = NewString("");
@@ -5420,7 +5417,7 @@ private:
 
     Printv(w->code, "  ", NULL);
     if (SwigType_type(result) != T_VOID) {
-      Printv(w->code, Swig_cresult_name(), " = ", NULL);
+      Printv(w->code, result_name, " = ", NULL);
     }
     Printv(w->code, callback_name, "(go_val", args, ");\n", NULL);
 
@@ -5447,7 +5444,7 @@ private:
 		     "Unable to use type %s as director method result\n", SwigType_str(result, 0));
       } else {
 	tm = Copy(tm);
-	Replaceall(tm, "$input", Swig_cresult_name());
+	Replaceall(tm, "$input", result_name);
 	Replaceall(tm, "$result", "c_result");
 	Printv(w->code, "  ", tm, "\n", NULL);
 	String *retstr = SwigType_rcaststr(result, "c_result");
@@ -5457,6 +5454,7 @@ private:
       }
       Delete(result_str);
     }
+    Delete(result_name);
   }
```

ajbouh:
I'll try the above patch and see how it works. I'm open to helping move things forward overall, but I haven't gotten the swig test suite running locally yet.

wsfulton:
@ajbouh are you able to finish this off?

ojwb:
(Actually this isn't specific to Go from comments above).

@ajbouh Did you make any further progress with this?

ojwb:
> Okay, this is a problem specific to directors and only in the additional director generated method. The convention for these methods is to use the user provided name for some historic reason instead of arg1, arg2 and result.

I wonder if these should be changed to use `arg1`, etc instead too.

Renaming `result` doesn't really solve the collision problem, it just addresses one case.  It's possible to collide with names in the target language's C API and we can't solve those by renaming.  E.g. consider this patch to the go director example:

```
--- a/Examples/go/director/director.h
+++ b/Examples/go/director/director.h
@@ -17,22 +17,22 @@ public:
        };
 
 protected:
-       virtual std::string Foo() {
+       virtual std::string Foo(int intgo) {
                return "Foo";
        };
 
-       virtual std::string Bar() = 0;
+       virtual std::string Bar(int) = 0;
 };
 
 
 class FooBarCpp : public FooBarAbstract
 {
 protected:
-       virtual std::string Foo() {
+       virtual std::string Foo(int intgo) {
                return "C++ " + FooBarAbstract::Foo();
        }
 
-       virtual std::string Bar() {
+       virtual std::string Bar(int) {
                return "C++ Bar";
        }
 };
```

Which leads to compilation errors such as:

```
example_wrap.cxx:326:8: error: expected ‘;’ before ‘swig_arg2’
  326 |   intgo swig_arg2;
      |        ^~~~~~~~~~
      |        ;
```

(It would need more tweaks to actually work anyway because `Foo` and `Bar` now take a parameter, but that's not relevant here.)

This is perhaps more of a theoretical concern as the API identifiers are mostly not things you'd naturally call parameters and typically have some sort of namespacing, but some APIs are much less clean than others in this regard, and it seems better to solve the full problem rather than just the case that's been reported so far.

This also shows that SWIG can already generate suitable names since it does for `Bar` where the parameter name is omitted in the C++ header - generated code is:

```
std::string SwigDirector_FooBarAbstract::Foo(int intgo) {
```

vs

```
std::string SwigDirector_FooBarAbstract::Bar(int arg0) {
```

I'll prepare a PR to do that.

wsfulton:
Fixing the generators to use Swig_cresult_name() has been the discussion to date and I think would be useful as the generated director code uses a **variable name** called result which is a common enough **parameter name** and can be easily avoided with the patch I included for Go. While parameter name clashes with target language **types** is of course is possible, it does not seem to be something that is a problem. The original parameter names are somewhat more user friendly and should be safe to keep as is in the as the generated header file is intended for external usage and is not just implementation detail. However, I'm torn as to whether the implementation in the generated .cxx file should be changed.

@ojwb, I'm not sure I follow what you intend your patch to do?

ojwb:
I'm suggesting the change to parameter names shown around the "vs" in my previous comment.  Good point about the header, I'll look at whether the names can be left there.  If they can this seems a clear improvement in robustness and better to address what's currently a mostly theoretical issue rather than wait for a user to have to hit a case and report it.  If we'd have to rename parameters in the header too I agree it's more debatable, though we could comment out the parameter names, e.g.: 

```
std::string SwigDirector_FooBarAbstract::Foo(int /* intgo */);
```

Regarding `result` collisions, there's also an old (and apparently ignored until now) user report of a collision between the `result` local variable SWIG declares and a `result` class in the API being wrapped:

https://sourceforge.net/p/swig/bugs/755/

The compile errors in that case are due to things like:

```
  result = (result *)new result();
```

which could be addressed by qualifying the type like so:

```
  result = (::result *)new ::result();
```

ojwb:
The collision on `result` has been reported again (against SWIG/Python in #2571).

I see above I was going to work up a new PR to rename the parameters in the `_wrap.cxx`, but was probably put off by wsfulton seeming not so keen, but then not responding to my clarification for what I was planning.

@wsfulton I can look at this, but not going to waste the time if you're then just going to object to the changes so a response would be useful.  I've assumed in the past that no response meant you were OK with something but that not being the case...

Issue #1068, opened 2017-08-09
C++ namespace declaration does not work inside the SWIG `%inline` block


luav:
According to the [SWIG docs](http://swig.org/Doc3.0/SWIG.html#SWIG_nn43) `%inline` block seems should support any valid C++ code. However, the following interface file `nstpl.i`:
```c++
%module nstpl
%{
#include "nstpl.hpp"
using namespace nst;
%}

%include "nstpl.hpp"
using namespace nst;
```
is compiled well, where is the expected shorter equivalent
```c++
%module nstpl

%inline %{
#include "nstpl.hpp"
using namespace nst;
%}
```
yields a wrapper generation error:
```
nstpl.i:13: Error: Nothing known about namespace 'nst'
```
The header `nstpl.hpp` can contain just the namespace definition:
```c++
namespace nst {
typedef int  Int;
}  // nst
```
Note that
```c++
%module nstpl

%inline %{
#include "nstpl.hpp"
%}
using namespace nst;
```
yields the same error. Thus, the namespaces defined in the files included into the `%inline` blocks are ignored, which is *definitely a bug*.

**Environment:**
SWIG Version 4.0.0
Compiled with g++ [x86_64-pc-linux-gnu]
Configured options: +pcre

ojwb:
Your first two examples aren't equivalent as `#include` is not the same as `%include`, so closing as invalid.

This is really not to do with namespaces, it's just about how `#include` works - if you run with `-includeall` then this works, because SWIG follows the `#include`.

luav:
Thanks for the clarification @ojwb. It would be nice to mention this in the [Inlined code blocks](http://swig.org/Doc3.0/SWIG.html#SWIG_nn43).

ojwb:
To propose an update to the manual, open a PR.

wsfulton:
BTW, the manual is in the Doc/Manual directory in case you aren't sure where the source for it is.

Issue #1049, opened 2017-08-03
Weird undocumented behavior of the "%shared_ptr" for the C++ templates


luav:
Documentation for the [%shared_ptr](http://swig.org/Doc3.0/Library.html#Library_std_shared_ptr) should be extended covering interaction with C++ templates.

If `Hierarchy` is a template then usage of `%shared_ptr(Hierarchy);` yields the following error:
```c++
error: type/value mismatch at argument 1 in template parameter list for 'template<class _Tp> class std::shared_ptr'
   std::shared_ptr< Hierarchy > tempshared1 ;
                              ^
compilation terminated due to -Wfatal-errors.
```
If `SHierarchy` is one of the instantiations of the `Hierarchy`, for example:
```c++
%template(SHierarchy) Hierarchy<SomeType>;
%template(AHierarchy) Hierarchy<AnotherType>;
```
then `%shared_ptr(SHierarchy);` results in the transparent shared_ptr also for `AHierarchy` (**unexpected**) and all other classes that are instances of the `Hierarchy` template, which should be documented and *should yield some warning* on the wrapper generation.

In the end, the right solution, which is **[not documented](http://swig.org/Doc3.0/Library.html#Library_std_shared_ptr)**:
```c++
%shared_ptr(Hierarchy<>);
```

ojwb:
Please can you submit a patch to improve the documentation?

luav:
The pull request for the documentation (manual) update has been performed.  
Still, it would be nice to *yield some warning on the wrapper generation* if the template specialization is used inside the `%shared_ptr` declaration, or to prevent such declarations at all.

wsfulton:
I just found a way to reproduce the problem. A slight tweak to the newly documented code to use template default arguments is:
```c++
// number.h
#include <memory>

template<int N=42> struct Number {
  int num;
  Number() : num(N) {}
  static std::shared_ptr<Number<N>> make() { return std::make_shared<Number<N>>(); }
};
```
```swig
%include <std_shared_ptr.i>

%shared_ptr(Number<10>);
%shared_ptr(Number<>);

%{
  #include "number.h"
%}
%include "number.h"

%template(Number10) Number<10>;
%template(Number42) Number<>;
```
A Java type wrapper class `SWIGTYPE_p_std__shared_ptrT_NumberT_42_t_t` is generated instead of `Number42` which is indicative of a type lookup problem in SWIG. I can't reproduce your ` error: type/value mismatch` error originally provided though.

[...]

PR #1086, opened 2017-09-11
Docs update according to #1068, #1049

luav:
`#include` directive inside the `%inline` blocks (#1068) and `%shared_ptr` usage for the templates (#1049) documented.


wsfulton:
Thanks. I've merged these in and edited. The `%shared_ptr` changes are however not correct and I've adapted your contribution to describe how it should work. I wasn't exactly sure how you came to the conclusion that you only need to use a single `%shared_ptr` for all template instantiations. I'll look at #1049 next, but for now I've changed your docs to how it should work and does work for simple template instantiations with a view to enhancing the docs later with specifics for default template arguments if needed.

luav:
Thank you for the adoption and explanation of how it should be. Without the docs I attempted to clarify that empirically overcoming some bugs... I will comment in #1049 to clarify the outlined conclusions.

Issue #1087, opened 2017-09-12
SWIG does not discriminate enum classes (C++11) [in function parameters]


luav:
Interface `fovrenum.i`:
```c++
%module fovrenum

%inline %{
enum class FileFormat: uint8_t {
	UNKNOWN = 0
};

%rename(toFFVal) to_val(FileFormat, bool bin);
int to_val(FileFormat flag, bool bin=false)
{
	return 0;
}

enum class ClsOutFmt: uint8_t {
	NONE
};

%rename(toCFVal) to_val(ClsOutFmt, bool bin);
int to_val(ClsOutFmt flag, bool bin=false)
{
	return 0;
}
%}
```

Note that overloaded functions `to_val()` take distinct enum **classes**, which are not discriminated by SWIG:
```sh
$ swig -v -c++ -python fovrenum.i
...
Generating wrappers...
fovrenum.i:36: Warning 509: Overloaded method to_val(ClsOutFmt) effectively ignored,
fovrenum.i:26: Warning 509: as it is shadowed by to_val(FileFormat).
```

**Environment:**
SWIG Version 4.0.0 (latest from the master branch)
Compiled with g++ [x86_64-pc-linux-gnu]
Configured options: +pcre

luav:
As it found, function declaration in SWIG (at least for `%rename`) distinguish default parameters from non-default unlike C++, which is not documented in the manual. So the following renaming works fine:
```c++
%rename(toFFVal) to_val(FileFormat, bool bin=false);
%rename(toCFVal) to_val(ClsOutFmt, bool bin=false);
```

I will describe this in the SWIG manual (docs)...

PR #1088, opened 2017-09-12
Docs update (%rename for functions with default parameters explained, see #1087)

luav:
%rename for functions with default parameters explained, see #1087

### DIFF
--- a/Doc/Manual/SWIG.html
+++ b/Doc/Manual/SWIG.html
@@ -1737,6 +1737,43 @@ careful about namespaces and your use of modules, you can usually
 avoid these problems.</p>
 
 <p>
+<tt>%rename</tt> directive respects function parameters and <i>discriminates default parameters</i> from non-default,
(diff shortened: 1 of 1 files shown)


wsfulton:
Yes, this area of the docs could be better, thanks for pointing out. I used your patch, but decided that this isn't quite the right place in the docs to put this info. That section is a gentle introduction to %rename and is for simple identifiers and not the full power of %rename for C++ which is covered in the SWIG and C++ chapter. The problem you had is ultimately covered in the SWIG and C++ chapter. I've thus referenced this from the %rename section you patched. Hopefully that is clearer for readers of the docs. The majority of relevant changes are in 9cc05a22f6c3007b92e5682aa4accd7d69c658c6.

BTW, for any future pull requests, please give the branch a suitable name, unlike 'luav-tmp2' as I normally merge the branch and the branch name appears in the main merge commit. I cherry-picked your change instead this time.

luav:
OK, thanks for the explanation of the branch naming.

PR #1092, opened 2017-09-14
Fix evil_diamond_prop test case. - gsoc2017

nihal95:
- This fixes the problem of no reaching parent class variables.

Note: Only the parent class is going to save extra variables created by __set

### DIFF
--- a/Source/Modules/php.cxx
+++ b/Source/Modules/php.cxx
@@ -1110,7 +1110,7 @@ public:
   /* Magic methods __set, __get, __isset is declared here in the extension.
      The flag variable is used to decide whether all variables are read or not.
    */
-  void magic_method_setter(Node *n, bool flag) {
+  void magic_method_setter(Node *n, bool flag, String *baseClassExtend) {
 
     if (magic_set == NULL) {
       magic_set = NewStringEmpty();
@@ -1118,6 +1118,9 @@ public:
       magic_isset = NewStringEmpty();
     }
     if (flag) {
+      if (Cmp(baseClassExtend, "Exception") == 0 || !is_class_wrapped(baseClassExtend)) {
+        baseClassExtend = NULL;
+      }
       Wrapper *f = NewWrapper();
       // Need arg info set for __get magic function with one variable.
       String * arginfo_code = NewString("0");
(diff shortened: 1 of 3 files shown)


Source/Modules/php.cxx
```diff
@@ -1193,10 +1203,17 @@ class PHP : public Language {
       Printf(f->code, "%s\n",magic_get);
       Printf(f->code, "\nelse if (strcmp(arg2->val,\"thisown\") == 0)\n{\n");
       Printf(f->code, "if(arg->newobject) {\nRETVAL_LONG(1);\n}\nelse {\nRETVAL_LONG(0);\n}\n}\n\n");
-      Printf(f->code, "else {\nif (!arg->extras) {\nRETVAL_NULL();\n}\n");
+      Printf(f->code, "else {\n");
+      if (baseClassExtend) {
+        Printf(f->code, "PHP_MN(%s___get)(INTERNAL_FUNCTION_PARAM_PASSTHRU);\n", baseClassExtend);
+        Printf(f->code, "if (Z_ISNULL_P(return_value) && arg->std.ce == SWIGTYPE_%s_ce) {\n", class_name);
```
ojwb: There are two reasons why `Z_ISNULL_P(return_value)` could be true here.  One is that the attribute isn't set on the parent class.  But the other reason is that it is set and its value is `NULL` - for example:

```
$ php7.0 -r '$a = new stdClass; $a->foo = NULL; var_dump($a);'
object(stdClass)#1 (1) {
  ["foo"]=>
  NULL
}
```

If an attribute is set to `42` via PHP it'll get set on the base class and also stored in the extras hash in the most derived class.  Then the C++ side updates the base class attribute to be `NULL`.  Then PHP read the attribute from the derived class, which fetches it from the base class, sees it is `NULL`, and returns `42` from the extras hash in the derived class instead, which is wrong - in this case `NULL` should be returned.

I continue to think that it would be better to just store "extra" attributes using a hash in the base class, and not ever store attributes twice.  If you store them twice, then you have to worry about something using the extra value when it's out of date.

ojwb:
BTW, the travis build is failing for PHP7 currently because the generated C code mixes declarations and code - that's valid as of C99, but SWIG aims to generate C90-compatible code (unless wrapping features which inherently imply a newer C standard), and in C90 you have to declare variables at the start of a block.

wsfulton:
Can the purpose of this patch be explained more fully? I'm not sure what it is trying to do.

The patch was updated to be C90 compatible and all the PHP tests are showing as passing on Travis and Appveyor @ojwb, is it good for merge now?

ojwb:
The patch is another step towards creating PHP classes via PHP's C API (instead of generating a `.php` script with the class definitions in).  Note it is not a PR for master, but rather for a branch we're using to help incremental review of this GSoC project.

I really need to get back to reviewing this work - I think it's how we want to be wrapping for PHP, and the bulk of the work has been done.

ojwb:
I still think the double tracking of attributes (in both the base class and derived class) needs fixing - it should actually be simpler to only track in base classes, reduces object storage overhead a little, and avoids weird corner cases.

But this fixes the only remaining failing testcase on the gsoc2017-php7-classes-via-c-api branch, so for now I think merging this is better than not merging it.

I've also cherry-picked a few commits from master to the branch to address issues with newer PHP and newer GCC.  Next step is to try to merge master to the branch.

Issue #315, opened 2015-01-29
python directors will not work if C++ exceptions are disabled


blaisegassend:
Currently python directors rely on C++ exceptions to raise various errors that can occur when trying to call a method that is implemented in Python. Code bases that do not allow exceptions can therefore not use python directors, which is unfortunate.

It appears that there are two uses of exceptions:
1. UnknownExceptionHandler.
2. Director*Exception::raise(PyObject *error, const char *msg).

My proposed solution is:
1. Get rid of UnknownExceptionHandler as it is dead code as far as I can tell.
2. Introduce a RaiseDirectorException() function which can be overridden by %feature(director:raise).

If anybody sees an issue with this solution let me know. I'll be submitting a patch shortly.

wsfulton:
I just want to make sure you have explored using the "director:except" feature?
The most flexible exception handling is done in Java which uses this feature plus "directorthrows" typemaps. I'd really like the other languages to use the same approach. Can you check that an implementation of this won't meet your needs, I can't say I've given your particular scenario much thought, but it might work. A consistent approach across the different languages (that are relevant) are a priority in accepting patches and so if what I've mentioned does not work, I'd like to understand why not. Some discussion of these were done in #91.

blaisegassend:
Hi William,

Unfortunately, there are four places where director.swig contains throw/try
statements, which simply cause compilation to fail. So a patch of some sort
is necessary so that those throws can be eliminated when some feature is
activated. Moreover, director:except only overrides one of the places where
the director code throws: when the callee has thrown an exception. Other
places that are not covered are: trying to call a pure method, type
mismatch in the value returned by python. So even with director:except you
can't get rid of all the throwing, and even if it did the code wouldn't
compile because methods containing throw are still present in the c++ file.

I had a quick look at directorthrows, and my understanding is that it is
unrelated to my needs. It seems to provide an alternate to director:except
to handle exceptions that happen in python/java. director:except is already
doing that fine for me. What I need to get rid of is the c++ exceptions
that are being raised when errors happen that are not due to an exception
in the python/java code. Am I understanding that correctly?

I am curious how Java is managing to do directors without c++ exceptions.
How is flow getting from the place where the java code returns to the place
that was called from java without using exceptions? Do you have an example
of a simple java director use in google3 so that I could look at the
generated code?

I'll share a patch in a bit so you can see what I'm up to.

Thanks,
Blaise

ojwb:
[...]

PR #1096, opened 2017-09-16
Show the original error if calling director method failed

vadz:
Previously, if calling the director method failed, the precise error
message was lost and just "Error detected" was shown to the user, making
it difficult to understand what the problem actually was.

Change the code to show the contents of the last Python error too to
make finding and fixing errors in Python director methods much simpler.

### DIFF
--- a/Lib/python/director.swg
+++ b/Lib/python/director.swg
@@ -269,6 +269,30 @@ namespace Swig {
     static void raise(const char *msg) {
       throw DirectorMethodException(msg);
     }
+
+    /* this should be called only if a Python error has occurred */
+    static void raise_after_pyerr(const char *msg) {
+      std::string full_msg(msg);
+
+      PyObject *exception, *v, *tb;
+      PyErr_Fetch(&exception, &v, &tb);
+      PyErr_NormalizeException(&exception, &v, &tb);
+
+      if (PyObject *s = PyObject_Str(v)) {
+        if (const char *cstr = SWIG_Python_str_AsChar(s)) {
+          full_msg += ": ";
+          full_msg += cstr;
+          SWIG_Python_str_DelForPy3(cstr);
+        }
+        Py_XDECREF(s);
+      }
+
+      Py_XDECREF(exception);
+      Py_XDECREF(v);
+      Py_XDECREF(tb);
+
+      raise(full_msg.c_str());
+    }
   };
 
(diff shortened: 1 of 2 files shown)


Lib/python/director.swg
```diff
@@ -269,6 +269,30 @@ namespace Swig {
...
+    /* this should be called only if a Python error has occurred */
+    static void raise_after_pyerr(const char *msg) {
+      std::string full_msg(msg);
+
+      PyObject *exception, *v, *tb;
+      PyErr_Fetch(&exception, &v, &tb);
+      PyErr_NormalizeException(&exception, &v, &tb);
+
+      if (PyObject *s = PyObject_Str(v)) {
+        if (const char *cstr = SWIG_Python_str_AsChar(s)) {
```
ojwb: How likely are these checks to fail?  We could avoid converting `msg` to `std::string` and back when we're just going to use it unaltered.

vadz: AFAIK they can never fail at all, the checks are just there to account for the "impossible" failures.

ojwb: OK, then there's no point complicating things.

It'd be good to have a testcase covering this change, but otherwise it looks sensible to me.

letmaik:
See my comment at https://github.com/swig/swig/issues/1117#issuecomment-340745356. I don't think this PR is the right approach.

degasus:
I have the opposite opinion of letmaik here. There are valid reasons why some exceptions must be handled in pure C++ (e.g. async worker threads calling broken python code), however I want to ask if the error message fetching should be done in the what() method instead. This removes the overhead of forwarded exceptions, and at the same time, it provides a way for C++ handlers to understand the issue.

ojwb:
In #1117 wsfulton said:
> I agree that we do need a better default and the default should be just to [throw] the original exception. The behaviour should be fully customisable too, for example it should be possible to not have any C++ exceptions as requested in https://github.com/swig/swig/issues/315.

Based on that I'm closing this as not the preferred approach.

@degasus said above:
> There are valid reasons why some exceptions must be handled in pure C++ (e.g. async worker threads calling broken python code), however I want to ask if the error message fetching should be done in the what() method instead. This removes the overhead of forwarded exceptions, and at the same time, it provides a way for pure C++ handlers to understand the issue.

I think you'd need to provide a custom `director:except` typemap if you want that.

Issue #1097, opened 2017-09-19
Transparent `%shared_ptr` fails compilation in some cases where non-transparent `shared_ptr` works fine


luav:
Transparent `%shared_ptr(Tmpl<Type>)` fails compilaiton in some cases where non-transparent shared_ptr works fine
```c++
%template(TmplType) Tmpl<Type>;
%template(TmplTypePtr) shared_ptr<Tmpl<Type>>;
```
---
Example for the `%shared_ptr(Graph<false>);`:
```sh
swig/wraps/daoc_java.cxx:6263:240: error: could not convert ‘result’ from ‘SwigValueWrapper<std::shared_ptr<std::__cxx11::list<daoc::Node<std::vector<daoc::Link<float, false>, std::allocator<daoc::Link<float, false> > > >, std::allocator<daoc::Node<std::vector<daoc::Link<float, false>, std::allocator<daoc::Link<float, false> > > > > > > >’ to ‘bool’
 :Node< std::vector< daoc::Link< daoc::LinkWeight,false > > > > >(result) : 0; 
                                                                            ^
compilation terminated due to -Wfatal-errors.
```
Code of the generated wrapper:
```c++
SWIGEXPORT jlong JNICALL Java_info_exascale_daoc_daocJNI_SGraph_1release_1_1SWIG_10(JNIEnv *jenv, jclass jcls, jlong jarg1, jobject jarg1_, jlong jarg2, jlong jarg3) {
  jlong jresult = 0 ;
  daoc::Graph< false > *arg1 = (daoc::Graph< false > *) 0 ;
  std::unordered_map< daoc::Id,daoc::Node< std::vector< daoc::Link< daoc::LinkWeight,false > > > * > *arg2 = (std::unordered_map< daoc::Id,daoc::Node< std::vector< daoc::Link< daoc::LinkWeight,false > > > * > *) 0 ;
  bool *arg3 = (bool *) 0 ;
  std::shared_ptr< Graph< false > > *smartarg1 = 0 ;
  SwigValueWrapper< std::shared_ptr< std::list< daoc::Node< std::vector< daoc::Link< float,false > > > > > > result;
  
  (void)jenv;
  (void)jcls;
  (void)jarg1_;
  
  smartarg1 = *(std::shared_ptr<  Graph<false> > **)&jarg1;
  arg1 = (Graph<false> *)(smartarg1 ? smartarg1->get() : 0); 
[...]

PR #1098, opened 2017-09-19
Side effects of %shared_ptr to the function overloading outlined (#1097)

luav:
Side effects of the transparent `%shared_ptr` to the function overloading outlined, see #1097

### DIFF
--- a/Doc/Manual/Library.html
+++ b/Doc/Manual/Library.html
@@ -1830,6 +1830,12 @@ System.out.println(val1 + " " + val2);
 </pre>
 </div>
 
+<p>
+<b>Note:</b> Transparent <tt>shared_ptr</tt> affects functions overloading, in particular overloading of the function having
+the transparent <tt>shared_ptr&lt;Type&gt;</tt> argument with the <tt>Type&amp;</tt> argument causes compilation errors.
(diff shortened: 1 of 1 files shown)


wsfulton:
If you have:
```c++
void UseType(std::shared_ptr<MyType> x);
void UseType(std::shared_ptr<MyType>* x);
```
SWIG will detect that they can't be overloaded as they use the same type in the target language and SWIG will ignore one of them with a warning to prevent compilation errors, eg:
```
example.i:48: Warning 516: Overloaded method UseType(std::shared_ptr< MyType > *) ignored,
example.i:47: Warning 516: using UseType(std::shared_ptr< MyType >) instead.
```
Similarly for:
```c++
void UseType(MyType x);
void UseType(MyType* x);
```
The typecheck typemaps are used to control this, but there is a bug/limitation comparing the smart pointer with non-smart pointers. I'll put a fix in for this.

luav:
Thank you @wsfulton ,
note that in my case the overloading was for `func(shared_ptr<T>)` vs `funct(T&)`, i.e. reference, not the pointer (which is internally processed similar to the pointer, but formally the function signature differs).

Issue #1100, opened 2017-09-21
`%template` and `%rename` prevent naming interface classes/functions with the C++ template name (break the former interfaces)


luav:
In the latest version of SWIG `%rename` and `%template` prevent alias creation for the template in the *global namespace* in case the alias name matches the template name from another namespace. It worked fine in the begin of August, and it seems was broken after 96e9941, see also #1093 (*update: no, it was broken by another commit...*).

The interface file `template_class.i`:
```c++
%module template_class
%{
#include <vector>
%}
%include <std_vector.i>

using std::vector;

%inline %{
namespace daoc {

using Id = unsigned;
using LinkWeight = float;

template <bool WEIGHTED=false>
struct InpLink {
	using Weight = LinkWeight;  //!< \copydoc LinkWeight

	constexpr static bool  IS_WEIGHTED = WEIGHTED;

	Id  id;  //!< Dest node id
	//constexpr static Weight  weight = 1; // SimpleLink<Weight>::weight;
	Weight  weight;  //!< Link weight

	InpLink(Id lid=0): id(lid)  {}
};

template <>
struct InpLink<true> {
	using Weight = LinkWeight;  //!< \copydoc LinkWeight

	constexpr static bool  IS_WEIGHTED = true;

	Id  id;  //!< Dest node id
	Weight  weight;  //!< Link weight

	InpLink(Id lid=0, Weight lweight=1)  // SimpleLink<Weight>::weight
	: id(lid), weight(lweight)  {}
};

}  // daoc
using namespace daoc;
%}

%template(SInpLink) InpLink<false>;  //!< Simple input link
%template(InpLink) InpLink<true>;  //!< Weighted input link

//!< Simple input links
%template(SInpLinks) vector<daoc::InpLink<false>>;  // Items<InpLink<false>>;
////!< Weighted input links
%template(InpLinks) vector<daoc::InpLink<true>>;  // Items<InpLink<true>>;
```
The wrapper generation yields the following warning:
```sh
...
Starting language-specific parse...
template_class.i:53: Warning 302: Identifier 'InpLink' redefined (ignored) (Renamed from 'InpLink< true >'),
template_class.i:24: Warning 302: previous definition of 'InpLink'.
Processing types...
```
[...]

PR #1101, opened 2017-09-23
Build revision (git commit) added in addition to the SWIG version

luav:
SWIG revision number (hash of the git commit) added in addition to the version, which allows to perform reproducible bug reporting and identification of the applied binary even if the repository was updated. The revision number also reduces the time to solve some issues, for example #1100 (need to find the commit hash of the working version of the executable).

It works in the following way:
```sh
$ ./swig -version

SWIG Version 4.0.0
SWIG Revision b3cca58

Compiled with g++ [x86_64-pc-linux-gnu]

Configured options: +pcre
```
```sh
$ ./swig -revision
b3cca58
```
If any uncommited files remained then the revision is appended with the `+` suffix:
```sh
$ ./swig -revision
6177a68+
```

The implementation is seamlessly integrated into the `autogen` to produce such compilation:
```sh
g++ -DHAVE_CONFIG_H   -DPACKAGE_REVISION="\"`git rev-parse HEAD | head -c7``git status --porcelain -uno | wc -l | sed -e '/^0/d' -e 's/.*/+/'`\"" -I../Source/Include ...
```
The revision number is automatically taken from the git repository as the current commit hash.  
For the tarball the revision is set to the `UNDEFINED_REVISION`, which is fine. Anyway, the tarballs are created for the release version, and the revision is useful for the frequent builds from the repository.


ojwb:
This seems problematic - while the git revision is dynamically substituted by make, `misc.c` won't get automatically rebuilt each time the git revision changes, only when it would get rebuilt for other reasons, and it's not a file which is likely to get rebuilt frequently for other reasons.  So this patch would result in a stale git revision being reported in many cases, which seems worse that not reporting one at all.

If we're going to do this, I think it needs to work more like this:

* a make rule which is always run by `make`, which sticks a define of the current git revision info into `gitrevision.h.tmp` and then compares that file to `gitrevision.h` and only does `mv gitrevision.h.tmp gitrevision.h` if the two files are different

* `misc.c` uses `#include "gitrevision.h"` to get the current revision

That way `misc.c` will always get rebuilt when the git revision changes, but we avoid it getting rebuilt on every invocation of `make`.

I'm not entirely sure this is all worth the effort though - it's kind of ugly that a "no-op" `make` still does stuff, and generally people who submit useful reports when building from git already say what git revision they're using.  The problem reports are those where there's no version information mentioned at all, and this won't help those because we aren't being told what `swig --version` says.

BTW, `head -c7` isn't portable, but truncating the SHA to 7 characters seems a bad idea anyway - there are already commits in the SWIG repo which have the same first 7 characters, for example:

```
$ git show d5dbb5e
error: short SHA1 d5dbb5e is ambiguous
hint: The candidates are:
hint:   d5dbb5e0 commit 2002-07-27 - Fixed interaction of keyword arguments and overloading.  Also, keyword arguments can be selectively enabled using %feature.
hint:   d5dbb5e2 commit 2005-12-26 - remove SWIG bug workaround
```

You can use `git rev-parse --short HEAD` to get enough digits to be unique, but that's not future proof as another commit could be added which collides between the build and the bug report.  Better to just include the full hash I think, then it's always clear which commit it is.

wsfulton:
I'd be interested to know what other C/C++ git based projects do to show the git revision number. There must be some common solutions out there given that it is nowadays so easy to build and obtain a non-official release version of code from github.

I havn't tried the patch, but suspect it'll break if git is not installed on a users machine. Surely it should use AC_CHECK_PROGS to look for git.

@ojwb's comments about the displayed revision being reliable need further thought/discussion. What I've done for this in the past on other projects (for Subversion based projects), is only display the official Subversion revision if the code is built by the Continuous Integration machines which provide the official built binaries. If the code is built in any other way, the Subversion revision number is replaced by something else less official (date/time stamp). For swig, we don't deploy the binaries (except for windows), so this isn't going to work.

luav:
@wsfulton :
> I havn't tried the patch, but suspect it'll break if git is not installed on a users machine.

No, when the revision is not defined then `UNDEFINED_REVISION` is shown. Thought, it might be better to write it in a compact way, like `NA`.

> display the official revision

The official git revision is too long. We can display it, but it is not more useful than the short revision.  
Yes, the short revision can be not unique, but you are always interested in the LATEST revision or know exactly the time gap by the used/supporting features. So it's fine, moreover `github` and `git` itself use exactly this approach (look at the [commit history](https://github.com/swig/swig/commits/master) on `github`)
```sh
/opt/repos/swig$ git rev-parse HEAD
ba45861b462e91e266456f510036768c2aa55be5

$ git rev-parse --short HEAD
ba45861
$ git log --pretty=format:'%h' -1
ba45861
```
@ojwb Yes, I will replace `head -c7` with `--short`, it is more optimal, but the result is exactly the same.

The main **point** to have the revision number, even latest (prelatest for non committed build) and approximate (anyway the ambiguity can be resolved easily) is to be able to **identify the sources corresponding to the available binary** if possible. It is very for bugs reporting and identification of the working/broken version of the previously built binaries.

The main usecase is the user's build from the `github` repository. It also works for any other sources distribution even without the installed git, just then `UNDEFINED_REVISION` is shown (`NA`). In the latter case it's possible to omit the revision output, but then there might be some confusion if you ask the user about the revision, which is not shown at all.

Source/Swig/typesys.c
```diff
@@ -817,8 +817,8 @@ SwigType *SwigType_typedef_resolve(const SwigType *t) {
 	Append(type, tsuffix);
 	Delete(targs);
 	Delete(tsuffix);
-	Delete(rtprefix);
       }
+      Delete(rtprefix);
```
luav: Memory leak fixed there for the case `!type && !SwigType_istemplate(base)`

ojwb:
> Yes, I will replace head -c7 with --short, it is more optimal, but the result is exactly the same.

It's not the same when abbreviating to 7 is ambiguous - then `--short` will give you more characters, while `head -c7` will give you an ambiguous abbreviated commit hash.

However, as I said I think it's unwise to abbreviate **at all** - even if a certain length is currently unambiguous, it could become ambiguous due to commits being pushed between the build and someone investigating a bug, and having to work out which of a set of commits which match a given abbreviated hash is likely to be the one which was used is just creating unnecessary work for ourselves.  Or worse, if the commit is on a private branch we might just end up looking at the wrong commit but not realise.

> Files inclusion to display the revision from the hardcoded string in the file is a BAD practice, because it does not catch any changes made by the user, which are not committed to the repository. Again, in my case "+" suffix is dynamically added, showing that the built sources based on that committed revision were updated and not committed.

I didn't suggest hardcoding string in a file, but rather generating a file dynamically (but such that its timestamp only updates when the content has actually changed).  There's no reason why the revision information in such a file can't contain a `+` suffix to indicate local changes.

> Another issue with the predefined strings or including files is when the user updates only some files (not the whole repository), which is catched dynamically by my approach, but not using the predefined strings/files.

Not sure what you're talking about here.  Perhaps you're mixing up git and subversion?  Subversion allows a "mixed" checkout, but git doesn't.

> it's better to have at least approximate information than not any information at all.

Unreliable information about what revision was used is worse than having to ask.

luav:
@ojwb 
> `--short` will give you more characters

No, it would required analysis of the whole commits tree. It just shortens to the [predefined length according to the official docs](https://git-scm.com/docs/git-rev-parse#git-rev-parse---shortlength).  
Anyway, thanks for the suggestion, the updated version contains `--short`, it is really a better and more efficient way.

> Subversion allows a "mixed" checkout, but git doesn't.

Git does allow to [checkout of any particular files / paths according to the official docs](https://git-scm.com/docs/git-checkout#git-checkout-emgitcheckoutem-p--patchlttree-ishgt--ltpathspecgt82308203). Or the same case occurs when you apply some patch, or edit some files not committing them.

The current version works for both git repository and tarballs. If you would like to re-implement the same functionality using files, then it's up to you, go on! But why haven't you done it yet and is there really any advantage of using the file(s)? - I see only several drawbacks...  
Otherwise, *why don't you merge this pull to the master* if you are satisfied with the functionality? - Just try it, check it locally. It seems to work for all cases and provides pretty useful functionality.

sethrj:
I want to point out that `git describe` is the idiomatic way of constructing version tags for use in software (not `git log --pretty=format:'%h' -1`). It also can be used to automatically inject the git version when releasing a repository with `git archive`, which solves the problem of saving a git version automatically into a *source-only* distribution.

To partly address @wsfulton's question about other projects' use and release of git version info, the Python tool [versioneer](https://github.com/warner/python-versioneer) is an excellent and thorough way of embedding git version/hash for Python, and it contains a good description of some of the subtleties of embedding git metadata into a distribution. It would be a useful prototype for an equivalent C++ tool if one doesn't exist.

ojwb:
Closing.  The approach taken by the patch here has problems as noted above, and they haven't been addressed for years.

vadz:
FWIW we've implemented a solution close to what @ojwb suggested above, i.e. we generate a `version.c` file containing (a function returning) the string returned by `git rev-parse HEAD` and only update this file when the string changes. This seems to work reasonably well, i.e. we didn't have any problems with it so far.

ojwb:
Well if someone cares enough to implement something along those lines for SWIG then cool, but keeping this open doesn't seem useful.

I'm not sold on this improving bug reports by showing the exact version used since IME SWIG bug reporters seem to either tell us precisely already or give no version information at all, and in the later case we don't get shown the `-version` output which would have this git commit information in if it was built from git.

However I can see users who built a version from git might sometimes find it helpful to know later the commit it was built from.

vadz:
The most useful aspect of this is having the full version embedded in the generated files, of course.

Anyhow, I'm not going to insist or anything, this is a patch I'll keep using locally but I have (much) bigger fish to (try to) merge into SWIG...

Issue #769, opened 2016-08-20
[Python] cannot import swig package in site-packages using swig >= 3.0.9


paleozogt:
This is somewhat similar to #691, however the details are slightly different (and its fix doesn't fix this problem).

I distribute my library as a package, where the wrapper py file is named `__init__.py`.  This worked fine with swig 3.0.8 and previous, but the changes to `swig_import_helper` seems to have broken it.

Here is a minimal example:

```
#!/bin/bash

# swig compile
mkdir pkg
echo "%module module" > module.i
echo "from . import module" > pkg/__init__.py
swig -python -outdir pkg module.i
mv pkg/module.py pkg/__init__.py
gcc -fPIC -shared -o pkg/_module.so module_wrap.c `python-config --cflags --ldflags`

# make a virtualenv to simulate whl installation
virtualenv -p `python-config --exec-prefix`/bin/python venv
cp -r pkg venv/lib/python2.7/site-packages

# can we import the package?
venv/bin/python -c 'import pkg'
```

which gives

```
Already using interpreter /usr/bin/python
New python executable in ~/Development/swig-example/.build/venv/bin/python
Installing setuptools, pip, wheel...done.
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "pkg/__init__.py", line 18, in <module>
    _module = swig_import_helper()
  File "pkg/__init__.py", line 17, in swig_import_helper
    return importlib.import_module(mname)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named _module
```

ojwb:
We also hit this with Xapian's Python bindings recently - we used to rename the generated `.py` file to `__init__.py` and this stopped working recently.  The SWIG docs seem to suggest that this isn't expected to work, so we assumed it was just by chance that it did before the recent changes:

> By default, SWIG generates proxy Python code – one *.py file for each *.i interface. The `__init__.py` files, however, are not generated by SWIG

see http://www.swig.org/Doc3.0/Python.html#Python_nn72

For now we're running the SWIG-generated wrappers through a rewriting script which replaces the insanely complicated conditional import code with just:

```
from . import _xapian
```

That works fine in this case for Python 2.6, 2.7 and 3.x.

[...]

PR #1105, opened 2017-09-27
Fix import swig package in site-packages using swig >= 3.0.9

mastersin:
Check `__name__.rpartition('.')` return for separator and return last
part of string if it doesn't same as expected.

See issue #769.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -846,7 +846,10 @@ public:
       Printv(default_import_code, "if _swig_python_version_info >= (2, 7, 0):\n", NULL);
       Printv(default_import_code, tab4, "def swig_import_helper():\n", NULL);
       Printv(default_import_code, tab8, "import importlib\n", NULL);
-      Printv(default_import_code, tab8, "pkg = __name__.rpartition('.')[0]\n", NULL);
+      Printv(default_import_code, tab8, "pkg_parts = __name__.rpartition('.')\n", NULL);
+      Printv(default_import_code, tab8, "pkg = pkg_parts[0]\n", NULL);
+      Printv(default_import_code, tab8, "if pkg_parts[1] != '.':\n", NULL);
+      Printv(default_import_code, tab8, tab4, "pkg = pkg_parts[2]\n", NULL);
       Printf(default_import_code, tab8 "mname = '.'.join((pkg, '%s')).lstrip('.')\n", module);
       Printv(default_import_code, tab8, "try:\n", NULL);
       Printv(default_import_code, tab8, tab4, "return importlib.import_module(mname)\n", NULL);


ojwb:
This breaks Python 2.4 (syntax error from the inline `if`: https://travis-ci.org/swig/swig/jobs/280310758).  While older Python isn't very interesting now, it'd be better to just use a multi-line if and keep this working in case anyone cares, at least if this change is going to be backported to 3.0.x.

Python 3 also failed on OS X, but due to travis flaky infrastructure reasons - I've restarted so we can see if there are any actual issues there (it passed on Linux, so likely not).

@wsfulton In #769 you concluded:
> Okay, let's not try support the 'new' requirement. If anything, let's just make the documentation clearer in this respect.

What are your thoughts on this patch?  It seems clean and very unlikely to break anything, but we've had bad experiences with changing this import logic and `%pythonimport` now provides an override.

mastersin:
2017-09-27 22:45 GMT+03:00 Olly Betts <notifications@github.com>:

> This breaks Python 2.4 (syntax error from the inline if:
> https://travis-ci.org/swig/swig/jobs/280310758). While older Python isn't
> very interesting now, it'd be better to just use a multi-line if and keep
> this working in case anyone cares, at least if this change is going to be
> backported to 3.0.x.
>
This is looks like a strange bug for python-2.4 due there are a check for
python version in original code - "if _swig_python_version_info >= (2, 7,
0):":

diff --git a/swig/Source/Modules/python.cxx b/swig/Source/Modules/python.cxx
index a6801fc..7aa08ba 100644
--- a/swig/Source/Modules/python.cxx
+++ b/swig/Source/Modules/python.cxx
@@ -846,7 +846,8 @@ public:
       Printv(default_import_code, "if _swig_python_version_info >= (2, 7,
0):\n", NULL);
       Printv(default_import_code, tab4, "def swig_import_helper():\n",
NULL);
       Printv(default_import_code, tab8, "import importlib\n", NULL);
-      Printv(default_import_code, tab8, "pkg =
__name__.rpartition('.')[0]\n", NULL);
+      Printv(default_import_code, tab8, "pkg_parts =
__name__.rpartition('.')\n", NULL);
+      Printv(default_import_code, tab8, "pkg = pkg_parts[0] if
pkg_parts[1] == '.' else pkg_parts[2]\n", NULL);
       Printf(default_import_code, tab8 "mname = '.'.join((pkg,
'%s')).lstrip('.')\n", module);
       Printv(default_import_code, tab8, "try:\n", NULL);
       Printv(default_import_code, tab8, tab4, "return
importlib.import_module(mname)\n", NULL);


> Python 3 also failed on OS X, but due to travis flaky infrastructure
> reasons - I've restarted so we can see if there are any actual issues there
> (it passed on Linux, so likely not).
>
> @wsfulton <https://github.com/wsfulton> In #769
> <https://github.com/swig/swig/issues/769> you concluded:
>
> Okay, let's not try support the 'new' requirement. If anything, let's just
> make the documentation clearer in this respect.
>
> What are your thoughts on this patch? It seems clean and very unlikely to
> break anything, but we've had bad experiences with changing this import
> logic and %pythonimport now provides an override.
>

-- 
Sin (Sinelnikov Evgeny)

ojwb:
> This is looks like a strange bug for python-2.4 due there are a check for python version in original code - "if _swig_python_version_info >= (2, 7, 0):":

It's a syntax error - Python presumably needs to be able to parse the code inside that branch even if though it's not executed.  For example, with Python 3.5 you get a syntax error for a non-executed `print` without parentheses:

```
$ python3.5
Python 3.5.4 (default, Sep  5 2017, 18:32:10) 
[GCC 7.2.0] on linux
>>> if False:
...     print "foo"
  File "<stdin>", line 2
    print "foo"
              ^
SyntaxError: Missing parentheses in call to 'print'
```

And it's easily worked around - just use a multi-line `if`.

wsfulton:
I've merged in #1282 which seems to fix #769. @mastersin, can you please close this issue when confirmed that it works for you? Thanks.

wsfulton:
No response, assuming fixed.

neinhus:
From my point of view it could lead to problems to do this:
        try:
            return importlib.import_module(mname)
        except ImportError:
            return importlib.import_module('_FSNativeTauIO')
Reason: the import error may be also that a depending shared library is not found. The error message thrown is "No module named _<name>"  but in my case the true error message was:  
ImportError: lib<somename>.so.<nr>: cannot open shared object file.
This can be difficult to debug without looking into the wrapper code.

wsfulton:
@neinhus, the various issues linked to here have been fixed for swig-4.0.0. There is no try except block anymore for the module import.

PR #1114, opened 2017-10-06
using double for long integers in R

fschlimb:
R does not support long integers (64bit), so there is no fast way to fully support them.
This uses doubles to represent long integers in R and converts them to/from 64bit integers on the C/C++ side. It issues warning if input long integer cannot be represented precisely (e.g. > 2^DBL_MANT_DIG).

@joequant

### DIFF
--- a/Lib/r/r.swg
+++ b/Lib/r/r.swg
@@ -60,13 +60,7 @@ SWIG_InitializeModule(0);
 	     unsigned int *, unsigned int[ANY],
              short *, short[ANY],
              signed short *, signed short[ANY],
-             unsigned short *, unsigned short[ANY],
-             long *, long[ANY],
-             signed long *, signed long[ANY],
-             unsigned long *, unsigned long[ANY],
-             long long *, long long[ANY],
-             signed long long *, signed long long[ANY],
-             unsigned long long *, unsigned long long[ANY]
+             unsigned short *, unsigned short[ANY]
              
 {
 { int _rswigi;
@@ -79,7 +73,13 @@ SWIG_InitializeModule(0);
 } 
 
 %typemap(in) float *, float[ANY],
-             double *, double[ANY]
+             double *, double[ANY],
+             long *, long[ANY],
+             signed long *, signed long[ANY],
+             unsigned long *, unsigned long[ANY],
+             long long *, long long[ANY],
+             signed long long *, signed long long[ANY],
+             unsigned long long *, unsigned long long[ANY]
              
 {
 {  int _rswigi;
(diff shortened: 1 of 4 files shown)


ojwb:
CI failures.  One's travis being flaky (which I restarted), but the R builds fail with what looks like a missing header (`DBL_MANT_DIG` is not defined).

fschlimb:
Seems travis is still flaky...

ojwb:
I've restarted the ones which flaked, but the R failure seems genuine:

```
checking r testcase arrays_dimensionless (with run test)
cat: ../../../../../Examples/test-suite/r/arrays_dimensionless_runme.Rout: No such file or directory
make[1]: *** [arrays_dimensionless.cpptest] Error 1
```

Also, octave failures look real too.

ojwb:
Oh, I see octave failures on git master too, so they're probably unrelated.

fschlimb:
I guess there are more things this PR is missing. Will continue when I have more time.

fschlimb:
The last failure comes from copyToR:
```
Error in (function (cl, name, valueClass)  : 
  assignment of an object of class "numeric" is not valid for @'ui' in an object of class "A"; is(value, "integer") is not TRUE
Calls: copyToR -> copyToR -> <Anonymous>
```
I am not sure I really understand how this works. The R code causing this is probably this:
```
if(!isGeneric("copyToR")) 
 setGeneric("copyToR",
            function(value, obj = new(gsub("Ref$", "", class(value)))) 
               standardGeneric("copyToR"
           ))
```
How or why this gets to an "numeric" I don't know. The get method in the wrapper clearly creates an integer:
```
SWIGEXPORT SEXP
R_swig_A_ui_get ( SEXP self, SEXP s_swig_copy)
{
[...]
  result = (unsigned int) ((arg1)->ui);
  r_ans = SWIG_From_int((int)(result));
  
  vmaxset(r_vmax);
  if(r_nprotect)  Rf_unprotect(r_nprotect);
  
  return r_ans;
}
```

@joequant: any idea?

fschlimb:
ok. Apparently the conversions from/to int were redirected to the ones for 'long'. Not sure why, but adding explicit ones for 'int' seems to do the right thing.

fschlimb:
the failures don't seem related to SWIG at all.

fschlimb:
Will you merge this or do you see need for anything else?

wsfulton:
C 'long' is not always 64-bits, it is very much platform specific. SWIG makes an assumption that long is 32-bits. While this is not always correct, there is no correct assumption! However, an assumption is required in order to wrap a 'long' and for consistency this assumption is consistent across all the target languages. 

SWIG is designed so that if this sort of assumption is not suitable for a user, typemaps can be used to change the way it works. In this case you'd provide overridden typemaps for long in your own code. This is in theory. If you can't wrap 'long' as if it was a double, for example, by using `%apply double { long};` then please raise a separate issue/patch.

PR #1115, opened 2017-10-06
Namespace quirks

fschlimb:
Classes with the same name in different namespace have not been treated correctly by SWIG.
The changes here resolve most of the issues related to this, including combining it with the use of "using namespace".
Previous attempts to get this into master as separate PRs failed as the different commits are only work together correctly and I don't want to spend time with figuring out what would be a correct order of separate PRs.

Related PRs are #810 and #742.

### DIFF
--- a/Source/Swig/symbol.c
+++ b/Source/Swig/symbol.c
@@ -391,8 +391,12 @@ void Swig_symbol_add_using(String * name, String * uname, Node * n) {
   h = Swig_symbol_clookup(uname,0);
   if (h && checkAttribute(h, "kind", "class")) {
     String *qcurrent = Swig_symbol_qualifiedscopename(0);
-    Append(qcurrent, "::");
-    Append(qcurrent, name);
+    if(qcurrent) {
+        Append(qcurrent, "::");
+        Append(qcurrent, name);
+    } else {
+        qcurrent = NewString(name);
+    }
     Setattr(symtabs, qcurrent, n);
     Delete(qcurrent);
   }
--- a/Source/Swig/symbol.c
+++ b/Source/Swig/symbol.c
@@ -391,11 +391,11 @@ void Swig_symbol_add_using(String * name, String * uname, Node * n) {
   h = Swig_symbol_clookup(uname,0);
   if (h && checkAttribute(h, "kind", "class")) {
     String *qcurrent = Swig_symbol_qualifiedscopename(0);
-    if(qcurrent) {
-        Append(qcurrent, "::");
-        Append(qcurrent, name);
+    if (qcurrent) {
+	Append(qcurrent, "::");
+	Append(qcurrent, name);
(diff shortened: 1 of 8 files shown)


ojwb:
There are some unnecessary whitespace changes in the patch, such as blank lines being removed and indentation changes on code lines you haven't otherwise changed  - please avoid these, as they make review harder, and also make tools like `git blame` harder to use to dig back into code history.

Some of the coding style is wrong, for example in `Source/CParse/templ.c`:

```
if( tmp ) {
```

This doesn't match the code elsewhere in that file (and in most of SWIG), which puts a space after `if` and no spaces between parentheses and what they contain, i.e.:

```
if (tmp) {
```

Also, your tab settings seem to be wrong for SWIG code, resulting in mis-indented code when viewed with the appropriate tab settings.  A tab character should be set to move to the next multiple of 8 column.

fschlimb:
Some of the failures are unrelated to SWIG, others seem unrelated to my changes.
Shall I issue a "fake" commit or can Travis be triggered explicitly without a commit?

wsfulton:
I've restarted the one job that was due to a network failure. The remaining failed jobs are real failures due to the patch (I think). I don't recognise them. Do they pass on your machine?

ojwb:
@fschlimb Did you see William's comment?

We can't merge a patch which causes tests to start failing.

fschlimb:
@ojwb sorry for the delay. I do not test go and R on my machine. The failures don't really look related to my patch (like in R it complains strangely about ```object 'true' not found```) and errors occur only in R and go. I can try to find some time this or next week to look into this. 
Is there anyone familiar with the R and go front-ends? Maybe a quick look gives them an immediate thought of what's happening...

Source/Swig/symbol.c
```diff
@@ -382,6 +382,26 @@ String *Swig_symbol_qualified_language_scopename(Symtab *n) {
   return result;
 }
 
+/* -----------------------------------------------------------------------------
+ * Swig_add_usingclass()()
```
ojwb: This comment doesn't match the name of the function actually added.

fschlimb:
I spent some time digging into this.

I finally stranded in ```SwigType_typedef_resolve_all``` (Swig/stype.c:1187). When called by the member-variable resolution (for the only member variable ```Manta::ColorSpace<(Manta::RGBTraits)>::data```)  it finds ```Traits::ComponentType``` as the type. This is incorrect because ```Traits``` is the template argument-name (from the template-class spec) and not the fully resolved basic data-type.

I suspect this particular case doesn't pose a problem for Python (for example) because its set-method accepts a ```PyObject``` (and not a C++ type) and maybe the fully resolved type is never needed elsewhere.

Any idea how to fix this?

fschlimb:
superseded by #1484

PR #1119, opened 2017-10-07
Fix constants involved parenthesised char literals

ojwb:
Fixes https://sourceforge.net/p/swig/bugs/1168/


ojwb:
Only test failure is unrelated to these changes, so merging.

Issue #405, opened 2015-05-04
[Node.js] Crash with method that contains return value and reference parameters


vicnet:
Code generated for a method that returns a value (int for example) and reference parameters crash in `SWIGV8_AppendOutput` of the first by-reference parameter.
It is due to the creation of the tuplet:
jsresult is built with:
`jsresult = SWIG_From_int((int)(result));`
and return tupplet is filled by the first by-reference parameter as:
`jsresult = SWIGV8_AppendOutput(jsresult, SWIG_From_long((*arg2)));`
that try to add an element in a `V8::Array` that is not an array but an integer value initially constructed.

PR #1121, opened 2017-10-08
Fix SWIGV8_AppendOutput

alecmev:
See #405.

I'm new to SWIG and V8, so, please, do review.


ojwb:
Thanks for the patch - it seems very plausible, and doesn't break any existing tests.

It'd be good to have a regression test though - `Examples/test-suite/li_typemaps.i ` looks like it has a suitable function for this if I understand the context:

```
struct Foo *out_foo(int a, int *OUTPUT) {
  struct Foo *f = new struct Foo();
  f->a = a;
  *OUTPUT = a * 2;
  return f;
}
```

So I think you'd just need to add `Examples/test-suite/javascript/li_typemaps_runme.js` (see the other `*_runme.js` in that directory) which tries to call the wrapped version of this function - before the patch that should explode, but with the patch it should work.

PR #1123, opened 2017-10-10
Update pyhead.swg

rafmav:
#define is for defining constants
for renaming a type, use typedef !

### DIFF
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -211,5 +211,5 @@ typedef destructor freefunc;
 #if PY_VERSION_HEX < 0x03020000
 #define PyDescr_TYPE(x) (((PyDescrObject *)(x))->d_type)
 #define PyDescr_NAME(x) (((PyDescrObject *)(x))->d_name)
-#define Py_hash_t long
+typedef long Py_hash_t
 #endif


ojwb:
Your patch is syntactically invalid.  Please actually test patches compile before opening a PR.

rafmav:
Thanks.
No sandbox, sad!

Its compiles here.
or need [-fpermissive]
...

rafmav:
This is what I have while building something using Swig:
 ==> Starting build()...  
[ 25%] Building CXX object
CMakeFiles/_ca_smoothing.dir/lib/ca_smoothingPYTHON_wrap.cxx.o
~/workspace/github/python-casmoothing/src/python-casmoothing/lib/ca_smoothingPYTHON_wrap.cxx:979:19:
error: declaration does not declare anything [-fpermissive] #define
Py_hash_t long ^ /usr/include/vtk/vtkPythonCompatibility.h:81:14: note:
in expansion of macro 'Py_hash_t' typedef long Py_hash_t; ^~~~~~~~~
make[2]: *** [CMakeFiles/_ca_smoothing.dir/build.make:72:
CMakeFiles/_ca_smoothing.dir/lib/ca_smoothingPYTHON_wrap.cxx.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:68:CMakeFiles/_ca_smoothing.dir/all]
Error 2 make: *** [Makefile:152:all] Error 2 ==> ERROR: A failure
occurred in build(). Aborting...

a "grep" on the swig repository gives me Lib/python/pyhead.swg
for Py_hash_t

I did this:
remove:
#define Py_hash_t long
(declaration does not declare anything !)
(AND it is already defined in vtkPythonCompatibility.h)
I wrote in place:
#ifndef Py_hash_t
typedef long Py_hash_t;
#endif
thus simplified on patch in...
typedef long Py_hash_t;

So what to do ?

ojwb:
> Thanks. No sandbox, sad!

If you by "sandbox" you mean CI builds, then there are CI builds, and they failed for your patch.

> I did this: remove: #define Py_hash_t long (declaration does not declare anything !) (AND it is already defined in vtkPythonCompatibility.h) I wrote in place: #ifndef Py_hash_t typedef long Py_hash_t; #endif thus simplified on patch in... typedef long Py_hash_t; So what to do ?

But the patch you submitted isn't what you describe here - the patch you submtted has no trailing semicolon on the `typedef` which means it's syntactically invalid.

Moral of the story - use git to commit the changes you tested to a branch and push them, then open a PR from the branch.  Avoid github's online editor (for code changes at least - for a typo in a text file I guess it's OK).

Also, it sounds like the `'#ifndef` you tested with is needed in your situation as otherwise vtk's define will turn your new code into `typedef long long`.  In fact the `#ifndef` is the most important part - whether it's a `typedef` or a `#define` makes no real difference.

rafmav:
Thanks.

Le Fri, 13 Oct 2017 05:13:12 +0000 (UTC),
Olly Betts <notifications@github.com> a écrit :

> > Thanks. No sandbox, sad!  
> 
> If you by "sandbox" you mean CI builds, then there are CI builds, and
> they failed for your patch.
> 
sandbox: learning tool or place for kids.
(french: "sand" : "sable", "sandbox : "bac à sable": place where young
children play and learn...)

> > I did this: remove: #define Py_hash_t long (declaration does not
> > declare anything !) (AND it is already defined in
> > vtkPythonCompatibility.h) I wrote in place: #ifndef Py_hash_t
> > typedef long Py_hash_t; #endif thus simplified on patch in...
> > typedef long Py_hash_t; So what to do ?  
> 
> But the patch you submitted isn't what you describe here - the patch
> you submtted has no trailing semicolon on the `typedef` which means
> it's syntactically invalid.
> 
Hanging typo sorry.
> Moral of the story - use git to commit the changes you tested to a
> branch and push them, then open a PR from the branch.  Avoid github's
> online editor (for code changes at least - for a typo in a text file
> I guess it's OK).
Agree: Github interface I already do not understand how it works!
> 
> Also, it sounds like the `'#ifndef` you tested with is needed in your
> situation as otherwise vtk's define will turn your new code into
> `typedef long long`.  In fact the `#ifndef` is the most important
> part - whether it's a `typedef` or a `#define` makes no real
> difference.
> 
I agree.

But, usually, at least before, "typedef" is / was for naming / renaming
a type, and "#define" for naming constants, extended for formula, that
are constants minus the algebra values in. Concept different for
typedef that defines a type, so a place / range in memory.

There is a difference while creating the program, if I remember:
#define is for the precompiler, all symbolics constants are replaced by
there values; on error, hard to find where; after, the compiler itself
take the hand, and does the "typedef", replacing too the symbolic value
by the type, "long" here, but I believe I remember in case of bug in
execution, it can be retrieve, that is not the case with a "#define",
but with a "#define", it won't compile... so it is best a problem of
where and when we want to track errors: while compiling or while
running...

Here, compiler stops saying "#define Py_hash_t long" does
nothing, and won't compile without the option "fpermissive", which is a
bad idea: to miss errors with "-fpermissive".
I said the problem was in "Swig", in reality there is not really one,
just the precompiler find a second time a definition of a
#define... that is already in "vtk" but perhaps the problem is also in
"vtk" that too : a "#define" instead of a "typedef"...

ojwb:
If some header you're including is defining `Py_hash_t` differently to SWIG - e.g. `#define Py_hash_t int` - then changing SWIG as you propose isn't going to help if that header is included first - SWIG's `typedef long Py_hash_t;` will become `typedef long int;` after preprocessor substitution, and that won't compile.

If the header defines `Py_hash_t` the same way SWIG does (i.e. `#define Py_hash_t long`) then that should currently compile OK (at least with GCC a repeat `#define` to exactly the same value is accepted without even a warning), but with your proposed change we'd get `typedef long long;` which similarly won't compile, so your change actually seems to make the situation worse.

Changing both vtk and SWIG to use `typedef` wouldn't help if they typedef it to different types.  It should work if they define it the same way, but that should already work as things are.

I'm not sure what a better fix is.  `#ifdef Py_hash_t` / `#undef Py_hash_t` / `#endif` before doing a `typedef long Py_hash_t;` would help for headers which also `#define Py_hash_t` but not for headers which have a conflicting typedef.

I'm going to close this PR - the patch as-is here doesn't even compile and there's not been any attempt to update it in 3 months.  I could apply it with the syntax error fixed, but as above that seems to make things worse.

Feel free to open a new PR if you have a fully thought through and tested patch to propose.

PR #1128, opened 2017-10-19
[C#] Add ToArray to std_vector.i

Liryna:


wsfulton:
Thanks, I've just merged it in.

degasus:
I think there might be a bug in this code. For a vector of vector of int, this generates:
```
  public int[][] ToArray() {
    int[][] array = new int[][this.Count];
    this.CopyTo(array);
    return array;
  }
```
int[][this.Count] is clearly wrong, but I don't see an automatical way to fix it.

wsfulton:
Presumably the problem is something like:
```swig
%template(VectorVectorInt) std::vector<std::vector<int>>;
```
Shouldn't the AllocateArray function be private? Why don't you put together a pull request and include a test case?

PR #1135, opened 2017-10-24
go: generate more idiomatic output (drop unnecessary return names)

tmc:
This change omits the AFAICT unused and unneeded _swig_ret named return values which is more idiomatic.

### DIFF
--- a/Source/Modules/go.cxx
+++ b/Source/Modules/go.cxx
@@ -1204,12 +1204,12 @@ private:
     // Write out the result type.
     if (info->is_constructor) {
       String *cl = exportedName(class_name);
-      Printv(parm_print, " (_swig_ret ", cl, ")", NULL);
+      Printv(parm_print, " ", cl, NULL);
       Delete(cl);
     } else {
       if (SwigType_type(info->result) != T_VOID) {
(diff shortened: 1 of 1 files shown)


Source/Modules/go.cxx
```diff
@@ -1822,15 +1822,15 @@ class GO:public Language {
     // Write out the result type.
     if (is_constructor) {
       String *cl = exportedName(class_name);
-      Printv(parm_print, " (_swig_ret ", cl, ")", NULL);
+      Printv(parm_print, " ", cl, "", NULL);
       if (first == NULL) {
 	first = NewString("_swig_ret");
```
ianlancetaylor: Note how if there are no parameters, only results, first will be set to _swig_ret, and then referenced below.  This change can't work in that case.

tmc: I see. Thanks for the review. This was admittedly a bit of a naive attempt.

ojwb: @tmc Are you planning to revise this patch, or should we just close it?

tmc: I’ll close it.

PR #1146, opened 2017-11-08
Pull master into Matlab

CameronDevine:
To fix an issue that I was having (That later turned out to be my own mistake) I merged the SWIG master branch into the Matlab branch. Since I did this work I thought I would file a pull request to make someone's life easier in the future.


ojwb:
Thanks for the PR, but it's causing a lot of the CI builds to fail.  I checked 3 of them and they were all due to the same issue:

```
../../../../../Examples/test-suite/python/../nested_structs.i:3: Error: Syntax error in input(1).
```

(Note, that's not a python-specific file due to the `..` in the path).

CameronDevine:
I fixed the problem which was previously causing the builds to fail. However many of the Python based builds failed because of the error:

````
WARNING: The following packages cannot be authenticated!
  libpython3.5-minimal libpython3.5-stdlib libpython3.5 python3.5-minimal
  libpython3.5-dev python3.5 python3.5-dev
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
The command "sudo apt-get -qq install python3.5-dev" failed. Retrying, 3 of 3.
````

This appears to be an issue with the continuous integration. What is the best way to fix this?

wsfulton:
Regarding the Travis failures, the very latest master is needed to work as one of our 3rd party repositories for Python (deadsnakes) was moved.

Matlab development has moved over to https://github.com/jaeandersson/swig/ and I think it has a recent pull from swig master. I don't want to conflict with their work and so won't pull this in. I suggest you use that instead. My hope is that this will eventually make its way back into the official SWIG repository. I think the matlab developers there have a bit more work to do.

PR #1150, opened 2017-11-13
Lua 5.3 integer support

Fsu0413:
Lua 5.3 added an integer type.
The original number type is now treated as real type.

e.g.:
```c
// push the number to Lua, in variable x
lua_pushnumber(L, (lua_Number)24);
```

```lua
-- x always is type of number, in Lua 5.3 it is real type
-- 24 is type of integer in Lua 5.3, in Lua 5.2 it is type of number
if x == 24 then
    print("Yes!")
else
    print("No!")
end
```

outputs "No!". Because x is now 24.0, not 24.
In Lua 5.2 and earlier this program should output "Yes." because both x and 24 is type of number.

This change breaks a lot of programs including swig-lua interface.
We can't use Lua 5.3 or later before this pr got merged to swig.

### DIFF
--- a/Lib/lua/lua.swg
+++ b/Lib/lua/lua.swg
@@ -20,8 +20,14 @@
 // this basically adds to a table of constants
 %typemap(consttab) int, unsigned int, short, unsigned short, long, unsigned long, unsigned char, signed char, bool, enum SWIGTYPE
        {SWIG_LUA_CONSTTAB_INT("$symname", $value)}
+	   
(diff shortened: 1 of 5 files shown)


ojwb:
All the Lua tests are failing in CI with:

```
../../../../../Lib/lua/lua.swg:179: Error: Syntax error in input(1).
```

Please run the test suite locally (`make check-lua-test-suite`) to catch such errors **before** pushing.  It'll save you a lot of time too.

ojwb:
@Fsu0413 Did you see my comment above about this breaking the Lua testsuite?

Fsu0413:
Sorry...I am not currently working on this issue.
Please wait until I found a way to solve it.

I'm closing this issue by now.

PR #1162, opened 2017-11-27
Incorporate template argument namespace

k3m3:
When declaring a templated type, the namespace of the template argument is ignored and the wrapper class is generated without the correct package declaration and placed in the root directory instead of the corresponding package directory. 

Currently (see example below), the following structure would be generated:

* foo
  * bar
    * TestyMcTestface.java
* baz
  * ter
    * TestyMcTestface.java
* FooTestyList.java
* BazTestyList.java

with the patch the correct structure is generated
* foo
  * bar
    * TestyMcTestface.java
    * FooTestyList.java
* baz
  * ter
    * TestyMcTestface.java
    * BazTestyList.java

Furhtermore, without the patch FooTestyList and BazTestyList only have the package declaration 'package testy;' whereas it should be 'package testy.foo.bar;' and 'package testy.baz.ter;'. 

Example: 
--- test.i --- 
%module test

%include <std_list.i>

%nspace foo::bar::TestyMcTestface;
%template(FooTestyList) std::list&lt;foo::bar::TestyMcTestface&gt;;

%nspace baz::ter::TestyMcTestface;
%template(BazTestyList) std::list&lt;baz::ter::TestyMcTestface&gt;;

%include "../test.hpp"
--- test.hpp ---

namespace foo { namespace bar {

class TestyMcTestface { };

} }

namespace baz { namespace ter {

class TestyMcTestface { };

} }
--- command --- 
swig -outcurrentdir -c++ -java -package testy test.i
--- END ---

### DIFF
--- a/Source/Modules/typepass.cxx
+++ b/Source/Modules/typepass.cxx
@@ -463,6 +463,16 @@ class TypePass:private Dispatcher {
 	  SwigType_typedef_class(fname);
 	  scopename = Copy(fname);
 	}
+        // Incorportate the template argument namespace, so the wrapper class package declaration and location is correct
+        SwigType* tan = SwigType_templateargs(name);
+        Replace(tan, "<(", "", DOH_REPLACE_ANY);
+        Replace(tan, ")>", "", DOH_REPLACE_ANY);
+        SwigType* ansname = Swig_scopename_prefix(tan);
+        Replace(ansname, "::", ".", DOH_REPLACE_ANY);
+        Setattr(n, "sym:nspace", ansname);
+        Delete(tan);
+        Delete(ansname);
+
 	Delete(deftype_name);
 	Delete(resolved_name);
       } else {


wsfulton:
I think you may have misunderstood what %nspace does. It recreates the C++ namespaces (in Java equivalent, that is packages) for the types you specify.

In your example, `std::list<baz::ter::TestyMcTestface>` does not have %nspace attached to it, and so it should remain in the 'root' as indeed it does. If you were to add:
```
%nspace std::list<baz::ter::TestyMcTestface>;
```
then you correctly get std/BazTestyList.java because `std::list<baz::ter::TestyMcTestface>` instantiates an instance of `list<baz::ter::TestyMcTestface>` in the std namespace and the name you have given the instantiation for Java is BazTestyList.

There is no easy way that I can think of to make something in the std namespace appear in the baz::ter namespace other than in pure C++, for example, using alias template. %nspace doesn't support moving types into a different namespace as far as the wrappers are concerned, it merely mirrors the C++ namespace structure.

k3m3:
Thanks for your prompt feedback. 
This seems to be a problem for us, since the generated java code does not compile due to CPtr's protected access modifier which is inaccessible to the proxy class because it is not in the same (sub)package/namespace where it should be. Exposing the CPtr public would be a very undesired option. 
It does not make much sense, at least to me ( no offense meant ), to have templated types in the root or their container namespace rather than their specialization types namespaces, since a typedef or derivation of the specialization would not make them part of the root or std namespace either.

```
namespace foo { namespace bar {

typedef std::list<baz::ter::TestMcTestface> FooTestyList;
// or 
class FooTestyList : public std::list<baz::ter::TestMcTestface>

} } 
```

wsfulton:
Sorry, I don't have a magic bullet for your requirements given the combined limitations in SWIG and Java. You could go back to not using %nspace / avoid subpackages and just use %rename any classes that have the same name in multiple namespaces so that they don't conflict. Otherwise, you'd need to come up with a proposal to enhance %nspace to do something more than just mirror the C++ namespaces. But I suggest you put together a proposal with lots of different use cases as to how it would work and get agreement on that before implementing it.

k3m3:
I will (re)evaluate if it is a viable option to go with %rename in a global namespace, but since the project to be wrapped has a tight structure and heavily relies on the namespaces I doubt it. 
Should I close this and open a new one if I come up with something to propose or leave it open ? 
Anyway, thank you for your time, its highly appreciated.

wsfulton:
Closing. Yes, please open a new issue if you put together a full proposal.

PR #1164, opened 2017-11-29
Add via file support

JPEWdev:
Arguments may be passed using a via file. The file is specified by
prefixing the file name with a '@' (e.g. '@args.txt'). All arguments in
the file (separated by whitespace) will be inserted into the argument
list in place of the via file. If the file does not exist, the argument
is passed as is.

This allows passing long arguments that may exceed the maximum
command line length

### DIFF
--- a/Doc/Manual/SWIG.html
+++ b/Doc/Manual/SWIG.html
@@ -163,6 +163,32 @@ can be obtained by typing <tt>swig -help</tt> or <tt>swig
 
 </pre></div>
 
+<p>
+Arguments may also be passed in a command-line options file (also known as a
+response file) which is useful if they exceed the system command line length
+limit. To do this, put the arguments in a file, then provide the file name
+prefixed with <tt>@</tt> like so:
+</p>
+
+<div class="shell"><pre>
+swig @<em>file</em>
+</pre></div>
+
+<p>
+The options read from the file are inserted in place of the file option. If the
+file does not exist, or cannot be read, then the option will be treated
+literally and not removed.
+</p>
+
+<p>
+Options in the file are separated by whitespace. A whitespace character may be
+included in an option by surrounding the entire option in either single or
+double quotes. Any character (including a backslash) may be included by
+prefixing the character to be included with a backslash. The file may itself
+contain additional <tt>@file</tt> options; any such options will be processed
+recursively.
+</p>
+
 <H3><a name="SWIG_nn3">5.1.1 Input format</a></H3>
 
 
--- a/Source/Modules/swigmain.cxx
+++ b/Source/Modules/swigmain.cxx
@@ -125,7 +125,7 @@ void SWIG_merge_envopt(const char *env, int oargc, char *oargv[], int *nargc, ch
 
   int argc = 1;
   int arge = oargc + 1024;
-  char **argv = (char **) malloc(sizeof(char *) * (arge));
+  char **argv = (char **) malloc(sizeof(char *) * (arge + 1));
   char *buffer = (char *) malloc(2048);
   char *b = buffer;
   char *be = b + 1023;
@@ -147,6 +147,7 @@ void SWIG_merge_envopt(const char *env, int oargc, char *oargv[], int *nargc, ch
   for (int i = 1; (i < oargc) && (argc < arge); ++i, ++argc) {
     argv[argc] = oargv[i];
   }
+  argv[argc] = NULL;
 
   *nargc = argc;
   *nargv = argv;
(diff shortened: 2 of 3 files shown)


wsfulton:
The gcc man page describes the equivalent functionality:

>	   @file
>	   Read command-line options from file.  The options read are inserted in place of the
>	   original @file option.  If file does not exist, or cannot be read, then the option will
>	   be treated literally, and not removed.
>
>	   Options in file are separated by whitespace.  A whitespace character may be included in
>	   an option by surrounding the entire option in either single or double quotes.  Any
>	   character (including a backslash) may be included by prefixing the character to be
>	   included with a backslash.  The file may itself contain additional @file options; any
>	   such options will be processed recursively.

Does the patch support all of this in the same way?

The Microsoft Windows developer tools (like visual c++) seem to call these 'response files'. Unless someone suggests another commonly known name for these, can you change the implementation to use this terminology and update the 'Basics' chapter (SWIG.html) "Running SWIG" section with this new functionality?

JPEWdev:
I borrowed the idea from the gcc syntax. I will update it to behave in the same way, rename it 'response files' (I think I borrowed 'via files' from the ARM compiler or something), and include it in the documentation.

Thanks for the feedback

ojwb:
While I actually do have a vague recollection that MSVC uses the term "response file", I'm not sure it is a helpful one to use here - this term doesn't seem to be used in the Unix world, it doesn't in itself explain that it's a file of command line options, and someone confronted with an unknown term might try a web search but googling for `"response file"` doesn't seem to return anything helpful - I get a load of stuff like "A response file is an XML file that contains data required to complete installation operations silently" (where the name makes a lot more sense - that's a file of canned responses).

GCC's docs don't really seem to name the concept, but describe it as "Read command-line options from file", so what about something self-descriptive like "command-line options file"?

Incidentally, I notice that GCC explicitly documents that `@file` works inside such a file but MSVC explicitly documents that it doesn't, so we can't be exactly compatible with both.

wsfulton:
'response file' is used quite consistently in the Microsoft documentation, google for 'msdn response file'. The gcc implementation was done after Microsoft did it and in the unix world there doesn't seem to be any identifiable name, so it seems reasonable to use, what seems to me, to be the original naming. 'command-line options file' is a much more obvious name, but we should add something like 'aka response file' into the documentation as well.

@JPEWdev am I correct in interpreting that your implementation adheres to the gcc implementation, noting @ojwb's comment that it is different to that in Visual C++? The docs you added seem to be a copy of the gcc man page I quoted earlier.

JPEWdev:
Yes. AFAIK this behaves identical to the gcc implementation.

JPEWdev:
Documentation updated

JPEWdev:
Fixed a bug where argv wasn't terminated with a NULL pointer. This should resolve the ci failures (and I think was a pre-existing bug if arguments were passed using SWIG_FEATURES).

JPEWdev:
Any update on this?

Doc/Manual/SWIG.html
```diff
@@ -163,6 +163,32 @@ <H2><a name="SWIG_nn2">5.1 Running SWIG</a></H2>
 
 </pre></div>
 
+<p>
+Arguments may also be passed in an command-line options file (also known as a
```
wsfulton: typo, should be 'a command-line options file'.

Source/Modules/swigmain.cxx
```diff
@@ -147,11 +148,90 @@ void SWIG_merge_envopt(const char *env, int oargc, char *oargv[], int *nargc, ch
...
+
+  new_argv[index] = (char *)malloc(option_len + 1);
+  memcpy(new_argv[index], start, option_len);
+  new_argv[index][option_len] = '\0';
+
+  *argc = new_argc;
+  *argv = new_argv;
+}
+
+void SWIG_merge_options_files(int *argc, char ***argv) {
```
wsfulton: This should be a static function without the `SWIG_` prefix. `SWIG_merge_envopt` was there before but is also wrong and ought to be corrected.

Source/Modules/swigmain.cxx
```diff
@@ -147,11 +148,90 @@ void SWIG_merge_envopt(const char *env, int oargc, char *oargv[], int *nargc, ch
...
+  char **new_argv = *argv;
+  int new_argc = *argc;
+  FILE *f;
+
+  i = 1;
+  while (i < new_argc) {
+    if (new_argv[i] && new_argv[i][0] == '@' && (f = fopen(&new_argv[i][1], "r"))) {
+      char c;
+      char* b;
+      char* be = &buffer[BUFFER_SIZE];
```
wsfulton: SWIG coding convention requires `char *b' and `char *be`.

wsfulton:
Thanks, I've merged into master.

PR #1166, opened 2017-11-30
Swig C Preprocessor: fix handling of // style comments inside macro arguments.

llongi:
Hi, SWIG currently fails with:
Error: Macro 'PACKED_STRUCT' expects 1 argument

on files that have macro arguments that contain // style comments, for example as in:
https://github.com/inilabs/libcaer/blob/master/include/events/polarity.h#L45

This is perfectly valid C syntax.
In commit 624ec3e1b7dd8908b37ee86bce6f60423c915bc5 related to swig/swig#974 this was fixed for /* ... */ style comments inside macro arguments, but the // style comments were not considered.
This commit completes that fix by also handling // style comments and discarding the comment part of the line.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,31 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.0 (in progress)
 ===========================
 
+2018-01-16: wsfulton
+            Expressions following a preprocessor directive must now be separated by whitespace
+            or non-numeric characters. This syntax change makes the SWIG preprocessor work like
+            the C preprocessor in this area.
+
+            For example, the following code used be accepted as valid syntax:
+              #if1
+              #define ABC 123
+              #endif
+
+            Now you get an error:
+              example.h:1: Error: Unknown SWIG preprocessor directive: if1 (if this is a block of
+              target language code, delimit it with %{ and %})
+              example.h:3: Error: Extraneous #endif.
+
+            The following is the correct syntax:
+              #if 1
+              #define ABC 123
+              #endif
+
+            The following of course also works:
+              #if(1)
+              #define ABC 123
+              #endif
+
 2018-01-15: wsfulton
             Fix issue #1183. Floating point exception evaluating preprocessor expressions
             resulting in division by zero.
--- a/Examples/test-suite/errors/pp_expressions_bad.i
+++ b/Examples/test-suite/errors/pp_expressions_bad.i
@@ -24,3 +24,20 @@
 #if 0
 #elif 8.8
 #endif
+
+/* Missing whitespace after preproc directive */
+#if123
+#endif
+
+#if456e
+#endif
+
+#if 0
+#warning This should not warn
+#elif1
+#warning This should also not warn
+#endif
+
+#if(1)
+#warning Warning okay: #if(1)
+#endif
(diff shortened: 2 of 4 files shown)


ojwb:
It'd be good to have a regression test to check this now works and continues to work in the future.

wsfulton:
What's a minimal example that this fixes please so that we can add it to our test-suite and accept the patch?

ojwb:
The example the reporter linked to from the original report seems to work fine with git master.  However, an example with a comma in the comment fails, so here's a test suite patch:

```
--- a/Examples/test-suite/preproc.i
+++ b/Examples/test-suite/preproc.i
@@ -389,6 +389,20 @@ TCX_PACKED (typedef struct tcxMessageBugImpl
 {
     int mBid; /**< Bid price and size, check PresentMap if available in message */
 }) tcxMessageBug;
+
+
+TCX_PACKED (typedef struct tcxMessageTestImplSlashSlash
+{
+    int mHeader; ///< comment
+}) tcxMessageTestSlashSlash;
+
+
+TCX_PACKED (typedef struct tcxMessageBugImplSlashSlash
+{
+    int mBid; ///< Bid price and size, check PresentMap if available in message
+}) tcxMessageBugSlashSlash;
+
+
 %}
 
 // Regression tests for https://github.com/swig/swig/pull/1111
```

The first case already works, the second case results in:

```
./../preproc.i:398: Error: Macro 'TCX_PACKED' expects 1 argument
./../preproc.i:410: Error: Syntax error in input(1).
```

And the patch here fixes that.

@llongi Are you actually seeing this fail for C++ comments which don't contain a comma?  If so, it would be really good to have a reduced testcase for such a situation.

llongi:
Indeed a minimal example is something like this:

```
#define SOME_MACRO(STRUCT_DECLARATION) STRUCT_DECLARATION

SOME_MACRO(
struct some_struct {
	// A C style comment, with a comma.
	int data1;
	// A C style comment with a parenthesis ) this time.
	int data2;
});
```

I have to admit I didn't even realize the comma had anything to do with it, I found the other bug report about the multi-line /* ... */ comments and ignoring them and noticed the same thing was not done for // ... comments and fixed that.
As far as I can see the fact that a //.. comment without a comma in it even actually parses is some unintended side-effect in the parser, as the block of code for ignoring comments (line 610) just does that for /*...*/ comments, else it assumes the rest of the line is NOT a comment and parses it as usual, where it then hits the comma and errors out. Same thing happens if you put for example a ')' in the comment (as above). Basically without my patch it just tries to parse the comment parts of an argument as if they were real code, and errors out if they don't look enough like it. That is conceptually wrong, so what my patch does, identify the comment and ignore it altogether, seems like the cleaner approach.

ojwb:
I think that as-is the comment gets copied across to the output when the macro is expanded, and then gets ignored later as a comment.  That may mean there are some corner cases we get wrong, though I'm struggling to construct an example (it seems to be hard to end up with the character sequence `/*` after preprocessor comment removal).

I've pushed a commit with the test coverage improvement above to the PR branch.  I also took the liberty of fixing the formatting of your commit message - the first line should ideally be no more than about 50 characters, and be followed by a blank line - otherwise commands like `git log --oneline` flatten the whole first paragraph onto one line and the result is an unreadable mess.

Not sure why github auto-closed the PR as a result - will reopen.

ojwb:
Hmm, seems I can't reopen.  And now I can't force push the old branch tip back to the PR branch to undo the previous force push (I get "permission denied" - I guess because the PR is now closed).  Stupid github.

I've instead rebased the changes onto current master and pushed that.  I was hoping to wait to check that the CI was green first, but I don't foresee any problems.

ojwb:
One slight blip thanks to scilab 5 not coping with symbols longer than 24 bytes.  I'm not set up to test scilab locally, and it's not clear from the error exactly what the constructed symbol name that is too long actually is, but I've renamed the new class to a shorter name and have pushed so hopefully that'll do the trick.

PR #1178, opened 2017-12-27
Forward declaration within typedef when names are identical.

### LINKED ISSUE #838: Incorrect generated code when tag name and typedef name are identical.

davedissian:
This pull request fixes #838 (sorry it took so long to re-do this PR).

I determined that `Language::classLookup` seems to fail when there exists a `cdecl` in the symbol table where the name is identical to another struct/class in the symbol table. This situation is similar to the case:

```
typedef struct Foo {
   ....
} Foo;
```

This is handled as a special case, as documented in symbol.c:

```
 * Due to the unified namespace for structures, special handling is performed for
 * the following:
 *
 *        typedef struct Foo {
 *
 *        } Foo;
 * 
 * In this case, the symbol table contains an entry for the structure itself.  The
 * typedef is left out of the symbol table.
```

This PR adds a similar workaround for the case:

```
typedef struct Foo Foo;
```

By replacing the broken `cdecl` node in the symbol table with the correct `classforward` node during parsing, we fix this issue nicely without having to add a more brittle hack to `Language::classLookup` and other places to work around this particular edge case.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,11 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.0 (in progress)
 ===========================
 
+2017-12-30: davedissian
+            Fixed a symbol lookup issue when encountering a typedef of a symbol from the tag
+            namespace to the global namespace when the names are identical, such as 'typedef
+            struct Foo Foo;'.
+
 2017-12-13: wsfulton
             [Perl] add missing support for directorfree typemaps.
 
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -490,6 +490,7 @@ CPP_TEST_CASES += \
 	throw_exception \
 	typedef_array_member \
 	typedef_class \
+	typedef_classforward_same_name \
 	typedef_funcptr \
 	typedef_inherit \
 	typedef_mptr \
@@ -676,6 +677,7 @@ C_TEST_CASES += \
 	string_simple \
 	struct_rename \
 	struct_initialization \
+	typedef_classforward_same_name \
 	typedef_struct \
 	typemap_subst \
 	union_parameter \
--- a/Examples/test-suite/java/typedef_classforward_same_name_runme.java
+++ b/Examples/test-suite/java/typedef_classforward_same_name_runme.java
@@ -1,0 +1,22 @@
+
+import typedef_classforward_same_name.*;
+
+public class typedef_classforward_same_name_runme {
+
+  static {
+    try {
+	    System.loadLibrary("typedef_classforward_same_name");
+    } catch (UnsatisfiedLinkError e) {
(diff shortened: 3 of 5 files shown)


davedissian:
Looks like tests are passing on Travis CI in general however a couple of configurations are broken regardless of my changes.

wsfulton:
The Travis testing is broken for many Python versions as the deadsnakes Ubuntu ppa we are using for the different Python versions is no longer available. I'll commit a fix for that soon. Meanwhile, ignore the Python failures.

wsfulton:
I'm not entirely comfortable with this patch which messes around with the parse tree and replaces a typedef with a forward declaration for this particular case. I've noted that similar special case `typedef struct Foo { ... } Foo;` you highlight is similar. Also, the very similar example to the one you are solving:
```
void function(Test* test);
struct Test {
      int foo;
};
typedef struct Test Test;
```
does work in SWIG unpatched. This is because `classLookup` finds the class/struct declaration instead of the typedef when calling `Swig_symbol_clookup`. The returned symbol from this function is actually a linked list and `classLookup` currently only looks at the first element in the list. The following patch is another way to fix this bug as it looks for the class node in the entire linked list:
```patch
diff --git a/Source/Modules/lang.cxx b/Source/Modules/lang.cxx
index 1e4a6bd..9a215b0 100644
--- a/Source/Modules/lang.cxx
+++ b/Source/Modules/lang.cxx
@@ -3296,6 +3296,14 @@ Node *Language::classLookup(const SwigType *s) const {
        break;
       if (Strcmp(nodeType(n), "class") == 0)
        break;
+      Node *sibling = n;
+      while (sibling) {
+       sibling = Getattr(sibling, "csym:nextSibling");
+       if (sibling && Strcmp(nodeType(sibling), "class") == 0)
+         break;
+      }
+      if (sibling)
+       break;
       n = parentNode(n);
       if (!n)
        break;
```
This seems less of a hack and more of a proper fix. However, a similar change is probably needed in other places. Could you try out this patch to see if it works in your real-life code base?

davedissian:
Thanks for your response. I was unsure whether this kind of patch would be acceptable as it is a bit of a hack, but I figured I'd create a PR anyway for discussion.

I tried your patch and it works great, and in hindsight it makes much more sense to implement the fix in that way - I missed the detail that the symbol table contains a linked list when the symbol names are the same. I applied your patch to my PR and updated my branch (removing the previous changes) with the additional test.

PR #1201, opened 2018-02-19
Updated tarball search to match convention

gr3mlinz:
pcre tarball naming convention has been updated on ftp://ftp.csx.cam.ac.uk/pub/software/programming/pcre/ causing the script to not find tarballs downloaded straight from the website.


gr3mlinz:
Mistake: PCRE dep not same as PCRE2 dep

PR #1213, opened 2018-03-21
Fix ruby %alias directive for native c functions

goatshriek:
Using the %alias directive on native C functions causes swig to segfault due to
a dereference of klass (which is NULL for native C functions) in the
defineAliases function of the Ruby module. This commit adds support for an alias
of native C functions for both separate module as well as global functions, as
well as three test cases for the %alias directive of the Ruby module.

Fixes:
mod.i
%module ruby_alias
%alias get_my_name "nickname,fullname";
%inline %{
const char *get_my_name(){
  return "Chester Tester";
}
%}
$ swig -ruby mod.i
Segmentation fault

Signed-off-by: Joel Anderson <joelanderson333@gmail.com>

### DIFF
--- a/Examples/test-suite/ruby/Makefile.in
+++ b/Examples/test-suite/ruby/Makefile.in
@@ -22,6 +22,7 @@ CPP_TEST_CASES = \
 	li_std_stack \
 	li_std_wstring \
 	primitive_types \
+	ruby_alias_method \
 	ruby_keywords \
 	ruby_minherit_shared_ptr \
 	ruby_naming \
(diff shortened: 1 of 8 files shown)


goatshriek:
I'll add a fix for the older versions of Ruby that are failing in Travis CI in a day or so.

goatshriek:
The test errors on versions of Ruby prior to 2.3.0 have been resolved: the original_name function did not exist in these, so a direct comparison of Method objects allowed programs to test for an alias being set to a given method.

wsfulton:
Thanks, especially for the comprehensive tests.

Harrypotter19876:
Wat

PR #1222, opened 2018-04-10
* Makefile.in (configfiles): Update URLs for latest configfiles.

bje-:
We haven't uploaded the latest version of these scripts to `ftp.gnu.org` in over 10 years. See <http://ftp.gnu.org/pub/gnu/config/README> for details. The best place to fetch them from is the Savannah git repository.


ojwb:
Thanks.  Also much better to be fetching them over https rather than unencrypted ftp.

PR #1223, opened 2018-04-10
python.cxx: Insert some additional newlines to suppress some PEP8 warnings

bje-:
This overcomes a couple of easily rectified warnings from the `example.i` file:

```
example.py:51:1: E302 expected 2 blank lines, found 1
example.py:90:1: E305 expected 2 blank lines after class or function definition, found 1
```

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -926,7 +926,7 @@ public:
       {
 	// Python-2.2 object hack
 	Printv(f_shadow,
-	       "\n", "def _swig_setattr_nondynamic(self, class_type, name, value, static=1):\n",
+	       "\n\n", "def _swig_setattr_nondynamic(self, class_type, name, value, static=1):\n",
 	       tab4, "if (name == \"thisown\"):\n", tab8, "return self.this.own(value)\n",
 	       tab4, "if (name == \"this\"):\n", tab8, "if type(value).__name__ == 'SwigPyObject':\n", tab4, tab8, "self.__dict__[name] = value\n",
 #ifdef USE_THISOWN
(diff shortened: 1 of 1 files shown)


vadz:
Is there any provision for disabling specific warnings from the file itself? I think it might be preferable to do it like this if possible, tweaking generated code to avoid code style warnings looks like a rather useless task to me and historically doing this created a few problems with all the noise due to insignificant changes in the code. So personally I'd rather keep the SWIG code simpler and the generated code smaller and just disable the warnings.

bje-:
You have convinced me of the error of my ways. ;-)  It is possible to exclude machine generated files with `flake8`, so I will. One counterargument is that it prevents people (and SWIG developers) from checking the quality of the generated files, removing deprecated idioms, etc. It is useful to rigorously check auto-generated script code, particularly when there is no compilation step.

ojwb:
FWIW, we already do PEP8 checks in the testsuite by running `pycodestyle` (if installed) with options `--ignore=E402,E501,E30,W291,W391`, so we are checking for many such things already (and that's run from CI).

It looks like `E30` in that list means `E30*` (I don't see this in the man page, but if I remove it I get a flood of `E301`, `E302` and `E305`), so it seems these pep8 violations are actually deliberately being allowed currently.

I'd lean towards fixing the blank lines to match pep8 myself, especially as the noise from that change is easy to filter out (`diff -B`) so the downside seems small.  And generally we accept patches to fix compiler warnings in the generated code, which seems a fairly similar situation.

If we do we should drop `E30` from that list though, to make sure that the change does what it's meant to and doesn't regress in future.

bje-:
That is correct: `E30` means `E30*`. I don't think it's documented.

vadz:
I'm aware of PEP8 checks the testsuite does precisely because I remember spending quite some time on making them pass with the code on the Doxygen branch (and then re-fixing them many times). I just don't see any real usefulness in having them. The comparison with compiler warnings doesn't seem valid to me because:

1. It's not that simple to disable the warnings for all compilers, while this is trivial to do with pep8.
1. A warning might indicate code which is actually incorrect, while pep8 is only about purely stylistic issues (AFAIK?) that it doesn't really make any sense to worry about in the generated code.

bje-:
PEP8 is a style guide, but `flake8` will find syntax errors, too.

vadz:
Syntax errors would definitely be worth checking for.

PR #1224, opened 2018-04-10
Some typos in configure

Dragonway:
Fix some typos in configure.ac help messages, add default value for uninitialized arg


Dragonway:
Looks like my changes can't break anything but python py3 doesn't pass checks (just notice).

ojwb:
Thanks.

PR #1227, opened 2018-04-13
%pythonappend docs and memory management example

jakecobb:
This is PR #929 with comments on original addressed and rebased onto the current master for a cleaner merge.


wsfulton:
This is useful info for the Python docs, thanks. 2315ed878b0cdac1d47853a7c526cacf2a83c23f makes some additions to the section added.

PR #1230, opened 2018-04-16
Fix off-by-one error

gjanssens:
The condition only applies to guile 1.8 and older so it should really
have been '<2' all along. As we already have such a conditional block
earlier up, merge the two together as well.


ojwb:
Some CI builds are still pending, but the guile build passed and this doesn't touch generic code so will merge.

BTW, if you're interested in fixing guile stuff, the macOS build is showing some issues which look like they're at least in part due to using removed APIs (`scm_listify`):

https://travis-ci.org/swig/swig/jobs/361492415

Issue #312, opened 2015-01-20
SWIG generates deprecated symbols for guile 2.0 and up


Giels:
Symbols such as scm_sym2var, or scm_module_lookup_closure and so on have been deprecated and have been removed from guile 2.1. As of version 3.0.4, SWIG still generates these symbols for guile bindings.

ojwb:
I don't believe SWIG currently has an active guile maintainer, so if you want to see changes to the guile backend, you probably need to come up with a patch, and ensure the test suite passes after applying it with currently used versions of guile.

The last major guile contribution was adding support for guile 2.0 in April 2013 by @gjanssens, so you might want to see if Geert is interested in collaborating on this.

ojwb:
Thanks for the analysis - indeed the only use (in `Lib/guile/guile_scm_run.swg`) is suitably protected:

```
#if defined (SCM_MAJOR_VERSION) && (SCM_MAJOR_VERSION < 2)

static SCM
scm_module_variable (SCM module, SCM sym)
{
  return scm_sym2var (sym, scm_module_lookup_closure (module), SCM_BOOL_F);
}
```

So I'm going to close this as there's no actual bug in this report.

@Giels - if you are seeing deprecated symbols which are used in a non-protected context in git master, please reopen.

> I don't know when swig plans its next release (configure is now primed for 4.0). These changes are worth passing to a wider audience though. Will there still be 3.0 maintenance releases ? If so, please do include my last two PR's.

I'm not sure what a realistic date for 4.0.0 is.  I don't think there's currently a plan for a 3.0.13, though it's going to be problematic if the gap extends too long.  Problem is I'm not sure we really have the developer time to usefully progress towards 4.0.0 and maintain 3.0.x in parallel.

PR #1232, opened 2018-04-18
Fix guile 2.2 support

gjanssens:
There were a few issues with this. These commits should take care of it. The macOS test with guile (which is using guile 2.2) succeeds now.
Running the tests locally on my Fedora 27 system (also using guile 2.2) works as well.

Note this is probably not a complete removal of all deprecated symbols as discussed in issue #312. It's sufficient to pass the test suite and also to work for gnucash (my main project).

**Edit** I just checked in more detail and I believe this PR does fix all of the known uses of deprecated symbols.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1704,13 +1704,20 @@ else
   fi
   if test -n "$GUILE_CONFIG" ; then
     if test x"$GUILE" = xyes; then
-      AC_MSG_CHECKING([for guile bindir])
-      guile_bindir="`$GUILE_CONFIG info bindir`"
-      AC_MSG_RESULT([$guile_bindir])
-      GUILE=$guile_bindir/guile
-      if ! test -f "$GUILE" ; then
-	GUILE=
-        AC_PATH_PROG(GUILE, guile)
+      AC_MSG_CHECKING([for guile executable])
+      if $GUILE_CONFIG info guile > /dev/null 2>&1; then
+        GUILE="`$GUILE_CONFIG info guile`"
+        AC_MSG_RESULT([$GUILE])
+      else
+        AC_MSG_RESULT([not in guile-config, constructing path])
+        AC_MSG_CHECKING([for guile bindir])
+        guile_bindir="`$GUILE_CONFIG info bindir`"
+        AC_MSG_RESULT([$guile_bindir])
+        GUILE=$guile_bindir/guile
+        if ! test -f "$GUILE" ; then
+          GUILE=
+          AC_PATH_PROG(GUILE, guile)
+        fi
       fi
     fi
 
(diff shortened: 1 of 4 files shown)


ojwb:
Thanks - this is great except it breaks the guile Linux CI build.

It looks like configure now finds an empty string for the guile executable: https://travis-ci.org/swig/swig/jobs/368163893#L657

gjanssens:
Hmm, I missed that :(

I have updated the PR to handle it properly. The guile linux CI build is already happy. Still waiting for it to complete all the way until guile OS X...

configure.ac
```diff
@@ -1705,44 +1705,62 @@ else
...
-    if test -z "$GUILE_CFLAGS" ; then
-      AC_MSG_CHECKING([for guile compile flags])
-      GUILE_CFLAGS="`$GUILE_CONFIG compile`" # Note that this can sometimes be empty
-      AC_MSG_RESULT([$GUILE_CFLAGS])
+    if test -f "$GUILE" ; then
+      # Test if guile-config and guile versions match. They should.
+      gc_version="`$GUILE_CONFIG --version 2>&1 | head -1`"
+      gc_version="${gc_version##* }"
+      g_version="`$GUILE --version | head -1`"
+      g_version="${g_version##* }"
```
ojwb: We don't currently rely on `${foo##bar}` anywhere in SWIG.

It is specified by POSIX, but `info autoconf` warns:

```
'${#VAR}'
'${VAR%WORD}'
'${VAR%%WORD}'
'${VAR#WORD}'
'${VAR##WORD}'
     Posix requires support for these usages, but they do not work with
     many traditional shells, e.g., Solaris 10 '/bin/sh'.
```

So probably prudent to use sed instead.  You can do the `head -1` at the same time, so it's no more external commands to run:

```
      gc_version="`$GUILE_CONFIG --version 2>&1 | sed 's/.* //;q'`"
      g_version="`$GUILE --version | sed 's/.* //;q'`"
```
```

File: configure.ac [hunk 2]
```diff
@@ -1705,44 +1705,62 @@ else
...
+      fi
+      if ! test -f "$GUILE" ; then
+        GUILE=
+        AC_PATH_PROG(GUILE, guile)
       fi
     fi
 
-    if test -f "$GUILE" ; then
+    if ! test -f "$GUILE" ; then
+      AC_MSG_WARN([no suitable guile executable found. Disabling Guile])
```
ojwb: It seems that `$GUILE` being empty means it's disabled, which matches how other similar variables work and generally seems to make sense.

However we also seem to have a second disabled state where `$GUILE` is non-empty but not a valid file path.  I think it would be cleaner to set `GUILE=` here and then replace `test -f "$GUILE"` below with `test -z "$GUILE"`.

configure.ac
gjanssens: Good point. I have changed the relevant commit.

configure.ac
gjanssens: Although I believe both ways of testing are correct I agree with your observation. Using the same test all the time makes the code much more readable.

gjanssens:
I have also rebased my branch on current master.

Issue #945, opened 2017-03-16
std::vector garbage collection allows use after free


jakecobb:
I noticed this in Python, haven't test other languages yet.  The `std::vector` support (from `std_vector.i`) for value (non-pointer) types allows the vector to be garbage collected while the managed layer still holds a pointer to one of its values.  Here is a SWIG example:

```
// test.i
%module test
%include "std_vector.i"
%include "std_string.i"

%inline %{
#include <vector>
#include <string>

struct Value {
  std::string name;
  Value(const std::string& name) : name(name) { }	
};

std::vector<Value> someValues() {
  std::vector<Value> values;
  values.emplace_back("test1");
  return values;
}
%}

%template(Values) std::vector<Value>;

```

In Python, you get some behavior like:
```python
>>> import test
>>> vals = test.someValues()
>>> v0 = vals[0]
>>> v0.name
'test1'
>>> del vals
>>> v0.name
''
```

Of course in my real use case the value has a `std::unique_ptr` leading to a crash on use after free.  An even shorter example of the problem when the `std::vector` is a temporary:

```python
>>> v = test.someValues()[0]
>>> v.name
''
```

It seems that either some back-reference is necessary to keep the vector from being freed while proxies to its values are alive or the index operator should be making a copy that the managed layer owns.

jakecobb:
Here is a possible workaround:

```
%fragment("container_owner_attribute", "header") {
  PyObject* container_owner_attribute() {
    static PyObject* name = SWIG_Python_str_FromChar("__owner");
    return name;
  }
}

%define %StdVector(_name, _type)
// override the default SWIGTYPE out typemap
%typemap(out, noblock=1, fragment="container_owner_attribute") _type *, _type &, _type &&, _type[] {
  %set_output(SWIG_NewPointerObj(%as_voidptr($1), $descriptor, $owner | %newpointer_flags));
  // set a reference to the owner
  PyObject_SetAttr($result, container_owner_attribute(), $self);
}
// now declare the template with the above typemap operating
%template(_name) std::vector<_type>;
// remove the typemap to revert to the default typemap
%clear _type *, _type &, _type &&, _type[];
%enddef

%StdVector(Values, Value);
```

Of course, this only helps for Python and assumes there is no pre-existing typemap on `_type` that needs to be preserved.  It does prevent the use-after-free unless the user explicitly deletes `__owner` from the `std::vector` wrapper object.

[...]

PR #1234, opened 2018-04-19
Python container memory handling fix

jakecobb:
Fixes Issue #945 

For Python container handling (std::vector and friends), prevent a use-after-free due to the index operator returning a reference.  Whenever we are returning a reference from the container (wrapped as an unowned-pointer in the default case), set an attribute pointing to the owning container so it won't be garbage collected while the script user holds the child element reference.  If the reference wrapper is pointer-owning or not a SWIG pointer wrapper then we assume a copy was made and do nothing.

### DIFF
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -49,7 +49,7 @@ namespace swig {
     static PyObject* attr = SWIG_Python_str_FromChar("__owner");
     SwigPyObject* swigThis = SWIG_Python_GetSwigThis(child);
     if (swigThis && (swigThis->own & SWIG_POINTER_OWN) != SWIG_POINTER_OWN) {
-      PyObject_GenericSetAttr(child, attr, owner);
+      PyObject_SetAttr(child, attr, owner);
       return true;
     }
     return false;
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -36,8 +36,17 @@
 
 %include <std_except.i>
 
-%fragment("reference_container_owner", "header") {
+%fragment("container_owner_attribute_init", "init") {
+  // thread safe initialization
+  swig::container_owner_attribute();
+}
+%fragment("reference_container_owner", "header", fragment="container_owner_attribute_init") {
 namespace swig {
+  PyObject* container_owner_attribute() {
+    static PyObject* attr = SWIG_Python_str_FromChar("__owner");
+    return attr;
+  }
+
   /**
    * Call to add a back-reference to the owning object when returning a 
    * reference from a container.  Will only set the reference if <code>child</code> 
@@ -46,10 +55,9 @@ namespace swig {
    * @return if the reference was set or not
    */
   bool reference_container_owner(PyObject* child, PyObject* owner) {
-    static PyObject* attr = SWIG_Python_str_FromChar("__owner");
     SwigPyObject* swigThis = SWIG_Python_GetSwigThis(child);
     if (swigThis && (swigThis->own & SWIG_POINTER_OWN) != SWIG_POINTER_OWN) {
-      PyObject_SetAttr(child, attr, owner);
+      PyObject_SetAttr(child, container_owner_attribute(), owner);
       return true;
     }
     return false;
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -48,28 +48,32 @@ namespace swig {
     return attr;
   }
 
-  // By default, do not add the back-reference (for value types)
-  // Specialization below will check the reference for pointer types.
-  template <typename Type>
-  bool reference_container_owner(PyObject* child, PyObject* owner)
-  { return false; }
+  template <typename T>
+  struct container_owner {
+    // By default, do not add the back-reference (for value types)
+    // Specialization below will check the reference for pointer types.
+    static bool reference(PyObject* child, PyObject* owner)
+    { return false; }
+  };
 
-  /**
-   * Call to add a back-reference to the owning object when returning a 
-   * reference from a container.  Will only set the reference if <code>child</code> 
-   * is a SWIG wrapper object that does not own the pointer.
-   *
-   * @return if the reference was set or not
-   */
   template <>
-  bool reference_container_owner<swig::pointer_category>(PyObject* child, PyObject* owner) {
-    SwigPyObject* swigThis = SWIG_Python_GetSwigThis(child);
-    if (swigThis && (swigThis->own & SWIG_POINTER_OWN) != SWIG_POINTER_OWN) {
-      PyObject_SetAttr(child, container_owner_attribute(), owner);
-      return true;
+  struct container_owner<swig::pointer_category> {  
+    /**
+     * Call to add a back-reference to the owning object when returning a 
+     * reference from a container.  Will only set the reference if <code>child</code> 
+     * is a SWIG wrapper object that does not own the pointer.
+     *
+     * @return if the reference was set or not
+     */
+    static bool reference(PyObject* child, PyObject* owner) {
+      SwigPyObject* swigThis = SWIG_Python_GetSwigThis(child);
+      if (swigThis && (swigThis->own & SWIG_POINTER_OWN) != SWIG_POINTER_OWN) {
+        PyObject_SetAttr(child, container_owner_attribute(), owner);
+        return true;
+      }
+      return false;
     }
-    return false;
-  }
+  };
 }
 }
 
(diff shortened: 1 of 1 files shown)


jakecobb:
This test was already failing before my changes so its failure is expected:

```
checking python testcase overload_numeric (with run test)
  File "overload_numeric_runme.py", line 37, in <module>
    check(nums.over(float("inf")), "float")
  File "overload_numeric_runme.py", line 11, in check
    raise RuntimeError("got: " + got + " expected: " + expected)
RuntimeError: got: double expected: float
make[1]: *** [overload_numeric.cpptest] Error 1
```

Any other failure in the Python test suite is not expected.

ojwb:
> This test was already failing before my changes so its failure is expected:

While that's not relevant to the proposed change, this failure is odd as it is passing in CI and also locally for me (using Python 2.7).  What Python version and platform are you seeing this failure with?

jakecobb:
System-provided Python 2.7.10 on OS X Sierra on a 2015 Macbook Pro:

```
$ system_profiler SPSoftwareDataType | grep Version
      System Version: macOS 10.12.5 (16F73)
      Kernel Version: Darwin 16.6.0
$ python --version
Python 2.7.10
$ which python
/usr/bin/python
```

ojwb:
The travis CI Python on macOS job passes - that is also running "python" which is 2.7.10:

https://travis-ci.org/swig/swig/jobs/368423500

I can't see anywhere which clearly shows if that's actually `/usr/bin/python` or `python` in some directory which is before that in `PATH`, but the location of the header files for it seem to indicate it is indeed the system version:
```
checking for Python header files... -I/System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7 -I/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config
```

The log shows it's a slightly newer OS version, though it seems surprising if that makes a difference:

```
ProductVersion:	10.12.6
```

jakecobb:
I checked against the 3.0.12 release tag (`rel-3.0.12`) and it fails for me there too, not just against `master`.   Should I open an issue to discuss this further since it is not related to this PR?

I have a newer clang than shown in that build log (I upgraded it vs what Xcode 8.1 shipped with):

```
$ clang --version
Apple LLVM version 9.0.0 (clang-900.0.39.2)
Target: x86_64-apple-darwin16.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Running `make` actually uses the `gcc` and `g++` aliases for clang, those show the same version but with this line:

```
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
```

Python is coming from the same location:

```
checking for Python header files... -I/System/Library/Frameworks/Python.framework/Versions/2.7/include/python2.7 -I/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config
checking for Python library directory... /System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/config
```

My hunch is it's related to this in the wrapper:

```cpp
/* Accept infinite as a valid float value unless we are unable to check if a value is finite */
#ifdef SWIG_isfinite
# define SWIG_Float_Overflow_Check(X) ((X < -FLT_MAX || X > FLT_MAX) && SWIG_isfinite(X))
#else
# define SWIG_Float_Overflow_Check(X) ((X < -FLT_MAX || X > FLT_MAX))
#endif
```

Because the failing test line is passing infinity:

```python
check(nums.over(float("inf")), "float")
```

wsfulton:
overload_numeric failure is fixed now, see #1239.

wsfulton:
> Can you explain the std::vector<int> case a bit? If the reference is being added, then the C++ int& is being wrapped by Swig as a non-owned pointer isn't it? Or is the run-time check alone adding the overhead?

Actually, it is **const** int& in these methods of std::vector (value_type is int for `std::vector<int>`:
```
  const value_type& __getitem__(difference_type i) const throw (std::out_of_range);
  const value_type& front() const;
  const value_type& back() const;
```
The typemaps for `const int&` _effectively_ wrap the integer 'by value' via a call to PyInt_FromLong. There is no SWIGTYPE in this case as the return is not being wrapped as a pointer. Hence a back-reference is NOT required. The runtime checking that the returned PyObject* is a SWIGTYPE is an overhead and not necessary. The fundamental concept of SWIG is designed around typemaps to control the generated code per type, and for type `const int&` the back-reference should not be generated.

> It took me quite a while to figure out where to stick that typemap so it worked properly. 

I'm not surprised, this exactly what I meant by "_hard to fathom what is going on given all the macro trickery, typemaps and fragment trickery_"

> Can you give any hints about how the type traits are maintained that I would need to implement suggestion 2, maybe an example of where they are used by the existing code?

Not easily, for reason stated above. Best if you run swig -E to look at the pre-processed code resulting from including std_vector.i and using %template as it is much easier to understand as you'll then see the type traits. Then try and find the sources given in the Lib/stl, Lib/typemaps, Lib/python directories. Start with Lib/std_vector.i. Hopefully you understand the concept of C++ type traits. If not, read around the subject, it is important in modern C++.

jakecobb:
So it looks like `swig::traits` tracks the relevant type information, using `swig::pointer_category` and `swig::value_category` as tags.  I've updated to use template specialization around this to only do the SWIG wrapper check on `pointer_category` objects.  I tested this against `std::vector<int>`, `std::vector<std::string>` (with `std_string.i` included), and a self-defined struct `std::vector<Foo>`.

Assuming this state is acceptable, are you happy with the multiple commits in the PR or do you prefer it be rebased down to a single commit?

wsfulton:
No need to squash the commits, unless you really want to.

Nearly there, I think. We need a template class for the specializations and not a template function. Have a read of http://www.gotw.ca/publications/mill17.htm. With the final implementation, ideally we should have an example of how a user can remove the addition of the back-reference. Maybe too much too ask, but would be good to know. This area of SWIG is so complex, I really don't like it!

jakecobb:
I'm not sure the function specialization matters because there is no overloading here and the template type parameter does not even participate in the argument list types, but it's an easy enough change so I went ahead and made it.  I suppose tag dispatch to overloaded functions without templates would also have worked.

I don't think the user should remove the reference because it's a memory safety issue.  However, I'm not sure exactly how they could if they wanted to.  They would need to stop the `ret` typemap from firing, but only in the context of containers to avoid blowing away any other `ret` typemaps that might be in affect for references to the same type.  My patch limits scope by placing the typemap inside the class definition and by using the `typedef`'d `value_type` for matching instead of the true type name.  But, I can't seem to override this from my `.i` file, here is an attempt:

```
%module swigtest

%include "std_string.i"
%include "std_vector.i"
%include "std_container.i"
%include "typemaps.i"

%inline {
  struct Foo {};
}

%typemap(ret) std::vector<Foo>::value_type&, Foo& {
  // remove the back reference?
}

%template(IntVector) std::vector<int>;
%template(FooVector) std::vector<Foo>;
%template(StringVector) std::vector<std::string>;

%typemap(ret) std::vector<Foo>::value_type&, Foo& {
  // remove the back reference?
}
```

The comment there never appears in the output and the back-reference is still added.

wsfulton:
I made a couple of minor changes, see 62417708fdb63c7eff0f2921adae898475d2f980 a628bf9b6f5bf79cc28cfa09dbb76279b4d97f77 be9d736597e359480dd98d05612e5dbc4747dd6d 9038a9987dc1440422a60aab8b943e42568be973. The second one shows how to solve overriding the typemaps (should a user want to, and I think it is important that a user is able to).  You'd override the typemap using one or other of the 'ret' typemaps below:
```
%module example

#if 0
%typemap(ret) const std::vector<Foo>::value_type&
#else
%typemap(ret) const std::vector<Foo>::value_type& std::vector<Foo>::__getitem__
#endif
{
  // remove the back reference
}

%include "std_string.i"
%include "std_vector.i"

%inline {
  struct Foo {
    int val;
    Foo(int val = 0) : val(val) {}
  };
}

%template(IntVector) std::vector<int>;
%template(FooVector) std::vector<Foo>;
%template(StringVector) std::vector<std::string>;
```
In order to understand why, run `swig -debug-tmsearch` and you'll see the typemap matching search:
```
../../../Lib/std/std_vector.i:91: Searching for a suitable 'ret' typemap for: std::vector< Foo >::value_type const &std::vector< Foo >::__getitem__
  Looking for: std::vector< Foo >::value_type const &std::vector< Foo >::__getitem__
  Looking for: std::vector< Foo >::value_type const &__getitem__
  Looking for: std::vector< Foo >::value_type const &
  Looking for: std::vector< Foo >::value_type &std::vector< Foo >::__getitem__
  Looking for: std::vector< Foo >::value_type &__getitem__
  Using: %typemap(ret) std::vector< Foo >::value_type &__getitem__
```
The one used above is the one now supplied by SWIG. Any of the preceding matches will take precedence, hence one of the two 'ret' typemaps given above will work. See http://swig.org/Doc4.0/Typemaps.html for more info.

More generically, you can override the typemap with this instead for all types:
```
%extend std::vector {
%typemap(ret) const value_type& __getitem__
{
  // remove the back reference
}
}
```
For completeness, overrides for `const value_type& front` and `const value_type& back` should also be added in.

Issue #1206, opened 2018-03-03
Linux: configure doesn't accept go version 1.10


woguls:
I have installed the currently "featured version" of go version 1.10, via https://golang.org/dl/
When generating the configure script I get:

```
checking for go... go
checking whether go version is too old... yes - minimum version is 1.2
```

For the moment I am able to manually change the configure script to accept go1.10 and everything seems to work.

ojwb:
Looks like the pattern in `configure.ac` doesn't expect two digit minor versions:

```
   case $go_version in
    go1.1* | go1.0* | go1 )
      AC_MSG_RESULT([yes - minimum version is 1.2])
```

That needs fixing, but should only mean that you can't run SWIG's testsuite for Go - you should still get a SWIG build that can generate Go bindings.

ojwb:
Fixed by e57d845312131320dc5be91f97f693f0a10c548e.

PR #1235, opened 2018-04-19
Fix go version matching in configure for go1.10

Rasie1:
`configure`'s version matching doesn't know about golang's version naming convention and sees go1.10 as older version. I fixed that and replaced matching strings to identify go1.10 as latest version and succeed and generating test suite.
See issue https://github.com/swig/swig/issues/1206


wsfulton:
Looks okay to me.

PR #1240, opened 2018-04-23
[Java] Suppress deprecation warning on finalize method

### LINKED ISSUE #1237: Deprecation warning when compiling finalize method in Java 9
I'm using Swig 3.0.12, I've recently updated my java installation:
```shell
$ java --version
openjdk 9.0.4
OpenJDK Runtime Environment (build 9.0.4+12-Debian-4)
OpenJDK 64-Bit Server VM (build 9.0.4+12-Debian-4, mixed mode)
```
and I get the following output when compiling the Java code generated by Swig:
```
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
```
```
...
/some/java/class/generated/by/swig/MyClass.java:44: warning: [deprecation] finalize() in Object has been deprecated
  protected void finalize() {
                 ^
...
Note: Some input files additionally use or override a deprecated API.
100 warnings
```

Manually adding `@SuppressWarnings("deprecation")` to all the `finalize` methods seems to work. Unfortunately this is not a solution as this is code generated by swig.

I don't know what's the best approach for swig here (generate the suppression annotation or find a replacement for `finalize`) as I'm not a Java expert.

In the mean time, does one see a workaround that could allow me to silence these warnings without having to manually patch the generated code each time I call swig?

sjoubert:
Java 9 deprecates the finalize method.
For now just suppress the deprecation warning.
Fixes #1237

### DIFF
--- a/Examples/test-suite/java_director.i
+++ b/Examples/test-suite/java_director.i
@@ -7,6 +7,7 @@
 %module(directors="1") java_director
 
 %typemap(javafinalize) SWIGTYPE %{
+  @SuppressWarnings("deprecation")
   protected void finalize() {
 //    System.out.println("Finalizing " + this);
     delete();
--- a/Examples/test-suite/java_throws.i
+++ b/Examples/test-suite/java_throws.i
@@ -192,6 +192,7 @@ try {
 
 // Need to handle the checked exception in NoExceptTest.delete()
 %typemap(javafinalize) SWIGTYPE %{
+  @SuppressWarnings("deprecation")
   protected void finalize() {
     try {
       delete();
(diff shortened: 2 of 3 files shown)


ojwb:
@wsfulton This looks like an easy win, but I don't know Java well so I'm hesitant to apply it without your OK.

wsfulton:
I'm in two minds about this. While this patch is helpful to make this kind of noise go away, it is masks a problem that needs addressing. I'd like to see a proper jdk9 solution using some of the suggestions in #1237 implemented as alternative typemaps. Depending on what comes up will determine whether we should just mask this problem or put in what will be a radically different way of managing wrapped C++ object from Java. While I don't have the time to investigate this right now, we should aim to either use this patch or provide a better solution before we release 4.0, so let's put this on hold until shortly before releasing 4.0. I've just tagged this patch with the swig-4.0 milestone, so it doesn't get overlooked.

PR #1242, opened 2018-04-25
Fixes #1225

### LINKED ISSUE #1225: Swig clutters Python-API with _swigregister objects
# Problem
When I create a Python wrapper for a C++ library (no -builtin flag) the resulting lib.py contains for each class (`CLASS`) something like this:

    CLASS_swigregister = _lib.CLASS_swigregister
    CLASS_swigregister(CLASS)

The extra CLASS_swigregister member of the API is annoying since users of that library do not necessarily know or care that the library was created with swig (of which I am a fan since years). Users see it in an IDE and do not know what to do.

# Possible solution

I have changed my rather large API with some simple regex to the following:

    _lib.CLASS_swigregister(CLASS)

This works without a problem (for me at least). Can SWIG create this single line as standard behaviour? Or is there any need for `CLASS_swigregister` residing in the public API?

philippkraft:
Python wrapper does no longer contain PythonClass_swigregister methods that aree not intended for users. The _swigregister call uses the internal lib_wrap.cxx interface.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -4632,8 +4632,8 @@ public:
 		   class_name);
 	  }
 	}
-	Printf(f_shadow_file, "%s_swigregister = %s.%s_swigregister\n", class_name, module, class_name);
-	Printf(f_shadow_file, "%s_swigregister(%s)\n", class_name, class_name);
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks.

PR #1246, opened 2018-04-28
Remove superfluous parens in generated Python scripts.

bje-:
The generated Python code contains some superfluous parentheses. This change removes them.


wsfulton:
Merged, thanks.

PR #1208, opened 2018-03-06
Add travis test for node 9

nshmyrev:
Pull request to demonstrate node 9 is broken in travis. Related to https://github.com/swig/swig/pull/968#issuecomment-370918993

PR #1248, opened 2018-05-02
Tests for V8 5.2+ support

jacobwgillespie:
This is an attempt at getting Travis CI testing multiple NodeJS versions for PR #968.  Most likely this PR won't merge and these changes will be incorporated into #968, but I am opening a PR to trigger the Travis CI build.


ojwb:
The new javascript builds all seem to have failed.  I've cancelled the rest of the builds - as William said in #1208, it's better when trying to sort out travis builds to run just the build you're trying to fix rather than all 76 (both for you, travis and anyone else trying to use it).

PR #1255, opened 2018-05-12
[Java] Warn when finalizers aren't generated.

captain5050:
A common bug pattern is to %ignoreall and then to forget to %unignore a
destructor. Without a finalizer the delete method isn't called and so
such bugs can be the source of memory leaks without runtime exceptions.
Warn when a finalizer isn't generated so that such memory leaks are easy
to avoid.


ojwb:
The branch here for some reason has 1569 commits changing 1621 files, but it looks like it's actually just the two commits before the final merge which are relevant and the rest are commits from master.

@wsfulton Does this seem a reasonable change?

If so I can probably apply them by cherry-picking.  And if not we should close this PR.

wsfulton:
The finalizer method is not absolutely needed and is often not generated and manual memory management is preferred for performance reasons and can be more reliable. Java 9 actually deprecated the finalizer method. So in summary, it is entirely reasonable not to have a finalizer and so issuing a warning when it is missing is not correct. Sorry, not accepted.

captain5050:
> 

This seems to assume that the warning is unconditional, the warning is only generated when swig wraps something with a destructor and that destructor cannot be run because of an %ignore. This is a *very* common bug and resultant memory leak. Finalizers are deprecated in favor of cleaners. When swig supports cleaners then this would make sense there as not cleaning up because of a %ignore is also a most likely a bug.

wsfulton:
@captain5050 I feel your pain as it sounds like you've suffered a leak problem. However, SWIG is doing exactly what it is told to, just like a C++ compiler allows you to shoot yourself in the foot. I ran your code change through our test-suite, which doesn't in all honesty have that much testing around destructors and it incorrectly warned 11 times. It warns in completely valid cases such as an explicit %ignore of a destructor, when a destructor is protected and when a destructor is private. I see this indicative of the noise that a SWIG user will have to deal with if this patch was accepted. I'm not really sure what to suggest to be honest other than improving the docs around the ignore all pattern.

captain5050:
> @captain5050 I feel your pain as it sounds like you've suffered a leak problem. However, SWIG is doing exactly what it is told to, just like a C++ compiler allows you to shoot yourself in the foot. I ran your code change through our test-suite, which doesn't in all honesty have that much testing around destructors and it incorrectly warned 11 times. It warns in completely valid cases such as an explicit %ignore of a destructor, when a destructor is protected and when a destructor is private. I see this indicative of the noise that a SWIG user will have to deal with if this patch was accepted. I'm not really sure what to suggest to be honest other than improving the docs around the ignore all pattern.

Thanks! So this patch stemmed from seeing the issue in Google where >17,000 build targets were implicated. Adding %nodefaultdtor was the standard fix. The question is should swig default to dangerous behavior, no finalizer created, and place a burden on developers to try to spot this kind of issue - which given the use of smart pointers and the like is pretty much impossible. Or should swig default to warning (not an error) of this issue and then place a burden on the programmer to opt into the dangerous no warning behavior. I'm pro warnings but I can see why they wouldn't be to everyone's taste.

wsfulton:
If this warning is really useful for some folk, then how about we add it to the list of warnings that are only turned on when `-Wextra` is used. The 'spammy' warnings are turned on with this option and I think this is one of them. In order to accept this patch, it needs to be even less spammy. I'm struggling to see the point of this warning when the destructor is private or protected, so the warning should only be emitted to when a destructor is explicitly ignored by a user. I'll await a new patch for this. Please make sure your pull request contains just the changes required for this change.

ojwb:
A `finalize()` method now triggers a **removal** warning with recent JDK versions (escalated from a **deprecation** warning that has been emitted since JDK 9), so it seems likely SWIG will need to stop generating them in the fairly near future (#1550 is work towards that).

Adding a warning when `finalize()` isn't generated may not be so useful as such at this point, though there may be an equivalent need with cleaners or whatever replaces it.

ojwb:
I'm going to close this as it seems to be abandoned and finalizers are going to be removed from some future Java version so adding a warning about them not being generated seems of decreasing utility as time goes on.

If despite this someone is keen to implemented a more suitable version of this warning (see https://github.com/swig/swig/pull/1255#issuecomment-1045218730) please open a new PR (or push updates to this one and we can reopen).

captain5050:
No problems with me in closing this as I've not moved it forward. Fwiw, the switch from finalizers to cleaners won't solve the problem the issue reports, namely that it is quite easy with the %ignoreall then %unignore style of exposing a C++ API to forget to %unignore the destructor and thereby generate a memory leak - whether the destructor be called by a finalizer or a cleaner. In a large code base I came across quite a number of occurrences, hence the issue/patch to try to eliminate bugs at compile time.

ojwb:
I suspect it's not specific to SWIG/Java even - there are probably other target languages which will leak if the destructor isn't wrapped.

PR #1256, opened 2018-05-12
[Octave] Add support for version 4.4

### LINKED ISSUE #1254: [Octave] Add support for version 4.4
Octave 4.4 has been released and (as is usual for any major Octave release) results in numerous API incompatibilities in the SWIG Octave module. I'm working to address these and have a PR tested and ready for merge.

kwwette:
- This PR add support to SWIG for Octave version 4.4
- Should not introduce any incompatibilities visible to the user of the SWIG-generated wrappings
- Closes #1254


ojwb:
The CI job for Octave 4.4 seems to be hitting the time limit:

```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

Unclear if it's just taking too long, or has got stuck in a loop or similar.

kwwette:
@ojwb I've added a commit to the PR:

https://github.com/swig/swig/pull/1256/commits/f862bca3b3eb586626ce16a653f1d07319d3d11d

This splits the tests for each Octave version into two Travis CI jobs, each running half of the test suite, which should ensure that they complete within the timeout limit.

wsfulton:
Splitting the test-suite is horrible. Can you please rather do a performance analysis as to why Octave is **so** much slower than anything else. 

It's already the only language that requires 2 makejobs on Travis. From what I can see the SWIG run time is comparable to other languages, so it must be the compilation time or octave interpreter runtime causing a problem. Can you tweak the Octave compilation flags so that the c++ compiler runs faster? Is there a header file include problem with the wrappers slowing it down?

kwwette:
@wsfulton I've removed the test-splitting patch. That said I don't really see an alternative. The Octave test suite is already compiled using `-O0`, so it's not spending extra time optimising the code. And I remember a while ago trying to reduce the number of included Octave/C++ headers to the minimum required.

If I take a random test suite wrapper (`octave_dim_wrap.cxx`) and dump the preprocessed code with `-E`, I get a file with ~111000 lines. If I then comment out all the `#include` statements, I then get a file with ~7000 lines. I imagine that's the reason compiling SWIG code generated for Octave is general is much slower than for other languages. I don't know what can be done about that, as I cannot control what header files Octave needs, what C++ headers those files includes, etc.

ojwb:
I agree splitting the testsuite doesn't seem a great approach.

While it might seem obvious that `-O0` would be fastest, that's not necessarily actually true as there are costs to processing the sometimes much larger generated code - for example on my dev box, `-O0` is actually a little slower than either `-O1` or `-O2` for the examples:

```
$ env CCACHE_DISABLE=1 time make check-octave-examples CXXFLAGS=-O0
checking Examples/octave/callback
checking Examples/octave/class
checking Examples/octave/constants
checking Examples/octave/contract
checking Examples/octave/enum
checking Examples/octave/extend
checking Examples/octave/funcptr
checking Examples/octave/funcptr2
checking Examples/octave/functor
checking Examples/octave/operator
checking Examples/octave/pointer
checking Examples/octave/reference
checking Examples/octave/simple
checking Examples/octave/template
checking Examples/octave/variables
84.92user 8.85system 1:26.25elapsed 108%CPU (0avgtext+0avgdata 347520maxresident)k
16inputs+474992outputs (0major+1858661minor)pagefaults 0swaps
$ env CCACHE_DISABLE=1 time make check-octave-examples CXXFLAGS=-O1
checking Examples/octave/callback
checking Examples/octave/class
checking Examples/octave/constants
checking Examples/octave/contract
checking Examples/octave/enum
checking Examples/octave/extend
checking Examples/octave/funcptr
checking Examples/octave/funcptr2
checking Examples/octave/functor
checking Examples/octave/operator
checking Examples/octave/pointer
checking Examples/octave/reference
checking Examples/octave/simple
checking Examples/octave/template
checking Examples/octave/variables
83.48user 8.74system 1:24.39elapsed 109%CPU (0avgtext+0avgdata 348028maxresident)k
16inputs+474992outputs (0major+1860922minor)pagefaults 0swaps
$ env CCACHE_DISABLE=1 time make check-octave-examples CXXFLAGS=-O2
checking Examples/octave/callback
checking Examples/octave/class
checking Examples/octave/constants
checking Examples/octave/contract
checking Examples/octave/enum
checking Examples/octave/extend
checking Examples/octave/funcptr
checking Examples/octave/funcptr2
checking Examples/octave/functor
checking Examples/octave/operator
checking Examples/octave/pointer
checking Examples/octave/reference
checking Examples/octave/simple
checking Examples/octave/template
checking Examples/octave/variables
83.19user 8.96system 1:24.23elapsed 109%CPU (0avgtext+0avgdata 348752maxresident)k
8inputs+474992outputs (0major+1862524minor)pagefaults 0swaps
```

So to tune for fastest test runs I'd suggest trying `-O1` instead, and also dropping `-g` as that can generate a lot of extra data for C++.

But really we need a bigger improvement than that, so perhaps we should look at making use of ccache to cache the results of all these compilations between CI runs - travis has ccache support, see: https://docs.travis-ci.com/user/caching/#ccache-cache

This solution is not perfect as any SWIG code change which causes the generated output to change for all the test cases will result in a test run which has to actually compile everything and so will likely time out, but (assuming the ccache still gets saved when that happens) such situations should heal themselves as the next testsuite run would benefit from the compilation work which the first already did.

kwwette:
@wsfulton @ojwb I've tried a few other approaches to reducing the Octave test suite runtime, including compiling with `-O1` and using precompiled headers. Unfortunately nothing reduced the runtime significantly.

Enabling ccache would probably help for builds which do not affect the Octave wrappings, as @ojwb suggested, but otherwise would still trigger a rebuild. And unfortunately I think Travis only saves the `.ccache/` on a successful job completion, and not on a timeout.

In any case perhaps this should be handled as a separate issue / PR.

kwwette:
Travis build has now passed after restarting timed-out Octave job.

@wsfulton Please merge this PR if you are okay with the changes.

wsfulton:
I'm not sure if your tests really are modifying the optimization level...
```
~/swig/github/swig/Examples/test-suite/octave (master *>)$ make using_private.cpptest CXXFLAGS=-O2
if [ -f ./using_private_runme.m ]; then echo "checking octave testcase using_private (with run test)" ; else echo "checking octave testcase using_private" ; fi
checking octave testcase using_private (with run test)
make -f ../../../Examples/Makefile SRCDIR='./' CXXSRCS='' SWIG_LIB_DIR='../../../Lib' SWIGEXE='../../../swig' INCLUDES='-I../../../Examples/test-suite' SWIGOPT='-outcurrentdir -I../../../Examples/test-suite' NOLINK=true TARGET='using_private' INTERFACEDIR='../' INTERFACE='using_private.i' octave_cpp
make[1]: Entering directory '/home/william/swig/github/swig/Examples/test-suite/octave'
env SWIG_LIB=../../../Lib  ../../../swig -c++ -octave -outcurrentdir -I../../../Examples/test-suite -o using_private_wrap.cxx ./../using_private.i
g++ -g -c -fpic -I. -I./ -O2 using_private_wrap.cxx   -I../../../Examples/test-suite   -D_FORTIFY_SOURCE=2 -I/usr/include/octave-4.0.0/octave/.. -I/usr/include/octave-4.0.0/octave -I/usr/include/hdf5/serial -I/usr/include/octave-4.0.0/octave/.. -I/usr/include/octave-4.0.0/octave -I/usr/include/hdf5/serial -I/usr/include/mpi  -g -Wformat -Werror=format-security -O0
```
In the last line there is both `-O2` and `-O0`. I'm not entirely sure what gcc does in this case.

PR #1257, opened 2018-05-14
Misc. typos

luzpaz:
found via `codespell` and `grep`


Doc/Manual/Scilab.html
```diff
@@ -820,7 +820,7 @@ <H3><a name="Scilab_wrapping_pointers">39.3.6 Pointers</a></H3>
 The user of a pointer is responsible for freeing it or, like in the example, closing any resources associated with it (just as is required in a C program).
 </p>
 
-<H4><a name="Scilab_wrapping_pointers_pointer_adresses">39.3.6.1 Utility functions</a></H4>
+<H4><a name="Scilab_wrapping_pointers_pointer_addresses">39.3.6.1 Utility functions</a></H4>
```
ojwb: If we're going to change this (and so break any existing external links and bookmarks) I think we should realign the anchor name with the current section title - i.e. `Scilab_wrapping_pointers_utility_functions`.

luzpaz: rebased. I didn't implement https://github.com/swig/swig/pull/1257#discussion_r188064972 because I didn't fully understand what the "ask" was

ojwb: Change the anchor to `Scilab_wrapping_pointers_utility_functions` so it matches the current heading it is attached to.

luzpaz: Done, and rebased again. Do you want me to squash commits or not ?

luzpaz:
Found a few more typos before merge. please review 6f69830 as well.

Issue #1259, opened 2018-05-15
cast between incompatible function types reported by gcc 8 for swig & python


D4N:
### Summary 

swig with Python seems to produce C source code that triggers some of the new warnings provided by gcc 8. 

### Reproducer

I am running swig 3.0.12 & gcc 8.1.1 both from the Fedora 28 repositories. The same issue appears on Archlinux, too.

It can be reproduced with the following two files:
test.c:
``` C
void triple_it(int *input) { *input *= 3 * (*input); }
```
and test.i:
```
%module test
%pointer_functions(int, intp);
%{
void triple_it(int *input);
%}
void triple_it(int *input);
```

Then running swig & gcc:
```
$ swig -o test_wrap.c -python test.i
$ gcc -o test.os -c -Wall -Wextra -Werror -O2 -fPIC test.c
$ gcc -o test_wrap.os -c -Wall -Wextra -Werror -O2 -fPIC -I/usr/include/python3.6m test_wrap.c
test_wrap.c:1819:23: error: cast between incompatible function types from ‘PyObject * (*)(PyObject *)’ {aka ‘struct _object * (*)(struct _object *)’} to ‘PyObject * (*)(PyObject *,
PyObject *)’ {aka ‘struct _object * (*)(struct _object *, struct _object *)’} [-Werror=cast-function-type]
   {(char *)"disown",  (PyCFunction)SwigPyObject_disown,  METH_NOARGS,  (char *)"releases ownership of the pointer"},
                       ^
test_wrap.c:1820:23: error: cast between incompatible function types from ‘PyObject * (*)(PyObject *)’ {aka ‘struct _object * (*)(struct _object *)’} to ‘PyObject * (*)(PyObject *,
PyObject *)’ {aka ‘struct _object * (*)(struct _object *, struct _object *)’} [-Werror=cast-function-type]
   {(char *)"acquire", (PyCFunction)SwigPyObject_acquire, METH_NOARGS,  (char *)"acquires ownership of the pointer"},
                       ^
test_wrap.c:1823:23: error: cast between incompatible function types from ‘PyObject * (*)(PyObject *)’ {aka ‘struct _object * (*)(struct _object *)’} to ‘PyObject * (*)(PyObject *,
PyObject *)’ {aka ‘struct _object * (*)(struct _object *, struct _object *)’} [-Werror=cast-function-type]
[...]

Issue #701, opened 2016-06-07
What should be the minimum Python version to support in swig-4.0?


wsfulton:
The swig-3.0.x release supposedly supports python 2.0 and is currently documented: 

_SWIG is compatible with most recent Python versions including Python 3.0 and Python 2.6, as well as older versions dating back to Python 2.0. For the best results, consider using Python 2.3 or newer._

We currently only test for python-2.4 to 3.5. 

One of the aims for the next version of SWIG (3.1) is to vastly reduce the number of options supported for Python wrappers. There are just over 50 command line options and they can be viewed by running `swig -python -help`. Many of these options only work for newer versions of python and not all the combinations can be tested. Many of the options require newer versions of Python.

This ticket is an opportunity for SWIG users to express their minimum version required. I would like to start by suggesting that 2.7 should be the minimum version supported as there is still a huge code base that has not moved to python-3 and most 2.x projects can move to 2.7. 2.7 is also a good stepping stone to moving to the very different python-3 world as it has some 3.x features backported into it. If anyone wants support for an older version please say so here and why.

Lastly we don't know when swig-3.1 will be released. It requires volunteers to implement the necessary changes, so if you are able to help, please say so.

mromberg:
Given that python 2.6 reached end of life in 2013:


https://www.python.org/dev/peps/pep-0361/

  It seems unreasonable for swig to have to support it or anything even older.  However 2.7 will be supported until 2020:

http://legacy.python.org/dev/peps/pep-0373/

  So, the suggestion for swig-3.1 to only support python-2.7+ seems very reasonable to me.

P.S.  Also, I may be able to help with swig-3.1.  Let me know what needs doing.

mruffalo:
It might also be worth it to specify a minimum version of Python 3. As per [PEP 392](https://www.python.org/dev/peps/pep-0392/) and [PEP 398](https://www.python.org/dev/peps/pep-0398/), Python 3.2 and 3.3 are now EOL, so supporting 2.7 and 3.4+ seems reasonable.

[...]

PR #1261, opened 2018-05-18
Rationalise the Python-specific options

ojwb:
This is a significant start on cleaning out the cruft ready for 4.0.0 as agreed in #701.  I haven't yet started to rationalise the command line options, but with this done we can make `-modern` the default and drop support for `-classic`, and then hopefully pick away at some of the others.

It removes some of the function casts which GCC8 warns about (#1259), but is not yet a complete fix for that.


Source/Modules/python.cxx
```diff
@@ -836,7 +834,7 @@ class PYTHON:public Language {
       Printf(default_import_code, tab8 tab4 "return importlib.import_module('%s')\n", module);
       Printf(default_import_code, tab4 "%s = swig_import_helper()\n", module);
       Printv(default_import_code, tab4, "del swig_import_helper\n", NULL);
-      Printv(default_import_code, "elif _swig_python_version_info >= (2, 6, 0):\n", NULL);
```
wsfulton: I think we should raise a RuntimeError if < 2.6 saying unsupported version to give a helpful error message to users trying to use older versions.

Also the C code should `#error` if < 2.6.

ojwb:
>
> I think we should raise a RuntimeError if < 2.6 saying unsupported version
> to give a helpful error message to users trying to use older versions.
>
OK.

> Also the C code should #error if < 2.6.
>
It already does.

>

Source/Modules/python.cxx
wsfulton: Ah, I just saw elsewhere you have already put in the check in the C code, so ignore that comment

wsfulton:
Hey @ojwb, please don't get carried away modifying python.cxx, the doxygen code needs merging first, which I'm working on now. Also, I think focus should be on getting -O working before removing anything more because that is what our eventual aim is. Until that is fully working, we can't be confident that we can just remove any of the non -O options.

ojwb:
This PR is blocked waiting for the doxygen branch to be merged, but meanwhile we're wasting a lot of travis resources on 6 CI builds for Python versions we're not intending to support, and so I've pushed a change to remove those builds (f7a255c38acff21cad01e300ba8d8169aea9c377) - this shouldn't cause problematic conflicts with the doxygen branch.

But that's why this PR now shows a conflict on `.travis.yml`.  I'll resolve that when I pick up work on this again.

ojwb:
The doxygen branch is now merged, so I've picked out the changes to drop support for Python < 2.7 and Python 3 < 3.4 and pushed those to master, then rebased this branch on current master so it just has the changes towards unpicking the thicket of Python options which we want to hack back.

mromberg:
> AIUI, the only area where -O doesn't currently work is with the nondynamic feature, which there seemed to be a firm consensus on removing However when I said I was actually going to remove it you told me to hold off on doing so.
> 
> I view the current incompatibility as being a problem with nondynamic rather than with -O, and I really don't see the point in investing effort in trying to fix a feature which seems to have no users and which nobody likes.

  Yes.   The nondynamic feature was one of the last things I tried to tackle when getting all the -O tests to work.   There is still an open pull request with my changes #718.   These changes did make the tests that failed pass at the time.   However, I don't think #718 should be applied at all.   It is mostly useful as a reference.  My opinion is that the nondynamic feature should be removed all together.   

  It makes the most sense to remove the nondynamic tests first.  This would as a side effect make -O also work.   Then the more detailed surgery of removing nondynamic and all the complicated meta class stuff can occur.

  It is very hard to imagine any working swig generated code that would break if swig were to remove the nondynamic feature.  To be safe it could be made a no op and or issue a warning.  

  What nondynamic does is cause python to throw an exception if any python code attempts to add new attributes to a swig generated class.   So, almost all user code that may have enabled nondynamic would not be doing this (adding new attributes).   The only code that would break is I suppose code that was expecting these exceptions and catching them.   And while I can't say for certain no one is doing this.   It would surprise me if anyone were.

  So, I think the tests for nondynamic should just be removed.   I also think the feature itself should be removed unless any super convincing argument can be made for it's existence.   The argument against it is that there is a lot of questionable code inserted into python wrappers to support it.  And this questionable code is being used even if the feature nondynamic is never turned on.

wsfulton:
Now #1293 needs addressing before we can focus on the full rationalisation of the Python options. Please add any thoughts about type errors to #1293.

wsfulton:
With #1293 fixed, I am now working on merging this patch into master. There are a number of conflicts, but I think I have it all working now. I'll do some more testing, then commit it tomorrow.

wsfulton:
@ojwb, thanks for the patch, closing.

Issue #1165, opened 2017-11-30
AOT support for C# directors


KallynGowdyYetiCGI:
Hello, I'm currently trying to use SWIG as a tool to generate C++ bindings for C# that is used in Unity 3D for mobile development. Of particular note is that I need to support calling C# code from C++.

Enter directors.

Unfortunately, the code that SWIG generates does not work in scenarios where JIT is not available because of the use of instance member methods attached to delegates. 

To explain it another way, when run on iOS (which disallows JIT compilation) IL2CPP throws an exception because it detects that I'm trying to marshal a function pointer from unmanaged code. (explanation below)

In normal .Net, the VM would simply emit the code necessary to handle the callback at runtime, but with IL2CPP (the backend that Unity 3D uses) we need to tell them about the callback ahead of time so that they can generate the code at compile time. We can do this with `MonoPInvokeCallbackAttribute`, but this has the noticeable drawback of only supporting callbacks to managed methods that are static. (See [this page](https://docs.unity3d.com/Manual/TroubleShootingIPhone.html) for more info; search for `"MonoPInvoke"`) 

This is where my major problem comes in. Declaring a class in C# that overrides methods declared by a class in C++ simply cannot work without changing how SWIG generates the C++ and C# director code.

Effectively, I need SWIG to generate directors that accept an additional pointer when calling `swig_connect_director(delegate)`. This pointer would be used as a way to reference methods that are tied to object instances.

For example, `swig_connect_director` would be redefined as:
```c++
void swig_connect_director(SWIG_Callback callback, void *delegate_ref) {
  swig_callback = callback;
  swig_delegate_ref = delegate_ref;
}
```
where `swig_delegate_ref` is a `void *` field on the `SwigDirector_XyzBase` class and `SWIG_Callback` is defined as:
```c++
typedef void (SWIGSTDCALL* SWIG_Callback0_t)(void *, args...);
```
(pointer to a function whose arguments accept `void *` and whatever arguments the actual method requires)

Then, on the C# side we would generate a static function which wraps the instance delegate.
```c#
// New delegate to represent the static method callback.
public delegate void SwigDelegateStaticAction (IntPtr delegate_ref, args...);

// Existing delegate which SWIG already creates
public delegate void SwigDelegateAction(args...);

// Instead of using SwigDirectorAction as the callback,
// we use a static method which will resolve the instance method for us.
[MonoPInvokeCallback]
private static void SwigDirectorStaticAction(IntPtr delegate_ref, args...)
{
[...]

PR #1262, opened 2018-05-18
Implemented Mono AOT directors compatibility

ZimM-LostPolygon:
References:
https://docs.microsoft.com/en-us/xamarin/ios/internals/limitations#Reverse_Callbacks
https://github.com/swig/swig/issues/1165
---

This commit is based on the `monotouch` branch, although pretty much nothing remained from it. `monotouch` didn't actually solve the problem of not using instance method delegates for callback. In this branch, the general idea is to introduce a level of indirection - instead of sending managed instance method delegate to the native side of a director, send a "dispatcher" static method delegate, in conjunction with an instance method handle. `GCHandle` is used as a handle.

Mono AOT compatibility mode is enabled with `-monoaotcompat` command line flag. It is not enabled by default since it comes with some performance cost.

I'm a bit concerned about the code maintainability, since changes are all around the C# module code. Perhaps we can add a separate integration test pass on Travis, with `-monoaotcompat` flag added? Currently, all tests pass just fine with it.

Obviously, some documentation will need to be added.


iBicha:
Would like for this merged as well, manual patching doesn't look nice as a solution
Thanks

wsfulton:
@ntkeep thanks for the contribution, I'll have a good look at it soon. 

Some detailed technical documentation including an example is indeed needed. It'll really help with the review too. I suggest you add a new test case and use that in the docs. The test-suite is able to handle flags being set per testcase. Add the test into Examples/test-suite/csharp/Makefile.in and set SWIGOPT, like the intermediary_classname test does at https://github.com/swig/swig/blob/f8ffc04215b1be9ade91d418c0e0cf309e1809a2/Examples/test-suite/csharp/Makefile.in#L47 

If you are working on Windows, you may find the test-suite challenging to work, so you could instead add a new Example under Examples/csharp instead.

ZimM-LostPolygon:
@wsfulton Just adding a new test case won't be enough. It would be much better to additionally run all *existing* C# tests with `-monoaotcompat` flag added, but I can't find a way to do that.

wsfulton:
Could you please change the option to simply `-monoaot`? Is AOT specific to mono-csc  and not supported by Microsoft's csc? Just wondering if the name is correct.

Set the env variable:
```
SWIG_FEATURES=-monoaot
```
See .travis.yml which sets this env a few times for Python and -builtin. For appveyor (Windows) testing, tweak appveyor.yml and set the SWIG_FEATURES env variable too.

Our regression testing is taking a long time already, so doubling it up for C# is not ideal. I think given what I've seen in this patch, one director test would cover (nearly) all the changes, or have I got this wrong? You could even use one of the other director tests, perhaps this:
```swig
// file Examples/test-suite/director_classes_monoaot.i
%module director_classes_monoaot
%include "director_classes.i"
```
and the Makefile:
```
director_classes_monoaot.cpptest: SWIGOPT += -monoaot 
```
Then copy director_classes_runme.cs to director_classes_monoaot_runme.cs and adjust. Even if we do have a test with this flag set and testing the entire test-suite, it'd still be very handy to have one testcase testing monoaot.

wsfulton:
Also, sorry @ZimM-LostPolygon, in my earlier response I mistakenly implied @ntkeep had contributed the patch when it was you.

wsfulton:
hi, @ZimM-LostPolygon, any response to my post in July? I'm looking to finalize the patches that go into SWIG 4 and will include this one if it is suitably tested.

ZimM-LostPolygon:
> Could you please change the option to simply -monoaot? Is AOT specific to mono-csc and not supported by Microsoft's csc? Just wondering if the name is correct.

I think it might be just `-aot`. `csc` doesn't support AOT compilation, but maybe it will someday. And the two biggest AOT compilers are Mono and Unity's IL2CPP, with IL2CPP using the same attributes as Mono.

wsfulton:
Accepting patches like this without tests and documentation is irresponsible and unprofessional. I can probably put together some tests. Can you not write a couple of paragraphs for inclusion in the docs? If you post it to this issue, I'll adapt it for incorporation into http://swig.org/Doc3.0/CSharp.html. It will need some justification for the option (when to use etc), some decent background technical details and links.

wsfulton:
Hi @ZimM-LostPolygon, I'd like to draw a line under this patch for the 4.0.0 release. As you said _Obviously, some documentation will need to be added._ I really need some outline docs from you, I can flesh them out if you knock up a few salient points.

ZimM-LostPolygon:
@wsfulton I'll try my best to fix and document this PR over the next weekend.

Mystfit:
Did this solution make its way into Swig 4.0? It looks to be the perfect solution to an issue I'm having with Unity, Swig, C# and Android. At the moment I have to patch the generated PInvoke file with [MonoPInvokeCallback(typeof(Delegate)] decorators. Am I correct in assuming that this PR automates the process?

JessHolle:
Any hope of getting this merged???

wsfulton:
If someone brings the patch up to standard, I'll merge it in. I think all the information for what is required is in this ticket. Seems that @ZimM-LostPolygon is not going to do this, so someone else interested in having this merge in, please step in and add the necessary improvements.

Mystfit:
@wsfulton I'll tackle it. From looking at this thread I'm gathering the required changes are 

- Rename the command line flag to -monoaot
- Create a test (Examples/test-suite/director_classes_monoaot.i)
- Add documentation for the command line flag and when to use it

I'll take a stab at giving some technical background information but I'm afraid I'm not quite as up to speed with writing some of the extra use-case tests and when to use this flag.  My personal use case for this PR has been to get Swig wrapped code functioning in Unity builds on Android devices and and this fix has achieved that by fixing some of the swig exception functions throwing errors on Android.

Would the best test be for me to write a test where a C# test program sends a "dispatcher" static method delegate, in conjunction with an instance method handle to a director which then calls back into managed code? (paraphrasing Zim's words). 

How much information do we need in the documentation to justify this feature? Would it be enough to summarize the information from [the Xamarin docs linked in the PR description?](https://docs.microsoft.com/en-us/xamarin/ios/internals/limitations#Reverse_Callbacks)

wsfulton:
@Mystfit, excellent. Possibly `-aot` is a better command-line option as I read that CoreRT will support AOT, but I couldn't work out if other C# compilers will support it. This is the sort of info, I'd expect to see in the docs too. 

Your example program sounds fine. I can port it into the SWIG test-suite. As much info in the docs as possible would be helpful. I'd imagine that might be a couple of paragraphs plus one or two classic motivating use cases.

The example needs to run even if AOT is not supported by the compiler. Just output a message saying C# compiler doesn't support AOT so the test still passes. So, detecting Mono for the runtime tests is needed (see #1576) or even better some sort of detection of AOT support in the compiler.

JessHolle:
Has anyone rebased this against the latest release or been considering doing so?

I'm wondering as having wasted too much time on handcrafted workarounds to this issue, I'm now left wondering if just building the pull request myself and using it isn't the best option at this point.

I don't know that I have time/energy/background to get this "over the finish line" in terms of getting this merged into SWIG.  But I am now about at the point of understanding the SWIG implementation sources.

cho-trackman:
Why did this stall?

JessHolle:
My understanding is that this stalled because no one had the combination of knowledge and time/energy to drag the pull request over the finish line, so to speak, i.e. get all documentation updates done, add/update tests, etc.

Exactly which pieces it stalled on is not clear to me, though, if that's what you're asking.  (And, yes, I'm still doing horrible workarounds for the lack of this capability.)

Nolram12345:
Any possible updates on this ?

devshgraphicsprogramming:
I think Blazor forbids passing delegates around, so this is needed for SWIG to run on Mono-WASM

Source/Modules/csharp.cxx
```diff
@@ -1934,15 +1947,29 @@ class CSHARP:public Language {
...
 	String *overname = Getattr(udata, "overname");
-	Printf(proxy_class_code, "    if (SwigDerivedClassHasMethod(\"%s\", swigMethodTypes%s))\n", method, methid);
+	if (mono_aot_compatibility_flag) {
+	  Printf(proxy_class_code, "    global::System.IntPtr swigDelegate%sgcHandlePtr = global::System.IntPtr.Zero;\n", methid);
+	}
+	Printf(proxy_class_code, "    if (SwigDerivedClassHasMethod(\"%s\", swigMethodTypes%s)) {\n", method, methid);
 	Printf(proxy_class_code, "      swigDelegate%s = new SwigDelegate%s_%s(SwigDirector%s);\n", methid, proxy_class_name, methid, overname);
+	if (mono_aot_compatibility_flag) {
+	  Printf(proxy_class_code, "      swigDelegate%sdispatcher = new SwigDelegate%s_%s_Dispatcher(SwigDirector%s_Dispatcher);\n", methid, proxy_class_name, methid, overname);
+	  Printf(proxy_class_code, "      global::System.Runtime.InteropServices.GCHandle swigDelegate%sgcHandle = global::System.Runtime.InteropServices.GCHandle.Alloc(swigDelegate%s, global::System.Runtime.InteropServices.GCHandleType.Weak);\n", methid, methid);
```
devshgraphicsprogramming, line 1957: if that handle was changed to a strong one then it would implement some of #2410 
but probably break everything else thats not in the spirit of #2410

wsfulton:
To everyone interested in this issue, can you please focus on the small amount of work required to get this pull request over the line. It's just about documentation and understanding when/why it would be needed. What I need is a couple of paragraphs covering some basics:

1. Why is this needed?
2. When would one use it?
3. What is broken requiring this?
4. What is faster/better with this patch?
5. What are the trade offs if using this patch?
6. How does one use it?
7. What are the system requirements / version of everything needed?
8. How does one test it - is it fully tested merely by running the entire test-suite with the additional `-monaot` command line?
9. ... you get the picture ... any other background information or might be useful for me to review it and for users to use it.

If you can't manage to edit the CSharp.html file, please just put the additional docs in a separate comment in this issue.

Thanks, then we can get it merged. I'm happy to resolve the merge issues etc, but need the above clarifications.

devshgraphicsprogramming:
## Quality issues (mostly GCHandle usage and naming)

### Naming and Quantity of Handles

I haven't looked much are the code (I'm not a SWIG regular) but it seems to me like there is (or there should be, I can't really tell) only one `GCHandle` to the derived object.

However it and its various representations are called `swigDelegate` which is an unfortunate prefix name, given that when one talks about delegates in C# one means methods/callbacks not the object itself, so I'd propose renaming that to avoid confusion.

### IMPORTANT: There's a resource leak

Even though the `GCHandle` is weak, I can't find any code that `Free`s it after allocation.

While a weak handle allows the object to be collected and the object won't leak, the handle itself must be freed no matter the type.

 > A new [GCHandle](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle?view=net-6.0) that protects the object from garbage collection. This [GCHandle](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle?view=net-6.0) must be released with [Free()](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.gchandle.free?view=net-6.0#system-runtime-interopservices-gchandle-free) when it is no longer needed.
 
 So that needs fixing.
 
 P.S. Please add a `%feature` or a flag to control the `GCHandle` type, its probably 5 minutes worth of work, and it would go a long way towards #2410

devshgraphicsprogramming:
We'll submit a PR that fixes the GCHandle memory leak.

wsfulton:
Hi @devshgraphicsprogramming, I've got some time to look at this again. 

> We'll submit a PR that fixes the GCHandle memory leak.

Do you have that? I'll try and resolve the conflicts if you are available to help with any follow up questions too.

ojwb:
>> What is broken requiring this?

> Can you rephrase the question?

@devshgraphicsprogramming It wasn't my question, but think a clearer version is probably "What problem(s) does using the `-monoaot` option solve?"

devshgraphicsprogramming:
> > > What is broken requiring this?
> 
> > Can you rephrase the question?
> 
> @devshgraphicsprogramming It wasn't my question, but think a clearer version is probably "What problem(s) does using the `-monoaot` option solve?"

the PR is so old I don't remeber much to do with it, also we probably can't contribute this due to licensing issues.

sunnyqeen:
I can try to explain why AOT option is needed.
Xamarin iOS and Unity use AOT to compile the C# code to native code to speed up the app running, and it's forced.
So that the AOT compiler needs to know which methods are called from C# to native and native to C#. 
This introduces MonoNativeFunctionWrapper and MonoPInvokeCallback, with these attributes, the compiler knowns the methods and generates the necessary native code, but the limitation is "only static methods are supported".

the swig generated classes for director use dynamitic delegate functions (instance functions) for callback. 
This works fine for JIT, but not AOT, as JIT could create functions at runtime. here is some code from \Examples\csharp\callback\Callback.cs
```
  private void SwigDirectorConnect() {
    if (SwigDerivedClassHasMethod("run", swigMethodTypes0))
      swigDelegate0 = new SwigDelegateCallback_0(SwigDirectorMethodrun); // it's dynamic,  not working for AOT
    examplePINVOKE.Callback_director_connect(swigCPtr, swigDelegate0);
  }

  private void SwigDirectorMethodrun() {
    run();
  }

  public delegate void SwigDelegateCallback_0();

  private SwigDelegateCallback_0 swigDelegate0;
```
This pr try to solve this by using additional static delegate functions.

wsfulton:
@sunnyqeen, thanks for your input. I think @devshgraphicsprogramming answered all the motivation questions I had. I can move that into the official docs, but there are no tests for this. It requires an enhancement to the Gitlab CI and if it passes with aot turned on and the normal C# runs don't fail before this pull request is going to be ready for final review and merge to master.

sunnyqeen:
This pr requires no new test, we could just pass the new option to the ci.
Passing SWIGOPT when doing make. here is the example ci.yml changes.
```
    - name: Test
      working-directory: build/build
      run: |
          source $GITHUB_WORKSPACE/Tools/CI-linux-environment.sh
          set -x

          if test -z "$SWIGLANG"; then
            make $SWIGJOBS check-ccache
            make $SWIGJOBS check-errors-test-suite
          else
            case "$SWIGLANG" in
              javascript)
                case "$ENGINE" in
                  v8 | jsc)
                    # Running tests using v8 or jsc involves creating a custom
                    # interpreter in Tools/javascript, which is currently broken
                    # for parallel builds (we attempt to update this interpreter
                    # while running, resulting in "Text file busy" error).
                    unset SWIGJOBS
                esac
                ;;
              ocaml)
                # `make check-ocaml-test-suite` fails with parallel make with:
                #
                # File "swig.ml", line 1:
                # Error: Could not find the .cmi file for interface swig.mli.
                #
                # Apparently we ought to be using `ocamldep` to generate make
                # dependencies.
                unset SWIGJOBS
                ;;
            esac

            # Stricter compile flags for examples. Various headers and SWIG generated code prevents full use of -pedantic.
            cflags="$($GITHUB_WORKSPACE/Tools/testflags.py --language $SWIGLANG --cflags --std=$CSTD --compiler=$CC)"
            cxxflags="$($GITHUB_WORKSPACE/Tools/testflags.py --language $SWIGLANG --cxxflags --std=$CPPSTD --compiler=$CC)"
            make check-$SWIGLANG-version
            make check-$SWIGLANG-enabled
            make $SWIGJOBS check-$SWIGLANG-examples CFLAGS="$cflags" CXXFLAGS="$cxxflags" PY_ABI_VER="$PY_ABI_VER"
            make $SWIGJOBS check-$SWIGLANG-test-suite CFLAGS="$cflags" CXXFLAGS="$cxxflags" PY_ABI_VER="$PY_ABI_VER"
            if [ "$SWIGLANG" = "csharp" ]; then
              make $SWIGJOBS check-$SWIGLANG-examples SWIGOPT="-monoaot" CFLAGS="$cflags" CXXFLAGS="$cxxflags" PY_ABI_VER="$PY_ABI_VER"
              make $SWIGJOBS check-$SWIGLANG-test-suite SWIGOPT="-monoaot" CFLAGS="$cflags" CXXFLAGS="$cxxflags" PY_ABI_VER="$PY_ABI_VER"
            fi
          fi
```

trittsv:
Hi, is someone still working on this? What is missing? Shall other people jump in working on it?

devshgraphicsprogramming:
> Hi, is someone still working on this? What is missing? Shall other people jump in working on it?

you need to rewrite it, I can't get a go ahead to release the source IIRC.

JessHolle:
> > Hi, is someone still working on this? What is missing? Shall other people jump in working on it?
> 
> you need to rewrite it, I can't get a go ahead to release the source IIRC.

I don't see that in the comments above...  I thought the main gaps initially were just around testing and documentation.  Or is this a matter of later fixes you provided but can't get approval to release?

I no longer work on the project where I became of this gap, but the amount of hand-coded boilerplate required to workaround Mono AOT limitations is quite obnoxious and error prone, while still clearly being very formulaic.  It seems like a very solvable problem that's really painful for the few (e.g. Unity users) in this situation.

devshgraphicsprogramming:
our fix works, you just need to cleanroom it

trittsv:
We also use this branch in some projects where we have to use Unity on android. Would be cool if it would be in a normal swig version.

@devshgraphicsprogramming can you specify more what exactly to clean up?

@wsfulton can you tell what is missing from your side?

PR #1266, opened 2018-06-03
AddAccessor - string length bugs leading to uninitialised variables

richardbeare:
This pull request is for records and testing. I do not intend this PR to be included, 
as other refactoring will supersede it.

There is a problem processing methods with short names. The
test looks for a suffix, length 4, testing whether it matches
"_get". If the length of the name is less than 4, the test
isn't applied so "isSet" does not get correctly initialised
and remains set to -1. This leads to methods with short names
not being processed correctly. This fix initialises isSet to 0
prior to attempting the test.

A second problem is addressed, in which a pointer is set relative
to the end of string without checking length.

The following tests now have different R files, with previously
incorrect set-style methods being removed:

abstract_access.R
class_scope_namespace.R
clientdata_prop_a.R
clientdata_prop_b.R
director_basic.R
director_enum.R
director_property.R
inherit_void_arg.R
multiple_inheritance_interfaces.R
template_extend_overload_2.R
template_partial_specialization.R
template_partial_specialization_typedef.R
variable_replacement.R

### DIFF
--- a/Source/Modules/r.cxx
+++ b/Source/Modules/r.cxx
@@ -1052,9 +1052,10 @@ int R::OutputMemberReferenceMethod(String *className, int isSet,
     String *item = Getitem(el, j);
     String *dup = Getitem(el, j + 1);
     char *ptr = Char(dup);
-    ptr = &ptr[Len(dup) - 3];
+    int n = Len(dup);
+    ptr = &ptr[n - 3];
 
-    if (!strcmp(ptr, "get"))
+    if ((n > 3) && !strcmp(ptr, "get"))
       varaccessor++;
 
     if (Getattr(itemList, item))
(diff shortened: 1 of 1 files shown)


Source/Modules/r.cxx
```diff
@@ -1052,9 +1052,10 @@ int R::OutputMemberReferenceMethod(String *className, int isSet,
     String *item = Getitem(el, j);
     String *dup = Getitem(el, j + 1);
     char *ptr = Char(dup);
-    ptr = &ptr[Len(dup) - 3];
+    int n = Len(dup);
+    ptr = &ptr[n - 3];
```
ojwb: If `n < 3` then this puts `ptr` before the start of the string, which is undefined behaviour if the string is at the start of the allocated buffer (I'm not sure how the DOH string stuff works, so maybe there's a count or something in front of the string data).

Either way, it would be cleaner to just move the setting of `ptr` inside the `n > 3` check.

richardbeare: You are quite right - I thought about this and decided to be lazy as I'm planning to replace a lot of this code anyway. No excuses - I will send an update soon.

richardbeare:
This patch corrected bugs in `AddAccessor` code which lead to uninitialized variables. I developed the fixes for the purpose of testing the refactored versions of the code. The refactoring is in PR #1268.

PR #1268, opened 2018-06-06
R member list trial simplify

richardbeare:
This a first pass at addressing issues with structure/class member access methods in the R module.

This is an alternative version of #1267 - I've been explicit about breaking the lists into types before calling OutputMemberReferenceMethod, rather than doing everything inside. I think it is a bit clearer.

### DIFF
--- a/Source/Modules/r.cxx
+++ b/Source/Modules/r.cxx
@@ -1862,11 +1862,7 @@ int R::functionWrapper(Node *n) {
     /* Add the name of this member to a list for this class_name.
        We will dump all these at the end. */
 
-    int nlen = Len(iname);
-    char *ptr = Char(iname);
-    bool isSet(0);
-    if (nlen > 4) isSet = Strcmp(NewString(&ptr[nlen-4]), "_set") == 0;
-
+    bool isSet(GetFlag(n, "memberset"));
 
     String *tmp = NewString("");
     Printf(tmp, "%s_%s", class_name, isSet ? "set" : "get");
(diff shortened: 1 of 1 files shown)


richardbeare:
@ojwb - I'm hoping to get back to this work soon. Any preferences as to the preferred approach?

I think this one is a bit clearer than #1267.

ojwb:
I've stepped back from actively working on SWIG for now, so I'm not really the best person to ask.

Probably best to consult @joequant regarding the R backend.

richardbeare:
@joequant , @wsfulton .

I'm hoping to get back to this work soon. Any preferences as to the preferred approach?

I think this one is a bit clearer than #1267.

wsfulton:
Hi @richardbeare. First could you please fix the coding convention problems, see http://swig.org/Doc3.0/Extending.html#Extending_coding_style_guidelines. The indent tool can help you identify them. Also don't comment out code, please delete it unless you have a good reason to keep it. I suspect that there are plenty of pre-existing coding convention problems in r.cxx which could be fixed, but please don't include lines with whitespace only changes in this patch as they'd be irrelevant to the proposed fix. Any whitespace changes would need a separate whitespace only change patch, but I don't propose you do that now.

This patch needs runtime tests for all the examples fixed. You identified quite a few in #1267. The runtime tests will really help me understand what is going on too, given my knowledge of the R module is very limited.

wsfulton:
Closing as I understand this is superceded by #1511.

PR #1274, opened 2018-06-09
Add support for std::map<> comparator template argument for Java

vadz:
Allow exporting maps using non-default comparison function, similar to
what was done for C# back in 9efaf954c71118d41ba9bf43ed371bbe83093564


wsfulton:
Looks fine. I'm just replicating the change across all languages and so will merge and commit it all later today.

PR #1279, opened 2018-06-28
 Avoid function-local statics that call Python code to avoid deadlocks

### LINKED ISSUE #1275: Python deadlock caused by C++ static initializer(s) used in pyrun.swg
Specifically: https://github.com/swig/swig/blob/master/Lib/python/pyrun.swg#L494

Being declared static causes a lock to be held by C++ during the execution of `SwigPyObject_TypeOnce()` which includes Python C API calls that ultimate allocate memory via the Python allocator and thus can trigger Python GC run.  A Python GC run can cause a GIL release while doing destruction of some objects which can in turn switch to a new thread which then calls into a SWIG wrapper hitting the exact same static `SwigPyObject_TypeOnce()` locked call... leading to a deadlock.

petrmitrichev:
This fixes #1275 , more specifically the following:

https://github.com/swig/swig/blob/master/Lib/python/pyrun.swg#L494 has a function-local static, so first time control passes through it a lock to be held by C++ during the execution of SwigPyObject_TypeOnce() which includes Python C API calls that ultimate allocate memory via the Python allocator and thus can trigger Python GC run. A Python GC run can cause a GIL release while doing destruction of some objects which can in turn switch to a new thread which then calls into a SWIG wrapper hitting the exact same static SwigPyObject_TypeOnce() locked call leading to a deadlock.

Reproducing this deadlock in a small testcase is not simple as we need to make sure garbage collection triggers exactly during the first creation of an object in a certain SWIG module, and that another thread interrupts said garbage collection.

### DIFF
--- a/Lib/python/pyinit.swg
+++ b/Lib/python/pyinit.swg
@@ -431,6 +431,11 @@ SWIG_init(void) {
     SwigPyBuiltin_AddPublicSymbol(public_interface, swig_const_table[i].name);
 #endif
 
+  SWIG_Py_None_global_Init();
+  SwigPyObject_type_global_Init();
+  SwigPyPacked_type_global_Init();
+  SWIG_Python_TypeCache_global_Init();
+
   SWIG_InstallConstants(d,swig_const_table);
 %}
 
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -259,6 +259,9 @@ _SWIG_Py_None(void)
   return none;
 }
 
+/* This used to be a function-local static variable, but accessing Python
+ * C API from inside a function-local static initializer can lead to deadlocks.
+ * https://github.com/swig/swig/issues/1275 */
 static PyObject *SWIG_Py_None_global = 0;
 
 SWIGRUNTIME void SWIG_Py_None_global_Init(void) {
@@ -506,6 +509,9 @@ SWIGRUNTIME void SwigPyObject_type_global_Init(void) {
   // Do nothing
 }
 #else
+/* This used to be a function-local static variable, but accessing Python
+ * C API from inside a function-local static initializer can lead to deadlocks.
+ * https://github.com/swig/swig/issues/1275 */
 static PyTypeObject *SwigPyObject_type_global = 0;
 
 SWIGRUNTIME void SwigPyObject_type_global_Init(void) {
@@ -859,6 +865,9 @@ SwigPyPacked_compare(SwigPyPacked *v, SwigPyPacked *w)
 
 SWIGRUNTIME PyTypeObject* SwigPyPacked_TypeOnce(void);
 
+/* This used to be a function-local static variable, but accessing Python
+ * C API from inside a function-local static initializer can lead to deadlocks.
+ * https://github.com/swig/swig/issues/1275 */
 static PyTypeObject* SwigPyPacked_type_global = 0;
 
 SWIGRUNTIME void SwigPyPacked_type_global_Init(void) {
@@ -1467,6 +1476,9 @@ SWIG_Python_SetModule(swig_module_info *swig_module) {
   }
 }
 
+/* This used to be a function-local static variable, but accessing Python
+ * C API from inside a function-local static initializer can lead to deadlocks.
+ * https://github.com/swig/swig/issues/1275 */
 static PyObject *SWIG_Python_TypeCache_global = 0;
 
 SWIGRUNTIME void SWIG_Python_TypeCache_global_Init(void) {
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -258,11 +258,24 @@ _SWIG_Py_None(void)
   Py_DECREF(none);
   return none;
 }
-SWIGRUNTIME PyObject * 
+
+static PyObject *SWIG_Py_None_global = 0;
+
+SWIGRUNTIME void SWIG_Py_None_global_Init(void) {
+  SWIG_Py_None_global = _SWIG_Py_None();
+}
+
+SWIGRUNTIME PyObject *
 SWIG_Py_None(void)
 {
-  static PyObject *SWIG_STATIC_POINTER(none) = _SWIG_Py_None();
-  return none;
+  assert(SWIG_Py_None_global);
+  return SWIG_Py_None_global;
+}
+
+#else
+SWIGRUNTIME void SWIG_Py_None_global_Init(void)
+{
+  // Do nothing
 }
 #endif
 
@@ -488,11 +501,21 @@ SwigPyObject_type(void) {
     assert(cd->pytype);
     return cd->pytype;
 }
+
+SWIGRUNTIME void SwigPyObject_type_global_Init(void) {
+  // Do nothing
+}
 #else
+static PyTypeObject *SwigPyObject_type_global = 0;
+
+SWIGRUNTIME void SwigPyObject_type_global_Init(void) {
+  SwigPyObject_type_global = SwigPyObject_TypeOnce();
+}
+
 SWIGRUNTIME PyTypeObject*
 SwigPyObject_type(void) {
-  static PyTypeObject *SWIG_STATIC_POINTER(type) = SwigPyObject_TypeOnce();
-  return type;
+  assert(SwigPyObject_type_global);
+  return SwigPyObject_type_global;
 }
 #endif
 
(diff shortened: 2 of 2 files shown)


petrmitrichev:
The (only two) failures on Travis CI seem unrelated ("The command "sudo apt-get -qq install php7.0-cli php7.0-dev" failed 3 times."), so I think this passes the tests.

wsfulton:
I restarted those failed Travis tests and they have passed.

wsfulton:
I've merged this patch in. I made a few minor changes around initialization order and simplified the implementation a bit.

jschueller:
Using master it seems we get segfaults since commit 4b4e018 (ok @ b3763c8), investigating.

petrmitrichev:
Thanks William for the quick merge!

The reason I've decided to have "_Init" methods in my original patch was to avoid the requirement for the code in pyrun.swg and pyinit.swg to be in very tight sync, which seems fragile and easy to break with future refactorings.

In particular, I was not sure if #ifdef values in pyinit.swg and pyrun.swg are guaranteed to evaluate to the same thing, given that SWIG_PYTHON_BUILD_NONE is actually defined in pyrun.swg: https://github.com/swig/swig/blob/master/Lib/python/pyrun.swg#L234 . A bit of a shot in the dark without a repro, but could this be the reason for the segfault (i.e. SWIG_PYTHON_BUILD_NONE actually not set when evaluating pyinit.swg)?

jschueller:
Your changes seem fine @petrmitrichev, it may be due to 4b4e018, @wsfulton follow-up change

petrmitrichev:
@jschueller Thanks for confirming!

In my previous comment I'm theorizing why @wsfulton 's follow-up change can be breaking things: he removed the _Init methods that I had, and instead inlined them into pyinit.swg. My theory is that #ifdef's can evaluate differently in pyinit.swg.

jschueller:
oh, ok

jschueller:
It seems to crash in SWIG_Python_GetSwigThis:
```
#0  0x00000000004b6824 in PyObject_GetAttr ()
#1  0x00007fffefdb82a8 in SWIG_Python_GetSwigThis (pyobj=<optimized out>) at /tmp/openturns/python/src/openturns/swig_runtime.hxx:2022
#2  0x00007fffefdb8351 in SWIG_Python_ConvertPtrAndOwn (obj=<optimized out>, ptr=ptr@entry=0x7fffffffe018, ty=0x7ffff1435500 <_swigt__p_OT__Mesh>, own=0x0, flags=0) at /tmp/openturns/python/src/openturns/swig_runtime.hxx:2072
```

wsfulton:
Oops! None of our tests break, so I'd like to see a reproducer. What SWIG options are you using? You also seem to be using an optimized build. Can you use a debug build so that the asserts can be used? Also what is the full stack trace?

The only thing I can think of is something to do with module unload/reload, which was something that slightly concerned me about the original patch. I should have mentioned it. Can you put a breakpoint in SWIG_Python_DestroyModule and see if that gets called before the crash? If the module is unloaded, then we really ought to be setting all of the statics related to this patch back to NULL, not just SWIG_This_global, or giving some more thought to module unloading. Which could open up a can of worms as there aren't many/any guarantees by Python in this area.

wsfulton:
Oh yes also, @petrmitrichev, your concerns about SWIG_PYTHON_BUILD_NONE were wisely conservative, but as it turns out unnecessary. SWIG_PYTHON_BUILD_NONE is set right at the top of the generated c/c++ file when the -buildnone option is used. In fact I'm quite intrigued as to why SWIG has this option of redefining `Py_None` to something else. Seems most peculiar and is potentially something to remove in swig-4 as part of our reducing complexity in removing the vast array of options for swig and python.

jschueller:
swig crashes in SWIG_Python_ConvertPtrAndOwn in our following code because the module gets destroyed too early (SWIG_Python_DestroyModule is called once):

```
void * ptr = 0;
  if (SWIG_IsOK(SWIG_ConvertPtr(inputMesh.get(), &ptr, SWIG_TypeQuery("OT::Mesh *"), 0)))
  {
    inputMesh_ = *reinterpret_cast< OT::Mesh * >(ptr);
  }
```
we dont use any python option, I'll try to write a simple example.

jschueller:
Hello again,

I wrote a small example to debug this:
https://github.com/jschueller/swig-field-example

I enabled travis here so you can see the logs:
https://travis-ci.org/jschueller/swig-field-example/builds/399542060

jschueller:
I'm starting to think that the problem is due to the use of SWIG_ConvertPtr in another compilation unit than the module: if I move it into openturns.i (dropping PythonFieldFunction.*xx) it does not segfault with master.

wsfulton:
CMake is obfuscating the commands being run. Could you get CMake to show the commands it is running on Travis so I can work out what is being run?

Run using 'python -v' to see debugging info about module cleanup. Try this to debug and probably fix it:
```diff
diff --git a/Lib/python/pyrun.swg b/Lib/python/pyrun.swg
index d8d2488..3c3a4be 100644
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1398,8 +1398,11 @@ SWIG_Python_DestroyModule(PyObject *obj)
       if (data) SwigPyClientData_Del(data);
     }
   }
+/*
   Py_DECREF(SWIG_This());
   SWIG_This_global = NULL;
+*/
+  printf("SWIG_Python_DestroyModule ..................\n");
 }
 
 SWIGRUNTIME void
@@ -1428,6 +1431,12 @@ SWIG_Python_TypeCache(void) {
   return SWIG_Python_TypeCache_global;
 }
 
+SWIGRUNTIME void
+SWIG_Python_TypeCache_DestroyModule(PyObject *obj)
+{
+  printf("SWIG_Python_TypeCache_DestroyModule ..................\n");
+}
+
 SWIGRUNTIME swig_type_info *
 SWIG_Python_TypeQuery(const char *type)
 {
@@ -1441,7 +1450,7 @@ SWIG_Python_TypeQuery(const char *type)
     swig_module_info *swig_module = SWIG_GetModule(0);
     descriptor = SWIG_TypeQueryModule(swig_module, swig_module, type);
     if (descriptor) {
-      obj = PyCapsule_New((void*) descriptor, NULL, NULL);
+      obj = PyCapsule_New((void*) descriptor, NULL, SWIG_Python_TypeCache_DestroyModule);
       PyDict_SetItem(cache, key, obj);
       Py_DECREF(obj);
     }
```

jschueller:
I enabled verbose builds so you can see the commands on travis

wsfulton:
Seems you are using -external-runtime. I havn't had time to look properly, but I'm not sure if this can be used as a header file given the global static variables which have been introduced with explicit initialization to replace the local static variables lazy initialization. They will need explicit initialization by your own code that uses the code generated by -external-runtime, but I'm not entirely sure this is the correct approach given that this was not previously needed. Needs some more thought, I'm now not sure this patch is a good one when an external runtime is being used. This happens to be highlighted by the similar changes I made to SWIG_This() although your own code using the external runtime happened not to use the functions you changed from being lazily initialized.

jschueller:
Ok. Let me know if you find an answer to this, whether I should use some initialization code, or if you decide to revert it, or if there is another way to convert that pyobject pointer back to a c++ pointer without using SWIG_ConvertPtr.

petrmitrichev:
I've modeled my change after the handling of SwigPyObject_stype variable (https://github.com/swig/swig/blob/f8ffc04215b1be9ade91d418c0e0cf309e1809a2/Lib/python/pyrun.swg#L465, https://github.com/swig/swig/blob/f8ffc04215b1be9ade91d418c0e0cf309e1809a2/Lib/python/pyinit.swg#L407), so consider changing its handling as well if you decide that SWIG_init is not the right place for such initialization.

wsfulton:
Sorry, I havn't got back to this sooner. I just am not sure without looking deeper whether or not your use of the generated code from -external-runtime is valid or not. However, I was planning on reverting the change and going back to the original lazily instantiated singletons implementation. Then make changes to explicitly call the singleton functions during module initialization. So this would result in a combination of the commit and the old way of doing it. You will need to explicitly initialize the singletons in your own code in order to avoid deadlocks as you seem to have a separate runtime.

jschueller:
Ok, I'll see if I I can keep up with your code.

jschueller:
The current master is ok regarding my test-case, thank you.

PR #1282, opened 2018-07-03
Make python shadow sub-modules runnable

rupertnash:
Python modules can be run as main from the command line with the -c flag
Currently this does not work as the python wrapper module does not correctly import the extension module. This commit makes the generated code consistent with PEP 366.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -855,7 +855,7 @@ public:
        */
       Printv(default_import_code, "def swig_import_helper():\n", NULL);
       Printv(default_import_code, tab4, "import importlib\n", NULL);
-      Printv(default_import_code, tab4, "pkg = __name__.rpartition('.')[0]\n", NULL);
+      Printv(default_import_code, tab4, "pkg = __package__ if __package__ else __name__.rpartition('.')[0]\n", NULL);
       Printf(default_import_code, tab4 "mname = '.'.join((pkg, '%s')).lstrip('.')\n", module);
       Printv(default_import_code, tab4, "try:\n", NULL);
       Printv(default_import_code, tab8, "return importlib.import_module(mname)\n", NULL);


wsfulton:
I think you meant the `-m' flag where you mentioned `-c'. 

To check the patch I created an empty file `__init__.py` in one of the examples (Examples/python/class) to make the example a package and then from the Examples/python directory, ran `python -m class.example`. This now runs successfully instead of failing with an ImportError.

@mromberg this is your area of expertise, any comments to add?

rupertnash:
You're right I mean `-m`!

mromberg:
At first glance, this patch looks ok.   But python's import machinery is very complicated.  Even python developers think so:

https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html

  The use of module attributes such as `__package__` need to take into account:
  * That the version of python may not even support packages at all.
   * That the attribute may not even exist
    * The attribute may be an empty string
    * The attribute may be None
    * The attribute may be some non-empty string.
  
  I think swig is finally dropping support for really, really old versions of python.  So, I think we can assume that all swig supported python versions do support packages.  But, I'm way less certain that they all support the `__package__` attribute in the same way.   It may even depend on exactly what loader is used.

https://www.python.org/dev/peps/pep-0366/

   I think that this patch should probably have a check for `__package__` existing before it is used.   And it would also be a very good idea to add some more cases to the swig test suite covering package related imports.   I'm fairly certain that the current tests don't cover all the cases.  Adding a new dependency by using the `__package__` attribute is bound to uncover cases where the swig tests don't cover cases used by clever swig users.

  And lastly, swig supports loading the C wrapper module from either the same package as the shadow module OR from the global namespace.   I think this should be deprecated.   But for historical reasons it needs to be covered.   So, each python package related test should cover both the case where the C module is in the right package and when it is just dumped into the global namespace.

  In summary, I think it would be good to add some more package tests as part of this patch :)

wsfulton:
SWIG 4 will support python 2.7 plus 3.2 and later. We test for all these versions, and the Travis + Appveyor tests all pass (the entire set of tests gets run when a patch is submitted, see the checks below). @mromberg, this includes all the package import tests you added in the Examples/python/import_packages directory.

Surely it is reasonable to expect `__package__` to always be present in the supported versions, PEP 0366 introduced it in 2.6 and 3.0. I don't understand how `__package__` could be completely absent in these versions. The other 3 cases listed by @mromberg are mentioned in PEP 0366, so agreed should be covered.

Regarding extra testing, I'm all for additional tests, but @mromberg, I thought you pretty much had most cases covered (barring `-m` testing). If not, I think you'll be best placed to help set up some more tests.

wsfulton:
#769 and #1105 are related issues.

mromberg:
I may be a bit gun shy after things went off the rails the last time the python import code got a tweak.  This time, swig will only have to support "newish" versions of python.  So, that may help a whole bunch.   I just remember that all the existing tests passed back then too and then we discovered that there were cases the tests did not account for.

  I would not be at all surprised if this were to happen again.   When I worked on the python zip importer, I found that many of these majick python attributes are set by individual code loaders.   So, each developer of a loader has to read these PEPs, understand them and correctly implement each feature.   Needless to say this does not always happen :).

  In any case, the current swig code does need an update here as what is there now is not working in this here.   If I have some time (I'm dealing with some family estate related issues right now) I may be able to add some more package related tests to swig if I can think of areas we may not have covered that may involve the `__package__` attribute.   Otherwise, it may be ok to cross our fingers and go with what we have (and add new test cases if something slips by).

wsfulton:
Right then, I'll merge this patch in soon. As ever, @mromberg, some extra tests when you are ready will be much appreciated.

wsfulton:
This also fixes the renaming of the generated module python file to `__init__.py` in a package directory as reported in #769. 

@mromberg, this is one case that should be additionally tested and added to the documentation. If you can't do this let me know.

wsfulton:
604ae7186bcd9d5b8a2b781d5dcf9d52c0c759a3 fixes this problem for the -builtin option.

@mromberg, you can see I also added in runtime tests for running `python -m` in this commit. There is one strange problem in the from_init1, from_init2, from_init3 tests though which I commented on here: https://github.com/swig/swig/blob/604ae7186bcd9d5b8a2b781d5dcf9d52c0c759a3/Examples/python/import_packages/from_init3/runme.py#L22-L27. If you've any ideas please let me know. Still need tests for renaming the generated module python file to `__init__.py`.

wsfulton:
51dadaeacde6de32ab945ef0c930e1f3d95f2ae0 adds a test to check the pure Python module can be renamed to `__init__.py` to turn the module into a Python package. I havn't documented this approach, but if anyone thinks it is a good idea we could.

PR #1283, opened 2018-07-04
Enable test case for python_pybuf

gmazzamuto:


wsfulton:
SWIG's minimum supported Python version is now 2.7, so the old buffer interface can be removed altogether which I've done in 4313c2c16829da32408b9e69b6a42ae7c8a0b112.

PR #1288, opened 2018-07-16
Fix garbage collection for guile >= 2.0.12

ChristopherHogan:
## The Problem
This [thread](https://lists.gnu.org/archive/html/guile-devel/2016-10/msg00012.html) discusses a change introduced to Guile 2.0.12 that changes the type of every smob to "finalized" just before calling its `free` function.

The `_wrap_delete_*` methods that SWIG generates call `SWIG_Guile_MustGetPtr` and `SWIG_Guile_MarkPointerDestroyed`, both of which end up calling `SWIG_Guile_ConvertPtr`. The issue is that `SWIG_Guile_ConvertPtr` expects an object of type `swig_tag` or `swig_collectable_tag`, but it gets a "finalized" smob instead.

## The Solution
This PR adds `SWIG_Guile_IsValidSmob`, which considers a "finalized" smob valid for the purposes of getting and marking pointers.  `swig_finalized_tag` is created the same way it's done in Guile, in `libguile/smob.c:clear_smobnum`. 

## Reproducer
Using Guile >= 2.0.12, add a call to the garbage collector at the end of `Examples/test-suite/schemerunme/reference_global_vars.scm` before the script exits:
```scheme
;class reference
(var-TestClass (createref-TestClass (new-TestClass 20)))
(if (!= (TestClass-num-get (value-TestClass (var-TestClass))) 20)
  (begin (display "Runtime test 13 failed.\n") (exit 1)))

(gc)

(exit 0)
```
and run
```bash
cd Examples/test-suite/guile && make reference_global_vars.cpptest
```
The test will report:
```
Backtrace:
           1 (primitive-load "/root/swig/Examples/test-suite/guile/.?")
           0 (gc)

ERROR: In procedure gc:
ERROR: In procedure delete-TestClass: Wrong type argument in position 1: #<finalized smob 20ee680>
make: *** [reference_global_vars.cpptest] Error 1
```
With the changes in this PR, the test completes without error.

### DIFF
--- a/Examples/test-suite/schemerunme/reference_global_vars.scm
+++ b/Examples/test-suite/schemerunme/reference_global_vars.scm
@@ -56,4 +56,12 @@
 (if (!= (TestClass-num-get (value-TestClass (var-TestClass))) 20)
   (begin (display "Runtime test 13 failed.\n") (exit 1)))
 
+; Test garbage collection on guile >= 2.0.12
+(if (or (and (string=? (major-version) "2")
+             (string=? (minor-version) "0")
+             (string>=? (micro-version) "12"))
+        (and (string>=? (major-version) "2")
+             (string>=? (minor-version) "1")))
+    (gc))
+
 (exit 0)
--- a/Lib/guile/guile_scm_run.swg
+++ b/Lib/guile/guile_scm_run.swg
@@ -99,6 +99,7 @@ SWIG_Guile_scm2newstr(SCM str, size_t *len) {
 static int swig_initialized = 0;
 static scm_t_bits swig_tag = 0;
 static scm_t_bits swig_collectable_tag = 0;
+static scm_t_bits swig_finalized_tag = 0;
 static scm_t_bits swig_destroyed_tag = 0;
 static scm_t_bits swig_member_function_tag = 0;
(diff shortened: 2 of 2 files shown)


wsfulton:
@ChristopherHogan, could you add that change to `Examples/test-suite/schemerunme/reference_global_vars.scm` to the patch? Does the gc reliably run, or is it like other language's garbage collectors and not guarantee to do anything? Or does the test require guile-2.0.12? If the gc is sure to run, can you code it up so it only runs if the version is >= 2.0.12? The version of Guile on Travis OSX is currerntly 2.2.4, so it'l get tested in this setup.

I'm no guile expert, @gjanssens, I don't suppose you could have a look to see if you think this change is okay?

ChristopherHogan:
@wsfulton, I added the test for guile >= 2.0.12.

> Does the gc reliably run, or is it like other language's garbage collectors and not guarantee to do anything?

It looks like guile's `(gc)` procedure calls `GC_gcollect` from `libgc`, and the docs for that just say "Explicitly force a garbage collection." That sounds to me like it will reliably run.

wsfulton:
Thanks, I'll merge this in.

PR #1290, opened 2018-07-21
Tests for V8 5.2+ support (reloaded)

furylynx:
This is another attempt at getting Travis CI testing multiple node versions for PR #968. The tests run fine except the python 3 and the lua test on OS X that should be unrelated to the changes made (see https://travis-ci.org/furylynx/swig).

**Edit, @wsfulton :**

> @furylynx, could you please write a paragraph summary of the changes in this issue for the merge commit and release notes/changes file?

This pull request provides fixes for deprecated and removed functions that were used by swig to generate javascript wrappers for node (7.x+) and V8 (5.2+).

Additionally, new Travis checks for different node versions are introduced (node 4,6,8 and 10). They make use of the Node Version Manager (v0.33.10) to install a particular node version that was specified for the test. If no particular node version was specified, tests use the nodejs package of the package manager.


wsfulton:
Excellent, nice work, I think this will do. I'll replicate the Travis install instructions on a local box to have a closer look at the build and then I reckon I can merge into master. @furylynx, could you please write a paragraph summary of the changes in this issue for the merge commit and release notes/changes file?

wsfulton:
On my machine, configure finds nodejs, not node:
```
checking for nodejs... nodejs
```
and so the wrong version is used as my nodejs is version v0.10.25. This simply because both are looked for in configure.ac, but in this order:
https://github.com/swig/swig/blob/8f224694e78cd17ecc3098be2f9f36378a54632e/configure.ac#L1508
What is the reason for looking for both? Are node and nodejs the same thing? Was nodejs renamed node at some point? Maybe the correct approach is to also install nodejs?? With an understanding to the background on this, these, then we ought to know what the fix might be.

furylynx:
Yes, in this context nodejs and node should be the same thing. 

> There is a naming conflict with the node package (Amateur Packet Radio Node Program), and the nodejs binary has been renamed from node to nodejs.

I guess your conflict occurs when installing node(js) both from package manager and by using nvm.

wsfulton:
If nodejs is the new name for node, is there not a way to get nvm to install it with the correct name?

furylynx:
The renaming apparently was done for systems potentially having the Amateur Packet Radio Node Program installed (mainly former ubuntu releases, that also provide the [nodejs-legacy](https://packages.ubuntu.com/trusty/all/nodejs-legacy/filelist) package which contains a symlink for node to the nodejs binary). Brew on MacOS only provides a node binary (lacking a nodejs binary). The nodejs package on recent ubuntu releases also provides a node binary and symlinks nodejs to it [[1](https://packages.ubuntu.com/artful/amd64/nodejs/filelist), [2](https://packages.ubuntu.com/bionic/amd64/nodejs/filelist)]. So after all, having only a node binary should be fine.

wsfulton:
Please provide a more accurate summary of these changes if this does not suffice:
https://github.com/swig/swig/blob/15e2bdac7a7357377ecaff074c7e490ebd20d1c4/CHANGES.current#L11

PR #1297, opened 2018-08-02
Fix performance when calling C# virtual methods.

rokups:
Previous implementation used reflection check on every call to a virtual method to verify if it was overridden by a subclass. While this works it is also pretty slow and accumulates fast. This check was replaced by checking if delegate associated with the virtual call is not null. This delegate is set only when subclass overrides a virtual method in question. Behaviour of generated code does not change. There still is a case where base class calling overridden virtual method would call C++ method instead of overriding C# method because call happens before SwigDirectorConnect() is called. This happens already and i do not think there is anything we can or should do. C# advices against calling virtuals from constructors anyway.

P.S. Swig codebase is full of mixed tabs and spaces which breaks code alignment if tab width if anything else than 4 spaces. Would you like a PR fixing that?

Edit:
~~Apparently my initial commit broke director_constructor test. This probably was related to condition mentioned above. So i re-added old check back and added delegate null check before it. This gives us a fast exit when virtual method is overridden. It is still slow if method is not overriden because metadata check will still happen. I am not entirely happy with it. Do you think adding `bool directorsInitialized` and do fast nullcheck when set to true while doing show metadata check otherwise is an acceptable solution?~~

### DIFF
--- a/Doc/Manual/CSharp.html
+++ b/Doc/Manual/CSharp.html
@@ -1537,6 +1537,7 @@ public class Base : global::System.IDisposable {
   internal Base(global::System.IntPtr cPtr, bool cMemoryOwn) {
     swigCMemOwn = cMemoryOwn;
     swigCPtr = new global::System.Runtime.InteropServices.HandleRef(this, cPtr);
+    SwigDirectorSetup();
   }
 
   internal static global::System.Runtime.InteropServices.HandleRef getCPtr(Base obj) {
@@ -1559,12 +1560,15 @@ public class Base : global::System.IDisposable {
   }
 
   public virtual uint UIntMethod(uint x) {
-    uint ret = examplePINVOKE.Base_UIntMethod(swigCPtr, x);
+    uint ret = (swigMethodOverriden0 ? examplePINVOKE.Base_UIntMethodSwigExplicitBar(swigCPtr, x) : examplePINVOKE.Base_UIntMethod(swigCPtr, x));
     return ret;
   }
 
   public virtual void BaseBoolMethod(Base b, bool flag) {
-    examplePINVOKE.Base_BaseBoolMethod(swigCPtr, Base.getCPtr(b), flag);
+    if (swigMethodOverriden1)
+      examplePINVOKE.Base_BaseBoolMethodSwigExplicitBar(swigCPtr, Base.getCPtr(b), flag);
+    else
+      examplePINVOKE.Base_BaseBoolMethod(swigCPtr, Base.getCPtr(b), flag);
     if (examplePINVOKE.SWIGPendingException.Pending)
       throw examplePINVOKE.SWIGPendingException.Retrieve();
   }
@@ -1573,18 +1577,19 @@ public class Base : global::System.IDisposable {
     SwigDirectorConnect();
   }
 
-  private void SwigDirectorConnect() {
-    if (SwigDerivedClassHasMethod("UIntMethod", swigMethodTypes0))
-      swigDelegate0 = new SwigDelegateBase_0(SwigDirectorUIntMethod);
-    if (SwigDerivedClassHasMethod("BaseBoolMethod", swigMethodTypes1))
-      swigDelegate1 = new SwigDelegateBase_1(SwigDirectorBaseBoolMethod);
-    examplePINVOKE.Base_director_connect(swigCPtr, swigDelegate0, swigDelegate1);
+  private void SwigDirectorSetup() {
+    var classType = typeof(Base);
+    swigMethodOverriden0 = examplePINVOKE.SwigDerivedClassHasMethod(this, "UIntMethod", classType, swigMethodTypes0);
+    swigMethodOverriden1 = examplePINVOKE.SwigDerivedClassHasMethod(this, "BaseBoolMethod", classType, swigMethodTypes1);
   }
 
-  private bool SwigDerivedClassHasMethod(string methodName, global::System.global::System.Type[] methodTypes) {
-    System.Reflection.MethodInfo methodInfo = this.GetType().GetMethod(methodName, methodTypes);
-    bool hasDerivedMethod = methodInfo.DeclaringType.IsSubclassOf(typeof(Base));
-    return hasDerivedMethod;
+  private void SwigDirectorConnect() {
+    SwigDirectorSetup();
+    if (swigMethodOverriden0)
+      swigDelegate0 = new SwigDelegateBase_0(SwigDirectorUIntMethod);
+    if (swigMethodOverriden1)
+      swigDelegate1 = new SwigDelegateBase_1(SwigDirectorBaseBoolMethod);
+    examplePINVOKE.Base_director_connect(swigCPtr, swigDelegate0, swigDelegate1);
   }
 
   private uint SwigDirectorUIntMethod(uint x) {
@@ -1603,6 +1608,9 @@ public class Base : global::System.IDisposable {
 
   private static global::System.Type[] swigMethodTypes0 = new global::System.Type[] { typeof(uint) };
   private static global::System.Type[] swigMethodTypes1 = new global::System.Type[] { typeof(Base), typeof(bool) };
+
+  private bool swigMethodOverriden0;
+  private bool swigMethodOverriden1;
 }
 </pre>
 </div>
--- a/Examples/test-suite/director_basic.i
+++ b/Examples/test-suite/director_basic.i
@@ -121,6 +121,10 @@ public:
   static Bar * call_pmethod(MyClass *myclass, Bar *b) {
     return myclass->pmethod(b);
   }
+
+   // Collisions with generated method names
+   virtual void Connect() { }
+   virtual void Setup() { }
 };
 
 template<class T>
--- a/Lib/csharp/boost_intrusive_ptr.i
+++ b/Lib/csharp/boost_intrusive_ptr.i
@@ -285,13 +285,13 @@
   }
 
 // Base proxy classes
-%typemap(csbody) TYPE %{
+%typemap(csbody, directorsetup="\n    SwigDirectorSetup();") TYPE %{
   private global::System.Runtime.InteropServices.HandleRef swigCPtr;
   private bool swigCMemOwnBase;
 
   PTRCTOR_VISIBILITY $csclassname(global::System.IntPtr cPtr, bool cMemoryOwn) {
     swigCMemOwnBase = cMemoryOwn;
-    swigCPtr = new global::System.Runtime.InteropServices.HandleRef(this, cPtr);
+    swigCPtr = new global::System.Runtime.InteropServices.HandleRef(this, cPtr);$directorsetup
   }
 
   CPTR_VISIBILITY static global::System.Runtime.InteropServices.HandleRef getCPtr($csclassname obj) {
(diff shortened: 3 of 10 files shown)


rokups:
Hey @wsfulton you were right in your email, this is a can of worms :)

Turns out my performance fix was not problematic and we indeed can avoid checking metadata completely. Delegate nullcheck is enough.

I managed to replicate issue of failing `director_constructor.i` test and uncovered a bug. `SwigDerivedClassHasMethod()` erroneously returned `true` when subclass had non-overriding virtual method with the same name and types. More details are in the commit message. I believe code that tests `director_constructor.i` does `virtual int test()` instead of `override int test()` and needs fixing, however i was unable to find where and how. Any tips where i could find code that drives `director_constructor.i` test?

wsfulton:
Have a look at http://swig.org/Doc3.0/Extending.html#Extending_test_suite for understanding the test-suite. I'm afraid that getting it to work under Windows requires Cygwin/MinGW or maybe even WSL now. It is much much easier under Linux (even for C#) and takes a few minutes to get the entire dev environment up and running on Ubuntu, see https://github.com/swig/swig/wiki/Getting-Started. 

You can view director_constructor.i online here: https://github.com/swig/swig/blob/master/Examples/test-suite/director_constructor.i. It contains a perfectly valid `virtual int test()=0;`. Note that there is no C# runtime test for this test-case (as there is no Examples/test-suite/csharp/director_constructor_runme.cs file).

rokups:
Mixing tabs and spaces on the same line..? Any idea if any editor does handle it automatically? Docs mention `make beautify-file` command, but using that creates much more adjustments on the lines other than i have modified. I guess bottom line is that working on the code is enough of mental strain already and worrying about when one should insert tab and when space is just too much pointless mental work. Maybe adopting style where tabs and spaces are not mixed on the same line could be something to consider in the future?

rokups:
Ok my previous implementation was complete crap so i redid it. It is a much smaller delta now with an elegant caching of `SwigDerivedClassHasMethod()` result in a boolean stored in a class instance. No more structs, no extra unneeded stuff stored in class instance either. Unlike with previous implementation documentation should need no changes as well.

I now consider this code in a good shape and ready for review.

wsfulton:
Maybe, I missed something, but what is the point of having the new `swigMethodOverriden` members as a cache? They are only used once per class instance as they are set and only used in `SwigDirectorConnect` from the constructor).

Re tabs, I think the convention was normal when it was chosen years ago. The oldest editors around support automated mixed tabs/spaces (vim and emacs). 8 spaces instead of 1 tab is sometimes used, but unfortunately introduces an inconsistency. Probably best to discuss coding convention in a separate issue or the swig-devel mailing list so that this issue doesn't go off-topic.

rokups:
`swigMethodOverriden` is used every time a virtual method is called. SWIG used to generate code like this:

```cs
  public virtual string ping() {
    string ret = SwigDerivedClassHasMethod("ping", swigMethodTypes0) ? director_basicPINVOKE.Foo_pingSwigExplicitFoo(swigCPtr) : director_basicPINVOKE.Foo_ping(swigCPtr));
    return ret;
  }
```
This does expensive metadata check every time virtual is called. This PR changes generated code to:
```cs
  public virtual string ping() {
    string ret = (((swigMethodOverriden0 ?? (swigMethodOverriden0 = SwigDerivedClassHasMethod("ping", swigMethodTypes0))).Value) ? director_basicPINVOKE.Foo_pingSwigExplicitFoo(swigCPtr) : director_basicPINVOKE.Foo_ping(swigCPtr));
    return ret;
  }
```
A nullable type is used for `swigMethodOverriden0` and only very first time method executes `swigMethodOverriden0` will be null, `??` operator will delegate execution to `(swigMethodOverriden0 = SwigDerivedClassHasMethod("ping", swigMethodTypes0))` where metadata check happens and is assigned to `swigMethodOverriden0`.

Note that we can not assign `swigMethodOverriden0` in `SwigDirectorConnect()` (or simply check for `swigDelegate0 != null`) because  `SwigDirectorConnect()` is executed only when user derives from a proxy class. However existence of virtual method needs to be checked in all cases. I tried this approach and it resulted in overflowing stack. You were right saying i am opening a can of worms :)

wsfulton:
Travis has been having lots of network problems over the past few days making it nearly impossible to get a build fully passing. Only the csharp builds are relevant for this patch, and so I've restarted those.

Sorry, I missed the other calls to `SwigDerivedClassHasMethod`. It's been many years since I fully grokked the C# director implementation.

SWIG shouldn't degrade thread-safety in the wrappers, and this doesn't look thread-safe to me. No doubt a thread local variable could fix this, but I wonder if it will provide much improvement then? Can you put together some metrics around the performance improvements for some simple functions that don't do a lot, such as `virtual void f() {}`?  That way, the performance overheads are removed for marshaling of types and code within the function implementation. Probably it needs to include a C++ only inheritance hierarchy as well as one with a C# class.  I'd like to verify the performance improvements myself on different machines if you could kindly make it publicly available somewhere.

The assignment of `hasDerivedMethod` has changed from:
```c#
    bool hasDerivedMethod = methodInfo.DeclaringType.IsSubclassOf(typeof(Employee));
```
to
```c#
    bool hasDerivedMethod = methodInfo.IsVirtual && !methodInfo.IsAbstract && methodInfo.DeclaringType.IsSubclassOf(typeof(Employee)) && methodInfo.DeclaringType != methodInfo.GetBaseDefinition().DeclaringType;
```
I don't see how this is related to improving performance. If anything it looks worse due to the additional introspection calls. Calling `methodInfo.IsVirtual` seems pointless because only virtual methods can be director methods.

rokups:
> Can you put together some metrics around the performance improvements

I will work on that :+1: 

> this doesn't look thread-safe to me

Hmm you are right. I was avoiding more proper implementation where `swigMethodOverriden0` would have to be set in every constructor. Thread safety gives a perfect reason to stop being lazy and do it that way. I will explore this solution after we look at performance numbers.

> The assignment of hasDerivedMethod has changed <...>I don't see how this is related to improving performance.

Ah that is not a performance fix but a bug fix. I left details in commit message. `SwigDerivedClassHasMethod` can be easy to confuse. Subclass may have a method with same name and same types which is:
* `virtual` but not `override`. In C# such method does not override virtual method of parent class. This is handled by `methodInfo.DeclaringType != methodInfo.GetBaseDefinition().DeclaringType` check.
* Not `virtual` and not `override`. Likewise in C# such method does not override virtual method of parent class.
* Is `abstract`. Directors rely on method having an implementation.

Considering `SwigDerivedClassHasMethod` is meant to execute only once per instance performance should not be of any concern.

rokups:
Test code:
```cs
    // warmup hit
    for (int i = 0; i < 1000; i++)
      b.noop();

    Stopwatch sw = new Stopwatch();
    sw.Start();
    int count = 0;
    while (sw.ElapsedMilliseconds < 10000)
    {
      b.noop();
      count++;
    }
    sw.Stop();
    Console.WriteLine($"{count} iterations in 10s");
```
master
5089924 iterations in 10s
4849378 iterations in 10s
5231168 iterations in 10s

Average: 5,056,823

PR
166077096 iterations in 10s
174917708 iterations in 10s
174475341 iterations in 10s

Average: 171,823,381

Speed increase on average: 34 times
Tested on mono 5.12.0 (linux)

rokups:
A little recap on reworked PR:
* `swigMethodOverridenN` is not a plain `bool`.
* `swigMethodOverridenN` booleans are initialized from every constructor (solves threadsafety issue)
* `SwigDerivedClassHasMethod` is modified to take instance and current class type as parameters and moved to PINVOKE class. This method is now shared between all classes instead of every class having their private version.
* Other changes removed from `SwigDerivedClassHasMethod`, will make another PR in the future and add tests then.
* Updated docs to properly reflect implementation of C# directors.

wsfulton:
I'm hesitant to accept these latest changes for these reasons:
1) The patch includes bug fixes when it is meant to be a performance improvement patch. Could you please separate out the SwigDerivedClassHasMethod fixes into a new patch to apply to the current code base and include tests that it fixes? I'd like to apply this simple fix first as it is tangential to the non-trivial performance improvements which are taking a few iterations to get right.
2) This is a rather sensitive area and there is quite a bit of change I don't understand such as the additional code in each virtual method. Please don't move code about in this patch (the stuff you moved to the PINVOKE class), I can't analyse the changes easily!
3) They are a performance degradation. Construction looks slower to me due to the additional PInvoke calls in the constructor. This will be significant for classes with many virtual methods. In many use cases, the virtual methods won't be called and so there is an unnecessary penalty to pay on construction. Please include performance figures for mere construction where the director class is both derived from and not derived from.
4) The typemap changes are not backwards compatible for anyone who has written their own typemaps. While I am willing to accept non-backwards compatible changes in SWIG 4, the reason for the change will require a significant improvement. 

I don't see any performance figures for the latest approach, but I suspect on balance you should go back to your earlier approach and make it thread-safe, that is, lazily calculate and cache whether a method is overridden. I'd like to understand (from performance figures) why this approach was abandoned.

Regarding naming for utility methods/variables that SWIG adds, they need to be prefixed with Swig/swig.

rokups:
Yes i completely understand and i am willing to keep polishing it until it is in acceptable shape :)

For starters i will move that fix out to a new small PR. Keeping PR laser-focused is not my strong side apparently.

> This is a rather sensitive area and there is quite a bit of change I don't understand such as the additional code in each virtual method.

What do you mean? There is no additional code. Only metadata check replaced with a boolean.

> They are a performance degradation. Construction looks slower to me due to the additional PInvoke calls in the constructor.

Ah constructor call looks confusing. `examplePINVOKE.SwigDerivedClassHasMethod` is not actual pinvoke call, its that pure C# method that resided in every class. PINVOKE class was not the best place for it indeed. This is still a costly operation though so performance concern is most likely valid.

> you should go back to your earlier approach and make it thread-safe, that is, lazily calculate and cache whether a method is overridden.

I was not sure how to do that. In previous approach i used nullable boolean (`bool?`) which is actually a heap-allocated object whose reference is stored in the object. `bool` on the other hand is primitive stored straight inside the class. But to make `bool?` access thread-safe we would have to lock it on every call to virtual method. I dont think that is a good idea either. If you have a better idea please do share.

wsfulton:
An enum will be stack allocated, so yes it would make more sense to use an enum for `swigMethodOverridenN`. Changing the enum size to `byte` might be worth considering. 

For thread-safety, you'd need to put a lock around the access to it and do some research on optimal thread-safe usage patterns and publish benchmarks of them with the other implementations you have. I think you will need to implement quite a few different versions and compare them all. I don't think you can attempt to store the data from two different threads as this usually results in undefined behaviour, unless you have some sort of inside knowledge of all CLR implementations and the hardware they are implemented on. You may want to research guaranteed .NET atomic operations instead.

ojwb:
Looks like the `SwigDerivedClassHasMethod` fixes were split off into #1323.

wsfulton:
#1323 containing the SwigDerivedClassHasMethod fixes has been merged. @rokups, if you want to pursue any further performance optimisations please pick up this conversation otherwise close the ticket. Thanks.

ojwb:
No response in nearly 2 months, so I'm going to close.

@rokups If you are interesting in continuing with any of the other changes proposed here, or any other performance optimisations that would be great - please just reply and we can reopen.  It just doesn't seem useful to keep the ticket open indefinitely.

Issue #1258, opened 2018-05-15
Unsupported variadic template usage


diorcety:
In cpp11_variadic_templates.i add the following lines at 70:
```

template <typename... BaseClasses>
void abcd(MultiInherit<BaseClasses...> &h)
```

The following error appears:
Error: Syntax error in input(1)

valparm or valexpr should support something like type PERIOD PERIOD PERIOD

wsfulton:
Okay, this has potential, there are no additional shift/reduce conflicts in the parser, which is good.

Re this change in cpp11_variadic_templates.i:
```swig
%template (Abcd2) Abcd<A,B>;
```
it needs to be
```swig
%template (Abcd1) Abcd<A>;
```
as only one template parameter is supported and I can't see the patch changing this. The Abcd2 code you show actually results in using MultiInherit1 and not MultiInherit2 in the generated code (a runtime test would flush this problem out, but is  easily spotted looking at the generated Java type names). Actually, this testcase is commented out in Examples/test-suite/common.mk as variadic template support is far from working well. 

The docs in the Doc/Manual/CPlusPlus11.html file (or last published at http://swig.org/Doc3.0/CPlusPlus11.html#CPlusPlus11_variadic_templates) will need updating. Those docs only talk about variadic template parameters for template classes, not for template functions. The docs will need an update to match what SWIG supports at some point, but it will be helpful if you describe the improvements in more detail. I think it just allows parsing of the 

The change results in seg faults in a few of the testcases in the test-suite, eg overload_template. 

I'd like the patch broken into two as the parser changes are also adding support for C++ variadic arguments:
```c++
void foo(const char* fmt, ...); // variadic arguments in C
void foo(const char* fmt...); // variadic arguments also in C++
```
A useful reference for this is at http://en.cppreference.com/w/cpp/language/variadic_arguments.

This change should first ensure that no vararg support is broken and the new syntax for parameters in a function (i.e. not template parameters) is working. Then a 2nd patch should build on this one to add variadic template argument support, which will require a few fixes in your patch as it stands.

diorcety:
https://github.com/diorcety/swig/commit/d0e38e99323d23e320bfab91e19a5a329e2faef1#diff-eaef211ed355012097bff4132a2f0168R5147 allows (MultiInherit<BaseClasses...> &h) which doesn't parse currently.
The two other modifications allows variadic template on functions

phcerdan:
[...]

PR #1311, opened 2018-08-20
Resolution issues

diorcety:


wsfulton:
There are various problems with this patch that the continuous integration testing shows. @diorcety, do please run the test-suite before you submit a patch.

There's no description fwith this patch or a reference to an issue describing a problem. Please provide plenty of information about the problem and the fix.

wsfulton:
For reference, I found background info and an earlier analysis of a similar patch in #1258.

PR #1314, opened 2018-08-24
fix python std_unordered_map

wangke1935:


wsfulton:
This adds duplicate code when using std_map.i. 2f31c3e94ecbf6f28b3e9d6ee64b31448e07941d in #1319 has been chosen in preference.

PR #1323, opened 2018-09-06
Improve correctness of SwigDerivedClassHasMethod()

rokups:
Improve correctness of SwigDerivedClassHasMethod() by making sure only methods that have `override` are used connected by director.

C# does not treat `virtual` methods of child classes as overriding (as opposed to c++). In order to override a method it must have `override` specified.

Previous version of this method treated `virtual void foo()` or `void foo()` in a subclass as methods that override virtual method of parent class. This resulted in `SwigDirectorConnect()` creating delegates and connecting them to a native class. Presence of such methods caused needless roundtrip through managed layer while in the end correct native function was called.

This is the old code flow:

```cpp
void SwigDirector_MyClass::nonOverride() {
  if (!swig_callbacknonOverride) {  // 0. swig_callbacknonOverride is set
    MyClass::nonOverride();
    return;
  } else {
    swig_callbacknonOverride();     // 1. this is called because wrapper mistakenly assumes overriding
  }
}

SWIGEXPORT void SWIGSTDCALL CSharp_director_basicNamespace_MyClass_nonOverrideSwigExplicitMyClass(void * jarg1) {
  MyClass *arg1 = (MyClass *) 0 ;

  arg1 = (MyClass *)jarg1;
  (arg1)->MyClass::nonOverride();   // 5. Correct method called in the end
}
```

```cs
  private void SwigDirectornonOverride() {
    nonOverride();                  // 2.
  }

  public virtual void nonOverride() {
    if (SwigDerivedClassHasMethod("nonOverride", swigMethodTypes4))                // 3. This returns `true`
        director_basicPINVOKE.MyClass_nonOverrideSwigExplicitMyClass(swigCPtr);    // 4. Native method of director class called explicitly
    else
        director_basicPINVOKE.MyClass_nonOverride(swigCPtr);
  }

```

This is the new code flow:

```cpp
void SwigDirector_MyClass::nonOverride() {
  if (!swig_callbacknonOverride) {  // 0. swig_callbacknonOverride is not set
    MyClass::nonOverride();         // 1. Calls correct method immediately
    return;
  } else {
    swig_callbacknonOverride();
  }
}
```

### DIFF
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -1953,7 +1953,7 @@ public:
 	Printf(proxy_class_code, "  private bool SwigDerivedClassHasMethod(string methodName, global::System.Type[] methodTypes) {\n");
 	Printf(proxy_class_code,
 	       "    global::System.Reflection.MethodInfo methodInfo = this.GetType().GetMethod(methodName, global::System.Reflection.BindingFlags.Public | global::System.Reflection.BindingFlags.NonPublic | global::System.Reflection.BindingFlags.Instance, null, methodTypes, null);\n");
-	Printf(proxy_class_code, "    bool hasDerivedMethod = methodInfo.DeclaringType.IsSubclassOf(typeof(%s));\n", proxy_class_name);
+	Printf(proxy_class_code, "    bool hasDerivedMethod = methodInfo.IsVirtual && methodInfo.DeclaringType.IsSubclassOf(typeof(%s)) && methodInfo.DeclaringType != methodInfo.GetBaseDefinition().DeclaringType;\n", proxy_class_name);
 	/* Could add this code to cover corner case where the GetMethod() returns a method which allows type
 	 * promotion, eg it will return foo(double), if looking for foo(int).
 	 if (hasDerivedMethod) {
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -1951,9 +1951,32 @@ public:
 	// Only emit if there is at least one director method
 	Printf(proxy_class_code, "\n");
 	Printf(proxy_class_code, "  private bool SwigDerivedClassHasMethod(string methodName, global::System.Type[] methodTypes) {\n");
-	Printf(proxy_class_code,
-	       "    global::System.Reflection.MethodInfo methodInfo = this.GetType().GetMethod(methodName, global::System.Reflection.BindingFlags.Public | global::System.Reflection.BindingFlags.NonPublic | global::System.Reflection.BindingFlags.Instance, null, methodTypes, null);\n");
-	Printf(proxy_class_code, "    bool hasDerivedMethod = methodInfo.IsVirtual && methodInfo.DeclaringType.IsSubclassOf(typeof(%s)) && methodInfo.DeclaringType != methodInfo.GetBaseDefinition().DeclaringType;\n", proxy_class_name);
+	Printf(proxy_class_code, "    global::System.Reflection.MethodInfo[] methodInfos = this.GetType().GetMethods(\n");
+	Printf(proxy_class_code, "        global::System.Reflection.BindingFlags.Public | global::System.Reflection.BindingFlags.NonPublic | global::System.Reflection.BindingFlags.Instance);\n");
+	Printf(proxy_class_code, "    foreach (global::System.Reflection.MethodInfo methodInfo in methodInfos) {\n");
+	Printf(proxy_class_code, "      if (methodInfo.DeclaringType == null)\n");
+	Printf(proxy_class_code, "        continue;\n\n");
+	Printf(proxy_class_code, "      if (methodInfo.Name != methodName)\n");
+	Printf(proxy_class_code, "        continue;\n\n");
+	Printf(proxy_class_code, "      var parameters = methodInfo.GetParameters();\n");
+	Printf(proxy_class_code, "      if (parameters.Length != methodTypes.Length)\n");
+	Printf(proxy_class_code, "        continue;\n\n");
+	Printf(proxy_class_code, "      bool parametersMatch = true;\n");
+	Printf(proxy_class_code, "      for (var i = 0; i < parameters.Length; i++) {\n");
+	Printf(proxy_class_code, "        if (parameters[i].ParameterType != methodTypes[i]) {\n");
+	Printf(proxy_class_code, "          parametersMatch = false;\n");
+	Printf(proxy_class_code, "          break;\n");
+	Printf(proxy_class_code, "        }\n");
+	Printf(proxy_class_code, "      }\n\n");
+	Printf(proxy_class_code, "      if (!parametersMatch)\n");
+	Printf(proxy_class_code, "        continue;\n\n");
+	Printf(proxy_class_code, "      if (methodInfo.IsVirtual && (methodInfo.DeclaringType.IsSubclassOf(typeof(%s))) &&\n", proxy_class_name);
+	Printf(proxy_class_code, "        methodInfo.DeclaringType != methodInfo.GetBaseDefinition().DeclaringType) {\n");
+	Printf(proxy_class_code, "        return true;\n");
+	Printf(proxy_class_code, "      }\n");
+	Printf(proxy_class_code, "    }\n\n");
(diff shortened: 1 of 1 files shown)


Examples/test-suite/csharp/director_basic_runme.cs
```diff
@@ -69,6 +71,14 @@ class MyOverriddenClass : MyClass {
       throw new Exception("null not received as expected");
     return b;
   }
+
+  public bool nonVirtual() {
```
wsfulton: This is not a valid use case, the C# compiler warns:
```
./director_basic_runme.cs(75,15): warning CS0114: `director_basicNamespace.MyOverriddenClass.nonVirtual()' hides inherited member `director_basicNamespace.MyClass.nonVirtual()'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword
director_basic/MyClass.cs(96,23): (Location of the symbol related to previous warning)
```
Please adjust the testcase to be valid. Does the patch fix this invalid use case or a valid use case where the `new` and `override` keywords have been added? If it only fixes invalid C# usage, I'll be inclined to reject it as the C# compiler already shows there is a problem. Also, please add code to the director_basic_runme.cs that calls nonVirtual() which works now and which failed (got stuck in a recursive loop) before the patch.

Examples/test-suite/csharp/director_basic_runme.cs
```diff
@@ -69,6 +71,14 @@ class MyOverriddenClass : MyClass {
...
+    throw new Exception("non-virtual overrides virtual method");
+  }
+
+  public virtual bool nonOverride() {
```
wsfulton: This also is not a valid use case, the C# compiler warns:
```
director_basic/MyClass.cs(96,23): (Location of the symbol related to previous warning)
./director_basic_runme.cs(79,23): warning CS0114: `director_basicNamespace.MyOverriddenClass.nonOverride()' hides inherited member `director_basicNamespace.MyClass.nonOverride()'. To make the current member override that implementation, add the override keyword. Otherwise add the new keyword
director_basic/MyClass.cs(100,23): (Location of the symbol related to previous warning)
```
Similar comments for nonVirtual...
Please adjust the testcase to be valid. Does the patch fix this invalid use case or a valid use case where the `new` and `override` keywords have been added? If it only fixes invalid C# usage, I'll be inclined to reject it as the C# compiler already shows there is a problem. Also, please add code to the director_basic_runme.cs that calls nonOverride() which works now and which failed (got stuck in a recursive loop) before the patch.

Examples/test-suite/csharp/director_basic_runme.cs
wsfulton: I think you need to change the C++ nonVirtual method into nonVirtual1 and add nonVirtual2 so that you have two different non-virtual methods to add to MyOverriddenClass, one should use `override` and the other `new`. Likewise for nonOverride.

Examples/test-suite/csharp/director_basic_runme.cs
rokups: "invalid usecase" is an overstatement. Compiler does not output an error after all, just a warning. Adding `new` keywords here is a right thing to do, but it only silences warnings and makes code more explicit. In the end compiled result is all the same and presence of `new` keyword does not affect compiled output. In short - patch does make a valid usecase more correct.

Examples/test-suite/csharp/director_basic_runme.cs
rokups: Hmm you probably mean that these additions would allow doing a test without `%typemap(cscode) MyClass` inject. Well that is a good idea!

rokups: Err on the second thought it wont work because methods can not be overloaded based on these keywords. In this case in mot sure what would be achieved by renaming of these methods. Could you please elaborate?

Examples/test-suite/csharp/director_basic_runme.cs
wsfulton: Warnings are often errors. In this case I don't know what the C# compiler is assuming when neither `new` nor `override` is present. It must be one or the other, do you know? Whichever it is, adding tests that add warnings to the SWIG test-suite is undesirable, so let's make the test explicit and test for both `new` and `override`.

Examples/test-suite/csharp/director_basic_runme.cs
wsfulton: The idea is to test both `override` and `new` in the user's derived class.

Examples/test-suite/csharp/director_basic_runme.cs
rokups: When neither `new` nor `override` is present a method in subclass hides a method in parent class. If instance variable is of subclass type it would call method defined in a subclass, however if instance variable is of superclass type it would call method defined in superclass. Indeed having both `new` and `override` does not make much sense and compiler outputs an error.

```cs
class Parent
{
    public virtual int foo1() { return 1; }
    public virtual int foo2() { return 2; }
    public virtual int foo3() { return 3; }
    public virtual int foo4() { return 4; }
}

class Child : Parent
{
    public int foo1() { return 10; }
    public new int foo2() { return 20; }
    public virtual int foo3() { return 30; }
    public override int foo4() { return 40; }
}

var child = new Child();
var parent = (Parent)child;

child.foo1();    // returns 10. Possibly unintentionally hidden method of parent class, produces warning
child.foo2();    // returns 20. Hiding of method in parent class is definitely intended, no warning.
child.foo3();    // returns 30. Does not override method in parent class.
child.foo4();    // returns 40. Does override method in parent class.

parent.foo1();    // returns 1
parent.foo2();    // returns 2
parent.foo3();    // returns 3
parent.foo4();    // returns 40
```

> let's make the test explicit and test for both `new` and `override`.

`new` just removes a warning. Nothing to test here. `override` is already tested by other testcases. This patch deals with method hiding. I believe all cases relevant to the patch are already covered.

Examples/test-suite/csharp/director_basic_runme.cs
rokups: I dont follow. Swig already generates code that calls right method, however it takes a slow path. Patch makes it generate code that takes a fast path when method hiding is intended. I dont see how this timing can be tested without doing something unreliable/complex.

Examples/test-suite/csharp/director_basic_runme.cs
wsfulton: Thanks, nice detailed explanation. Your change to the testcase is helpful too.

Examples/test-suite/csharp/director_basic_runme.cs
wsfulton: You've removed the warning and replaced with good C# code, so am okay with this change. There are lots of ways that the director code can be called. I've used some clues about what you said and replicated the fast/slow path and added this via the calls to `call_nonVirtual` in commit  b90e3ae35d9a9c6ccb5119cb4b84b288baa75443. That's what I was looking for to make the review of exactly what you are fixing easier.

wsfulton:
I nearly merged this patch. For now I have committed your tests from this patch into master. 

I thought about possible regressions not yet covered in the test-suite and I have added one such case in commit b90e3ae35d9a9c6ccb5119cb4b84b288baa75443. It's where a user overloads a method with a mix of override and new. Your patch will need to address the regression. The main problem with your patch is that it only considers the possibility of one method where there is no overloading. There is also a similar problem (not added as a test though) where the wrapped director C++ function is overloaded or has default arguments and the user's C# class does not override all the methods. I'm not entirely sure what the best approach is to be honest.

rokups:
Hey, did you have a chance to look at the fix?

wsfulton:
Sorry, not yet, but I will.

ojwb:
@wsfulton Looks like this is just awaiting a final review from you (and is probably blocking #1297 which it seems it was split out from).

wsfulton:
Sorry, this fell off my radar. It took quite some understanding it again, but looks okay as the new version of `SwigDerivedClassHasMethod` now detects when a C++ virtual method is not overridden in a C#'s derived class.

The new implementation does a parameter comparison, so covers C++ methods with default parameters as separate C# methods are generated for each defaulted argument.

Issue #1116, opened 2017-10-06
Enums with hex values parsed incorrectly for R


davidcreasy:
If I specify a value for enums using 'hex', then the enumerations all have a value of 0.
The same file works OK with Perl, Java, C# and Python, so I think it's just a bug for R?
Simple example:

// enums.h
class testEnum {
    public:
        enum SHAPES {
          SQUARE,
          TRIANGLE,
          RECTANGLE
        };
        enum COLOURS {
          RED   = 0x00FF0000,
          GREEN = 0x0000FF00,
          BLUE  = 0x000000FF
        };
        enum NUMBERS {
            ONE = 1,
            TWO = 2,
            THREE = 3
        }
        testEnum(bool notEmpty) { 
        }
    private:    
};

And a standard .i file:
/* File : enums.i */
%module enums

%{
#include "\tmp\enums.h"
%}

/* Let's just grab the original header file here */
%include "\tmp\enums.h"

SWIG'd with: 
swig.exe  -c++ -r  enums.i

Produces this in the .R file:
defineEnumeration('_testEnum__SHAPES',
                    .values = c(
                        'SQUARE' = 0, 
                        'TRIANGLE' = 1, 
                        'RECTANGLE' = 2
))
defineEnumeration('_testEnum__COLOURS',
                    .values = c(
                        'RED' = 0, 
                        'GREEN' = 0, 
                        'BLUE' = 0
))
defineEnumeration('_testEnum__NUMBERS',
                    .values = c(
                        'ONE' = 1, 
                        'TWO' = 2, 
                        'THREE' = 3
))

ojwb:
It is indeed R-specific.  Interestingly your example actually seems to work on Linux, presumably due to differences in `sscanf`'s handling of `%lf`.  However, the handling here is buggy in several ways.

In `Sources/Modules/r.cxx` this is the code:

```
  Printv(scode, "defineEnumeration('", mangled_tdname, "'",
         ",\n",  tab8, tab8, tab4, ".values = c(\n", NIL);

  Node *c;
  int value = -1; // First number is zero
  for (c = firstChild(n); c; c = nextSibling(c)) {
    //      const char *tag = Char(nodeType(c));
    //      if (Strcmp(tag,"cdecl") == 0) {
    name = Getattr(c, "name");
    String *val = Getattr(c, "enumvalue");
    if(val && Char(val)) {
      int inval = (int) getNumber(val);
      if(inval == DEFAULT_NUMBER)
        value++;
      else
[...]

PR #1328, opened 2018-09-19
R enums2018

richardbeare:
An implementation of R enumerations initialised by calls to C/C++ code, as discussed in #1116.

As mentioned by @wsfulton, calling wrapper functions to determine the underlying integer value is the only reliable approach. The project that originally prompted this work is SimpleITK, which features an extremely [complex enumerations](https://github.com/SimpleITK/SimpleITK/blob/e582aa5ee0539b48aeda32596cd1abf5e790ecd9/Code/Common/include/sitkPixelIDValues.h#L90).

This PR is a first step. The problems I'm aware of, and looking into are as follows:

Namespaces are mangled into the R side enumeration structures. These structures are hidden variables in a special environment designed to implement a kind of type-safety, as R doesn't have a native enumeration type. Namespaces are also included in certain type-tags, used by some of the R code. These tags are also related to code that isn't user visible.

These naming issues aren't new problems - they have always been part of the R enumeration binding structure

A walk-through to illustrate how it operates (using SimpleITK).

The following function is defined to create a dictionary-like structure in a hidden
environment. The name has a .\_\_ prefix to hide it:

```r
defineEnumeration =
function(name, .values, where = topenv(parent.frame()), suffix = "Value")
{
   # Mirror the class definitions via the E analogous to .__C__
  defName = paste(".__E__", name, sep = "")
  delayedAssign(defName,  .values, assign.env = where)

  if(nchar(suffix))
    name = paste(name, suffix, sep = "")

  setClass(name, contains = "EnumerationValue", where = where)
}
```
This PR adds the `delayedAssign` call to enable calls to this function to be made before the shared library is loaded. Loading the shared library usually happens after the main file is sourced and this trick makes it easier to include swig code in a package.

Two conversion functions are provided - note that both require a type, which we will see is derived from the C++ type, to determine which environment to use.

```r
enumToInteger <- function(name,type)
{
   if (is.character(name)) {
   ans <- as.integer(get(paste(".__E__", type, sep = ""))[name])
   if (is.na(ans)) {warning("enum not found ", name, " ", type)}
   ans
   } 
}

enumFromInteger =
function(i,type)
{
  itemlist <- get(paste(".__E__", type, sep=""))
  names(itemlist)[match(i, itemlist)]
}
```

Functions that use enums as arguments:

``` r
`CastImageFilter_SetOutputPixelType` = function(self, pixelID)
{
  if (inherits(self, "ExternalReference")) self = slot(self,"ref") 
  pixelID = enumToInteger(pixelID, "_itk__simple__PixelIDValueEnum"); 
  
  if(length(pixelID) > 1) {
    warning("using only the first element of pixelID");
  };
  
  ;ans = .Call('R_swig_CastImageFilter_SetOutputPixelType', self, pixelID, PACKAGE='SimpleITK');
  
  if (is.null(ans)) {
    return(invisible(ans))
  } else {
# should be a NULL - otherwise an exception
    stop("Exception in SITK - check warnings")
  }
  
  ans
  
}

attr(`CastImageFilter_SetOutputPixelType`, 'returnType') = 'void'
attr(`CastImageFilter_SetOutputPixelType`, "inputTypes") = c('_p_itk__simple__CastImageFilter', 'character')
class(`CastImageFilter_SetOutputPixelType`) = c("SWIGFunction", class('CastImageFilter_SetOutputPixelType'))

```

or return enums, call these function, with the code being constructed via typemaps.

```r
`Image_GetPixelID` = function(self, .copy = FALSE)
{
  if (inherits(self, "ExternalReference")) self = slot(self,"ref") 
  ;ans = .Call('R_swig_Image_GetPixelID', self, as.logical(.copy), PACKAGE='SimpleITK');
  ans = enumFromInteger(ans, "_itk__simple__PixelIDValueEnum");
  
  ans
  
}
attr(`Image_GetPixelID`, 'returnType') = 'character'
attr(`Image_GetPixelID`, "inputTypes") = c('_p_itk__simple__Image')
class(`Image_GetPixelID`) = c("SWIGFunction", class('Image_GetPixelID'))

```
The various attributes get used for overloading.

That completes the overview, now for the questions.

Is the presence of namespaces in the various type tags in R a problem? They are, as far as I know, not user visible (or at least not obvious). I've been looking into ways of removing them, and it is non-trivial as they are derived from the types in the swig parse tree at points where namespace information isn't always available. I'm still working on this to see if there is a way around it, but let me know if you think it is unnecessary.

### DIFF
--- a/Doc/Manual/R.html
+++ b/Doc/Manual/R.html
@@ -189,9 +189,46 @@ of the proxy class baggage you see in other languages.
 
 
 <p>
-enumerations are characters which are then converted back and forth to
-ints before calling the C routines. All of the enumeration code is
-done in R.
+R doesn't have a native enumeration type. Enumerations are represented
+as character strings in R, with calls to R functions that convert back
+and forth between integers.
+
+The details of enumeration names and contents are stored in hidden R
+environments, which are named according the the enumeration name - for
+example, an enumeration colour:
+<div class="code"><pre>
+enum colour { red=-1, blue, green = 10 };
+</pre></div>
+
+will be initialized by the following call in R:
+<div class="code"><pre>
+defineEnumeration("_colour",
+ .values=c("red" = .Call('R_swig_colour_red_get',FALSE, PACKAGE='enum_thorough'),
+"blue" = .Call('R_swig_colour_blue_get',FALSE, PACKAGE='enum_thorough'),
+"green" = .Call('R_swig_colour_green_get',FALSE, PACKAGE='enum_thorough')))
+
+</pre></div>
+
+which will create an environment named <tt>.__E___colour</tt>. The enumeration
+values are initialised via calls to C/C++ code, allowing complex
+values for enumerations to be used. Calls to the C/C++ code require
+the compiled library to be loaded, so a  <tt>delayedAssign</tt> is employed
+within <tt>defineEnumeration</tt> in order to allow the code to be easily used in R
+packages.
+
+The user typically does not need to access the enumeration lookup
+functions or know the name of the enumeration type used by
+R. Attributes containing the type information are attached by swig to
+functions requiring enumeration arguments or returning enumeration
+values, and those attributes are used to identify and access the
+appropriate environments and thus translate between characters
+and integers.
+
+The relevant functions, for debugging purposes, are <tt>enumToInteger</tt> and
+</tt>enumFromInteger</tt>.
+
+Anonymous enumerations are not supported.
+
 </p>
 
 </body>
--- a/Doc/Manual/R.html
+++ b/Doc/Manual/R.html
@@ -227,7 +227,9 @@ and integers.
 The relevant functions, for debugging purposes, are <tt>enumToInteger</tt> and
 </tt>enumFromInteger</tt>.
 
-Anonymous enumerations are not supported.
+Anonymous enumerations are ignored by the binding generation process,
+leaving no way of accessing the value of anonymous enumerations from R
+code.
 
 </p>
 
--- a/Source/Modules/r.cxx
+++ b/Source/Modules/r.cxx
@@ -256,6 +256,7 @@ static void replaceRClass(String *tm, SwigType *type) {
   String *tmp = getRClassName(type);
   String *tmp_base = getRClassName(type, 0);
   String *tmp_ref = getRClassName(type, 1, 1);
+
   Replaceall(tm, "$R_class", tmp);
   Replaceall(tm, "$*R_class", tmp_base);
   Replaceall(tm, "$&R_class", tmp_ref);
@@ -579,7 +580,6 @@ String * R::createFunctionPointerHandler(SwigType *t, Node *n, int *numArgs) {
   for (i = 0; p; p = nextSibling(p), ++i) {
     String *arg = Getattr(p, "name");
     String *lname;
-
     if (!arg && Cmp(Getattr(p, "type"), "void")) {
       lname = NewStringf("arg%d", i+1);
       Setattr(p, "name", lname);
@@ -640,6 +640,9 @@ String * R::createFunctionPointerHandler(SwigType *t, Node *n, int *numArgs) {
       }
       Replaceall(tm, "$1", name);
       Replaceall(tm, "$result", "r_tmp");
+      if (debugMode) {
+        Printf(stdout, "Calling Replace A: %s\n", Getattr(p,"type"));
+      }
       replaceRClass(tm, Getattr(p,"type"));
       Replaceall(tm,"$owner", "0");
       Delete(lstr);
@@ -785,6 +788,11 @@ int R::cDeclaration(Node *n) {
 ***/
 int R::top(Node *n) {
   String *module = Getattr(n, "name");
+
+  if (debugMode) {
+    Printf(stdout, "<Top> %s\n", module);
+  }
+
   if(!Rpackage)
     Rpackage = Copy(module);
   if(!DllName)
@@ -1225,7 +1233,12 @@ int R::enumDeclaration(Node *n) {
     String *name = Getattr(n, "name");
     ename = getRClassName(name); 
     if (debugMode) {
-      Printf(stdout, "enumDeclaration: %s, %s, %s, %s\n", name, symname, nspace, ename);
+      Node *N = getCurrentClass();
+      String *cl = NewString("");
+      if (N) {
+        cl=getEnumClassPrefix();
+      }
+      Printf(stdout, "enumDeclaration: %s, %s, %s, %s, %s\n", name, symname, nspace, ename, cl);
     }
     Delete(name);
     // set up a call to create the R enum structure. The list of
@@ -1286,6 +1299,9 @@ int R::enumvalueDeclaration(Node *n) {
   {
     // Wrap C/C++ enums with constant integers or use the typesafe enum pattern
     SwigType *typemap_lookup_type = parent_name ? parent_name : NewString("enum ");
+    if (debugMode) {
+      Printf(stdout, "Setting type: %s\n", Copy(typemap_lookup_type));
+    }
     Setattr(n, "type", typemap_lookup_type);
     
     // Simple integer constants
@@ -1311,19 +1327,27 @@ int R::enumvalueDeclaration(Node *n) {
 **************************************************************/
 int R::variableWrapper(Node *n) {
   String *name = Getattr(n, "sym:name");
-
+  if (debugMode) {
+    Printf(stdout, "variableWrapper %s\n", n);
+  }
   processing_variable = 1;
   Language::variableWrapper(n); // Force the emission of the _set and _get function wrappers.
   processing_variable = 0;
 
 
   SwigType *ty = Getattr(n, "type");
+  String *nodeType = nodeType(n);
   int addCopyParam = addCopyParameter(ty);
 
   //XXX
   processType(ty, n);
 
-  if(!SwigType_isconst(ty)) {
+  if (nodeType && !Strcmp(nodeType, "enumitem")) {
+    /* special wrapper for enums - don't want the R _set, _get functions*/
+    if (debugMode) {
+      Printf(stdout, "variableWrapper enum branch\n");
+    }
+  } else if(!SwigType_isconst(ty)) {
     Wrapper *f = NewWrapper();
     Printf(f->def, "%s = \nfunction(value%s)\n{\n",
 	   name, addCopyParam ? ", .copy = FALSE" : "");
@@ -1341,6 +1365,8 @@ int R::variableWrapper(Node *n) {
   return SWIG_OK;
 }
 
+/*************************************************************
+**************************************************************/
 
 void R::addAccessor(String *memberName, Wrapper *wrapper, String *name,
 		    int isSet) {
@@ -1787,6 +1813,9 @@ int R::functionWrapper(Node *n) {
       SwigType *resolved =
 	SwigType_typedef_resolve_all(resultType);
       if (expandTypedef(resolved)) {
+        if (debugMode) {
+          Printf(stdout, "Setting type: %s\n", resolved);
+        }
 	Setattr(p, "type", Copy(resolved));
       }
     }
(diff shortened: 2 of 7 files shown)


hjmjohnson:
@richardbeare As always, nice work Richard.

Source/Modules/r.cxx
```diff
@@ -1199,75 +1214,111 @@ int R::OutputArrayMethod(String *className, List *el, File *out) {
...
+    // This will have content if the %nspace feature is set on
+    // the input file
+    String *nspace = Getattr(n, "sym:nspace"); // NSpace/getNSpace() only works during Language::enumDeclaration call
+    String * ename;
 
-  if(!tdname || Strcmp(tdname, "") == 0) {
+    String *name = Getattr(n, "name");
+    ename = getRClassName(name); 
+    if (debugMode) {
+      Node *N = getCurrentClass();
```
wsfulton: N is not a coding convention we use. Use a proper variable name, such as 'current_class'.

Source/Modules/r.cxx
```diff
@@ -1199,75 +1214,111 @@ int R::OutputArrayMethod(String *className, List *el, File *out) {
...
+      String *cl = NewString("");
+      if (N) {
+        cl=getEnumClassPrefix();
```
wsfulton:
```suggestion
        cl = getEnumClassPrefix();
```

Source/Modules/r.cxx
```diff
@@ -1199,75 +1214,111 @@ int R::OutputArrayMethod(String *className, List *el, File *out) {
...
+      }
+      Printf(stdout, "enumDeclaration: %s, %s, %s, %s, %s\n", name, symname, nspace, ename, cl);
+    }
+    Delete(name);
+    // set up a call to create the R enum structure. The list of
+    // individual elements will be built in enum_code
+    enum_values=0;
```
wsfulton:
```suggestion
    enum_values = 0;
```

Source/Modules/r.cxx
```diff
@@ -1199,75 +1214,111 @@ int R::OutputArrayMethod(String *className, List *el, File *out) {
...
+    // Emit each enum item
     Language::enumDeclaration(n);
-    return SWIG_OK;
+      
+    Printf(enum_def_calls, "defineEnumeration(\"%s\",\n .values=c(%s))\n\n",
+           ename, enum_values);
```
wsfulton: Should be one line

Source/Modules/r.cxx
```diff
@@ -1199,75 +1214,111 @@ int R::OutputArrayMethod(String *className, List *el, File *out) {
...
+    Delete(ename);
+    //Delete(symname);
+  }
+  return SWIG_OK;
+}
+/*************************************************************
+**************************************************************/
+int R::enumvalueDeclaration(Node *n) {
+  if (getCurrentClass() && (cplus_mode != PUBLIC)) {
+    Printf(stdout , "evd: Not public\n");
```
wsfulton:
```suggestion
    Printf(stdout, "evd: Not public\n");
```

Source/Modules/r.cxx
```diff
@@ -1199,75 +1214,111 @@ int R::OutputArrayMethod(String *className, List *el, File *out) {
...
-  String *mangled_tdname = SwigType_manglestr(tdname);
-  String *scode = NewString("");
-
-  Printv(scode, "defineEnumeration('", mangled_tdname, "'",
-	 ",\n",  tab8, tab8, tab4, ".values = c(\n", NIL);
+  Swig_require("enumvalueDeclaration", n, "*name", "?value", NIL);
+  String *symname = Getattr(n, "sym:name");
+  String *value = Getattr(n, "value");
+  String *name = Getattr(n, "name");
+  Node   *parent = parentNode(n);
```
wsfulton:
```suggestion
  Node *parent = parentNode(n);
```

wsfulton:
Hi @richardbeare. I have put in some comments about code formatting, which as a future maintainer, you need to get learn and get right please. 

The patch does not have any tests for the fixes. This is critical too for accepting patches. As this changes a around enums, can you please port the enum_thorough runtime testcase to R. Either use Java, C#, perl or D as the one to convert to R as these are very thorough.

Can you also please add in documentation for this pull request to R.html as no doubt it will be needed and a summary in CHANGES.current.

richardbeare:
Thanks for the reviews, I'll get onto it asap. You are correct - there are many non standard parts in r.cxx, and most of the tests only check that the R code is syntactically correct, without actually running it.

Source/Modules/r.cxx
```diff
@@ -1307,6 +1366,8 @@ int R::variableWrapper(Node *n) {
   return SWIG_OK;
 }
 
+/*************************************************************
+**************************************************************/
```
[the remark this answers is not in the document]
wsfulton: Can you please use our comment formatting approach  where each line begins '` * `'. Here and elsewhere where you have made changes.

wsfulton:
When you've added the documentation, I'll merge this over. Please be aware that we will be releasing SWIG 4.0 this month, so time is running out to get this done. If you need help, let me know. 

const enum reference support is missing and in the testcase you mentioned that you need to put in const reference typemaps. Are you planning that next? Look for 'const enum SWIGTYPE &' typemaps in the other languages. You ought to be able to copy the R 'enum SWIGTYPE' typemaps and adapt slghtly. For the 'in' typemap, just add in a temp var to hold the enum value.

richardbeare:
OK,
What is the correct variable to use to define where the scripts are run? $(SCRIPTDIR)/$(SCRIPTPREFIX) appears to reference the source location of the scripts, which isn't where the .Rout files get written. Thus I was getting a failure from cat not finding the file.

wsfulton:
Replace `$(SCRIPTDIR)` with `.` in the two instances of `|| (cat $(SCRIPTDIR)/$(SCRIPTPREFIX)$${f}$(SCRIPTSUFFIX)out ; false)`

richardbeare:
OK, easy, expecting some cmake style CURRENT_BINARY_DIR  variables.

Where is the place to edit documentation?

wsfulton:
Doc/Manual/R.html

richardbeare:
Makefile fixed and some enumeration documentation added.

Struggling with enum references. Symptom is the type that gets deduced and used as attributes on the R side includes _p in the name, so calls to enumToInteger are given the wrong typename. There doesn't appear to be a way of fixing it in the scoercein typemap families.

Perhaps an in typemap will fix things by using the temporary variable, but I haven't found the right part of another language to use as a model yet.

wsfulton:
I've added this to the swig-4.0 milestone so we need to get this sorted quickly. For enum references, I just want to check that you are only attempting to get the const enum references working? Non-const should be wrapped as a pointer with _p in the name. For each `enum SWIGTYPE` typemap, have you provided an equivalent `const enum SWIGTYPE &` typemap where the latter provides wrappers as if passed by value?

richardbeare:
Most of that side of things haven't been changed by this patch, so I'm having to find my way through it. The current problem is that some wrapping code is created, but the tags used on the R side to identify which hidden environment to use to convert between integers and strings are wrong.

wsfulton:
Is it this line that is not working in R::enumDeclaration:
```c++
    ename = getRClassName(name); 
```

richardbeare:
`getRClassName` is used everywhere. I think may be part of the issue here, but not the entire story. 

Initially I thought that an `rtype` typemap would do the trick. Those typemaps are responsible for creating the tags used to dispatch to the correct lookup function. There are a couple of options for manipulating the RClass keyword, but none of them did the trick. I don't think it is necessary to change that code. Instead I think the idea is to have the right typemap closer to the C end, as you suggested earlier, I think, so that R sees an enum rather than enum &, then the rest should work as expected. I just need to figure out where that is.

richardbeare:
Here's an example of the problem. Below is an example of one of the functions created in the enum_thorough test. The issue is the name of the class in enumToInteger calls, which are created via an scoercein typemap and the enumFromInteger call created via the scoerceout typemap. The name in quotes is used to look up a hidden environment in R that contains the enumeration values, but the correct name doesn't include the _p prefix. Thus there's an R side error when this is executed.
``` r
`SpeedClass_speedTest7` = function(self, s, .copy = FALSE)
{
  if (inherits(self, "ExternalReference")) self = slot(self,"ref") 
  s = enumToInteger(s, "_p_SpeedClass__speed"); 
  ;ans = .Call('R_swig_SpeedClass_speedTest7', self, s, as.logical(.copy), PACKAGE='enum_thorough');
  ans = enumFromInteger(ans, "_p_SpeedClass__speed");
  
  ans
  
}
```

The name is constructed via the `replaceRClass` function, which takes a typemap and a type. The call for the scoercein case above:

typemap:  s = enumToInteger(s, "$R_class")
the following options are created via calls to getRClassName: _p_SpeedClass__speed : _p_SpeedClass__speed : _p_p_SpeedClass__speed.

However getRClassName looks to have been hacked around in the past. There's a middle argument that is't used and a manual appending of _p if upref is requested.

It looks to me as though the support for syntax like "$&R_class" isn't there, and should be added.

I've added some typemaps to support the lower level C/C++ parts, and they seem OK. 

Currently the options for "$R_class" and "$*R_class" are the same, which I think is the problem. As far as I can tell, "$*R_class" isn't used in any of the existing typemaps, so I suspect this part was never completed.

@wsfulton  Thoughts?

richardbeare:
@wsfulton - I have extended the patch to fix the reference problem. I suspect my change may repair a very old bug. There's more work to be done to verify whether the change can replace the problem function in all scenarios, but I suspect it can. As far as I can tell, most of the difference is around dereferencing the name in the R_class, and there are no typemaps where this is done, apart from the one I just added.

richardbeare:
The modifications are able to correctly generate the SimpleITK R binding. One of the enumerations used in SimpleITK prompted this work over 5 years ago! This SimpleITK [PR](https://github.com/SimpleITK/SimpleITK/pull/697) is a place holder for the work, with the resulting R package passing all tests.

wsfulton:
@richardbeare did you see the Travis failure:
```
cpp11_director_enums_wrap.cpp:1458:35: error: cannot convert ‘Color’ to ‘int’ for argument ‘1’ to ‘SEXPREC* Rf_ScalarInteger(int)’
   r_ans = Rf_ScalarInteger(*result); 
```
Locally, you need to run: `./configure enable-cpp11-testing` to turn on these c++11 tests.

Doc/Manual/R.html
```diff
@@ -189,9 +189,46 @@ <H2><a name="R_nn7">33.7 Enumerations</a></H2>
...
+R. Attributes containing the type information are attached by swig to
+functions requiring enumeration arguments or returning enumeration
+values, and those attributes are used to identify and access the
+appropriate environments and thus translate between characters
+and integers.
+
+The relevant functions, for debugging purposes, are <tt>enumToInteger</tt> and
+</tt>enumFromInteger</tt>.
+
+Anonymous enumerations are not supported.
```
wsfulton: They generate code that compiles. In what way are they not supported? Are the symbols introduced by anonymous enums not available at all? Other languages wrap them as simple numeric values.

richardbeare: As far as I know, the R module has always ignored them, but I will check further. There's certainly no direct way of implementing them via the R environment approach used by the rest of the enum framework.

richardbeare: @wsfulton - regarding the cpp11 problems - I think the issue is that I need to do
```cpp
r_ans = Rf_ScalarInteger(static_cast<int>(*result)); 
```

However I presume I should only turn that on for cpp11 - what is the appropriate set of guards to use in `r.swg`?

richardbeare: I have just confirmed that the old/original R enumeration framework ignored the anonymous enumerations and doesn't provide any way of accessing them. This PR, which is mostly about correctly initialising values for enumerations, doesn't change that. I will update the doc to be clearer.

Not sure at the moment how tricky it is to do something more useful - might be able to get a basic variable wrapper approach to do something.

wsfulton: Okay, so anonymous enums ought to be wrapped by the equivalent to an R read-only integer variable / constant.

wsfulton: The `const enum SWIGTYPE&` needs modifying with the cast. The cast can always be used and I've made a commit to fix this in 39e3c3b9d774a483fbec8f0fcf6c136833a79bf8. I'm merging this in shortly, so thanks for sorting this all out.

PR #1334, opened 2018-10-04
Set thread name when attaching to Android JVM in SWIG Java director

hasinoff:
Otherwise the native thread name gets clobbered with "Thread-XX", which
makes debugging hard. This issue seems to be specific to the Android JVM.
Someone proposed a similar fix to SWIG:
https://github.com/swig/swig/pull/831
but abandoned it.

### DIFF
--- a/Lib/java/director.swg
+++ b/Lib/java/director.swg
@@ -6,7 +6,12 @@
  * ----------------------------------------------------------------------------- */
 
 #if defined(DEBUG_DIRECTOR_OWNED) || defined(DEBUG_DIRECTOR_EXCEPTION)
+#include <errno.h>
+#include <string.h>
 #include <iostream>
+#endif
+#if defined(__ANDROID__)
+#include <pthread.h>
 #endif
 
 #include <exception>
@@ -139,12 +144,26 @@ namespace Swig {
         void **jenv = (void **)&jenv_;
 #endif
         env_status = director_->swig_jvm_->GetEnv((void **)&jenv_, JNI_VERSION_1_2);
+        JavaVMAttachArgs args;
+        args.version = JNI_VERSION_1_2;
+        args.group = NULL;
+        args.name = NULL;
+#if defined(__ANDROID__)
+        char thread_name[16];  // MAX_TASK_COMM_LEN=16 is hard-coded in the kernel.
+        if (pthread_getname_np(pthread_self(), thread_name, sizeof(thread_name)) == 0) {
+          args.name = thread_name;
+        } else {
+#if defined(DEBUG_DIRECTOR_OWNED)
+          std::cout << "JNIEnvWrapper: Couldn't get thread name: " << strerror(errno) << std::endl;
+#endif
+        }
+#endif
 #if defined(SWIG_JAVA_ATTACH_CURRENT_THREAD_AS_DAEMON)
         // Attach a daemon thread to the JVM. Useful when the JVM should not wait for
         // the thread to exit upon shutdown. Only for jdk-1.4 and later.
-        director_->swig_jvm_->AttachCurrentThreadAsDaemon(jenv, NULL);
+        director_->swig_jvm_->AttachCurrentThreadAsDaemon(jenv, &args);
 #else
-        director_->swig_jvm_->AttachCurrentThread(jenv, NULL);
+        director_->swig_jvm_->AttachCurrentThread(jenv, &args);
 #endif
       }
       ~JNIEnvWrapper() {


hasinoff:
Just realized that pthread_getname_np is only available on Android API level 26+. Added logic to fix this, but it would be better to support older versions too.

The alternative approach in the abandoned PR here
https://github.com/swig/swig/pull/831/commits/bd277b726da04b05ecfc64f488612e86b6815415
using prctl looks like it should work (that's how pthread_getname_np itself is implemented in bionic), but the HAVE_PRCTL macro doesn't seem to be defined by the NDK and I'm not sure yet about its availability in older versions.

wsfulton:
I'd rather the thread name wasn't set by default as it adds a slight overhead for each upcall which seems unnecessary and some consistency across platforms is desirable. Implementing the thread name for all platforms is rather challenging to get right. I've had a stab at it for unix including recent versions of linux, Android and macos using both of your attempts. The pthread one may not always work, but the prctl approach not in all linux kernels. The patch I've provided, requires a user to set the `SWIG_JAVA_USE_THREAD_NAME` macro to turn on the implementation in the patch, see the changes to director_threads.i. How does this patch work for you? 

```patch
From 3195c3e4dac605aaa9ee877ea956151bc3e15830 Mon Sep 17 00:00:00 2001
From: William S Fulton <wsf@fultondesigns.co.uk>
Date: Mon, 8 Oct 2018 21:29:11 +0100
Subject: [PATCH] Java directors - more generic thread name setting

Activated if a user sets SWIG_JAVA_USE_THREAD_NAME.
Implementations provided for linux/android/macos/unix.
A user's implementation will be used if SWIG_JAVA_GET_THREAD_NAME is
defined. It must implement the function:
  namespace Swig {
    SWIGINTERN int GetThreadName(char *name, size_t len);
  }
---
 Examples/test-suite/director_thread.i              | 20 +++++++-
 .../test-suite/java/director_thread_runme.java     |  6 +++
 Lib/java/director.swg                              | 55 +++++++++++++++++-----
 3 files changed, 69 insertions(+), 12 deletions(-)

diff --git a/Examples/test-suite/director_thread.i b/Examples/test-suite/director_thread.i
index 260bce7..cb999ec 100644
--- a/Examples/test-suite/director_thread.i
+++ b/Examples/test-suite/director_thread.i
@@ -7,6 +7,11 @@
 %module(directors="1") director_thread
 #endif
 
+%begin %{
+#define SWIG_JAVA_USE_THREAD_NAME
+//#define DEBUG_DIRECTOR_THREAD_NAME
+%}
+
 %{
 #ifdef _WIN32
 #include <windows.h>
@@ -89,10 +94,23 @@ extern "C" {
         fprintf(stderr, "pthread_create failed in run()\n");
         assert(0);
       }
+      int setname = pthread_setname_np(thread, "MyThreadName");
+      if (setname != 0) {
+        fprintf(stderr, "pthread_setname_np failed in run()\n");
+        assert(0);
+      }
 %#endif
       MilliSecondSleep(500);
     }
-    
+
+    static bool namedThread() {
+%#ifdef _WIN32
+      return false;
+%#else
+      return true;
+%#endif
+    }
+
     virtual void do_foo() {
       val += 1;
     }
diff --git a/Examples/test-suite/java/director_thread_runme.java b/Examples/test-suite/java/director_thread_runme.java
index c67d410..9bc4a93 100644
--- a/Examples/test-suite/java/director_thread_runme.java
+++ b/Examples/test-suite/java/director_thread_runme.java
@@ -30,6 +30,12 @@ class director_thread_Derived extends Foo {
   }
 
   public void do_foo() {
+    // Not all operating systems can name threads, so only test on those that can
+    if (Foo.namedThread()) {
+      String threadName = Thread.currentThread().getName();
+      if (!threadName.equals("MyThreadName"))
+        throw new RuntimeException("Unexpected thread name: " + threadName);
+    }
     setVal(getVal() - 1);
   }
 }
diff --git a/Lib/java/director.swg b/Lib/java/director.swg
index ba06f9e..d4c88d5 100644
--- a/Lib/java/director.swg
+++ b/Lib/java/director.swg
@@ -5,19 +5,51 @@
  * methods can be called from C++.
 
-#if defined(__ANDROID__)
+#if defined(DEBUG_DIRECTOR_OWNED) || defined(DEBUG_DIRECTOR_EXCEPTION) || defined(DEBUG_DIRECTOR_THREAD_NAME)
+#include <iostream>
+#endif
+
+#include <exception>
+
+#if defined(SWIG_JAVA_USE_THREAD_NAME)
+
+#if !defined(SWIG_JAVA_GET_THREAD_NAME)
+namespace Swig {
+  SWIGINTERN int GetThreadName(char *name, size_t len);
+}
+
+#if defined(__linux__)
+
 #include <sys/prctl.h>
+SWIGINTERN int Swig::GetThreadName(char *name, size_t len) {
+  (void)len;
+#if defined(PR_GET_NAME)
+  return prctl(PR_GET_NAME, (unsigned long)name, 0, 0, 0);
+#else
+  (void)name;
+  return 1;
 #endif
+}
 
-#if defined(DEBUG_DIRECTOR_OWNED) || defined(DEBUG_DIRECTOR_EXCEPTION)
-#if defined(__ANDROID__)
-#include <errno.h>
-#include <string.h>
+#elif defined(__unix__)
+
+#include <pthread.h>
+SWIGINTERN int Swig::GetThreadName(char *name, size_t len) {
+  return pthread_getname_np(pthread_self(), name, len);
+}
+
+#else
+
+SWIGINTERN int Swig::GetThreadName(char *name, size_t len) {
+  (void)len;
+  (void)name;
+  return 1;
+}
 #endif
-#include <iostream>
+
 #endif
 
-#include <exception>
+#endif
 
 namespace Swig {
 
@@ -151,13 +183,14 @@ namespace Swig {
         args.version = JNI_VERSION_1_2;
         args.group = NULL;
         args.name = NULL;
-#if defined(__ANDROID__)
+#if defined(SWIG_JAVA_USE_THREAD_NAME)
         char thread_name[16];  // MAX_TASK_COMM_LEN=16 is hard-coded in the kernel.
-        if (prctl(PR_GET_NAME, (unsigned long)thread_name, 0, 0, 0) == 0) {
+        if (Swig::GetThreadName(thread_name, sizeof(thread_name)) == 0) {
           args.name = thread_name;
+#if defined(DEBUG_DIRECTOR_THREAD_NAME)
+          std::cout << "JNIEnvWrapper: thread name: " << thread_name << std::endl;
         } else {
-#if defined(DEBUG_DIRECTOR_OWNED) || defined(DEBUG_DIRECTOR_EXCEPTION)
-          std::cerr << "JNIEnvWrapper: Couldn't set thread name: " << strerror(errno) << std::endl;
+          std::cout << "JNIEnvWrapper: Couldn't set Java thread name" << std::endl;
 #endif
         }
 #endif
-- 
2.7.4
```

wsfulton:
@hasinoff do you know of a way we could get SWIG on Android properly tested using Travis? If you do, some help there would be much appreciated as there isn't any testing on Android but is becoming an increasingly important platform.

vadz:
Just FYI, (very) recent versions of Windows 10 support getting and setting thread names using [SetThreadDescription()](https://docs.microsoft.com/en-us/windows/desktop/api/processthreadsapi/nf-processthreadsapi-setthreaddescription). Previous versions supported setting it using the `0x406D1388` exception hack, but not getting it, so I guess this can't be used in this code.

hasinoff:
Thanks, @wsfulton. A few comments:

- I haven't verified whether the JVM resetting the thread name is a problem on platforms other than Android. Does the new "MyThreadName" test fail on current test platforms when the new SWIG_JAVA_USE_THREAD_NAME macro is unset?

- It seems like upcalls are relatively rare and expensive in practice (including the cost of the client Java code), such that the overhead of getting and setting the thread name should be small. Would you consider making the new thread name behavior the default, and inverting the meaning of the macro to SWIG_JAVA_DEFAULT_JVM_THREAD_NAME or similar? That way setting a macro would only be needed for the presumably unusual case where preserving original thread names would noticeably degrade upcall performance.

- On Apple platforms, the maximum thread name length seems to be 64 (including the trailing \0), not 16 like on other Linux-based platforms. Supporting evidence in the definition of MAXTHREADNAMESIZE here:
https://opensource.apple.com/source/xnu/xnu-4570.31.3/osfmk/mach/thread_info.h

- I don't have experience with Travis, but it does look like there's support for Android, building using the Android SDK and testing using an emulator:
https://docs.travis-ci.com/user/languages/android/

wsfulton:
I suspect the JVM makes up a thread name "Thread-XX" on all platforms if JavaVMAttachArgs::name is not supplied. I observed this on Windows 7 and Linux. 

We test regularly on Linux, Mac, Windows. I tested the new "MyThreadName" test on Linux, but it will pass on Windows as the thread name isn't checked for on Windows, see `namedThread` in the patch. A Windows 10 version 1607 solution could be added using the info @vadz gave.

The main purpose of directors is to make this upcall. If you run the test, you'll see that the thread name setting is done on each upcall. The reason is the use of JNIEnvWrapper in each upcall.

I would consider setting the thread name by default only if we had a tried and tested approach for obtaining the thread name on all platforms that will definitely compile and will not result in any runtime problems. If you can find such code, we can add it in... maybe from the JVM source code itself??? Even if we found such code, we can't add on the burden to a user of running an autoconf like 'configure' step for the code to compile. The dependent system headers must always be available. The irony is the JVM provides cross-platform threading code, but we can't access it until the native code's thread has been attached to the JVM!

Let's change to `thread_name[64]` for the longer Apple thread names.

For Travis Android support, the link shows support for a pure Android platform `language: android`, which we can't use as we use a mixed language platform via `language:  cpp`. I suspect it will take a non-trivial amount of effort to get it working. If there are any Android users keen to make sure SWIG continues to work well on Android, please help. Probably best via a separate Github issue to discuss.

wsfulton:
@hasinoff, I'll update your pull request with the mods I put in and make the `thread_name[64]` change and merge to master. Shout if you object or wish to discuss further before I merge.

hasinoff:
SGTM, thanks.

PR #1335, opened 2018-10-11
PHP: Make reserved keywords to reserved function only

AlexanderGabriel:
Hi,

SWIG renames variables or properties of objects like empty to c_empty.
I quess, because they are listet as reservced keywords (http://php.net/manual/en/reserved.keywords.php) but f.e. "empty" is listed as "empty()" which allows to use empty as variable or property of objects but not as function or method.

If i use $empty as variable, my IDE does not complaint (which is not a proof i know) and i can run this code:
grafik

grafik

This can be repeated with all of the keywords which are listet as "keyword()":

grafik

grafik

Full source code here: https://github.com/AlexanderGabriel/SWIGCmakePHPExampleWindows/blob/master/test_keywords.php

I patched SWIG (see changes) to allow these keywords f.e. as empty but still not empty() (this will not work/throw an error), made a c-lib and tested it:
grafik (c-source-file)
grafik (Interface-File)

grafik (PHP-Testfile test_keywords.php)
grafik (output)

There is a project "mapserver" which has a property "empty" on an object "webObj" in their php-extension since more than 10 years and this worked without any problems: https://mapserver.org/mapscript/php/phpmapscript.html#webobj If that was wrong, this wouldnt have worked the last 10 years.

Would be cool if you accept my PR to make this work with swig-based extensions.

Thank you.

Kind regards,
Alex

### DIFF
--- a/Lib/php/phpkw.swg
+++ b/Lib/php/phpkw.swg
@@ -50,7 +50,6 @@ PHPKW(do);
 PHPKW(echo); // "Language construct"
 PHPKW(else);
 PHPKW(elseif);
-PHPKW(empty); // "Language construct"
 PHPKW(enddeclare);
 PHPKW(endfor);
 PHPKW(endforeach);
@@ -830,6 +829,7 @@ PHPFN(cosh);
 PHPFN(count);
 PHPFN(current);
 PHPFN(each);
+PHPFN(empty);
 PHPFN(end);
 PHPFN(exp);
 PHPFN(extract);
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks for the patch.

There's a slight compatibility consideration - if people are relying on the existing automatic renaming then they'll get a different wrapped API now.  But if we target this at 4.0, I think that's fine.  The automatic renaming is really just a way to avoid generating bad code, and I'd think most people will add manual renames to nicer names - for those who don't, the new major version should be enough of a signal.

I suggest we just leave this alone for PHP5 as that's reaches EOL at the end of 2018.

Did you look at any of the other entries in this file to see if any others are misclassified?

AlexanderGabriel:
Thank you for merging.
I only checked the entries on http://php.net/manual/en/reserved.keywords.php ending with ().

PR #1339, opened 2018-10-17
Documentation for argument handling refers to wrong chapter.

hakonhagland:
Chapter 10 (Argument Handling), first section, refers to Chapter 3 (Getting started on
Windows) but should presumably be Chapter 5 (SWIG Basics)?


wsfulton:
Thanks.

PR #1355, opened 2018-11-21
Resolve types of the classes nested inside template classes

### LINKED ISSUE #1353: Wrong abstractness detection for the methods of nested classes inside templates
This is a bit esoteric, but also pretty bad as wrong output is silently generated: if you have a template class with some nested classes defined inside it, name resolution doesn't work correctly for these classes, at least when they're used as function parameters. One (but almost surely not the only) bad consequence of this is that non-abstract classes are wrongly treated as being abstract and so their ctors are not generated. Here is the smallest example:

```
%module ab

%inline %{
template <int>
struct OuterTemplate

template <>
struct OuterTemplate<1>
{
  struct AbstractBase
  {
    virtual bool IsSameAs(const AbstractBase& other) const = 0;
    virtual ~AbstractBase() {}
  };

  struct ConcreteDerived : public AbstractBase
  {
    ConcreteDerived() : m_value(0.) {}
    explicit ConcreteDerived(double value) : m_value(value) {}

    virtual bool IsSameAs(const AbstractBase& other) const
    {
      return m_value == static_cast<const ConcreteDerived&>(other).m_value;
    }

    double m_value;
  };
};
%}
%template(OuterTemplate1) OuterTemplate<1>;
```

Compiling this with e.g. `swig -csharp  -c++ ab.i` (but C# part doesn't matter, the bug is in language-independent part and also happens with e.g. Java) generates `ConcreteDerived` without any ctors (BTW, this example also triggers another long-standing bug with a bogus warning 521, but this is not related).

I've tried debugging this and what I see is that `Allocate::is_abstract_inherit()` doesn't manage to resolve the type correctly, i.e. it is called with `q(const).f(r.q(const).AbstractBase)`, not qualified with `OuterTemplate<(1)>::`, and doesn't find this type anywhere. Because of this it doesn't match the actual parameter of `ConcreteDerived::IsSameAs()` which is `r.q(const).OuterTemplate<(1)>::AbstractBase`, and so the method is considered as pure virtual and the entire class as abstract.

It looks like the real problem is that `decl` attribute of the base class declaration uses the unqualified parameter type in the first place, as this doesn't happen for just a simple (non-template) outer class, but I haven't been able to find out why does this happen. I'll continue investigating, but if anybody (surreptitiously glancing at William) has any idea about this, it would be very welcome, as I still have trouble navigating SWIG internals.

vadz:
Do the same thing for the template classes as for the normal classes
during the type resolve pass, i.e. call classDeclaration() from
TypePass::templateDeclaration(), in order to fully qualify the names of
nested types used as method parameters inside template classes too.

Also amend classDeclaration() to prevent it from processing base classes
for the templates as this would result in errors in absence of %template
directives for the base templates.

Add a test checking for the situation described above.

See #1353.

---

I freely admit I don't fully understand what am I doing here, several things in `TypePass::classDeclaration()` remain completely mysterious to me. But at least this fix makes the new test pass, when it failed before.

Unfortunately this doesn't fix #1353 because in my case an enum value of an enum defined inside a namespace is used and this enum value is not fully qualified, i.e. we're still searching for a function taking `q(const).f(r.q(const).NSpace::OuterTemplate<(TemplArg_1)>::AbstractBase)` and so don't find `q(const).f(r.q(const).NSpace::OuterTemplate<(NSpace::TemplArg_1)>::AbstractBase)` overriding it.

I'm still struggling to understand why/how do we end up with the fully qualified enum value in the derived class declaration but not in the base one, and any help with this would still be very welcome. But in the meanwhile I'd like to test this PR to see if it breaks anything else -- if not, it should be applied, as it does fix a real bug, even though it doesn't fix the bug which affects me personally.

### DIFF
--- a/Examples/test-suite/csharp/nested_in_template_runme.cs
+++ b/Examples/test-suite/csharp/nested_in_template_runme.cs
@@ -1,0 +1,10 @@
+using System;
+using nested_in_templateNamespace;
+
+public class runme {
+    static void Main() {
+        var cd = new OuterTemplate1.ConcreteDerived(8.8);
+        if (cd.m_value != 8.8)
+            throw new Exception("ConcreteDerived not created correctly");
+    }
+}
--- a/Examples/test-suite/nested_in_template.i
+++ b/Examples/test-suite/nested_in_template.i
@@ -1,0 +1,34 @@
+%module nested_in_template
+
+#if !defined(SWIGCSHARP) && !defined(SWIGJAVA)
+%feature("flatnested");
+#endif
+
+%inline %{
+template <int>
+struct OuterTemplate;
+
+template <>
+struct OuterTemplate<1>
+{
+  struct AbstractBase
+  {
+    virtual bool IsSameAs(const AbstractBase& other) const = 0;
+    virtual ~AbstractBase() {}
+  };
+
+  struct ConcreteDerived : AbstractBase
+  {
+    ConcreteDerived() : m_value(0.) {}
+    explicit ConcreteDerived(double value) : m_value(value) {}
+
+    virtual bool IsSameAs(const AbstractBase& other) const {
+      return m_value == static_cast<const ConcreteDerived&>(other).m_value;
+    }
+
+    double m_value;
+  };
+};
+%}
+
+%template(OuterTemplate1) OuterTemplate<1>;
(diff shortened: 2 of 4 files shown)


vadz:
Oops, sorry, a last minute change (which I was so sure couldn't matter) broke even those test suites which I did test. Apparently I understand this stuff even less than I thought...

PR #1356, opened 2018-11-22
Add unordered_set|map and set to Java

bkotzz:
Based off the implementation for the `std_map.i` that currently exists, this PR adds the equivalent for STL containers `unordered_set`, `unordered_map`, `set`. Also adds tests for the four classes.

PTAL @wsfulton


wsfulton:
Hi @bkotzz. Implementations for set, and the unordered containers are very welcome, thanks.

The next release is SWIG 4.0.0 where there is the possibility to introduce non-backwards compatible changes.
What I would like is for the Java std::map wrappers to implement Java's java.util.Map<K,V> so that it looks and feels like a Java container. The approach of making a C++ container look and feel like a Java container has already been done for std::vector and std::list. Incidentally, std::map for C# is also complete, but std::set is missing. I'd like to see every wrapped method have a runtime test and suggest using Examples/test-suite/csharp/li_std_map_runme.cs as a reference and convert it into Java and extending where necessary.

Anyway, this will require renaming method like set to put, empty to isEmpty, etc. All the other methods in the java.util.Map<K,V> interface will also need to be implemented. The bad news is it extra work in the development and testing and will be a more complex implementation. The good news is that a lot of the implementation can be cribbed from std_vector.i and std_list.i.

The same needs doing for std::set where the wrappers should implement java.util.Set<E> (I think this is most appropriate). Hopefully this will be fairly easy and again can mostly be cribbed from the other containers.

Finally, the std::unordered_* implementation and tests will be a fairly simple copy/paste from std::map and std::set.

Will you be able to tackle this? If you are, it will provide a huge boost as I'm not going to release SWIG 4.0.0 until this is done. You're clearly getting comfortable with the dev work from what I can see in the patch.

bkotzz:
Hi @wsfulton, that’s a good idea. I do find it a bit annoying when these proxy classes don’t match the typical Java containers, so I see the benefit here. I think I can take this on, but not sure I have time right now. What is the timeline for the version 4 release? I’ll have more time to work on this around Christmas time - think we should go ahead with this change to get something out there, and then pick this back up around that time?

wsfulton:
Hopefully 4.0 will be ready in a month or two, so I'll wait until you have refactored this. Sorry, but I only pull finished work into master.

wsfulton:
Hi @bkotzz, please be aware that each time you update the pull request it is added to a queue of CI testing which kicks off nearly 100 builds requiring 19 hours of compute time on Appveyor and Travis. It will be helpful if you test any changes locally and provide your updated patch of changes when you are confident that the development is ready for review.

I also see some force pushes to this public pull request which I'm sure you will appreciate is not good practice for tracking changes. Maybe you want to develop this on a new branch and submit a new clean pull request when ready.

wsfulton:
Hi @bkotzz. We're getting close to releasing SWIG 4.0.0. Will you be able to finish off std::map anytime soon?

bkotzz:
@wsfulton please take a look

bkotsopoulossc:
Hey @wsfulton, any feedback here?

bkotsopoulossc:
Sounds good, I just don't want to be the bottleneck for the release, so just checking if there's any work items for me. I'll stay tuned

wsfulton:
Hi @bkotsopoulossc. I've had a close look now. Congratulations, it looks truly great. I'll add a couple of minor comments as a review shortly. I've got just one main observation/query:

 Java's equivalent to std::map looks like it is TreeMap as it is a sorted map and implements the NavigableMap interface (which in turn inherits from SortedMap) as well as extending AbstractMap.  How about wrapped std::map also implementing NavigableMap?  So instead of "The std::map container looks and feels much like a java.util.HashMap from Java", it'll look and feel like a java.util.TreeMap. Java API's taking either a SortedMap or NavigableMap interface can then be passed a wrapped std::map.

Similarly for wrapped std::set to additionally implement NavigableSet?

However, the unordered_map and unordered_set are okay as HashMap and HashSet are unordered containers.

Lib/java/std_map.i
```diff
@@ -12,60 +14,188 @@
...
+    iterator begin();
+    iterator end();
+    %extend {
+      %fragment("SWIG_MapSize");
+
+      jint sizeImpl() const throw (std::out_of_range) {
+        return SWIG_MapSize(self->size());
+      }
 
-// Legacy macros (deprecated)
```
wsfulton: Please leave all these legacy macros in. There are a number of these across different languages and I'll remove them all at the same time one day.

Lib/java/std_map.i
```diff
@@ -12,60 +14,188 @@
...
+
+    return setToReturn;
+  }
+%}
+
+  public:
+    map();
+    map(const map<KeyType, ValueType, Comparator >&);
+
+    struct iterator {
```
wsfulton: The iterator class needn't be public, so add this inside class definition on the next line:
    %typemap(javaclassmodifiers) iterator "protected class"

Lib/java/std_map.i
```diff
@@ -12,60 +14,188 @@
...
-          if (i != self->end())
-            return i->second;
-          else
-            throw std::out_of_range("key not found");
+        std::map<KeyType, ValueType, Comparator >::iterator getNextUnchecked() {
+          std::map<KeyType, ValueType, Comparator >::iterator copy = (*$self);
+          return ++copy;
+        }
+
+        bool isNot(const iterator other) const {
```
wsfulton: Pass by reference would be better:
```suggestion
        bool isNot(const iterator &other) const {
```

Lib/java/std_map.i
```diff
@@ -12,60 +14,188 @@
...
+%javamethodmodifiers std::map::iterator::setValue "private";
+
 namespace std {
 
-  template<class K, class T, class C = std::less<K> > class map {
-    public:
-      typedef size_t size_type;
-      typedef ptrdiff_t difference_type;
-      typedef K key_type;
-      typedef T mapped_type;
```
wsfulton: These typedefs are a fundamental definition of the container and should not be deleted.

Lib/java/std_map.i
bkotsopoulossc: In general C++ iterations are lightweight objects and the STL typically passes them around by value (take a look at the declarations of any container constructor or algorithm that takes in a begin/end iterator range). But up to you, I can change it

bkotsopoulossc:
@wsfulton I can look at implementing NavigableMap as well but that has a fairly big surface area and will take some time

wsfulton:
How about addressing the minor issues mentioned above and then I'll merge to master? You can then add NavigableMap afterwards. It should be an enhancement and not change any of the APIs you already have. I have a few more cosmetic changes that I can do straight after the merge, which you may want to wait for if we take this approach.

Lib/java/std_map.i
bkotzz: It's not as simple as that - `const_iterator` and `iterator` are different types - you can't pass an `iterator` into a function that takes `const_iterator`. This struct that this member belongs to is `iterator` not `const_iterator`. To switch to `const_iterator` means that the `setValue` function in the extension part is no longer valid because a `const_iterator` cannot mutate the value it points to.

bkotzz:
@wsfulton any chance of getting this in? I really would like to use this in my project and it’s been up for a few months now

wsfulton:
I've merged it in actually, but havn't pushed it yet. Am just making some other STL changes too before before pushing (possibly later today).

wsfulton:
There you go, all merged. Many thanks for this well worked enhancement. If you fancy working the same magic on std::multiset and std::multiset too and unordered equivalents, please do.

bkotzz:
Thanks @wsfulton. What’s the timeline for the version 4 release?

wsfulton:
About now, but there are some regressions reported on the swig-devel mailing list which need fixing first. Keep an eye on it here: https://github.com/swig/swig/milestone/1

Issue #1024, opened 2017-07-18
Custom __new__ for shadow class


fflexo:
```
%module shadow_new

%inline %{
class _bar {};
typedef _bar *bar;

bar get_bar() { return new _bar(); }
%}

%extend _bar {
  %pythoncode %{
    def __new__(cls,*args,**kwargs):
      print('In new for class: %s' % str(cls))
      return super().__new__(cls)
  %}
}
```

And a quick test of that:

```
import shadow_new
shadow_new.get_bar()
```

Built with:
```
swig3.0 -python -py3 -c++ -Wall shadow_new.i
g++ -Wall shadow_new_wrap.cxx -o_shadow_new.so \
 -I/usr/include/python3.4 -shared -fPIC
python3 run.py
```
Gives the following with SWIG 3.0.10:
```
Segmentation fault (core dumped)
```
I was expecting that if I'd done something wrong then it would at least give me a cryptic error message instead of segfault, although what I actually want to do is influence the creation of the shadow class.

Some poking in a debugger suggests that the last code not from libpython was this line, from `SWIG_Python_NewShadowInstance`:
```
    inst = ((PyTypeObject*) data->newargs)->tp_new((PyTypeObject*) data->newargs, Py_None, Py_None);
```

Behaviour seems the same with a relatively recent build of the trunk. With SWIG 2.x it gives the following output instead:
```
Traceback (most recent call last):
  File "run.py", line 2, in <module>
    shadow_new.get_bar()
TypeError: object() takes no parameters
```
although I'm none the wiser still.

wsfulton:
I can't reproduce your swig 2.x output using swig2.0.12. `object()` is valid python3 code despite the error message you show. Probably this is the result of undefined behaviour in that version, which manifests as a seg fault with swig-3.0.10.

This issue is fixed now though for swig 4, see #1321. The output of your test case is no longer a seg fault, but:
```
In new for class: <class 'example._bar'>
```
which looks like correct to me.

PR #1357, opened 2018-11-22
fix for creating new shadow class in python 3.6

chlandsi:
I ran into the issue with overriding `__new__` described in #1024 with Swig 3.0.12 and fixed it for python 3.7 by applying 0165180, but this created a new error "SystemError: Objects/tupleobject.c:81: bad argument to internal function" in python 3.6. This patch fixes the issues I had for python 3.5, 3.6 and 3.7.

### DIFF
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1205,12 +1205,16 @@ SWIG_Python_NewShadowInstance(SwigPyClientData *data, PyObject *swig_this)
 #if PY_VERSION_HEX >= 0x03000000
     PyObject *empty_args = PyTuple_New(0);
     if (empty_args) {
-      inst = ((PyTypeObject *)data->newargs)->tp_new((PyTypeObject *)data->newargs, empty_args, Py_None);
-      Py_DECREF(empty_args);
-      if (inst) {
-        PyObject_SetAttr(inst, SWIG_This(), swig_this);
-        Py_TYPE(inst)->tp_flags &= ~Py_TPFLAGS_VALID_VERSION_TAG;
+      PyObject *empty_kwargs = PyDict_New();
+      if (empty_kwargs) {
+        inst = ((PyTypeObject *)data->newargs)->tp_new((PyTypeObject *)data->newargs, empty_args, empty_kwargs);
+        Py_DECREF(empty_kwargs);
+        if (inst) {
+          PyObject_SetAttr(inst, SWIG_This(), swig_this);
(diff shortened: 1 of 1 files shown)


wsfulton:
Ping @chlandsi

chlandsi:
Sorry, I've been busy with other stuff. I'll try to come up with a minimal example within the week. So far I can say that this has happened with different builds of CPython 3.6 on Mac and Linux (Ubuntu 16.04) with no other swig flags than `-py3`, and the latest release version of swig.

chlandsi:
The code to reproduce is more or less the same as in #1024:
```
%module(directors="0") mod;

%inline %{
class Foo {
  public:
    virtual ~Foo() {}
    Foo() {}
};

Foo* get_foo() {return new Foo();}
%}

%extend Foo {
    %pythoncode %{
    def __new__(cls, *args, **kwargs):
        print('in Foo.__new__()')
        return super().__new__(cls)

    def __init__(self):
        print('in Foo.__init__()')
    %}
};
```

Calling 
```
mod.get_foo()
foo = mod.Foo()
```
on the module generated by swig 3.0.12, with or without #0165180 applied, gives the following result on Release cpython 3.5.5, 3.6.5 and 3.7.1:

```
# Python 3.5.5
$ swig -Wall -v -c++ -python -py3 src/mod.i
$ g++ --std=c++17 -shared -fPIC src/mod_wrap.cxx -o src/_mod.so -I/home/user/.pyenv/versions/3.5.5/include/python3.5m/

>> in Foo.__new__()
>> in Foo.__new__()
>> in Foo.__init__()

# Python 3.6.5
$ swig -Wall -v -c++ -python -py3 src/mod.i
$ g++ --std=c++17 -shared -fPIC src/mod_wrap.cxx -o src/_mod.so -I/home/user/.pyenv/versions/3.6.5/include/python3.6m/

>> Traceback (most recent call last):
>>   File "test.py", line 3, in <module>
>>     foo1 = mod.get_foo()
>> SystemError: Objects/tupleobject.c:81: bad argument to internal function

# Python 3.7.1
$ swig -Wall -v -c++ -python -py3 src/mod.i
$ g++ --std=c++17 -shared -fPIC src/mod_wrap.cxx -o src/_mod.so -I/home/user/.linuxbrew/include/python3.7m

>> in Foo.__new__()
>> in Foo.__new__()
>> in Foo.__init__()
```

With the patch from this PR, the result is as expected on all Python versions.

wsfulton:
Perfect. I could recreate this using Ubuntu python 3.6.7. I'll add this as a test case and merge to master shortly. Thanks.

PR #1358, opened 2018-11-22
Fix handling of abstract base classes nested inside templates

### LINKED ISSUE #1353: Wrong abstractness detection for the methods of nested classes inside templates

vadz:
Code handling %template in the parser created a totally new top-level
module child of namespace type when handling templates inside a
namespace and copied the nodes from the previously parsed C++ template
declaration to it. However copies of this node kept their original
values of "abstracts" attribute, which contained pointers to the classes
in the original template declaration, i.e. outside of the subtree
created for the instantiated template. This, in turn, meant that during
the types resolution pass, the code in TypePass did not update the types
used in the methods of the classes appearing in the "abstracts" List,
even though it did update the types for the children of the instantiated
template subtree.

And this finally resulted in wrongly detecting overridden virtual
methods as abstract in Allocate::is_abstract_inherit() during the next
pass, as the signatures of the overridden method -- using resolved types
-- and of the method from the class pointed to by "abstract" -- using
the original types from C++ code -- didn't match.

Resolve this simply by not copying "abstracts" attributes when creating
the template subtree and doing another pass over this tree to recreate
them using the new nodes, just as it's already done for "defaultargs"
attribute, presumably for similar reasons. Note that doing another pass
over the tree is not as efficient as doing everything in a single pass,
but merging the new update_abstracts() with update_defaultargs() is not
completely obvious, so for now keep it simple and optimize it later if
necessary.

Also, add a test checking for the situation described above.

Closes #1353.


vadz:
@wsfulton Can I merge this in myself if you don't have time for it, please? I'm quite sure that this version of the fix is correct and it does fix a pretty bad (because wrappers for C++ ctors are just silently not generated) bug.

wsfulton:
Code looks fine. Explanation of the problem and solution is great. Go ahead merge. I will at some point check the output of the `make partialcheck-python-test-suite` before and after the change just to verify the change, so you might want to do so as well, otherwise the patch will need further work. 

Just one small comment about the test and use of `double m_value`. Can you please change this to an int as comparing for exact double precision values is not good practice?

wsfulton:
There is one other thing you need to do and that is add nested_in_template to Examples/test-suite/common.mk and make sure the test-suite doesn't break. I'm assuming you tested Python,C#,Java, but you'll need to check all the other languages too.

vadz:
Oops, sorry for forgetting to update `common.mk` and thanks for pointing it out, done now. Also replaced double with int.

And I've compared the results of running `make partialcheck-python-test-suite` and there were no changes (BTW, I just did this manually with `diff -qr swig{-master,}/Examples/test-suite/python`, is there some more automatic way to do it with SWIG makefiles?).

I'll merge this once the CI builds finish.

vadz:
Sorry, just to avoid any confusion: I made a mistake and the example still fails with `-builtin`. However it's still true that it is completely unrelated to this commit and is probably due to 0f61c5a84743268d808bd8da372a27f8e2e88cbc

vadz:
I did make the change but have somehow forgot to commit this file, sorry. Will create a new PR to retest the build with it in a moment.

vadz:
I'm really sorry, I have no idea what has happened here, apparently I've somehow overwritten my changes while rebasing/amending. I've readded this to #1364 too, thanks for noticing it.

PR #1359, opened 2018-11-23
Fix bogus warnings about illegal dtor names for dtors of nested classes inside templates

vadz:
The import commit here is the second one, the first one is just some trivial cleanup. The changes in the second commit are actually rather small too, just view it ignoring whitespace so that the indented code doesn't appear as being changed.

### DIFF
--- a/Source/CParse/templ.c
+++ b/Source/CParse/templ.c
@@ -49,7 +49,7 @@ void Swig_cparse_debug_templates(int x) {
  * template parameters
  * ----------------------------------------------------------------------------- */
 
-static void cparse_template_expand(Node *n, String *tname, String *rname, String *templateargs, List *patchlist, List *typelist, List *cpatchlist) {
+static void cparse_template_expand(Node *templnode, Node *n, String *tname, String *rname, String *templateargs, List *patchlist, List *typelist, List *cpatchlist) {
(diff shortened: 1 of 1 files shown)


vadz:
I'm not sure if this has anything to do with Vladimir Kalinin's changes, I think this place was just not updated (because how could anybody know that it had to be updated?), but any additional review would be welcome, of course.

And the `nested_in_template` test case does indeed generate this warning without this PR. So I'd prefer to apply this one too when I'll merge the other one if you don't mind (I'll address your comments there a.s.a.p. but not today).

wkalinin:
Indeed these changes have nothing to do with mine. That is, I just didn't notice this case. The fix seems to be fine as far as I can understand, and it is very localized anyway, doing exactly what's described.

vadz:
@wsfulton Is it ok for me to merge this too?

wsfulton:
This fixes more than a bogus warning. The destructor wrapper was missing altogether before the fix. Looks like it occurred for any nested class within a template, but only when the destructor was explicitly declared. That is, there was no problem with an implicit destructor. I've just merged this in, thanks.

@wkalinin. I'm also surprised we didn't find this case when the nested class support was added... never mind!

Issue #1360, opened 2018-11-23
Leak of SWIG var link object (C++/python bindings)


adr26:
I have noticed that some of our tests were detecting a memory leak on process exit (running on Windows):
 
```Detected memory leaks!
Dumping objects ->
{177} normal block at 0x035A1AD0, 25 bytes long.
Data: <DataBlock_EmptyD> 44 61 74 61 42 6C 6F 63 6B 5F 45 6D 70 74 79 44 
{176} normal block at 0x035A2C98, 16 bytes long.
Data: <  Z p    }      > D0 1A 5A 03 70 C4 00 10 F0 7D 09 10 00 00 00 00 
Object dump complete.
```
 
I traced this down to the reference count of `_SWIG_globals` not returning to `0` when our .pyd was unloaded, and hence `swig_varlink_dealloc()` not running.

wsfulton:
Fixed by #1361

PR #1361, opened 2018-11-23
#1360: Leak of SWIG var link object

adr26:
Fix reference counting on `_SWIG_globals` to allow var link to be freed on module unload.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,11 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.0 (in progress)
 ===========================
 
+2018-11-23: adr26
+            [Python] #1360 Leak of SWIG var link object
+
+            Fix reference counting on _SWIG_globals to allow var link to be freed on module unload.
+
 2018-11-22: rupertnash
             [Python] #1282 Make generated module runnable via python -m (PEP 366 conforming)
 
--- a/Lib/python/pyinit.swg
+++ b/Lib/python/pyinit.swg
@@ -227,11 +227,7 @@ SWIG_Python_addvarlink(PyObject *p, const char *name, PyObject *(*get_attr)(void
 SWIGINTERN PyObject *
 SWIG_globals(void) {
   static PyObject *_SWIG_globals = 0; 
-  if (!_SWIG_globals) {
-    _SWIG_globals = SWIG_newvarlink();
-  } else {
-    Py_INCREF(_SWIG_globals);
-  }
+  if (!_SWIG_globals) _SWIG_globals = SWIG_newvarlink();  
   return _SWIG_globals;
 }
 
--- a/Lib/python/pyinit.swg
+++ b/Lib/python/pyinit.swg
@@ -331,6 +331,9 @@ SWIGEXPORT
 #endif
 SWIG_init(void) {
   PyObject *m, *d, *md, *globals;
+
+  (void)globals;
+
 #if PY_VERSION_HEX >= 0x03000000
   static struct PyModuleDef SWIG_module = {
     PyModuleDef_HEAD_INIT,
--- a/Lib/python/pyinit.swg
+++ b/Lib/python/pyinit.swg
@@ -332,8 +332,6 @@ SWIGEXPORT
 SWIG_init(void) {
   PyObject *m, *d, *md, *globals;
 
-  (void)globals;
-
 #if PY_VERSION_HEX >= 0x03000000
   static struct PyModuleDef SWIG_module = {
     PyModuleDef_HEAD_INIT,
@@ -385,6 +383,8 @@ SWIG_init(void) {
   metatype = SwigPyObjectType();
   assert(metatype);
 #endif
+
+  (void)globals;
 
   /* Create singletons now to avoid potential deadlocks with multi-threaded usage after module initialization */
   SWIG_This();
--- a/Lib/python/pyinit.swg
+++ b/Lib/python/pyinit.swg
@@ -227,7 +227,11 @@ SWIG_Python_addvarlink(PyObject *p, const char *name, PyObject *(*get_attr)(void
 SWIGINTERN PyObject *
 SWIG_globals(void) {
   static PyObject *_SWIG_globals = 0; 
-  if (!_SWIG_globals) _SWIG_globals = SWIG_newvarlink();  
+  if (!_SWIG_globals) {
+    _SWIG_globals = SWIG_newvarlink();
+  } else {
+    Py_INCREF(_SWIG_globals);
+  }
   return _SWIG_globals;
 }
 
@@ -326,7 +330,7 @@ SWIGEXPORT
   void
 #endif
 SWIG_init(void) {
-  PyObject *m, *d, *md;
+  PyObject *m, *d, *md, *globals;
 #if PY_VERSION_HEX >= 0x03000000
   static struct PyModuleDef SWIG_module = {
     PyModuleDef_HEAD_INIT,
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -3316,21 +3316,21 @@ public:
     getf = NewWrapper();
     setf = NewWrapper();
 
-    Printf(f_init, "\t globals = SWIG_globals();\n");
-    Printf(f_init, "\t if (!globals) {\n");
-    Printf(f_init, "     PyErr_SetString(PyExc_TypeError, \"Failure to create SWIG globals.\");\n");
-    Printf(f_init, "#if PY_VERSION_HEX >= 0x03000000\n");
-    Printf(f_init, "\t   return NULL;\n");
-    Printf(f_init, "#else\n");
-    Printf(f_init, "\t   return;\n");
-    Printf(f_init, "#endif\n");
-    Printf(f_init, "\t }\n");
-
     /* If this is our first call, add the globals variable to the
        Python dictionary. */
 
     if (!have_globals) {
+      Printf(f_init, "\t globals = SWIG_globals();\n");
+      Printf(f_init, "\t if (!globals) {\n");
+      Printf(f_init, "     PyErr_SetString(PyExc_TypeError, \"Failure to create SWIG globals.\");\n");
+      Printf(f_init, "#if PY_VERSION_HEX >= 0x03000000\n");
+      Printf(f_init, "\t   return NULL;\n");
+      Printf(f_init, "#else\n");
+      Printf(f_init, "\t   return;\n");
+      Printf(f_init, "#endif\n");
+      Printf(f_init, "\t }\n");
       Printf(f_init, "\t PyDict_SetItemString(md, \"%s\", globals);\n", global_name);
+      Printf(f_init, "\t Py_DECREF(globals);\n");
       if (builtin)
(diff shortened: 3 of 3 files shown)


wsfulton:
Hi @adr26. If you look at the Examples/python/simple example, you can see that it wraps a few global C variables. The patch has introduced some code bloat with unnecessary duplicate resetting of the `globals` variable in `SWIG_init`. Can you revert this to go back to a one liner to set up each global variable wrapper in the `SWIG_addvarlink` call? BTW, you've dropped the spaces after the commas in the `SWIG_addvarlink` call.

Would you mind explaining how this has fixed a memory leak? I've probably missed something, but from what I can see, the patch adds one `Py_INCREF(_SWIG_globals)` in `SWIG_globals()` call to every one `Py_DECREF(globals);` in `SWIG_init()`. Previously these inc/dec calls weren't present but the net effect was the same. Where are the extra Py_DECREF calls to fix the leak?

adr26:
Hi @wsfulton, thanks for the comments and sorry for the inauspicious start!

I have used the `Examples/python/simple` example as the basis of answering your questions.

1. Code bloat / unnecessary re-setting of `globals`.

I have reverted `SWIG_globals()` to just return the result of `SWIG_newvarlink()`, as we can simply add the required `Py_DECREF(globals)` inside the `if (!have_globals)`, after having called `PyDict_SetItemString()` to increment the reference count from 1->2.

I have left all the other accesses still using the new local variable `globals`, which I added. For a debug build, doing this will save generating many extra calls to `SWIG_globals()` and explicitly tells the compiler that another call to `SWIG_newvarlink()` won't be necessary. However, this point is moot on a release build where the compiler would hopefully inline all the calls to `SWIG_globals` anyway, so if you feel strongly that you'd like me to eliminate the new local variable `globals` and revert to multiple `SWIG_globals()` calls as a further step, I could do so (though I prefer it how it is now).

2. Spacing in the `SWIG_addvarlink()` call.

Mea culpa, this is fixed.

3. Explanation of the requirement for `Py_DECREF(globals)`.

Simply put, when [`PyObject_New`](https://docs.python.org/2/c-api/allocation.html?highlight=pyobject_new#c.PyObject_New) is called, it returns a new object with a reference count of 1. There must be a balancing `Py_DECREF()` of this object, taking its reference count from 1->0, in order for the [`tp_dealloc` API](https://docs.python.org/2/c-api/typeobj.html?highlight=tp_dealloc#c.PyTypeObject.tp_dealloc) to be called on this object.

The key point is then to see that all the other transitions on the object reference count for the `swig_varlinkobject` are balancing pairs and so without a `Py_DECREF(globals)` to balance the original `PyObject_NEW()` call, the `swig_varlinkobject` is never correctly deallocated.

The easiest way to show this is to use the `Examples/python/simple` example to construct an object reference count trace (shown here with my fix in place, so that the reference count does transition ->0).
Consider simply loading the example module, then exiting to firce it to be unloaded:
```
C:\workspaces\swig\Examples\python\simple>python
Python 2.7.15 (v2.7.15:ca079a3ea3, Apr 30 2018, 16:22:17) [MSC v.1500 32 bit (Intel)] on win32
>>> import _example
>>> ^Z
```
Then, attaching a debugger and logging the callstacks to show the transitions on the reference counts (and using VS2017's nice mixed native/python debug facilities!), we can see how the extra `Py_DECREF(globals)` allows the reference count to transition back to 0 and therefore allows the `swig_varlinkobject` to be freed:
```
\Examples\python\simple\example_wrap.c:     swig_varlinkobject *result = PyObject_NEW(swig_varlinkobject, swig_varlink_type());

refcnt: 0x00000001
>    _example.pyd!SWIG_Python_newvarlink()                                                                                    Line 3351   C
     _example.pyd!SWIG_globals()                                                                                              Line 3378   C
     _example.pyd!init_example()                                                                                              Line 3609   C
     [Python to Native Transition]
     <stdin>!<module> Line 1                                                                                                              Python
     [Native to Python Transition]
     python.exe!__tmainCRTStartup()                                                                                           Line 586    C
     kernel32.dll!@BaseThreadInitThunk@12()                                                                                   Unknown
     ntdll.dll!__RtlUserThreadStart()                                                                                         Unknown
     ntdll.dll!__RtlUserThreadStart@8()                                                                                       Unknown

\Examples\python\simple\example_wrap.c:   PyDict_SetItemString(md, "cvar", globals);

refcnt: 0x00000002
>    python27.dll!dict_set_item_by_hash_or_entry(_object * op, _object * key, long hash, PyDictEntry * ep, _object * value)   Line 793    C
     python27.dll!PyDict_SetItem(_object * op, _object * key, _object * value)                                                Line 848    C
     python27.dll!PyDict_SetItemString(_object * v, const char * key, _object * item)                                         Line 2523   C
     _example.pyd!init_example()                                                                                              Line 3612   C
     [Python to Native Transition]
     <stdin>!<module>                                                                                                         Line 1      Python
     [Native to Python Transition]
     python.exe!__tmainCRTStartup()                                                                                           Line 586    C
     kernel32.dll!@BaseThreadInitThunk@12()                                                                                   Unknown
     ntdll.dll!__RtlUserThreadStart()                                                                                         Unknown
     ntdll.dll!__RtlUserThreadStart@8()                                                                                       Unknown

\Examples\python\simple\example_wrap.c:   Py_DECREF(globals);

refcnt: 0x00000001
>    _example.pyd!init_example()                                                                                              Line 3613   C
     [Python to Native Transition]
     <stdin>!<module>                                                                                                         Line 1      Python
     [Native to Python Transition]
     python.exe!__tmainCRTStartup()                                                                                           Line 586    C
     kernel32.dll!@BaseThreadInitThunk@12()                                                                                   Unknown
     ntdll.dll!__RtlUserThreadStart()                                                                                         Unknown
     ntdll.dll!__RtlUserThreadStart@8()                                                                                       Unknown

refcnt: 0x00000002
>    python27.dll!PyDict_Merge(_object * a, _object * b, int override)                                                        Line 1677   C
     python27.dll!PyDict_Copy(_object * o)                                                                                    Line 1747   C
     python27.dll!_PyImport_FixupExtension(char * name, char * filename)                                                      Line 597    C
     python27.dll!_PyImport_LoadDynamicModule(char * name, char * pathname, _iobuf * fp)                                      Line 68     C
     python27.dll!load_module(char * name, _iobuf * fp, char * pathname, int type, _object * loader)                          Line 1937   C
     python27.dll!import_submodule(_object * mod, char * subname, char * fullname)                                            Line 2725   C
     python27.dll!load_next(_object * mod, _object * altmod, char * * p_name, char * buf, int * p_buflen)                     Line 2539   C
     python27.dll!import_module_level(char * name, _object * globals, _object * locals, _object * fromlist, int level)        Line 2248   C
     python27.dll!PyImport_ImportModuleLevel(char * name, _object * globals, _object * locals, _object * fromlist, int level) Line 2312   C
     python27.dll!builtin___import__(_object * self, _object * args, _object * kwds)                                          Line 50     C
     python27.dll!PyCFunction_Call(_object * func, _object * arg, _object * kw)                                               Line 85     C
     python27.dll!PyObject_Call(_object * func, _object * arg, _object * kw)                                                  Line 2547   C
     python27.dll!PyEval_CallObjectWithKeywords(_object * func, _object * arg, _object * kw)                                  Line 4241   C
     python27.dll!PyEval_EvalFrameEx(_frame * f, int throwflag)                                                               Line 2643   C
     python27.dll!PyEval_EvalCodeEx(PyCodeObject * co, _object * globals, _object * locals, _object * * args, int argcount, _object * * kws, int kwcount, _object * * defs, int defcount, _object * closure) Line 3604    C
     [Python to Native Transition]
     <stdin>!<module>                                                                                                         Line 1    Python
     [Native to Python Transition]
     python.exe!__tmainCRTStartup()                                                                                           Line 586    C
     kernel32.dll!@BaseThreadInitThunk@12()                                                                                   Unknown
     ntdll.dll!__RtlUserThreadStart()                                                                                         Unknown
     ntdll.dll!__RtlUserThreadStart@8()                                                                                       Unknown

refcnt: 0x00000001
>    python27.dll!insertdict_by_entry(_dictobject * mp, _object * key, long hash, PyDictEntry * ep, _object * value)          Line 519    C
     python27.dll!insertdict(_dictobject * mp, _object * key, long hash, _object * value)                                     Line 556    C
     python27.dll!dict_set_item_by_hash_or_entry(_object * op, _object * key, long hash, PyDictEntry * ep, _object * value)   Line 795    C
     python27.dll!PyDict_SetItem(_object * op, _object * key, _object * value)                                                Line 848    C
     python27.dll!_PyModule_Clear(_object * m)                                                                                Line 139    C
     python27.dll!PyImport_Cleanup()                                                                                          Line 512    C
     python27.dll!Py_Finalize()                                                                                               Line 487    C
     python27.dll!Py_Main(int argc, char * * argv)                                                                            Line 690    C
     python.exe!__tmainCRTStartup()                                                                                           Line 586    C
     kernel32.dll!@BaseThreadInitThunk@12()                                                                                   Unknown
     ntdll.dll!__RtlUserThreadStart()                                                                                         Unknown
     ntdll.dll!__RtlUserThreadStart@8()                                                                                       Unknown

refcnt: 0x00000000
>    _example.pyd!swig_varlink_dealloc(swig_varlinkobject * v)                                                                Line 3245    C
     python27.dll!dict_dealloc(_dictobject * mp)                                                                              Line 1086    C
     python27.dll!dict_dealloc(_dictobject * mp)                                                                              Line 1086    C
     python27.dll!Py_Finalize()                                                                                               Line 487    C
     python27.dll!Py_Main(int argc, char * * argv)                                                                            Line 690    C
     python.exe!__tmainCRTStartup()                                                                                           Line 586    C
     kernel32.dll!@BaseThreadInitThunk@12()    Unknown
     ntdll.dll!__RtlUserThreadStart()    Unknown
     ntdll.dll!__RtlUserThreadStart@8()    Unknown
```
In particular, you can note that without the `Py_DECREF(globals)`, there would be no 1->0 reference count transition when the extension was unloaded, and `swig_varlink_dealloc()` would not be called.

If in `\Lib\python\pyinit.swg`, you add:
```c++
+#include <crtdbg.h>

SWIG_init(void) {
  PyObject *m, *d, *md, *globals;
+  int flags = 0;

...

  (void)globals;

+  flags |= _CRTDBG_ALLOC_MEM_DF;
+  flags |= _CRTDBG_LEAK_CHECK_DF;
+  _CrtSetDbgFlag(flags);
```
then you can turn on the Microsoft CRT's debug heap leak tracking facility. With this enabled, and my fix applied, there is no leak reported on exit in the debug output.

If, however, you then comment out the `Py_DECREF(globals)`, the debug output shows:
```
Detected memory leaks!
Dumping objects ->
{116} normal block at 0x009A42E0, 4 bytes long.
 Data: <Foo > 46 6F 6F 00 
{115} normal block at 0x00999750, 16 bytes long.
 Data: < B  P           > E0 42 9A 00 50 B3 9D 0F D0 B3 9D 0F 00 00 00 00 
Object dump complete.
The program '[0x22B0] python.exe' has exited with code 0 (0x0).
```
where you can see the leaked heap blocks are indeed the `swig_varlinkobject` and its name ("Foo" in this example).

Hopefully this makes everything clear. I have sync'd the latest changes from master, so I hope the CI should be fixed as well, having picked up your version check fix.

Please let me know if I need to clarify further, and thank you for your input!

Yours, Andrew R

wsfulton:
Thanks for all the details. It's merged now.

PR #1363, opened 2018-11-25
Remove checks for Python version being >= 2.7.0 in generated code

vadz:
These checks are unnecessary as the generated code throws an exception
if Python version is < 2.7.0 anyhow, so simplify the code by removing
them as we know that the test is always true.

Incidentally, this fixes the problem with generating invalid code when
using relative imports with -builtin in master due to using the
undeclared _swig_python_version_info -- as now this code doesn't use
_swig_python_version_info at all.

---

As the current master doesn't pass Travis CI tests, I wanted to fix this. The problem is due to using `_swig_python_version_info` before it's imported, so initially I did this:
```diff
diff --git a/Source/Modules/python.cxx b/Source/Modules/python.cxx
index 19798b5f8..f2e89c3e9
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -898,15 +898,13 @@ public:

     if (shadow) {
       Swig_banner_target_lang(f_shadow_py, "#");
-      if (Len(f_shadow_begin) > 0)
-       Printv(f_shadow_py, "\n", f_shadow_begin, "\n", NIL);
+      Printv(f_shadow_begin, "\nfrom sys import version_info as _swig_python_version_info\n", NULL);
+      Printv(f_shadow_begin, "if _swig_python_version_info < (2, 7, 0):\n", NULL);
+      Printv(f_shadow_begin, tab4, "raise RuntimeError('Python 2.7 or later required')\n\n", NULL);
+      Printv(f_shadow_py, f_shadow_begin, NIL);
       if (Len(f_shadow_after_begin) > 0)
       Printv(f_shadow_py, f_shadow_after_begin, "\n", NIL);

-      Printv(f_shadow_py, "\nfrom sys import version_info as _swig_python_version_info\n", NULL);
-      Printv(f_shadow_py, "if _swig_python_version_info < (2, 7, 0):\n", NULL);
-      Printv(f_shadow_py, tab4, "raise RuntimeError('Python 2.7 or later required')\n\n", NULL);
-
       if (moduleimport) {
        Replaceall(moduleimport, "$module", module);
        Printv(f_shadow_py, moduleimport, "\n", NIL);
```
which did fix the problem by ensuring that the import always comes first, but then I noticed that these checks of `_swig_python_version_info` seem to be completely unnecessary any more and so we can just remove them, which is what this commit does.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -733,12 +733,9 @@ public:
          */
         Printf(default_import_code, "# pull in all the attributes from %s\n", module);
         Printv(default_import_code, "if __name__.rpartition('.')[0] != '':\n", NULL);
-        Printv(default_import_code, tab4, "if _swig_python_version_info >= (2, 7, 0):\n", NULL);
-        Printv(default_import_code, tab8, "try:\n", NULL);
-        Printf(default_import_code, tab8 tab4 "from .%s import *\n", module);
-        Printv(default_import_code, tab8 "except ImportError:\n", NULL);
-        Printf(default_import_code, tab8 tab4 "from %s import *\n", module);
-        Printv(default_import_code, tab4, "else:\n", NULL);
+        Printv(default_import_code, tab4, "try:\n", NULL);
+        Printf(default_import_code, tab8 "from .%s import *\n", module);
+        Printv(default_import_code, tab4 "except ImportError:\n", NULL);
         Printf(default_import_code, tab8 "from %s import *\n", module);
         Printv(default_import_code, "else:\n", NULL);
         Printf(default_import_code, tab4 "from %s import *\n", module);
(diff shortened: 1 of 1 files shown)


wsfulton:
@vadz, sorry, I missed checking this builtin example with my last commit. Checking the minimum Python version at runtime is a new more general helpful check for users committed recently, so I'd like to leave it in. I've fixed it in e729a868b91a11eabf88b5c420351a80b43a07fd.

vadz:
The same thing seems to have been done independently in 865dc1e71c3d2f3e35197d2cdd188c4b878cbb7d, so closing this one.

vadz:
Oops, sorry, I saw your commit but not your reply, please ignore my previous comment.

PR #1366, opened 2018-11-26
Try to locate php-config if there is no php-config7.X

wizzard:
Gentoo based distros have `php-config` configuration file installed in the same directory as `php` executable.
This patch adds a fallback to `php-config` if no `php-config7.X` configuration file is found for PHP7.

---
This change is [Reviewable](https://reviewable.io/reviews/swig/swig/1366)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2015,7 +2015,12 @@ else
       dnl /usr/bin/php7.0 -> /usr/bin/php-config7.0
       case $PHP in
         *7.*)
-          PHPCONFIG=`echo "$PHP"|sed 's/7\...*$/-config&/'` ;;
+          PHPCONFIG=`echo "$PHP"|sed 's/7\...*$/-config&/'`
+          # try to see if there is php-config in the same directory
+          if test ! -f $PHPCONFIG; then
+            PHPCONFIG=$PHP-config
+          fi
+          ;;
         *)
           PHPCONFIG=$PHP-config ;;
       esac


wsfulton:
Seems reasonable to me. I'll merge across if @ojwb doesn't object.

@wizzard, did you do anything in particular for this pull request to show the 'Reviewable' button?

wizzard:
@wsfulton didn't do anything special, committed from the console as I always do. This button never displayed in my other pulls.
I'm not sure, but maybe the reason is that I registered on reviewable site a while ago and their application is in my "Authorized OAuth Apps" in GitHub settings?

wsfulton:
Oops, completely missed the CI failures on Travis. Clearly they shouldn't break.

ojwb:
The CI failure is because the `-f` check fails when there's `PHP` is specified without a full path and so overrides `PHPCONFIG` from a working value to a non-working one.

But looking more closely, the code being patched is currently:

```
      case $PHP in
        *7.*)
          PHPCONFIG=`echo "$PHP"|sed 's/7\...*$/-config&/'` ;;
        *)
          PHPCONFIG=$PHP-config ;;
      esac
```

So if `$PHP` doesn't have a version in it, we just append `-config` already!

AFAICS, there's no bug here to fix.  If there is a bug, please tell us exactly what configure command line you're using, what `php*` programs are on `PATH`, and what `PHP` and `PHPCONFIG` values configure ends up trying to use.

ojwb:
No response from submitter for 4 weeks so closing.

If you still believe there's a bug here, please provide the information requested in my previous comment.

Issue #1368, opened 2018-11-27
[Python] AV in tp_print caused by mismatched Python/extension CRT usage


adr26:
The Python type API [`tp_print`](https://docs.python.org/2/c-api/typeobj.html?highlight=tp_print#c.PyTypeObject.tp_print) is documented to take a parameter of type `FILE*`:

`int tp_print(PyObject *self, FILE *file, int flags)`

If a Python extension, built using SWIG, uses a different CRT from that used to build Python itself (e.g. on Windows, the extension was built using a different version of Visual Studio than was used to build the Python release), then the Python release binaries invoking an implementation of `tp_print` from the extension will provide a parameter of type `FILE` from a CRT which is incompatible with the CRT used by the extension itself.

To see this in action, simply compiling the example `\Examples\python\simple` and attempting to print the module's dictionary as follows will create an AV:
```
C:\workspaces\swig\Examples\python\simple>python
Python 2.7.15 (v2.7.15:ca079a3ea3, Apr 30 2018, 16:22:17) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import _example
>>> print _example.__dict__
{'cvar':
```
Attaching a debugger, you see:
```
Exception thrown at 0x77B62E19 (ntdll.dll) in python.exe: 0xC0000005: Access violation writing location 0x00000014.
Unhandled exception at 0x77B62E19 (ntdll.dll) in python.exe: 0xC0000005: Access violation writing location 0x00000014.

>   ntdll.dll!RtlpWaitOnCriticalSection()                                                                                     Unknown
    ntdll.dll!RtlpEnterCriticalSectionContended()                                                                             Unknown
    ntdll.dll!_RtlEnterCriticalSection@4()                                                                                    Unknown
    _example.pyd!_lock_file(_iobuf * stream)                                                                                  Line 141    C++
    _example.pyd!__acrt_lock_stream_and_call::__l3::<lambda>()                                                                Line 253    C++
    _example.pyd!__crt_seh_guarded_call<int>::operator()<void <lambda>(void),int <lambda>(void) &,void <lambda>(void) >(__acrt_lock_stream_and_call::__l3::void <lambda>(void) && setup, common_vfprintf::__l11::int <lambda>(void) & action, __acrt_lock_stream_and_call::__l4::void <lambda>(void) && cleanup) Line 197    C++
[...]

PR #1369, opened 2018-11-28
[Python] #1368: AV in tp_print caused by mismatched Python/extension CRT usage

adr26:
Removes use of `tp_print` throughout, for reasons explained in #1368.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,13 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.0 (in progress)
 ===========================
 
+2018-12-04: adr26
+            [Python] #1368 AV in tp_print caused by mismatched Python/extension CRT usage
+
+            Remove all use of tp_print, as this API uses a FILE*, which can be
+            mismatched when modules are built with different C libraries from
+            the main python executable.
+
 2018-12-04: wsfulton
             [Python] #1282 Fix running 'python -m'  when using 'swig -builtin'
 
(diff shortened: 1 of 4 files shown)


adr26:
Hi @wsfulton,

Could you take a look at these changes (and #1368) and let me know if you think this is the right thing way to fix that issue?

Kind regards,

Andrew R

wsfulton:
Some observations...
```python
x = example.cvar
print(x)
print(str(x))
print(repr(x))
```
Used to output (when wrapping three global variables, `Eek`, `Bar`, `Foo`):
```
Swig global variables (Eek, Bar, Foo)

(Eek, Bar, Foo)
<Swig global variables>
```
Now the output is:
```
(Eek, Bar, Foo)
(Eek, Bar, Foo)
<Swig global variables>
```
and for when `x` is a packed SWIG object (such as a callback constant), the
output changes from
```
<Swig Packed at _5644f71f2d7f00000000000000000000_m_A__f_int__int>
_5644f71f2d7f00000000000000000000_m_A__f_int__int
<Swig Packed at _5644f71f2d7f00000000000000000000_m_A__f_int__int>
```
to
```
_aac4557d5d7f00000000000000000000_m_A__f_int__int
_aac4557d5d7f00000000000000000000_m_A__f_int__int
<Swig Packed at _aac4557d5d7f00000000000000000000_m_A__f_int__int>
```
More specifically, the output changes for Python 2 only. Python 3 no longer calls tp_print, see https://docs.python.org/3.7/c-api/typeobj.html?highlight=tp_print#c.PyTypeObject.tp_print. By dropping the calls to tp_print, the patch results in consistent output in both Python 2 and Python 3. So although the output changes for both cvar and packed objects, this is good. There is no change in output for SWIG proxy classes nor builtin types (when using -builtin) as using `print(x)`, where x is one of these objects, was already showing the same output as `print(str(x))`.

wsfulton:
Thanks for the patch. I've merged this to master now. I reverted the removal of the tp_print slot in python.cxx, see 90221812c67bb94180e27c6f58afbb99bf40fc5b. I also removed the tp_print support functions that are no longer used, see b8fc71b5f875f85297cf79c7cb85e339e96c04c0.

PR #1376, opened 2018-12-13
Travis CI hardening

tlby:
`apt-get update` producing a `Hash sum mismatch` seems to come from three common sources:

 1) ancient apt version trying to interact with modern repositories
 2) misbehaving apt proxies
 3) malformed apt-get resource files

I don't think 1 is true in our case, we can't do much about 2, but 3 is an easy fix provided here.

In another PR I got some Travis CI failures at the `apt-get update` step.  It looks like someone requested a retest on my behalf because 12 hours later the build went green seemingly on its own.  If that problem is common, then this update _may_ help.


tlby:
@wsfulton this patch tries to make [these errors](https://travis-ci.org/tlby/swig/jobs/466770670) a little less common.  If this is not a common problem, then feel free to discard this PR.

wsfulton:
@tlby, that was me who restarted the broken builds in other PR (#1375). The intermittent Travis failures with 'apt-get' are frustrating, and I tend to just restart the failed builds if need be.

PR #1384, opened 2018-12-26
%attribute usage leads to SWIG warnings on macros within C++ header files

gkv311:
Using %attribute within project .i file like this:
```
%attribute(Quantity_Color, double, red, Red);
```
leads to SWIG emitting warning(s) on C++ header file like this:
```
src/Aspect\Aspect_DisplayConnection.hxx(23) : Warning 202: Could not evaluate expression '!defined(_WIN32) && (!defined(__APPLE__) || defined(MACOSX_USE_GLX)) && !defined(__ANDROID__) && !defined(__QNX__)'
```

Wrapped class has nothing to do with specific header file, and SWIG generation is clean if %attribute is not used anywhere:
```
#if !defined(_WIN32) && (!defined(__APPLE__) || defined(MACOSX_USE_GLX)) && !defined(__ANDROID__) && !defined(__QNX__)
  #include <InterfaceGraphic.hxx>
#endif
```

I have put macros checks inside %attribute into brackets.
I don't know what is the difference, but it solved the issue for me.
```
-#if #SetMethod != #AttributeName
+#if (#SetMethod != #AttributeName)
```

### DIFF
--- a/Lib/typemaps/attribute.swg
+++ b/Lib/typemaps/attribute.swg
@@ -144,7 +144,7 @@
 %define %attribute_custom(Class, AttributeType, AttributeName, GetMethod, SetMethod, GetMethodCall, SetMethodCall)
   %ignore Class::GetMethod();
   %ignore Class::GetMethod() const;
-  #if #SetMethod != #AttributeName
+  #if (#SetMethod != #AttributeName)
     %ignore Class::SetMethod;
   #endif
   %extend Class {
@@ -197,7 +197,7 @@
 // User macros
 
 %define %attribute(Class, AttributeType, AttributeName, GetMethod, SetMethod...)
-  #if #SetMethod != ""
+  #if (#SetMethod != "")
     %attribute_custom(%arg(Class), %arg(AttributeType), AttributeName, GetMethod, SetMethod, self_->GetMethod(), self_->SetMethod(val_))
   #else
     %attribute_readonly(%arg(Class), %arg(AttributeType), AttributeName, GetMethod, self_->GetMethod())
@@ -205,7 +205,7 @@
 %enddef
 
 %define %attribute2(Class, AttributeType, AttributeName, GetMethod, SetMethod...)
-  #if #SetMethod != ""
+  #if (#SetMethod != "")
     %attribute_custom(%arg(Class), %arg(AttributeType), AttributeName, GetMethod, SetMethod, &self_->GetMethod(), self_->SetMethod(*val_))
   #else
     %attribute_readonly(%arg(Class), %arg(AttributeType), AttributeName, GetMethod, &self_->GetMethod())
@@ -213,7 +213,7 @@
 %enddef
 
 %define %attributeref(Class, AttributeType, AttributeName, AccessorMethod...)
-  #if #AccessorMethod != ""
+  #if (#AccessorMethod != "")
     %attribute_custom(%arg(Class), %arg(AttributeType), AttributeName, AccessorMethod, AccessorMethod, self_->AccessorMethod(), self_->AccessorMethod() = val_)
   #else
     %attribute_custom(%arg(Class), %arg(AttributeType), AttributeName, AttributeName, AttributeName, self_->AttributeName(), self_->AttributeName() = val_)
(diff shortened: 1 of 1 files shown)


vadz:
Sorry but this change really doesn't make any sense to me, why should parsing of this preprocessor condition be affected by using or not using attribute at all? Can you try reproducing the problem with a small and simple example?

gkv311:
See sample attached.
[swig_test.zip](https://github.com/swig/swig/files/2710392/swig_test.zip)

```
%include "attribute.i"
#define Standard_DEPRECATED(a)
#define DEFINE_STANDARD_RTTIEXT(a,b)
#define DEFINE_STANDARD_HANDLE(a,b)
%attribute(Aspect_DisplayConnection, bool, value, getValue)
%{
#include <Aspect_DisplayConnection.hxx>
%}
%include <Aspect_DisplayConnection.hxx>
```
```
swig.exe -c++ -Wextra -w309 -csharp -fcompact -fvirtual -DSWIG2_CSHARP -D_WIN32 "-I%~dp0src" -outdir "%~dp0Wrappers" "%~dp0wrap_test.i"
```

http://git.dev.opencascade.org/gitweb/?p=occt.git;a=blob;f=src/Aspect/Aspect_DisplayConnection.hxx;h=d2080a7b58ea441da9cad98fcd479aac1a46d268;hb=HEAD

wsfulton:
There is possibly something wrong in the preprocessor and this might work around it. Needs investigation, but my knowledge on the preprocessor isn't the best, so I'll defer. The code that @gkv311 put brackets around might even be a preprocessor extension.

I can reproduce on linux by unzipping the file and then simply running:
```
 $ swig -c++ -Wextra -w309 -csharp -fcompact -fvirtual -DSWIG2_CSHARP -D_WIN32 "-Isrc" -outdir "." "wrap_test.i"
src/Aspect_DisplayConnection.hxx:23: Warning 202: Could not evaluate expression '!defined(_WIN32) && (!defined(__APPLE__) || defined(MACOSX_USE_GLX)) && !defined(__ANDROID__) && !defined(__QNX__)'
src/Aspect_DisplayConnection.hxx:23: Warning 202: Error: 'Can't mix strings and integers in expression'
src/Aspect_DisplayConnection.hxx:42: Warning 202: Could not evaluate expression '!defined(_WIN32) && (!defined(__APPLE__) || defined(MACOSX_USE_GLX)) && !defined(__ANDROID__) && !defined(__QNX__)'
src/Aspect_DisplayConnection.hxx:42: Warning 202: Error: 'Can't mix strings and integers in expression'
```

gkv311:
Please let me know if I need registering a bug for this issue, or this pull request is enough (I'm not very familiar with github workflow).

wsfulton:
This pull request logs this as a bug. The pull request as it stands look more like a workaround than a fix. I may be wrong, but in order to accept the pull request, I need to know if this is a real fix or if the fix needs to be done in the SWIG preprocessor. Further info such as a comparison of SWIG's preprocessed output against the output from a C preprocessor (say gcc's) would help.

ojwb:
I think I've tracked down this bug to `expr.c`:

```
@@ -392,6 +409,7 @@ int Preprocessor_expr(DOH *s, int *error) {
          goto extra_rparen;
        stack[sp - 1].op = EXPR_VALUE;
        stack[sp - 1].value = stack[sp].value;
+       stack[sp - 1].svalue = stack[sp].svalue;
        sp--;
        break;
       default:
```

With that change `(X)` in a preprocessor conditional expression takes the same string vs int type as the `X` inside and the problem here goes away.  The reason why this testcase triggers this is presumably because it happens to leave `svalue` in that stack entry set.

I'll sort out a testcase, and also check that there aren't any other missing initialisations of fields in stack entries.

ojwb:
Minimised testcase:

```
%module test
#if "" != ""
#endif
#if 1 && (!1)
#endif
```

```
$ swig -module test -c++ -Wextra -python reduced.i
reduced.i:4: Warning 202: Could not evaluate expression '1 && (!1)'
reduced.i:4: Warning 202: Error: 'Can't mix strings and integers in expression'
```

ojwb:
And an alternative version where a single expression breaks:

```
%module test
#if ((("" != ""))) && (1 && (!1))
#endif
```

```
$ swig -module test -c++ -Wextra -python reduced.i
reduced2.i:2: Warning 202: Could not evaluate expression '((("" != ""))) && (1 && (!1))'
reduced2.i:2: Warning 202: Error: 'Can't mix strings and integers in expression'
```

ojwb:
Aha, and we can make a testcase which fails without needing `-Wextra -Werror` because when the preprocessor can't evaluate the expression it treats it as `0`:

```
%module test
#if "" != ""
#endif
#if 1 && (!0)
// Should go here
#else
# error BUG
#endif
#if ((("" == ""))) || (1 && (!1))
// Should go here
#else
# error BUG
#endif
```

```
$ swig -module test -python errout.i 
errout.i:7: Error: CPP #error "BUG". Use the -cpperraswarn option to continue swig processing.
errout.i:12: Error: CPP #error "BUG". Use the -cpperraswarn option to continue swig processing.
```

I've verified that `svalue` is always set when it needs to be aside from the one missing place, so I'll put all the pieces together and push a fix.  I'll also add some comments where I had to work stuff out for myself for the benefit of the next person to venture into this code.

Issue #364, opened 2015-03-24
ocaml test suite needs some love


amitkumarghatwal:
Hi All,

I have been able to build/install SWIG successfully. The test suite gets configured as below:
/*************_/
The SWIG test-suite and examples are configured for the following languages:
ocaml perl5
/**_***********/

Perl5 test cases are passing , but when i run "make -k check" . i am getting the below errors.

/*************_/
checking Examples/ocaml/simple
File "swig.ml", line 159, characters 54-57:
Warning 20: this argument will not be used by the function.
/bin/sh: camlp4: command not found
File "swigp4.ml", line 18, characters 8-10:
Error: Syntax error
/**_***********/

From where do we install - camlp4 package ? Any pointers in this regards would be helpful.
Regards,
Amit

ojwb:
You don't say what platform you're using, but on Debian, `camlp4` is in the package also called `camlp4`.

However, the ocaml test suite isn't in great shape currently, so you will still see failures - see my recent mailing list posts about that:

http://thread.gmane.org/gmane.comp.programming.swig.devel/23627

If you know enough about ocaml to solve those, help would be appreciated.  I know practically nothing about it.

ojwb:
We should be probing for camlp4 rather than using the literal tool name in the makefile - I've fixed that, but it won't help the underlying testsuite problems.

amitkumarghatwal:
Hello ojwb,

Sorry about not mentioning my setup. I have tried  running ocaml test suite on SLES12 platform. I personally don't have much info on camlp4.

I did try the ocaml test suite again after ur latest commit. It fails with different errors now though as below : 
/*********************************************/
checking Examples/ocaml/simple
File "swig.ml", line 159, characters 54-57:
Warning 20: this argument will not be used by the function.
File "swigp4.ml", line 18, characters 8-10:
Error: Syntax error
../../Makefile:898: recipe for target 'ocaml_static' failed
make[3]: *** [ocaml_static] Error 2
/*********************************************/

Has anyone able to successfully run the "ocaml test suite" on any platform ?

Regards,
Amit

ojwb:
I doubt anyone has recently.  I had a go at a `git bisect` but didn't get anywhere useful.  I suspect it's a combination of changes to the ocaml toolchain which SWIG needs updating for, new testcases which highlight existing flaws in SWIG's ocaml backend, and possibly changes to the generic SWIG code which the ocaml backend hasn't been full updated for.

[...]

PR #1386, opened 2018-12-29
[OCaml] Update OCaml for out-of-source builds

ZackerySpytz:
This is one of the steps needed to fix the OCaml test suite.

In addition, disable the creation of toplevels by default in the OCaml
examples (toplevels are currently broken).

### DIFF
--- a/Examples/Makefile.in
+++ b/Examples/Makefile.in
@@ -917,8 +917,8 @@ ocaml_static: $(SRCDIR_SRCS)
 	$(SWIG) -ocaml $(SWIGOPT) -o $(ISRCS) $(INTERFACEPATH)
 	$(OCC) -g -c -ccopt -g -ccopt "$(INCLUDES)" $(ISRCS) $(SRCDIR_SRCS)
 	$(OCC) -g -c $(INTERFACE:%.i=%.mli)
-	$(OCC) -g -c $(INTERFACE:%.i=%.ml)
-	test -z "$(PROGFILE)" || $(OCC) $(OCAMLPP) -c $(PROGFILE)
+	$(OCC) -w -U -g -c $(INTERFACE:%.i=%.ml)
+	test -z "$(PROGFILE)" || $(OCC) -o $(PROGFILE:%.ml=%) $(OCAMLPP) -c $(SRCDIR)$(PROGFILE)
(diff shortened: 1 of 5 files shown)


wsfulton:
A couple of observations that need addressing:
1) If the `_runme.ml` file has an error, make does NOT fail.
2) The `_runme.cmi` and `_runme.cmo` are generated into the source tree for out-of-source builds. This is fundamentally wrong. These files are then moved into the build tree. The source tree might get contaminated if the mv fails. Also if the mv fails, the make should fail. Redirecting the mv output to /dev/null shouldn't be done, it makes debugging very hard in the event of errors! For in-source builds, the mv is unnecessary. All round, the output should go directly into the build tree, not the source tree.

Having said all that, it is great to see the output in Travis now as Travis does out-of-source builds. Without Travis working, OCaml was on the list of languages to be removed in the forthcoming 4.0.0 release as we are not going to maintain any languages that aren't properly setup for testing. So you may just save OCaml from the chop... big thanks.

ZackerySpytz:
Thank you, @wsfulton, for the comments. I'll respond soon.

ZackerySpytz:
> If the _runme.ml file has an error, make does NOT fail.

Actually, `make` did fail in this case, but for an unrelated reason: the OCaml compiler removes the `.cmo` file (if it exists) before generating a new one, and if compilation of the `*_runme.ml` fails, the missing `.cmo` file will end up causing a failure (this was far from preferable). It should properly fail now (for what it's worth, this was an issue with the preexisting code).

> Actually, the test-suite was working on Travis out-of-source builds

The test suite was not working for out-of-source builds. On Travis, (for example, https://travis-ci.org/swig/swig/jobs/471124176#L1187) one can see that runme tests fail with the following:
```
checking ocaml testcase class_ignore (with run test)
File "class_ignore.ml", line 93, characters 13-14:
Warning 11: this match case is unused.
File "_none_", line 1:
Error: Cannot find file ../../../../../Examples/test-suite/ocaml/class_ignore.cmo
```
 The changes in Examples/test-suite/ocaml/Makefile.in fix this.

> These messages no longer appear:

I added `-w -S` to temporarily disable them, as they fill up the build logs with cruft. The OCaml module needs to generate cleaner code, but I think the OCaml test suite should be fixed first.

wsfulton:
Thanks for explanations. This certainly looks better now, so I'll merge it across. I had some problems initially as I tested with `-j8`. Probably it is a lurking issue that was always there. I've added this info to #364.

PR #1390, opened 2018-12-31
Install xml files

### LINKED ISSUE #1388: Lib/xml is not installed
The Lib/xml/{typemaps.i,xml.swg} files are not installed (the lib-languages var does not contain xml in Makefile.in)
Probably they either are missing at install or useless and should be removed.

xantares:
I suppose it fixes #1388


wsfulton:
The files are actually empty, but it makes sense that Lib/xml files are installed in case this changes in the future.

PR #1393, opened 2019-01-03
[OCaml] Fix some of the OCaml examples

ZackerySpytz:
Update some of the OCaml examples so that they run successfully with
`make check-ocaml-examples` (important for the CI). Some of the
examples were written to depend on passed arguments.

Temporarily disable the broken std_string example.

`make check-ocaml-examples` should now pass on Travis.


wsfulton:
Hi @ZackerySpytz, this is very helpful, thanks. 

One of the important requirements for a SWIG target language is to have the 'simple' and 'class' examples. I don't suppose you'd be able to add then convert the 'class' example from one of the other languages, such as Examples/python/class to give the equivalent output?

There is also one minor issue with the OCaml examples and that is the naming of the example scripts were renamed a while back to all be called runme. So all the various names such as example_prog.ml and example.ml should be renamed runme.ml. The example Makefiles should also be simpler as a result. Could you make this change for this pull request too please?

All the module names should also be `%module example`. I see string_from_ptr is called `foolib`.

ZackerySpytz:
Thank you for reviewing and merging my pull requests, @wsfulton. I will make the changes you've requested.

ZackerySpytz:
@wsfulton When creating the OCaml class example, I encountered a severe bug in the OCaml module: 
the getters and setters of public member variables do not work. I'll create a PR for this bug. A hacky fix would require just a small change in ocaml.cxx, but a proper fix would require a lot of changes in ocaml.cxx, which is risky considering the poor state of the OCaml test suite (and the lack of runme tests).

wsfulton:
A quick fix for now is okay given the immature and unknown state of the OCaml module. Looks like the fix is in #1400 which I'll merge in. Thanks.

wsfulton:
This fixes the examples so I've merged it over. Thanks. Could you please add in the 'class' example separately?

ZackerySpytz:
Thanks for the merge.
> Could you please add in the 'class' example separately?

I'll create a PR for that, but #1400 needs to be sorted out first.

ZackerySpytz:
With #1400 merged, I've created #1407 for the OCaml class example.

PR #1398, opened 2019-01-06
Add missing includes to library and test cases

sethrj:
Many of these tests implicitly required the target language library files to include the headers upstream. This change allows tests in future language modules to pass without requiring those modules' base library to unconditionally include these headers.


sethrj:
Thanks @ojwb !

PR #1399, opened 2019-01-06
Add support for parsing curly braces in typemap kwargs

sethrj:
This enables more advanced typemap construction (using macro concatenation) inside of typemap keyword arguments, which (for Java, C#, and Fortran) can require extra complexity.

### DIFF
--- a/Doc/Manual/Typemaps.html
+++ b/Doc/Manual/Typemaps.html
@@ -734,7 +734,7 @@ these methods is described later.
 </p>
 
 <p>
-<em>modifiers</em> is an optional comma separated list of <tt>name="value"</tt> values. 
+<em>modifiers</em> is an optional comma separated list of <tt>name="value"</tt> or <tt>name={value}</tt> values. 
 These are sometimes to attach extra information to a typemap and is often target-language dependent.
 They are also known as typemap attributes.
 </p>
@@ -803,6 +803,7 @@ Here are some examples of valid typemap specifications:
 
 /* Typemap with modifiers */
 %typemap(in, doc="integer") int "$1 = scm_to_int($input);";
+%typemap(in, doc={embedded MACRO will be expanded}) int "$1 = scm_to_int($input);";
 
 /* Typemap applied to patterns of multiple arguments */
 %typemap(in) (char *str, int len),
--- a/Examples/test-suite/typemap_args.i
+++ b/Examples/test-suite/typemap_args.i
@@ -1,0 +1,32 @@
+%module typemap_args
+
+%define MYTYPECHECK_int_YAY    SWIG_TYPECHECK_INTEGER %enddef
+%define MYTYPECHECK_double_YAY SWIG_TYPECHECK_DOUBLE  %enddef
+
+%fragment("includeme"{int}, "header") %{
+#define INCLUDED_INT_TYPECHECK
+%}
+%fragment("includeme"{double}, "header") %{
+#define INCLUDED_DOUBLE_TYPECHECK
+%}
+
+%define %add_typecheck(TYPE)
+%typemap(typecheck, precedence={MYTYPECHECK_ ## TYPE ## _YAY}, fragment="includeme"{TYPE}) TYPE {
+  $1 = (TYPE)$1;
+}
+%enddef
+
+%add_typecheck(int)
+%add_typecheck(double)
+
+%inline %{
+int foo(int a) { return a; }
+double foo(double a) { return a * 2; }
+%}
+
+%{
+/* If these two lines don't compile out, it's because the
+ * necessary fragment wasn't included */
+INCLUDED_INT_TYPECHECK;
+INCLUDED_DOUBLE_TYPECHECK;
+%}
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -507,6 +507,7 @@ CPP_TEST_CASES += \
 	typedef_sizet \
 	typedef_struct_cpp \
 	typedef_typedef \
+	typemap_args \
 	typemap_arrays \
 	typemap_array_qualifiers \
 	typemap_delete \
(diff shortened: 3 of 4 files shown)


Doc/Manual/Typemaps.html
```diff
@@ -734,7 +734,7 @@ <H3><a name="Typemaps_defining">12.2.1 Defining a typemap</a></H3>
 </p>
 
 <p>
-<em>modifiers</em> is an optional comma separated list of <tt>name="value"</tt> values. 
+<em>modifiers</em> is an optional comma separated list of <tt>name="value"</tt> or <tt>name={value}</tt> values.
```
ojwb, line 737: It seems the hblock style (e.g. `name=%{value with a " in it%}`) is also now supported, which seems reasonable for consistency with typemap bodies, but it ought to be documented here too.

Examples/test-suite/typemap_args.i
```diff
@@ -0,0 +1,32 @@
...
+int foo(int a) { return a; }
+double foo(double a) { return a * 2; }
+%}
+
+%{
+/* If these two lines don't compile out, it's because the
+ * necessary fragment wasn't included */
+INCLUDED_INT_TYPECHECK;
+INCLUDED_DOUBLE_TYPECHECK;
+%}
```
ojwb, line 32: And the hblock style should be tested too.

Also I think it would make sense to test that macros aren't expanded in `foo=""` and `foo=%{%}` as well as that they are in `{}`.

ojwb:
@wsfulton Are you happy with this as a new feature?  Github says about you mentioned in in #1195 but I couldn't find where.

Doc/Manual/Typemaps.html
ojwb: We should probably also say "since SWIG 4.1.0" or whatever for new features like this.

sethrj:
@ojwb It's been a long time since I've looked at this, but I believe on my fork I ended up having to apply a fix (and add a unit test) for this feature. I will update this branch with any further changes I've made, merge master in to ensure all the unit tests still pass with upstream changes, and then ping you for review again? (This will be one of the small PRs I reference in https://github.com/swig/swig/pull/1195#issuecomment-1026632267 .)

ojwb:
Sounds good, though I'd really like William's thumbs up on this as a new feature.  The functionality seems useful to me, and it's nice that it follows the existing syntax for typemap bodies rather than inventing a whole new syntax, but I don't want to get into commit/revert wars.

wsfulton:
I vaguely remember it, but also can't work out what I thought of it at the time. Looking at it now, I'm wondering if the current preprocessor can not be used as is instead of adding in what is a rather quirky way of doing the same thing. For example `%add_typecheck` in the example can be quite simply rewritten as follows:
```swig
%define %add_typecheck(TYPE)
%typemap(typecheck, precedence=MYTYPECHECK_ ## TYPE ## _YAY, fragment="includeme" #TYPE) TYPE {
  $1 = (TYPE)$1;
}
%enddef
```
resulting in the preprocessed output that looks okay to me:
```swig
/*@SWIG:example.i,13,%add_typecheck@*/
%typemap(typecheck, precedence=         70     , fragment="includeme""int") int {
  $1 = (int)$1;
}
/*@SWIG@*/
/*@SWIG:example.i,13,%add_typecheck@*/
%typemap(typecheck, precedence=       90      , fragment="includeme""double") double {
  $1 = (double)$1;
}
```

ojwb:
Ah yes, using existing cpp-style features like that does seem cleaner.

@sethrj Does that work for you?  If so, let's close this.

sethrj:
@ojwb @wsfulton I recall my need for this feature evolving over time, but I think the simplest motivating example is a macro for setting up fundamental types, which have different "in" and "out" behaviors.
```lex
%define %fortran_intrinsic(CTYPE, FTYPE, FKIND)
  %fortran_apply_typemaps(FORTRAN_INTRINSIC_TYPE, CTYPE)
  
  // Regular values are passed as pointers and returned by value
  %typemap(ctype, in={CTYPE*}) CTYPE
   %{CTYPE%}
  %typemap(imtype, in={FTYPE(FKIND), intent(in)}) CTYPE
   %{FTYPE(FKIND)%}
```
I believe the special character `*` prevents the typemap from being unquoted, but it looks like I can use string concatenation in this case:
```
  %typemap(ctype, in=#CTYPE ## "*") CTYPE
   %{CTYPE%}

  %typemap(imtype, in=#FTYPE ## "(" ## #FKIND ## "), intent(in)") CTYPE
   %{FTYPE(FKIND)%}
```
Of course, the proliferation of `#` signs here makes the typemap even less legible.

I will convert the typemaps if you think the proposed extension is too unusual to justify as part of the main parser, but I think the proposed syntax is a reasonable extension to SWIG's parser given that `%typemap(in) {TYPE*}` was devised as a simpler alternative to `%typemap(in) #TYPE ## "*"`.

ojwb:
You shouldn't need to token paste strings, which reduces the proliferation of `#` significantly:

```
%typemap(ctype, in=#CTYPE "*") CTYPE
%{CTYPE%}

%typemap(imtype, in=#FTYPE "(" #FKIND "), intent(in)") CTYPE
%{FTYPE(FKIND)%}
```

There are existing uses of this in both the library and testsuite so it should work, for example:

```
Examples/test-suite/preproc.i:     %typemap(argout, doc="($arg <vector of <" #SCM_TYPE ">>)") 
Lib/guile/list-vector.i:     %typemap(in, doc="$NAME is a vector of " #SCM_TYPE " values")
```

sethrj:
Oh! That's good to know. I'll see if there are any other use cases I missed and if there's nothing new that this actually enables I will close it. Thanks!

sethrj:
Thanks for the recommendation @ojwb @wsfulton ; it looks like I am indeed able to use the feature that's already there. I will close this PR.

PR #1403, opened 2019-01-09
[OCaml] Use int64_t instead of int64 for OCaml versions >= 4.03.0

### LINKED ISSUE #1194: Unknown int64 error in compiling ocaml examples
Hi! I am new to swig and want to wrap c++ program in ocaml.
I cloned swig from master branch and tried an ocaml example.
Got the following results:

    ➜  ocaml git:(master) ✗ cd std_vector
    ➜  std_vector git:(master) ✗ ls
    Makefile  example.h example.i runme.ml
    ➜  std_vector git:(master) ✗ make
    /Library/Developer/CommandLineTools/usr/bin/make -f ../../Makefile SRCDIR='' SRCS='' \
 	PROGFILE='runme.ml' TARGET='example' INTERFACE='example.i' \
	ocaml_static_cpp
    rm -rf swig.mli swig.ml swigp4.ml && env SWIG_LIB=../../../Lib  ../../../swig -ocaml -co swig.mli 
    2>/dev/null && env SWIG_LIB=../../../Lib  ../../../swig -ocaml -co swig.ml 2>/dev/null && env 
    SWIG_LIB=../../../Lib  ../../../swig -ocaml -co swigp4.ml 2>/dev/null &&  ocamlc -c swig.mli &&  
    ocamlc -c swig.ml &&  ocamlc -I ` camlp4 -where` -pp "camlp4o pa_extend.cmo q_MLast.cmo" -
    c swigp4.ml
    env SWIG_LIB=../../../Lib  ../../../swig -ocaml -c++  -o example_wrap.cxx example.i
    cp example_wrap.cxx example_wrap.c
    ocamlc -cc 'g++ -Wno-write-strings' -g -c -ccopt -g -ccopt "-xc++ " example_wrap.c
    example_wrap.c:1039:51: error: use of undeclared identifier 'int64'
            swig_type_info *ty = (swig_type_info *)(long)SWIG_Int64_val(SWIG_Field(v,1));
                                                         ^
    example_wrap.c:845:31: note: expanded from macro 'SWIG_Int64_val'
    #define SWIG_Int64_val(v) (*((int64 *) SWIG_Data_custom_val(v)))
                              ^
    example_wrap.c:1039:51: error: expected expression
    example_wrap.c:845:38: note: expanded from macro 'SWIG_Int64_val'
    #define SWIG_Int64_val(v) (*((int64 *) SWIG_Data_custom_val(v)))

    ➜  std_vector git:(master) ✗ gcc --version
    gcc (Homebrew GCC 6.3.0_1) 6.3.0
    Copyright (C) 2016 Free Software Foundation, Inc.
    This is free software; see the source for copying conditions.  There is NO
    warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

The problem is that int64 does not seem to be defined anywhere.
I got these results on macOS SIerra and ocamlc version 4.05.0
 and got the same one when using clang 4.0 through setting -cc clang++ in ocamlc options.
Does this issue relate to my environment?

ZackerySpytz:
OCaml's int64 type was replaced with the C99 int64_t in OCaml 4.03.0.
https://github.com/ocaml/ocaml/commit/b868c05ec91a7ee193010a421de768a3b1a80952

Closes #1194.

### DIFF
--- a/Lib/ocaml/ocamldec.swg
+++ b/Lib/ocaml/ocamldec.swg
@@ -23,6 +23,7 @@ SWIGEXT {
 #include <caml/callback.h>
 #include <caml/fail.h>
 #include <caml/misc.h>
+#include <caml/version.h>
 
 #define caml_array_set swig_caml_array_set
 
@@ -101,9 +102,17 @@ SWIGEXT {
 
 
 #ifndef ARCH_ALIGN_INT64
+#if OCAML_VERSION >= 40300
+#define SWIG_Int64_val(v) (*((int64_t *) SWIG_Data_custom_val(v)))
+#else
 #define SWIG_Int64_val(v) (*((int64 *) SWIG_Data_custom_val(v)))
+#endif
+#else
+#if OCAML_VERSION >= 40300
+CAMLextern int64_t Int64_val(caml_value_t v);
 #else
 CAMLextern int64 Int64_val(caml_value_t v);
+#endif
 #define SWIG_Int64_val(v) Int64_val(v)
 #endif
 


ZackerySpytz:
On Travis, the patch simply doesn't work:

```
example_wrap.c:741:26: fatal error: caml/version.h: No such file or directory
 #include <caml/version.h>
```
I'm not sure why.  The patch worked fine when I ran `make check-ocaml-test-suite` locally with 
both OCaml 4.01.0 (the version used on Travis) and OCaml 4.07.0.

wsfulton:
Suggests that caml/version.h is a new file and not available in 4.01.0? Can you check OCaml release notes?

wsfulton:
Here are the list of files in ocaml-nox on Ubuntu Trusty OCaml 4.01.0 (used by Travis) https://packages.ubuntu.com/trusty/amd64/ocaml-nox/filelist
Here are the list of files in ocaml-nox on Ubuntu Xenial (Ocaml 4.02.3) https://packages.ubuntu.com/xenial/amd64/ocaml-nox/filelist

It could be a packaging problem or simply a new file??

ZackerySpytz:
I haven't been able find any helpful details in the OCaml source, the OCaml git log, or https://github.com/ocaml/ocaml/blob/trunk/Changes.

wsfulton:
I did a bit of hunting around in the ocaml github repo and deduced that version.h is generated by https://github.com/ocaml/ocaml/blob/4.01.0/byterun/Makefile.common#L113 in 4.01.0 but I don't think it was actually used in this version. There is also a script https://github.com/ocaml/ocaml/blob/4.02.0/tools/make-version-header.sh in 4.02.0 which generates the headers.

I downloaded the 4.01.0 and 4.02.0 sources and built them using ` ./configure --prefix /tmp/ocamlinstall-4.01.0 && make world -s && make install`. Sure enough version.h is only in 4.02.0. Comparing the two directories containing the header files, I can see a few macros that were introduced in 4.02.0 in files that SWIG is already including:

misc.h: CAMLassert
alloc.h: value caml_alloc_sprintf

There are others via indirect includes. Anyway, it seems reasonable to say that version.h is available only if one of these is defined. I've chosen CAMLassert, so now we set OCAML_VERSION to 0 implying < 40200. See 48d9e3bc0204acf61c8da31ab03040515393086b

wsfulton:
If there is a cleaner way to deduce the version, then let's do that.

PR #1407, opened 2019-01-12
[OCaml] Add a class example 

ZackerySpytz:
It is based on the Python class example. The Makefile is based on the
Makefile used in the OCaml shapes example.

### DIFF
--- a/Examples/ocaml/check.list
+++ b/Examples/ocaml/check.list
@@ -1,11 +1,12 @@
 # see top-level Makefile.in
+argout_ref
+class
+contract
+scoped_enum
+shapes
 simple
 #std_string
 std_vector
 stl
-argout_ref
-shapes
-contract
-scoped_enum
 string_from_ptr
 strings_test
--- a/Examples/ocaml/class/Makefile
+++ b/Examples/ocaml/class/Makefile
@@ -1,0 +1,31 @@
+TOP          = ../..
+SWIGEXE      = $(TOP)/../swig
+SWIG_LIB_DIR = $(TOP)/../$(TOP_BUILDDIR_TO_TOP_SRCDIR)Lib
+SWIGOPT      =
+SRCS         = example.c
+TARGET       = example
+INTERFACE    = example.i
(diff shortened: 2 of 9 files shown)


wsfulton:
Okay, great! The shapes example seems redundant now with the class example added, so let's remove that. The shapes example, from which you based this is a bit odd as it is c++ code in an example.c file, which I've never seen before. The common class example should be using example.cxx too. Can you 'fix' that? I can take a look if it looks tricky.

ZackerySpytz:
> The shapes example seems redundant now with the class example added, so let's remove that.

I believe it would be best to keep the shapes example -- it is currently the only example of the use of Directors with OCaml.

> The common class example should be using example.cxx too.

No, there is a reason that example.c is named as such, and it is stated in the [OCaml
module's documentation](http://www.swig.org/Doc3.0/Ocaml.html#Ocaml) under "Compiling the code": `ocamlc` cannot handle .cxx, .cc, or .cpp files.

wsfulton:
Sorry, I completely missed the shapes examples is a director example. All the other languages supporting directors have the callback and extend examples for demonstrating directors. I'd appreciate those replacing the shapes examples if you fancy that.

I can see `%feature("director")` in Examples/ocaml/stl/example.i. I suspect that is just cruft.

Thanks for pointing out the ocamlc file naming restriction. Seems peculiar, but so be it.

ZackerySpytz:
Thanks for the merge. Adding the callback and extend examples for OCaml sounds good.

ZackerySpytz:
@wsfulton I've created #1415 for the callback example. I hit a bug in OCaml's director support when attempting to create the extend example. At this point, I'm not sure of the best way to fix it.

wsfulton:
I see, having the callback example is a step forward in the meantime, nice.

You're obviously putting in a lot of effort into the OCaml module and this is all great. May I suggest you join the swig-devel mailing list and if you would like to continue improving the OCaml module, introduce yourself and mention what you'd like to do going forwards and any motivation of interest.

PR #1410, opened 2019-01-14
[OCaml] Fix and re-enable the string_simple, minherit, and unions tests

ZackerySpytz:
copy_string() is a macro in the OCaml C API, so rename the function
in string_simple.i to copy_str(). Add a runtime test.

The minherit runtime test was fixed by b64d685.

Use the proper syntax for accessing member variables in
unions_runme.ml

For string_simple, an `ifdef` for OCaml could be used instead of the renaming.

### DIFF
--- a/Examples/test-suite/javascript/string_simple_runme.js
+++ b/Examples/test-suite/javascript/string_simple_runme.js
@@ -3,7 +3,7 @@ var string_simple = require("string_simple");
 // Test unicode string
 var str = "olé";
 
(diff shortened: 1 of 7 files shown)


wsfulton:
> For string_simple, an ifdef for OCaml could be used instead of the renaming.
I like to have as few preprocessor ifdef statements in the tests as possible, so renaming the variable and adjusting the other runtime tests to use the new name, as you've done, is my preferred approach too.

PR #1414, opened 2019-01-16
Better handling for parameters clashing with keywords for all languages

### LINKED ISSUE #1272: Python docstring: argument names conflicting with python keywords, are systematically turned into argN
Hi,

SWiG 2.0.12 used to rename python keyword-named C/C++ arguments, to an underscore-prefixed version, but SWiG 3.0.12 doesn't do than anymore (by default.)

Assuming an interface file such as:

    %module test
    %feature(autodoc,0);
    %feature("compactdefaultargs");
    %inline %{
    int process(int from);
    %}

using 2.0.12 we would get:

  """process(_from) -> int"""

But since the following changelist:

    commit fdc6bbeda387fb49ab7dcdbb0f17e4645db3115c
    Author: Vadim Zeitlin <vz-swig@zeitlins.org>
    Date:   Sat Aug 9 16:33:23 2014 +0200

this has changed and we now get:

    """process(arg1) -> int"""

…which unfortunately is less useful.

Using the `-keyword` option, apparently gets us to the old behavior. But unfortunately this has side-effects, other than just making the documentation more useful (in particular, well, enabling keyword arguments, which isn't something we desire)

Am I mistaken thinking that, regardless of whether or not kwargs are enabled for a given wrapper, the docstring shouldn't change? (or in any case not become less useful)

Thanks!

vadz:
As previously discussed, it seems like it would make sense to generalize Python handling of parameters clashing with the language keywords to all languages, so this PR does just this (and incidentally fixes #1272). I've tested it for the languages I use locally and it seems to work fine -- but I won't be surprised at all if it breaks something for the ones I didn't test...

Note that I couldn't add a part of the [proposed test](https://github.com/swig/swig/issues/1272#issuecomment-452201273):
```python
if process(_from=10) != 10:
```
doesn't work, neither before nor after my changes. I'm not sure how are keyword arguments supposed to be handled, but they just don't work here. This:
```python
if process(10) != 10:
```
works, but doesn't seem like a useful test.

### DIFF
--- a/Examples/test-suite/autodoc.i
+++ b/Examples/test-suite/autodoc.i
@@ -148,3 +148,11 @@ bool is_python_builtin() { return true; }
 bool is_python_builtin() { return false; }
 #endif
 %}
+
+// Autodoc Python keywords
+%warnfilter(SWIGWARN_PARSE_KEYWORD) process;
+%feature(autodoc,0) process;
+%feature("compactdefaultargs") process;
+%inline %{
+int process(int from) { return from; }
+%}
--- a/Examples/test-suite/python/keyword_rename_runme.py
+++ b/Examples/test-suite/python/keyword_rename_runme.py
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
 import keyword_rename
 keyword_rename._in(1)
+keyword_rename._in(_except=1)
 keyword_rename._except(1)
+keyword_rename._except(_in=1)
--- a/Examples/test-suite/errors/swig_typemap_warn.stderr
+++ b/Examples/test-suite/errors/swig_typemap_warn.stderr
@@ -3,5 +3,7 @@ swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int arg3 
 swig_typemap_warn.i:7: Warning 1001: Test warning for 'out' typemap for double mmm (result) - name: mmm symname: mmm &1_ltype: double * descriptor: SWIGTYPE_double
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int abc (arg1) - argnum: 1 &1_ltype: int * descriptor: SWIGTYPE_int
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int arg3 (arg3) - argnum: 3 &1_ltype: int * descriptor: SWIGTYPE_int
+swig_typemap_warn.i:7: Warning 314: 'def' is a python keyword, renaming to '_def'
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int abc (arg1) - argnum: 1 &1_ltype: int * descriptor: SWIGTYPE_int
 swig_typemap_warn.i:7: Warning 1000: Test warning for 'in' typemap for int arg3 (arg3) - argnum: 3 &1_ltype: int * descriptor: SWIGTYPE_int
+swig_typemap_warn.i:7: Warning 314: 'def' is a python keyword, renaming to '_def'
--- a/Examples/test-suite/keyword_rename.i
+++ b/Examples/test-suite/keyword_rename.i
@@ -4,6 +4,8 @@
 
 %module keyword_rename
 
+%feature("kwargs");
+
 #pragma SWIG nowarn=SWIGWARN_PARSE_KEYWORD
 
 %inline %{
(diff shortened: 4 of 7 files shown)


vadz:
I was just going to start fixing the errors in Travis builds, do you prefer to do it yourself or should I still go ahead and do it? I don't mind either choice, but I wouldn't want to duplicate the effort.

Examples/test-suite/autodoc.i
```diff
@@ -148,3 +148,11 @@ bool is_python_builtin() { return true; }
...
 #endif
 %}
+
+// Autodoc Python keywords
+%warnfilter(SWIGWARN_PARSE_KEYWORD) process;
+%feature(autodoc,0) process;
+%feature("compactdefaultargs") process;
+%inline %{
+int process(int from) { return from; }
+%}
```
wsfulton: Could you please add more tests combining "kwargs" and "compactdefaultargs" using keywords in the argument list? And also use optional parameters in the C++ function with keywords.

Examples/test-suite/python/keyword_rename_runme.py
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env python
 import keyword_rename
 keyword_rename._in(1)
+keyword_rename._in(_except=1)
```
wsfulton: `%feature("kwargs");` in keyword_rename.i should fix this for -fastproxy/-builtin builds.

vadz:
Just a ping to let you know that the tests have passed after the latest changes, so IMHO this should be ok to merge for SWIG 4.0.

wsfulton:
Thanks. There's a problem I've just spotted which means it isn't possible to suppress the warning using the usual warning suppression techniques (%warnfilter). `Swig_name_make` is for symbol renaming and is part of the general SWIG naming implementation. This includes all the %rename/%ignore machinery. The problem is extending it to parameters when it is only implemented for globally visible types and symbols. The usual %ignore/%rename/%feature patterns do not work because there is no C++ syntax to fully qualify a function's parameter name from outside the function.

Am pondering what to do. I'm tempted to just suppress the warning for parameter names until/if we implement something for attaching features to parameters.

vadz:
> Am pondering what to do. I'm tempted to just suppress the warning for parameter names until/if we implement something for attaching features to parameters.

First of all, I think the warnings are something that you really want to fix in any case (`_from` is better than `arg1`, but renaming the original parameter to something else so that it's both readable and not ugly would be even better), so I don't think it's horrible to just always give them to gently encourage people to fix them (and this is why I also don't really like suppressing these warnings unconditionally -- people wouldn't be even aware their parameters got mangled then). Second, I think we could attach something ("ignore-param-rename" feature?) to the function and check for it and disable the warnings for the parameters of this function if it's present. But I can't deny that this is rather ugly. Unfortunately I don't have any other ideas neither.

wsfulton:
Right now, I'm keen to get SWIG-4.0.0 out the door asap and so will defer putting in place a proper hooking up of parameter naming to %rename/%ignore to a later version. 

Renaming of duplicate parameters and adding in parameter names 'argN' has always silently occurred and this is the precedent. Before this change Python emitted a warning that could not be suppressed in a controlled manner (apart from editing a header which in many cases is not possible). This patch, as is, extends this bug to all the other languages, which is undesirable. Trivial warnings that can't be suppressed are generally a nuisance, if not universally hated, so I think we should silently suppress them for the time being. I'll code this up today before merging the patch. I like the fact that this patch renames a keyword in a sensible way rather than using 'argN'.

vadz:
> Renaming of duplicate parameters and adding in parameter names 'argN' has always silently occurred and this is the precedent. 

Good point.

> Before this change Python emitted a warning that could not be suppressed in a controlled manner (apart from editing a header which in many cases is not possible). This patch, as is, extends this bug to all the other languages, which is undesirable. Trivial warnings that can't be suppressed are generally a nuisance, if not universally hated, so I think we should silently suppress them for the time being. I'll code this up today before merging the patch.

OK, thanks. I won't be able to do it today, but if you don't manage to find time for it today neither, please let me know and I can try doing it myself during the week-end.

Thanks!

wsfulton:
@vadz, would you be able to port these parameter name improvements for autodoc to Ruby and Octave? They use very similar code to Python. I think it means fdc6bbeda387fb49ab7dcdbb0f17e4645db3115c, which introduced the use of makeParameterName into Python, and the recent Python changes (including those in this patch) copying over to Ruby and Octave.

PR #1417, opened 2019-01-17
[OCaml] Fix dead code generation in overloaded function wrappers

ZackerySpytz:
The OCaml module was generating dead code in the wrappers for
overloaded functions. Only the generated dispatch function needs to
allocate an array for the passed arguments.

In addition, add overload_extend, overload_rename and overload_subtype
runtime tests.


ZackerySpytz:
The Travis failures are unrelated to this patch.

wsfulton:
That's okay. Sometimes Travis has network problems. I restarted the failed builds, so it should go green shortly.

PR #1418, opened 2019-01-17
[OCaml] Improve the error message for incorrect overloaded function calls

ZackerySpytz:
List the possible prototypes in the error message. The code was taken
from python.cxx's dispatchFunction().

### DIFF
--- a/Source/Modules/ocaml.cxx
+++ b/Source/Modules/ocaml.cxx
@@ -733,7 +733,18 @@ public:
 	       "argv = (CAML_VALUE *)malloc( argc * sizeof( CAML_VALUE ) );\n"
 	       "for( i = 0; i < argc; i++ ) {\n" "  argv[i] = caml_list_nth(args,i);\n" "}\n", NIL);
 	Printv(df->code, dispatch, "\n", NIL);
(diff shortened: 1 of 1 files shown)


Source/Modules/ocaml.cxx
```diff
@@ -733,7 +733,18 @@ class OCAML:public Language {
 	       "argv = (CAML_VALUE *)malloc( argc * sizeof( CAML_VALUE ) );\n"
 	       "for( i = 0; i < argc; i++ ) {\n" "  argv[i] = caml_list_nth(args,i);\n" "}\n", NIL);
 	Printv(df->code, dispatch, "\n", NIL);
-	Printf(df->code, "caml_failwith(\"No matching function for overloaded '%s'\");\n", iname);
+	Node *sibl = n;
+	while (Getattr(sibl, "sym:previousSibling"))
+	  sibl = Getattr(sibl, "sym:previousSibling");
+	String *protoTypes = NewString("");
```
wsfulton: Can you avoid caml case please. The convention is all lower case variable names.

wsfulton: I see you copied it from another module, so not to worry. Those are wrong too :(

PR #1427, opened 2019-01-22
[OCaml] Fix %allowexception

ZackerySpytz:
OCaml's variableWrapper() wasn't calling emit_action_code() for
in/out typemaps, which meant that %allowexception was being ignored.

In addition, remove all comments in the typemaps in Lib/ocaml. In the
case of the allowexcept test, one of the typemap comments caused
compilation to fail because it became nested within another comment
in an %exception block.

Re-enable the allowexcept test.
Add allowexcept_runme.ml.


ZackerySpytz:
The Travis failure is unrelated to this patch.

PR #1429, opened 2019-01-24
[OCaml] Eliminate use of -Wno-write-strings

ZackerySpytz:
Don't convert string literals to char * in the strings_test example.

In constantWrapper(), use SwigType_str() instead of SwigType_lstr()
in order to keep const qualifiers.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,11 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.0 (in progress)
 ===========================
 
+2019-01-28: ZackerySpytz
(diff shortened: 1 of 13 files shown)


wsfulton:
Seems reasonable to me, SWIG 4.0.0 would  a good time to increase the minimum version. Please make sure this is clearly documented.

ZackerySpytz:
Thank you, @wsfulton. I've pushed a commit that updates the OCaml module's documentation and adds an entry in CHANGES.current (which is flagged as a potential incompatibility). I based the CHANGES.current entry on the one that describes the removal of Python 3.0 and 3.1 support.

By the way, the CI failures for the other commits in this pull request were due to network issues.

PR #1439, opened 2019-01-30
[OCaml] Fix reference typemaps for std::string

ZackerySpytz:
Fix warnings in the extend_template_method, li_std_string, and
template_methods tests. std::string was missing a typecheck typemap.

Add extend_template_method_runme.ml, li_std_string_runme.ml, and
template_methods_runme.ml.


wsfulton:
CI shows Examples/ocaml/strings_test failing, could you take a look at that please?

ZackerySpytz:
@wsfulton I'm sorry. Recently, I've just been running `make check-ocaml-test-suite -j4` before creating an OCaml-related PR, but that's not enough.

Thank you for merging my other pull requests -- I greatly appreciate that.

wsfulton:
Hi @ZackerySpytz, there are merge conflicts and it looks like this pull request has broken the examples, so I can't mrege this.

I see OCaml is using the same typemaps for const and non-const string&. Usually we don't have these typemaps the same as the non-const string reference could be an input, an output or an input and output parameter, where the const string reference can only be an input parameter. Hence const string& and string passed by value are equivalent from the target language. What is the intention for non-const string from the OCaml language side? Normally we don't provide a typemap for non-const string reference and it is up to a user to apply the const string& typemaps to string& if the intention is to use it as an input only parameter, but the users needs to pro-actively decide. All the target languages I can think of have immutable strings, so using a string& doesn't map naturally to a higher-level language function parameter.

ZackerySpytz:
I'm aware that this PR is broken. I don't know how to fix it.

I'll remove the typemaps for string &.

ojwb:
> I'm aware that this PR is broken. I don't know how to fix it.

I think this probably didn't get looked at again because of this comment, but actually CI passed back then.

I merged master and resolved one conflict, I think correctly, and pushed to the PR branch but now CI fails.  I think I'm out of my depth here as I don't know ocaml at all.  It doesn't help that we've switch CI systems thanks to the travispocalypse.

@ZackerySpytz Could you take a look?  The work here seems useful, and it would be good to get it merged.  It passed CI before so hopefully it isn't far off working now.

ojwb:
> All the target languages I can think of have immutable strings, so using a string& doesn't map naturally to a higher-level language function parameter.

FWIW, for PHP `std::string&` does map naturally to a higher-level language function parameter - i.e. a PHP string with the parameter marked as "pass by reference" (which is what SWIG does by default).

I've merged latest master to the branch again (the conflict was just both master and branch adding typemaps to the file, but nothing that overlapped in function, only in location in the file).

Testing locally, I found and fixed a failure for `cpp20_spaceship_operator` due to not passing `$(CXXFLAGS)`.

The other local failure is:

```
checking ocaml testcase template_methods (with run test)
Fatal error: exception Invalid_argument("Incorrect number of arguments passed to 'Klass_KlassTMethodBool'")
make[1]: *** [Makefile:79: template_methods.cpptest] Error 2
```
`Examples/test-suite/ocaml/template_methods_runme.ml` is new from this PR, and comparing with the `_runme.java` (which is the only existing implementation) it seems that the problem is that we should be passing no parameters rather than `true` in two cases.  Fixing that the test passes.  (I'm a bit confused that apparently CI passed originally, but I guess I must have been wrong about that or something, but this testcase seems to be clearly wrong before my change.)

So fingers crossed this passes CI, in which case we can finally merge this!

ojwb:
Oh, except we don't have working CI for ocaml currently, sigh.

ojwb:
I cancelled the PR CI run as it wouldn't actually test ocaml.

I've cherry-picked the CXXFLAGS fix since that bug is present on git master too, then squash-merged the rest (the history is so drawn out that this seemed the best option).

PR #1440, opened 2019-01-30
Stop defaulting to Tcl

ojwb:
SWIG now requires a target language to be specified instead of
defaulting to wrapping for Tcl.  Specifying swig --help without
a target language now just shows the generic help.  The -nolang
option has been removed.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,12 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.0 (in progress)
 ===========================
 
+2019-01-31: olly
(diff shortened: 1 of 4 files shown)


ojwb:
Any objections to merging this?

(`-nolang` isn't documented anywhere and only seems to be provided to allow things like `swig -nolang -help`; I can't find any sign of anything using it - I checked https://codesearch.debian.net and some general web searches).

wsfulton:
Fine by me. I was thinking of doing the same as it happens!

ojwb:
Cherry-picked to master in ec72026285729f5b73f85505a1fd44263816d739.

PR #1442, opened 2019-01-31
[OCaml] Add support for the docstring option in the module directive

ZackerySpytz:
If given, the OCaml module will place the docstring at the very
beginning of the generated mli file, where it can be read by the
OCamldoc tool.

The implementation is based on the equivalent features in the Python
and Ruby modules.

It would be nice to have tests for this feature, but that would require
parsing generated mli files or [OCamldoc](https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html) output (see also #582, where
it was decided that tests were not needed in a similar case for the Ruby
module).

### DIFF
--- a/Doc/Manual/Contents.html
+++ b/Doc/Manual/Contents.html
@@ -1397,6 +1397,10 @@
 </ul>
 <li><a href="Ocaml.html#Ocaml_nn31">Exceptions</a>
 </ul>
+<li><a href="Ocaml.html#Ocaml_nn32">Documentation Features</a>
+<ul>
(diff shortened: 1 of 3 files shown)


ZackerySpytz:
The Travis build is close to the halfway point, but I see six unrelated failures already (the OCaml job passed).

wsfulton:
I think I restarted the failing Travis jobs a while back, so it is all green now. If you can think of any low-level approach to testing the documentation comments, that would. In Python we use the inspect module, inspect.getdoc to check the documentation comments. If there is anything equivalent in OCaml, consider using that.

PR #1452, opened 2019-02-06
Fix %apply for anonymous template instantiations

sethrj:
Any `%apply` directives inside a templated class were being ignored for classes instantiated anonymously with `%template() CLASS<T>`. This was inconsistent with how `%typemap` was treated.

(This fix is *independent* from the Fortran PR https://github.com/swig/swig/pull/1195; nothing there currently needs this fix.)


wsfulton:
Interesting little bug. Nice patch, thanks.

Issue #1460, opened 2019-02-09
%rename("$ignore", regexmatch$name=...) now results in parameter renaming


ojwb:
Trying out SWIG git master on Xapian's bindings, I found an unhelpful change in behaviour, which I would argue is a regression - the new behaviour is clearly not useful since it results in invalid Python code.

To demonstrate:

```
olly@gemse:~/git/swig$ cat dollarignore.i
%module dollarignore;
%rename("$ignore", regexmatch$name="_$") "";
%inline %{
int hide_this_one_(int x);
int wrap_me(int class_);
%}
olly@gemse:~/git/swig$ git checkout 3f5c17824c5f20023bac58f7ebfc8de8532d6881
HEAD is now at 3f5c1782 Improve handling parameters clashing with language keywords
olly@gemse:~/git/swig$ ./autogen.sh && make -s && ./preinst-swig -python dollarignore.i
+ test -d Tools/config
+ aclocal -I Tools/config
+ autoheader
+ automake --add-missing --copy --force-missing
+ autoconf
+ cd CCache
+ autoreconf
 cd .. && /bin/bash ./config.status Source/Makefile depfiles
config.status: creating Source/Makefile
config.status: executing depfiles commands
olly@gemse:~/git/swig$ tail -4 dollarignore.py
def wrap_me($ignore):
    return _dollarignore.wrap_me($ignore)


```

The pattern rename has been applied to a function parameter which has been renamed to `$ignore`, which is invalid Python code.

Prior to 3f5c17824c5f20023bac58f7ebfc8de8532d6881 things worked more helpfully - SWIG apparently realised that the special meaning of `$ignore` doesn't apply to a function parameter so it didn't try to rename it:

```
olly@gemse:~/git/swig$ git checkout 3f5c17824c5f20023bac58f7ebfc8de8532d6881^
Previous HEAD position was 3f5c1782 Improve handling parameters clashing with language keywords
HEAD is now at fa0c3fe5 Merge branch 'ZackerySpytz-OCaml-cpp_nodefault-constructor_rename-runtime-tests'
olly@gemse:~/git/swig$ ./autogen.sh && make -s && ./preinst-swig -python dollarignore.i
+ test -d Tools/config
+ aclocal -I Tools/config
+ autoheader
+ automake --add-missing --copy --force-missing
+ autoconf
+ cd CCache
+ autoreconf
 cd .. && /bin/bash ./config.status Source/Makefile depfiles
config.status: creating Source/Makefile
[...]

PR #1462, opened 2019-02-10
Fix for wrongly renaming parameters to "$ignore" in Python

vadz:
This fixes the regression from #1460 and should be really merged before 4.0.

TIA!


wsfulton:
Looks good to merge to me if the CI tests all pass.

Source/Swig/naming.c
```diff
@@ -1377,12 +1377,15 @@ void Swig_name_rename_add(String *prefix, String *name, SwigType *decl, Hash *ne
...
-/* Create a name applying rename/namewarn if needed */
-static String *apply_rename(String *newname, int fullname, String *prefix, String *name) {
+/* Create a name for the given node applying rename/namewarn if needed */
+static String *apply_rename(Node* n, String *newname, int fullname, String *prefix, String *name) {
   String *result = 0;
   if (newname && Len(newname)) {
     if (Strcmp(newname, "$ignore") == 0) {
-      result = Copy(newname);
+      /* $ignore doesn't apply to parameters and while it's rare to explicitly write %ignore directives for them they could be caught by a wildcard ignore using
+         regex match, just ignore the attempt to ignore them in this case */
```
ojwb: I think there's actually a more common situation that's this bug would have manifested in - e.g. you write `%ignore prefix;` intending to ignore a function `prefix()` but there's a parameter `prefix` on a different function.

vadz:
CI tests passed, so I've merged this.

PR #1463, opened 2019-02-10
Fixed typo in Perl5 docs.

hakonhagland:
Fixed typo in section 33.7.2 "Perl5 typemaps". There is no function set_setiv()
in the perl api, the correct name is sv_setiv().


wsfulton:
@tlby, presumably this is okay?

PR #1469, opened 2019-02-13
Add support for C++17 hexadecimal floating literals

ZackerySpytz:
Add support for C++17 hexadecimal floating literals.


ZackerySpytz:
This patch doesn't give an error for invalid hexadecimal floating literals with no digits (e.g. `0x.p+6`). Also, there's no error for `0x4.p+` or `4.e+`, but those cases aren't directly related to this patch. SWIG's handling of float literals is not as strict as it could be.

wsfulton:
Thanks, there are plenty of other areas where SWIG accepts incorrect code, a few corner cases like this are okay to slip by. We rely mostly on the C/C++ compiler catching incorrect code.

PR #1474, opened 2019-02-18
[OCaml] Add missing INPUT, OUTPUT, and INOUT typemaps for primitives

ZackerySpytz:
The typemaps are based on PHP's.


wsfulton:
Could you please change typemaps.i to only provide the INPUT, OUTPUT and INOUT typemaps. A user must explicitly %include typemaps.i to obtain these. This is what happens in the other target languages. The main OCaml typemaps could go into another file. Typically the main typemaps would go into ocaml.swg and the support functions (they look like the runtime?) currently in ocaml.swg would go into another file like ocamlrun.swg. Anyway, arrange the file structure however you like, except typemaps.i should not be included by default. Perhaps we should make this commit to add the support first and then rename the files, but there is a merge conflict to deal with first.

ZackerySpytz:
GitHub annoyingly shows this PR as causing a huge amount of churn (I see `+945 −889`). This PR actually has only ~100 insertions and less than 100 deletions, and most changes were in the first commit. This might have something to do with the fact that typemaps.i was renamed and then recreated.

ZackerySpytz:
Hi, @wsfulton. `Lib/ocaml/typemaps.i` now only contains INPUT, OUTPUT, and INOUT typemaps, and I have also addressed the rest of your comment. I plan to merge this PR at some point in the next few days.

PR #1477, opened 2019-02-19
[OCaml] Add %feature(thread) to release the OCaml master lock

ZackerySpytz:
The caml_release_runtime_system() and caml_acquire_runtime_system()
functions are provided by the OCaml C API to release and acquire the
master lock.  This can be helpful for multi-threaded applications.

### DIFF
--- a/Doc/Manual/Contents.html
+++ b/Doc/Manual/Contents.html
@@ -1824,6 +1824,7 @@
 <ul>
 <li><a href="Ocaml.html#Ocaml_nn33">Module docstring</a>
 </ul>
+<li><a href="Ocaml.html#Ocaml_nn34">%feature("thread")</a>
 </ul>
 </div>
 <!-- INDEX -->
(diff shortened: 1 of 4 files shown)


ojwb:
I know little to nothing about ocaml, but this seems plausible to me.  There really should be a testcase exercising the new feature though, if possible one which shows that the master lock does get released and then reclaimed.

wsfulton:
@ZackerySpytz this looks okay, but a simplified version of the Python `-threads` option and associated features. Most of the documentation is at http://swig.org/Doc4.0/Python.html#Python_multithreaded. I would at least try and get the feature names the same as in Python before merging.

ojwb:
> @ZackerySpytz this looks okay, but a simplified version of the Python `-threads` option and associated features. Most of the documentation is at http://swig.org/Doc4.0/Python.html#Python_multithreaded. I would at least try and get the feature names the same as in Python before merging.

That seems sensible.

@ZackerySpytz are you still interested in getting this merged?

PR #1480, opened 2019-02-21
Fix some rejections of valid floating-point literals

ZackerySpytz:
Some valid floating-point literals were giving
"Error: Syntax error in input(1)".


wsfulton:
Thanks, that's a useful fix.

PR #1481, opened 2019-02-21
Updated java std_map to support Obj derivatives by importing original for use in class

zphensley42:
I am not sure what branch this should go to, so I just made changes from master -> master. Our projects are using a more modified version than this, but I felt this was a necessary change for mainline.

If your native wrappers provide their own 'Object' class, the lack of an import here ends up using the same 'Object' class in the generated package, thus causing compilation errors as it attempts to cast your generated 'Object' to other java.lang.Object derivatives, like String.


wsfulton:
I think I'd prefer a global replace of Object with java.lang.Object instead. I don't like to 'use up' the javaimports typemap so that it is free for users to use.

zphensley42:
Fair. I'll update my pr later today or tomorrow.

PR #1483, opened 2019-02-22
Fix the handling of director classes with final methods

### LINKED ISSUE #564: Directors try to handle methods marked as final in a subclass.
I used `%feature("director") MyClass;`. `MyClass` has a virtual method `virtual void MyClass::foo();` that is marked `final` in a subclass: `void Subclass::foo() final;`. SWIG still attempts to make `Subclass::foo()` a director method. This can be fixed by using `%feature("nodirector") MyClass::foo;`, but that is undesirable, as other subclasses of `MyClass` in my target language might want to implement `foo()`.

ZackerySpytz:
Generated SwigDirector_* classes were attempting to override
methods marked as final.

In addition, give a warning if the destructor of a director class is
final.

Closes #564.

### DIFF
--- a/Doc/Manual/Warnings.html
+++ b/Doc/Manual/Warnings.html
@@ -535,6 +535,7 @@ example.i(4) : Syntax error in input(1).
 <li>522. Use of an illegal constructor name '<em>name</em>' in %extend is deprecated, the constructor name should be '<em>name</em>'.
 <li>523. Use of an illegal destructor name '<em>name</em>' in %extend is deprecated, the destructor name should be '<em>name</em>'.
 <li>524. Experimental target language. Target language <em>language</em> specified by <em>lang</em> is an experimental language. Please read about SWIG experimental languages, <em>htmllink</em>.
+<li>525. The director base class '<em>name</em>' or the destructor of director base class '<em>name</em>' is marked as final.
 </ul>
 
 <H3><a name="Warnings_doxygen">18.9.6 Doxygen comments (560-599)</a></H3>
--- a/Examples/test-suite/cpp11_final_directors.i
+++ b/Examples/test-suite/cpp11_final_directors.i
@@ -1,0 +1,18 @@
+%module(directors="1") cpp11_final_directors
+
+%warnfilter(SWIGWARN_PARSE_KEYWORD) final;
+
+%director Derived;
+
+%inline %{
+struct Base {
+  virtual void basemeth() final {}
+  virtual ~Base() {}
+};
+
+struct Derived : Base {
+  virtual int derivedmeth() final { return 1; }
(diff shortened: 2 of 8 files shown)


Examples/test-suite/cpp11_final_directors.i
```diff
@@ -0,0 +1,18 @@
+%module(directors="1") cpp11_final_directors
+
+%warnfilter(SWIGWARN_PARSE_KEYWORD) final;
```
ojwb: This should only be needed if something gets wrapped as the identifier `final` in the target language (e.g. if there was a method, class or variable called `final`) but that's not the case here.  That's because `final` is a keyword in some target languages (e.g. Java).

I tried removing this line and your new testcase still seems to pass.

ZackerySpytz: Thank you for your comment, @ojwb. You are absolutely right.

Also, this PR needs to add cpp11_final_directors to Examples/test-suite/common.mk.

Doc/Manual/Warnings.html
```diff
@@ -535,6 +535,7 @@ <H3><a name="Warnings_nn14">18.9.5 Code generation (500-559)</a></H3>
 <li>522. Use of an illegal constructor name '<em>name</em>' in %extend is deprecated, the constructor name should be '<em>name</em>'.
 <li>523. Use of an illegal destructor name '<em>name</em>' in %extend is deprecated, the destructor name should be '<em>name</em>'.
 <li>524. Experimental target language. Target language <em>language</em> specified by <em>lang</em> is an experimental language. Please read about SWIG experimental languages, <em>htmllink</em>.
+<li>525. The director base class '<em>name</em>' or the destructor of director base class '<em>name</em>' is marked as final.
```
wsfulton: The documentation of this warning is different to the actual warning issued. Perhaps you intended to issue this warning whenever a final method was parsed and then removed the warning?

It looks like a class with a final destructor and is marked with feature director will not be a director class and then you get the warning. I think the coding of the implementation in the patch is best wrt expected warnings and behaviour. I'd like to see a couple of lines of documentation in the Directors section stating what is probably pretty obvious. That is virtual final methods are ignored without warning. Any class marked with director issues the warning and is equivalent to marking it with nodirector.

wsfulton:
I got a bit carried away and finished this patch, fixing the seg faults in C#. The combined changes are in bc0645ce2bc5afb19c62e4c6246397b9c2080d37.

Doc/Manual/Warnings.html
ZackerySpytz: I had intended to use this warning when support for final classes is added to SWIG. Marking a class' dtor as final is effectively the same as marking the whole class as final.

PR #1484, opened 2019-02-22
Namespace quirks

fschlimb:
continuation of #1115 (rebasing my branch seemed to confuse CI)

### DIFF
--- a/Source/CParse/templ.c
+++ b/Source/CParse/templ.c
@@ -329,6 +329,7 @@ int Swig_cparse_template_expand(Node *n, String *rname, ParmList *tparms, Symtab
 
     if (tp) {
       Symtab *tsdecl = Getattr(n, "sym:symtab");
+      String *tsname = Getattr(n, "sym:name");
       while (p && tp) {
 	String *name, *value, *valuestr, *tmp, *tmpr;
 	int sz, i;
@@ -376,11 +377,9 @@ int Swig_cparse_template_expand(Node *n, String *rname, ParmList *tparms, Symtab
 	      We will not replace template args if a type/class exists with the same
 	      name which is not a template.
 	    */
-	    Node * __n = Swig_symbol_clookup(s, 0);
-	    String * __nn = __n ? Getattr(__n, "sym:name") : 0;
-	    String * nn = __nn ? Getattr(n, "sym:name") : 0;
-	    int __eq = __nn && nn ? Strcmp(nn, __nn) : 1;
-	    if(__eq!=0 || Getattr(__n, "templatetype")) {
+	    Node * tynode = Swig_symbol_clookup(s, 0);
+	    String *tyname  = tynode ? Getattr(tynode, "sym:name") : 0;
+	    if (!tyname || !tsname || !Equal(tyname, tsname) || Getattr(tynode, "templatetype")) {
 	      SwigType_typename_replace(s, name, dvalue);
 	      SwigType_typename_replace(s, tbase, iname);
 	    }
--- a/Source/CParse/templ.c
+++ b/Source/CParse/templ.c
@@ -50,6 +50,7 @@ void Swig_cparse_debug_templates(int x) {
  * patching typenames and other aspects of the node according to a list of
  * template parameters
  * ----------------------------------------------------------------------------- */
+
 static void cparse_template_expand(Node *templnode, Node *n, String *tname, String *rname, String *templateargs, List *patchlist, List *typelist, List *cpatchlist) {
   static int expanded = 0;
   String *nodeType;
--- a/Source/CParse/templ.c
+++ b/Source/CParse/templ.c
@@ -260,7 +260,7 @@ int Swig_cparse_template_expand(Node *n, String *rname, ParmList *tparms, Symtab
   String *tbase;
   String *scopename = 0;
   String *tmp = tscope ? Getattr(tscope, "name") : 0;
-  if( tmp ) {
+  if (tmp) {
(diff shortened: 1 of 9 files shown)


fschlimb:
This is the continuation of #1115; I opened a new PR because rebasing my branch seemed to confuse CI.

I simplified the code and also reduced to changes related to the template/namespace issue. Most of the removed pieces from #1115 went into #1488.

I added a more extensive test-case also covering the case from #810.  I will check with our project to see if the new SWIG code-base together with this makes #810 unnecessary. I'll leave #810 open until I clarified. 

@wsfulton @ojwb Any other comments/thoughts/requests?

Source/CParse/templ.c
```diff
@@ -371,11 +370,20 @@ int Swig_cparse_template_expand(Node *n, String *rname, ParmList *tparms, Symtab
...
-	    SwigType_typename_replace(s, name, dvalue);
-	    SwigType_typename_replace(s, tbase, iname);
-	    /*      Printf(stdout,"'%s'\n", s); */
+	    /*
+	      The approach of 'trivially' replacing template arguments is kind of fragile.
+	      In particular if types with similar name in different namespaces appear.
+	      We will not replace template args if a type/class exists with the same
+	      name which is not a template.
+	    */
+	    Node * __n = Swig_symbol_clookup(s, 0);
```
ojwb: Identifiers containing a double underscore anywhere are reserved (https://en.cppreference.com/w/cpp/language/identifiers) so these variables really need renaming to not use reserved names.

Source/CParse/templ.c
```diff
@@ -371,11 +370,20 @@ int Swig_cparse_template_expand(Node *n, String *rname, ParmList *tparms, Symtab
...
+	    String * __nn = __n ? Getattr(__n, "sym:name") : 0;
+	    String * nn = __nn ? Getattr(n, "sym:name") : 0;
+	    int __eq = __nn && nn ? Strcmp(nn, __nn) : 1;
```
ojwb: This is confusing code.  The variable is called `__eq` which suggests it's true if equal, but (when both the arguments are non-NULL) it's set to the result of `Strcmp()`, which is `0` if equal and non-zero if non-equal.  So if the strings are equal, it seems `__eq` is false!?!

Better to compare the strings with `Equal(a, b)` and make the sense of the variable match what its name suggests.

Source/CParse/templ.c
```diff
@@ -371,11 +370,20 @@ int Swig_cparse_template_expand(Node *n, String *rname, ParmList *tparms, Symtab
...
+	    if(__eq!=0 || Getattr(__n, "templatetype")) {
```
ojwb: The whitespace here doesn't match the SWIG code (should have a space after `if` and spaces around `!=`).

Source/CParse/templ.c
fschlimb: done

Source/CParse/templ.c
fschlimb: done

Source/CParse/templ.c
fschlimb: done

fschlimb:
@ojwb thanks for your review. I made the suggested changes.

ojwb:
Thanks.  Looks OK to me now, but I think it would be good for @wsfulton to review too.

fschlimb:
@wsfulton, any chane this will ever make it to master?

ojwb:
I've pushed a minor whitespace tweak to the branch (the other pre-function comment blocks have a line after them before the function, at least in this file, so removing this one seems undesirable) which should also mean we'll get up to date CI results, but annoyingly a whole slew of builds seem to have failed for some sort of unrelated CI system reason.  

I'm hesitant to merge this without a thumbs up from @wsfulton as I'm not quite sure I've got my head around exactly what the problem is that this addresses.  From this comment I gather it's to do with doing a textual search and replace to expand template arguments but I haven't fully understood what goes wrong:

```
              The approach of 'trivially' replacing template arguments is kind of fragile.
	      In particular if types with similar name in different namespaces appear.
	      We will not replace template args if a type/class exists with the same
	      name which is not a template.
```

Maybe we can make the substitution a bit smarter for a better fix.

fschlimb:
I guess the fundamental problem is that SWIG tries to implement full C++ compiler frontend and uses such "trivial" replacements at various places (not only here). I am actually surprised how well this works for so many codes. Chapeau! Template and template argument substitution is a very complex issue with many subtleties. Trying to solve this would probably require a fundamentally different approach.
As far as I remember this fix does not introduce any regression (in the test suite) but makes SWIG support a larger set of possible namespace/template constructs.

wsfulton:
I've added this to the swig-4.1 milestone for review before we release the next version.

wsfulton:
fa00622614d35716b429b12aa9dc1bde8e4b586d contains the majority of this patch, that is, the changes in templ.c and the test case. The only observable fix I could see from the testcase was for Bar:
```c++
              namespace A { class Bar {}; }
              namespace B { template<typename T, typename U> class Bar : public A::Bar {}; }
 ```
I couldn't see any observable difference for the changes in typepass.cxx in either the test-suite nor the additional test case here; the change doesn't look unreasonable, but I'd like an example of the specialization that requires the change in order to accept it.

fschlimb:
Are you saying that the test here works correctly with fa00622?

wsfulton:
Yes, fa00622 includes the testcase from this pull request.

Issue #655, opened 2016-04-12
failing to identify class-method in using directive


fschlimb:
SWIG doesn't find a method for 'using' in class which is used via 'using' itself.
The following simple code gives the error

``````
swig -c++ -python using_member.i
using_member.i:16: Warning 315: Nothing known about 'A::get'.```
``````

```
%module using_member;

namespace interface1
{
    class A
    {
    public:
        int get(int);
    };
}
using interface1::A;

class B : public A
{
 public:
    using A::get;
    int get(double);
};
```

Note that `using A::interface1::get;` works fine. Unfortunately fixing it this way in the C++ code is not an option.

Also note that SWIG correctly identifies `A` when it's declared as the base-class, but not when searching for the method in the `using` directive.

Because of the latter this looks like it could be a simple fix. I'd be willing to invest some time in fixing it myself but would need some hints as I am not familiar with the SWIG code.

Any idea and/or help?

**Comment by wsfulton** [comment]
This will be a problem in the symbol table management and lookup (symbol.c). There is probably a Swig_symbol_clookup call which fails and causes the warning. I suspect there is a missing recursive using lookup (ie not following all using nodes).

For learning about SWIG source and its quirks, take a look at http://swig.org/Doc3.0/Extending.html and the development docs, especially the debugging section in Doc/Devel/internals.html#7. I notice that when using the -debug-csymbols option, the output is:

```
example.i:19: Warning 315: Nothing known about 'A::get'.
CSYMBOLS start  =======================================
===================================================
 -
  swig (namespace)
  interface1 (namespace)
  A (using)
  std (namespace)
  B (class)
===================================================
std::initializer_list -
  T (templateparm)
===================================================
swig -
  SwigPtr_PyObject (class)
  SwigVar_PyObject (class)
===================================================
interface1 -
  A (class)
===================================================
std -
  initializer_list<(T)> (template)
  initializer_list (template)
===================================================
B -
  get (using)
```
So looks like ```B::get``` has been added into the symbol table, so the problem is likely to be in the lookup.

**Comment by wsfulton** [comment]
BTW, there are other 'using' bugs that you might stumble across and be aware of while researching this. If you remove all the namespaces from your code, then consider this:
```c++
class B : public A
{
 public:
    using A::get; // line 4
    int get(double, double); // line 5
};
```
If you swap lines 4 and 5 around, the generated code for overloading doesn't work (for one of the swapped combinations, I don't recall which).

tamuratak:
[...]

PR #1488, opened 2019-02-26
better handling of using declarations

fschlimb:
Fix for #655

### DIFF
--- a/Source/Swig/symbol.c
+++ b/Source/Swig/symbol.c
@@ -384,12 +384,15 @@ String *Swig_symbol_qualified_language_scopename(Symtab *n) {
 
 /* -----------------------------------------------------------------------------
  * Swig_symbol_add_using()
+ *
+ * Adds a node to the C symbol table for a using declaration.
+ * Used for using-declarations within classes/structs.
  * ----------------------------------------------------------------------------- */
 
 void Swig_symbol_add_using(String * name, String * uname, Node * n) {
   Hash *h;
-  h = Swig_symbol_clookup(uname,0);
-  if (h && checkAttribute(h, "kind", "class")) {
+  h = Swig_symbol_clookup(uname, 0);
+  if (h && (checkAttribute(h, "kind", "class") || checkAttribute(h, "kind", "struct"))) {
     String *qcurrent = Swig_symbol_qualifiedscopename(0);
     if (qcurrent) {
 	Append(qcurrent, "::");
--- a/Source/Swig/symbol.c
+++ b/Source/Swig/symbol.c
@@ -383,6 +383,26 @@ String *Swig_symbol_qualified_language_scopename(Symtab *n) {
 }
 
 /* -----------------------------------------------------------------------------
+ * Swig_symbol_add_using()
+ * ----------------------------------------------------------------------------- */
+
+void Swig_symbol_add_using(String * name, String * uname, Node * n) {
+  Hash *h;
+  h = Swig_symbol_clookup(uname,0);
+  if (h && checkAttribute(h, "kind", "class")) {
+    String *qcurrent = Swig_symbol_qualifiedscopename(0);
+    if (qcurrent) {
+	Append(qcurrent, "::");
+	Append(qcurrent, name);
+    } else {
+	qcurrent = NewString(name);
+    }
+    Setattr(symtabs, qcurrent, n);
+    Delete(qcurrent);
+  }
+}
+
+/* -----------------------------------------------------------------------------
  * Swig_symbol_newscope()
  *
  * Create a new scope.  Returns the newly created scope.
@@ -1074,7 +1094,19 @@ static Node *symbol_lookup_qualified(const_String_or_char_ptr name, Symtab *symt
 	  Delete(qalloc);
 	return st;
       }
-      n = symbol_lookup(name, st, checkfunc);
+      if (checkAttribute(st, "nodeType", "using")) {
+	String *uname = Getattr(st, "uname");
+	if (uname) {
+	  st = Getattr(symtabs, uname);
+	  if (st) {
+	    n = symbol_lookup(name, st, checkfunc);
+	  } else {
+	    fprintf(stderr, "Error: Found corrupt 'using' node\n");
+	  }
+	}
+      } else if (Getattr(st, "csymtab")) {
+	n = symbol_lookup(name, st, checkfunc);
+      }
     }
     if (qalloc)
       Delete(qalloc);
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -539,6 +539,7 @@ CPP_TEST_CASES += \
 	using_directive_and_declaration_forward \
 	using_extend \
 	using_inherit \
+	using_member \
 	using_namespace \
 	using_namespace_loop \
 	using_pointers \
--- a/Examples/test-suite/python/using_member_runme.py
+++ b/Examples/test-suite/python/using_member_runme.py
@@ -1,0 +1,10 @@
+from using_member import *
+
+b = B()
+assert b.get(int(1)) == 10
+assert b.get(float(1)) == 20
+
+bb = BB()
+assert bb.greater(int(1)) == 0
+assert bb.greater(float(1)) == 1
+assert bb.great(True) == 2
--- a/Examples/test-suite/using_member.i
+++ b/Examples/test-suite/using_member.i
@@ -1,0 +1,52 @@
+%module using_member;
+
+%rename(greater) one::two::three::interface1::AA::great(int);
+%rename(greater) one::two::three::interface1::AA::great(float);
+
+%inline %{
+namespace interface1
+{
+    class A
+    {
+    public:
+        int get(int) {return 10;}
+    };
+}
+using interface1::A;
+
+class B : public A
+{
+ public:
+    using A::get;
+    int get(double) {return 20;}
+};
+
+
+namespace one {
+    namespace two {
+        namespace three {
+            namespace interface1
+            {
+                class AA
+                {
+                public:
+                    int great(int) {return 0;}
+                    int great(float) {return 1;}
+                };
+            }
+            using interface1::AA;
+        }
+    }
+    namespace twotwo {
+        namespace threetwo {
+            class BB : public two::three::AA
+            {
+            public:
+                using two::three::AA::great;
+                int great(bool) {return 2;}
+                int jj() {return 3;};
+            };
+        }
+    }
+}
+%}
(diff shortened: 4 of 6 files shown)


fschlimb:
This fixes #655 and is a partial continuation of #1115.

It includes test with a runtime-check (python) for correctness.

@wsfulton @ojwb Any other comments/thoughts/requests?

Source/Swig/symbol.c
```diff
@@ -382,6 +382,26 @@ String *Swig_symbol_qualified_language_scopename(Symtab *n) {
   return result;
 }
 
+/* -----------------------------------------------------------------------------
+ * Swig_symbol_add_using()
+ * ----------------------------------------------------------------------------- */
+
+void Swig_symbol_add_using(String * name, String * uname, Node * n) {
+  Hash *h;
+  h = Swig_symbol_clookup(uname,0);
```
ojwb: For consistency I'd put a space after the `,` here...

Source/Swig/symbol.c
```diff
@@ -382,6 +382,26 @@ String *Swig_symbol_qualified_language_scopename(Symtab *n) {
...
+    if (qcurrent) {
+	Append(qcurrent, "::");
+	Append(qcurrent, name);
+    } else {
+	qcurrent = NewString(name);
+    }
+    Setattr(symtabs, qcurrent, n);
+    Delete(qcurrent);
+  }
+}
```
ojwb: I think this new function would benefit greatly from a documentation comment explaining what it does and why - I'm not entirely sure I understand what's going on here.  The description of `Swig_symbol_clookup()` is:

```
 * Swig_symbol_clookup()
 *
 * Look up a symbol in the symbol table.   This uses the C name, not scripting
 * names.   Note: If we come across a using declaration, we follow it to
 * to get the real node. Any using directives are also followed (but this is
 * implemented in symbol_lookup()).
```

From the code it seems we get the "real node" for the "using" symbol, and if that's a class we qualify a different name and add a symtab entry for that, but I'm unclear why we're doing that.  I'm also left unsure why this is only done for classes (in particular it seems kind can be `"struct"` which in C++ is just a class with default visibility of `public`, but why is this not done for other kinds too?)

ojwb: If I do `s/class/struct/` on `using_member.i` then the test fails, but all the classes in there only have public members, so that ought to make no difference:

```
$ make using_member.cpptest
if [ -f ./using_member_runme.py ]; then echo "checking python testcase using_member (with run test)" ; else echo "checking python testcase using_member" ; fi
checking python testcase using_member (with run test)
make -f ../../../Examples/Makefile SRCDIR='./' CXXSRCS='' SWIG_LIB_DIR='../../../Lib' SWIGEXE='../../../swig' INCLUDES='-I../../../Examples/test-suite' SWIGOPT='-outcurrentdir -I../../../Examples/test-suite' NOLINK=true TARGET='using_member' INTERFACEDIR='../' INTERFACE='using_member.i' python_cpp
make[1]: Entering directory '/home/olly/git/swig/Examples/test-suite/python'
env SWIG_LIB=../../../Lib  ../../../swig -python  -c++ -outcurrentdir -I../../../Examples/test-suite -o using_member_wrap.cxx ./../using_member.i
./../using_member.i:20: Warning 315: Nothing known about 'A::get'.
./../using_member.i:45: Warning 315: Nothing known about 'two::three::AA::great'.
g++ -c -fpic -I. -I./ -I/usr/include  using_member_wrap.cxx   -I../../../Examples/test-suite  -I/usr/include/python2.7 -I/usr/lib/python2.7/config
g++ -shared  -I/usr/include     using_member_wrap.o    -o _using_member.so
make[1]: Leaving directory '/home/olly/git/swig/Examples/test-suite/python'
pycodestyle --ignore=E252,E30,E402,E501,E731,W291,W391 using_member.py
if [ -f ./using_member_runme.py ]; then env LD_LIBRARY_PATH=.:$LD_LIBRARY_PATH PYTHONPATH=.:.:$PYTHONPATH  python using_member_runme.py; fi
  File "using_member_runme.py", line 4, in <module>
    assert b.get(int(1)) == 10
AssertionError
make: *** [Makefile:122: using_member.cpptest] Error 1
```

fschlimb: The new nodes are specifically handling using declarations for class methods. Without this fix SWIG would simply not handle them and issue warnings like:
`Warning 315: Nothing known about 'A::get'`

Of course it should also handle struct methods, thanks for testing this out.

Do you think there is a better way to handle this?

fschlimb: I added a brief description and support for structs.

Source/Swig/symbol.c
fschlimb: done

fschlimb:
@wsfulton, any chane this will ever make it to master?

wsfulton:
Hey @ojwb, I take it there was no change in output in the partial-test-suite for say Python & Java with this patch and did you verify the various symbol table changes made sense?

@fschlimb Sorry for not responding earlier. I should have communicated at the time of this patch, it was too risky for patch releases 4.0.x and @ojwb is doing a great job at pushing forward with getting missing patches attended to for 4.1.0.

Couple of things overlooked:
- We don't use assert in the Python runtime tests.
- There is a checkAttribute missing checking for unions. Multiple calls to checkAttribute are best done via a temp variable to be more efficient.

ojwb:
> I take it there was no change in output in the partial-test-suite for say Python & Java with this patch and did you verify the various symbol table changes made sense?

There's no difference in the output, though that makes me realise the new testcase isn't actually getting run - I'll fix that.

SWIG's symbol tables make little sense to me anyway, so I think you'd better check that part...

ojwb:
> There's no difference in the output, though that makes me realise the new testcase isn't actually getting run - I'll fix that.

I'm wrong, it is being run (I was confused by the diff not showing the generated files being added, but that's because I ran the new version first and the generated files were still there when I ran the old).

ojwb:
> * We don't use assert in the Python runtime tests.

Maybe we're not meant to, but we do:

```
Examples/test-suite/python/cpp11_template_explicit_runme.py:assert isinstance(t1,int)
Examples/test-suite/python/cpp11_template_explicit_runme.py:assert isinstance(t2,cpp11_template_explicit.A)
Examples/test-suite/python/cpp11_template_explicit_runme.py:assert isinstance(t3,cpp11_template_explicit.TemperInt)
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_ret_c_shared_ptr(a) ==  1
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_ret_c_shared_ptr(b) == -1
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_ret_c_by_value(a)   ==  1
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_by_value(a)                  == 5
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_by_ref(a)                    == 6
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_by_pointer(a)                == 7
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_by_pointer_ref(a)            == 8
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_value(a)       == 9
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_ref(a)         == 10
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_pointer(a)     == 11
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_pointer_ref(a) == 12
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_by_pointer_with_null(a)                == -2
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_by_pointer_ref_with_null(a)            == -3
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_value_with_null(a)       == -4
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_ref_with_null(a)         == -5
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_pointer_with_null(a)     == -6
Examples/test-suite/python/li_boost_shared_ptr_director_runme.py:assert call_take_c_shared_ptr_by_pointer_ref_with_null(a) == -7
```

I'll push a fix for all of those.

ojwb:
Sigh, and fixing that reveals that `b.get(int(1))` gives 20 and not 10 like the testcase expects.

It seems that the `using` doesn't actually pull in the base class form of `get()` like you'd hope.

I need to stop for now.

Aside: I wonder if there is a way to "poison" assert to prevent this happening again.

ojwb:
> Sigh, and fixing that reveals that `b.get(int(1))` gives 20 and not 10 like the testcase expects.
> 
> It seems that the `using` doesn't actually pull in the base class form of `get()` like you'd hope.
> 
> I need to stop for now.

After confusing myself for a bit, this stopped happening.  I think I must have had a dirty tree.

> Aside: I wonder if there is a way to "poison" assert to prevent this happening again.

Testing, it seems `assert` does assert unless `-O` is passed, which we don't do when running the testsuite so if there is an easy way to enforce not using `assert` in the Python tests then that'd be nice, but it's not essential.

I've now pushed this fix.

ojwb:
> There is a checkAttribute missing checking for unions. Multiple calls to checkAttribute are best done via a temp variable to be more efficient.

You can't have a union as a baseclass of a class so I'm not seeing how you could use `using` on a member of a `union`.

Or are you suggesting `Swig_symbol_add_using()` should assert that it's not been pointed at a `union`?

wsfulton:
I've reverted this patch and put in a more correct fix in cb963a14401b05af36a58a2b76eba3f368354ffd as the symbol tables were not right. For example with `-debug-csymbols`, the output was:
```
...
===================================================
A -
===================================================
one::two::three::AA -
```
While that was an improvement compared to before the patch, it should include the class members that the using declaration introduces, like this:
```
...
===================================================
A -
  get (cdecl)
===================================================
one::two::three::AA -
  great (cdecl)
  great (cdecl)
```
Similarly for `-debug-symbols` for the renamed symbols.

With the new approach, the normal lookup mechanisms just work, because the class members can now be found via the using node, and there is no need for the changes to `symbol_lookup_qualified()`. The change adds the using declarations to the symbol tables in a near identical way that a typedef to a class does.

@fschlimb let me know if you get any surprises in your large code base.

fschlimb:
@wsfulton Thanks for finding a better fix, I never felt mine was particularly nice :)
Unfortunately I cannot verify your solution with the code which triggered the issues since the project no longer exists.

PR #1489, opened 2019-03-02
Include all template parameters for std_unordered_map Python macro

johnwason:
The current std_unordered_map Python macro currently only has two template arguments, K and T. It should have five: K, T, Hash, Pred, and Alloc. This patch corrects the missing template arguments.


wsfulton:
Thanks. I also add them in for the other C++11 STL classes.

PR #1494, opened 2019-03-06
[Perl] Fully qualify aliases in pm file

tlby:
I found a limitation trying to rename symbols into different Perl packages.
```
%module "Some::Lib"
...
%rename("%(regex:/^SOMELIB_(\\S+)$/Const::\\1/)s", %$isconstant) "";
```
With `-noproxy` the constants matched by this rename end up in `Some::Lib::Const::`, but with `-proxy` they end up in `Const::`.  This patch will make the wrapper fully qualify the aliases emitted in `-proxy` mode to make the behavior consistent.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 .gitattributes export-ignore
 .gitignore export-ignore
+CHANGES.current merge=union
--- a/Examples/test-suite/perl5/rename_qualified_runme.pl
+++ b/Examples/test-suite/perl5/rename_qualified_runme.pl
@@ -1,0 +1,36 @@
+use strict;
+use warnings;
+use Test::More tests => 12;
+BEGIN { use_ok('rename_qualified') }
+require_ok 'rename_qualified';
+
+sub try (&$$) {
+  my($test, $want, $name) = @_;
+  my $tb = Test::Builder->new();
+  my $got = eval { $test->() };
+  if($@) {
+    my $err = $@;
+    $err =~ s/^/    /mg;
+    $tb->diag($err);
+    return $tb->ok(0, $name);
+  }
+  return $tb->is_eq($got, $want, $name);
+}
+
+TODO: {
+  local $TODO = q(need to diagnose "Can't wrap XXX unless renamed to a valid identifier.");
+  my $s = eval { rename_qualified::New::Struct->new() };
+  try { $s->{InstanceVariable} } 111, 'InstanceVariable';
+  try { $s->InstanceMethod() } 222, 'InstanceMethod';
+  eval { $s->{NewInstanceVariable} = 1111 };
+  try { $s->{NewInstanceVariable} } 1111, 'InstanceVariable write';
+  try { rename_qualified::New::Struct::StaticMethod() } 333, 'StaticMethod';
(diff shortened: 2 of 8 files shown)


ojwb:
Looks OK to me, though it'd be good to have test coverage for the change (i.e. a test which fails if the change is reverted) to demonstrate it works and ensure it keeps working.

tlby:
Of course, I should lock this behavior down with some test coverage.  Thanks!

wsfulton:
@tlby, can you add in test coverage for the end of the week in order for this to be included in the final 4.0.0 release.

tlby:
That's a great idea.  I will try to get this ready shortly.

tlby:
The .gitattributes addition is supposed to help avoid merge conflicts by using an alternate merge strategy for just the CHANGES.current file.

.gitattributes
```diff
@@ -1,2 +1,3 @@
 .gitattributes export-ignore
 .gitignore export-ignore
+CHANGES.current merge=union
```
wsfulton: The CHANGES.current merge conflicts are a real nuisance so we do need a better solution, but this ought to go into a separate pull request. I don't quite know if it will work, the git documentation https://git-scm.com/docs/gitattributes says it will merge in a random order, but we want the dates to be done in correct date/time order.

Examples/test-suite/perl5/rename_qualified_runme.pl
```diff
@@ -0,0 +1,36 @@
...
+    my $err = $@;
+    $err =~ s/^/    /mg;
+    $tb->diag($err);
+    return $tb->ok(0, $name);
+  }
+  return $tb->is_eq($got, $want, $name);
+}
+
+TODO: {
+  local $TODO = q(need to diagnose "Can't wrap XXX unless renamed to a valid identifier.");
```
wsfulton: This warning message happens because you can't have colons in a symbol name. The %rename should then be ignored. If it is being accepted, it is a bug.

I think you are trying to use %rename to put symbols into an additional Perl package namespace. This is an abuse of %rename and should not be allowed. I don't know what the package support is like in SWIG for Perl, but should probably be used instead. We don't have a mechanism in SWIG for moving symbols into a different target namespace equivalent (like a Perl package). That is a whole new dimension of support to be added if deemed useful (I don't think it is). Perl should support %nspace first.

Can you please modify the patch to test the fix without this unsupported use of %rename?

.gitattributes
tlby: Sure thing, I will cut this apart and research if we can implement a custom merge strategy that would preserve ordering of the timestamps in the entries.

Examples/test-suite/perl5/rename_qualified_runme.pl
tlby: You're right, I am trying to use %rename to move related groups of functionality into separate packages.  It sounds like Perl handling needs to be more aggressive about checking valid identifiers than it currently is to prevent this abuse.  As evidenced by the one case I found working, constants need some additional validation.  I'll work out what needs to be done there.

tlby:
Sounds like the premise of this patch, to unify the difference between `-proxy` and `-noproxy` in this edge case is flawed.  Perl handling should be stricter to prevent this edge case from being allowed in the first place.

PR #1495, opened 2019-03-07
Add std::set<> typemaps for C#

vadz:
Create new Lib/csharp/std_set.i based on the existing std_map.i and run
li_std_set unit test for C# as well.

Notice that the set operations defined by the base ISet<> interface are
not implemented yet.

---

This might be too late for 4.0, but OTOH it's a pretty straightforward addition closing a surprising lack of these typemaps for C# (they exist for Java and Python, for example, and usually C# provides everything Java does).

Note that `ISet<>` is only available in .NET 3.5 or later, but I don't think this is a problem as you can just avoid using these typemaps if compatibility with ancient .NET versions is important.

### DIFF
--- a/Lib/csharp/std_set.i
+++ b/Lib/csharp/std_set.i
@@ -1,9 +1,13 @@
 /* -----------------------------------------------------------------------------
- * std_map.i
+ * std_set.i
  *
- * SWIG typemaps for std::set<T>
+ * SWIG typemaps for std::set<T>.
  *
- * The C# wrapper is made to look and feel like a C# System.Collections.Generic.HashSet<>.
+ * Note that ISet<> used here requires .NET 4 or later.
+ *
+ * The C# wrapper implements ISet<> interface and shares performance
+ * characteristics of C# System.Collections.Generic.SortedSet<> class, but
+ * doesn't provide quite all of its methods.
  * ----------------------------------------------------------------------------- */
 
 %{
--- a/Lib/csharp/std_set.i
+++ b/Lib/csharp/std_set.i
@@ -71,16 +71,84 @@ class set {
     }
   }
 
-  public void ExceptWith(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
-  public void IntersectWith(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
-  public bool IsProperSubsetOf(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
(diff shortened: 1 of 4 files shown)


Lib/csharp/std_set.i
```diff
@@ -0,0 +1,239 @@
+/* -----------------------------------------------------------------------------
+ * std_map.i
```
wsfulton:
```suggestion
 * std_set.i
```

Lib/csharp/std_set.i
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * SWIG typemaps for std::set<T>
+ *
+ * The C# wrapper is made to look and feel like a C# System.Collections.Generic.HashSet<>.
```
wsfulton: Wouldn't `System.Collections.Generic.SortedSet<>` be a better model to base this off?

wsfulton: Add in:
```
 * Requires .NET 4 minimum.
```

Lib/csharp/std_set.i
```diff
@@ -0,0 +1,239 @@
...
+  public bool IsProperSupersetOf(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
+  public bool IsSubsetOf(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
+  public bool IsSupersetOf(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
+  public bool Overlaps(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
+  public bool SetEquals(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
+  public void SymmetricExceptWith(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
+  public void UnionWith(global::System.Collections.Generic.IEnumerable<$typemap(cstype, T)> other) { throw new global::System.Exception("TODO"); }
+
```
wsfulton: Disappointing that these are not implemented, but I'll accept the patch with this hole. Someone else will hopefully sort it out.

vadz:
I've provided the naive implementations of the set-theoretic methods and updated the header comment (thanks!).

I also agree that `SortedSet<>` would be a better counterpart to `std::set<>` in C# (the only reason I didn't realize it is because I don't know C#...), but it has even more methods that I don't need currently, so I didn't implement them. What is there is already useful enough to be included in SWIG IMO and new methods can always be added later.

PR #1503, opened 2019-03-19
[OCaml] Fix race conds when running the examples with parallel make

ZackerySpytz:
Race conditions could occur when running the examples with parallel make.

### DIFF
--- a/Examples/ocaml/shapes/Makefile
+++ b/Examples/ocaml/shapes/Makefile
@@ -20,7 +20,7 @@ static:
 	PROGFILE='$(PROGFILE)' OBJS='$(OBJS)' \
 	ocaml_static_cpp
 
-toplevel:
+toplevel: static
 	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' SRCS='$(SRCS)' \
 	SWIG_LIB_DIR='$(SWIG_LIB_DIR)' SWIGEXE='$(SWIGEXE)' \
 	SWIGOPT='$(SWIGOPT)' TARGET='$(TARGET)' INTERFACE='$(INTERFACE)' \
--- a/Examples/ocaml/string_from_ptr/Makefile
+++ b/Examples/ocaml/string_from_ptr/Makefile
@@ -19,7 +19,7 @@ static:
 	PROGFILE='$(PROGFILE)' OBJS='$(OBJS)' \
 	ocaml_static
 
-toplevel:
+toplevel: static
 	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' SRCS='$(SRCS)' \
 	SWIG_LIB_DIR='$(SWIG_LIB_DIR)' SWIGEXE='$(SWIGEXE)' \
 	TARGET='$(TARGET)' INTERFACE='$(INTERFACE)' \
--- a/Examples/ocaml/strings_test/Makefile
+++ b/Examples/ocaml/strings_test/Makefile
@@ -23,7 +23,7 @@ dynamic:
 	PROGFILE='$(PROGFILE)' TARGET='$(TARGET)' INTERFACE='$(INTERFACE)' \
 	ocaml_static_cpp
 
-toplevel:
+toplevel: static
 	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' SRCS='$(SRCS)' \
 	SWIG_LIB_DIR='$(SWIG_LIB_DIR)' SWIGEXE='$(SWIGEXE)' \
 	PROGFILE='$(PROGFILE)' TARGET='$(TARGET)' INTERFACE='$(INTERFACE)' \


wsfulton:
This doesn't work for me. I get output like this:
```
~/swig/github/swig $ make check-ocaml-examples -j8
checking Examples/ocaml/argout_ref
checking Examples/ocaml/callback
checking Examples/ocaml/contract
checking Examples/ocaml/class
checking Examples/ocaml/scoped_enum
checking Examples/ocaml/simple
checking Examples/ocaml/shapes
checking Examples/ocaml/std_vector
File "swigp4.ml", line 1:
File "swigp4.ml", line 1:
Error: I/O error: /tmp/ocamlpp330f2c989090aeda81b97f22c56de960: File exists
Error: I/O error: /tmp/ocamlpp330f2c989090aeda81b97f22c56de960: File exists
File "swigp4.ml", line 1:
Error: I/O error: /tmp/ocamlpp330f2c989090aeda81b97f22c56de960: File exists
../../Makefile:876: recipe for target 'ocaml_static' failed
```
If I clean and run it again, I noticed that the tmp file is always the same name. It does not exist when starting the build and so seems to be required by each of the parallel builds. I went into the Examples/ocaml/class directory and touched this file and gave it root only permissions and built just this one example (so no parallel make). As the file can't be removed/created it failed of course. But it helped me diagnose where:
```
~/swig/github/swig/Examples/ocaml/class $ make clean check
make -f ../../Makefile SRCDIR='' TARGET='example' ocaml_clean
make[1]: Entering directory '/home/william/swig/github/swig/Examples/ocaml/class'
rm -f *_wrap* *~ .~* *.cmo *.cmi *.mli example.ml runme runme_top swig.ml swigp4.ml
rm -f core 
rm -f *.o *.so
make[1]: Leaving directory '/home/william/swig/github/swig/Examples/ocaml/class'
make -f ../../Makefile SRCDIR='' SRCS='example.c' \
SWIGOPT='' TARGET='example' INTERFACE='example.i' \
PROGFILE='runme.ml' OBJS='example.o' \
ocaml_static_cpp
make[1]: Entering directory '/home/william/swig/github/swig/Examples/ocaml/class'
(test -f swig.mli   || env SWIG_LIB=../../../Lib  ../../../swig -ocaml -co swig.mli 2>/dev/null) && (test -f swig.ml    || env SWIG_LIB=../../../Lib  ../../../swig -ocaml -co swig.ml 2>/dev/null) && (test -f swigp4.ml  || env SWIG_LIB=../../../Lib  ../../../swig -ocaml -co swigp4.ml 2>/dev/null) && (test -f swig.cmi   ||  ocamlc -c swig.mli) && (test -f swig.cmo   ||  ocamlc -c swig.ml) && (test -f swigp4.cmi ||  ocamlc -I ` camlp4 -where` -pp "camlp4o pa_extend.cmo q_MLast.cmo" -c swigp4.ml)
File "swigp4.ml", line 1:
Error: I/O error: /tmp/ocamlpp330f2c989090aeda81b97f22c56de960: File exists
../../Makefile:906: recipe for target 'ocaml_static_cpp' failed
```
And so narrowed it down even further to:
```
~/swig/github/swig/Examples/ocaml/class $ ocamlc -I ` camlp4 -where` -pp "camlp4o pa_extend.cmo q_MLast.cmo" -c swigp4.ml
File "swigp4.ml", line 1:
Error: I/O error: /tmp/ocamlpp330f2c989090aeda81b97f22c56de960: File exists
```
So it looks like it happens when this line in swigp4.ml is executed by ocamlc:
```ocaml
open Camlp4
```
My version of OCaml is 4.05.0. Looks like a flaw in ocamlc not being able to open this file at the same time as other instances of ocamlc trying to open the same filename or maybe it is something specific to Camlp4? Anyway it is a pretty major parallelism flaw. @ZackerySpytz, can you please find out what the problem is and perhaps you can think of a work-around? I'd like to get this into the swig-4.0.0 build as it breaks SWIG's parallel 'make check'.

ZackerySpytz:
Hi, @wsfulton. Thanks for the debugging.

With OCaml 4.07.0 and camlp4 4.07.0, this patch fixes the race conditions when creating toplevels.
I can't reproduce this other race condition that you've mentioned.

To be frank, I'm in over my head here. I don't have much experience with race conditions, and I don't think I will be able to fix this issue.

I certainly don't want this issue to hold back the release of SWIG 4.0.0. Perhaps it would be best to disable the OCaml module when running `make check`? The OCaml module is experimental, and I think it makes sense if experimental modules are excluded from `make check`, no?

wsfulton:
The kind of workaround I was thinking that you may be able to offer would be around the use of camlp4 or the point of swigp4.ml. I didn't know what it is, why it is there, although I just found something in SWIG's Ocaml.html. It is documented and optional for SWIG, but I've no idea on the importance of it. 

We've gone to a lot of trouble to make sure that 'make check' does actually work for experimental modules, so I don't want to go back on that. I'll try think of some other workaround in the makefiles, like forcing a serial make build for the ocaml examples. 

What would also be interesting is if this problem is only in 4.05.0.

wsfulton:
Also do the examples actually require swigp4.ml given it is documented as optional?

wsfulton:
I worked around the problem in b092055ca87a9e6f9d489d967d2ff27870687ce4 and 166b683e8918c655dcfa9c9c3d99727bbf6288ef

ZackerySpytz:
I'm sorry that I didn't respond to your latest comments on this issue, William. `swigp4.ml` is necessary for most of the examples. A lot of example code would have to be rewritten without it.

PR #1511, opened 2019-04-10
R member list trial simplify2019

richardbeare:


richardbeare:
This is a rebase of  PR #1268. It refactors code that was previously using string matching to generate get/set methods, rather than properties of the parse tree. The most obvious problem with the original approach was that methods ending with "set" or "get" were handled incorrectly.

The changes in output code from this patch are probably best compared with those of PR #1510 , which corrected some bugs in the existing methods (mainly to do with string lengths in comparisons and looking for `set` instead of `_set`

Samples of differences:
`Examples/test-suite/abstract_typedef2.i` includes a method named `set`. In the old approach this was  interpreted as an internally generated accessor function, resulting in `$<-` and `[[<-` being generated in the R code. This no longer happens.
Previously, as a result of the name and interpretation as an accessor function, the `set` method would not be available via the `$` method. Now it is.

Related problems occurred when methods were named "get" - `Examples/test-suite/director_string.i`. In the old approach these functions were incorrectly added to the vaccessors list. Now they aren't.

richardbeare:
@wsfulton Thought I'd better touch base again to see how I should handle this PR. I think it is important to get this one is as it fixed several issues. I did a bunch of work on style etc, but I'm starting to forget where I'm up to. Is the best course a rebase to somewhere specific so you can look in more detail?

wsfulton:
Hi @richardbeare. Looks like it is still work in progress with commented out code, can you tidy it up? Shouldn't there be a test showing the fix working? I mentioned this in #1268. Could you also please close this aforementioned pull request and any others that are superceded by this one.

Can you think of any backwards compatibility issues if we use this?

It would be nice to expand the docs on each improvement you make as the R docs are so bare minimum. Perhaps you can include a simple section on how member variables are used from R in the patch?

BTW, I don't understand the title of this pull request and how it is related to the fix. Perhaps it uses some R terminology I'm not familiar with.

richardbeare:
This patch doesn't add new functionality, but fixes a part that wasn't working properly.
 
I'll check the details, but here's my memory:

1. Title to do with names of functions and data structures in the R module. The patch relates to removing the use of method names to drive the wrapping logic. The R module code used the term "member list" to define the list of class methods that were being wrapped for a given class. "Simplify" referred to this being my second pass, producing what I thought was a clearer implementation. Happy to rename if there's a convention.

2. Tests : existing test cases do cover the patch. I've generated some examples of where the results were wrong previously and are now correct, but I'm not sure how best to convey this. I think the description in the PR covers most of the general detail. Basically there were many methods spread through the test suite that were incorrectly wrapped under the old implementation and are now correct.

3. Will have another scan for commented out code.

Do I rebase against a specific release?

richardbeare:
@wsfulton 
There were two sets of work that build into this PR that complicate the explanation. I'll attempt to describe the issues/fixes in chronological order. The initial bug addressed was the use of string matching to detect "set" and "get" at the end of names of class methods, on the grounds that swig generated function accessors named with such endings. In fact the comments in the original versions said that the patterns were "_set" and "_get", but did not test this correctly. The side effect of this approach was that class methods that happened to have matching endings were treated as swig generated accessors. We noticed this in the simpleitk project when methods named GetOffset were not wrapped correctly (meaning we were not able to call using syntax like Im$GetOffset()). In addition there was some incorrect initialization in the string logic meaning that some tests weren't correct. PR #1510 "fixed" some of these issues using the string based design, partly to serve as a baseline.

This patch uses what I hope is a more correct way of processing class methods, based on the internal swig structures rather than names.  This
has highlighted many examples of previously incorrect code, mostly
caused by misidentification of class methods as accessors. The upshot is that the
various R level accessors have changed. In `abstract_access` the original code allowed
the following, incorrect, calls:

```r
dd <- D()
dd$z <- 2
dd[["z"]] <- 2
```
These access methods are no longer supported by the R interface.

I've added a runtime test for `abstract_access` to confirm this.

richardbeare:
@wsfulton Anything else you want me to do on this one?

richardbeare:
@wsfulton Had a chance to check this out? Anything you'd like me to look at further?

wsfulton:
Hi @richardbeare . Am looking at this now. I'm comparing the generated code before and after the change. I don't understand some of the changes, such as the changes in the inherit_void testcase:
```diff
--- original/inherit_void_arg.R	2020-01-30 07:24:03.615158787 +0000
+++ inherit_void_arg.R	2020-01-30 07:34:16.118717205 +0000
@@ -213,31 +213,17 @@
 class(`A_f`) = c("SWIGFunction", class('A_f'))
 
 # Start of accessor method for A
-setMethod('$<-', '_p_A', function(x, name, value)
+setMethod('$', '_p_A', function(x, name)
 
 {
   accessorFuns = list('f' = A_f);
   ;        idx = pmatch(name, names(accessorFuns));
   if(is.na(idx)) 
-  return(callNextMethod(x, name, value));
+  return(callNextMethod(x, name));
   f = accessorFuns[[idx]];
-  f(x, value);
-  x;
-}
-
-
-);
-setMethod('[[<-', c('_p_A', 'character'),function(x, i, j, ..., value)
-
-{
-  name = i;
-  accessorFuns = list('f' = A_f);
-  ;        idx = pmatch(name, names(accessorFuns));
-  if(is.na(idx)) 
-  return(callNextMethod(x, name, value));
-  f = accessorFuns[[idx]];
-  f(x, value);
-  x;
+  function(...){
+    f(x, ...)
+  };
 }
 ```
Firstly, in R, what is the difference between "$<-" and "$" methods. Secondly, why has this changed given your patch?

File: Source/Modules/r.cxx [hunk 1]
```diff
@@ -1815,14 +1830,9 @@ int R::functionWrapper(Node *n) {
     /* Add the name of this member to a list for this class_name.
        We will dump all these at the end. */
 
-    int n = Len(iname);
-    char *ptr = Char(iname);
-    bool isSet(0);
-    if (n > 4) isSet = Strcmp(NewString(&ptr[n-4]), "_set") == 0;
+    bool isSet(GetFlag(n, "memberset"));
```
wsfulton:
```suggestion
    bool isSet = GetFlag(n, "memberset") ? true : false;
```
To fix the warning compiling with Visual Studio

richardbeare:
I'll get into more detail over the next few days, but start with the R details.

1) the $ operator in R is for list access. i.e
```R
listobj$element
```

returns the list entry named "element" from the variable listobj. Lists (and vectors) in R can have named elements and thus act a lot like python dictionaries.

The following notation is also allowed to assign to list elements:

```R
listobj$newelement <- 5
```
Will set the entry named newelement to 5.

It is also possible to overload the $ operator for other data types so that the same notation can be used. When this is done the access operator is implemented as a function named `$` and the assignment object is implemented as a function named `$<-`. There may also be functions named `[` and `[<-` for the bracket operators and `[[` and `[[<-` for double bracket operators.

I'll need to check your query in detail, but my memory is that there were many cases in which the old code was incorrectly generating assignment operators (for things that should not have had assignment operators). I think there are some comments about that in the test cases I created.

wsfulton:
I see, so this is the way attribute accessors are provided. The changes remove this incorrect attribute access.

If you could enhance R.html with these basics as to how global and member variables are wrapped and a bit about the implementation, it would really help. The R docs are currently in a very poor state and users of SWIG (and maintainers like me) often aren't all that familiar with the target language, so some examples of basic wrapping would be useful.

wsfulton:
I've merged it in, thanks for the contribution. I've tried to sum up this change for users in f0c922928765b9c1e972f07abec450b11e23dbf5. Please suggest any improvements in this ticket and I'll make the corrections.

PR #1512, opened 2019-04-10
fixing memleak when using stl containers of shared_ptr objects in python

ddurham2:
swig/swig#773 fixing memleak of shared_ptr objects in python with creating a [wrapped] vector\<shared_ptr\<Foo\>\> from a list of shared_ptr\<Bar\> where Bar is derived from Foo.

### DIFF
--- a/Lib/python/pystdcommon.swg
+++ b/Lib/python/pystdcommon.swg
@@ -45,13 +45,23 @@ namespace swig {
   template <class Type>
   struct traits_asptr {   
     static int asptr(PyObject *obj, Type **val) {
-      Type *p = 0;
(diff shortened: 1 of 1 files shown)


lballabio:
@wsfulton : for what is worth, I patched SWIG 4.0.1 with this change, regenerated wrappers for QuantLib (we had run into this problem, see https://github.com/lballabio/QuantLib-SWIG/issues/212) and the code I threw at them ran without ill effects.  The leak vanished as expected.  Unless you see other problems, I'd second adding this to next release.

wsfulton:
Thanks for the nudge, I'm looking at this next.

PR #1515, opened 2019-04-18
Fix parsing of enums with trailing comma with -doxygen

### LINKED ISSUE #1514: Doxygen parsing fails if last enum item is followed by comma
Example of enum with comma after the last item and doxygen comment, which fails:
<pre>
    /**
     * Testing comments after enum items
     */
     enum SomeAnotherEnum3
     {
       SOME_ITEM_100, ///< Post comment for the first item
       SOME_ITEM_200, ///< Post comment for the second item
       SOME_ITEM_300, ///< This comma is valid code but parsing fails with -doxygen flag
     };
</pre>
This is valid C/C++ code and when SWIG is run without -doxygen flag there is no error.
SWIG 4.0.0. beta1

vadz:
To correctly parse Doxygen post-comments after the trailing comma,
switch to right recursion in "enumlist" production rule definition: this
does consume more stack space when parsing, but makes the rules much
easier to write and to understand and hopefully shouldn't result in any
problems with real code (which shouldn't have thousands of enums items
in it).

Closes #1514.

### DIFF
--- a/Examples/test-suite/doxygen_parsing_enums.i
+++ b/Examples/test-suite/doxygen_parsing_enums.i
@@ -32,4 +32,9 @@
 		SOME_ITEM_30  ///< Post comment for the third item
 	};
 
+	enum SomeEnumWithTrailingComma
+	{
+		SOME_ITEM_100, ///< Post comment after comma.
+		SOME_ITEM_200, ///< Post comment after last comma.
+	};
 %}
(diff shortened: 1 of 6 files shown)


ZackerySpytz:
It seems that there are test failures...

markok314:
My enums are successfully parsed now, but can't check the output, because SWIG crashes at a later phase now (segmentation fault), but this might be related to some other issue. I'll investigate it and provide more info next week.

vadz:
> It seems that there are test failures...

Sorry, I forgot to commit the changes to Java test suite.

wsfulton:
Great, thanks, I've merged this for the final 4.0.0 release.

PR #1517, opened 2019-04-19
Don't crash if type is not available in Python Doxygen code

### LINKED ISSUE #1516: SWIG core-dumps if @return doxygen directive is given in struct comment.
The following snippet can be used to reproduce the problem:

<pre>
/**
 * My struct.
 *
 * @return this crashes SWIG
 */
struct MyStruct {
};
</pre>

I know it makes no sense to use @return directive in this context, but crashing is a bit too strong reaction.

vadz:
Fix crash if "@return" Doxygen tag was used on a node without any return
type (such as a class, for example). Ignoring it might not be the best
thing to do, but it's definitely better than crashing and it's not
really clear what else could be done anyhow.

Closes #1516.

### DIFF
--- a/Examples/test-suite/doxygen_misc_constructs.i
+++ b/Examples/test-suite/doxygen_misc_constructs.i
@@ -104,6 +104,9 @@
 
     };
 
+    /// @return This is a bad place for this tag, but it should be ignored.
+    struct StructWithReturnComment {};
+
     /**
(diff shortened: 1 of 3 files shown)


vadz:
Updated after rebasing on d67c133c4a85c72f77bce245b7db4ae047202396 to avoid the unrelated tests failures.

PR #1518, opened 2019-04-23
Troubleshooting section for Doxygen comments

markok314:
I added a section describing possible problems with SWIG when command line switch -doxygen is used and possible solutions. I added an example, which causes error _Error in input_ but is trivial to fix in code comment.


wsfulton:
Thanks, committed for swig 4.0.0

PR #1521, opened 2019-04-26
Use "git-describe" as SWIG version in configure

vadz:
This allows to distinguish between the different pre-release versions of
SWIG 4.0.0 instead of identifying all of them as just "4.0.0".

Update the code producing SWIG_VERSION preprocessor symbol value to
ignore everything after the first dash character in PACKAGE_VERSION.

---

I think this is quite useful to have as otherwise all recent SWIG versions identify itself as "4.0.0", which is not very informative.

### DIFF
--- a/Source/Modules/main.cxx
+++ b/Source/Modules/main.cxx
@@ -900,8 +900,11 @@ int SWIG_main(int argc, char *argv[], const TargetLanguageModule *tlm) {
   Preprocessor_define((DOH *) "SWIG 1", 0);
   Preprocessor_define((DOH *) "__STDC__", 0);
 
-  // Set the SWIG version value in format 0xAABBCC from package version expected to be in format A.B.C
+  // Set the SWIG version value in format 0xAABBCC from package version expected to be in format A.B.C[-suffix]
   String *package_version = NewString(PACKAGE_VERSION); /* Note that the fakeversion has not been set at this point */
+  if (char* const dash = strchr(Char(package_version), '-')) {
+    *dash = '\0';
+  }
   char *token = strtok(Char(package_version), ".");
   String *vers = NewString("SWIG_VERSION 0x");
   int count = 0;
--- a/configure.ac
+++ b/configure.ac
@@ -2,7 +2,7 @@ dnl Process this file with autoconf to produce a configure script.
 dnl The macros which aren't shipped with the autotools are stored in the
 dnl Tools/config directory in .m4 files.
 
-AC_INIT([swig],[4.0.0],[http://www.swig.org])
+AC_INIT([swig],patsubst(m4_esyscmd_s([git describe --tags --abbrev=7 --dirty]),[^rel-]),[http://www.swig.org])
 AC_PREREQ(2.60)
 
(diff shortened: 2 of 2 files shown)


wsfulton:
git is not a dependency for users running autogen.sh, for example downloading the source tarball from github with this patch, I get:
```
$ ./autogen.sh
+ test -d Tools/config
+ aclocal -I Tools/config
sh: 1: git: not found
```
This is an unacceptable backwards step.

Once autogen.sh has run, the information reported by `swig -version` is then out of date, for example, when local changes have been made to the code.

These kind of problems have been discussed in #1101. As suggested there, the information is unreliable, which is worse than the current approach which is accepting `swig -version` is only a rough guide. 

I think we should look for another project that has solved the ambiguities of providing reliable version information from git based code and base a solution on that. Right now we know that the version display is approximate and unless we can absolutely provide a reliable version information this kind of change is a false security blanket. Versioneer was mentioned in #1101, but the impression I have right now is that this problem is complicated and not completely solved, but I welcome being proved wrong! 

This issue was raised to distinguish the beta 4.0.0 version from the final 4.0.0 version. This is only the 2nd time we've put out a beta release... it's only done for major changes of which we may never have another - it certainly felt like the most challenging release to put wrt backwards compatibility to me. Maybe we can skip beta releases and just use version numbers instead in the future?

vadz:
This is a bit orthogonal, but I'm pretty sure that the release archives should include `configure` and `Makefile.in` instead of requiring the users to have `auto{conf,make}` installed. And yes, this does mean that they need to be created separately instead of counting on GitHub auto-generated links, but this is just the way it is.

Sorry for somehow missing #1101, now that I read through it, I think @ojwb's idea there is indeed the way to go. I'm not sure if it's worth pursuing however, if you're looking for "absolutely reliable version information", I think it's hopeless anyhow. OTOH personally I believe having reasonably accurate version information would be nice and very useful and the only thing to use for security purposes is the hash of the file and not its version anyhow.

If you'd be interested in continuing with this, please let me know what the goal should be: we can have (almost) reliable version information at the price of extra build steps that would be always executed even if they're not really needed, or we could have a (almost) zero-overhead version information that could get out of sync but that you could get resynchronize by just doing `rm .version` or something like this manually.

Otherwise I'll just stick with this patch locally and won't bother you any more...

vadz:
I don't think it's useful to keep this open as there doesn't seem to be any interest in this.

ojwb:
> This is a bit orthogonal, but I'm pretty sure that the release archives should include `configure` and `Makefile.in` instead of requiring the users to have `auto{conf,make}` installed. And yes, this does mean that they need to be created separately instead of counting on GitHub auto-generated links, but this is just the way it is.

Replying to a fairly old comment, but the download tarball linked to from http://www.swig.org/download.html **is** created separately and **does** contain `configure`, `Makefile.in`, etc.

I guess you're looking at the tarballs which github seem to insist on creating for tags pushed to repos hosted there - i.e. https://github.com/swig/swig/releases.  In other projects I've found these an occasional source of user confusion, but I've never found a way to disable them.  One just has to learn to ignore those it seems...

PR #1530, opened 2019-05-06
Fix `std::wstring` return value marshalling for C#

smithx:


wsfulton:
Looks good to me, so I've grabbed your fixes, thanks. 

I've ported the critical parts of your example into our test framework. We focus on the test-suite and keep the examples to a minimum. BTW, the unicode in your example didn't display properly for me in a DOS command prompt nor in powershell (Windows 7), but it did on a Linux bash shell.

View the changes: https://github.com/swig/swig/compare/a9a7b03...10f0329.

PR #1533, opened 2019-05-07
Java/CSharp: Fix upcasting for shared_ptrs of templated types in SWIGSmartPtrUpcast

etse-dignitas:
This PR addresses an issue we've found where SWIGSmartPtrUpcast does not create a shared_ptr of the correct parent type for shared_ptrs of templated child classes. Based on the class hierarchy provided in the accompanying test case (see **cpp11_shared_ptr_template_upcast.i**), SWIG generates the following upcast functions for the Java wrapper:

```c++
SWIGEXPORT jlong JNICALL Java_cpp11_1shared_1ptr_1template_1upcast_cpp11_1shared_1ptr_1template_1upcastJNI_Derived_1SWIGSmartPtrUpcast(JNIEnv *jenv, jclass jcls, jlong jarg1) {
    jlong baseptr = 0;
    std::shared_ptr< Derived > *argp1;
    (void)jenv;
    (void)jcls;
    argp1 = *(std::shared_ptr< Derived > **)&jarg1;
    *(std::shared_ptr< Base > **)&baseptr = argp1 ? new std::shared_ptr< Base >(*argp1) : 0;
    return baseptr;
}

SWIGEXPORT jlong JNICALL Java_cpp11_1shared_1ptr_1template_1upcast_cpp11_1shared_1ptr_1template_1upcastJNI_PrintableDerived_1SWIGSmartPtrUpcast(JNIEnv *jenv, jclass jcls, jlong jarg1) {
    jlong baseptr = 0;
    std::shared_ptr< Printable< Derived > > *argp1;
    (void)jenv;
    (void)jcls;
    argp1 = *(std::shared_ptr< Printable< Derived > > **)&jarg1;
    *(std::shared_ptr< Printable< Derived > > **)&baseptr = argp1 ? new std::shared_ptr< Printable< Derived > >(*argp1) : 0;
    return baseptr;
}
```

Note how `Derived_1SWIGSmartPtrUpcast` returns a baseptr with the correct upcast type `std::shared_ptr< Base >` (`Derived` inherits `Base`). 

However,  `PrintableDerived_1SWIGSmartPtrUpcast` returns a baseptr of the same type as the pointer argument - `std::shared_ptr< Printable< Derived > >` - instead of `std::shared_ptr< Derived >` (even though `Printable< Derived >` inherits from `Derived`). The equivalent problem occurs when wrapping C# with SWIG.

I've traced the issue back to `upcastsCode` in the java/csharp.cxx modules. When invoking `Replaceall(bsmart, rclassname, rbaseclass);`, the value of `bsmart` is actually an expanded version of the class type:

```c++
std::shared_ptr<(Printable<(Derived)>)>   //bsmart
```

While the values of `rclassname` and `rbaseclass` are not expanded:

```c++
Printable< Derived > //rclassname
Derived                     //rbclassname
```
The comparison of the expanded class type vs. non-expanded class type results in `Replaceall` replacing nothing, since the strings don't match. I suspect this issue may occur for other variations of class type expansions, but our project was seeing issues primarily with templates. Without virtual inheritance, this type discrepancy didn't have any noticeable impact when running the test case. However, with virtual inheritance, the type discrepancy results in a segfault when trying to access any inherited member function through the wrapper object in Java/C#.

This fix proposes un-expanding the class type before invoking `Replaceall` for a more comparable comparison when replacing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -432,10 +432,7 @@ install:
   - if test -n "$CPP11"; then CONFIGOPTS+=(--enable-cpp11-testing --without-maximum-compile-warnings "CXXFLAGS=-std=c++11 -Wall -Wextra" "CFLAGS=-std=c11 -Wall -Wextra") && export CSTD=c11 && export CPPSTD=c++11; fi
   - if test -n "$CPP14"; then CONFIGOPTS+=(--enable-cpp11-testing --without-maximum-compile-warnings "CXXFLAGS=-std=c++14 -Wall -Wextra" "CFLAGS=-std=c11 -Wall -Wextra") && export CSTD=c11 && export CPPSTD=c++14; fi
   - if test -n "$CPP17"; then CONFIGOPTS+=(--enable-cpp11-testing --without-maximum-compile-warnings "CXXFLAGS=-std=c++17 -Wall -Wextra" "CFLAGS=-std=c17 -Wall -Wextra") && export CSTD=c17 && export CPPSTD=c++17; fi
-  - ls -la $(which $CC)
-  - ls -la $(which $CXX)
-  - $CC --version
-  - $CXX --version
+  - ls -la $(which $CC) $(which $CXX) && $CC --version && $CXX --version
 script:
   - echo 'Configuring...' && echo -en 'travis_fold:start:script.1\\r'
   - if test -n "$SWIGLANG"; then CONFIGOPTS+=(--without-alllang --with-$WITHLANG); fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -429,12 +429,12 @@ before_install:
 install:
   - if test "$TRAVIS_OS_NAME" = "linux"; then source Tools/travis-linux-install.sh; fi
   - if test "$TRAVIS_OS_NAME" = "osx"; then source Tools/travis-osx-install.sh; fi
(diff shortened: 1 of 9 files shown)


wsfulton:
Hi @etse-dignitas. Well identified and great to see a testcase included. Could you please make a couple of minor tweaks ...

1. In the runtime tests, check the return values of GetResult and GetFormatted.
2. The manipulation of types should be done on `SwigType*` not `String*` types. So the bug is SwigType_typedef_resolve_all is taking a `String*` instead of a `SwigType*`. The values passed to it should be the `SwigType*` not the stringified name that is returned from SwigType_namestr. The last two parameters of upcastsCode will need correcting to SwigType* and SwigType_namestr only called when needed. Do the Replaceall using the `SwigType*` values.

Can squash your commits so that there is no debugging code commits in the pull request. Thanks.

wsfulton:
ping @etse-dignitas, did you get a chance to make these small changes?

etse-dignitas:
@wsfulton sorry for the long delay, I got sidetracked by other issues. Unfortunately I have been transferred over to another project, and I don't know when I will be able to work on this again (possibly months). If you or anyone else wants to take this over, feel free to. Otherwise, it will be some time before I can get to address and fully test these changes again.

wsfulton:
While looking at the `SwigType_typedef_resolve_all` fix, I noticed that this function needs enhancing for templates but I think we should avoid using it altogether but using the base class' "feature:smartptr" instead. See commits 564a91864a84046609bedcfc7cb5e5560fd62377 and b3da344765625f22847bd56ee22b392dd16c4c84 for future work.

PR #1534, opened 2019-05-09
Add "onlywrapped" feature

sethrj:
Automatically hide selected function signatures if they use unwrapped SWIG types.


sethrj:
Sorry, wrong target repository 😞

PR #1535, opened 2019-05-09
Add JS Native Directive API

TekuConcept:
Adds `%native(function)` api support for JavaScript. Custom node functions can now be written and included in the same swig module such as functions with function callbacks as parameters.
Deprecates #763 

**Additions Include:**
- native directive support for javascript
- updated JS documentation
- native directive example project
- native directive testcase

### DIFF
--- a/Examples/javascript/native/example.i
+++ b/Examples/javascript/native/example.i
@@ -4,100 +4,37 @@
 
 %wrapper
 %{
-#include <cstring>
-#include <string>
-
-
-#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
-  typedef v8::Handle<v8::Value> V8ReturnValue;
-  typedef v8::Arguments V8Arguments;
-# define V8_RETURN(val) return scope.Close(val)
-#else
-  typedef void V8ReturnValue;
-  typedef v8::FunctionCallbackInfo<v8::Value> V8Arguments;
-# define V8_RETURN(val) args.GetReturnValue().Set(val); return
-#endif
-#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032318)
-#    define V8_UNDEFINED() v8::Undefined()
-#else
-#    define V8_UNDEFINED() v8::Undefined(v8::Isolate::GetCurrent())
-#endif
-#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032117)
-#    define V8_HANDLESCOPE() v8::HandleScope scope
-#elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032224)
-#    define V8_HANDLESCOPE() v8::HandleScope scope(v8::Isolate::GetCurrent());
-#else
-#    define V8_HANDLESCOPE() v8::HandleScope scope(v8::Isolate::GetCurrent());
-#endif
-
-
-// 'unused' attribute only necessary for GNUC < v3.4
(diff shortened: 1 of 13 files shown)


wsfulton:
Could you please complete the testcase and add in a native_directive_runme.js file, something like the one for Java, native_directive_runme.java?

wsfulton:
New examples need adding to the Examples/javascript/check.list file. Can you add the new native example to this list and then it will be run by our Continuous Integration framework on Travis. I'm afraid that this will require it to work with all 3 Javascript engines.

Examples/javascript/native/example.i
```diff
@@ -0,0 +1,104 @@
...
+#    define V8_UNDEFINED() v8::Undefined(v8::Isolate::GetCurrent())
+#endif
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032117)
+#    define V8_HANDLESCOPE() v8::HandleScope scope
+#elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032224)
+#    define V8_HANDLESCOPE() v8::HandleScope scope(v8::Isolate::GetCurrent());
+#else
+#    define V8_HANDLESCOPE() v8::HandleScope scope(v8::Isolate::GetCurrent());
+#endif
+
```
wsfulton: Did you see the macros in javascriptrun.swg which provide SWIGV8_HANDLESCOPE, SWIGV8_RETURN? Probably better to just use those instead.

TekuConcept: Good idea! I actually was aware of those, but I was fence-sitting on using them. The idea of tightly coupling objects.

I will update them accordingly!

TekuConcept:
Yay! It passes!

@wsfulton I replaced the javascript example with something simpler than callbacks since the focus of this is simply to demonstrate the native directive's functionality and use-case.

Note: This PR depends on #1569 which fixes a node-gyp dependency bug for node-v4 builds.

Source/Modules/javascript.cxx
```diff
@@ -768,6 +786,22 @@ int JSEmitter::emitWrapperFunction(Node *n) {
...
 }
 
+int JSEmitter::emitNativeFunction(Node *n) {
+  String *wrapname = Getattr(n, "wrap:name");
+  // ismember never seems to be the case;
+  // it is technically possible to add native member functions,
+  // just not at the moment? leaving this as an option for later;
+  // the code will automatically defaulting to static space
+  if (GetFlag(n, "ismember") != 0)
+    Setattr(n, "feature:extend", "1"); // member space
```
wsfulton: Setting this flag is reserved for the parser when parsing %extend. So this code needs to be reimplemented. %extend has nothing to do with %native.

Are you trying to filter out non-static member functions used with %native? Does one of the testcases fail without this flag setting?

TekuConcept: That is what I was asking about several months ago on the mailing list. (June 26)
When I wrote the function implementation I referenced the python implementation, which at the time also used it.

The python implementation had set the value of `feature:extend` to 1, which I observed made the function a member function. However, when writing the test case for the JS impl., I saw that Java's function were global / static. Hence why I was confused.

Looking over the python implementation now, it looks like the line was removed as a correction. My JS implementation should work without the modification of `feature:extend`. I will remove it.

wsfulton: Thanks, I'll go with this then. For the future, this is a public code base, please use commit message that say something more useful than "Dev Checkpoint".

TekuConcept: Sorry; I was under the impression that I would eventually need to flatten all my commits into one before the PR was merged. I didn't want to waste time naming every little fix if the commits were just going to be absorbed, otherwise I would have given them each unique names.

PR #1545, opened 2019-05-23
Add support for appending to a typemap definition

### LINKED ISSUE #1541: Allow appending to a typemap
Not for the first time, I'm running into a problem because I need to define some aspect of custom behaviour typemap for a wrapped class using a typemap which is already defined elsewhere to implement another aspect. I can't easily just define a typemap doing both things at once because the existing typemap is inside a reusable macro and all workarounds are really ugly (it doesn't help that the existing code is already quite complicated).

It would be really helpful if I could just write `typemap(cscode, append=1) MyClass { ... }` instead. And I'm not the first person wishing something like this were possible, see e.g. [this SO question](https://stackoverflow.com/questions/12103206/is-it-possible-to-add-text-to-an-existing-typemap-in-swig).

So would a patch implementing this be acceptable? And, if so, would you have any hints about the best way to do this? In particular, I'm not sure if this should be handled at the parser level or in `typemap.c`, what do you think?

vadz:
If append=1 modifier is used, append to the existing typemap definition
rather than replacing it.

This is especially useful for cscode/javacode typemaps which can be
naturally combined.

Closes #1541.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.1 (in progress)
 ===========================
 
+2019-05-23: vadz
+            Add append=1 modifier for typemap definitions.
+
 2019-05-22: ferdynator
 	    [PHP] #1528 Don't add a closing '?>' PHP tag to generated files.
 	    PSR-2 says it MUST be omitted for files containing only PHP.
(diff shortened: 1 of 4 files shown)


ojwb:
I think this would be helpful for #1554.

vadz:
@wsfulton Thanks for merging my other recent PRs, but I wonder if you could have overlooked this one. It's pretty trivial, yet, IMHO, quite useful. Could you please look at it if you've really just didn't notice it or do you have any objections to merging it? TIA!

wsfulton:
@vadz, sorry, I've always got a backlog of issues to deal with and tend to deal with the easiest ones first, but I've commented on #1541 now.
@ojwb if you think this solution would help #1554, sorry, I also need to look at it which looks horribly non-trivial.

PR #1547, opened 2019-05-25
Doxygen comment whitespace fixes for python

mcfarljm:
Fixes and cleanup of the whitespace in the translated comments generated for python with the doxygen flag:

- Fix inconsistent indentation of content for `\code` command between inline and block usage
- Fix incorrect indentation for `\code` when using an alternative doxygen comment style that includes extra spaces
- Remove extraneous newlines in the generated comments surrounding `\code`, `\verbatim`, and `\f` math commands
  - Also ensure that generated pydoc output does contain an empty line in front of code and math blocks, which is required for sphinx
  - The end result is that the structure of the pydoc output should resemble that of the original doxygen comments, except that newlines are added before code and math blocks if necessary
- Remove extra newline added at the start of generated docstrings
- Update doxygen python `_runme.py` files for new expected output
- Add new test case `doxygen_basic_translate_style2.i` to test handling of the alternative doxygen comment style, which was not previously represented in the test cases

### DIFF
--- a/Source/Doxygen/pydoc.cxx
+++ b/Source/Doxygen/pydoc.cxx
@@ -140,14 +140,48 @@ static void trimWhitespace(string &s) {
 
 // Erase the first character in the string if it is a newline
 static void eraseLeadingNewLine(string &s) {
-  if ((! s.empty()) && s[0] == '\n')
+  if (!s.empty() && s[0] == '\n')
     s.erase(s.begin());
 }
 
 // Erase the last character in the string if it is a newline
 static void eraseTrailingNewLine(string &s) {
-  if ((! s.empty()) && s[s.size() - 1] == '\n')
+  if (!s.empty() && s[s.size() - 1] == '\n')
     s.erase(s.size() - 1);
+}
+
+// Check the generated docstring line by line and make sure that any
+// code and verbatim blocks have an empty line preceding them, which
+// is necessary for Sphinx.  Additionally, this strips any empty lines
+// appearing at the beginning of the docstring.
+static string padCodeAndVerbatimBlocks(const string &docString) {
+  std::string result;
+
+  std::istringstream iss(docString);
+
+  // Initialize to false because there is no previous line yet
+  bool lastLineWasNonBlank = false;
+  
+  for (string line; std::getline(iss, line); result += line) {
+    if (!result.empty()) {
+      // Terminate the previous line
+      result += '\n';
+    }
+
+    const size_t pos = line.find_first_not_of(" \t");
+    if (pos == string::npos) {
+      lastLineWasNonBlank = false;
+    } else {
+      if (lastLineWasNonBlank &&
+	  (line.compare(pos, 13, ".. code-block") == 0 ||
+	  line.compare(pos, 7, ".. math") == 0)) {
+	// Must separate code or math blocks from the previous line
+	result += '\n';
+      }
+      lastLineWasNonBlank = true;
+    }
+  }
+  return result;
 }
 
 /* static */
@@ -862,6 +896,9 @@ String *PyDocConverter::makeDocumentation(Node *n) {
 
     // remove the last '\n' since additional one is added during writing to file
     eraseTrailingNewLine(pyDocString);
+
+    // ensure that a blank line occurs before code or math blocks
+    pyDocString = padCodeAndVerbatimBlocks(pyDocString);
 
     if (m_flags & debug_translator) {
       std::cout << "\n---RESULT IN PYDOC---" << std::endl;
--- a/Source/Doxygen/pydoc.cxx
+++ b/Source/Doxygen/pydoc.cxx
@@ -519,7 +519,7 @@ void PyDocConverter::handlePlainString(DoxygenEntity &tag, std::string &translat
 }
 
 void PyDocConverter::handleTagVerbatim(DoxygenEntity &tag, std::string &translatedComment, const std::string &arg) {
-  translatedComment += arg + " ";
+  translatedComment += arg;
   for (DoxygenEntityListCIt it = tag.entityList.begin(); it != tag.entityList.end(); it++) {
     translatedComment += it->data;
   }
--- a/Source/Doxygen/pydoc.cxx
+++ b/Source/Doxygen/pydoc.cxx
@@ -493,6 +493,11 @@ void PyDocConverter::handleCode(DoxygenEntity &tag, std::string &translatedComme
   std::string code;
   handleTagVerbatim(tag, code, arg);
 
+  // Try and remove leading newline, which is present for block \code
+  // command:
+  if (code[0] == '\n')
+    code.erase(code.begin());
+
   translatedComment += codeIndent;
   for (size_t n = 0; n < code.length(); n++) {
     if (code[n] == '\n') {
--- a/Source/Doxygen/pydoc.cxx
+++ b/Source/Doxygen/pydoc.cxx
@@ -138,6 +138,18 @@ static void trimWhitespace(string &s) {
     s.erase(lastNonSpace + 1);
 }
 
+// Erase the first character in the string if it is a newline
+static void eraseLeadingNewLine(string &s) {
+  if ((! s.empty()) && s[0] == '\n')
+    s.erase(s.begin());
+}
+
+// Erase the last character in the string if it is a newline
+static void eraseTrailingNewLine(string &s) {
+  if ((! s.empty()) && s[s.size() - 1] == '\n')
+    s.erase(s.size() - 1);
+}
+
 /* static */
 PyDocConverter::TagHandlersMap::mapped_type PyDocConverter::make_handler(tagHandler handler) {
   return make_pair(handler, std::string());
@@ -422,13 +434,11 @@ void PyDocConverter::handleParagraph(DoxygenEntity &tag, std::string &translated
 void PyDocConverter::handleVerbatimBlock(DoxygenEntity &tag, std::string &translatedComment, const std::string &) {
   string verb = translateSubtree(tag);
 
-  if ((! verb.empty()) && verb[0] == '\n')
-    verb.erase(verb.begin());
-  
+  eraseLeadingNewLine(verb);
+
   // Remove the last newline to prevent doubling the newline already present after \endverbatim
   trimWhitespace(verb); // Needed to catch trailing newline below
-  if ((! verb.empty()) && verb[verb.size()-1] == '\n')
-    verb = verb.substr(0, verb.size()-1);
+  eraseTrailingNewLine(verb);
   translatedComment += verb;
 }
 
@@ -505,8 +515,7 @@ void PyDocConverter::handleCode(DoxygenEntity &tag, std::string &translatedComme
 
   // Try and remove leading newline, which is present for block \code
   // command:
-  if ((! code.empty()) && code[0] == '\n')
-    code.erase(code.begin());
+  eraseLeadingNewLine(code);
 
   translatedComment += codeIndent;
   for (size_t n = 0; n < code.length(); n++) {
@@ -529,8 +538,7 @@ void PyDocConverter::handleCode(DoxygenEntity &tag, std::string &translatedComme
   // For block commands, the translator adds the newline after
   // \endcode, so try and compensate by removing the last newline from
   // the code text:
-  if ((! translatedComment.empty()) && translatedComment[translatedComment.size()-1] == '\n')
-    translatedComment = translatedComment.substr(0, translatedComment.size()-1); // use translatedComment.pop_back() in C++ 11
+  eraseTrailingNewLine(translatedComment);
 }
 
 void PyDocConverter::handlePlainString(DoxygenEntity &tag, std::string &translatedComment, const std::string &) {
@@ -852,9 +860,7 @@ String *PyDocConverter::makeDocumentation(Node *n) {
   if (!pyDocString.empty()) {
 
     // remove the last '\n' since additional one is added during writing to file
-    if (pyDocString[pyDocString.size() - 1] == '\n') {
-      pyDocString.erase(pyDocString.size() - 1);
-    }
+    eraseTrailingNewLine(pyDocString);
 
     if (m_flags & debug_translator) {
       std::cout << "\n---RESULT IN PYDOC---" << std::endl;
--- a/Source/Doxygen/pydoc.cxx
+++ b/Source/Doxygen/pydoc.cxx
@@ -496,7 +496,6 @@ void PyDocConverter::handleCode(DoxygenEntity &tag, std::string &translatedComme
   IndentGuard indent(translatedComment, m_indent);
 
   trimWhitespace(translatedComment);
-  translatedComment += '\n';
 
   // Use the current indent for the code-block line itself.
   translatedComment += indent.getFirstLineIndent();
--- a/Source/Doxygen/pydoc.cxx
+++ b/Source/Doxygen/pydoc.cxx
@@ -498,7 +498,7 @@ void PyDocConverter::handleCode(DoxygenEntity &tag, std::string &translatedComme
   translatedComment += ".. code-block:: c++\n\n";
 
   // For now on, use extra indent level for all the subsequent lines.
-  codeIndent += m_indent;
+  codeIndent = m_indent;
 
   std::string code;
   handleTagVerbatim(tag, code, arg);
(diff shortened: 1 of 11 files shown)


Source/Doxygen/pydoc.cxx
```diff
@@ -488,11 +498,16 @@ void PyDocConverter::handleCode(DoxygenEntity &tag, std::string &translatedComme
   translatedComment += ".. code-block:: c++\n\n";
 
   // For now on, use extra indent level for all the subsequent lines.
-  codeIndent += m_indent;
+  codeIndent = m_indent;
```
vadz: Are you sure about this change? It looks logical to me to add to the existing indent, i.e. I don't see why was the old code wrong. The commit message speaks about a new unit test showing, but it doesn't seem to have been added yet?

And if this is really correct, the comment above needs to be updated after this change, it's not an "extra" indent level any longer. In fact, the whole `codeIndent` variable should be just removed if we really can just use `m_indent` instead.

mcfarljm: Thanks for the review.  As far as I can tell this change is correct.  The indentation handling in this code is a little bit confusing, and took me some time (and lots of printf debugging) to wrap my head around as I'm not the original author.  Above the modified line in question, `codeIntent` stores the indentation necessary for adding text to `translatedComments`.  With the standard doxygen style represented in the test cases, this would be an empty string.  At the line in question the current value of `m_indent` is set as 4 spaces by the indent guard.  Those spaces are added so that the translated comments in the code block are indented by that amount.  So with the standard doxygen comment style, the change that I introduced would have no effect.

However, the problem is that with an alternative doxygen style (see below), there is a space in front of everything that was handled by the parser.  This extra space shows up in the variable `codeIndent`, and it is needed when directly adding a new line of text to `translatedComment`.  However, I don't believe it is correct to then use that extra space plus the 4 indentations in front of the translated content, because the translated content itself already contains that extra space in front.  This results in 5 space indentation within code blocks when using the alternative comment style.

Having said that, it might be worth adding some additional comments to the code for better clarification.  What workflows are typically used to update pull requests?  In this particular case a new commit could make sense, but in other cases is a rebase and force push to the branch used?

I did attempt to add a new test case in 91a90b8d2, but it's my first time so I certainly may have missed something!  The new case is doxygen_basic_translate_style2.i, and I updated common.mk.

Here is what the alternative comment style looks like:
<pre>
/** Line 1
 *  Line 2
 */
</pre>
This results in an extra leading space in content read by the parser.  It seems to be mostly already handled correctly by the translator, but as I mentioned indentation within a code block was not correct.  And this comment style was not previously being tested.  (By the way, is it preferable to commit the new test case before the fix, so that the failure to achieve the intended behavior can be more clearly demonstrated?)

mcfarljm: > 
> 
> The commit message speaks about a new unit test showing, but it doesn't seem to have been added yet?

I just pushed an update 64b2113e6 because there was an oversight in the original commit of the new test case, doxygen_basic_translate_style2.  The python _runme.py file with expected results was referencing an out of date module name that I had since changed.  This may explain questions about the new unit test, which would have failed (was passing on my local machine due to old files hanging around...)

With regards to my earlier question about how to make updates to the PR: I'm happy to rebase this (and some of the other suggested cleanup I'm working on) and force push the updates to clean up the history if that is preferred to new commits.

mcfarljm:
> 
> 
> 
> Mostly they look good to me and the only comment I have is that it might be useful to add helper functions for stripping newlines at the beginning/end of strings instead of duplicating these checks (which are easy to write incorrectly, as the history of these changes shows, with `!str.empty()` check forgotten in the initial commit).

Thanks again for the review and good suggestions.  In addition to the unit test correction mentioned above (64b2113e6), I also pushed two new commits to use helper functions for stripping newlines, and to clarify comments relating to code indent.  I believe that it is ready now unless you want a rebase to clean up the history.

vadz:
Thanks for the explanations and the cleanup! This looks good to me now, but let's wait for @wsfulton to have a look at this and merge it.

mcfarljm:
Great.  I actually found one more whitespace change that I think should be added.  I should have noticed this before, but the translator puts an extra newline before the Python code block.  Right now, the following translation happens:
<pre>
This is a code block:
\code
some code
\endcode
</pre>
becomes
<pre>
This is a code block:

.. code-block:: c++

    some code
</pre>

But I recommend that we go ahead and remove that preceding newline also, in order to more closely match the structure of the user's original comments:
<pre>
This is a code block:
.. code-block:: c++

    some code
</pre>

If the user wants that newline separation, they simply include it in the original doxygen comments.  Otherwise, without removing this, there is no way to get Python output that doesn't include the extra newline.

This is a trivial fix to the code as it only involves removing one line that explicitly adds this newline to the translated comments (and then updating the doxygen python test output).  Unless there are objections, I would like to push this final commit before the merge.

By the way I am also working on updates to recognize and handle Python doctest code examples (that start with `>>>` and are not indented), and having this more precise whitespace will help enable very clean looking pydoc output.  For example:
<pre>
Python example:
\code{.py}
>>> 1 +1
2
\endcode
</pre>
becomes
<pre>
Python example:
>>> 1 +1
2
</pre>

mcfarljm:
I went ahead and pushed the final commit to remove the extra newline inserted before the code block, as explained in my previous comment.  (So much extra whitespace!)  Should now be ready for final approval.

I am currently working on some more interesting updates for handling the `\param` and `\code` doxygen commands.  I will plan to submit them as a new pull request once this one is merged.

mcfarljm:
I may have spoken to soon.  Suggest that we put this one on hold to review how sphinx handles some of these cases, and what the desired behavior is for swig doxygen pydoc output.  From a quick test, it looks like sphinx does require an empty line before code and math constructs.  Meaning the example I gave above:
<pre>
This is a code block:
.. code-block:: c++

    some code
</pre>
wouldn't be processed correctly.  This suggests that probably going back to the previous behavior of adding a line before code and math blocks is the right thing.  Doxygen seems to generate the same output whether the newline is there or not (at least for `\code`), so one could expect to process code in either style, but it appears the empty line does need to be there for sphinx.

I'll do some more testing with sphinx and get a better description of specifically which commits in this PR are affected (believe that all changes to whitespace within the blocks are good).  Then we can figure out what the right path forward is.

mcfarljm:
OK, so I did more testing of how sphinx handles pydoc comments in these scenarios:

- For both `..code-block` and `.. math`, sphinx only renders correctly if there is an empty line before the command.
- There is no need for an empty line after the command.  Even code or math at the very end of the comment with no newline is fine.

The commits that affect this behavior are:

- 08fc4a02b: This commit removes the extra newline before block math, making it possible to generate pydoc that will not render correctly in sphinx
- 347656566: This commit removes the extra newline before the code block, making it possible to generate pydoc that will not render correctly in sphinx

(FYI, it's easy to see exactly how all of these commits are changing the output by looking at the test diffs.)

I think my recommendation would be to not add newlines before the code and math commands, i.e. using this branch as it currently stands.  The argument for this is that it results in better consistency between the look of the doxygen code and the raw pydoc output.  As the branch stands now, translating this:
<pre>
/** 			   
 * Here is a code block: 
 *			   
 * \code		   
 * the code		   
 * \endcode		   
 */
</pre>
will produce:
<pre>
Here is a code block:  
                       
.. code-block:: c++    
                       
    the code
</pre>

I could see an argument for adding the newline before though, in case the doxygen content was:
<pre>
/** 			   
 * Here is a code block: 		   
 * \code		   
 * the code		   
 * \endcode		   
 */
</pre>

Incidentally, with the Napoleon extension for handling numpy style documentation, sphinx will process code or math that comes immediately after a section:
<pre>
Example
-------
.. code-block:: c++    
                       
    the code
</pre>
And I have translation of `\par Example` to this type of section heading planned as a future enhancement in conjunction with a "numpy" feature option from swig.

Are there other opinions on what the behavior for swig should be with respect to adding an empty line before code and math blocks in python?  Ideally, I suppose we could analyze the content and only add the line if necessary (I will take a look at whether that is feasible, but it might not be simple).

vadz:
Sorry, I should have remembered this before, but I've indeed already run into the problem with Sphinx requiring these empty lines. For me personally (but I'm sure not only me) it's very important that Sphinx handles this correctly because this is what the end goal is: have nice Sphinx-generated documentation for the users. AFAIU the current branch breaks this, so I'm definitely for removing the offending commits.

mcfarljm:
> 
> 
> AFAIU the current branch breaks this, so I'm definitely for removing the offending commits.

Just to be clear, it depends on how the doxygen comments are written, and the following style is compatible with the current branch changes:
<pre>
/** 			   
 * Here is a code block: 
 *
 * \code
</pre>

However, I agree that the priority should be on making sure the changes don't break any code (comments) that were previously working.

Let me take a week or two to see if I can find a solution that only adds the extra newline if necessary, so that the generated pydoc is valid for sphinx but also avoids adding additional newlines when not needed.  For me, the pydoc itself is also part of the "product", not only the sphinx docs.  I find myself accessing the docstrings very frequently using the Python online help, so any effort to "beautify" the raw pydoc is I think worthwhile (whereas always adding the extra newline would produce a lot of space in the pydoc if one used the above doxygen style for code blocks).

If I can't find a relatively easy solution to conditionally add the newlines, then I'll go ahead and remove the offending commits.

vadz:
I see, I didn't realize that it was still working provided the source comment was in the correct format.

Unfortunately I don't know if all our comments are like this or not. Perhaps the change would be acceptable if we could at least output a warning for code blocks that we know won't be parsed correctly by Sphinx after translation. But always generating Sphinx-friendly output, as you plan to do, would, of course, be even better.

Sorry but I really don't know this code as well (I just fixed some bugs in it, but am not the original author), so I can't give you any advice about how to do it, unfortunately, so I can only wish you good luck!

mcfarljm:
I pushed a fix that makes sure that an empty line appears before both code and math blocks.  Both of the following:
<pre>
/** 			   
 * Here is a code block: 
 *
 * \code
</pre>
or
<pre>
/** 			   
 * Here is a code block: 
 * \code
</pre>
will be rendered as:
<pre>
Here is a code block:  
                       
.. code-block:: c++    
</pre>
The extra line is only added if needed.

Please review, and then hopefully this branch is ready to merge.

mcfarljm:
Well, I thought that the last commit resolved the issue of ensuring an empty line before the code block, but I have since come across a case that it doesn't handle:
<pre>
/** 			   
 * \par Example
 * \code
</pre>
I may poke around with this some more, but I suspect that we should back out the commits (now 3) that affect this behavior and go back to always adding an empty line in front of the code and math blocks.

Is there a preferred approach for undoing these commits?  Revert with new commits or move to a new branch to clean up the history?  Either way, once we go back I would like to add some comments to the original source code to explain why the newlines are necessary.

mcfarljm:
I worked out a better solution that postprocesses the generated pydoc at the end by scanning it line by line and inserting the necessary newlines in front of code and math blocks as needed.  Unlike the solution from daf812148, this should work regardless of the type of content appearing before the blocks.

I have not yet pushed the new commit.  For a clean history, I could rebase to remove daf812148 and then add the new commit and force push.  But I wanted to check first if that is OK before doing a force push.

vadz:
I think it's fine to force push for your own PRs unless you know that someone else has started working on your branch -- and to the best of my knowledge, this is not the case here, so IMHO you can force push just fine.

TIA!

mcfarljm:
OK, great, I went ahead and force pushed an update to replace just the last commit, so that we now have 80c93dcb5.  This commit scans the generated pydoc and adds a newline in front of code and/or math blocks if necessary.  This is verified through new test cases in doxygen_translate_all_tags.i.

I have also undated the comment at the top of this page to better describe the scope of this pull request.

At this point, any other final review is appreciated, and hopefully we can get these updates merged!

mcfarljm:
Hi @vadz, just checking whether there is anything else I need to do in order for this to be merged.  From my perspective it is ready to go.  Thanks!

mcfarljm:
> 
> Also, just to put together all my comments about `padCodeAndVerbatimBlocks()` function, I'd write it like this (pseudo-codish):

Great suggestion!  Not only is it much simpler, but it also eliminates yet another empty line that was being added at the start of the generated docstrings (there are some nuances about this that I will describe in a separate comment)!

I went ahead and incorporated all of your suggestions, rebased, and force pushed so that they now appear in daad5d664.

mcfarljm:
Regarding the newline that was previously appearing at the beginning of the generated docstrings: I believe the reason for this is that with the Javadoc comment style, the parser is picking up the first empty line of the comment, for example:
```c++
/**
 * \brief
 * Brief description.
```

Based on the logic you suggested (which is used in commit daad5d664), that first line gets "filtered out" in `padCodeAndVerbatimBlocks` because `result.empty()` is true (and actually in the example I'm showing here, 2 empty lines get filtered out because `\brief` on its own line gets converted to an empty line).  The changes can clearly be seen in the test output `doxygen_basic_translate.py`, which no longer has an empty line at the start of the docstrings.  Incidentally, the algorithm that checks the pydoc test output was not affected by this change, so test results did not need to be updated.

One edge case that I noticed occurs when the "alternative" comment style is used that starts on the first line.  If that style is used in conjunction with a `\brief` command appearing on its own line:
```c++
/** \brief
 *  Brief description.
```
Then the first line that `padCodeAndVerbatimBlocks` sees is actually a single space, not an empty line (with this comment style, the strings fed to pydoc.cxx all have this space in front).  So in this particular case, the generated docstring actually does end up with an empty line at the front (which can be seen in the `doxygen_basic_translate_style2.py` test output).

The catch is that with the alternative comment style, this is only an issue when a `\brief` command is used on a line by itself, but that's not actually how I would use the command (and it's not the way it is shown in the doxygen documentation).  So although the logic in `padCodeAndVerbatimBlocks` (or elsewhere in pydoc.cxx) could certainly be updated to handle this case and remove that line, it would be a fair bit more complex than the current logic, and it's addressing a case that does not seem to be very important (arguably a non-idiomatic use of the `\brief` command).

wsfulton:
@vadz and @mcfarljm are you both happy with the state of this pull request? If so, I'll start my review. The issue is getting rather long for me to work out the current state.

vadz:
I think it's good to go, I've tested it with our (complicated and extensive) wrappers and the results are as expected, i.e. a few unnecessary blank lines have disappeared.

Thanks!

mcfarljm:
> 
> 
> @vadz and @mcfarljm are you both happy with the state of this pull request? If so, I'll start my review. The issue is getting rather long for me to work out the current state.

Yes, I'm happy with it.  It obviously turned into a larger effort than expected, but I think overall it produces cleaner output for translated comments.  The very first comment at the top of this page is up to date and summarizes the changes.

wsfulton:
Given sphinx compatibility is important, shouldn't we be testing using sphinx?

wsfulton:
All looks okay to me for merge to master. 

I would like to see some testing using Sphinx put in though so that there are no regressions in the future. Not sure if that will be easily achieved as I get the impression Sphinx is not so easily setup/configured. Any thoughts on that?

vadz:
Sphinx is not that difficult to setup/configure, but it's yet another dependency to have and I'm not sure it's worth it: we know that if we generate doc comments in the correct format, Sphinx will do its job and are not really interested in testing Sphinx itself. So having test cases verifying that the output of SWIG is what we expect, and what Sphinx expects, should be enough IMO.

mcfarljm:
I don't think it's necessary to test with sphinx.  I think the goal is really to ensure that we are getting the expected output from swig, and that doesn't require sphinx, as long as the test cases are defined appropriately.  If there is an example in the test cases that doesn't work with sphinx but should, then that is an issue with how the test case is written.  (And everything that is in the doxygen test cases now may not necessarily be meaningful in sphinx -- there are some unusual doxygen commands being tested.)

The bigger question is probably what specific doxygen content/commands do we want to test, and how do we want them to be translated into pydoc/sphinx?  This has to be addressed in the test cases regardless of whether sphinx is part of the tests or not.  Right now there is some basic testing of math and code content, which the test cases ensure gets translated appropriately.  But as an example, the doxygen command `\par <title>` doesn't currently get translated into anything meaningful to sphinx.  One thing that I'm already working on is `%feature("numpy")` to enable translation specifically into numpy format docstrings, which use a more compact format for parameter definitions (much more friendly when viewing the Python docs directly) and use standard sections such as "Examples", "Notes", and "See Also" (which I think would best be translated from doxygen `par` commands).

PR #1552, opened 2019-05-30
[Java] Add vector constructor with size

bkotzz:
The vector class for Java has a few constructors:
- No-args
- size plus element to initialize with
- another vector
- an array of the jstype (int not Integer)
- an Iterable of the jboxtype (Integer not int)

But there doesn't exist an option to initialize a vector a certain size without setting all the elements. Only reserve is available after already creating the vector.

This exposes the constructor that just takes a size.

Then, to make the the array constructor more efficient, we can initialize the vector to the right size before the loop, and in the loop just use set (which is just operator[] in c++). This avoids the constant re-allocs of doing `push_back` in a loop.

I also added a helper function to the tests, similar to maps/sets, to make them less verbose.

PTAL @wsfulton

### DIFF
--- a/Examples/test-suite/java/li_std_vector_runme.java
+++ b/Examples/test-suite/java/li_std_vector_runme.java
@@ -30,13 +30,6 @@ public class li_std_vector_runme {
     checkThat(v1.size() == 1);
     checkThat(!v1.isEmpty());
 
-    java.util.List<Integer> sizeInit = new IntVector(10);
-    checkThat(sizeInit.size() == 10);
-    sizeInit.set(0, 55);
-    sizeInit.set(9, 66);
-    sizeInit.add(77);
-    checkThat(sizeInit.size() == 11);
-
     int sum = 0;
     for (int n : v1) {
       checkThat(n == 123);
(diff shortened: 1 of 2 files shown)


wsfulton:
I'm afraid that this constructor was purposefully taken out. See extract from CHANGES file:
https://github.com/swig/swig/blob/9c3faeb0ee6c588bc4772bd186554796cc1e446e/CHANGES.current#L78-L100
The test-suite should fail with your change. Can you use the alternatives mentioned above?

bkotzz:
@wsfulton interesting, I did not realize that. I had only run the vector test case, not the other ones. I still think we should improve the array constructor to avoid the repeated allocations. I’ll prune down the change and have that one use reserve based on the size before the loop.

bkotzz:
@wsfulton I’ve stripped down the change to just be a reserve in the constructor to avoid repeated allocations. All tests are passing except one job that timed out. Is there a way to restart that one without re-running the whole suite?

bkotzz:
All good now, @wsfulton please take a look

wsfulton:
I've restarted the failed job on Travis a while back for you.

wsfulton:
Thanks for the contribution, I've squash merged your pull request in 55e835e

PR #1554, opened 2019-06-02
WIP: Rework the java typemaps to avoid using finalize()

### LINKED ISSUE #1550: Replace use of Java finalize()
In #1237 we discussed providing a way to avoid use of Java finalize() by using `PhantomReference` or `java.lang.ref.Cleaner` instead (the latter requires JDK 9, the former is supported since JDK 1.2).

I've started an attempt to implement this, so opening a ticket so that others know I am, and to have somewhere to track progress.

So far I have a hacked up version of the java typemaps such that `java.lang.ref.Cleaner` is used instead of generating a `finalize()` method.  Most of the testsuite passes, but from the failures it seems perhaps destruction isn't actually happening.

Once this is working, I'll try to adjust it to use `PhantomReference` instead.

ojwb:
Instead we use java.lang.ref.Cleaner (but will probably switch to
PhantomReference once this is working properly as Cleaner needs
JVM 9 which PhantomReference works with much older JVM versions).

Currently four testcases still fail, and I haven't attempted to
update the documentation at all.

Fixes #1550.

### DIFF
--- a/Examples/test-suite/java/director_pass_by_value_runme.java
+++ b/Examples/test-suite/java/director_pass_by_value_runme.java
@@ -14,7 +14,6 @@ public class director_pass_by_value_runme {
 
   private static void WaitForGC() {
     System.gc();
-    System.runFinalization();
     try {
       java.lang.Thread.sleep(10);
     } catch (java.lang.InterruptedException e) {
--- a/Examples/test-suite/java/java_director_runme.java
+++ b/Examples/test-suite/java/java_director_runme.java
@@ -16,7 +16,6 @@ public class java_director_runme {
   private static void WaitForGC()
   {
     System.gc();
-    System.runFinalization();
     try {
       java.lang.Thread.sleep(10);
     } catch (java.lang.InterruptedException e) {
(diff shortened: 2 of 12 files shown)


aartigao:
I was evaluating to implement this for our SWIG wrappers. This standard solution fits perfect, though. Any insight of when it will be merged to master?

ojwb:
@aartigao 

> I was evaluating to implement this for our SWIG wrappers.

It's not too hard to implement this with current SWIG, though some post-processing of the generated wrapper is required (or at least I couldn't see how to avoid it) - e.g. see https://github.com/xapian/xapian/compare/master...ojwb:java-cleaner for how I'm testing this in Xapian.

That testing hasn't thrown up any real problems with the `Cleaner` approach.  It is still possible to build up a significant backlog of garbage objects in some situations, but it's better behaved than the finalizer-based approach is.

> This standard solution fits perfect, though. Any insight of when it will be merged to master?

See the discussion in #1550 - I think it's really up to @wsfulton at this point.

ojwb:
I'm somewhat reluctantly going to close this as it's been left to bitrot for long enough that there are now conflicts, and we have an open issue for this (#1550).

PR #1570, opened 2019-06-27
Fix class name for nested classes

IsaacPascual:
The feature %interface_impl from swiginterface.i differ on generated function
name from JNI wrapper class when using nested classes without flatnested.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -324,6 +324,7 @@ CPP_TEST_CASES += \
 	nested_directors \
 	nested_comment \
 	nested_ignore \
+	nested_inheritance_interface \
 	nested_in_template \
 	nested_scope \
 	nested_template_base \
--- a/Examples/test-suite/csharp/nested_inheritance_interface_runme.cs
+++ b/Examples/test-suite/csharp/nested_inheritance_interface_runme.cs
@@ -1,0 +1,32 @@
+using System;
+using System.Collections.Generic;
+using nested_inheritance_interfaceNamespace;
+
+public class nested_inheritance_interface_runme {
+
+  static string SortArrayToString(string[] types) {
+    Array.Sort<string>(types);
+    return string.Join(" ", types);
+  }
+
+  static string SortArrayToString(Type[] types) {
+    List<string> stypes = new List<string>();
+    foreach (Type t in types)
+      stypes.Add(t.Name);
+    return SortArrayToString(stypes.ToArray());
+  }
+
+  public static void Main() {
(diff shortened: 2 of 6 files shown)


IsaacPascual:
Sorry for brach name I dont see how to change it.

wsfulton:
Can you give a full example of what this fixes please? Also a new testcase would be much appreciated, perhaps copy and modify one of the Examples/test-suite/multiple_inheritance* testcases.

IsaacPascual:
Well it fail in a big-complex project. Im trying to reproduce in a small project but is more difficult than I think.

IsaacPascual:
My problem is that code on ModuleJNI.java:
`
  public final static native long pack_B_I_GetInterfaceCPtr(long jarg1);
`
And on class B ( nested of A ) has not equal name for functions:
`
public long I_GetInterfaceCPtr() {
    return pack.ModuleJNI.pack_A_B_I_GetInterfaceCPtr(swigCPtr);
}
`

All the calls to Swig_name_member go with getClassPrefix() I try and it solve my problem.

IsaacPascual:
This is a basic case:
[example.tar.gz](https://github.com/swig/swig/files/3339053/example.tar.gz)

Module.h
```cpp
class Interface
{
};

class Class
{
public:
	class Nested : public Interface
	{
	};

};
```

Module.i
```
%module Module

%include <swiginterface.i>

%interface_impl(Interface);

%include "Module.h"
```

Makefile
```makefile
all:
	@mkdir -p pack
	swig -c++ -java -package pack -outdir pack/ Module.i

test:
	@grep -r --color "Nested_Interface_GetInterfaceCPtr" pack/

clean:
	rm -rf pack Module_wrap.cxx
```

IsaacPascual:
The problem is that the method on `pack/ModuleJNI.java`:
```java
public final static native long Nested_Interface_GetInterfaceCPtr(long jarg1);
```
and the use generated on `pack/Class.java`:
```java
return ModuleJNI.Class_Nested_Interface_GetInterfaceCPtr(swigCPtr);
```
dont have the same name!

You can see the problem with:
`make; make test`

IsaacPascual:
I try the test on `master` and fail but not in `patch-1`

IsaacPascual:
Ups some problem with csharp... but I only want this test for java... Some help please my first time writing tests for swig...

IsaacPascual:
Well! I fix it for `csharp` and implement a test case. It gonna be merged? or there is something more to do??

wsfulton:
Looks good to me, thanks for adding in the test-case. I fixed a minor C# compiler warning, see 20d96531e3a5fef442a13dee55020d53a452d7f8.

PR #1573, opened 2019-07-04
Python: Unify string quoting

cher-nov:
**Before:**

BEFORE

**After:**

AFTER

This is a pretty common notation in Python 3: https://stackoverflow.com/a/56190

Actually it could be nice to use this notation everywhere in Python generator, so if the SWIG maintainers agree, I would expand my PR with it, or make an additional one.

### DIFF
--- a/Examples/python/import_packages/namespace_pkg/normal.py
+++ b/Examples/python/import_packages/namespace_pkg/normal.py
@@ -11,7 +11,7 @@ def run_except_on_windows(commandline, env=None):
 print(" Starting subtest " + os.path.basename(__file__))
 
 # Package brave found under one path
-sys.path.insert(0, 'path1')
+sys.path.insert(0, "path1")
 
 from brave import robin
 print("  Finished from brave import robin")
--- a/Examples/python/import_packages/namespace_pkg/nstest.py
+++ b/Examples/python/import_packages/namespace_pkg/nstest.py
@@ -11,7 +11,7 @@ def run_except_on_windows(commandline, env=None):
         print("  Finished running: " + commandline)
 
 def copyMods():
-    dirs = ['path1', 'path2', 'path3']
+    dirs = ["path1", "path2", "path3"]
 
     # Clean out any old package paths
     for d in dirs:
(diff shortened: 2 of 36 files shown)


vadz:
I'm not really sure if this has any real benefit (the convention described in the linked SO answer seems reasonable, but I don't think it's universal), but OTOH I don't see any reason to not apply this neither.

If we do want to standardize on using single quotes in SWIG-generated code, we probably should have a comment somewhere in `Python.cxx` explaining this.

cher-nov:
> I'm not really sure if this has any real benefit (the convention described in the linked SO answer seems reasonable, but I don't think it's universal), but OTOH I don't see any reason to not apply this neither.

This is where syntax highlighting stumbles.

vadz:
> This is where syntax highlighting stumbles.

Strange, why should syntax highlighting have problems with double-quoted strings? Or, at least, more problems than with single-quoted ones?

cher-nov:
> Strange, why should syntax highlighting have problems with double-quoted strings? Or, at least, more problems than with single-quoted ones?

This is not an error, it just highligth it with the same color as natural language strings when we use different colors for double-quoted and single-quoted strings, according to the aforementioned notation. :)

wsfulton:
I took a look at a few files and it looks like SWIG mostly generates double quotes. Also after running the test-suite, I get:
```
$ grep '"' Examples/test-suite/python/*.py | wc -l
8516
$ grep "'" Examples/test-suite/python/*.py | wc -l
3782
```
Quotes on one line ought to be the same style, so happy to have a patch that fixes that. PEP8 says
> In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it. 

My vote is for double quotes given that seems to be mostly what is used already and I unapologetically prefer it given it is what C/C++ programmers use and whom SWIG is targetting and we need a reason to choose one over the other that isn't going to start a flame war.

I suggest using a linter to check the patch is good. Maybe https://github.com/zheller/flake8-quotes.

cher-nov:
> My vote is for double quotes given that seems to be mostly what is used already and I unapologetically prefer it given it is what C/C++ programmers use and whom SWIG is targetting and we need a reason to choose one over the other that isn't going to start a flame war.

Maybe add an option for the Python generator? Run-time or at least compile-time.

wsfulton:
We have option bloat in SWIG which is causing various maintenance headaches and we recently been trying to cut down number of options. Adding this as an option doesn't really add any value, so I say let's have just one quote style.

cher-nov:
@wsfulton I've updated the PR, please review.

wsfulton:
Looks good, I'll merge this in. Did you use any linters to check the output? Would be nice to also have a linter tool to check we keep consistent quotes in our CI builds.

Issue #616, opened 2016-02-17
C# bool array does not seem to work correctly


gpetrou:
Doing some tests with int and bool arrays it seems that int arrays work well, but bool do not.
The test method does not seem to copy the result correctly.

C#
`

```
     int[] intSource = { 1, 2, 3 };
     int[] intTarget = new int[intSource.Length];

     bool[] boolSource = { true, true, false, true, true, false, true };
     bool[] boolTarget = new bool[boolSource.Length];

     TestClass.myArrayCopyInt(intSource, intTarget, intTarget.Length);
     TestClass.myArrayCopyBool(boolSource, boolTarget, boolTarget.Length);

      foreach (int i in intTarget)
             Console.Write("{0} ", i);

      foreach (bool i in boolTarget)
            Console.Write("{0} ", i);
```

`

c++
`

```
   class TESTClass
    {
    static void myArrayCopy( int* intSource, int* intTarget, int nitems );
    static void myArrayCopy1( bool* boolSource, bool* boolTarget, int nitems );
  }

  void TestClass::myArrayCopy( int* intSource, int* intTarget, int nitems  ) {
      for (int i = 0; i < nitems; i++ ) {
     intTarget[ i ] = intSource[ i ];
   }
  }

  void TestClass::myArrayCopy1( bool* boolSource, bool* boolTarget, int nitems  ){
          for (int i = 0; i < nitems; i++ ) {
            boolTarget[ i ] = boolSource i ];
          }
    }
```

`

interface
`
         %module TestClassWrapper

```
     %{
         TestClass.h"
      %}  

       %include <wchar.i>
       %include <typemaps.i>
        %include <windows.i>
          %include <std_vector.i>
         %include "arrays_csharp.i"

      %apply int INPUT[]  { int* intSource }
       %apply int OUTPUT[] { int* intTarget }

    %apply bool INPUT[]  { bool* boolSource }
    %apply bool OUTPUT[] { bool* boolTarget }

      %include "TestClass.h"  
```

`

myd7349:
TestClass.h
```cpp
class TestClass
{
public:
    static void myArrayCopy(int* intSource, int* intTarget, int nitems);
    static void myArrayCopy1(bool* boolSource, bool* boolTarget, int nitems);
};
```

myd7349:
SWIG's boolean array typemaps is added after issue #229 . It is problematic, however.

geefr:
[...]

PR #1576, opened 2019-07-06
Change C# bool[] typemaps to marshall as 1-byte

geefr:
Default marshalling for bool[] now uses 1-byte entries in the array, to
ensure array contents is as expected in C++.

When running under mono csharp_lib_arrays_bool testcase will fail
due to an apparent bug in mono. Works correctly under Microsoft's
runtime. See https://github.com/mono/mono/issues/15592

### DIFF
--- a/Examples/test-suite/csharp/Makefile.in
+++ b/Examples/test-suite/csharp/Makefile.in
@@ -37,9 +37,6 @@ CPP11_TEST_CASES = \
 	cpp11_shared_ptr_upcast \
 	cpp11_strongly_typed_enumerations_simple \
 
-# bool[] typemaps don't work correctly when running under mono
-FAILING_CPP_TESTS = csharp_lib_arrays_bool
-
 include $(srcdir)/../common.mk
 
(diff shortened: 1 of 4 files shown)


wsfulton:
Note: fixes #616.

@geefr, could you please make a change to detect the Mono .net framework at runtime and then skip the test? https://stackoverflow.com/questions/721161/how-to-detect-which-net-runtime-is-being-used-ms-vs-mono looks useful. When it gets fixed in Mono, we can then later add in testing on Mono (need to detect the mono .net framework version number that it was fixed in and later at runtime).

geefr:
Sure, should have it done next week. Didn't realise the ci builds tested against microsoft's runtime aswell.

wsfulton:
Looks good, thanks. I can see the "Tests are running on mono, failing bool[] tests skipped" message only on the Travis Linux and MacOS tests, but not on the Appveyor Windows environment. As this is working as expected, I will now remove the message for the merge to master as we keep the tests quiet.

PR #1580, opened 2019-07-10
Fix nested class name

IsaacPascual:
When merge nested classes and interfaces via `%interface` from `<swiginterface.i>` the generated JNI function names differ from definition and use making java and csharp generated code unusable.

More examples #1570 (I change branch name).


wsfulton:
Github closed this ticket after I merged #1570. Pull requests are identical. This one does have a better branch name, I guess I should have used this pull request instead in hindsight. Unfortunately I pulled the patch before the branch got renamed.

Issue #1568, opened 2019-06-25
C# `set` code generation does not support `int`


nickwinder:
C# `set` code generation does not support int, or any type other than `string`. 

# Details
When trying to define a `std::set` with type of `int` the C# generated is incorrect because the `Add` method is hardcoded with type of `string` as seen 
https://github.com/swig/swig/blob/d3ae85ee734ec701cea0f67f0ccf3d187199819c/Lib/csharp/std_set.i#L33

When changing this method to support `int` or any other non-nullable type further issues are produced. This is because the type is defined as non-nullable. 
https://github.com/swig/swig/blob/d3ae85ee734ec701cea0f67f0ccf3d187199819c/Lib/csharp/std_set.i#L195
but set to null in 
https://github.com/swig/swig/blob/d3ae85ee734ec701cea0f67f0ccf3d187199819c/Lib/csharp/std_set.i#L202

Further to this when returning the type from the `Current` method it should be cast after a null check is done. 
https://github.com/swig/swig/blob/d3ae85ee734ec701cea0f67f0ccf3d187199819c/Lib/csharp/std_set.i#L215

# Reproduce

For simple reproduction please see example repo.
https://github.com/nickwinder/swig-c-sharp-set-bug

### Interface
```
%module Set
%{
#include "cSharpIntSet.hpp"
%}

%include "std_set.i"

%include "cSharpIntSet.hpp"

namespace std {
    %template(set_int) set<int>;
};

```

### Header
```
#pragma once

#include <set>

class cSharpIntSet {
  std::set<int> getSet();
};
```

### Source
```
#include "cSharpIntSet.hpp"

std::set<int> cSharpIntSet::getSet() { return std::set<int>(); }
```

vadz:
As you can probably see, this typemap was written as a generalization of custom typemaps for `std::set<std::string>` and it hasn't been fully generalized to work with the other types, sorry about this.

I'll try to fix it, but I'm not sure how to deal with non-nullable types, would you have any ideas about this? I guess the best would be to use `T?` for `currentObject`, but I don't think we have any typemap that would allow doing this...

nickwinder:
> sorry about this.
All good, no need to apologize. 

> I guess the best would be to use T? for currentObject

For a simple test I did the something similar, so I believe line 195
```diff
- private $typemap(cstype, T) currentObject;
+ private $typemap(cstype, T)? currentObject; 
```
and line 215
```diff
- return currentObject; 
[...]

PR #1583, opened 2019-07-11
Fix C# and Java std::set<> typemaps for primitive types

vadz:
This takes care of the issues discussed in #1568.

The changes here are pretty simple and minimal.


wsfulton:
Looks, good, thanks.

PR #1584, opened 2019-07-11
Fix missing value for first item of enums with trailing comma

### LINKED ISSUE #1566: Swig generates wrong Java (type unsafe) enum for C++ enum with trailing comma
E.g. for
```C++
enum RatioLimitMode
{
    GlobalRatioLimit,
    SingleRatioLimit,
    UnlimitedRatio,
};
```
following Java code will be generated:
```Java
public final static class RatioLimitMode {
    public final static int GlobalRatioLimit = ;
    public final static int SingleRatioLimit = GlobalRatioLimit + 1;
    public final static int UnlimitedRatio = SingleRatioLimit + 1;
}
```
GlobalRatioLimit has no value.
If you remove trailing comma, it is correctly set to 0. However C++ allows trailing commas in enums since C++11.

vadz:
The value of the first item of an enum with a trailing comma after its
last item was not correctly initialized to 0 any more after the changes
of 74adaa5738368ce473870451422f4a7175b6a736 (see #1515) because "_last"
attribute was not set correctly in this case.

Do set it for the last item when it's followed by a comma too and add
more unit tests checking for this.

Closes #1566.

### DIFF
--- a/Examples/test-suite/java/java_enums_runme.java
+++ b/Examples/test-suite/java/java_enums_runme.java
@@ -13,6 +13,12 @@ public class java_enums_runme implements stuff {
 
   public static void main(String argv[]) 
   {
+      if (WithTrailingComma.Second != 1)
+          throw new RuntimeException("Incorrect value for Second");
+
+      if (WithTrailingCommaAndIgnoredFirstItem.SecondNonIgnoredOne != 2)
+          throw new RuntimeException("Incorrect value for SecondNonIgnoredOne");
+
       int number = 200;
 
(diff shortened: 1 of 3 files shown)


wsfulton:
I've restarted the Appveyor PY3 test that failed, assuming it was an infrastructure problem. The python_docstring test failed to finish.

wsfulton:
I ran valgrind over the python_docstring testcase (both swig and python execution) with the patch and couldn't see any problems, so most likely an Appveyor transient problem.

Issue #1586, opened 2019-07-12
Python 3.8 removed _PyObject_GC_UNTRACK


gicmo:
See fedora bug [rhbz#1705450](https://bugzilla.redhat.com/show_bug.cgi?id=1705450), where renderdoc fails to compile with Python 3.8 beta 2 because the generated code is using [`_PyObject_GC_UNTRACK`](https://docs.python.org/3/c-api/gcsupport.html#c._PyObject_GC_UNTRACK), which is deprecated since 3.6 and got remove in 3.8.

wsfulton:
Fixed now, see #1587.

PR #1587, opened 2019-07-12
Use PyObject_GC_UnTrack for python >= 3.6

gicmo:
The [PyObject_GC_UnTrack]( https://docs.python.org/3/c-api/gcsupport.html#c._PyObject_GC_UNTRACK) macro got deprecated in 3.6 and finally removed in 3.8. Thus for all python versions >= 3.6 instead used `PyObject_GC_UnTrack`.

Closes issue #1586

### DIFF
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -117,7 +117,7 @@ SwigPyBuiltin_FunpackSetterClosure (PyObject *obj, PyObject *val, void *closure)
 
 SWIGINTERN void
 SwigPyStaticVar_dealloc(PyDescrObject *descr) {
-  _PyObject_GC_UNTRACK(descr);
+  PyObject_GC_UnTrack(descr);
   Py_XDECREF(PyDescr_TYPE(descr));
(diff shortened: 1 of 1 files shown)


Lib/python/builtin.swg
```diff
@@ -117,7 +117,11 @@ SwigPyBuiltin_FunpackSetterClosure (PyObject *obj, PyObject *val, void *closure)
 
 SWIGINTERN void
 SwigPyStaticVar_dealloc(PyDescrObject *descr) {
+#if PY_VERSION_HEX >= 0x03060000
+  PyObject_GC_UnTrack(descr);
+#else
   _PyObject_GC_UNTRACK(descr);
```
wsfulton:
```suggestion
  PyObject_GC_UnTrack(descr);
```
No need for the PY_VERSION_HEX check as we support 2.7, 3.2+ and PyObect_GC_UnTrack is available in 2.7 onwards. Please resubmit the pull request with this change to make sure the failing CI tests are fixed. Thanks.

PR #1596, opened 2019-07-20
Lua: userdata tostring should point to wrapped memory

sjml:
This is actually a fix to some functionality that I submitted many years ago. :) At the time I set the string conversion to output the userdata address, but since that points to an internal SWIG structure, it's way more useful to the user to point to the actual memory being wrapped in that userdata.

### DIFF
--- a/Lib/lua/luarun.swg
+++ b/Lib/lua/luarun.swg
@@ -1040,7 +1040,7 @@ SWIGINTERN int  SWIG_Lua_class_tostring(lua_State *L)
 /*  there should be 1 param passed in
   (1) userdata (not the metatable) */
   const char *className;
-  void* userData;
+  swig_lua_userdata* userData;
   assert(lua_isuserdata(L,1));  /* just in case */
   userData = lua_touserdata(L,1); /* get the userdata address for later */
   lua_getmetatable(L,1);    /* get the meta table */
@@ -1049,7 +1049,7 @@ SWIGINTERN int  SWIG_Lua_class_tostring(lua_State *L)
   lua_getfield(L, -1, ".type");
   className = lua_tostring(L, -1);
 
-  lua_pushfstring(L, "<%s userdata: %p>", className, userData);
+  lua_pushfstring(L, "<%s userdata: %p>", className, userData->ptr);
   return 1;
 }
 
--- a/Lib/lua/luarun.swg
+++ b/Lib/lua/luarun.swg
@@ -1049,7 +1049,7 @@ SWIGINTERN int  SWIG_Lua_class_tostring(lua_State *L)
   lua_getfield(L, -1, ".type");
   className = lua_tostring(L, -1);
 
-  lua_pushfstring(L, "<%s userdata: %p>", className, userData->ptr);
+  lua_pushfstring(L, "<%s userdata: %p::%p>", className, userdata, userData->ptr);
   return 1;
 }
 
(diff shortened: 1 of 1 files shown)


wsfulton:
Hi @sjml. We're short of Lua developers nowadays, so can you go through a few basics for me please to accept this patch. Can you please show a simple interface file, simple script and the output comparing it to the previous output. Maybe it is easier to just add a line to one of the examples (Examples/lua/class) would be a good one.

What is the difference in the two memory areas? I'm pretty sure other languages show the pointer to the underlying C/C++ struct.

Surely the output should be similar to one of the other scripting languages. For example, the same class example for Python shows one of the objects as follows:
```
<example.Circle; proxy of <Swig Object of type 'Circle *' at 0x7fb31f5d2540> >
```

sjml:
> Hi @sjml. We're short of Lua developers nowadays, so can you go through a few basics for me please to accept this patch. Can you please show a simple interface file, simple script and the output comparing it to the previous output. Maybe it is easier to just add a line to one of the examples (Examples/lua/class) would be a good one.

Sure, I can take a stab at that. I'll comment here again when I've done so. 

> What is the difference in the two memory areas? I'm pretty sure other languages show the pointer to the underlying C/C++ struct.

Without this patch, it's printing the memory address of the Lua object which wraps the C/C++ pointer as userdata. While that's not useless, it's obviously way more useful to know the address of the underlying memory. 

> Surely the output should be similar to one of the other scripting languages. For example, the same class example for Python shows one of the objects as follows:
> 
> ```
> <example.Circle; proxy of <Swig Object of type 'Circle *' at 0x7fb31f5d2540> >
> ```

This is a little more verbose than the Lua wrapping I've written, which I tried to make match the usual Lua output. For example, a function variable simply outputs `function: 0x7fdb6240af20`, and a table (Lua structure) `table: 0x7fdb627008d0`. I'm open to the more verbose output, though, if you would prefer. Obviously, changing a `printf` string is pretty trivial.

sjml:
Oh, actually, Examples/lua/class already tests this functionality. Before the patch, the first few lines printed say: 

```
Creating some objects:
    Created circle	<Circle userdata: 0x7f9e31c08dc8>
    Created square	<Square userdata: 0x7f9e31c09848>
```

Notably, that memory address is pointing to the Lua object. 

After the latest update to this PR, they say:
```
Creating some objects:
    Created circle      <userdata of type 'Circle *' at 0x7fdc0f409560>
    Created square      <userdata of type 'Square *' at 0x7fdc0f409900>
```

With those memory addresses pointing to the underlying structure. 

How does that look to you, @wsfulton?

vadz:
FWIW this is definitely something that would be helpful when debugging C++/Lua code (I mean showing the C++ object address, the format of the string doesn't really matter).

Thanks for fixing this!

sjml:
Looks like the travis failure is some unrelated thing with PHP, by the way.

wsfulton:
Don't worry about the php failure, I've restarted the failed php job, so it will probably go green shortly.

wsfulton:
It certainly seems more useful to have the underlying C/C++ struct/class pointer being displayed like other languages. Whether this conforms to Lua conventions/expectations isn't entirely clear to me. I'll merge it in on the weekend if there are no further objections.

sjml:
> It certainly seems more useful to have the underlying C/C++ struct/class pointer being displayed like other languages. Whether this conforms to Lua conventions/expectations isn't entirely clear to me. I'll merge it in on the weekend if there are no further objections.

I think it should be ok... there are astonishingly few hard conventions with Lua, so I think SWIG should feel free to create its own. I was only referring to the rather terse output the standard runtime gives when printing, say, a function value: 

```lua
f = function() end
print(f) --> function: 0x7fe07bd07c90
```

But since these objects are a little more complicated, a more verbose output seems warranted. That it brings Lua more in line with other SWIG wrappings seals the deal. 

Thanks for merging!

PR #1601, opened 2019-07-28
Fix invalid code generated for "%constant enum EnumType

rokups:
Previously it generated code like:

```cs
public static readonly LightType DefaultLighttype = (LightType).DefaultLighttype_get();
```


wsfulton:
Code change in SWIG sources looks okay. Can you please fix the testcase to be valid C++ so that the Travis test-suite passes. Thanks.

rokups:
Done. All tests pass.

wsfulton:
Thanks

PR #1604, opened 2019-07-30
Correct %init documentation for C#/Java

vadz:
For these languages, %init doesn't inject the code into the
initialization function (because there is none), but just puts it into
the global scope instead.

[skip ci]

---

I initially thought to fix this by generating the wrapper struct automatically, but realized that it would break any existing code. I don't see any backwards compatible way to fix this, so I decided to at least document it.

### DIFF
--- a/Doc/Manual/SWIG.html
+++ b/Doc/Manual/SWIG.html
@@ -3438,6 +3438,18 @@ initialization on module loading, you could write this:
 %}
 </pre></div>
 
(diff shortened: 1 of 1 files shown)


wsfulton:
Yes, probably best not try to retrofit this now. For Java, the code could have been put in JNI_OnLoad, but users probably have this coded up already. (I certainly have done this).

PR #1608, opened 2019-08-07
Doxygen param and code option handling for Python

mcfarljm:
Various improvements to how doxygen ```\param``` and ```\code``` commands are handled:
- Recognize ```\param[in]```, ```[out]```, and ```[in,out]```, and include it as a flag in generated pydoc output
- Flag optional arguments in generated pydoc output
- Recognize language option to ```\code``` command and add to generated pydoc ```code-block``` output (currently supports Python, Java, C, and C++, defaulting to C++ as before)
- Special handling for python doctest code blocks
- Improved handling of ```\code``` content to support additional characters

### DIFF
--- a/Doc/Manual/Doxygen.html
+++ b/Doc/Manual/Doxygen.html
@@ -829,155 +829,107 @@ These are suppressed with their content just printed out (if the tag has any
 sense, typically text content).
 Here is the list of these tags:
 </p>
+
 <div class="diagram">
-<table border="0" summary="Unsupported Java Doxygen Tags">
-<tr>
-  <th align="left">Unsupported Doxygen tags</th>
-</tr>
-<tr>
-<td>\addindex</td>
-<td>\addtogroup</td>
-<td>\anchor</td>
-<td>\attention</td>
-</tr>
-<tr>
-<td>\brief</td>
-<td>\bug</td>
-<td>\callgraph</td>
-<td>\callergraph</td>
-</tr>
-<tr>
-<td>\class</td>
-<td>\copybrief</td>
-<td>\copydetails</td>
-<td>\copydoc</td>
-</tr>
-<tr>
-<td>\date</td>
-<td>\def</td>
-<td>\defgroup</td>
-<td>\details</td>
-</tr>
-<tr>
-<td>\dir</td>
-<td>\dontinclude</td>
-<td>\dot</td>
-<td>\dotfile</td>
-</tr>
-<tr>
-<td>\enddot</td>
-<td>\endhtmlonly</td>
-<td>\endinternal</td>
-<td>\endlatexonly</td>
-</tr>
-<tr>
-<td>\endmanonly</td>
-<td>\endmsc</td>
-<td>\endrtfonly</td>
-<td>\endxmlonly</td>
-</tr>
-<tr>
-<td>\enum</td>
-<td>\example</td>
-<td>\extends</td>
-</tr>
-<tr>
-<td>\file</td>
-<td>\fn</td>
-<td>\headerfile</td>
-<td>\hideinitializer</td>
-</tr>
-<tr>
-<td>\htmlinclude</td>
-<td>\htmlonly</td>
-<td>\implements</td>
-<td>\include</td>
-</tr>
-<tr>
-<td>\includelineno</td>
-<td>\ingroup</td>
-<td>\internal</td>
-<td>\invariant</td>
-</tr>
-<tr>
-<td>\interface</td>
-<td>\latexonly</td>
-<td>\line</td>
-<td>\mainpage</td>
-</tr>
-<tr>
-<td>\manonly</td>
-<td>\memberof</td>
-<td>\msc</td>
-<td>\mscfile</td>
-</tr>
-<tr>
-<td>\name</td>
-<td>\namespace</td>
-<td>\nosubgrouping</td>
-<td>\package</td>
-</tr>
-<tr>
-<td>\page</td>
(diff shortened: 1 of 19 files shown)


wsfulton:
@mcfarljm this looks great for Python, so happy to have this. However, I'm not inclined to accept anything where Doxygen command support diverges between the two supported languages, Python and Java. Can you please make similar additions to javadoc.cxx that you put in pydoc.cxx? I appreciate that there may be some additional escaping issues to deal with, but if you need any assistance, mention it in this issue.

mcfarljm:
@wsfulton It depends on what kind of support you have in mind for Java.  Changes were made to ```doxyparser.cxx``` so that the command options are recognized regardless of which language is being used.  For example this doxygen construct:
```
 * \param[in] b parameter with intent(in)
```
Was previously not recognized and produced this javadoc:
```
b parameter with intent(in)
```
but is now:
```
@param b parameter with intent(in)
```

Similarly for code blocks, this:
```
 * \code{.py}
 * print('hello')
 * \endcode
```
was not recognized and was translated as:
```
 <br>
 .py}<br>
 print('hello')
```
but is now:
```
 <br>
 {@code 
  print('hello')
  }
```

So this update supports both of these commands for Java, but the difference is that the additional information (parameter direction/intent and code language) is ignored when writing the javadoc, whereas it is utilized in the pydoc.  The pydoc format has a separate field for parameter type, so that seemed like a natural place to include the parameter direction in the comments.  There is no corresponding parameter type section in the javadoc, so it's unclear where the intent information should go.  Same story for code language: the sphinx code block construct used in the pydoc has a place for the code language, but as far as I know javadoc does not.  Similar situation for optional arguments, since I added the flag as part of the parameter type description in pydoc.  Unclear where it should go in Java.  It's probably less important for Java anyway due to the way optional arguments are handled via overloading.

One divergence is that the new test ```doxygen_code_blocks.i``` doesn't have a "runme" file for Java.  I created it mainly to test the code language options, which get ignored by Java.  It also has a code block with characters such as angle brackets and newlines -- looking at it now this seems to get handled correctly in Java.  So we could add ```doxygen_code_blcoks_runme.java``` if you want.  And if anyone has some specification for how they would like parameter intent, code language, or optional argument flags to be included in the javadoc I'm happy to incorporate that.

wsfulton:
Thanks, I've got a clearer picture now with your further explanations. I agree that there is no obvious equivalent in Javadoc for the optional parameter direction in `\param` and the optional language in `\code`. Effectively ignoring these is about all that can be done. 

I think a bit more documentation for these enhancements in Doxygen.html is needed, such as the ignoring just mentioned and updating the list of (un)supported tags to include `\code` and `\endcode` for Python.

I'd like to see equivalent runme files for both Java and Python, so please add in the missing doxygen_code_blocks_runme.java. Even if it ignores the optional information, we need a regression test to make sure works as designed / documented.

wsfulton:
Very nice. Just one thing, the format of the Unsupported tags table and Supported tags table is different now. The Unsupported tags has a dot list item for each tag which seems unnecessary and inconsistent to the Supported tags table. I'm afraid the wkhtmltopdf tool we use to translate html to pdf for the pdf documentation does not support: `style="column-count:4;"`. Not a huge deal, but it creates one long list instead of 4 columns. In summary, can you please remove the list items dots and if possible avoid html style elements that wkhtmltopdf does not support.

mcfarljm:
I force-pushed an update to remove the bullets from the tag lists using:
```
<ul style="list-style-type:none;column-count:4;">
```

In terms of `column-count` not being supported for PDF creation, I'm not sure what better options there are.  The previous approach (which I wasn't able to install the patched wkhtmltopdf myself to test this with PDF conversion) used manual table formatting that in my opinion made it impractical to maintain the lists -- as evidenced by the fact that the lists previously were only partially in alphabetical order and contained duplicate entries.  The manual table formatting also set the items by row instead of by column, which I find harder to read for what is really just a single list.

I suggest keeping the current format assuming it doesn't break the PDF rendering and at least produces a valid (although not ideally formatted) list.  I think that's preferable to using a table that is very difficult to change and decidedly less natural to read.  The next best option may just be to list them out in paragraph form.

gieseanw:
I'm really hoping to get the support for `@param[in]` as soon as possible. Any way to break that functionality into its own PR and merge it?

mcfarljm:
@wsfulton I wanted to check back in on this pull request and see if we can get it merged, otherwise it may go stale and get forgotten.  Based on the above discussion, the only question was about formatting of the supported/unsupported tags tables.  As it currently stands, I have kept them as lists and left in the `column-count` tag, with the thinking being that even though `column-count` is not supported by wkhtmltopdf, maybe having it render as a single long list for the PDF is acceptable.  Alternatively, if you prefer, we could just list the items out in paragraph form.  Either way, my feeling is that the previous approach of using manually formatted table columns is not maintainable because inserting an entry requires reformatting the whole table (and I believe this is the reason the current version has out of order and duplicate entries).

wsfulton:
The pdf document generated ignores the `column-count`, I guess that's not the end of the world, it still displays all the tags but just in one long list instead. Just to let you know I'm looking to merge this shortly for swig-4.0.2.

gbala88:
> I'm really hoping to get the support for `@param[in]` as soon as possible. Any way to break that functionality into its own PR and merge it?

Hi, I'm trying to translate the doxygen having `@param[in]` in the C code to the corresponding python version. I hit an error stating its an unknown doxygen command. Any chance there is a working example for the same?

Issue #1611, opened 2019-08-12
--without arguments don't disable language specific examples/features


ngie-eign:
A number of specific languages have hooks, e.g., `--without-tcl` for Tcl bindings/examples. Specifying the flags should control doesn't whether or not a feature and its examples are installed, but it doesn't currently per [this piece of Makefile.in](https://github.com/swig/swig/blob/master/Makefile.in#L438) (the issue is more widespread beyond this point; this is just an example).

ojwb:
As documented at http://swig.org/Doc4.0/Preface.html these configure options only control which languages the testsuite gets run for:

> Note that the --without-xxx options, where xxx is a target language, have minimal effect. All they do is reduce the amount of testing done with 'make check'. The SWIG executable and library files installed cannot currently be configured with a subset of target languages.

wsfulton:
Indeed, closing. Also see comments in #1612.

PR #1612, opened 2019-08-13
Conditionally install language bindings with `install-lib` target

ngie-eign:
Prior to this change, all library bindings would be installed with
`install-lib`. This is not desired with swig installs that want to
distribute a more stripped down version of the swig package.

Conditionally install the files based on the command line arguments
provided to configure, e.g., honor `--without-tcl` for Tcl, etc.

Closes: #1611
Signed-off-by: Enji Cooper <enji.cooper@dell.com>

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -435,11 +435,83 @@ install-main:
 	@echo "Installing $(DESTDIR)$(BIN_DIR)/`echo $(TARGET_NOEXE) | sed '$(transform)'`@EXEEXT@"
 	@$(INSTALL_PROGRAM) $(TARGET) $(DESTDIR)$(BIN_DIR)/`echo $(TARGET_NOEXE) | sed '$(transform)'`@EXEEXT@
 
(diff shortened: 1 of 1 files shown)


wsfulton:
The `--without-<lang>` options force skips the languages, but it isn't the only way that the `@SKIP_<LANG>@` flags are set. This change will unintentionally skip languages from being installed because the `@SKIP_<LANG>@` flags are set if the given language is also not found nor appears to be correctly configured on the machine building SWIG. These flags are only really intended for running 'make check'. An implementation to conditionally install languages should be absolutely explicit from command-line options given and not have any side effect of the state of the environment/machine that built SWIG.

When a user attempts to use a language (that was not installed), a cryptic error message is seen. So this patch only partially implements a conditional install. More work would need to be done,  disabling the language specific code so that a user-friendly error appears if an attempt is made to use a disabled language. I would also expect to see a new swig executable option included that displays how the configuration was changed from the default.

wsfulton:
Closing as I don't think this is going anywhere. If all the comments are addressed, feel free to re-open.

PR #1613, opened 2019-08-13
Support stable Python ABI

vadimcn:


wsfulton:
Were you expecting any tests to pass? Looks like every test fails with ` # error Oh noes!`

vadimcn:
> Were you expecting any tests to pass? 

I am not expecting this PR to be merged as-is, I created for people on the thread to comment on proposed changes.    So far there were no takers... Maybe you'd care to take a look? :)

The error is injected on purpose, so I could be sure I was really testing with the `-py3-stable-abi` switch. 
A big piece that's missing is support for `-builtin`.   I am not sure I fully understand yet what's going on in that code.

aundro:
Hi all,

we (www.hex-rays.com) are long-time users of SWiG, and we will soon have to ship our product with support for Python 3.
In an attempt to ship our product with code that doesn't expect a specific version of Python to be installed (we use the one from the system), we would be interested in getting this Py_LIMITED_API-related work, into SWiG.

I can see that the work on this branch is not finished, but at a first glance it seems like a good start.
Can anyone elaborate on what still needs to be done, and perhaps pointers to get me started?
(I don't suppose the PyBuffer-related bits will find a solution any time soon, since the relevant CPython functions are (AFAICT) still not part of the "limited API". 
But for the rest, perhaps I can give a hand?)

vadimcn:
@wsfulton, @vadz: ping

vadz:
Sorry, I still didn't have time to test this in my real projects using SWIG, which I'd like to do. But just looking at the changes they do seem fine, thanks.

aundro:
I think there is still an issue, on windows.
SWiG generates the following code:

  public:
    Guard(PyThread_type_lock & mutex) : mutex_(mutex) {
      PyThread_acquire_lock(mutex_, WAIT_LOCK);
    }

    ~Guard() {
      PyThread_release_lock(mutex_);
    }
  };

and

    PyThread_type_lock Director::swig_mutex_own = PyThread_allocate_lock();

but `python3.lib` (and `python3.dll`) on windows doesn't export those three `PyThread_` functions.
I don't really see how to work around this, except maybe having SWiG use its own mutexes?

ojwb:
Are you saying Python no longer provides a public thread C API at all?  Or is this only with `Py_LIMITED_API`?

The SF patch referenced is now at: https://sourceforge.net/p/swig/patches/175/

It was long ago, but I think the key motivation was probably that it was broken on a platform I don't use (but people using the code I generate with SWIG want to use) and it was simpler to fix it with a cross-platform solution.

However it seemed (and still seems) cleaner to use Python's threading abstraction layer rather than heading straight for a platform-specific threading layer.  I'd imagine all POSIX platforms have pthreads by now, but Microsoft of course have their own different APIs for almost everything.  And if there's more than one thread implementation available on a platform I'd rather not have to try to think about what happens if Python and the module use different ones.  I don't know if any current platforms do, but e.g. Linux used to have both LinuxThreads and NPTL.

vadimcn:
If a function isn't in the limited API, build should have failed at the compilation stage.  
Looks like [PyThread_allocate_lock](https://github.com/python/cpython/blob/master/Include/pythread.h#L34) is **not** excluded.   However, I don't see it among [the functions exported from python3.dll](https://github.com/python/cpython/blob/master/PC/python3.def), so this might be a CPython bug.

vadimcn:
I wonder if it needs to be a separate lock, or can we simply reuse the GIL?

ojwb:
If attempting to use the GIL, beware of the simplified GIL State API (which I think SWIG still uses by default) as this can cause problems (or at least used to - it's possible things have evolved since) - e.g. see https://code.google.com/archive/p/modwsgi/wikis/ApplicationIssues.wiki and search for the "Python Simplified GIL State API" heading.

For Xapian we actually replace SWIG's code with our own version which uses the Python thread API but not the GIL State API (and as a bonus does nothing if `PyEval_ThreadsInitialized()` says threads aren't in use, which according to an old commit message gives a 5-10% speed-up in non-threaded code, though obviously that'll vary a lot depending on how much work each wrapped call does).  I've never got around to trying to clean that up and submit it back to SWIG though.

vadimcn:
@ojwb: the article you referenced indeed suggests that extensions using simple gil interface may deadlock if used in environment with multiple sub interpreters.  However it does not explain the exact mechanism of this happening, so I was unable to tell whether this condition would apply to our case.   
In any event, the article seems to say that this is a general problem with extensions using the simplified gil api (including swig extensions), so adding another use of gil should not make things any worse?

aundro:
@ojwb I should have been more clear, sorry. The function is indeed not excluded from the Py_LIMITED_API, but linking will fail due to it not being exposed in python3.dll

@vadimcn I guess it can be considered a CPython bug. I've found this relevant discussion: https://mail.python.org/pipermail/python-dev/2015-April/139124.html

But, the way I see it, whether or not it is a CPython bug (and ideally will be fixed in the future) the whole idea of using Py_LIMITED_API is that we can ship SWiG-generated wrappers that will run against existing Python 3 installations (provided they are not ancient) as well as future installations.

Therefore, I don't think it's a great course of action to expect CPython developers to fix it, and then use it, since that means SWiG would only be able to support Python 3.9 & later.

Cory-Kramer:
What is the status of this pull request? Looks like the CI builds/tests have passed. Are there still issues being investigated?

ojwb:
> In any event, the article seems to say that this is a general problem with extensions using the simplified gil api (including swig extensions)

It affects SWIG-generated extensions by default, but it's possible (and fairly easy) to override that and get an extension which doesn't use this problematic simplified GIL locking API.

> so adding another use of gil should not make things any worse?

Using the GIL is not a problem in itself.  The problem is using the simplified GIL locking API (`PyGILState_Ensure()`, etc) to do it.  [The python 3 docs](https://docs.python.org/3/c-api/init.html#non-python-created-threads) warn about this exact issue here:

> _Note that the PyGILState_*() functions assume there is only one global interpreter (created automatically by Py_Initialize()). Python supports the creation of additional interpreters (using Py_NewInterpreter()), but mixing multiple interpreters and the PyGILState_*() API is unsupported._

I'd really recommend against adding further uses of the simplified GIL API - I think it's a mistake SWIG is using it at all - but in practical terms what I really care about is that it doesn't become harder to stop SWIG using it in the bindings I maintain.

Currently in our interface file we just need to define `SWIG_PYTHON_NO_USE_GIL`, and also add suitably replacement implementations by defining `SWIG_PYTHON_THREAD_BEGIN_BLOCK`, `SWIG_PYTHON_THREAD_END_BLOCK`, `SWIG_PYTHON_THREAD_BEGIN_ALLOW` and `SWIG_PYTHON_THREAD_END_ALLOW`.

vadimcn:
I don't think anything is blocking this PR.   It may be incompatible with `-threads` switch (only if "directors" feature is enabled, as I understand), but that can be improved later, IMO.

ojwb:
If it's going to be incompatible with `-threads` that really should be documented and attempts to use incompatible combinations rejected with a clear error message (like you already did for `-builtin`).

Generating code which fails to compile or fails to work is just going to generating support load, and we already struggle to keep up with tickets.

ojwb:
You should probably also remove the `[WIP]` marker if you think it's ready.

Cory-Kramer:
Is this pull request ready to be merged?
Thanks again for taking the time to make this update, much appreciated!

vadimcn:
Any ideas how to make AppVeyor produce a more useful error message?

ojwb:
> Any ideas how to make AppVeyor produce a more useful error message?

No, but it also seems to be failing on master currently - see this build from 11 days ago: https://ci.appveyor.com/project/swig/swig/builds/28735588

The exact same commit apparently built OK 18 days ago: https://ci.appveyor.com/project/swig/swig/builds/28555074

So some change in the appveyor environment I guess.

vadz:
> So some change in the appveyor environment I guess.

Yes, I've already [commented on this](https://github.com/swig/swig/pull/1668#issuecomment-554556850). I.e. basically `2to3` is not available, or at least not detected by configure, any longer. I haven't had the time to look at why does this happen though.

vadimcn:
Well, other than that, this PR is ready to be merged, I think.

Cory-Kramer:
Any update on resolving the CI issue? Our team would love for this feature to be delivered.

vadz:
Thanks for the update!

I'd like to test this in our project, but I wonder how do you manage link dependencies under Unix: we take them from `python3-config --ldflags` and this outputs the version-dependent library name, e.g. `libpython3.7m.so.1.0`, so the generated module is still bound to this particular Python version. And there doesn't seem to be any `libpython3.so` at all, so how do you handle this? The only way to allow the same module to work with multiple Python versions that I see right now is to implement runtime linking support (via `dlopen` and friends), did you find any better solution by chance?

vadz:
Unfortunately I have some troubles making this work under MSW too. First of all, `Py_LIMITED_API` must be defined as at least `0x03040000` as otherwise `PyType_GetSlot()` used by the generated code is not available. I think we can live with this (i.e. I don't see how to avoid using it), but the documentation should be upfront about only supporting 3.4+ and having to define `Py_LIMITED_API` appropriately.

Second problem doesn't have much to do with this PR per se, but I'm using typemaps for mapping our own date class to `PyDateTime` and they don't compile at all with limited API. I'll have to fix this somehow before I can really test it fully.

vadimcn:
>  so the generated module is still bound to this particular Python version. And there doesn't seem to be any libpython3.so at all, so how do you handle this? 

Ah, that: you need to build the extension dylib without linking to libpython3.X (except on Windows).  The dynamic linker will still be able to resolve needed symbols as long as there's any suitable Python runtime in the process.
- On Linux make sure that unresolved symbols in the output dylib are not forbidden via `-z,defs`.
- On MacOS, link with `-undefined dynamic_lookup` to achieve the same effect.
- On Windows, link to `python3.dll`.

> First of all, Py_LIMITED_API must be defined as at least 0x03040000

It is defined, [here](https://github.com/swig/swig/pull/1613/files#diff-3cc91cccc4a3a707448c1f5993015108R640).   Are you including Python.h before SWIG-generated code includes it?

vadz:
> > so the generated module is still bound to this particular Python version. And there doesn't seem to be any libpython3.so at all, so how do you handle this?
> 
> Ah, that: you need to build the extension dylib without linking to libpython3.X (except on Windows). 

Yes, of course, thanks, it's definitely better than implementing runtime linking manually, don't know why I didn't think about this. I rather like using `-z defs` and it's a pity that this is not going to be possible any more, but I guess there is nothing to do about it (other than convincing Python developers/packagers to include `libpython3.so` in Python distributions as is done under MSW).

>     * On Linux make sure that unresolved symbols in the output dylib are not forbidden via `-z,defs`.
> 
>     * On MacOS, link with `-undefined dynamic_lookup` to achieve the same effect.
> 
>     * On Windows, link to `python3.dll`.

FWIW I think this should be described in the docs.

> > First of all, Py_LIMITED_API must be defined as at least 0x03040000
> 
> It is defined, [here](https://github.com/swig/swig/pull/1613/files#diff-3cc91cccc4a3a707448c1f5993015108R640). Are you including Python.h before SWIG-generated code includes it?

Oops, yes, this shouldn't be a problem for normal code, you're right. I just misread the documentation and thought I had to define it myself. I still think the discussion of this in the docs is rather confusing, I'd just remove the paragraph speaking about it -- if all this is done internally in the generated code, SWIG users don't really need to care about it.

vadz:
I've finished rewriting my typemaps and can confirm that this works fine under MSW: I've built the module with 3.6.3 and checked that the unit tests pass with 3.6.3 (sic), 3.4.1, 3.7.6 and 3.8.1.

Note that with Python 3.8 you need to explicitly call [add_dll_directory()](https://docs.python.org/3/library/os.html#os.add_dll_directory) if the actual DLL is not in the same directory as the module itself, which cost me quite some time to find out (especially because it wasn't the case for the previous Python versions), so I wonder if it could be worth mentioning this in the docs as well.

But other than the docs updates (that I could do myself if wanted), I think this should be merged because it really helps a lot and I'm going to switch all my projects using SWIG to use `-py3-limited-abi` a.s.a.p.

Thanks again for this PR!

vadimcn:
Okay, I've added a note on linkage.  I think that add_dll_directory() is a Python 3.8 issue, not something specific to stable ABI or even to SWIG.

Doc/Manual/Python.html
```diff
@@ -7281,11 +7281,22 @@ <H3><a name="Python_3_stable_abi">32.12.5 Python 3 Stable ABI</a></H3>
...
+<H4>Linkage</H4>
+<p>
+Version-independent modules have special requirements when it comes to linking: they should not depend on the versioned Python dynamic library, otherwise the dynamic 
+linker will refuse to load them into a different version of Python runtime.  
+Here's <a href="https://www.python.org/dev/peps/pep-0384/#linkage">PEP 384 guidance on linkage</a>.  In practical terms, it mean this:
+<ul>
+  <li>On Windows, link to the version-independent <tt>python3.lib</tt>.</li>
+  <li>On Linux, don't link to <tt>libpython</tt> at all, leaving all imported Python symbols unresolved.  Make sure that unresolved symbols are not disallowed via "<tt>-z defs</tt>" linker option.</li>
+  <li>On MacOS, also don't link to <tt>libpython</tt>, but you will need to explicitly allow unresolved symbols via "<tt>-undefined dynamic_lookup</tt>" option.  
+      If you re not involing linker directly, you may need to instruct your compiler to do so. For example, GCC-compatible complers will need this flag:
```
ojwb: "you're" and "invoking"

Doc/Manual/Python.html
```diff
@@ -7281,11 +7281,22 @@ <H3><a name="Python_3_stable_abi">32.12.5 Python 3 Stable ABI</a></H3>
...
+      "<tt>-Wl,-undefined,dynamic_lookup"</tt>.</li>
```
ojwb: The closing `"` should really be after the `</tt>` here.

Doc/Manual/Python.html
vadimcn: thanks!

vadz:
Sorry, but I'd like to return to the thread/directors issues previously discussed above as more extensive testing showed that it affects me as well. I've opened [an issue](https://bugs.python.org/issue39123) about this on Python bugtracker but I agree that even if they fix it today, it doesn't help us, so we need some kind of workaround and I've created #1687 for it. IMO it should be merged before/together with this one to avoid link-time errors when using directors.

vadz:
FWIW I'd really like to see this one (and #1687) merged. The conflict in the manual is trivial, but I could resolve it if it could help.

vadz:
@vadimcn I don't know if I should push to your branch or create another PR based on it, but I'd like to fix another minor but somewhat annoying issue: compiling SWIG-generated code now results in warnings about unused `SwigMethods_proxydocs`, which are indeed correct because it's not used when `Py_LIMITED_API` is defined. We can either take its definition inside `#ifndef Py_LIMITED_API` check too (as references to it are inside such checks now) or even just omit generating it entirely when `py3_stable_abi` is on. The former would probably be more compatible with the way things are currently done.

vadimcn:
@vadz: Sure, either way works for me.

okainov:
Looking forward to have it merged and working...

chris-decker08:
This would be of a GREAT benefit to me.  How can I assist in getting this merged?

jschueller:
hi, @vadimcn, could you rebase on top of master ? Files are conflicting.
I'd like to test your branch.

vadimcn:
@jschueller: Looks like only the documentation html is in conflict.  This should not prevent you from trying out this change.

vadimcn:
@vadz: are you still working to land this PR?

vadz:
Sorry, but unfortunately I can't really do much. I could rebase it and solve the minor issue mentioned in https://github.com/swig/swig/pull/1613#issuecomment-587211571 but normally only @wsfulton merges PRs and he hasn't shown any sign of life since quite some time (it would be reassuring to hear something just to know that you're ok, William!).

FWIW I'm using this in a couple of my projects and am very happy with it, thanks again for your changes!

Lib/python/pyinit.swg
```diff
@@ -200,12 +203,29 @@ swig_varlink_type(void) {
...
+  PyType_Slot slots[] = {
+    { Py_tp_dealloc, (void*)swig_varlink_dealloc },
+    { Py_tp_getattr, (void*)swig_varlink_getattr },
+    { Py_tp_setattr, (void*)swig_varlink_setattr },
+    { Py_tp_repr, (void*)swig_varlink_repr },
+    { Py_tp_str, (void*)swig_varlink_str },
+    { Py_tp_doc, (void*)varlink__doc__ },
+    { 0, NULL }
+  };
+  PyType_Spec spec = {};
```
bosmacs, line 216: These empty initializer lists pose a problem for some compilers, notably more recent versions of MSVC, or when using strict compile options (e.g. gcc with `-Wpedantic`).

schribl:
I am also using the code in this PR (#1613 ) and #1687 on Windows and Linux with Python 3.6-3.8 and I did not encounter any issues for a couple of month now. So thanks to @vadimcn and @vadz for this code!

@wsfulton is there any chance this will be consider to go into the mainline and can I maybe help to support this two PR?

wsfulton:
I've added this to the swig-4.1 milestone (consideration for next release).

Lib/python/pyinit.swg
```diff
@@ -200,12 +203,29 @@ swig_varlink_type(void) {
       return NULL;
   }
   return &varlink_type;
+#else
+  PyType_Slot slots[] = {
```
smistad, line 207: I get the following error with GCC 7 and 8 from this code:
```
In function ‘PyTypeObject* swig_varlink_type()’:
error: structured binding declaration cannot have type ‘PyType_Slot’
     PyType_Slot slots[] = {
                      ^~
note: type must be cv-qualified ‘auto’ or reference to cv-qualified ‘auto’
error: empty structured binding declaration
error: expected unqualified-id before ‘=’ token
     spec.slots = slots;
                ^
error: expected primary-expression before ‘;’ token
     spec.slots = slots;
                       ^
```
Not sure what is actually causing the error, as it only complains about this specific list initialization of PyType_Slot and not the others. Any idea on how to fix this?

vadz:
Note that f6d97d2338c79490513637a46e2e2d9ec80b8707, which is part of #1716, broke using this PR with master. The use of `PyUnicode_AsUTF8()` needs to be made conditional limited ABI not being used or the min version being >= 3.10, as this function has only become part of limited ABI in Python 3.10.

okainov:
Friendly reminder, is there any update on this? Would be great to have it in!

ojwb:
@vadz Apparently the missing `PyThread_allocate_lock()` and related functions are exported as of Python 3.10 (see https://bugs.python.org/msg396342) so maybe that's a sensible minimum Python version for SWIG to require for supporting this anyway.

vadz:
Yes, thanks, I've seen that we can have a proper solution with 3.10, but we've been using this with anything >= 3.5 for, what, a couple of years already by now, so I don't think it's such a problem to support older versions. The real problem is that there is still no SWIG release integrating this, but I don't know what could I do to help with this (other than maintaining my own fork of SWIG which I'm not quite ready to do yet...).

schnudi:
The stable Python ABI support would also for older (>=3.5) Python versions be really nice.

vadz:
@wsfulton If I rebase this on latest master and add the CI builds using `-py3-stable-abi` to GitHub Actions workflow, will you be able to merge this in observable future or would it be a waste of time to do it?

schnudi:
Push, I'm still looking forward to get the stable abi working with swig. If I could implement or test something, let me know.

Lib/python/pyinit.swg
ojwb: This is indeed not ISO C and we should avoid it:

```
$ echo 'struct { int x; } a = {};'|gcc -pedantic -x c - -o /dev/null
<stdin>:1:23: warning: ISO C forbids empty initializer braces [-Wpedantic]
```

The reason for this is presumably to 0-initialise members not explicitly initialised below.  The [Python docs for this structure](https://docs.python.org/3/c-api/type.html#c.PyType_Spec) explicitly enumerate the members, and the only one not explicitly initialised here is `itemsize`, so we should be able to drop the `{}` and add `spec.itemsize = 0;` below.

There seem to be two more instances of this.

ojwb:
`PyTuple_SET_ITEM` can presumably just be defined like `PyTuple_GET_ITEM` already is:

```
#ifdef Py_LIMITED_API
# define PyTuple_GET_ITEM PyTuple_GetItem
```

While I think you're overstating the importance of this (I use SWIG with Python and have not found a need for this) but it does seem like it ought to be merged, and looks plausible.  I'm not comfortable with doing that myself though.

vadz:
> `PyTuple_SET_ITEM` can presumably just be defined like `PyTuple_GET_ITEM` already is:
> 
> ```
> #ifdef Py_LIMITED_API
> # define PyTuple_GET_ITEM PyTuple_GetItem
> ```

Yes, this is what I did. This is not totally trivial, however, because the function and the macro have _different_ semantics (which seems like an awful idea to me but what do I know about API design), with the function decref'ing the previous item, unlike the macro. It seems that it doesn't matter for its use in the SWIG code, so we can do it like this, but it's not very nice to depend on this continuing to be the case in the future and we may end up silently starting to crash in limited API builds if master adds another use of this macro in some place where the difference matters. Which means I know need to track this somehow and remember to check for this when I merge with master the next time, which is something that I really could do without...
 
> While I think you're overstating the importance of this (I use SWIG with Python and have not found a need for this)

Do you really ship binaries for all of Python 3.4, 3.5, 3.6, 3.7, 3.8, 3.9 and 3.10?? We only did this for 3.[4-6] which were the only versions used before we started using this PR and it was already bad enough, I can't imagine doing it now.

> but it does seem like it ought to be merged, and looks plausible. I'm not comfortable with doing that myself though.

I understand this perfectly, but I'm just frustrated because I don't understand what's missing here, it looks like this could, and should, have been merged a very long time but nothing happens, even though other, vastly less important PRs, do get merged from time to time.

Anyhow, it's probably too late to do it this year. Hopefully it will get merged in 2022...

ojwb:
> Do you really ship binaries for all of Python 3.4, 3.5, 3.6, 3.7, 3.8, 3.9 and 3.10??

No, I don't ship binaries.

As to what binary-shipping Linux distros do, I know that Debian and derivates handle this by building for each supported minor version (which is pretty much all automated by the packaging tools).  There's typically only a few supported at a time - e.g. Debian unstable currently builds for 3.9 and 3.10 (the quick way to see is to check which python3*-dev packages python3-all-dev depends on).  This stable ABI would probably be helpful for someone creating a new binary Linux distro from scratch now, but I'd imagine all the existing ones built machinery to handle this long ago.

henryiii:
Even using cibuildwheel to build all your binaries (3.6-3.10, no one is using 3.4 or really even 3.5, those are well past EOL), having 1 binary instead of 5 reduces your PyPI upload requirements by **5x**, and keeps you from having to upload new wheels on every Python release in October. So it's a pretty big deal. Though it's pretty lightly used in practice (even still has some almsot-bugs, like not supporting platform tags, so you can't put different OSs and architectures in the same folder, like for a zipapp), largely because most systems don't really support it (this PR for SWIG, I think Cython 3 alphas have some support, pybind11 can't at the moment, etc)

smistad:
I have been using swig with this PR for a while. It is a real life-saver when you have a large binary python wheel. Uploading such a wheel for every python version to pypi would quickly use up the project size limits. Thank you so much for this work @vadimcn and @vadz! 

I hope this will eventually get merged into swig, so far I have been using a fork with this PR, together with https://github.com/swig/swig/pull/1687 which is also very useful.

degasus:
> Do you really ship binaries for all of Python 3.4, 3.5, 3.6, 3.7, 3.8, 3.9 and 3.10?? 

On Windows, we currently ship 2.7, 3.5, 3.6, 3.7, 3.8, 3.9 and there is a request for 3.10. All of them for x86 and for x64. We've patched the generated .py file to look for a path in the Windows registry, set the sys.path corresponding to the python version, and continue with the pyd.

I'm talking about 3 MB of generated cxx file and a 3.7 MB pyd for each of them (2.9 MB for x86). All of this to connect a commercial DLL of 5.5 MB. Oh, is there a point in talking about .pdb or the compilation time of a 3 MB cxx file?

So I'd like to answer: Yes!

vadz:
I'm not really sure how much sense does it make to continue this discussion, but the question is not whether it's possible to ship all these binaries -- the answer is definitely "yes" and, as I wrote, I used to do it myself. But if the question is whether I want to go back to it instead of shipping just a single binary for all Python 3 versions, then the answer is even more definitely "no".

degasus:
Sure, I'm on your side here. I hope my point was clear. It is insane being forced to ship every single 3.x pyd. I'm all for the stable API.

vadz:
This probably should be closed in favour of #2190 too now.

PR #1616, opened 2019-08-19
[Java] Add support for throwing IllegalStateException

bkotzz:
Adds support for throwing https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalStateException.html through the Java helper function.

PTAL @wsfulton


wsfulton:
You can write your own simple helper function to throw exceptions of any type. Nevertheless, I'll add this one in.

PR #1619, opened 2019-08-29
Fix error in generated code for Python in MSVC 2019.

emminizer:
Visual Studio 2019 release builds:
error C4703: potentially uninitialized local pointer variable 'p' used


emminizer:
This is very similar to https://github.com/swig/swig/commit/82969b07557bd7212b20373841406d4489dc5d78 but fixes part of Python.  I was directly impacted by the std_map, std_multimap, and std_pair.  I saw the pattern and fixed the other files too, although I was not impacted by those.

PR #1622, opened 2019-09-03
Add support for the <=> operator (C++20)

ZackerySpytz:
C++20 will introduce a `<=>` operator.

### DIFF
--- a/Doc/Devel/scanner.html
+++ b/Doc/Devel/scanner.html
@@ -204,6 +204,7 @@ SWIG_TOKEN_LESSTHAN          &lt;
 SWIG_TOKEN_GREATERTHAN       &gt; 
 SWIG_TOKEN_LTEQUAL           &lt;= 
 SWIG_TOKEN_GTEQUAL           &gt;= 
+SWIG_TOKEN_LTEQUALGT         &lt;=&gt;
 SWIG_TOKEN_NOT               ~ 
 SWIG_TOKEN_LNOT              ! 
 SWIG_TOKEN_LBRACKET          [ 
--- a/Doc/Manual/CPlusPlus20.html
+++ b/Doc/Manual/CPlusPlus20.html
@@ -34,6 +34,12 @@ Work has only just begun on adding C++20 support.
 
 <H2><a name="CPlusPlus20_core_language_changes">10.2 Core language changes</a></H2>
 
+<H3><a name="CPlusPlus20_spaceship_operator">10.2.1 Spaceship operator</a></H3>
+
+<p>
+SWIG correctly parses the spaceship operator <tt>&lt;=&gt;</tt>, and it should work in
+constant experssions.
+</p>
 
 <H2><a name="CPlusPlus20_standard_library_changes">10.3 Standard library changes</a></H2>
 
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -170,6 +170,7 @@ CPP_TEST_CASES += \
 	cpp17_nested_namespaces \
 	cpp17_nspace_nested_namespaces \
 	cpp17_u8_char_literals \
+	cpp20_spaceship_operator \
 	curiously_recurring_template_pattern \
 	default_args \
 	default_arg_expressions \
--- a/Examples/test-suite/cpp20_spaceship_operator.i
+++ b/Examples/test-suite/cpp20_spaceship_operator.i
@@ -1,0 +1,19 @@
+%module cpp20_spaceship_operator
+
+// Tests are designed so that code compiles with C++98 compilers
+
+%{
+#if __cplusplus >= 202002L
+#define CPP20 1
+#endif
+%}
+
+#ifdef CPP20
+
+#define ALIEN (4 <=> 2)
+
+#else
+
+#define ALIEN (-1)
+
+#endif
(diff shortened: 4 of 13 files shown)


ZackerySpytz:
This patch does not yet have any unit tests. Also, parts of the documentation still need to be updated.

ojwb:
Looks good so far.

> This patch does not yet have any unit tests. Also, parts of the documentation still need to be updated.

Were you intending to add tests and docs?

ZackerySpytz:
Yes.

wsfulton:
I'll merge this in when tests are available.

hjmjohnson:
@ZackerySpytz Do you have an ETA for the tests?

ojwb:
I've rebased onto current master, fixed some issues, and added docs and tests.

It seems `<=>` isn't supported in preprocessor expressions (which makes sense as it doesn't add anything useful there) - at least GCC and clang don't support it, so I've added an explicit error message for that case to SWIG too.

ojwb:
I tried to add a `_runme.php` but the new testcase doesn't generate code compilable with a non-C++20 compiler.

Will try to address, but it seems tricky to make a testcase for this which works with a non-C++20 compiler while actually testing everything - I don't see how to test `<=>` works in a constant as the value is parsed by SWIG which doesn't know whether the compiler supports C++20 - e.g. this for PHP at least:

```
#define ALIEN (4 <=> 2)
```

Will also add tests that `operator<=>` can be wrapped, and improve the manual docs a bit.

ojwb:
Having stared at this some more, I just don't see how I can test `<=>` support well with the requirement that the testcase also works with a non-C++20 compiler (which seems to be how all the C++14 and up tests work at the moment).

Here's what I'd write if it were conditional on C++20, much like the C++11 tests all seem to be:

```
%module cpp20_spaceship_operator

%inline %{
struct A {
  int v;

  explicit A(int v_) : v(v_) { }
};

%rename(spaceship) operator<=>;

#define ALIEN (0 <=> 1)
const int SPACE = 3 <=> 3;
%constant int COMET = 4 <=> 2;

int operator<=>(const A& a, const A& b) {
  return a.v - b.v;
}
%}
```

Looking at `cpp14_binary_integer_literals.i` I see I could take the approach there and test `int alien = (0 <=> 1);`, etc instead, though that seems to reduce the breadth of the test coverage.  Really it'd be better to test that case **as well** as the constant case.

However I really don't see a way to usefully test that SWIG can wrap `operator<=>` in the new style - if SWIG wraps it then it will generate a call to it, which isn't going to compile without a C++20 compiler.  The existing C++14 and C++17 testcases seem to just happen to be for features which can be tested well in the new style.

@wsfulton Is it OK to start a "conditional on C++20" list of testcases?  Or do you have any better ideas for how to test this?

wsfulton:
Yes, copying the C++11 style testing approach with a separate cpp20_xxx list of tests for adding in C++20 testing is okay. Grab latest serial 14 of [ax_cxx_compile_stdcxx.html](https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html) and replace the current serial 12 we have will provide C++20 compiler detection. Then copying the HAVE_CXX11 and replicate with a HAVE_CXX20 should do the job.

We could do the same for the C++14 and C++17 testing to make it consistent if you fancy that, although it is handy being able to test C++14 and C++17 features locally with older compilers. Leaving in the C++98 compatibility test code would make the rework easier, but it would be okay to rip it out too.

ojwb:
OK, sounds good.

> We could do the same for the C++14 and C++17 testing to make it consistent if you fancy that, although it is handy being able to test C++14 and C++17 features locally with older compilers. Leaving in the C++98 compatibility test code would make the rework easier, but it would be okay to rip it out too.

OK - I probably will look at making everything consistent, but I think as a second step.

The requirements for fully testing SWIG with C++14 and C++17 locally are likely to be met on a typical developer machine these days, which wasn't so true when this mechanism was added 3.5 years ago - C++14 was pretty much complete in GCC5/clang3.4/MSVC2015 (https://en.cppreference.com/w/cpp/compiler_support/14); C++17 in GCC7/clang5/some update to MSVC2017 (with the library support tending to lag more, but that's less likely to matter in SWIG testing) (https://en.cppreference.com/w/cpp/compiler_support/17).  In Ubuntu LTS default gcc terms, that's xenial for C++14 and bionic for C++17, with bionic the oldest Ubuntu LTS still in full support (and two newer LTS releases since).

> [ax_cxx_compile_stdcxx.html](https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html)

Related, I noticed recently while updating Xapian to use this macro instead of the older one specific to C++11 that the new macro rejects any compiler which doesn't define `__cplusplus` to say it supports the requested C++ version.  MSVC still always reports `199711L` by default (because Microsoft, and [even more stupidly they claim to have fixed this, but the fix only applies if you pass a special option](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/)), so currently SWIG's configure won't try to run any C++11 tests with MSVC.

We patched the macro for Xapian to address this (the current stable Xapian release requires C++11 so we noticed because this broke MSVC compilation) but I've not got around to submitting a patch to the macro upstream yet.

ojwb:
> I've not got around to submitting a patch to the macro upstream yet.

Now done: https://github.com/autoconf-archive/autoconf-archive/pull/255

PR #1626, opened 2019-09-09
Extend std::auto_ptr<> support to Ruby

vadz:
This is trivial as exactly the same typemap as for Python can be used
for Ruby too, all the differenced are abstracted by the unified typemap
library.

---

I don't know if it's better to copy the file or add a common file included from both Python and Ruby files. I'd prefer the latter, but `shared_ptr` typemaps are present in both languages directories, in spite of being 99% identical, so I did the same thing for this one too.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.1 (21 Aug 2019)
 ===========================
 
+2019-09-09: vadz
+            [Ruby] Add std::auto_ptr<> typemaps.
+
 2019-08-20: TekuConcept
             [Javascript] #1535 Add %native support to Javascript.
 
--- a/Examples/test-suite/li_std_auto_ptr.i
+++ b/Examples/test-suite/li_std_auto_ptr.i
@@ -12,7 +12,7 @@
 #endif
 %}
 
-#if defined(SWIGCSHARP) || defined(SWIGJAVA) || defined(SWIGPYTHON)
+#if defined(SWIGCSHARP) || defined(SWIGJAVA) || defined(SWIGPYTHON) || defined(SWIGRUBY)
 
 %include "std_auto_ptr.i"
 
--- a/Examples/test-suite/ruby/li_std_auto_ptr_runme.rb
+++ b/Examples/test-suite/ruby/li_std_auto_ptr_runme.rb
@@ -1,0 +1,43 @@
+#!/usr/bin/env ruby
+
+require 'swig_assert'
+
+require 'li_std_auto_ptr'
+
+k1 = Li_std_auto_ptr::makeKlassAutoPtr("first")
+k2 = Li_std_auto_ptr::makeKlassAutoPtr("second")
+swig_assert_equal_simple(Li_std_auto_ptr::Klass::getTotal_count(), 2)
+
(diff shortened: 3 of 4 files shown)


ojwb:
CI is failing for Ruby 1.9 and 2.6 (interestingly the oldest and newest we test, while working for those in between).  The failures don't look like travis issues.

vadz:
Thanks, I've somehow completely missed these failures (I wonder if there is any way to configure Travis to notify the PR author about errors?).

I'll try to install these Ruby versions and debug because right now I have no idea where do they come from.

ojwb:
> I wonder if there is any way to configure Travis to notify the PR author about errors?

[The docs](https://docs.travis-ci.com/user/notifications#configuring-email-notifications) say:

> Pull Request builds do not trigger email notifications.

So it seems not.

The situation is annoying for both sides - the maintainer usually has to comment to point out the CI failure; or to avoid this, the submitter has to know they won't be automatically notified and then poll all the PRs they submit having given a chance to let the CI run (which for SWIG is typically several hours later).

wsfulton:
Hi @vadz, you can use rvm to test these versions. Running `rvm install ruby-2.6` should do it.

I suspect the tests fail because the garbage collector cannot be relied upon for deleting the object. There are similar problems in C# and Java where we try very hard to force the gc to run and eventually give up and just print an error message instead of failing the test if the object is not deleted, see https://github.com/swig/swig/blob/master/Examples/test-suite/java/li_std_auto_ptr_runme.java. I don't know if Ruby has some more defined gc behaviour we can invoke/enforce. If not, the same approach will be needed for testing.

vadz:
> Hi @vadz, you can use rvm to test these versions. Running `rvm install ruby-2.6` should do it.

Yes, I'll try to do it some time soon, thanks.
 
> I suspect the tests fail because the garbage collector cannot be relied upon for deleting the object. 

Probably, but I don't know what can I do about it, considering that the tests already contain explicit calls to `GC.start` (which were needed as otherwise the tests were failing even when running locally).

> There are similar problems in C# and Java where we try very hard to force the gc to run and eventually give up and just print an error message instead of failing the test if the object is not deleted, see https://github.com/swig/swig/blob/master/Examples/test-suite/java/li_std_auto_ptr_runme.java. I don't know if Ruby has some more defined gc behaviour we can invoke/enforce. If not, the same approach will be needed for testing.

OK, I didn't think about just ignoring the failure to destroy the objects, but I'll do it if I can't find anything better, thanks for the advice.

vadz:
@wsfulton I implemented your suggestion and also call `GC.start` many times in a loop now. Unfortunately this still isn't always enough and, in fact, this seems to be orthogonal to the version of Ruby itself: e.g. the previously failing Ruby 1.9 and 2.7 tests pass now, but the previously passing test using Ruby 2.0 now gives the warning message.

FWIW I can confirm that in my own code all the objects are released, eventually, so the typemaps work as expected, it's just GC which doesn't consider them being worth collecting, apparently. And there doesn't seem to be any way to tell it to do it other than perhaps by actually allocating some big amount of memory in the object (using Ruby-specific allocation macros, so that its GC knows about it), which really doesn't seem to be worth it.

So could you please merge this into the upcoming 4.1? It's a simple enhancement that really shouldn't break anything. TIA!

PR #1629, opened 2019-09-16
Add support for MY_CXT in perl5.

hakonhagland:
Adds support for thread-safe file-static variables in Perl 5.

Please let me know how to improve this pullrequest.

### DIFF
--- a/Source/Modules/perl5.cxx
+++ b/Source/Modules/perl5.cxx
@@ -419,6 +419,29 @@ public:
       Printf(f_header,"#define SWIG_init    %s\n\n", boot_name);
       Printf(f_header,"#define SWIG_name   \"%s::%s\"\n", cmodule, boot_name);
       Printf(f_header,"#define SWIG_prefix \"%s::\"\n", cmodule);
+
+      /* Support for static varibles when using threads, see perldoc perlxs */
+      Printf(f_header, "#define MY_CXT_KEY \"%s::_guts\" XS_VERSION\n", module);
+      Printf(f_init, "/* Support for static variables when using threads,");
+      Printf(f_init, " see \"perldoc perlxs\" */\n");
+      Printf(f_init, "\n#ifdef __cplusplus\nextern \"C\" {\n#endif\n");
+      Printf(f_init, "\n#ifdef SWIG_USE_MY_CXT\n");
+      Printf(f_init, "XS(XS_%s_CLONE) {\n", underscore_module);
+      Printf(f_init, "\t dVAR; dXSARGS;\n");
+      Printf(f_init, "\t MY_CXT_CLONE;\n");
+      Printf(f_init, "\t XSRETURN_EMPTY;\n");
+      Printf(f_init, "}\n");
+      Printf(f_init, "#endif\n\n");
+      Printf(f_init, "SWIGRUNTIME void\n");
+      Printf(f_init, "SWIG_MyCxtInit(void) {\n");
+      Printf(f_init, "#ifdef SWIG_USE_MY_CXT\n");
+      Printf(f_init, "\t MY_CXT_INIT;\n");
+      Printf(f_init, "\t newXS(\"%s::CLONE\", XS_%s_CLONE,  (char*)__FILE__);\n",
+          module, underscore_module);
+      Printf(f_init, "#endif\n");
+      Printf(f_init, "}\n\n");
+      Printf(f_init, "#ifdef __cplusplus\n}\n#endif\n");
+
       Delete(boot_name);
     }
 
--- a/Doc/Manual/Perl5.html
+++ b/Doc/Manual/Perl5.html
@@ -78,6 +78,7 @@
 <li><a href="#Perl5_nn52">Overhead and code bloat</a>
 <li><a href="#Perl5_nn53">Typemaps</a>
 </ul>
+<li><a href="#Perl5_nn54">Threads and static variables</a>
 </ul>
 </div>
 <!-- INDEX -->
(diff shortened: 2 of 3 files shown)


wsfulton:
@tlby , could you look at this pull request please.

Source/Modules/perl5.cxx
```diff
@@ -419,6 +419,29 @@ class PERL5:public Language {
       Printf(f_header,"#define SWIG_init    %s\n\n", boot_name);
       Printf(f_header,"#define SWIG_name   \"%s::%s\"\n", cmodule, boot_name);
       Printf(f_header,"#define SWIG_prefix \"%s::\"\n", cmodule);
+
+      /* Support for static varibles when using threads, see perldoc perlxs */
```
ojwb: Comment typo: "varibles"

tlby:
I'm still working on this.  The VM I had for running support grid tests (against all the perl versions) melted and setting up a fresh one had some unexpected challenges.  Making headway, but slower than expected.

tlby:
The grid test environment is back and this patch did pass the tests with all versions, however we don't have test coverage of this behavior yet, and perl v5.6 is still supported so I would like to verify graceful degradation or meaningful errors when feature is enabled.

Also, since this doesn't seem to change the behavior of how SWIG translates declarations into interface code, I suspect this could be expressed as a collection of SWIG macros in the standard library, rather than a patch to the swig program itself.  I'll try to work out something equivalent there and see if I hit any snags.

tlby:
This macro seems to be working fairly well
```
%define %START_MY_CXT
%{
#define MY_CXT_KEY SWIG_prefix "::_guts" XS_VERSION
START_MY_CXT;
%}
%init %{
MY_CXT_INIT;
%}
%inline %{
static
void CLONE(...) {
  MY_CXT_CLONE;
}
%}
%enddef
```
Use looks very similar to this PR
```
%{
  typedef struct {
    int foo;
  } my_cxt_t;
%}

%START_MY_CXT;

%init %{
  MY_CXT.foo = 0;
%}
```
I think it might make sense to add this to `Lib/Perl5/perlmacros.swg` since according to perlxs documentation
> It is [...] strongly recommended that these macros be used by all XS modules that make use of static data.

I'd like to get some test coverage over this feature before we call any work here finished.

ojwb:
> perl v5.6 is still supported

I guess you mean "by SWIG"?  Upstream it's well past being supported:

https://perldoc.perl.org/perlpolicy.html#MAINTENANCE-AND-SUPPORT

The SWIG docs currently recommend 5.8 or later:

> For the best results, it is recommended that SWIG be used with Perl 5.8 or later. We're no longer testing regularly with older versions, but Perl 5.6 seems to mostly work, while older versions don't.

The oldest version actually tested by ci currently seems to be Perl 5.18.

The last 5.6 release was "5.6.2         2003-Nov-15", so over 16 years old now.  5.8.0 was released in 2002.  The oldest Perl I've seen installed anywhere in recent years is mingw which still seems to have 5.8.8, and even that's very much an outlier.

If keeping support for 5.6 requires much effort and/or makes it harder to improve support for more modern versions that are actually in use, bumping the minimum requirement to 5.8 seems entirely reasonable.

tlby:
> If keeping support for 5.6 requires much effort and/or makes it harder to improve support for more modern versions that are actually in use, bumping the minimum requirement to 5.8 seems entirely reasonable.

Yes, by SWIG.  In my last push for major enhancements William asked for support back to 5.0, but I could not find anything earlier than 5.6 that would build, so we compromised there.  I am still testing SWIG with 5.6 and up, even though I haven't figured out how to get ci to handle it.

h2xs sometimes provides backfill for new apis in ppport.h, I will see if they did anything for `MY_CXT` there.  If not, then I won't do any heroics to make 5.6 work, but it should at least be investigated.

ojwb:
Travis allows picking a specific Perl version via Perlbrew:

https://docs.travis-ci.com/user/languages/perl/

That says "Perl versions earlier than 5.8 are not supported."

We now require CI coverage to support a target language - it seems reasonable to extend that to requiring CI for the oldest version we can realistically claim to support.  It's great that you're testing more versions by hand, but that's extra work for you and means we don't get to see regressions a patch causes before we merge it.

FWIW, back in 2011, I tested which Perl versions **actually worked** with SWIG and found the majority of the test suite passed with 5.6 but older versions didn't work well at all - as a result I updated the docs to recommend 5.8 or later (f320ea1d4a83a58b3e0873824eaf693df62a9916).

I appreciate William's concerns, and agree we need to consider carefully before dropping support for old versions, but also we have limited resources.  If people actually want support for Perl 5.6 (or older) they need to help make that happen, the same as for most things in SWIG.

I think we also need to let people who've stepped up to maintain a backend make reasonable calls on which versions to support.  If we insist that to maintain a language you have to maintain it for ancient versions too we're going to burn out volunteers.

wsfulton:
Seems like perl 5.6 can't be tested and there is some doubt over this patch, so I will defer this until swig-4.1.0 unless @tlby gives it the thumbs up now.

vadz:
FWIW I really don't see why would anyone care about Perl 5.6 in 2020. It's ridiculously outdated.

wsfulton:
Agreed, and hence for the swig-4.0.2 patch releases, we aim not to break support for the versions that are known to be okay. Hence I've deferred to swig-4.1.0 where we can change explicitly state the minimum version that is supported.

tlby:
I should have written this down more clearly.  My hesitation about this patch is not strongly about reverse compatibility⁠ — I suspect that could be handled without great effort, but @ojwb is right, since Travis CI can't test that far back it's likely not worth the effort.  I think it would be healthy if we moved to a v5.8 or v5.10 minimum for perl.  There were large changes in that era and we could drop a lot of bits of swig's Perl support I frankly don't understand if this happened.

The things I was hoping to clean up for this patch are things my early contributions were flagged for as well:
- This work appears to be fully expressible using templates, and I think we aim to avoid patches to the language modules when that's true.  I think I cooked up a functional equivalent above.
- The work has no test case, and I've tried to write one myself, but struggle to work out how to force the bad thread synchronization this code intends to guard against.  The Perl documentation about the feature is somewhat vague about what exactly it guards against.  This is where I'm stuck right now.

ojwb:
We said we'd defer this to 4.1.0 so I've set that as the milestone.

ojwb:
> The things I was hoping to clean up for this patch are things my early contributions were flagged for as well:
> 
>     * This work appears to be fully expressible using templates, and I think we aim to avoid patches to the language modules when that's true.  I think I cooked up a functional equivalent above.

By "templates" I assume you mean stuff in `Lib/perl5/` rather than C++ templates?

If it can be done cleanly that does seem better than generating code from `perl5.cxx`.

>     * The work has no test case, and I've tried to write one myself, but struggle to work out how to force the bad thread synchronization this code intends to guard against.  The Perl documentation about the feature is somewhat vague about what exactly it guards against.  This is where I'm stuck right now.

A testcase which checks the code generated compiles and works in normal use would be sufficient I think.  If it's possible to trigger the bad situation that this feature aims to defend against even better, but if it's related to thread synchronization I'd imagine it could be hard to trigger and I don't think it's helpful to block merging this indefinitely because of that.

ojwb:
I've now made 5.8.0 the oldest Perl we aim to support.

I've also unset the milestone.  This change could still go in 4.1.0 if it was ready (some test coverage is really needed, and maybe reworking to use the `%define %START_MY_CXT` approach suggested above), but this is a non-invasive change and unlikely to break anything that doesn't use the new mechanism so it could reasonably be merged for any SWIG version rather than having to wait for the next x.y.0 release.

PR #1650, opened 2019-10-23
Add keys() method to JavaScript wrappers of std::map

vadz:
As the commit message explains, this is very far from ideal, but then std containers typemaps for JS are in a rather sorry state. I don't have the possibility to change them all, unfortunately, but at least adding this method makes `map` usable in JS as previously it was simply impossible to find which elements it contained.

### DIFF
--- a/Examples/test-suite/javascript/li_std_map_runme.js
+++ b/Examples/test-suite/javascript/li_std_map_runme.js
@@ -1,0 +1,36 @@
+var li_std_map = require("li_std_map");
+
+var collectionSize = 20;
+var midCollection = collectionSize / 2;
+
+// Set up an string to int map
+var simap = new li_std_map.StringIntMap();
+for (var i = 0; i < collectionSize; i++) {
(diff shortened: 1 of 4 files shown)


wsfulton:
Hi @vadz. I'm afraid that this is more of a quick fix hack not suitable for the SWIG library so am rejecting it as is. 

`std::map` does not depend on `std::vector`. We should not add this dependency into the wrappers either. SWIG does not add in additional STL dependencies to any other STL containers and I don't think this approach should change.

What the other languages do is provide access to the keys via iterators or return the full set of keys using a language specific types. For example 
https://github.com/swig/swig/blob/7ba19e758632c8aeddcf82ddf48a05f34e218bc9/Lib/python/std_map.i#L213
https://github.com/swig/swig/blob/7ba19e758632c8aeddcf82ddf48a05f34e218bc9/Lib/ruby/std_map.i#L165
The same approach should be used for Javascript.

PR #1651, opened 2019-10-23
Add minimal std::auto_ptr<> typemaps to JavaScript too

vadz:
Allow returning auto_ptr<> from wrapped functions.

Add a unit test checking minimal functionality and, if running under
"nodejs --expose-gc", that the objects are destroyed correctly.


wsfulton:
Merged, see commits 8c1c01f5e6550121715fdf3e1fb99b870450efc9...a9731251a4234a390993208098cdcef9c358372d

Issue #1643, opened 2019-10-08
swig -python -doxygen segfaults at "@param ..."


dweindl:
Hi, I am trying to use the `-doxygen` option for generating PyDoc comments. I stumbled over PyDocConverter seemingly having trouble handling `@param ...` inside the documentation of variadic functions.

Minimal example (tested with SWIG 4.0.1):
`test.i`:
```
%module test

/**
 * @brief Test fun
 * @param ... extra args
 */
void test_fun(...);
```
`swig -doxygen -python -c++  test.i`
-> Program received signal SIGSEGV, Segmentation fault.

See stack trace:
```
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:65
#1  0x00007ffff7e8b73c in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::compare(char const*) const () from /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00005555555a897a in std::operator==<char, std::char_traits<char>, std::allocator<char> > (__rhs="...", __lhs=<optimized out>) at /usr/include/c++/8/bits/basic_string.h:6108
#3  std::operator!=<char, std::char_traits<char>, std::allocator<char> > (__rhs="...", __lhs=<optimized out>) at /usr/include/c++/8/bits/basic_string.h:6110
#4  PyDocConverter::getParamType (this=<optimized out>, param="...") at Doxygen/pydoc.cxx:421
#5  0x00005555555aac8b in PyDocConverter::handleTagParam (this=0x555555721a50, tag=..., translatedComment="\nTest fun\n") at /usr/include/c++/8/bits/basic_string.h:936
#6  0x00005555555a8d07 in PyDocConverter::translateSubtree[abi:cxx11](DoxygenEntity&) (this=0x555555721a50, doxygenEntity=...) at /usr/include/c++/8/bits/stl_list.h:209
#7  0x00005555555ac4d6 in PyDocConverter::makeDocumentation (this=0x555555721a50, n=<optimized out>) at Doxygen/pydoc.cxx:889
#8  0x000055555559e9cd in DoxygenTranslator::getDocumentation (this=0x555555721a50, node=0x555555bc6570, indentationString=0x0) at Doxygen/doxytranslator.cxx:58
[...]

PR #1654, opened 2019-10-27
Fix for Issue #1643

mcfarljm:
Fixes Issue #1643, in which swig -doxygen would segfault when processing a doxygen comment for a parameter whose name does not appear in the actual argument list.  Test cases have been added to doxygen_basic_translate.i, and expected output updated for Python and Java.


vadz:
Looks good to me, thanks!

Issue #1647, opened 2019-10-21
/**/ sigfault SWIG when enabling -doxygen


ColinChargyBentley:
Hi,
When using SWIG 4.0.1 with -doxygen, if my code contains "/**/" sigfault with the following call stack:
```
#0  __strncpy_ssse3 () at ../sysdeps/x86_64/multiarch/strcpy-ssse3.S:2745
#1  0x0000555555590584 in DohNewStringWithSize (so=0x555555f67573, len=-1) at DOH/string.c:1115
#2  0x0000555555567add in yylook () at CParse/cscanner.c:457
#3  0x0000555555567cfd in yylex () at CParse/cscanner.c:556
#4  0x000055555556e081 in yyparse () at CParse/parser.c:4752
#5  0x000055555556cefa in Swig_cparse (f=0x7ffff7f53910) at CParse/parser.y:1235
#6  0x0000555555640436 in SWIG_main (argc=42, argv=0x555555959f70, tlm=0x555555944620 <modules+672>) at Modules/main.cxx:1194
```
Note than len is -1 because in#2:
- loc is "/**/"
- begin is 3
- end is 2

So end - begin = -1.
Such comments should be ignored and shloudn't trigger SWIG to crash.
In the meantime, I'll remove /**/ from my code.
Regards,
Colin Chargy

mcfarljm:
Fixed with PR #1656.

PR #1656, opened 2019-10-30
Fix doxygen crash with empty comment (Issue #1647)

mcfarljm:
Fix for Issue #1647.  Parsing "/**/" with -doxygen would result in a crash due to
calculation of comment start/end that does not work for an empty
comment.  Fixed by catching this case prior to processing.  Added
simple regression test to doxygen_basic_translate.


wsfulton:
Sorry about the multiple open/close I mistakenly did, but thank you for the contribution.

Issue #1632, opened 2019-09-20
Missing new line in PyDoc


huckw:
Using Swig 4.0.1 and -doxygen the source file has the following documentation...

```
			/////////////////////////////////////////////////////////////////////
			/// @brief Compute slanted edge MTF
			/// Compute the slanted edge MTF's on an image containing multiple squares, each with four edges
			///
			/// @param TestImage image containing the squares
			/// @param OverSamplingFactor integer oversampling factor
			/// @param PixelPitch the pixel pitch in units of \todo {enter the pixel pitch units}
			/// @param side - either Left or Right (first letter must be upper-case)
			/// @param ResultsMTF the resulting MTF data structure
			/// @return true on success, false otherwise
			/////////////////////////////////////////////////////////////////////
			bool SlantedEdgeMTF(cv::Mat &TestImage, const float OverSamplingFactor, const float PixelPitch, const std::string Side, DataMTF &ResultsMTF);
```
And the generate .py docs look like this...
```
    def SlantedEdgeMTF(self, *args):
        r"""
        *Overload 1:*
         Compute slanted edge MTF Compute the slanted edge MTF's on an image containing multiple squares, each with four edges :type TestImage: :py:class:`Mat`
        :param TestImage: image containing the squares :type OverSamplingFactor: float
        :param OverSamplingFactor: integer oversampling factor :type PixelPitch: float
        :param PixelPitch: the pixel pitch in units of TODO: {enter the pixel pitch units} :param side: - either Left or Right (first letter must be upper-case) :type ResultsMTF: :py:class:`DataMTF`
        :param ResultsMTF: the resulting MTF data structure :rtype: boolean
        :return: true on success, false otherwise
        """
       
```
The `:type` and `:rtype` stuff is at the end of the previous line instead of on its own line.  Sometimes even a `:param` ends up on a previous line (like the "side" parameter)!  Am I doing something incorrectly?

mcfarljm:
This appears to be an issue with Swig parsing the Doxygen comment style that you are using.  If you use this style:
```
/** 
 * comments
 */
```
then swig should process it correctly (the `:type` entries will come before the `:param` entries, each on their own line)

The reason that your example is missing the `:type` declaration for the `side` parameter is that there is a case mismatch between the Doxygen comment and the actual parameter declaration.

I will try to look into what is needed for swig to support this doxygen comment style, but in the meantime using the other style is a workaround.

huckw:
[...]

PR #1659, opened 2019-11-02
Fix newline handling for doxygen "///" comments

mcfarljm:
Fix for Issue #1632.  Newlines were not handled correctly when processing doxygen `///` style comments.  Unlike the other styles, the scanner processes each line individually, which results in all of the newlines being stripped out.  The fix adds newlines back in for the comments as they are processed in `cscanner.c`.  A new test case, `doxygen_basic_translate_style3.i` has also been added, mirroring that of the other two supported comment styles.

Previously, the following:
```
/// Test for a parameter with difficult type
/// (mostly for python)
/// @param a Very strange param
void function7(Shape::superType *a[10])
{
}
```
was translated as:
```
Test for a parameter with difficult type (mostly for python) :type a: :py:class:`Shape`
:param a: Very strange param
```
After this fix, it is translated as:
```
Test for a parameter with difficult type
(mostly for python)
:type a: :py:class:`Shape`
:param a: Very strange param
```

It would be good for someone that is familiar with `cscanner.c` to double-check the logic that I have used to modify the `String *cmt`:
```
cmt = NewStringf("%s ", cmt);
```
There may be a better way to append to a string, and as far as I can tell, there is no API for cleaning up the memory created from such a call (and there doesn't seem to be other such cleanup being done in `cscanner.c` anyway).

There is one edge case that still doesn't work correctly with this comment style, which is the use of comments within a code block.  This is because of the way that comment characters are stripped out of the front of the string in `doxyparser.cxx`.  I have modified the case where this was being tested and added a note about the limitation for future reference (it may be less straightforward to address).

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -622,6 +622,7 @@ DOXYGEN_TEST_CASES += \
 	doxygen_basic_notranslate \
 	doxygen_basic_translate \
 	doxygen_basic_translate_style2 \
+	doxygen_basic_translate_style3 \
 	doxygen_ignore \
 	doxygen_misc_constructs \
 	doxygen_nested_class \
--- a/Examples/test-suite/doxygen_basic_translate_style3.i
+++ b/Examples/test-suite/doxygen_basic_translate_style3.i
@@ -1,0 +1,95 @@
+%module doxygen_basic_translate_style3
+
+%include "doxygen_basic_translate.h"
+
+%inline %{
+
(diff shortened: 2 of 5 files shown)


wsfulton:
Thanks for the fix. 
> It would be good for someone that is familiar with cscanner.c to double-check the logic that I have used to modify the String *cmt:

I fixed the memory leak fixed in ae1e736c1b2aaefa7664a6d64829dba0c2ca72e2.

Issue #1044, opened 2017-07-31
SWIG does not respect move constructor for the returning value


luav:
Return by value with defined move constructor and deleted copy constructor fails on compilation because the constructor semantic is not respected.  
SWIG interface file:
```c++
%module vector_mbmvctor
%{
#include <utility>  // move()
%}

%inline {
struct Member {
	int  val;
	
	Member(): val(0)  {}
	
	Member(const Member&)=delete;
	Member(Member&&)=default;
	
	Member& operator=(const Member&)=delete;
	Member& operator=(Member&&)=default;
	
	Member release()  { return std::move(*this); }
};
}
```
The generated wrapper for `Member release()`  is:
```c++
SWIGINTERN PyObject *_wrap_Member_release(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  ...
  result = (arg1)->release();
  resultobj = SWIG_NewPointerObj((new Member(static_cast< const Member& >(result))), SWIGTYPE_p_Member, SWIG_POINTER_OWN |  0 );
  return resultobj;
fail:
  return NULL;
}
```
The wrapper yields the following compilation error:
```
./vector_mbmvctor_wrap.cxx: In instantiation of ‘SwigValueWrapper<T>& SwigValueWrapper<T>::operator=(const T&) [with T = Member]’:
./vector_mbmvctor_wrap.cxx:4661:10:   required from here
./vector_mbmvctor_wrap.cxx:32:73: error: use of deleted function ‘Member::Member(const Member&)’
 igValueWrapper& operator=(const T& t) { SwigMovePointer tmp(new T(t)); pointer 
```

It could be fixed respecting the constructor semantics via `is_move_constructible()`:
```c++
#if __cplusplus >= 201103L
  #include <type_traits>
  using std::is_move_constructible;
  #define MOVE_SEMANTICS
#endif  // No MOVE_SEMANTICS
...
  result = (arg1)->release();
  resultobj = SWIG_NewPointerObj(new (
#ifdef MOVE_SEMANTICS
    is_move_constructible<Member>::value
    ? Member(move(result)) :
#endif  // MOVE_SEMANTICS
    Member(result)
  ) , SWIGTYPE_p_Member, SWIG_POINTER_OWN |  0 );
...
```

luav:
The manual workaround is to create another member function that
- either returns the value by reference / pointer, retaining the value ownership by the original object,
- or returns created unique/shared_ptr to which the value is moved, transferring the ownership to the caller.

[...]

PR #1663, opened 2019-11-06
Some progress on issue #1044: Wrap move-only classes

d-b-w:
This includes two commits. The first commit I'm feel pretty confident of: it creates a test case demonstrating the issue, and addresses the first compiler error that pops up. The line in `Examples/test-suite/cpp11_move_only.i` that triggers the error is commented out right now, so I _could_ push just the first commit if you want.

The second commit allows my test case to _work_, but I don't know what ramifications it will have in other places. I'm including it primarily to highlight where the problem occurs.

### DIFF
--- a/Lib/swig.swg
+++ b/Lib/swig.swg
@@ -660,6 +660,8 @@ namespace std {
 #ifdef __cplusplus
 %insert("runtime") %{
 #ifdef __cplusplus
+#include <utility>
+
 /* SwigValueWrapper is described in swig.swg */
 template<typename T> class SwigValueWrapper {
   struct SwigMovePointer {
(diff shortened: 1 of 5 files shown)


Lib/swig.swg
```diff
@@ -673,7 +675,14 @@ template<typename T> class SwigValueWrapper {
...
   SwigValueWrapper& operator=(const T& t) { SwigMovePointer tmp(new T(t)); pointer = tmp; return *this; }
-  operator T&() const { return *pointer.ptr; }
+  SwigValueWrapper& operator=(T&& t) {
+    if (pointer.ptr != &t) {
+      SwigMovePointer tmp(new T(std::move(t)));
+      pointer = tmp;
+    }
+    return *this;
+  }
+  operator const T&() const { return *pointer.ptr; }
```
ojwb: These changes produce generated code that requires a C++11 compiler.  I think you need to conditionalised the new code based on checking the value of `__cplusplus` so that it still works as currently for a pre-C++11 compiler.

I think your CI failures are due to CI testing that things work with `-std=c++98` (at least in part, and perhaps entirely).

d-b-w: aha! Thanks.

ojwb:
This was recently addressed by 088dc6e8707b45872c687c980349ece6866c0bfd (which also seems to handle factory functions, which the patch here apparently doesn't).

PR #1670, opened 2019-11-20
Fix missing-field-initializers warning with Py3.8

jschueller:
Python 3.8 adds tp_vectorcall, tp_print is added for compat just for 3.8
https://github.com/python/cpython/pull/13185/files#diff-b5db2632fa7acaf3b44abb56f18b9615


wsfulton:
Thanks, it was merged and combined with #1696 and 489629fd532593cbf856a22990b520f770561984.

PR #1672, opened 2019-11-24
Octave 4.4 bugfix

friedrichatgc:
In octave 4.4 the function
`virtual bool is_object() const`
was renamed to
`virtual bool isobject() const`
which is overload by the SWIG octave runtime.
The first commit fixes this, such that for octave 4.4 still the right overload from SWIG is called.

The second commit extends the octave operator test by the concatenation operator. This operator requires that isobject returns true which was not the case when using octave 4.4.


wsfulton:
The Travis tests (Octave 4.4) show a failure:
```

error: invalid conversion of multi-dimensional struct to scalar struct

error: called from

    ../../../../../Examples/octave/operator/runme.m at line 47 column 3

error: octave_base_value::parent_class_name_list(): wrong type argument 'swig_ref'

error: called from

    ../../../../../Examples/octave/operator/runme.m at line 47 column 3
```
This will need fixing before merging to master.

friedrichatgc:
Yes, sure. I will try to fix this. However, this test is working well on my machine locally with octave 4.4 but one other test fails which works on travis. Do you know about any test instabilities?

wsfulton:
You can see the test results on Travis at the bottom of this issue.

wsfulton:
The `isobject` rename was introduced in octave-4.4.0 and so this line:
```diff
+#if SWIG_OCTAVE_PREREQ(4,2,0)
```
should be
```diff
+#if SWIG_OCTAVE_PREREQ(4,4,0)
```

wsfulton:
Should be fixed by #2020 for Octave >= 6, but not for versions 4.4 to 6.0.

wsfulton:
ping @friedrichatgc. I'd like this fixed for the upcoming swig-4.1 release. Any news on your fix. A patch for versions >= 4.4 and < 6.0 is what is needed.

friedrichatgc:
I have ported this patch now to current master by:
- fixing the wrong version check (4.2 -> 4.4)
- enabled the new check with octave "horzcat" operator on VPATH builds (the new github actions CI seems to use VPATH build)

However, enabling this leads to a crash in the check Examples/octave/module_load: loading in a local (function) scope leads to a crash in octave >=4.4 <6.
In octave 6 even the checks from an unmodified master do not compile on my system.

I have tested different octave version using docker images with the following Dockerfile
https://gist.github.com/friedrichatgc/c06eeea96473b0d1cbc744e5a723b3c1

wsfulton:
Hi @friedrichatgc. Yes, indeed, there is something in your modifications that lead to module_load example failing for >=4.4 to <6. This is a regression, so really needs fixing. module_load works using master for this range of versions. Does module_load also fail for 6.0-6.3? The Github Actions shows it passing for 6.4.

friedrichatgc:
I have now switched my tests to the docker-hub images from gnuoctave/octave which provides more octave versions.
The latest commit in this PR (f0bb77221a3c0a88e65d87b6a611b9e180b16108) has the following results per octave version:
4.2.2: all checks OK
4.4.1,5.1.0,5.2.0: all checks OK except module_path which crashes
6.1.0,6.2.0,6.3.0,6.4.0: all checks OK
I have no idea why my changes break newer octave 4 versions but not octave 6. With the current master branch my use-case will work at least with octave >=6. So its fine for me to reject this PR (I can remove the workarounds in my project when switching to octave 6)

friedrichatgc:
Update!
The crash occurs in the on-exit code after an octave assert call in the .m file is false.
This crash happens in all octave versions if an assert is false.

The assert is triggered since the octave "exists" function behaves differently in different version.
See also the commit message of https://github.com/swig/swig/pull/1672/commits/d0121bff6ec45a2bdffb2d6d2a1889d49038feb0
which fixes the check module_path by adapting the asserts in the test dependent on the octave version.

wsfulton:
The changing behaviour of `exists` doesn't feel right across different versions of Octave. @kwwette, you've modified this example a lot over the years for different versions of Octave and you added in these `exists` checks. What do you think about this pull request, or perhaps the example is too sensitive to changes in Octave and should be dumbed down?

kwwette:
@wsfulton Those checks could be simplified to just
```
assert(exist("swigexample"));
```
It's probably sufficient to just check that an identifier `swigexample` exists, rather than needing to check what kind of identifier it is, which has evidently changed for newer Octave versions.

ojwb:
OK, I'm going to merge this (since the CI failure is appveyor and not related) and then push a commit to simplify the checks as kwwette suggests above.

wsfulton:
@kwwette is suggesting to use `assert(exist("swigexample"))`, but this patch does not do that. It tries to check how many objects are in existence which behave differently in different versions. The idea is to make the test more robust, so the patch needed simplifying before merging.

ojwb:
@wsfulton I pushed that change to the branch before merging - see a60655a86bd4a708c3273f39fa3cd7f2c67ddc90.

wsfulton:
Oh, sorry, I'm not being clear enough. What I envisaged is instead of all this:
```octave
  if compare_versions(version(), '4.4', '>=') && compare_versions(version(), '6', '<')
    % in octave >= 4.4 < 6 'swigexample' is recognized as variable (1) instead of a .oct file (3) on the second call of 'swigexample'
    assert(exist("swigexample") == 3 || exist("swigexample") == 1);
  else
    assert(exist("swigexample") == 3);
```
replace with what @kwwette suggested:
```octave
assert(exist("swigexample"))
```
The thinking is that different versions of Octave will return different ideas of how "swigexample" exists, but anything > 0 indicates it does exist. See https://octave.sourceforge.io/octave/function/exist.html.

ojwb:
Yes that is exactly what I did already.

ojwb:
Oh hang on. Something odd has happened as the commit I linked to seems to
reverse the change not make it. Will investigate when not on phone.

ojwb:
@wsfulton OK, looks like I must have merged head from friedrichatgc:octave-4-4-bugfix rather than the extra commit I'd made locally.  I dug the missing commit out from my reflog and have cherry-picked and pushed as 1a03840172fef655e5e752c14195756185e4668d.  Thanks for spotting and sorry for the miscommunication.

wsfulton:
No problem, got there in the end :+1:

PR #1681, opened 2019-12-16
Fix SWIG crash in during python generation with -doxygen

burnss887:
If SWIG find next construction, that will be crash:
```c++
  /** \details
    Returns the matrix which maps the WCS to
    the plane coordinate system.

    \param normal [in]  The normal to the plane.
    \param plane [in]  The plane.

    \remarks
    If normal is specified, the origin of the plane is (0,0,0)
  */
  static ECS worldToPlane( const GeVector3d& normal );
  static ECS worldToPlane( const GePlane& plane );
```
I add two checking for empty std::string, and null pointer.

Please, accept this fix.

Regards,
Ivan.


vadz:
The fix looks reasonable, but I don't quite understand how does the name end up being null in the first place in the snippet above, as it seems to be specified for both parameters. It would be nice to extend one of the Doxygen tests to check for this.

P.S. Travis CI failure is spurious and is due to `com.sun.javadoc` disappearance in OpenJDK 13 (we really need to do something about this, but it's unrelated to this PR).

burnss887:
I Think, the problem is in comment, which located before two functions. But, when Swig parse doxygen comments, that he expect to see two arguments for first function. But it see only one argument in function, and return null pointer. And than in callstack called strlen() function which no processing null pointers, and throw unhandled exception.

vadz:
Again, this makes sense, but it would be great to have a precise test case for it, i.e. just add an example of such comment to one of the existing tests that resulted in a crash before and doesn't after applying this PR. I'll try to do it myself when I have time, but if you could please do it, it would be great. TIA!

wsfulton:
Can you please fix this pull request so that it addresses the Python issue only. It seems to have 'grown' some C# changes. Is the additional testing request going to be addressed?

wkalinin:
Probably was intended to be put in a separate branch, no sense to mix these changes.

burnss887:
Sorry, yes, intended to be put in a separate branch, no sense to mix these changes. 
I reverted merges.

wsfulton:
I can't replicate the bug. I can't even get the testcase to go through the code modified in doxyparser.cxx. I used this testcase and the `-doxygen` command line argument:
```
%inline %{
struct ECS {};
struct GeVector3d {};
struct GePlane {};
/** \details
    Returns the matrix which maps the WCS to
    the plane coordinate system.

    \param normal [in]  The normal to the plane.
    \param plane [in]  The plane.

    \remarks
    If normal is specified, the origin of the plane is (0,0,0)
  */
  static ECS worldToPlane( const GeVector3d& normal );
  static ECS worldToPlane( const GePlane& plane );
%}
```
Can you please provide input that triggers the bug in order for me to consider this pull request. If you find it hard, then run swig with the `-E` option on your real project and dump the resulting preprocessed output somewhere for us to download.

jmcfarland-swri:
These changes touch some of the same code as my earlier PR #1654 (which does add a new test case).  We're probably addressing the same underlying issue.  I'll try to take a closer look when I have some time.

mcfarljm:
These changes touch some of the same code as my earlier PR #1654 (which does add a new test case). We're probably addressing the same underlying issue. I'll try to take a closer look when I have some time.

mcfarljm:
I tested this from a few different commits, including aa47b4c438421595b2764044b54502ad378cd7e5 (the parent commit of the PR change).  I used basically the same input as William, and I also wasn't able to reproduce the crash.  With that input, I don't get any comment translation, apparently due to the second overloaded function definition not having any doxygen comments.  If I comment out the second `worldToPlane` declaration, then swig produces this output:
``` python
def worldToPlane(plane):
    r"""

    Returns the matrix which maps the WCS to
    the plane coordinate system.

    :param normal: [in]  The normal to the plane.
    :type plane: :py:class:`GePlane`
    :param plane: [in]  The plane.

    Remarks: 
    If normal is specified, the origin of the plane is (0,0,0)
    """
    return _burns_pr.worldToPlane(plane)
```

Note that removing the argument name altogether, e.g.:
``` c++
static ECS worldToPlane( const GeVector3d& );
```
will cause `swig -doxygen` to crash.  This is Issue #1643, fixed by PR #1654.

I also confirmed that as William noted, with this input, the program doesn't go into the code that was changed in `doxyparser.cxx`.  Instead, it takes the first branch here: 
https://github.com/swig/swig/blob/9cfeb592edc39dbe702cb56b571ed9e29b3a8276/Source/Doxygen/doxyparser.cxx#L1079-L1082
If there's an issue with the code in `addDoxyCommand` (which there certainly may be), it would be good to have a test case that illustrates the bug.

Also note that the parameter direction flag in these examples is not correct doxygen syntax.  Perhaps that was intentionally being done because `swig -doxygen` did not previously support parameter direction, but that was recently added.

burnss887:
I attach testfile for reproduce (file test.i). 
[python_doxy_crash.zip](https://github.com/swig/swig/files/4071188/python_doxy_crash.zip)

Also, found if function contains in only type of parameter (without parameter name) , then generated code (see: Overload 2) not contains information about Python type (:py:class:'GePlane'). Note: I also attached generated python file.
 
image

wsfulton:
#1654 fixes the nameless parameter crash in python_doxy_crash.zip. The fix there is very similar to the pydoc.cxx fix in this patch, so is now redundant in this patch. I still havn't seen a testcase that requires the changes in this patch for doxyparser.cxx. Please close this issue if you can't segfault using the latest in git. Thanks.

wsfulton:
@burnss887 Your C++ declaration is missing the parameter name, so SWIG is unable to find the type of the parameter because it is unnamed. The solution is to name the parameter in your C++ declaration. I can't see any other solution really.

burnss887:
Problem resolved. Thanks!

PR #1683, opened 2019-12-18
Reimplement Doxygen unit tests for Java using Java 9+ API

vadz:
My Java skills are below average, so the code could almost certainly profit from being reviewed by someone with better language knowledge, but at the very least it works (tested with Java 11 and 14) and allows us to reenable the currently failing test run under macOS, so I'd like to merge this if you don't mind.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1351,10 +1351,10 @@ fi
 # Check Java version: we require Java 9 or later for Doxygen tests.
 if test -n "$JAVAC"; then
     AC_MSG_CHECKING(if java version is 9 or greater)
-    javac_version=`$JAVAC -version 2>&1`
+    javac_version=`"$JAVAC" -version 2>&1`
     java_version_num=`echo $javac_version | sed -n 's/^javac //p'`
     if test -z "$java_version_num"; then
-        AC_MSG_WARN([unknown format for Java version returned by $JAVAC ($javac_version)])
+        AC_MSG_WARN([unknown format for Java version returned by "$JAVAC" ($javac_version)])
         JAVA_SKIP_DOXYGEN_TEST_CASES=1
         AC_MSG_RESULT(unknown)
     else
--- a/configure.ac
+++ b/configure.ac
@@ -1408,15 +1408,6 @@ if test -z "$JAVA_HOME" && test -n "$JAVA_HOME_MAYBE" ; then
   fi
 fi
 
-# Javadoc support required for the Java test-suite is available by default in jdk9+ and in tools.jar in earlier jdk versions
-AC_MSG_CHECKING(for java tools.jar)
-if test -n "$JAVA_HOME" && test -r "$JAVA_HOME/lib/tools.jar" ; then
-  JAVA_TOOLS_JAR="$JAVA_HOME/lib/tools.jar"
-  AC_MSG_RESULT([$JAVA_TOOLS_JAR])
-else
-  AC_MSG_RESULT(not found)
-fi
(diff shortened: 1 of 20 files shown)


vadz:
Oh. I carefully checked all Travis CI builds to confirm that we were not using Java 8 in any of them any longer, but didn't think about AppVeyor ones, which still do. I'll have to disable Doxygen tests when using Java 8 or earlier...

vadz:
This should be good to go now (I've restarted a couple of completely unrelated Travis CI jobs that died for unknown reasons).

configure.ac
```diff
@@ -1348,6 +1348,33 @@ else
   JAVAC="$JAVACBIN"
 fi
 
+# Check Java version: we require Java 9 or later for Doxygen tests.
+if test -n "$JAVAC"; then
+    AC_MSG_CHECKING(if java version is 9 or greater)
+    javac_version=`$JAVAC -version 2>&1`
```
wsfulton:
```suggestion
    javac_version=`"$JAVAC" -version 2>&1`
```
To fix Appveyor output:
```
configure: WARNING: unknown format for Java version returned by C:/Program Files (x86)/Java/jdk1.8.0/bin/javac (./configure: line 8548: C:/Program: No such file or directory)
```

wsfulton:
Great to have this upgraded to Java 9. I had just started looking at what was required and the change looks fine. I was under the impression for the javadoc parsing we could have used Java 6 code, I think the Java 9 restriction you have is down to the Doclet. Anyway, I don't think it is important and restricting to Java 9 minimum is fine for these tests.

I checked the output of the old and new tools running this script in Examples/test-suite/java:
```bash
outfile=/tmp/doxygen_comments.txt
rm -f $outfile

doxytests="doxygen_alias doxygen_basic_notranslate doxygen_basic_translate doxygen_basic_translate_style2 doxygen_ignore doxygen_misc_constructs doxygen_nested_class doxygen_parsing_enums doxygen_parsing_enums_proper doxygen_parsing_enums_simple doxygen_parsing_enums_typesafe doxygen_parsing_enums_typeunsafe doxygen_parsing doxygen_translate_all_tags doxygen_translate doxygen_translate_links"

for t in $doxytests ; do
    make -s $t.cpptest
done

for t in $doxytests ; do
    echo "........... $t .............." >> $outfile
    java CommentParser $t >> $outfile
done
```
and comparing outputs. There are a number of `\n`, `<br>` and whitespace differences. There seems to be quite a lot of working around these diffs in the code which I havn't quite understood, but suspect they are okay. Probably a cleanup in this area is needed to update the tests with the new output at a later date. 

At first, I thought there were missing comments in the new code, but it looks like the output is just in a different order. I'll try and analyse it a bit better later, but need to rush off right now.

configure.ac
vadz: Thanks, I missed this one.

vadz:
Thanks for looking at this!

Yes, its the javadoc doclets API we used for "parsing" (in quotes because we really didn't parse anything but just let javadoc do it for us) is what got removed in Java 13 and had to be replaced with Java 9+-only new doclet API. I'm not sure what could be the alternative, however: we definitely don't want to do any actual parsing ourselves, do we?

The output is indeed different and I was a bit worried about it initially, but IMO the important thing is that `make check-doxygen` works, so I didn't bother with `\n`s and `<br>`s. As for the comment order changes, we could easily avoid them by sorting them but, again, I don't think it really matters.

PR #1687, opened 2019-12-23
Make directors implementation for Python work with limited API

vadz:
Don't use Python thread functions in directors implementation code, as
they're not exported from Python version-independent library under
Windows, see https://bugs.python.org/issue39123

This morally reverts 4aaf48d65055bd68cc55e81f915af39693e9a7a6, although
this commit uses a somewhat different (and cleaner) implementation of
the mutex abstraction and so it isn't a literal revert.

---

See #1613.

### DIFF
--- a/Lib/python/director.swg
+++ b/Lib/python/director.swg
@@ -14,6 +14,22 @@
 #include <vector>
 #include <map>
 
+#if defined(SWIG_PYTHON_THREADS)
+/*  __THREAD__ is the old macro to activate some thread support */
+# if !defined(__THREAD__)
+#   define __THREAD__ 1
+# endif
+#endif
+
+#ifdef __THREAD__
+# if defined(_WIN32)
+#   define WIN32_LEAN_AND_MEAN /* skip OLE and other unnecessary headers */
+#   define NOMINMAX /* avoid most common clash with windows.h contents */
+#   include <windows.h>
+# else
+#   include <pthread.h>
+# endif
+#endif
 
 /*
   Use -DSWIG_PYTHON_DIRECTOR_NO_VTABLE if you don't want to generate a 'virtual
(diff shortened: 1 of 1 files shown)


ojwb:
I'm rather luke warm on this.

In particular, aren't the python thread functions available on all other platforms?

And `#include <windows.h>` is excessively namespace polluting - we should at least define the magic macros which curtail that to reduce the chances of this breaking user interface files which currently work.

vadz:
> In particular, aren't the python thread functions available on all other platforms?

They probably are, but OTOH it's not clear whether they're supposed to be part of limited API or not, so I'm not sure how safe is it to use them. It also just seemed nicer and more symmetric to use pthreads functions directly for symmetry (FWIW Python doesn't have any other implementations than Win32 and pthreads neither, but their mutexes are actually semaphores, so the new version is simpler and should have less overhead, if it matters).

> And `#include <windows.h>` is excessively namespace polluting - we should at least define the magic macros which curtail that to reduce the chances of this breaking user interface files which currently work.

Do you mean just `#define NOMINMAX` or something more? Because I don't really want to duplicate the entire contents of [this file](https://github.com/wxWidgets/wxWidgets/blob/v3.1.3/include/wx/msw/winundef.h) here...

ojwb:
> Do you mean just `#define NOMINMAX` or something more?

That and `#define WIN32_LEAN_AND_MEAN` are the ones I had in mind.  We should probably `#undef` them again afterwards so that users with interface files which explicitly `#include <windows.h>` get to decide what they want from it.

There are a ridiculous number of other `NO`-prefixed macros one can define (e.g. see https://stackoverflow.com/questions/1394910/how-to-tame-the-windows-headers-useful-defines) - defining all except whatever is needed to get the thread functions would minimise the risk of collisions (short of providing our own prototypes) but I'm not sure it's worth the complication to go further.

> OTOH it's not clear whether they're supposed to be part of limited API or not, so I'm not sure how safe is it to use them.

Exactly what's in/not in the limited API really should be clearly defined - if it isn't that really needs rectifying by the Python developers.

> It also just seemed nicer and more symmetric to use pthreads functions directly for symmetry (FWIW Python doesn't have any other implementations than Win32 and pthreads neither, but their mutexes are actually semaphores, so the new version is simpler and should have less overhead, if it matters).

Python could be ported to a platform with its own thread API though, and assuming pthreads means that SWIG would then also need porting - currently it should work as-is.  Perhaps that's just a problem to solve if it actually happens though.

vadz:
> > Do you mean just `#define NOMINMAX` or something more?
> 
> That and `#define WIN32_LEAN_AND_MEAN` are the ones I had in mind.

I've updated the PR to include these ones.

> We should probably `#undef` them again afterwards so that users with interface files which explicitly `#include <windows.h>` get to decide what they want from it.

This is not really useful, as including the header the second time wouldn't do anything unless they manually `#undef _WINDOWS_` and if they do this, they could undef the other symbols too.

> > OTOH it's not clear whether they're supposed to be part of limited API or not, so I'm not sure how safe is it to use them.
> 
> Exactly what's in/not in the limited API really should be clearly defined - if it isn't that really needs rectifying by the Python developers.

It should, but my bug report remains without any real response so far, so I still don't know the answer here.

> > It also just seemed nicer and more symmetric to use pthreads functions directly for symmetry (FWIW Python doesn't have any other implementations than Win32 and pthreads neither, but their mutexes are actually semaphores, so the new version is simpler and should have less overhead, if it matters).
> 
> Python could be ported to a platform with its own thread API though, and assuming pthreads means that SWIG would then also need porting - currently it should work as-is. Perhaps that's just a problem to solve if it actually happens though.

Yes, this is how I feel about it.

ojwb:
> >  We should probably #undef them again afterwards so that users with interface files which explicitly #include <windows.h> get to decide what they want from it.

> This is not really useful [...]

Ah yes - the user would actually have to make sure they `#include <windows.h>` before SWIG does.

vadz:
This probably should be closed now in favour of #2190.

vadz:
Part of #2190 now.

PR #1688, opened 2019-12-24
Changed return type of swig_this() to size_t.

friedrichatgc:
The type long may be 4 bytes but swig_this() must return the address of
the object as an integer. Using size_t ensures that the return type can
store a pointer.

### DIFF
--- a/Lib/octave/octrun.swg
+++ b/Lib/octave/octrun.swg
@@ -507,10 +507,10 @@ SWIGRUNTIME void swig_acquire_ownership_obj(void *vptr, int own);
 	delete this;
     }
 
-    long swig_this() const {
+    size_t swig_this() const {
       if (!types.size())
-	return (long) this;
-      return (long) types[0].second.ptr;
+	return (size_t) this;
+      return (size_t) types[0].second.ptr;
     }
     const char* help_text() const {
       if (!types.size())


vadz:
I don't know anything about Octave module, but, just out of curiosity, why can't this function return `void*`? FWIW in practice `size_t` should work, i.e. be big enough to contain `void*`, on all the usual platforms, but in theory this doesn't necessarily have to be the case.

friedrichatgc:
According the c++ standard size_t must be able to store any none-member pointer. So, size_t should be valid on all platforms.
swig_this is used to get the object pointer in the interpreter (octave). Since, octave can handle different types of integers but no pointers the swig_this function in octave must return a integer.
This swig_this function in octrun.swg is only used in octruntime.swg. Hence, it can also return void* (as proposed) since its return value is converted in octruntime.swg for octave to octave_uint64 using the cast to unsigned long long.
So void* should also work from my point of view but the current approach using long is definitively wrong, at least on Windows 64bit where long is just 4 bytes but pointers are 8 bytes.

ojwb:
An old ticket I know but I stumbled over it and thought this was worthy of comment:

> According the c++ standard size_t must be able to store any none-member pointer.

I'm fairly sure that's not actually the case.  E.g. https://en.cppreference.com/w/cpp/types/size_t says:

> On many platforms (an exception is systems with segmented addressing) std::size_t can safely store the value of any non-member pointer, in which case it is synonymous with [std::uintptr_t](https://en.cppreference.com/w/cpp/types/integer).

Maybe Octave doesn't support such platforms though - I'm not aware of anything currently mainstream which has such a `size_t`.

PR #1695, opened 2020-01-04
Verbatim comments translation for C#

wkalinin:
Added literal doxygen comments translation for C#, which is enough for our needs. Hope other C# users would find that useful.

### DIFF
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -1430,15 +1430,15 @@ public:
 	if (!GetFlag(n, "firstenumitem"))
 	  Printf(enum_code, ",\n");
 
-  if (doxygen && doxygenTranslator->hasDocumentation(n)) {
-    String* doxygen_comments = doxygenTranslator->getDocumentation(n, "  ");
-    if (comment_creation_chatter)
-      Printf(enum_code, "/* This was generated from enumvalueDeclaration() */");
-    Printv(enum_code, "\n /** ", Char(doxygen_comments), " */\n", NIL);
-    Delete(doxygen_comments);
-  }
+	if (doxygen && doxygenTranslator->hasDocumentation(n)) {
+	  String* doxygen_comments = doxygenTranslator->getDocumentation(n, "  ");
+	  if (comment_creation_chatter)
+	    Printf(enum_code, "/* This was generated from enumvalueDeclaration() */");
+	  Printv(enum_code, "\n /** ", Char(doxygen_comments), " */\n", NIL);
+	  Delete(doxygen_comments);
+	}
 
-  if (csattributes)
+	if (csattributes)
 	  Printf(enum_code, "  %s\n", csattributes);
 
 	Printf(enum_code, "  %s", symname);
@@ -1555,7 +1555,7 @@ public:
       String* doxygen_comments = doxygenTranslator->getDocumentation(n, "  ");
       if (comment_creation_chatter)
         Printf(constants_code, "/* This was generated from constantWrapper() */");
-      Printv(constants_code, " /** ", Char(doxygen_comments), " */ ", NIL);
+      Printv(constants_code, " /** ", Char(doxygen_comments), " */\n", NIL);
       Delete(doxygen_comments);
     }
 
@@ -1968,7 +1968,7 @@ public:
       String* doxygen_comments = doxygenTranslator->getDocumentation(n, 0);
       if (comment_creation_chatter)
         Printf(proxy_class_def, "/* This was generated from emitProxyClassDefAndCPPCasts() */");
-      Printv(proxy_class_def, " /** ", Char(doxygen_comments), " */ \n", NIL);
+      Printv(proxy_class_def, " /** ", Char(doxygen_comments), " */\n", NIL);
       Delete(doxygen_comments);
     }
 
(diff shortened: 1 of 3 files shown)


Source/Modules/csharp.cxx
```diff
@@ -155,6 +161,7 @@ class CSHARP:public Language {
       director_method_types(NULL),
       director_connect_parms(NULL),
       destructor_call(NULL),
+      structuralComments(NULL),
```
vadz: This results in
```
Source/Modules/csharp.cxx:164:7: warning: field 'structuralComments' will be initialized after field 'output_file' [-Wreorder]
```
which is harmless here but still would be better to fix.

Source/Modules/csharp.cxx
```diff
@@ -46,6 +47,8 @@ class CSHARP:public Language {
   bool global_variable_flag;	// Flag for when wrapping a global variable
   bool old_variable_names;	// Flag for old style variable names in the intermediary class
   bool generate_property_declaration_flag;	// Flag for generating properties
+  bool doxygen;			//flag enabling comment processing
+  bool comment_creation_chatter; //flag for getting information about where comments were created
```
vadz: Was this supposed to be left in? Because there doesn't seem to be any way to set it, so it doesn't look useful in its current form.

Source/Modules/csharp.cxx
```diff
@@ -1434,6 +1514,22 @@ class CSHARP:public Language {
     return SWIG_OK;
   }
 
+  /* -----------------------------------------------------------------------
+ * doxygenComment()
+ * Simply translates the doxygen comment and places it into the appropriate
```
vadz: I'm not sure if "translates" here is really correct, as it doesn't seem to be doing any translation, does it?

Source/Modules/csharp.cxx
```diff
@@ -1456,6 +1552,15 @@ class CSHARP:public Language {
     Swig_save("constantWrapper", n, "value", NIL);
     Swig_save("constantWrapper", n, "tmap:ctype:out", "tmap:imtype:out", "tmap:cstype:out", "tmap:out:null", "tmap:imtype:outattributes", "tmap:cstype:outattributes", NIL);
 
+    //translate and write comment if flagged
```
vadz: Sorry, could you please explain what do you mean by "if flagged" here?

repo-alt:
This is actually very old code, added when doxygen was just released in it's own branch. I'll fix the warning and remove the debug output.
I have no idea what comments may be translated to, our documentation tool supports Doxygen in C# just fine, so I needed to only copy the comments.
There are quite a bit of doxygen tests in Java test suite, what test would you recommend as essential?

Source/Modules/csharp.cxx
repo-alt: this is a copy from the java module, no idea what it actually means, probably "is enabled?", not very useful in that case

Source/Modules/csharp.cxx
repo-alt: it is also no way to set it in Java module, but it may be set at runtime while debugging

Source/Modules/csharp.cxx
repo-alt: It depends on the translator, just default being verbatim copy. I mean, technically C# module knows that the translator is doing nothing only in module constructor (that may change if someone will introduce an alternative).

wkalinin:
This is actually very old code, added when doxygen was just released in it's own branch. I'll fix the warning and remove the useless comments
I have no idea what comments may be translated to, our documentation tool supports Doxygen in C# just fine, so I needed to only copy the comments.
There are quite a bit of doxygen tests in Java test suite, what test would you recommend as essential?
(Sorry for the spam, did not relogin into real account)

Source/Modules/csharp.cxx
wkalinin: It depends on the translator, just default being verbatim copy. I mean, technically C# module knows that the translator is doing nothing only in module constructor (that may change if someone will introduce an alternative).

Source/Modules/csharp.cxx
wkalinin: Blind copy from the Java module, probably means "if enabled", removed now.

Source/Modules/csharp.cxx
wkalinin: Maybe set from the debugger (just like in Java module)

ojwb:
I think it would be best to have all the doxygen testcases for each supported language.

While there are quite a few, it looks like most of the hard work for a new language is implementing something equivalent to `CommentParser` which gets hold of the output doxygen comments, normalises it, and compares with the expected version.

The rest is an essentially mechanical conversion of the testcases, and the bulk of that could be scripted.

emmenlau:
Doxygen support for C# would be great to have, so just kindly bumping this up again...

wsfulton:
Converting and adding the Java doxygen tests for this C# doxygen support is what is required for this to be included into SWIG. The testcases that need converting from Java to C# are: 
```
$ ls Examples/test-suite/java/doxygen_*runme.java
Examples/test-suite/java/doxygen_alias_runme.java
Examples/test-suite/java/doxygen_basic_notranslate_runme.java
Examples/test-suite/java/doxygen_basic_translate_runme.java
Examples/test-suite/java/doxygen_basic_translate_style2_runme.java
Examples/test-suite/java/doxygen_basic_translate_style3_runme.java
Examples/test-suite/java/doxygen_code_blocks_runme.java
Examples/test-suite/java/doxygen_ignore_runme.java
Examples/test-suite/java/doxygen_misc_constructs_runme.java
Examples/test-suite/java/doxygen_nested_class_runme.java
Examples/test-suite/java/doxygen_parsing_enums_proper_runme.java
Examples/test-suite/java/doxygen_parsing_enums_simple_runme.java
Examples/test-suite/java/doxygen_parsing_enums_typesafe_runme.java
Examples/test-suite/java/doxygen_parsing_enums_typeunsafe_runme.java
Examples/test-suite/java/doxygen_parsing_runme.java
Examples/test-suite/java/doxygen_translate_all_tags_runme.java
Examples/test-suite/java/doxygen_translate_links_runme.java
Examples/test-suite/java/doxygen_translate_runme.java
```
Converting Java code to C# code is normally quite easy.

emmenlau:
@wkalinin you have already done a lot for C# doxygen support! Would you be willing to push this further? I don't feel confident that I'm ready to start here, but doxygen support would be really nice to have in master...

emmenlau:
Dear @wkalinin, based on your nice work in this PR, I have continued with C# documentation support in #2421. My PR adds a few extensions on top of yours. I left your author name etc intact, and hope your appreciate my continued work.

Please let me know if this is in your best interest? Then we can possibly close one of the two merge requests and continue the collaborative effort in the remaining merge request. Your input and help is appreciated!

ojwb:
Closing in favour of #2421.

ojwb:
I've just merged #2835 which implemented support for this.  If there's anything useful this branch does which the merged branch doesn't, please open an issue (or better still submit a patch!)

PR #1696, opened 2020-01-05
Adding tp_vectorcall field to PyTypeObject for Python version 3.8

shadchin:


wsfulton:
Thanks, it was combined with #1670 and 489629fd532593cbf856a22990b520f770561984.

PR #1700, opened 2020-01-10
Check Py{Bytes,String}_AsStringAndSize() for failure

### LINKED ISSUE #1349: coverity: PyString_* functions' return value unchecked sometimes
```
/* Warning: This function will allocate a new string in Python 3,
 * so please call SWIG_Python_str_DelForPy3(x) to free the space.
 */
SWIGINTERN char*
SWIG_Python_str_AsChar(PyObject *str)
{
  char *newstr = 0;
  str = PyUnicode_AsUTF8String(str);
// 1. Condition str, taking true branch.
  if (str) {
    char *cstr;
    Py_ssize_t len;
// CID 187882 (#6 of 6): Unchecked return value (CHECKED_RETURN)
// 2. check_return: Calling Bytes_AsStringAndSize without checking return value (as is done elsewhere 198 out of 209 times).
    PyBytes_AsStringAndSize(str, &cstr, &len);
    newstr = (char *) malloc(len+1);
    memcpy(newstr, cstr, len+1);
    Py_XDECREF(str);
  }
  return newstr;
#else
  return PyString_AsString(str);
#endif
}
```
This function comes from `Lib/python/pyhead.swg`.

ZackerySpytz:
PyBytes_AsStringAndSize() and PyString_AsStringAndSize() were not
being checked for failure.

Closes #1349.

### DIFF
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -43,12 +43,11 @@ SWIG_Python_str_AsChar(PyObject *str)
   if (str) {
     char *cstr;
     Py_ssize_t len;
-    if (PyBytes_AsStringAndSize(str, &cstr, &len) == -1)
-      return NULL;
-    newstr = (char *) malloc(len+1);
-    if (!newstr)
-      return NULL;
-    memcpy(newstr, cstr, len+1);
+    if (PyBytes_AsStringAndSize(str, &cstr, &len) != -1) {
+      newstr = (char *) malloc(len+1);
+      if (newstr)
+        memcpy(newstr, cstr, len+1);
+    }
     Py_XDECREF(str);
   }
   return newstr;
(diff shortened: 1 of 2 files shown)


Lib/python/pyhead.swg
```diff
@@ -43,8 +43,11 @@ SWIG_Python_str_AsChar(PyObject *str)
   if (str) {
     char *cstr;
     Py_ssize_t len;
-    PyBytes_AsStringAndSize(str, &cstr, &len);
+    if (PyBytes_AsStringAndSize(str, &cstr, &len) == -1)
+      return NULL;
     newstr = (char *) malloc(len+1);
+    if (!newstr)
+      return NULL;
```
wsfulton: Please change to return only at the end of the function to void a leak due to `Py_XDECREF(str)` possibly not being called.

PR #1705, opened 2020-01-20
Add missing preprocessor defines to the docs list

nightlark:
Searching through the history, these are the defines I found that aren't mentioned in the documentation and didn't find a reference to deprecation:
```
SWIGD
SWIG_D_VERSION
SWIGGO
SWIGGO_CGO
SWIGGO_GCCGO
SWIGGO_INTGO_SIZE
BUILDING_NODE_EXTENSION
SWIGPYTHON_PY3
SWIGPYTHON_BUILTIN
SWIG_RUBY_AUTORENAME
```

These preprocessor defines had a commit saying they were deprecated, but I didn't see them in a commit or changelog saying they were removed:
```
SWIGPERL5
SWIGTCL8
```

These preprocessor defines all have a commit or comment in the change log saying they were removed:
```
SWIGALLEGROCL
SWIGCFFI
SWIGCHICKEN
SWIGCLISP
SWIGSEXP
SWIGPIKE
SWIGMODULA3
SWIGUFFI
```

### DIFF
--- a/Doc/Manual/Preprocessor.html
+++ b/Doc/Manual/Preprocessor.html
@@ -121,7 +121,6 @@ SWIGJAVA                        Defined when using Java
 SWIGJAVASCRIPT                  Defined when using Javascript
 SWIG_JAVASCRIPT_JSC             Defined when using Javascript for JavascriptCore
 SWIG_JAVASCRIPT_V8              Defined when using Javascript for v8 or node.js
-BUILDING_NODE_EXTENSION         Defined when using Javascript for node.js
 SWIGLUA                         Defined when using Lua
 SWIGMZSCHEME                    Defined when using Mzscheme
 SWIGOCAML                       Defined when using OCaml
--- a/Doc/Manual/Preprocessor.html
+++ b/Doc/Manual/Preprocessor.html
@@ -121,6 +121,7 @@ SWIGJAVA                        Defined when using Java
 SWIGJAVASCRIPT                  Defined when using Javascript
 SWIG_JAVASCRIPT_JSC             Defined when using Javascript for JavascriptCore
 SWIG_JAVASCRIPT_V8              Defined when using Javascript for v8 or node.js
+SWIG_JAVASCRIPT_NODE_V8         Defined when using Javascript for node.js
 SWIGLUA                         Defined when using Lua
 SWIGMZSCHEME                    Defined when using Mzscheme
 SWIGOCAML                       Defined when using OCaml
--- a/Doc/Manual/Preprocessor.html
+++ b/Doc/Manual/Preprocessor.html
@@ -110,11 +110,18 @@ SWIG_VERSION                    Hexadecimal (binary-coded decimal) number contai
                                 such as 0x010311 (corresponding to SWIG-1.3.11).
 
 SWIGCSHARP                      Defined when using C#
+SWIGD                           Defined when using D
+SWIG_D_VERSION                  Unsigned integer target version when using D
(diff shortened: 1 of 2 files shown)


vadz:
Thanks, FWIW I agree that these symbols should be documented.

nightlark:
Any suggested changes?

nightlark:
@vadz any additional changes you want to see made to this PR?

vadz:
No, this looks good to me, thanks, but normally PRs are merged by @wsfulton, so it's not really my opinion which counts.

nightlark:
Okay, thanks! I'll keep an eye out for a review by wsfulton with what changes are needed.

Doc/Manual/Preprocessor.html
```diff
@@ -110,11 +110,18 @@ <H2><a name="Preprocessor_condition_compilation">10.3 Conditional Compilation</a
...
+SWIGGO_CGO                      Defined when using Go for cgo
+SWIGGO_GCCGO                    Defined when using Go for gccgo
+SWIGGO_INTGO_SIZE               Size of the Go type int when using Go (32 or 64)
 SWIGGUILE                       Defined when using Guile
 SWIGJAVA                        Defined when using Java
 SWIGJAVASCRIPT                  Defined when using Javascript
 SWIG_JAVASCRIPT_JSC             Defined when using Javascript for JavascriptCore
-SWIG_JAVASCRIPT_V8              Defined when using Javascript for v8 or node.js 
+SWIG_JAVASCRIPT_V8              Defined when using Javascript for v8 or node.js
+BUILDING_NODE_EXTENSION         Defined when using Javascript for node.js
```
ojwb: This isn't a SWIG-specific define - it's something node requires us to define, not something we should be documenting here:

https://github.com/nodejs/node/blob/master/src/node_api.h#L4

ojwb: Hmm, except it's indeed defined for the SWIG preprocessor, not just when the generated wrapper code is being built.

Having a define to specify that we're generating bindings for node.js rather than v8 without node.js seems useful, but it really should be suitably namespaced - e.g. `SWIG_JAVASCRIPT_NODE` or  `SWIG_JAVASCRIPT_NODE_V8` something like that.

nightlark: Do you want a `SWIG_JAVASCRIPT_NODE*` SWIG preprocessor define added as an alias for `BUILDING_NODE_EXTENSION`, or should `BUILDING_NODE_EXTENSION` also get changed so it's only defined when the generated wrapper code is built?

vadz: Changing the existing define could silently break the currently working code, so I wouldn't do this. IMO defining this symbol shouldn't be a problem in practice, even if it's not prefixed by `SWIG_` -- code dealing with Node should be aware of this symbol anyhow, presumably.

nightlark: I removed `BUILDING_NODE_EXTENSION` from the docs and added a `SWIG_JAVASCRIPT_NODE_V8` preprocessor define.

wsfulton:
Thanks. I've taken some of these changes, tweaked and added some corrections, see f97b37a3162c15eca54d2171c36c75125ad3754e...ae0efd3d742ad083312fadbc652d4d66fdad6f4d.

I've removed the additional Javascript macro as Javascript is probably going to change quite a lot in the major release 4.1, so any new macros can be added in with this work, eg in #1520.

PR #1594, opened 2019-07-18
D: Replace deprecated imports (package std.c)

msteinbeck:
This PR fixes #1593. Tested with DMD (v2.086.1), GDC (8.2.0), and LDC (1.16.0).

PR #1709, opened 2020-01-24
D: Replace deprecated imports (package std.c). Fixes #1593

### LINKED ISSUE #1593: D: Replace broken imports in generated source code
With newer versions of dmd (I'm using v2.086.1) the generated source code is not compilable anymore due to broken imports. The following two imports must be replaced:

- `static import std.c.string;` -> `static import core.stdc.string;`
- `import std.c.linux.linux;` -> `import core.sys.posix.dlfcn;`

msteinbeck:
This PR replaces #1594.

I tried to fix the remaining errors but I have no experience with D. However, I was able to locate the file (and statement) that causes one of the errors. `Source/Modules/d.cxx`: 

```cpp
// Helper function to determine if a method has been overridden in a
// subclass of the wrapped class. If not, we just pass null to the
// director_connect_function since the method from the C++ class should
// be called as usual (see above).
// Only emit it if the proxy class has at least one method.
if (first_class_dmethod < curr_class_dmethod) {
    Printf(proxy_class_body_code, "\n");
    Printf(proxy_class_body_code, "private bool swigIsMethodOverridden(DelegateType, FunctionType, alias fn)() %s{\n", (d_version > 1) ? "const " : "");
    Printf(proxy_class_body_code, "  DelegateType dg = &fn;\n");
    Printf(proxy_class_body_code, "  return dg.funcptr != SwigNonVirtualAddressOf!(FunctionType, fn);\n");
    Printf(proxy_class_body_code, "}\n");
    Printf(proxy_class_body_code, "\n");
    Printf(proxy_class_body_code, "private static Function SwigNonVirtualAddressOf(Function, alias fn)() {\n");
    Printf(proxy_class_body_code, "  return cast(Function) &fn;\n");
    Printf(proxy_class_body_code, "}\n");
}
```

The parameter `alias fn` cannot be called because the method `swigIsMethodOverridden` is `const`. Maybe an experienced D programmer knows how to fix this.

### DIFF
--- a/Lib/d/wrapperloader.swg
+++ b/Lib/d/wrapperloader.swg
@@ -40,9 +40,10 @@ private {
   } else {
     version(D_Version2) {
       static import std.conv;
+    } else {
+      static import std.c.string;
     }
     static import std.string;
-    static import std.c.string;
   }
 
   version(D_Version2) {
@@ -112,7 +113,7 @@ private {
     version(Tango) {
       import tango.sys.Common;
     } else version(linux) {
-      import std.c.linux.linux;
+      import core.sys.posix.dlfcn;
     } else {
       extern(C) {
         const RTLD_NOW = 2;


buckle2000:
When will this get approved

wsfulton:
ping @dnadlinger , the D maintainer for approval.

dnadlinger:
Apologies; I only rarely keep an eye on SWIG PRs these days – please feel free to ping me more aggressively. (It would be great if somebody would take it on themselves to develop D support actively.)

I've updated the commit with a better fix that doesn't break D1 (not that it would matter these days) – feel free to merge once green.

PR #1713, opened 2020-01-30
Enable nullable reference types in C#

tillsch:
In C# 8 nullable reference types are [disabled in generated files](https://github.com/dotnet/roslyn/blob/70e158ba6c2c99bd3c3fc0754af0dbf82a6d353d/docs/features/nullable-reference-types.md#generated-code) by default, and generators should enable them by inserting `#nullable enable` or `#nullable restore` into each file.

I added the CLI option `-enablenullable` to insert `#nullable enable` into all generated C# files after the banner. This allows people to use nullable references types without breaking backward compatibility.

### DIFF
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -654,7 +654,7 @@ public:
     Printf(f, "//\n");
     Swig_banner_target_lang(f, "//");
     Printf(f, "//------------------------------------------------------------------------------\n\n");
-    if(enable_nullable_flag) {
+    if (enable_nullable_flag) {
       Printf(f, "#nullable enable\n\n");
     }
   }
@@ -4598,10 +4598,10 @@ extern "C" Language *swig_csharp(void) {
 const char *CSHARP::usage = "\
 C# Options (available with -csharp)\n\
      -dllimport <dl> - Override DllImport attribute name to <dl>\n\
+     -enablenullable - Enable nullable in all C# generated files\n\
      -namespace <nm> - Generate wrappers into C# namespace <nm>\n\
      -noproxy        - Generate the low-level functional interface instead\n\
                        of proxy classes\n\
      -oldvarnames    - Old intermediary method names for variable wrappers\n\
      -outfile <file> - Write all C# into a single <file> located in the output directory\n\
-     -enablenullable - Enable nullable in all C# generated files\n\
 \n";
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -1,4 +1,4 @@
-/* -----------------------------------------------------------------------------
+ /* -----------------------------------------------------------------------------
  * This file is part of SWIG, which is licensed as a whole under version 3 
  * (or any later version) of the GNU General Public License. Some additional
  * terms also apply to certain portions of SWIG. The full details of the SWIG
@@ -46,6 +46,7 @@ class CSHARP:public Language {
   bool global_variable_flag;	// Flag for when wrapping a global variable
   bool old_variable_names;	// Flag for old style variable names in the intermediary class
   bool generate_property_declaration_flag;	// Flag for generating properties
+  bool enable_nullable_flag;
 
   String *imclass_name;		// intermediary class name
   String *module_class_name;	// module class name
@@ -122,6 +123,7 @@ public:
       global_variable_flag(false),
       old_variable_names(false),
       generate_property_declaration_flag(false),
+      enable_nullable_flag(false),
       imclass_name(NULL),
       module_class_name(NULL),
       imclass_class_code(NULL),
@@ -254,6 +256,9 @@ public:
 	} else if (strcmp(argv[i], "-oldvarnames") == 0) {
 	  Swig_mark_arg(i);
 	  old_variable_names = true;
+	} else if (strcmp(argv[i], "-enablenullable") == 0) {
+	  Swig_mark_arg(i);
+	  enable_nullable_flag = true;
 	} else if (strcmp(argv[i], "-outfile") == 0) {
 	  if (argv[i + 1]) {
 	    output_file = NewString("");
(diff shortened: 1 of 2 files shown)


Source/Modules/csharp.cxx
```diff
@@ -4595,4 +4603,5 @@ C# Options (available with -csharp)\n\
                        of proxy classes\n\
      -oldvarnames    - Old intermediary method names for variable wrappers\n\
      -outfile <file> - Write all C# into a single <file> located in the output directory\n\
+     -enablenullable - Enable nullable in all C# generated files\n\
```
wsfulton: New option is not in alphabetical order

Source/Modules/csharp.cxx
```diff
@@ -649,6 +654,9 @@ class CSHARP:public Language {
     Printf(f, "//\n");
     Swig_banner_target_lang(f, "//");
     Printf(f, "//------------------------------------------------------------------------------\n\n");
+    if(enable_nullable_flag) {
```
wsfulton:
```suggestion
    if (enable_nullable_flag) {
```
Code style fix

Source/Modules/csharp.cxx
```diff
@@ -254,6 +256,9 @@ class CSHARP:public Language {
 	} else if (strcmp(argv[i], "-oldvarnames") == 0) {
 	  Swig_mark_arg(i);
 	  old_variable_names = true;
+	} else if (strcmp(argv[i], "-enablenullable") == 0) {
```
wsfulton: Please document this new option in CSharp.html, including a sentence or two as to why it is needed (for those who have not heard of the preprocessor directive) and a link to the MS docs.

wsfulton:
Actually, can we not always generate `#nullable enable` conditional on c# version 8 or later ? Then there is no need for the new command line option. 

BTW, does the generated code simply fail to compile with c# > 8?

vadz:
FWIW I'd also prefer to just generate the directive unconditionally if possible, I don't want to have to modify my projects to pass a new option to SWIG (which wouldn't work with the older versions) and I don't see why would anybody else.

Source/Modules/csharp.cxx
tillsch: Updated!

tillsch:
Without `#nullable enable` it will only fail to compile if one explicitly uses the nullable features, such as the `?.` operator.

Depending on what you're currently doing with swig, enabling nullable reference types in the generated code will give you a bunch of warnings. After all, better warnings is one of the purposes of nullable reference types. When I tried it with our project, it was hundreds of warnings. That's why I thought making it opt-in would be better.
It is also reasonable to use C# without nullable reference types, so there should be some way to enable/disable it. But adding `#nullable restore` (defaulting back to your project settings) might work for most people, not sure.

Also, I fixed my mistakes. 😊

tillsch:
@wsfulton could you have another look? I fixed all the issues you mentioned. 😄

wsfulton:
The main concern I have in this patch is that it generates code in a very inflexible way. I'd prefer an approach where a user could add in this code with more flexibility. For example, they may want to support more than one C# version in which case the C# code won't compiler with C# < 8. Given C# has very limited preprocessor capabilities and no way to detect the version of C# supported during compile time, the solution would then be for a project to manually define a C# macro in the build system, such as `CSHARP_VER8` and then use this macro:

```c#
#if CSHARP_VER8
#nullable restore
#endif
```

The C# code is then distributable to work with different C#/.net framework versions. So we need a solution that a user can control the code generated at the beginning of each C# file. I'll think of the best SWIG approach to do this and then report back here.

GenuineAster:
Hi @wsfulton , I'm taking over from Till here. We are using a forked version of Swig internally and would ideally like to switch to mainline Swig as it makes for less internal dependencies for us to manage. What can I do to move this pull request forward? Have you had the opportunity to consider different approaches to this issue?

macmade:
Any news on this feature?

adj123:
Hi @wsfulton, last we heard on this you said you'd get back with some thoughts on the best approach here. Can we see movement on this PR, whether it's from yourself or a green light for the community to take it from here?

It's coming up to 4 years since NRTs have been released into the C# language and they're now a standard for well-maintained codebases; it's an active pain point for consumers of C# Swig libraries that they cannot give the nullability guidance they want to their consumers.

Is there a workaround that can be used in the meantime, eg. some directive to inject custom code into the header of every file? (I'm not too familiar with Swig but I imagine this could have been added before for custom `using` statement imports?)

FWIW I don't really understand the arguments above about attempting to infer whether to add NRTs based on the generated C# lang version - NRTs are an opt-in feature and it's perfectly possible for a project to also have them enabled on <C#8 and disabled on >=C#8 (indeed the latter is very common). There's no real way to know if they should be enabled in a project other than by explicit declaration by the project/swig generator, so the initial proposal of adding a flag to swig seems the most appropriate to me.

roetlich:
Hi @adj123, I'm the person who originally opened this PR (using a different Github account). I didn't see any way to do this within swig back then, that's why I forked SWIG instead :sweat:. But I guess in newer .Net versions `<Nullable>enable</Nullable>` in the .csproj file also works.

wsfulton:
I've earmarked this for the 4.2 release and will review before then.

adj123:
@wsfulton Fantastic, thanks

@roetlich Hi! I might have misunderstood you but isn't the reason this PR is necessary because the project-level annotation _isn't_ sufficient for generated files, as you mention in your opening description?

roetlich:
@adj yes, getting this merged would be a lot better in a lot of cases

roetlich:
@wsfulton I guess this PR was closed because the mutate company is being silly. Should I recreate it?

wsfulton:
I'm not sure why @mutate-ab closed this. Feel free to resubmit this pull request.

roetlich:
New PR is here: https://github.com/swig/swig/pull/2681

Issue #862, opened 2017-01-01
Use PyUnicode_AsUTF8AndSize() for Python >= 3.3


ojwb:
Currently typemaps for Python 3 have to allocate a temporary buffer to get UTF-8 data out of Python.  I noticed that Python 3.3 added API functions to allow access to a UTF-8 representation which is cached by the object - this has two significant benefits:

 * We don't need to allocate memory ourselves and have to worry about the allocation failing (though currently we don't - the result of `malloc()` isn't checked against `NULL` here) and then have to ensure the allocation is released again
 * Avoids pointless repeated conversions to UTF-8 of the same string, which will be a big speed-up in some cases

I had a quick go at converting, but the test-suite segfaulted so I'm opening a ticket so this doesn't just get forgotten.

Here are the API docs:

```
char* PyUnicode_AsUTF8AndSize(PyObject *unicode, Py_ssize_t *size)

    Return a pointer to the UTF-8 encoding of the Unicode object, and store the size of the encoded representation (in bytes) in size. The size argument can be NULL; in this case no size will be stored. The returned buffer always has an extra null byte appended (not included in size), regardless of whether there are any other null code points.

    In the case of an error, NULL is returned with an exception set and no size is stored.

    This caches the UTF-8 representation of the string in the Unicode object, and subsequent calls will return a pointer to the same buffer. The caller is not responsible for deallocating the buffer.

    New in version 3.3.

char* PyUnicode_AsUTF8(PyObject *unicode)

    As PyUnicode_AsUTF8AndSize(), but does not store the size.

    New in version 3.3.
```

wsfulton:
Merge of #1716 closes this.

PR #1716, opened 2020-02-01
Use PyUnicode_AsUTF8() for Python >= 3.3

ZackerySpytz:
See #862.

### DIFF
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -37,7 +37,9 @@
 SWIGINTERN char*
 SWIG_Python_str_AsChar(PyObject *str)
 {
-#if PY_VERSION_HEX >= 0x03000000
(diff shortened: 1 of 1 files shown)


ojwb:
This looks good as far as it goes, but there are also calls to `PyUnicode_AsUTF8String()` in `Lib/python/pyrun.swg` and `Lib/python/pystrings.swg`.

wsfulton:
The code in `Lib/python/pyrun.swg` doesn't require a `char *`, so I don't see any need to remove the use of `PyUnicode_AsUTF8String()` and replace it with `PyUnicode_AsUTF8()`.

I'm not entirely sure if `SWIG_AsCharPtrAndSize` is worth changing to remove the use of `PyUnicode_AsUTF8String()`, but suggest it is done in a separate patch if necessary.

PR #1718, opened 2020-02-03
documentation about R accessors, with examples

richardbeare:
Short extension to the R section of the documentation, with an example of accessors and how
they are called from R.

Mentioned the use of finalizers and support for stl objects.


wsfulton:
Nice. I'll merge in shortly.

Issue #692, opened 2016-06-01
Please add support for unique_ptr


ashwin:
There is support for auto_ptr in v3.0.9, but no support yet for unique_ptr.

myd7349:
Which target languages are you using SWIG for?

ashwin:
Python

AnushS4t1qbit:
This is definitely a popular feature. 

I am a little wary of volunteering to tackle this, since I am sure there is a good reason why this hasn't been implemented yet. 

If I were to go about making this happen though, where would be a good starting point, the implementation for `shared_ptr`?

ojwb:
> I am sure there is a good reason why this hasn't been implemented yet

I doubt that reason is complexity though.  SWIG is developed by volunteers, and there are always more things that could be done than there is time to do them.

`std::auto_ptr` might be a better starting point, as `std::unique_ptr` is perhaps more akin to that.

wsfulton:
See Lib/python/std_auto_ptr.i. As unique_ptr is so similar to auto_ptr,
there isn't much to change. However, there is only support for outputs, not
input. Probably the input ('in') typemap should check that the reference
count is exactly one and then take over ownership from python.

lichti81:
I would be happy that there is also support for unique_ptr in C#

KrisThielemans:
Are there any updates on this? We're also interested in Python. I realise that doing this an input seems quite hard. It might need a copy of the underlying object as Python etc users will not expect their objects to disappear I guess.

There's an interesting approach on [stackoverflow](https://stackoverflow.com/a/27699663/3467846)

vadz:
FWIW #1722 adds support for `unique_ptr<>` in the "output" direction. I don't see much sense in doing it in the "input" direction but maybe I'm missing something.

KrisThielemans:
Great news about the PR. Do you want to continue the discussion there?

My understanding of input vs output:
- output: `unique_ptr<Obj> create_obj();`
- input: `void use_obj(unique_ptr<Obj>);`

We're currently debating the latter on https://github.com/UCL/STIR/issues/485 where we try to find a safe way to let a class store a ptr. Our current thinking goes like this
```c++
class A
{
  private:
    shared_ptr<Obj> sptr;
  public:
    A(unique_ptr<Obj> p):  sptr(p)
    {}
   shared_ptr<const Obj> get_sptr() const { return sptr;}
};
```
In Python (or whatever wrapped language), we'd like to be able to do
```python
[...]

PR #1722, opened 2020-02-11
Allow using auto_ptr<> typemaps for other similar types too

vadz:
Generalize %auto_ptr macro so that it could be used with other types,
e.g. std::unique_ptr<> or even boost::unique_ptr<> using the same
approach as is already used for std::shared_ptr<>, i.e. by allowing to
predefine the class name and/or namespace different from the default one
before including std_auto_ptr.i.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.2 (in progress)
 ===========================
 
+2020-02-11: vadz
+            [C#, Java, Python, Ruby] Allow using auto_ptr<> typemaps for unique_ptr<> too.
+
 2020-02-06: wsfulton
             [Python] #1673 #1674 Fix setting 'this' when extending a proxy class with __slots__.
 
--- a/Doc/Manual/Library.html
+++ b/Doc/Manual/Library.html
@@ -2007,6 +2007,16 @@ directly supported.
 </p>
 
 <p>
+Moreover, the same support can be also used for the (non-deprecated)
+<tt>std::unique_ptr</tt>. To use it, just predefine
+<tt>SWIG_AUTO_PTR_CLASSNAME</tt> as <tt>unique_ptr</tt> <em>before</em>
+including <tt>std_auto_ptr.i</tt> file. For completeness, notice that it's
+also possible to predefine <tt>SWIG_AUTO_PTR_NAMESPACE</tt> to something
+different from its default <tt>std</tt> value, e.g. to use
+<tt>boost::unique_ptr</tt> class.
+</p>
+
+<p>
 A typical example of use would be
 </p>
 <div class="code">
--- a/Examples/test-suite/li_std_auto_ptr.i
+++ b/Examples/test-suite/li_std_auto_ptr.i
@@ -1,43 +1,36 @@
 %module li_std_auto_ptr
 
-%{
-#if __GNUC__ >= 5 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 8)
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations" // auto_ptr deprecation
-#endif
-
-#if defined(__clang__)
-#pragma clang diagnostic push
-// Suppress 'auto_ptr<>' is deprecated
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-#endif
-%}
-
 #if defined(SWIGCSHARP) || defined(SWIGJAVA) || defined(SWIGPYTHON) || defined(SWIGRUBY)
 
+// Prefer using std::unique_ptr<> if it's available, as auto_ptr<> generates
+// deprecation warnings with C++14 compilers and may not exist at all when
+// using C++17.
+//
+// As we can't test for C++11 support at SWIG time, we use an extra level of
+// indirection, with SwigTest::auto_ptr being defined as either auto_ptr<> or
+// unique_ptr<> below.
+#define SWIG_AUTO_PTR_NAMESPACE SwigTest
 %include "std_auto_ptr.i"
 
 %auto_ptr(Klass)
 
 %{
-#if __cplusplus < 201703L
 #include <memory>
-#else
-// Simple std::auto_ptr implementation for testing after its removal in C++17
-namespace std {
-  template <class T> class auto_ptr {
-    T *ptr;
-    public:
-      auto_ptr(T *ptr = 0) : ptr(ptr) {}
-      auto_ptr(auto_ptr&& a) : ptr(a.ptr) { a.ptr = 0;}
-      ~auto_ptr() { delete ptr; }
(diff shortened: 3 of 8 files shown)


wsfulton:
Adding `std::unique_ptr` support is great. I think the support for it should be more conventional though.

Normally the STL libraries are available via a %include of the appropriate library, so in this case it should be `%include <std_unique_ptr.i>` rather than the somewhat quirky and unusual approach provided by this patch (defining `SWIG_AUTO_PTR_CLASSNAME` and then including a different std library file). 

This 'fix' could be done fairly easily be using the majority of the patch and adding in the macros inside of `std_unique_ptr.i`. While I understand that the approach taken is based on the shared_ptr for boost and std, I would rather see the std_auto_ptr.i files simply copied into std_unique_ptr.i and appropriately modified for unique_ptr and maintained separately. There was (and possibly is) still a case to support `boost::shared_ptr` as well as `std::shared_ptr`, but I don't see a need to support `boost::unique_ptr`.

Regards the tests, I would create different tests for different shared pointers and only test `unique_ptr` if compiler is > c++11 and do the opposite for `auto_ptr`.

vadz:
Thanks for looking at this and sorry if I'm misunderstanding you, but I really don't want to duplicate all `auto_ptr.i` files as `unique_ptr.i` if this is what you propose. IMO the last thing SWIG typemaps library needs is more duplication.

I can, of course, add `unique_ptr.i`, and I agree that it's a better idea, but it would still be implemented via inclusion of `auto_ptr.i` internally. I don't see why should SWIG users care about this, so hopefully this would be acceptable, but please let me know if you agree before I do this.

Concerning the tests, I'm also not sure what's the problem with the current approach. I could add C++11-only tests for `unique_ptr`, but we don't currently have a way to build some tests for only C++ < 17, and it doesn't seem worth to add them just to have `auto_ptr` tests. IMO the current test provides good test coverage at rather low cost.

lichti81:
Is with this issue #692 solved?

lichti81:
I agree with @wsfulton .
I also think it would be better if the various ptr wrappers were independent of each other.
If they can only be distinguished by magic definitions you can't use them in combination.
Especially with auto_ptr and unique_ptr it can happen that they are mixed in one header file. Especially with classes that have grown historically. Older methods still work with auto_ptr, newer methods already work with unique_ptr.

Furthermore I am not sure if auto_ptr and unique_ptr can be mapped equally.
I have not looked at the typemap but the behaviour of the two ptr in C++ is not identical. The main difference is that with unique_ptr the ownership is passed. See for example also here: https://stackoverflow.com/questions/3451099/stdauto-ptr-to-stdunique-ptr

If you ask me, I believe that you cannot hide this ownership semantics from the user.
At least in C# there is no such logic.
But I could imagine that in C# a unique_ptr classe is provided as well. A reference to the inner object is only available if you have the ownership, otherwise it is null.
It is also important that when you delete the C# object the C++ object is not deleted as well. The delete must always be done by the unique_ptr (in C++ or C#) which has the ownership.

vadz:
Just FYI: I'm not sure if it's worth doing anything here before @wsfulton can address the questions from my [previous comment](https://github.com/swig/swig/pull/1722#issuecomment-592950617) as I could be misunderstanding what needs to be done here. So for now I'll just keep using this in my own code and while I'd still be ready to spend time on integrating this in SWIG proper, I don't want to waste both your and my time on doing something before understand what this something is, exactly.

KrisThielemans:
There doesn't seem to be a way to upvote a PR, so I'll do it via this comment (sorry for the noise). Being able to wrap `unique_ptr` is essential for many projects, including ours.

wsfulton:
Oops, sorry I missed that this is waiting for my response. For clarification:

* Yes, copy and modify std_auto_ptr.i resulting in std_unique_ptr.i (that does not include std_auto_ptr.i).
* With regard to testing, it is challenging writing tests for different standards, but possible...
  * Testing for c++>=11 can be achieved by adding a cpp11_std_unique_ptr.i test and add it to CPP11_TEST_CASES in common.mk.
  * No need to change li_std_auto_ptr.i as it already has some simple hacks for testing in different C++ standards and works for c++<17.

@lichti81, once full support is added (that is, 'in' typemaps for function parameters), then yes, I think you are correct that the implementation will be different for these two smart pointers. To be honest, I think the full implementation for inputs is going to be challenging as the proxy class will no longer hold ownership to the object.

vadz:
As I wrote before:

> I really don't want to duplicate all `auto_ptr.i` files as `unique_ptr.i` if this is what you propose. IMO the last thing SWIG typemaps library needs is more duplication.

Anybody should feel free to create the new `std_unique_ptr.i` duplicating `std_auto_ptr.i` if they feel like it, of course, but I just have no motivation in doing something I consider not just useless but actively counter-productive. I'm not closing this PR in case it can be useful to the others, but neither do I plan to do anything about it, so please feel free to close it yourself if you're adamant about doing it like this.

KrisThielemans:
I have virtually no experience with SWIG development and typemaps, but it seems to me that if the suggested modifications work for both `auto_ptr` and `unique_ptr` without code duplication, there's value in that. I am not sure about @lichti81's point that the current scheme might run into trouble when using both types (wouldn't you be able to include the file twice with 2 different `define SWIG_AUTO_PTR_CLASSNAME`?).

I also agree that this is something that the user shouldn't see. Especially if the support needs to diverge later (e.g. by only adding `in` typemaps for `unique_ptr`).

@wsfulton what about moving the modified files to something like `*_auto_unique_ptr*` which would be internal, and having `std_auto_ptr.i` and `std_unique_ptr.i` et al set `SWIG_AUTO_PTR_CLASSNAME` and include the relevant file. This avoids copying content, and is future proof. (It also seems the less amount of work 😄 ).

The only other way that I see to avoid copying the code is by using macros, but that seems very hard to get right.

I have no clue about the testing. sorry.

lichti81:
I'm pretty sure that for a full support of `std::unique_ptr` a other kind of typemap is needed. A unique_ptr must be seen more like a container than a pointer. This means the usage is than similar to vector. Example:
`%template(MyTypeUniquePtr) unique_ptr<MyType>;`

Which creates also a class `MyTypeUniquePtr` in the proxy.
And the memory management is than fully inside the container implementation, which allows to use it also for **in** parameter/arguments.
When a managed object is pushed into the unique_ptr-container the container takes the ownership. When container fall out of scope, the managed part of the inner object is deleted by the garbage collector the unmanaged part by the C++ unique_ptr in the destructor of the container. When the unique_ptr in C++ moves the content to a other unique_ptr the pointer of the original container is set to nullptr and nothing is deleted.

Unfortunately I'm not familiar enough in writing typedefs to create such a typemap.
I hope somebody of the community has the needed skills.

KrisThielemans:
I certainly am not qualified to comment on that. However, I think this PR has working functionality which adds minimal but sorely needed functionality to SWIG. Until there's something better, I'd vote for converting it to something that works for @wsfulton and merging it in 😉

PR #1725, opened 2020-02-14
Catch exceptions by reference rather than by value.

dmach:
Fixes -Wcatch-value gcc warnings.


wsfulton:
For some reason, my gcc-8 and gcc-9 on Ubuntu doesn't accept `-Wcatch-value` as an option. Nevertheless, thanks, looks fine to me. I have changed to catch these by const reference instead of non-const reference though.

PR #1726, opened 2020-02-16
Fix reading options files on platforms with unsigned char

vadz:
This fixes EOF detection on platforms where char is unsigned, as
comparing it with EOF could never return true there.

Thanks gcc for the warning "comparison is always true due to limited
range of data type [-Wtype-limits]".


wsfulton:
This warning is seen in the s390x architecture, see #1803.

PR #1727, opened 2020-02-16
Make Python buffer typemaps compatible with limited API

vadz:
Use deprecated PyObject_As{ReadmWrite}Buffer() functions instead of
PyObject_GetBuffer() not available in limited API. Even though these
functions are deprecated and are not supposed to be used, using them
should be fine for now (up to Python 3.8) and, hopefully, buffer API
will be stabilized before they are removed in a future Python version.

---

This is another fix I needed for one of my projects after applying #1613 (in addition to #1687). Using of deprecated functions is not great, but it's still incomparably better than nothing and I'm pretty sure there is no other possible solution with the current state of Python limited ABI (the [problem is known](https://bugs.python.org/issue10181) since  10 years already without anything happening for the last 8 or so).

### DIFF
--- a/Lib/python/pybuffer.i
+++ b/Lib/python/pybuffer.i
@@ -12,14 +12,36 @@
  *      }
  */
 
+/* Note that in Py_LIMITED_API case we have no choice, but to use deprecated
+ * functions, as they provides the only way to access buffer data with limited
+ * API, which doesn't include Py_buffer definition. We also disable the
+ * warnings about doing this because they're not useful in our case.
+ */
+
 %define %pybuffer_mutable_binary(TYPEMAP, SIZE)
 %typemap(in) (TYPEMAP, SIZE) {
   int res; Py_ssize_t size = 0; void *buf = 0;
+%#ifdef Py_LIMITED_API
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic push
+    %#pragma GCC diagnostic ignored "-Wdeprecated"
+  %#elif defined(_MSC_VER)
+    %#pragma warning(push)
+    %#pragma warning(disable: 4996)
+  %#endif
+  res = PyObject_AsWriteBuffer($input, &buf, &size);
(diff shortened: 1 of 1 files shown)


vadz:
Part of #2190 now.

wsfulton:
Seriously, introducing deprecated methods!?? If the limited ABI doesn't have the functionality, then either don't use this functionality with the limited ABI or work with the Python community to find an alternative.

vadz:
I don't know why are you looking at this one when I closed it and we've been discussing this in #2190 already, but I think you're missing the point: these functions are part of the stable ABI and so it absolutely doesn't matter that they're marked as deprecated in the header (which is, of course, a mistake by Python developers, but this is neither here nor there) because they can never be removed and it's perfectly fine to use them.

I.e. the limited ABI does have this functionality and it's the only way to work with the buffers when using it.

So I have trouble understanding your point here, but, for the record, yes, I'm completely serious. Advising not to use buffers (which is, as you certainly know, the only reasonably efficient way to exchange big amounts of data with Python) just because these functions have been marked with a deprecation macro in the headers doesn't seem very serious to me, however.

wsfulton:
I find it really hard to believe that the limited ABI contains deprecated methods given that the whole point of the ABI is it won't change. How does one remove the deprecated methods in something that won't change??! Complete nonsense! I just can't believe the Python devs have done this. I assume there must be something missing in this logic, but if not, then we'll have to go with it. It still feels like we should get clarification from the Python devs as to what the thinking here is and the solution. Are there any useful links around this catch-22?

vadz:
Python developers are aware of the problem and, as mentioned [here](https://github.com/swig/swig/pull/2190#discussion_r799024887), have just finally solved it 3 days ago by stabilizing the new buffer API. So in the future we will have the possibility to switch to the new API once it becomes reasonable to require Python 3.11.

But today (and for a few more years I'm afraid) these functions remain the only way to access the buffers when using stable ABI. And, as I think I already mentioned, it's not clear if they're ever going to be removed at all, considering the whole point of the ABI stability guarantee.

And yes, all this doesn't make much sense but it is what it is. I have no idea how/why did it happen, sorry.

PR #1728, opened 2020-02-17
Add support for LUA lightuserdata to SWIG_Lua_ConvertPtr 

ryannevell:
SWIG_Lua_ConvertPtr was missing a check if the given object was of lightuserdata type. It would attempt to unwrap it as a regular userdata type, which would result in an invalid pointer address being returned.

This change includes a test case to demonstrate the issue, and an update to SWIG_Lua_ConvertPtr to handle the lightuserdata case.


wsfulton:
We don't have a Lua maintainer at the moment, but from what I can see this is okay -  a lightuserdata type is just a wrapper around a C pointer. 

Thanks.

PR #1738, opened 2020-02-27
Scilab - make generated gateways compatible with 6.1.0

### LINKED ISSUE #1737: SWIG Scilab generated gateways are not compatible with Scilab 6.1.0
Scilab 6.1.0 introduced zero output argument detection which allows any gateway to have a special behavior without output argument. This effectively change the behavior of `CheckOutputArgument(pvApiCtx, 1, 1)` where the minimal number of output arguments could not be ignored (as it was the case before Scilab 6.1).

IMHO the code generator should be updated to only use `CheckOutputArgument(pvApiCtx, 0, 1)` everywhere.

davidcl:
This fix #1737 but using the correct 0 or 1 output argument syntax. For any function mapped `foo`, this will handle both:
```scilab
foo();
a = foo();
```

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.2 (in progress)
 ===========================
 
+2020-02-27: davidcl
+            [Scilab] #1737 makes generated gateway compatible with Scilab 6.1.0.
+
 2020-02-18: ryannevell
             [Lua] #1728 Add support for LUA lightuserdata to SWIG_Lua_ConvertPtr.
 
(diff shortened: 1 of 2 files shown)


wsfulton:
In the Travis CI testing above, the first example hangs when run, so I'm afraid there is something drastically broken when using this patch.

davidcl:
You are correct, there are some issues on the Scilab part of the build system/test runner; I started to investigate some of them and both Scilab 6.0 and Scilab 6.1 seem to be broken on my [travis-ci build](https://travis-ci.org/davidcl/swig/builds/656319433) with stock Scilab versions.

I will first add a pull request to the build system with 5.5.2, 6.0.2 and 6.1.0 builds before pushing modifications.

PR #1739, opened 2020-02-27
Scilab - fix #745 use SWIG_<module>_Init() function to init the module

### LINKED ISSUE #745: Swig + scilab creates Init function that conflicts with internal function name
Hi,

If you have a C file called foo.c with the function name

Foo_Init() {

}

And then you have a SWIG file defined as follows:

%module Foo

You will get an error because Scilab Swig will create a function called Foo_Init() and this causes a compilation error.

[example.zip](https://github.com/swig/swig/files/359534/example.zip)

davidcl:
This pull request Closes #745

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.0.2 (in progress)
 ===========================
 
+2020-02-27: davidcl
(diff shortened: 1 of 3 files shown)


ojwb:
This seems a clear improvement so while a review from someone who knows scilab would be good, I don't see it makes sense to block merging it any longer.

PR #1702, opened 2020-01-14
Upgrade SWIG to support NodeJS v12 / V8 v7.6, v8 v7.8 runtime

lissyx:
WIP, trying to get Travis to run ?

PR #1736, opened 2020-02-27
Remove support for old Node.js versions and stand-alone v8

yegorich:
Start removing support for stand-alone v8 as also Node.js older than v6.x.

Fixes https://github.com/swig/swig/issues/1735.

Issue #1735, opened 2020-02-27
Remove support for umaintained Node.js versions


yegorich:
Is there any plan to get rid of the old Node.js versions?

Latest Version | Release Date
------------ | -------------
0.10.48 | 2016-10-18
0.12.18 | 2017-02-22
4.9.1 | 2018-03-29
6.17.1 | 2019-04-03
8.17.0 | 2019-12-17

The background is, I'm trying to take over https://github.com/swig/swig/pull/1702. Getting rid of these old versions would bump v8 to 5.1.x thus, it will greatly simplify the code. For example, we could safely change all v8::Handle occurrences to v8::Local (see [here](https://www.electronjs.org/blog/nodejs-native-addons-and-electron-5) for more information).

I don't know how many people still use these old Node.js versions with SWIG. Hence, it is just a suggestion.

lissyx:
@yegorich I guess from the discussion on https://github.com/swig/swig/issues/1520 it would be fine.

yegorich:
But can this still be done for SWIG 4.x?

According to that discussion only v0.10 and v0.12 should be dropped. But can we also agree on removing v4 as well?

I am using [nvm](https://github.com/nvm-sh/nvm) to manage multiple Node.js versions and I have trouble installing node-gyp for versions older than v6.

@wsfulton @mochizk

yegorich:
If I look at Debian libv8 [packages](https://packages.debian.org/search?keywords=libv8), it looks like it will be replaced via Node.js. See [libnode-dev](https://packages.debian.org/buster/libnode-dev) package.

yegorich:
Do I also understand it correctly that if we remove support for the stand-alone v8 engine, we don't need `Tools/javascript/v8_shell.cxx`?

vadz:
FWIW we only use Node 10 or higher, and AFAICS dropping all versions before 8 would simplify the code even more. This could be too drastic, of course...

wsfulton:
Please see #1520 for a discussion on removal of older versions of Javascript. Removal of the older versions missed the boat for the swig-4.0.x series, so as it stands this patch can only be used for swig-4.1.0 and later. I intend to release-4.0.2 quite soon. Right now there is no target date for 4.1 though as we don't have many patches for the larger 4.1 release, see https://github.com/swig/swig/milestone/3. A patch to swig-4.0.1 for adding in support for newer Javascript versions that don't work, #1735, is very much encouraged though.

yegorich:
[...]

PR #1746, opened 2020-03-02
Prepare SWIG for Node.js v12

yegorich:
This PR replaces all `v8::Handle` occurrences to `v8::Local` for all Node.js releases where these both types are the same i.e. `v8::Handle` is just an alias to `v8::Local`.

It also introduces changes for new Node.js v12 related semantics.

This PR is based on https://github.com/swig/swig/pull/1702.

Fixes https://github.com/swig/swig/issues/1760, fixes https://github.com/swig/swig/issues/1520

### DIFF
--- a/Examples/javascript/native/example.i
+++ b/Examples/javascript/native/example.i
@@ -15,7 +15,7 @@ int placeholder() { return 0; }
     static SwigV8ReturnValue JavaScript_do_work(const SwigV8Arguments &args) {
         SWIGV8_HANDLESCOPE();
         const int MY_MAGIC_NUMBER = 5;
-        v8::Handle<v8::Value> jsresult =
+        SWIGV8_VALUE jsresult =
             SWIG_From_int(static_cast< int >(MY_MAGIC_NUMBER));
         if (args.Length() != 0)
             SWIG_exception_fail(SWIG_ERROR, "Illegal number of arguments.");
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -423,11 +423,14 @@ fail:
   SWIGV8_FUNCTION_TEMPLATE $jsmangledname_class_0 = SWIGV8_CreateClassTemplate("$jsname");
   $jsmangledname_class_0->SetCallHandler($jsctor);
   $jsmangledname_class_0->Inherit($jsmangledname_class);
-  $jsmangledname_class_0->SetHiddenPrototype(true);
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
+  $jsmangledname_class_0->SetHiddenPrototype(true);
   v8::Handle<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction();
-#else
+#elif (SWIG_V8_VERSION < 0x0705)
+  $jsmangledname_class_0->SetHiddenPrototype(true);
   v8::Local<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction();
+#else
+  v8::Local<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked();
 #endif
 %}
 
@@ -439,7 +442,12 @@ fail:
  * ----------------------------------------------------------------------------- */
 %fragment("jsv8_register_class", "templates")
 %{
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
   $jsparent_obj->Set(SWIGV8_SYMBOL_NEW("$jsname"), $jsmangledname_obj);
+#else
+  $jsparent_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("$jsname"), $jsmangledname_obj);
+#endif
+
 %}
 
 /* -----------------------------------------------------------------------------
@@ -459,7 +467,11 @@ fail:
  * ----------------------------------------------------------------------------- */
 %fragment("jsv8_register_namespace", "templates")
 %{
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
   $jsparent_obj->Set(SWIGV8_SYMBOL_NEW("$jsname"), $jsmangledname_obj);
+#else
+  $jsparent_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("$jsname"), $jsmangledname_obj);
+#endif
 %}
 
 /* -----------------------------------------------------------------------------
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -11,7 +11,7 @@
 static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args) {
   SWIGV8_HANDLESCOPE();
   
-  v8::Handle<v8::Object> self = args.Holder();
+  SWIGV8_OBJECT self = args.Holder();
   $jslocals
   if(args.Length() != $jsargcount) SWIG_exception_fail(SWIG_ERROR, "Illegal number of arguments for $jswrapper.");
   $jscode
@@ -53,7 +53,7 @@ static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args) {
   SWIGV8_HANDLESCOPE();
   
   OverloadErrorHandler errorHandler;
-  v8::Handle<v8::Value> self;
+  SWIGV8_VALUE self;
 
   // switch all cases by means of series of if-returns.
   $jsdispatchcases
@@ -78,7 +78,7 @@ fail:
 static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args, V8ErrorHandler &SWIGV8_ErrorHandler) {
   SWIGV8_HANDLESCOPE();
   
-  v8::Handle<v8::Object> self = args.Holder();
+  SWIGV8_OBJECT self = args.Holder();
   $jslocals
   if(args.Length() != $jsargcount) SWIG_exception_fail(SWIG_ERROR, "Illegal number of arguments for $jswrapper.");
   $jscode
@@ -226,7 +226,7 @@ static SwigV8ReturnValue $jswrapper(v8::Local<v8::Name> property, const SwigV8Pr
 #endif
   SWIGV8_HANDLESCOPE();
   
-  v8::Handle<v8::Value> jsresult;
+  SWIGV8_VALUE jsresult;
   $jslocals
   $jscode
   SWIGV8_RETURN_INFO(jsresult, info);
@@ -271,7 +271,7 @@ fail:
 static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args) {
   SWIGV8_HANDLESCOPE();
   
-  v8::Handle<v8::Value> jsresult;
+  SWIGV8_VALUE jsresult;
   $jslocals
   if(args.Length() != $jsargcount) SWIG_exception_fail(SWIG_ERROR, "Illegal number of arguments for $jswrapper.");
 
@@ -296,7 +296,7 @@ fail:
 static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args) {
   SWIGV8_HANDLESCOPE();
   
-  v8::Handle<v8::Value> jsresult;
+  SWIGV8_VALUE jsresult;
   OverloadErrorHandler errorHandler;
   $jscode
 
@@ -320,7 +320,7 @@ static SwigV8ReturnValue $jswrapper(const SwigV8Arguments &args, V8ErrorHandler 
 {
   SWIGV8_HANDLESCOPE();
   
-  v8::Handle<v8::Value> jsresult;
+  SWIGV8_VALUE jsresult;
   $jslocals
   $jscode
   SWIGV8_RETURN(jsresult);
@@ -374,7 +374,7 @@ fail:
 %fragment("jsv8_define_class_template", "templates")
 %{
   /* Name: $jsmangledname, Type: $jsmangledtype, Dtor: $jsdtor */
-  v8::Handle<v8::FunctionTemplate> $jsmangledname_class = SWIGV8_CreateClassTemplate("$jsmangledname");
+  SWIGV8_FUNCTION_TEMPLATE $jsmangledname_class = SWIGV8_CreateClassTemplate("$jsmangledname");
   SWIGV8_SET_CLASS_TEMPL($jsmangledname_clientData.class_templ, $jsmangledname_class);
   $jsmangledname_clientData.dtor = $jsdtor;
   if (SWIGTYPE_$jsmangledtype->clientdata == 0) {
@@ -420,11 +420,15 @@ fail:
 %fragment("jsv8_create_class_instance", "templates")
 %{
   /* Class: $jsname ($jsmangledname) */
-  v8::Handle<v8::FunctionTemplate> $jsmangledname_class_0 = SWIGV8_CreateClassTemplate("$jsname");
+  SWIGV8_FUNCTION_TEMPLATE $jsmangledname_class_0 = SWIGV8_CreateClassTemplate("$jsname");
   $jsmangledname_class_0->SetCallHandler($jsctor);
   $jsmangledname_class_0->Inherit($jsmangledname_class);
   $jsmangledname_class_0->SetHiddenPrototype(true);
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
   v8::Handle<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction();
+#else
+  v8::Local<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction();
+#endif
 %}
 
 /* -----------------------------------------------------------------------------
@@ -444,7 +448,7 @@ fail:
  * ----------------------------------------------------------------------------- */
 %fragment("jsv8_create_namespace", "templates")
 %{
-  v8::Handle<v8::Object> $jsmangledname_obj = SWIGV8_OBJECT_NEW();
+  SWIGV8_OBJECT $jsmangledname_obj = SWIGV8_OBJECT_NEW();
 %}
 
 /* -----------------------------------------------------------------------------
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -423,13 +423,12 @@ fail:
   SWIGV8_FUNCTION_TEMPLATE $jsmangledname_class_0 = SWIGV8_CreateClassTemplate("$jsname");
   $jsmangledname_class_0->SetCallHandler($jsctor);
   $jsmangledname_class_0->Inherit($jsmangledname_class);
-#if (SWIG_V8_VERSION < 0x0705)
-$jsmangledname_class_0->SetHiddenPrototype(true);
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
+  $jsmangledname_class_0->SetHiddenPrototype(true);
   v8::Handle<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction();
-#else
+#elif (SWIG_V8_VERSION < 0x0705)
+  $jsmangledname_class_0->SetHiddenPrototype(true);
   v8::Local<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction();
-#endif
 #else
   v8::Local<v8::Object> $jsmangledname_obj = $jsmangledname_class_0->GetFunction(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked();
 #endif
--- a/Lib/javascript/v8/javascriptinit.swg
+++ b/Lib/javascript/v8/javascriptinit.swg
@@ -68,16 +68,16 @@ SWIG_V8_GetModule(void *) {
 // TODO: is it ok to do that?
 extern "C"
 #if (NODE_MODULE_VERSION < 0x000C)
-void SWIGV8_INIT (v8::Handle<v8::Object> exports)
+void SWIGV8_INIT (SWIGV8_OBJECT exports)
 #else
-void SWIGV8_INIT (v8::Handle<v8::Object> exports, v8::Handle<v8::Object> /*module*/)
+void SWIGV8_INIT (SWIGV8_OBJECT exports, SWIGV8_OBJECT /*module*/)
 #endif
 {
   SWIG_InitializeModule(static_cast<void *>(&exports));
 
   SWIGV8_HANDLESCOPE();
   
-  v8::Handle<v8::Object> exports_obj = exports;
+  SWIGV8_OBJECT exports_obj = exports;
 %}
 
 
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -67,6 +67,11 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_NUMBER_NEW(num) v8::Number::New(num)
 #define SWIGV8_OBJECT_NEW() v8::Object::New()
 #define SWIGV8_UNDEFINED() v8::Undefined()
+#define SWIGV8_ARRAY v8::Handle<v8::Array>
+#define SWIGV8_FUNCTION_TEMPLATE v8::Handle<v8::FunctionTemplate>
+#define SWIGV8_OBJECT v8::Handle<v8::Object>
+#define SWIGV8_OBJECT_TEMPLATE v8::Handle<v8::ObjectTemplate>
+#define SWIGV8_VALUE v8::Handle<v8::Value>
 #define SWIGV8_NULL() v8::Null()
 #else
 #define SWIGV8_ARRAY_NEW() v8::Array::New(v8::Isolate::GetCurrent())
@@ -80,6 +85,11 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_NUMBER_NEW(num) v8::Number::New(v8::Isolate::GetCurrent(), num)
 #define SWIGV8_OBJECT_NEW() v8::Object::New(v8::Isolate::GetCurrent())
 #define SWIGV8_UNDEFINED() v8::Undefined(v8::Isolate::GetCurrent())
+#define SWIGV8_ARRAY v8::Local<v8::Array>
+#define SWIGV8_FUNCTION_TEMPLATE v8::Local<v8::FunctionTemplate>
+#define SWIGV8_OBJECT v8::Local<v8::Object>
+#define SWIGV8_OBJECT_TEMPLATE v8::Local<v8::ObjectTemplate>
+#define SWIGV8_VALUE v8::Local<v8::Value>
 #define SWIGV8_NULL() v8::Null(v8::Isolate::GetCurrent())
 #endif
 
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -9,8 +9,10 @@
 
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031803)
 #define SWIGV8_STRING_NEW2(cstr, len) v8::String::New(cstr, len)
-#else
+#elif (SWIG_V8_VERSION < 0x0706)
 #define SWIGV8_STRING_NEW2(cstr, len) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), cstr, v8::String::kNormalString, len)
+#else
+#define SWIGV8_STRING_NEW2(cstr, len) (v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), cstr, v8::NewStringType::kNormal, len)).ToLocalChecked()
 #endif
 
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
@@ -47,12 +49,18 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_THROW_EXCEPTION(err) v8::ThrowException(err)
 #define SWIGV8_STRING_NEW(str) v8::String::New(str)
 #define SWIGV8_SYMBOL_NEW(sym) v8::String::NewSymbol(sym)
+#elif (SWIG_V8_VERSION < 0x0706)
+#define SWIGV8_ADJUST_MEMORY(size) v8::Isolate::GetCurrent()->AdjustAmountOfExternalAllocatedMemory(size)
+#define SWIGV8_CURRENT_CONTEXT() v8::Isolate::GetCurrent()->GetCurrentContext()
+#define SWIGV8_THROW_EXCEPTION(err) v8::Isolate::GetCurrent()->ThrowException(err)
+#define SWIGV8_STRING_NEW(str) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), str, v8::String::kNormalString)
+#define SWIGV8_SYMBOL_NEW(sym) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), sym, v8::String::kNormalString)
 #else
 #define SWIGV8_ADJUST_MEMORY(size) v8::Isolate::GetCurrent()->AdjustAmountOfExternalAllocatedMemory(size)
 #define SWIGV8_CURRENT_CONTEXT() v8::Isolate::GetCurrent()->GetCurrentContext()
 #define SWIGV8_THROW_EXCEPTION(err) v8::Isolate::GetCurrent()->ThrowException(err)
-#define SWIGV8_STRING_NEW(str) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), str)
-#define SWIGV8_SYMBOL_NEW(sym) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), sym)
+#define SWIGV8_STRING_NEW(str) (v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), str, v8::NewStringType::kNormal)).ToLocalChecked()
+#define SWIGV8_SYMBOL_NEW(sym) (v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), sym, v8::NewStringType::kNormal)).ToLocalChecked()
 #endif
 
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032318)
@@ -117,7 +125,7 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue()
 #define SWIGV8_WRITE_UTF8(handle, buffer, len) (handle)->WriteUtf8(buffer, len)
 #define SWIGV8_UTF8_LENGTH(handle) (handle)->Utf8Length()
-#else
+#elif (SWIG_V8_VERSION < 0x0706)
 #define SWIGV8_TO_OBJECT(handle) (handle)->ToObject(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
 #define SWIGV8_TO_STRING(handle) (handle)->ToString(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
 #define SWIGV8_NUMBER_VALUE(handle) (handle)->NumberValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
@@ -125,7 +133,16 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
 #define SWIGV8_WRITE_UTF8(handle, buffer, len) (handle)->WriteUtf8(v8::Isolate::GetCurrent(), buffer, len)
 #define SWIGV8_UTF8_LENGTH(handle) (handle)->Utf8Length(v8::Isolate::GetCurrent())
+#else
+#define SWIGV8_TO_OBJECT(handle) (handle)->ToObject(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
+#define SWIGV8_TO_STRING(handle) (handle)->ToString(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
+#define SWIGV8_NUMBER_VALUE(handle) (handle)->NumberValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
+#define SWIGV8_INTEGER_VALUE(handle) (handle)->IntegerValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
+#define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue(v8::Isolate::GetCurrent())
+#define SWIGV8_WRITE_UTF8(handle, buffer, len) (handle)->WriteUtf8(v8::Isolate::GetCurrent(), buffer, len)
+#define SWIGV8_UTF8_LENGTH(handle) (handle)->Utf8Length(v8::Isolate::GetCurrent())
 #endif
+
 
 /* ---------------------------------------------------------------------------
  * Error handling
@@ -371,10 +388,11 @@ SWIGRUNTIME void SWIGV8_SetPrivateData(SWIGV8_OBJECT obj, void *ptr, swig_type_i
   cdata->handle.MarkIndependent();
 #elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032100)
   cdata->handle.MarkIndependent(v8::Isolate::GetCurrent());
-#else
+#elif (SWIG_V8_VERSION < 0x0706)
   cdata->handle.MarkIndependent();
+// Looks like future versions do not require that anymore:
+// https://monorail-prod.appspot.com/p/chromium/issues/detail?id=923361#c11
 #endif
-
 }
 
 SWIGRUNTIME int SWIG_V8_ConvertPtr(SWIGV8_VALUE valRef, void **ptr, swig_type_info *info, int flags) {
@@ -422,8 +440,12 @@ SWIGRUNTIME SWIGV8_VALUE SWIG_V8_NewPointerObj(void *ptr, swig_type_info *info, 
   }
 #endif
 
-//  v8::Handle<v8::Object> result = class_templ->InstanceTemplate()->NewInstance();
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903) || (SWIG_V8_VERSION < 0x0705)
   v8::Local<v8::Object> result = class_templ->InstanceTemplate()->NewInstance();
+#else
+  v8::Local<v8::Object> result = class_templ->InstanceTemplate()->NewInstance(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked();
+#endif
+
   SWIGV8_SetPrivateData(result, ptr, info, flags);
 
   SWIGV8_ESCAPE(result);
@@ -646,8 +668,10 @@ SWIGV8_VALUE SWIGV8_NewPackedObj(void *data, size_t size, swig_type_info *type) 
   cdata->handle.MarkIndependent();
 #elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032100)
   cdata->handle.MarkIndependent(v8::Isolate::GetCurrent());
-#else
+#elif (SWIG_V8_VERSION < 0x0706)
   cdata->handle.MarkIndependent();
+// Looks like future versions do not require that anymore:
+// https://monorail-prod.appspot.com/p/chromium/issues/detail?id=923361#c11
 #endif
 
   SWIGV8_ESCAPE(obj);
@@ -674,6 +698,9 @@ SWIGV8_VALUE SWIGV8_AppendOutput(SWIGV8_VALUE result, SWIGV8_VALUE obj) {
 
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903) || (SWIG_V8_VERSION < 0x0706)
   arr->Set(arr->Length(), obj);
+#else
+  arr->Set(SWIGV8_CURRENT_CONTEXT(), arr->Length(), obj);
+#endif
 
   SWIGV8_ESCAPE(arr);
 }
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -49,17 +49,18 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_THROW_EXCEPTION(err) v8::ThrowException(err)
 #define SWIGV8_STRING_NEW(str) v8::String::New(str)
 #define SWIGV8_SYMBOL_NEW(sym) v8::String::NewSymbol(sym)
+#elif (SWIG_V8_VERSION < 0x0706)
+#define SWIGV8_ADJUST_MEMORY(size) v8::Isolate::GetCurrent()->AdjustAmountOfExternalAllocatedMemory(size)
+#define SWIGV8_CURRENT_CONTEXT() v8::Isolate::GetCurrent()->GetCurrentContext()
+#define SWIGV8_THROW_EXCEPTION(err) v8::Isolate::GetCurrent()->ThrowException(err)
+#define SWIGV8_STRING_NEW(str) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), str, v8::String::kNormalString)
+#define SWIGV8_SYMBOL_NEW(sym) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), sym, v8::String::kNormalString)
 #else
 #define SWIGV8_ADJUST_MEMORY(size) v8::Isolate::GetCurrent()->AdjustAmountOfExternalAllocatedMemory(size)
 #define SWIGV8_CURRENT_CONTEXT() v8::Isolate::GetCurrent()->GetCurrentContext()
 #define SWIGV8_THROW_EXCEPTION(err) v8::Isolate::GetCurrent()->ThrowException(err)
-#if (SWIG_V8_VERSION < 0x0706)
-#define SWIGV8_STRING_NEW(str) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), str, v8::String::kNormalString)
-#define SWIGV8_SYMBOL_NEW(sym) v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), sym, v8::String::kNormalString)
-#else
 #define SWIGV8_STRING_NEW(str) (v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), str, v8::NewStringType::kNormal)).ToLocalChecked()
 #define SWIGV8_SYMBOL_NEW(sym) (v8::String::NewFromUtf8(v8::Isolate::GetCurrent(), sym, v8::NewStringType::kNormal)).ToLocalChecked()
-#endif
 #endif
 
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032318)
@@ -124,18 +125,22 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue()
 #define SWIGV8_WRITE_UTF8(handle, buffer, len) (handle)->WriteUtf8(buffer, len)
 #define SWIGV8_UTF8_LENGTH(handle) (handle)->Utf8Length()
+#elif (SWIG_V8_VERSION < 0x0706)
+#define SWIGV8_TO_OBJECT(handle) (handle)->ToObject(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
+#define SWIGV8_TO_STRING(handle) (handle)->ToString(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
+#define SWIGV8_NUMBER_VALUE(handle) (handle)->NumberValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
+#define SWIGV8_INTEGER_VALUE(handle) (handle)->IntegerValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
+#define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
+#define SWIGV8_WRITE_UTF8(handle, buffer, len) (handle)->WriteUtf8(v8::Isolate::GetCurrent(), buffer, len)
+#define SWIGV8_UTF8_LENGTH(handle) (handle)->Utf8Length(v8::Isolate::GetCurrent())
 #else
 #define SWIGV8_TO_OBJECT(handle) (handle)->ToObject(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
 #define SWIGV8_TO_STRING(handle) (handle)->ToString(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()
 #define SWIGV8_NUMBER_VALUE(handle) (handle)->NumberValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
 #define SWIGV8_INTEGER_VALUE(handle) (handle)->IntegerValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
+#define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue(v8::Isolate::GetCurrent())
 #define SWIGV8_WRITE_UTF8(handle, buffer, len) (handle)->WriteUtf8(v8::Isolate::GetCurrent(), buffer, len)
 #define SWIGV8_UTF8_LENGTH(handle) (handle)->Utf8Length(v8::Isolate::GetCurrent())
-#if (SWIG_V8_VERSION < 0x0706)
-#define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue(SWIGV8_CURRENT_CONTEXT()).ToChecked()
-#else
-#define SWIGV8_BOOLEAN_VALUE(handle) (handle)->BooleanValue(v8::Isolate::GetCurrent())
-#endif
 #endif
 
 
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -173,7 +173,7 @@ public:
         SWIGV8_THROW_EXCEPTION(err);
     }
   }
-  v8::Handle<v8::Value> err;
+  SWIGV8_VALUE err;
 };
 
 /* ---------------------------------------------------------------------------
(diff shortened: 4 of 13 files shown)


yegorich:
@lissyx the tests for older Node.js versions are OK but I have overlooked a `v8::Handle` somewhere. Unfortunately I cannot find this place. Can you help me to find it?

```
../example_wrap.cxx:1736:5: error: ‘Handle’ in namespace ‘v8’ does not name a template type
 v8::Handle<v8::Value> SWIG_From_int  (int value)
     ^~~~~~
../example_wrap.cxx: In function ‘SwigV8ReturnValue _wrap_placeholder(const SwigV8Arguments&)’:
../example_wrap.cxx:1763:14: error: ‘SWIG_From_int’ was not declared in this scope
   jsresult = SWIG_From_int(static_cast< int >(result));
              ^~~~~~~~~~~~~
../example_wrap.cxx: In function ‘SwigV8ReturnValue JavaScript_do_work(const SwigV8Arguments&)’:
../example_wrap.cxx:1783:13: error: ‘SWIG_From_int’ was not declared in this scope
             SWIG_From_int(static_cast< int >(MY_MAGIC_NUMBER));
             ^~~~~~~~~~~~~
```

Where does `SWIG_From_int()` come from?

lissyx:
> @lissyx the tests for older Node.js versions are OK but I have overlooked a `v8::Handle` somewhere. Unfortunately I cannot find this place. Can you help me to find it?
> 
> ```
> ../example_wrap.cxx:1736:5: error: ‘Handle’ in namespace ‘v8’ does not name a template type
>  v8::Handle<v8::Value> SWIG_From_int  (int value)
>      ^~~~~~
> ../example_wrap.cxx: In function ‘SwigV8ReturnValue _wrap_placeholder(const SwigV8Arguments&)’:
> ../example_wrap.cxx:1763:14: error: ‘SWIG_From_int’ was not declared in this scope
>    jsresult = SWIG_From_int(static_cast< int >(result));
>               ^~~~~~~~~~~~~
> ../example_wrap.cxx: In function ‘SwigV8ReturnValue JavaScript_do_work(const SwigV8Arguments&)’:
> ../example_wrap.cxx:1783:13: error: ‘SWIG_From_int’ was not declared in this scope
>              SWIG_From_int(static_cast< int >(MY_MAGIC_NUMBER));
>              ^~~~~~~~~~~~~
> ```
> 
> Where does `SWIG_From_int()` come from?

I'm sorry, but I have no idea.

yegorich:
In theory, I have found the place. It comes from `Lib/javascript/v8/javascriptprimtypes.swg`.

```
/* int */

%fragment(SWIG_From_frag(int),"header") {
SWIGINTERNINLINE
#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
v8::Handle<v8::Value> SWIG_From_dec(int)(int value)
#else
v8::Local<v8::Value> SWIG_From_dec(int)(int value)
#endif
{
  return SWIGV8_INT32_NEW(value);
}
}
```
 But it looks like this statement doesn't work:

`#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)`

It doesn't matter, what I put as an expression, it is always true.  @wsfulton , @vadz what am I missing?

vadz:
Sorry, I didn't have time to actually test this, but as a wild guess, could this be due to using `{...}` delimiters? Inside them, SWIG handles preprocessing directives itself, see the "Preprocessor" section of the manual. And I think the intention here is rather to put these `#if`s into the generated code, in which case you would use `%{....%}` around the entire fragment or use `%#` inside it if you do need SWIG to preprocess some part of it.

yegorich:
@vadz thanks for the hint. Wrappers for int, long etc. are now working. But now I have a problem with `SWIG_From_char`. I cannot understand where it comes from. `char` is not covered in `Lib/javascript/v8/javascriptprimtypes.swg`. It is not handled in `Lib/javascript/v8/javascriptstrings.swg` either.

```
SWIGINTERNINLINE v8::Handle<v8::Value>
  SWIG_From_char  (char c)
  {
    return SWIG_FromCharPtrAndSize(&c,1);
  }


  SWIGINTERNINLINE v8::Handle<v8::Value>
  SWIG_FromCharPtr(const char *cptr)
  {
    return SWIG_FromCharPtrAndSize(cptr, (cptr ? strlen(cptr) : 0));
  }

```

```
../example_wrap.cxx:1791:22: error: ‘Handle’ in namespace ‘v8’ does not name a template type
 SWIGINTERNINLINE v8::Handle<v8::Value>
                      ^~~~~~
../example_wrap.cxx:1798:22: error: ‘Handle’ in namespace ‘v8’ does not name a template type
 SWIGINTERNINLINE v8::Handle<v8::Value>
                      ^~~~~~
../example_wrap.cxx: In function ‘SwigV8ReturnValue _wrap_CCONST(v8::Local<v8::Name>, const SwigV8PropertyCallbackInfo&)’:
../example_wrap.cxx:1869:16: error: ‘SWIG_From_char’ was not declared in this scope
     jsresult = SWIG_From_char(static_cast< char >('x'));
                ^~~~~~~~~~~~~~

```

yegorich:
Looks like it affects all types not covered in `Lib/javascript/v8/javascriptprimtypes.swg` i.e. some default handling that I cannot figure out:

* `SWIG_AsVal_unsigned_SS_int`
* `SWIG_AsPtr_std_string`
* `SWIG_AsVal_signed_SS_char`
* etc.

yegorich:
Adding @p2k and @mochizk to the discussion.

yegorich:
I have found the problem. It is in `Lib/javascript/v8/javascripttypemaps.swg`:

```
#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
#define SWIG_Object                     v8::Handle<v8::Value>
#else
#define SWIG_Object                     v8::Local<v8::Value>
#endif
```

But here once again, I don't understand what syntax to use. The above block is not working.

vadz:
Here you just can't define `SWIG_Object` differently at SWIG time, SWIG output should be the same for whichever version of Node the code ends up being compiled with. But you can, of course, define `SWIG_Object` in the generated C++ code differently depending on this condition.

There seems to be some confusion between SWIG-time and compile-time phases in this PR...

yegorich:
@vadz I am very confused with different phases indeed :-)

So where would you place this conditional definition? `javascripttypemaps.swg` seems to be the first include in the `javascript.swg`.

vadz:
It's not so much where they're placed which counts, but how do they translate to the output. The `V8_MAJOR_VERSION` checks only make sense in the generated C/C++ code, not for SWIG.

wsfulton:
@yegorich, the following might help: view the output of the SWIG preprocessor using SWIG -E (for SWIG time). View the output of the c++ compiler using gcc -E. Both gcc and swig run the code through their associated preprocessors before the actual swig compiler or C++ compiler processes the code.

wsfulton:
SWIG's preprocessor docs might help too: http://swig.org/Doc4.0/Preprocessor.html.

yegorich:
@wsfulton I have added `-E` to the `Examples/javascript/example.mk`:

`SWIGOPT        = -$(JSENGINE) -DV8_VERSION=$(JSV8_VERSION) -E`

But I cannot find the occurrences of  `SWIG_Object` in the output aside from the one in comments. So it was already processed.

The problem is, `SWIG_Object` is the core element that will be used in the common "recipes" like `Lib/typemaps/primtypes.swg` and `Lib/typemaps/ptrtypes.swg`. So if we want to have it customized, we have to make it version-dependent in the  `javascripttypemaps.swg`.

We already have some mappings for int, bool, etc. They are defined via %fragment where we can insert compile time `#if/#else` expressions. But reading  [http://swig.org/Doc4.0/Preprocessor.html](http://swig.org/Doc4.0/Preprocessor.html), I cannot find any construct for the simple define and from what I see in `Lib/typemaps/ptrtypes.swg`, it won't work this way.

It is very unfortunate that Node.js has deprecated such a fundamental datatype as `v8::Handle`. There is no such situation with the other languages supported by SWIG. Hence, the only way I see for now, is to drop support for Node.js older than v4 and stand-alone v8.

Let me know if I see it wrong.

So far the issue with the `SWIG_Object` is the last obstacle for adding support for v12/13 aside from ignoring some warnings.

tomleavy:
@yegorich I put in a PR on your branch that I think resolves the issue you were having with `SWIG_Object`. I ran the CI from my branch and it seemed to pass on all node versions

yegorich:
@tomleavy it did the trick for me too. Thank you very much.

yegorich:
@wsfulton @vadz the JavaScript tests are all green. Could you review this PR again? Thanks.

tomleavy:
@wsfulton @vadz bump

Lib/javascript/v8/javascriptcode.swg
```diff
@@ -435,7 +478,12 @@ fail:
  * ----------------------------------------------------------------------------- */
 %fragment("jsv8_register_class", "templates")
 %{
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903) || (SWIG_V8_VERSION < 0x0706)
```
vadz, line 481: This is very confusing because 2 different conventions for `SWIG_V8_VERSION` are used in the same expression. I'd use `V8_MAJOR_VERSION` and `V8_MINOR_VERSION` in the RHS, as the intention is to deprecate `SWIG_V8_VERSION` and use native v8 constants for node `>= 4.3.0`.

Examples/javascript/native/example.i
```diff
@@ -15,7 +15,11 @@ int placeholder() { return 0; }
     static SwigV8ReturnValue JavaScript_do_work(const SwigV8Arguments &args) {
         SWIGV8_HANDLESCOPE();
         const int MY_MAGIC_NUMBER = 5;
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
         v8::Handle<v8::Value> jsresult =
+#else
+        v8::Local<v8::Value> jsresult =
+#endif
```
vadz, line 22: I'd strongly consider defining `SwigV8Object` typedef or whatever allowing to write this just as
```suggestion
        SwigV8Object<v8::Value>::type jsresult =
```
to avoid repeating this preprocessor condition over and over and over again.

(ideal would be to avoid the `::type` part, but I don't see how to do it without C++11 `using`).

Lib/javascript/v8/javascriptinit.swg
```diff
@@ -70,14 +70,18 @@ extern "C"
 #if (NODE_MODULE_VERSION < 0x000C)
 void SWIGV8_INIT (v8::Handle<v8::Object> exports)
 #else
-void SWIGV8_INIT (v8::Handle<v8::Object> exports, v8::Handle<v8::Object> /*module*/)
+void SWIGV8_INIT (v8::Local<v8::Object> exports, v8::Local<v8::Object> /*module*/)
```
vadz, line 73: This looks like an incompatible change to me, i.e. whichever versions of v8 that uses init function with 2 parameters of `v8::Handle` type would be broken by this change. I am not sure which versions these are and whether we care about them or not though.

Lib/javascript/v8/javascriptrun.swg
```diff
@@ -67,6 +75,7 @@ typedef v8::PropertyCallbackInfo<v8::Value> SwigV8PropertyCallbackInfo;
 #define SWIGV8_NUMBER_NEW(num) v8::Number::New(num)
 #define SWIGV8_OBJECT_NEW() v8::Object::New()
 #define SWIGV8_UNDEFINED() v8::Undefined()
+#define SWIGV8_OBJECT v8::Handle<v8::Value>
```
vadz, line 78: This seems like a very confusing name, knowing that we also use `v8::Handle<v8::Object>` in a couple of places. Why not call it `SWIGV8_VALUE` instead?

yegorich: I see your point and will rename it.

Examples/javascript/native/example.i
yegorich: What about using `SWIGV8_VALUE` instead of introducing a new typedef?

vadz: If we have `SWIGV8_VALUE`, we'd also need `SWIGV8_OBJECT` and `SWIGV8_FUNCTIONHANDLER` and maybe more, as we use `v8::Handle<>` or `v8::Local<>` with a few types, not just `v8::Value`. So it seems like a typedef would be more flexible.

But having a few macros would still be better than writing these `#if`s everywhere, so if the typedef solution doesn't work for some reason, doing what you propose would still be an improvement IMO.

yegorich: Somehow I failed to use typedef hence this my current version with macros.

Lib/javascript/v8/javascriptinit.swg
yegorich: This code is from 2013. I believe 0x000C means Node.js v0.12. I have replaced everything with `SWIGV8_OBJECT` macro.

Lib/javascript/v8/javascriptcode.swg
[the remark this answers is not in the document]
yegorich: You're right. I have removed it the tests are still OK.

vadz:
Thanks for the changes! The macros are a bit painful to look at, but still much better than the `#if` checks, so IMO this is good enough to be merged, especially as lack of support for newer Node versions is a real problem in practice.

Let's hope that @wsfulton can look at and merge this soon.

Lib/javascript/v8/javascriptcode.swg
yegorich: For the future, I suggest to use `NODE_MODULE_VERSION`. See its [description](https://nodejs.org/en/download/releases/#ref-1)

ojwb:
> I hope ojwb can correct me if I'm wrong, but I think we really just need to define `V8_XXX_VERSION` for the ancient Node versions that didn't have them and use them, and only them, exclusively.

Defining `V8_XXX_VERSION` for older V8 which doesn't define it for itself seems a bit unclean to me - we'd be messing with macros in someone else's name space.

Practically, older V8 versions presumably won't check for these macros so I wouldn't expect problems there, but it could be problematic if user interface files (or other included headers) check for them in this sort of fashion:

```
#ifdef V8_MAJOR_VERSION
// Do stuff for V8 >= 4.3.0
#else
// Do stuff for older V8
#endif
```

(Are V8 versions < 4.3.0 still relevant now?)

yegorich:
I don't think that V8 versions are still relevant. At least there are no submitted issues.

As discussed in https://github.com/swig/swig/issues/1735 and https://github.com/swig/swig/pull/1736, support for these versions will be removed from the next major SWIG version. This change would reduce the usage of macros hugely.

vadz:
> > I hope ojwb can correct me if I'm wrong, but I think we really just need to define `V8_XXX_VERSION` for the ancient Node versions that didn't have them and use them, and only them, exclusively.
> 
> Defining `V8_XXX_VERSION` for older V8 which doesn't define it for itself seems a bit unclean to me - we'd be messing with macros in someone else's name space.

Yes, but I think that it's not a problem in practice, _especially_ considering that we're going to drop support for these versions soon anyhow.
 
> Practically, older V8 versions presumably won't check for these macros so I wouldn't expect problems there, but it could be problematic if user interface files (or other included headers) check for them in this sort of fashion:
> 
> ```
> #ifdef V8_MAJOR_VERSION
> // Do stuff for V8 >= 4.3.0
> #else
> // Do stuff for older V8
> #endif
> ```

If we care about this, we need to define `SWIG_V8_MAJOR_VERSION` that would belong to us and which we'd use instead of `V8_MAJOR_VERSION`. I'd just really like to have one single symbol, no matter what it's called. In fact, as already mentioned, I'd like to have one `SWIG_V8_CHECK_VERSION(x,y,z)` macro that would evaluate to 1 if V8 version is `>= x.y.z` and 0 otherwise. This should be already possible to do right now.

> (Are V8 versions < 4.3.0 still relevant now?)

I don't think so. Ubuntu 16.04 LTS (i.e. not even the last one) comes with [Node 4.2.6](https://packages.ubuntu.com/xenial/nodejs) which, according to this [useful table](https://nodejs.org/en/download/releases/) uses V8 4.5.

chenhengqi:
@yegorich  After replace all `v8::Handle` with `v8::Local` manually,  another error occurs
```
error: 'LocalScope' is not a member of 'v8'
```

Any ideas ?

brgirgis:
This branch could handle successfully one of my non-trivial projects.

chenhengqi:
Nice job. I apply these changes manually to my project and it works.

Hope this can be merge soon.

brgirgis:
Any specific reason why this branch is not in yet?

vadz:
@wsfulton Any chance of merging this for 4.0.2? This is really required to make SWIG JS backend useful in 2020.

nullr0ute:
I'm still getting errors when trying to build mraa (https://github.com/eclipse/mraa) and upm (https://github.com/eclipse/upm) with this patch against nodejs 12.16.3:

<pre>
[ 39%] Building CXX object src/javascript/CMakeFiles/mraajs.dir/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx.o
cd /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript && /usr/lib64/ccache/c++  -Dmraajs_EXPORTS -I/home/perobins/fedora/packages/mraa/mraa-2.1.0/api -I/home/perobins/fedora/packages/mraa/mraa-2.1.0/api/mraa -I/home/perobins/fedora/packages/mraa/mraa-2.1.0/include -I/usr/include/json-c -I/usr/include/node -I/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/..  -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -Wall -Werror=main -Wformat -Wmain -Wuninitialized -Winit-self -Wnon-virtual-dtor -Woverloaded-virtual -Wreorder -fPIC    -DBUILDING_NODE_EXTENSION -DSWIGJAVASCRIPT=TRUE -std=gnu++11 -o CMakeFiles/mraajs.dir/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx.o -c /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'v8::Local<v8::Value> SWIGV8_AppendOutput(v8::Local<v8::Value>, v8::Local<v8::Value>)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1479:56: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, uint32_t, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
 1479 |   arr->Set(SWIGV8_CURRENT_CONTEXT(), arr->Length(), obj);
      |                                                        ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3509:37: note: declared here
 3509 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context, uint32_t index,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'void SWIGV8_AddStaticFunction(v8::Local<v8::Object>, const char*, void (* const&)(const v8::FunctionCallbackInfo<v8::Value>&))':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1554:151: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
 1554 |   obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW(symbol), SWIGV8_FUNCTEMPLATE_NEW(_func)->GetFunction(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked());
      |                                                                                                                                                       ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'void SWIGV8_AddStaticVariable(v8::Local<v8::Object>, const char*, SwigV8AccessorGetterCallback, SwigV8AccessorSetterCallback)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1566:87: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::SetAccessor(v8::Local<v8::Context>, v8::Local<v8::Name>, v8::AccessorNameGetterCallback, v8::AccessorNameSetterCallback, v8::MaybeLocal<v8::Value>, v8::AccessControl, v8::PropertyAttribute, v8::SideEffectType, v8::SideEffectType)', declared with attribute 'warn_unused_result' [-Wunused-result]
 1566 |   obj->SetAccessor(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW(symbol), getter, setter);
      |                                                                                       ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3604:37: note: declared here
 3604 |   V8_WARN_UNUSED_RESULT Maybe<bool> SetAccessor(
      |                                     ^~~~~~~~~~~
In file included from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1818:
/home/perobins/fedora/packages/mraa/mraa-2.1.0/api/mraa/gpio.hpp: In static member function 'static void mraa::Gpio::v8isr(uv_work_t*, int)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/api/mraa/gpio.hpp:178:63: error: no matching function for call to 'v8::Function::Call(v8::Local<v8::Object>, int&, v8::Local<v8::Value> [1])'
  178 |         f->Call(SWIGV8_CURRENT_CONTEXT()->Global(), argc, argv);
      |                                                               ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:4298:43: note: candidate: 'v8::MaybeLocal<v8::Value> v8::Function::Call(v8::Local<v8::Context>, v8::Local<v8::Value>, int, v8::Local<v8::Value>*)'
 4298 |   V8_WARN_UNUSED_RESULT MaybeLocal<Value> Call(Local<Context> context,
      |                                           ^~~~
/usr/include/node/v8.h:4298:43: note:   candidate expects 4 arguments, 3 provided
In file included from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1818:
/home/perobins/fedora/packages/mraa/mraa-2.1.0/api/mraa/gpio.hpp: In member function 'mraa::Result mraa::Gpio::useMmap(bool)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/api/mraa/gpio.hpp:319:77: warning: 'mraa_result_t mraa_gpio_use_mmaped(mraa_gpio_context, mraa_boolean_t)' is deprecated [-Wdeprecated-declarations]
  319 |         return (Result) mraa_gpio_use_mmaped(m_gpio, (mraa_boolean_t) enable);
      |                                                                             ^
In file included from /home/perobins/fedora/packages/mraa/mraa-2.1.0/api/mraa/gpio.hpp:29,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1818:
/home/perobins/fedora/packages/mraa/mraa-2.1.0/api/mraa/gpio.h:289:26: note: declared here
  289 | DEPRECATED mraa_result_t mraa_gpio_use_mmaped(mraa_gpio_context dev, mraa_boolean_t mmap);
      |                          ^~~~~~~~~~~~~~~~~~~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'SwigV8ReturnValue _wrap_I2c_read(const SwigV8Arguments&)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:8541:54: error: no matching function for call to 'New(char*, int&)'
 8541 |     jsresult = node::Buffer::New((char*) arg2, result)->handle_;
      |                                                      ^
In file included from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1807:
/usr/include/node/node_buffer.h:49:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, size_t)'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                        ^~~
/usr/include/node/node_buffer.h:49:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:52:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, v8::Local<v8::String>, node::encoding)'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:52:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:57:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t, node::Buffer::FreeCallback, void*)'
   57 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:57:40: note:   candidate expects 5 arguments, 2 provided
/usr/include/node/node_buffer.h:64:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t)'
   64 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:64:40: note:   candidate expects 3 arguments, 2 provided
/usr/include/node/node_buffer.h:69:44: note: candidate: 'v8::MaybeLocal<v8::Uint8Array> node::Buffer::New(v8::Isolate*, v8::Local<v8::ArrayBuffer>, size_t, size_t)'
   69 | NODE_EXTERN v8::MaybeLocal<v8::Uint8Array> New(v8::Isolate* isolate,
      |                                            ^~~
/usr/include/node/node_buffer.h:69:44: note:   candidate expects 4 arguments, 2 provided
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'SwigV8ReturnValue _wrap_I2c_readBytesReg(const SwigV8Arguments&)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:8709:54: error: no matching function for call to 'New(char*, int&)'
 8709 |     jsresult = node::Buffer::New((char*) arg3, result)->handle_;
      |                                                      ^
In file included from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1807:
/usr/include/node/node_buffer.h:49:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, size_t)'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                        ^~~
/usr/include/node/node_buffer.h:49:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:52:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, v8::Local<v8::String>, node::encoding)'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:52:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:57:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t, node::Buffer::FreeCallback, void*)'
   57 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:57:40: note:   candidate expects 5 arguments, 2 provided
/usr/include/node/node_buffer.h:64:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t)'
   64 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:64:40: note:   candidate expects 3 arguments, 2 provided
/usr/include/node/node_buffer.h:69:44: note: candidate: 'v8::MaybeLocal<v8::Uint8Array> node::Buffer::New(v8::Isolate*, v8::Local<v8::ArrayBuffer>, size_t, size_t)'
   69 | NODE_EXTERN v8::MaybeLocal<v8::Uint8Array> New(v8::Isolate* isolate,
      |                                            ^~~
/usr/include/node/node_buffer.h:69:44: note:   candidate expects 4 arguments, 2 provided
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'SwigV8ReturnValue _wrap_Spi_write(const SwigV8Arguments&)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:10090:54: error: no matching function for call to 'New(char*, int&)'
10090 |     jsresult = node::Buffer::New((char*) result, arg3)->handle_;
      |                                                      ^
In file included from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1807:
/usr/include/node/node_buffer.h:49:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, size_t)'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                        ^~~
/usr/include/node/node_buffer.h:49:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:52:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, v8::Local<v8::String>, node::encoding)'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:52:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:57:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t, node::Buffer::FreeCallback, void*)'
   57 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:57:40: note:   candidate expects 5 arguments, 2 provided
/usr/include/node/node_buffer.h:64:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t)'
   64 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:64:40: note:   candidate expects 3 arguments, 2 provided
/usr/include/node/node_buffer.h:69:44: note: candidate: 'v8::MaybeLocal<v8::Uint8Array> node::Buffer::New(v8::Isolate*, v8::Local<v8::ArrayBuffer>, size_t, size_t)'
   69 | NODE_EXTERN v8::MaybeLocal<v8::Uint8Array> New(v8::Isolate* isolate,
      |                                            ^~~
/usr/include/node/node_buffer.h:69:44: note:   candidate expects 4 arguments, 2 provided
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'SwigV8ReturnValue _wrap_Uart_read(const SwigV8Arguments&)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:10670:54: error: no matching function for call to 'New(char*, int&)'
10670 |     jsresult = node::Buffer::New((char*) arg2, result)->handle_;
      |                                                      ^
In file included from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:1807:
/usr/include/node/node_buffer.h:49:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, size_t)'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                        ^~~
/usr/include/node/node_buffer.h:49:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   49 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate, size_t length);
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:52:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, v8::Local<v8::String>, node::encoding)'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:52:57: note:   no known conversion for argument 1 from 'char*' to 'v8::Isolate*'
   52 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                            ~~~~~~~~~~~~~^~~~~~~
/usr/include/node/node_buffer.h:57:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t, node::Buffer::FreeCallback, void*)'
   57 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:57:40: note:   candidate expects 5 arguments, 2 provided
/usr/include/node/node_buffer.h:64:40: note: candidate: 'v8::MaybeLocal<v8::Object> node::Buffer::New(v8::Isolate*, char*, size_t)'
   64 | NODE_EXTERN v8::MaybeLocal<v8::Object> New(v8::Isolate* isolate,
      |                                        ^~~
/usr/include/node/node_buffer.h:64:40: note:   candidate expects 3 arguments, 2 provided
/usr/include/node/node_buffer.h:69:44: note: candidate: 'v8::MaybeLocal<v8::Uint8Array> node::Buffer::New(v8::Isolate*, v8::Local<v8::ArrayBuffer>, size_t, size_t)'
   69 | NODE_EXTERN v8::MaybeLocal<v8::Uint8Array> New(v8::Isolate* isolate,
      |                                            ^~~
/usr/include/node/node_buffer.h:69:44: note:   candidate expects 4 arguments, 2 provided
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx: In function 'void mraa_initialize(v8::Local<v8::Object>, v8::Local<v8::Object>)':
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12594:100: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12594 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("uint8Array"), _exports_uint8Array_obj);
      |                                                                                                    ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12599:88: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12599 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("Gpio"), _exports_Gpio_obj);
      |                                                                                        ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12604:86: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12604 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("I2c"), _exports_I2c_obj);
      |                                                                                      ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12609:86: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12609 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("Pwm"), _exports_Pwm_obj);
      |                                                                                      ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12614:86: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12614 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("Spi"), _exports_Spi_obj);
      |                                                                                      ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12619:86: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12619 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("Aio"), _exports_Aio_obj);
      |                                                                                      ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12624:88: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12624 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("Uart"), _exports_Uart_obj);
      |                                                                                        ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
/home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:12629:86: warning: ignoring return value of 'v8::Maybe<bool> v8::Object::Set(v8::Local<v8::Context>, v8::Local<v8::Value>, v8::Local<v8::Value>)', declared with attribute 'warn_unused_result' [-Wunused-result]
12629 | exports_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("Led"), _exports_Led_obj);
      |                                                                                      ^
In file included from /usr/include/node/node.h:67,
                 from /home/perobins/fedora/packages/mraa/mraa-2.1.0/src/javascript/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx:171:
/usr/include/node/v8.h:3504:37: note: declared here
 3504 |   V8_WARN_UNUSED_RESULT Maybe<bool> Set(Local<Context> context,
      |                                     ^~~
make[2]: *** [src/javascript/CMakeFiles/mraajs.dir/build.make:86: src/javascript/CMakeFiles/mraajs.dir/CMakeFiles/mraajs.dir/mraajsJAVASCRIPT_wrap.cxx.o] Error 1
make[2]: Leaving directory '/home/perobins/fedora/packages/mraa/mraa-2.1.0'
make[1]: *** [CMakeFiles/Makefile2:669: src/javascript/CMakeFiles/mraajs.dir/all] Error 2
make[1]: Leaving directory '/home/perobins/fedora/packages/mraa/mraa-2.1.0'
make: *** [Makefile:164: all] Error 2
</pre>

vadz:
Are you sure those `node::Buffer::New()` calls don't come from your own typemaps? I don't see anything like this in SWIG sources/typemaps.

nullr0ute:
> Are you sure those `node::Buffer::New()` calls don't come from your own typemaps? I don't see anything like this in SWIG sources/typemaps.

TBH I'm not sure, it's not my project, I'm just trying to build it. It builds fine on the old NodeJS builds and it was identified that the swig issue was the issue for the newer nodejs

vadz:
@nullr0ute You need to ask the original project developers to look at this then, or at the very least try to reproduce the problem in the SWIG test suite.

yegorich:
@wsfulton I have seen that you already made 4.0.2 release notes but can we still get this PR into 4.0.2?

yegorich:
@wsfulton how should we proceed? Merge this PR or agree on what Node.js versions to support and work on https://github.com/swig/swig/pull/1736?

wsfulton:
Sorry it didn't make it into 4.0.2, I really needed to get it out the door with minimal risk and change in versions of target languages supported as it was a patch release. I'm going to try work out what is going on here next.

What is the high level summary for Node versions supported? There was talk above of not supporting the current set of supported Node versions and it would be a lot easier to add support for v12 if 

> ... the only way I see for now, is to drop support for Node.js older than v4 and stand-alone v8

The SWIG test-suite for these older version seems to pass on Travis though with this pull request.

The next SWIG version will be 4.1.0 and dropping support for old versions of Node is quite okay and probably a good thing for Node. Can the patch or the code base be simplified by dropping support for versions < 4 as hinted at here and in #1520.

yegorich:
This PR enables support for the new Node.js versions without loosing the support for the old ones. It means SWIG supports versions from 0.10.x to v13. 

Almost every Node.js version introduces some non-compatible structures, calls, etc. This results in an ever-growing number of `#if's`:

```C
#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031710)
  object.Dispose();
#elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031900)
  object.Dispose(isolate);
#elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x032100)
  object->Dispose(isolate);
#elif (V8_MAJOR_VERSION-0) < 5
  object->Dispose();
#endif
``` 

Hence, removing support for the obsolete Node.js versions would significantly simplify the code. It is also important to drop the versions, where `v8::Handle` is not an alias for `v8::Local`. It seems like Node.js 6.5.x was the first version to provide such an alias. 

The 6.x.x was  EOL on 2019-04-03 i.e. more than a year ago. 8.x.x was also stopped in 2019. Meanwhile we have the following versions that receive updates: 10, 12 and 14.

Another goody in dropping the old versions is that we can use `NODE_MODULE_VERSION` to distinguish between incompatible API versions. As for now, we use a mixed version scheme. 

If we decide to drop the old versions, it makes more sense to make it https://github.com/swig/swig/pull/1736 as this PR makes changes to support all versions.

shrinktofit:
Vote for this.

asherkin:
Is there anything that can be done to help move this forward? It seems to have stalled for another 2 months after the last batch of questions were addressed.

arsnyder16:
@wsfulton Where does this stand?  

We are within 6 months of node 10 being end of life, and node consumers will need to move to a later versions soon.

jeanlescure:
I have a blocker due to the fact that a paid geolocation library my company needs is using SWIG to generate their Node version of the package, and even though they seem to be using this PR to generate the bindings for Node v12, we are **required** to update to Node v14 which is the current LTS.

lissyx:
> I have a blocker due to the fact that a paid geolocation library my company needs is using SWIG to generate their Node version of the package, and even though they seem to be using this PR to generate the bindings for Node v12, we are **required** to update to Node v14 which is the current LTS.

FTR we are still using something close from that PR and successfully generate up to NodeJS v15 bindings.

jeanlescure:
> > I have a blocker due to the fact that a paid geolocation library my company needs is using SWIG to generate their Node version of the package, and even though they seem to be using this PR to generate the bindings for Node v12, we are **required** to update to Node v14 which is the current LTS.
> 
> FTR we are still using something close from that PR and successfully generate up to NodeJS v15 bindings.

I opened #1898 since the build of the aforementioned library fails on NodeJS v14 using the bindings produced by this PR of SWIG :shrug:

tungntpham:
@yegorich This is great stuff. I have tested the fixes on Node 12 and Node 14 and it works nicely for us. While reviewing the code, we did some refactoring of the changes to reduce the code size and potentially improve the readability. I have created a pull request [here](https://github.com/yegorich/swig/pull/3) if you could take a look. Thanks.

yegorich:
@tungntpham  it is now almost a year since I have started working on this PR. Meanwhile we have switched to [ref-napi](https://github.com/node-ffi-napi/node-ffi-napi). So feel free to take the whole PR over.

tungntpham:
@Yegorich Thanks. I will start taking this over. May message you for further details.

arsnyder16:
@wsfulton  any chance we can get this into 4.1? Node 10 end of life is 2021-04-30

https://nodejs.org/en/about/releases/

tomleavy:
I second that. I've been using this branch to codegen for production servers with heavy volume and it's been great. Making it official would make my life a lot easier

rubyjedi:
This patch works well, and I can confirm it is functional with Node 15 on Docker(Alpine).

I also added changes to call IsNothing() on any Set() methods that specify a Local Context, which helps suppress GCC warning about unused return values.

rubyjedi/swig@f296c81f

dot-asm:
> I also added changes to call IsNothing() on any Set() methods that specify a Local Context, which helps suppress GCC warning about unused return values.

Formally speaking, the right thing to do is to actually use the value. Because there ought to be a rationale behind tagging methods in question with `V8_WARN_UNUSED_RESULT`. One can even argue that playing tricks like suggested one is actually inappropriate, because by swiping the problem under the rug we effectively prevent people from **ever** having a proper look at the problem.

Well, there is a chance that ignoring the return value is actually an appropriate thing to do under the circumstances, but then the commit message ought to explain **why** is it the case. As opposed to sounding more like "it's annoying [so let's just search-n-paste]." And in case it turns out to be a correct solution, I'd say that `ToChecked()` might be a slightly better choice. In essence, `IsNothing()` is a question, not an assertion. I mean the normal usage pattern would be like `if (x.IsNothing()) { oops() }`. BTW, if one was to actually act upon the return value after all, one would have to use `ToChecked()`. And note that `ToChecked()` does the "oops" part...

wsfulton:
Hi, I'll give this a good review in the coming days. Right now I can't consider merging this in as the Javascript node 12 tests fail on Travis, so please update the pull request with whatever is needed to fix it.

vadz:
FWIW this seems trivial to fix, we could either conditionally use the new function or just disable `-Wdeprecated-declarations` locally or even globally.

BTW, we should also consider migrating to GitHub Actions from Travis CI, the latter is practically unusable for open source projects now and I think it's only going to get worse, not better. Switching should be simple enough, I've already done it for a few projects and could do it for SWIG too if you agree.

wsfulton:
@vadz, your experiences on updating Travis CI sounds very useful. Let's discuss separately, how about as a Github discussion or old school swig-devel mailing list?

vadz:
Whatever you prefer, @wsfulton (but please ping me if you start a discussion as I'm not sure I'd be notified about it otherwise).

dot-asm:
Great news! As kind of a prior notice I have a request queued up for making initialization context-aware. This allows to instantiate the extension multiple times, most notably in multiple thread contexts. It's not fully-fledged in sense that it doesn't facilitate all the necessary hooks, but it's a step in desired direction... Well, it actually works for our purposes, for purely computational subroutines that is.

wsfulton:
@yegorich @tungntpham @dot-asm @lissyx who has taken over responsibility for this patch?

I can't work out if this is the patch to use or #1736 or #1746. If there is any doubt, swig 4.1 can drop versions older than NodeJS v4.0.0, as discussed in #1520. I'll even consider dropping support for versions older than v6.0.0. I would like to see the code cleaned up to drop the support of these older versions, but that can come later if need be. 

Anyone going to update the generation of this code:
```
error: ‘v8::Local<v8::Value> v8::Object::Get(uint32_t)’ is deprecated
```
and similar seen in Travis testing? It probably breaks even newer versions given the generated code is using a deprecated API. @lissyx you mention v15. Can we add Travis tests for v14 (can be after this patch is merged in if need be).

lissyx:
> @yegorich @tungntpham @dot-asm @lissyx who has taken over responsibility for this patch?
> 
> I can't work out if this is the patch to use or #1736 or #1746. If there is any doubt, swig 4.1 can drop versions older than NodeJS v4.0.0, as discussed in #1520. I'll even consider dropping support for versions older than v6.0.0. I would like to see the code cleaned up to drop the support of these older versions, but that can come later if need be.
> 
> Anyone going to update the generation of this code:
> 
> ```
> error: ‘v8::Local<v8::Value> v8::Object::Get(uint32_t)’ is deprecated
> ```
> 
> and similar seen in Travis testing? It probably breaks even newer versions given the generated code is using a deprecated API. @lissyx you mention v15. Can we add Travis tests for v14 (can be after this patch is merged in if need be).

I'm sorry, but I'm really unable to commit any time on looking into that now. All I can share is that we use https://github.com/lissyx/swig/tree/taskcluster for NodeJS v15 successfully, but it's only tested on our usage.

tungntpham:
> > @yegorich @tungntpham @dot-asm @lissyx who has taken over responsibility for this patch?
> > I can't work out if this is the patch to use or #1736 or #1746. If there is any doubt, swig 4.1 can drop versions older than NodeJS v4.0.0, as discussed in #1520. I'll even consider dropping support for versions older than v6.0.0. I would like to see the code cleaned up to drop the support of these older versions, but that can come later if need be.
> > Anyone going to update the generation of this code:
> > ```
> > error: ‘v8::Local<v8::Value> v8::Object::Get(uint32_t)’ is deprecated
> > ```
> > 
> > 
> > and similar seen in Travis testing? It probably breaks even newer versions given the generated code is using a deprecated API. @lissyx you mention v15. Can we add Travis tests for v14 (can be after this patch is merged in if need be).
> 
> I'm sorry, but I'm really unable to commit any time on looking into that now. All I can share is that we use https://github.com/lissyx/swig/tree/taskcluster for NodeJS v15 successfully, but it's only tested on our usage.

I tried to spend some time on this, but currently a bit stuck with other work. Everyone is welcome to give it a go first. Thanks.

vadz:
I'll get to this sooner or later if nobody else does because we're using this PR extensively with more changes for our own wrappers and I'd really like to get it into SWIG but IMO the priority should be to merge this just suppressing the deprecation warnings for now. We can always remove the use of deprecated API later, it's not _really_ a problem to use it and having an actually usable JS backend in SWIG is much more important. Just my 2c.

lissyx:
> > > @yegorich @tungntpham @dot-asm @lissyx who has taken over responsibility for this patch?
> > > I can't work out if this is the patch to use or #1736 or #1746. If there is any doubt, swig 4.1 can drop versions older than NodeJS v4.0.0, as discussed in #1520. I'll even consider dropping support for versions older than v6.0.0. I would like to see the code cleaned up to drop the support of these older versions, but that can come later if need be.
> > > Anyone going to update the generation of this code:
> > > ```
> > > error: ‘v8::Local<v8::Value> v8::Object::Get(uint32_t)’ is deprecated
> > > ```
> > > 
> > > 
> > > and similar seen in Travis testing? It probably breaks even newer versions given the generated code is using a deprecated API. @lissyx you mention v15. Can we add Travis tests for v14 (can be after this patch is merged in if need be).
> > 
> > 
> > I'm sorry, but I'm really unable to commit any time on looking into that now. All I can share is that we use https://github.com/lissyx/swig/tree/taskcluster for NodeJS v15 successfully, but it's only tested on our usage.
> 
> I tried to spend some time on this, but currently a bit stuck with other work. Everyone is welcome to give it a go first. Thanks.

Oh and I should mention that this branch include some tentative Travis CI work, but I abandonned it given the lead time (24h+) of travis and my lack of ability to properly reproduce the same issues locally to investigate: it was working on my system, and blowing away on Travis :/.

yegorich:
Unfortunately, I'm short on time too. I think, getting rid of Node.js versions older than v6.x should be made first or just part of this PR. This will greatly simplify both the code and CI. Just replacing `v8::Handle` occurrences to `v8::Local` would fix most of the incompatibilities.

dot-asm:
> the priority should be to merge this just suppressing the deprecation warnings for now.

I would agree with this. I can't dedicate undivided attention, but I'll do my best to follow it up.

wsfulton:
Committed. I've put into the release notes that v6 is now the minimum supported version. Patches to remove support up to v6 are welcome in order to simplify the code. Likewise, patches to fix the deprecated APIs are also welcome.

lissyx:
> Committed. I've put into the release notes that v6 is now the minimum supported version. Patches to remove support up to v6 are welcome in order to simplify the code. Likewise, patches to fix the deprecated APIs are also welcome.

Citing myself from https://github.com/swig/swig/issues/1520#issuecomment-785833516 it's possible for some people it will even work upto NodeJS v15 :)

arsnyder16:
@wsfulton My 2 cents.  If we swig is looking to future proofing the javascript nodejs code generation then its needs to get away from generating v8 and move to node-addon-api, or n-api. n-api was introduced in node version 8.x

If a maintainer is found it would be ideal if that was their focus. The reason the node team created these was to avoid the issues that swig experiences nearly every major release of node. 

https://github.com/nodejs/node-addon-api#node-addon-api-module
https://nodejs.github.io/node-addon-examples/

dot-asm:
> Committed. I've put into the release notes that v6 is now the minimum supported version. Patches to remove support up to v6 are welcome in order to simplify the code. Likewise, patches to fix the deprecated APIs are also welcome.

See #1956. More to come...

dot-asm:
> move to node-addon-api

Not that I have a say, but first thing to recognize is that it's a Node-specific thing. So the question would be if it means that there would have to be a **second** Javascript back-end? Of course one might make a case that Node is the only one that should be supported, but till one does, introducing additional back-end would be undesired...

arsnyder16:
That's fair point. There are already multiple javascript back ends including JavascriptCore/node(mostly via v8)/pure v8.  I can't speak to the amount of swig users actually producing pure v8, but there seems to be an enough of a node user base that it might be worthwhile to support node directly instead of supporting node through v8.  The node generation while mostly rooted in v8 does include node specific items such as node.h

Most of the changes that are happening in the javascript end of swig seem to be driven by supporting newer nodejs versions.  They don't seem to be driven by the direct v8 releases but more as a side effect of node releases.

dot-asm:
> Most of the changes that are happening in the javascript end of swig seem to be driven by supporting newer nodejs versions. They don't seem to be driven by the direct v8 releases but more as a side effect of node releases.

Not that it matters, but that's not how I view it. Each node.js release uses specific v8 version, so you really rather follow v8 versioning anyway. Either way, as already said, it's not like I have a say, it's something for maintainers to consider in the end:-)

dot-asm:
See #1957, #1958.

chenhengqi:
@wsfulton  Can you release a new version with this commit included ?

wsfulton:
We're working towards a new release and have just about recovered from our CI system Travis going offline and replacing it with Github Actions. We can't release without full testing. There are quite a few patches to be incorporated for the next release so will be a few months still.

Issue #1747, opened 2020-03-02
Error with quotes in Doxygen comment


mdouze:
Hi 

We are using SWIG to wrap a large C++ library (https://github.com/facebookresearch/faiss), thanks for this awesome tool!

We are planning to migrate to SWIG 4 to translate the Doxygen comments to python docstrings. 

It seems that the docstring generated when there are double quotes in the doxygen comments can be invalid in 4.0.1 and the current github version. 

To repro:  

```
matthijs@devvm4948:/tmp $ cat > tt.i
%module tt

/// blabla "oh"
void f();
matthijs@devvm4948:/tmp $ /data/users/matthijs/swig4/swig/installed/bin/swig  -doxygen -python -c++ tt.i
matthijs@devvm4948:/tmp $ python3 tt.i
  File "tt.i", line 1
    %module tt
    ^
SyntaxError: invalid syntax
```

vadz:
Thanks for the bug report, but you confused me for a moment there as I was wondering why would you run python on SWIG input file. The real problem is, of course:
```
% python3 tt.py
  File "tt.py", line 66
    r""" blabla "oh""""
                      ^
SyntaxError: EOL while scanning string literal
```

I.e. this happens when there is quote at the end of Doxygen comment and could be worked around by just adding a space before the closing `"""` in SWIG code. I wonder if we could do it unconditionally or could this have any negative consequences (that I don't see right now)?

mdouze:
ah sorry, yes it's a copy/paste error. 

A few additional tests: 

- if the quotes are at the beginning it's not a problem because SWIG inserts a space in front

- it triple quotes """" are used in the string it fails again. Not sure how that could happen in "normal" code.

vadz:
Both problems should be fixed in #1748.

wsfulton:
Shouldn't we generate the following if the last character is double quote:
```python
r""" blabla "oh""" '"'
```
or
```python
r""" blabla "oh""" "\""
```
I don't think we should solve it by adding in a space.

If I've understood the patch correctly, the string will then have a visible backslash in it??? I may have that wrong though.

vadz:
> Shouldn't we generate the following if the last character is double quote:
> 
> ```python
> r""" blabla "oh""" '"'
> ```
> 
> or
> 
> ```python
> r""" blabla "oh""" "\""
> ```

This would indeed work. It's, of course, rather ugly, so perhaps it would be better to generate
```python
r''' blabla "oh"''''
```
instead?

[...]

PR #1748, opened 2020-03-03
Fix quoting issues with docstrings in the generated Python

vadz:
Single-line Doxygen comments ending with a double quote resulted in
syntactically-invalid Python docstrings in the output, so escape the
last quote in this case to avoid it.

See #1747.


mdouze:
Looks good!

vadz:
I've updated the changes after the discussion, the new version shouldn't have any backslashes anywhere any more.

PR #1749, opened 2020-03-04
Tools/testflags.py: resolve pylint and pep8 issues

yegorich:
Add comments, reduce the line length, move imports to the beginning
of the file etc.


wsfulton:
This is a backwards step in my opinion and makes the code less readable. Sorry, but I'm rejecting this one.

PR #1751, opened 2020-03-05
Scilab 6.0.2 and 6.1.0 compatible gateway

### LINKED ISSUE #1737: SWIG Scilab generated gateways are not compatible with Scilab 6.1.0

davidcl:
This fix #1737 by upgrading the Scilab code generator to have Scilab 6.0 and 6.1 compatible semantics. The test suite has been upgraded to detect the scilab version when needed.

The Travis CI support #1742 should probably be merged before this pull request.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,7 +180,10 @@ jobs:
           VER: '3.0'
           CPPSTD: c++11
         - SWIGLANG: scilab
-          os: ubuntu-18.04 # scilab-6.1 in ubuntu-20.04 not yet working
+          VER: '5.5.2'
+        - SWIGLANG: scilab
+          os: ubuntu-18.04 # scilab 6.0
+        - SWIGLANG: scilab
         - SWIGLANG: tcl
         # c++11 testing
         - SWIGLANG: csharp
@@ -213,7 +216,9 @@ jobs:
           CPPSTD: c++11
         - SWIGLANG: scilab
           CPPSTD: c++11
-          os: ubuntu-18.04 # scilab-6.1 in ubuntu-20.04 not yet working
+          os: ubuntu-18.04 # scilab 6.0
+        - SWIGLANG: scilab
+          CPPSTD: c++11
         - SWIGLANG: tcl
           CPPSTD: c++11
         # c++14 testing
@@ -249,7 +254,9 @@ jobs:
           CPPSTD: c++14
         - SWIGLANG: scilab
           CPPSTD: c++14
-          os: ubuntu-18.04 # scilab-6.1 in ubuntu-20.04 not yet working
+          os: ubuntu-18.04 # scilab 6.0
+        - SWIGLANG: scilab
+          CPPSTD: c++14
         - SWIGLANG: tcl
           CPPSTD: c++14
         # c++17 testing (using gcc11)
@@ -298,7 +305,6 @@ jobs:
         - SWIGLANG: scilab
           CPPSTD: c++17
           GCC: 11
-          os: ubuntu-18.04 # scilab-6.1 in ubuntu-20.04 not yet working
         - SWIGLANG: tcl
           CPPSTD: c++17
           GCC: 11
--- a/Source/Modules/scilab.cxx
+++ b/Source/Modules/scilab.cxx
@@ -11,10 +11,11 @@
  * Scilab language module for SWIG.
  * --------------------------------------------------------------------------*/
 
+#include <cstddef>
+#include <cstdlib>
 #include "swigmod.h"
 
 static const int SCILAB_IDENTIFIER_NAME_CHAR_MAX = 24;
-static const int SCILAB_VARIABLE_NAME_CHAR_MAX = SCILAB_IDENTIFIER_NAME_CHAR_MAX - 4;
 
 static const char *usage = (char *) " \
 Scilab options (available with -scilab)\n \
@@ -25,7 +26,6 @@ Scilab options (available with -scilab)\n \
      -buildersources <files>                - Add the (comma separated) files <files> to the builder sources\n \
      -builderverbositylevel <level>         - Set the builder verbosity level to <level> (default 0: off, 2: high)\n \
      -gatewayxml <gateway_id>               - Generate gateway xml with the given <gateway_id>\n \
-     -targetversion <scilab_major_version>  - Generate for Scilab target (major) version (default: 5)\n \
 \n";
 
 
@@ -40,11 +40,11 @@ protected:
 
   String *variablesCode;
 
-  int targetVersion;
-
   bool generateBuilder;
   File *builderFile;
   String *builderCode;
+  String *builderCode5;
+  String *builderCode6;
   int builderFunctionCount;
 
   List *sourceFileList;
@@ -67,6 +67,9 @@ protected:
   bool createLoader;
   File *loaderFile;
   String *loaderScript;
+  String *loaderScript5;
+  String *loaderScript6;
+  int loaderFunctionCount;
 public:
 
   /* ------------------------------------------------------------------------
@@ -74,8 +77,6 @@ public:
    * ----------------------------------------------------------------------*/
 
   virtual void main(int argc, char *argv[]) {
-    targetVersion = 5;
-
     generateBuilder = false;
     sourceFileList = NewList();
     cflags = NewList();
@@ -140,12 +141,6 @@ public:
           createGatewayXML = true;
           gatewayID = NewString(argv[argIndex + 1]);
           Swig_mark_arg(argIndex + 1);
-        } else if (strcmp(argv[argIndex], "-targetversion") == 0) {
-          if (argv[argIndex + 1] != NULL) {
-            Swig_mark_arg(argIndex);
-            targetVersion = atoi(argv[argIndex + 1]);
-            Swig_mark_arg(argIndex + 1);
-          }
         }
       }
     }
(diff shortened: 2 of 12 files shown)


wsfulton:
ping @smarchetto, this is one for you to look at.

wsfulton:
This pull request introduces a new command line option `-targetversion`. We avoid version options as much as possible and use the C preprocessor to handle version differences. Can you please look at using this instead.

davidcl:
I removed the `targetversion` command-line argument introduced in 2017 and use runtime detection on loader/builder files to detect what are the expected library entry points depending on the Scilab version. IMHO this is similar to the C preprocessor approach.

wsfulton:
Hi @davidcl. I'm not sure what to make of this. Two of the four Scilab versions are broken during Travis testing. We can't have regressions like this.

Can @smarchetto (the Scilab maintainer) please opine on the pull request.

The removal of `-targetversion` is a good thing as it looks like the generated code will deal with all versions... this is how the other SWIG modules mostly work.

configure.ac
```diff
@@ -1231,18 +1231,18 @@ else
 
   if test -n "$SCILAB"; then
     # Check for Scilab version (needs api_scilab so needs version 5.3.3 or higher)
-    SCILAB_FULL_VERSION=`$SCILAB -version | head -1 | sed -e 's|Scilab version \"\(.*\)\"|\1|g'`
+    SCILAB_FULL_VERSION=`$SCILAB -nwni -version | head -1 | sed -e 's|Scilab version \"\(.*\)\"|\1|g'`
```
ojwb, line 1234: It seems unnecessarily confusing that `SCILAB_FULL_VERSION` here becomes `SCILAB_VERSION` in the makefiles under `Examples`, but there's also a `SCILAB_VERSION` here which is different (if I follow, the former is e.g. `6.0.1` while the latter is `601`).

I'd suggest renaming the `SCILAB_VERSION` here to something like `SCILAB_VERSION_NO_DOTS` and then `SCILAB_FULL_VERSION` here can just be `SCILAB_VERSION`.

Source/Modules/scilab.cxx
```diff
@@ -1136,18 +1095,67 @@ class SCILAB:public Language {
...
+   * Create a Scilab small identifier to be used by Scilab 5
+   * ----------------------------------------------------------------------- */
+
+  String* createSmallIdentifierName(String* name, int outputLen = SCILAB_IDENTIFIER_NAME_CHAR_MAX) {
+    char* s;
+    if (DohCheck(name)) {
+      s = (char *) Data(name);
+    } else {
+      s = (char *) name;
+    }
```
ojwb, line 1137: The implementation of `Data(obj)` includes `DohCheck(obj)` and returns `(void *) obj` for you if that returns false.  There's also a macro `Char(X)` which is defined as `((char *) Data(X))` so you can replace the code above with just:

```
char * s = Char(name);
```

configure.ac
davidcl: ok I will rename

Source/Modules/scilab.cxx
davidcl: thanks for the information, I will use `Char(X)` then

Source/Modules/scilab.cxx
```diff
@@ -938,33 +874,51 @@ class SCILAB:public Language {
...
    * saveBuilderFile()
    * ----------------------------------------------------------------------- */
 
   void saveBuilderFile(String *gatewayName) {
-    Printf(builderCode, "];\n");
+    Printf(builderCode5, "];\n");
+    Printf(builderCode6, "];\n");
+
+    Printf(builderCode, "ver = getversion('scilab');\n");
+    Printf(builderCode, "if ver(1) <= 5 & ver(2) <= 5 & ver(3) <= 2 then\n");
```
ojwb, line 910: This check looks wrong to me - for example, 5.4.3 and 4.6.0 would be treated as "6.0.0 or more" (perhaps those aren't actually scilab releases, but my point is that the logic of the check is wrong).

Why not just test `if ver(1) < 6 then` here?

davidcl: you are correct, this seems a logic bug

.github/workflows/ci.yml
```diff
@@ -215,10 +212,6 @@ jobs:
           CPPSTD: c++11
         - SWIGLANG: scilab
           CPPSTD: c++11
-          os: ubuntu-18.04
-        - SWIGLANG: scilab
-          CPPSTD: c++11
-          os: ubuntu-20.04
```
ojwb: After this change it looks like we're only testing the scilab version in `ubuntu-20.04`, which is 6.1.0.  Before we were testing 6.0.1 from `ubuntu-18.04` too.

wsfulton:
Also, can you please add in better user information about what all the changes in scilab.cxx are about are needed in the CHANGES.current file.

davidcl:
There is a rebase/merge issue on SWIG_xxxx_Init naming. I will fix that

.github/workflows/ci.yml
davidcl: You are correct, after looking at other languages, I removed `os` selection. IMHO it will be more maintainable in the future to avoid specific checking (as the code is now handling both version).

Should I explicitly check for 6.0.1 on ubuntu-18.04 to pass this commit and remove it later ?

Source/Modules/scilab.cxx
```diff
@@ -938,33 +872,51 @@ class SCILAB:public Language {
...
+    Printf(builderCode, "  // version is less or equal to 5.5.2\n");
+    Printf(builderCode, "  \n");
+    Append(builderCode, builderCode5);
+    Printf(builderCode, "  \n");
+    Printf(builderCode, "else\n");
+    Printf(builderCode, "  // version is 6.0.0 or more\n");
+    Printf(builderCode, "  \n");
+    Append(builderCode, builderCode6);
+    Printf(builderCode, "  \n");
+    Printf(builderCode, "end\n");
```
ojwb, line 918: If I follow, `builderCode5` is essentially the same as `builderCode6` except that identifiers are truncated (or slightly a slightly more complex adjustment for some cases) to be at most 24 characters.

Does that mean that if there aren't any long identifiers in an API that `builderCode5` will be identical to `builderCode6`?

If so, it seems worthwhile to have a special-case for that as it should give a smaller wrapper in many common cases - something like:

```
    if (Equal(builderCode5, builderCode6)) {
      Append(builderCode, builderCode6);
    } else {
      // The code above
    }
```

And probably similarly for `loaderScript5` vs `loaderScript6`.

.github/workflows/ci.yml
davidcl: I will add scilab 5.5 from scilab.org as the ci-linux-install.sh script has been updated to handle that.

Source/Modules/scilab.cxx
davidcl: Good idea ! done in the latest push

ojwb:
CI is failing for scilab 5.5.2:

https://github.com/swig/swig/runs/5223440768?check_suite_focus=true

Looks like configure fails to detect scilab:

https://github.com/swig/swig/runs/5223440768?check_suite_focus=true#step:5:130

davidcl:
The build and tests are now passing for 6.0 and 6.1 ; build is ok for 5.5.2 downloaded from scilab.org but some tests are failing. I will investigate.

ojwb:
> The build and tests are now passing for 6.0 and 6.1 ; build is ok for 5.5.2 downloaded from scilab.org but some tests are failing

Note that the build itself should be independent of the versions of languages found (and also whether languages are found or not) by configure - those languages should only be used when the testsuite or examples are run.  If SWIG built with some scilab versions installed but not others that would be a sign that there's an unwanted build-time use of the installed scilab.

davidcl:
Tests looks ok now.

There is an issue on Scilab 5.5.2 with the scilabconst() feature. Its variable is undefined on [Scilab 5.5.2](https://github.com/swig/swig/runs/5244733283?check_suite_focus=true#step:7:575) whereas it works on [Scilab 6](https://github.com/swig/swig/runs/5244733341?check_suite_focus=true#step:7:575). This might be a pre-existing bug in the code to manage scilabconst().

davidcl:
> Note that the build itself should be independent of the versions of languages found

I fully agree with you and that's the whole purpose of removing the `targetversion` argument in this pull request. However in the tests, there is not such check ; the same Scilab version is used at ./configure time and test time. Is there a test with a shared swig build (with all lang) and multiple lang runtimes on the system ?

ojwb:
>> Note that the build itself should be independent of the versions of languages found
>
> I fully agree with you and that's the whole purpose of removing the targetversion argument in this pull request.

That's related but not quite the same.  Removing `-targetversion` was about **generating wrapper code** that isn't specific to a particular scilab version.

My point here is that `configure` only detects the various target language interpreters (or toolchains for compiled languages) for the testsuite, and the `swig` executable should not depend on the results of those tests.

> However in the tests, there is not such check ; the same Scilab version is used at ./configure time and test time. Is there a test with a shared swig build (with all lang) and multiple lang runtimes on the system ?

No.  Theoretically it'd be a nice check that people haven't incorporated uses of the detected language headers or whatever into the `swig` executable, but in practical terms we don't end up with such code getting merged because such uses are obvious when the code is reviewed.

So I think just stick to the existing framework where SWIG gets rebuilt in CI for each language tested.  We use ccache so the redundant compilations should actually be very cheap, and the CI run time likely dominated by actually running the testcases.

ojwb:
Great, thanks - merged.

PR #1759, opened 2020-03-17
%feature("cs:argdef")

crhilton:
We need a way to specify default arguments for functions.  Sometimes
we can just take the arguments straight from c++ literals.  Sometimes
you can't, and need to specify those by hand.

So to make this work we:

 1. Shutoff handling all "defaultargs" variants of methods and constructors
    which have cs:argdef defined for their node.  This gets us a single
    constructor or method and skips the variants.
 2. Parse and spit out valid csharp arguments with defaults where specified.
    We allow `'*'` to specify all arguments, and `$value` to specify that the
    c++ literal is okay.  We always grab a specific named argument before
    falling back to `'*'`, and if you specify `'*:$value'` we ignore it if
    $value is not specified in c++.  If you specify `name:$value`, then you
    asked for it and we spit out the empty string for `$value` if there is none.

### DIFF
--- a/Source/CParse/parser.y
+++ b/Source/CParse/parser.y
@@ -6844,6 +6844,9 @@ exprcompound   : expr PLUS expr {
 		 $$.val = NewStringf("%s%s",qty,scanner_ccode);
 		 Clear(scanner_ccode);
 		 $$.type = SwigType_type(qty);
+		 if ($$.type == T_USER) {
+		   $$.type = T_UNKNOWN;
+		 }
 		 Delete(qty);
                }
                ;


vadz:
Thanks, this looks interesting and quite useful, although the feature syntax looks a bit unwieldy to me. I'm not sure if I can really propose something better though.

P.S. I've edited the description to avoid italicizing a big chunk of text and to remove the irrelevant commented out names of the files with conflicts.

crhilton:
> Thanks, this looks interesting and quite useful, although the feature syntax looks a bit unwieldy to me. I'm not sure if I can really propose something better though.
 
I'm not in love with it either, and am open to other ideas.  It's definitely awkward.

ojwb:
A way to do this which fits better with existing SWIG features would be something like:

```
%typemap(csdefaultarg) int z "4"
```

A limitation is that such a typemap would apply to any `int` parameter named `z` (while in some cases that's what you want in which case this is helpful, but in others it's not).  You could set this typemap before the function(s) you want to apply it to and clear it after, but that can be awkward to arrange if there's a mix of such functions in a single API header.

This limitation is also awkward for existing usage of `in` typemaps so it'd be more widely helpful to address it (or if there's a way to already, document it better).  Perhaps an attribute to specify the function (which would could match like a typemap matches), e.g.:

```
%typemap(csdefaultarg, function="Foo::bar") int z "4"
```

@wsfulton Thoughts?

wsfulton:
The proposed syntax is peculiar and not like anything I've seen before. The typemap suggestion is awkward as stated and makes it nearly impossible to use in a class which uses overloaded methods and repeatedly uses the same variable names.

Maybe the best solution is to use the fact the parameters of a feature are essentially a dictionary of named value pairs and use these to specify overrides. So by default use the C++ default value as parsed, assuming it compiles as C#, otherwise use the named key value pairs if specified as an override. For example:

```swig
%feature("cs:argdef", d=123) Foo::Foo;
%feature("cs:argdef", x=99) Foo::bar;

const int default_val = 123;
struct Foo {
  Foo(int a, int b=1, int c=2, const int& d=default_val) {}
  int bar(int x, int y=2, int z=3) { return x+y+z; }
};
```
to generate the proxy class wrappers:
```csharp
  public Foo(int a, int b=1, int c=2, int d=123) : this(examplePINVOKE.new_Foo__SWIG_0(a, b, c, d), true) {
  }

  public int bar(int x=99, int y=2, int z=3) {
    int ret = examplePINVOKE.Foo_bar__SWIG_0(swigCPtr, x, y, z);
    return ret;
  }
```

crhilton:
@wsfulton I think I've successfully implemented your proposed syntax.

If you start with just:
%feature("cs:argdef") Foo::bar;
Then you get the c++ default arguments literally translated over.

If you name individual arguments:
%feature("cs:argdef", d=1) Foo::bar;
Then you get defaults for everything but d, and d becomes the literal 1.

If you want to specify something that would make sense in csharp, but not in c++, you quote it:
%feature("cs:argdef", d="Math.Pi") Foo::bar;
That gets you d=Math.Pi in csharp.

If you want a string literal you have to include quotes in your literal in the swig bindings:
%feature("cs:argdef", d="\"a string\"") Foo::bar;


That's the behavior I got by using Getattr(the node, "feature:cs:argdef:the-parameter-name") for each argument for the function/method we're generating a proxy for inside of printArgumentDeclaration in csharp.cxx.  This does end up with quite a bit less code, and more straightforward code (no need for a - probably broken - custom parsing routine).  


I need to finish up documentation and tests before I resubmit.  Does the name of this ("cs:argdef") also need improvement?  Is the string escaping thing confusing for others aside from me? I don't have a great idea for how to avoid it using this syntax.  But it feels weird that literal ints and floats come over, but literal strings are interpreted as literal csharp code.  But I also need that to be true so you can put in literal csharp code (which is a use case we've had).

wsfulton:
The string quoting was something that I wasn't 100% sure of. I actually tried a few default, exactly like `Math.PI` and it is okay, I suspect/hope! all C# expressions are going to be parsed by SWIG. The example I tried this morning was the following most complex constant I could think of quickly and it worked:
```
%feature("cs:argdef", d=System::abc.x) Foo::Foo;
```
Maybe using your quote approach is better, not sure.

The name `cs:argdef` isn't great, I'll think of a better one...

ojwb:
> The typemap suggestion is awkward as stated and makes it nearly impossible to use in a class which uses overloaded methods and repeatedly uses the same variable names.

I think you missed my "which would could match like a typemap matches" note, so that's supported - e.g.:

```
%typemap(csdefaultarg, function="Foo::bar(int, int, int)") int z "4"
class Foo {
public:
   int bar(int x, int y, int z); // Adds a default here
   int bar(int z); // Doesn't add a default here
};
```

Your approach works here, but doesn't seem to extend to providing a way to address the similar issue with not being able to target an `in` typemap to a particular function, e.g. consider:

```
%typemap(in) int z "...";

// This is in a third party API header so we can't easily change the parameter names or add SWIG directives inside the class.
class Foo {
public:
   int bar(int x, int y, int z); // We want to apply custom "in" typemap to z here.
   int bar(int z); // But what if we don't want to also apply it here?
};
```

I've had to workaround this problem myself before.

crhilton:
I have no idea how to extend typemaps like that.  I'd need some pointers on how to do that, if that was the preferred route.

ojwb:
Re the naming, I'd suggest spelling out "default" in full as "def" could be short for "definition" in this context.

ojwb:
> Is the string escaping thing confusing for others aside from me? I don't have a great idea for how to avoid it using this syntax. But it feels weird that literal ints and floats come over, but literal strings are interpreted as literal csharp code.

The more helpful way to think about this is that the double quotes aren't part of the value - they are just how a parameter value is specified - **BUT** they can be omitted for some (simpler) values - so these are completely equivalent:

```
%feature("cs:argdef", d=1) Foo::bar;
%feature("cs:argdef", d="1") Foo::bar;
```

This is somewhat like quoting in a shell script.

Needing to backslash-escape double quotes in the value is annoying, but I think you can use raw strings to avoid that - e.g.:

```
%feature("cs:argdef", d=R"--("a string")--") Foo::bar;
```

Whether that's actually more readable is perhaps arguable, but it's at least not a SWIG-specific syntax.

crhilton:
I'm forming a new pull request with all of this.

PR #1766, opened 2020-03-25
fixed some missing semicolons. 

khoran:
Each one caused a compile error in generated R code. I discovered this while generating wrappers for R for the OpenBabel project. In the case of the r.cxx change, the newline didn't seem to make it into the output, causing two lines of code to appear on one line without a semicolon between them.

### DIFF
--- a/Lib/r/rtype.swg
+++ b/Lib/r/rtype.swg
@@ -94,20 +94,20 @@
   %{  $input = enumToInteger($input, "$R_class"); %}
 
 %typemap(scoercein) SWIGTYPE, SWIGTYPE *, SWIGTYPE *const, SWIGTYPE &, SWIGTYPE &&
- %{ if (inherits($input, "ExternalReference")) $input = slot($input,"ref") %}
+ %{ if (inherits($input, "ExternalReference")) $input = slot($input,"ref"); %}
 
 /*
 %typemap(scoercein) SWIGTYPE *, SWIGTYPE *const
-  %{ $input = coerceIfNotSubclass($input, "$R_class") %}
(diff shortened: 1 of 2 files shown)


wsfulton:
Hi @khoran, do you have a simple line or two of input C/C++ code that is fixed by the patch?

khoran:
I'm sorry, it's not easy to find the relevant source line of C++. The generated R file is 35k lines long, and I'm not super familiar with the OpenBabel code base myself. 

Here is an example of some bad generated R code: 
```
 # Start of OBSquarePlanarConfig_specified_set
   `OBSquarePlanarConfig_specified_set` = function(self, s_specified) {
     if (inherits(self, "ExternalReference")) self = slot(self,"ref")  s_specified = as.logical(s_specified);
     ;.Call('R_swig_OBSquarePlanarConfig_specified_set', self, s_specified, PACKAGE='openbabel_R'); }
   attr(`OBSquarePlanarConfig_specified_set`, 'returnType') = 'void'
   attr(`OBSquarePlanarConfig_specified_set`, "inputTypes") = c('_p_OpenBabel__OBSquarePlanarConfig', 'logical')
   class(`OBSquarePlanarConfig_specified_set`) = c("SWIGFunction", class('OBSquarePlanarConfig_specified_set'))
```
The missing semicolon is on the third line, right after `self = slot(self,"ref")`. I think this chunk relates to the code in [include/openbabel/stereo/squareplanar.h](https://github.com/openbabel/openbabel/blob/openbabel-3-0-0/include/openbabel/stereo/squareplanar.h). Not certain though.  Sorry I can't be more helpful. 

The code I'm generating from is [Openbabel 3](https://github.com/openbabel/openbabel).

Issue #252, opened 2014-10-26
"Error: Syntax error in input(3)" when any identifier is called "complex"


ThomasJClark:
This .i file produces a syntax error:

```
struct foo {
    int complex;
};
```

This one doesn't:

```
struct foo {
    int simple;
};
```

I guess this is because SWIG assumes that I'm using the `complex` macro, even though I didn't include complex.h.  `complex` isn't a reserved word in C, so it's perfectly valid to use it as an identifier.

ojwb:
If you aren't using `<complex.h>` anywhere, then you can define complex to something else which at least allows SWIG to process this example, and may provide a workaround for at least some situations:

```
$ swig -module xxx complex.i
complex.i:2: Error: Syntax error in input(3).
$ swig -module xxx -Dcomplex=complex_ complex.i
$
```

I'm wondering if SWIG should be changed to understand `_Complex` and not treat `complex` as a reserved keyword.  Then if you want to wrap an API which uses `complex`, you can just use `#define complex _Complex` in your SWIG interface file.

wsfulton:
This is fixed now and required one small change in addition to the patch in #1770.

PR #1770, opened 2020-04-02
Properly handle C99 complex types even in C++ mode

### LINKED ISSUE #1487: Fail to parse: typedef float _Complex
A standard header file generated by cgo when creating a shared library contains the following type definition:
```
typedef float _Complex GoComplex64;
typedef double _Complex GoComplex128;
```

When running with 
`swig -r example.i`
fails with 
`Error: Syntax error - possibly a missing semicolon.`
I would want a systematic/automatic way of solving the problem. Running Swig 3.0.12 on MacOS.

lpsinger:
Use the `_Complex` keyword rather than the `complex` macro.

Fixes #1487.

### DIFF
--- a/Examples/test-suite/complextest.i
+++ b/Examples/test-suite/complextest.i
@@ -137,7 +137,7 @@
   {
     return conj(a);
   }
-}
+%}
 
 
 complex Conj4(complex a);
--- a/Examples/test-suite/complextest.i
+++ b/Examples/test-suite/complextest.i
@@ -119,18 +119,6 @@
   {
     return conjf(a);
   }
-
-
-  complex Conj4(complex a)
-  {
-    return conj(a);
-  }
-
-
-  _Complex Conj5(_Complex a)
-  {
-    return conj(a);
-  }
 }
 
 
--- a/Examples/test-suite/complextest.i
+++ b/Examples/test-suite/complextest.i
@@ -119,19 +119,31 @@
   {
     return conjf(a);
   }
+}
 
 
-  complex Conj4(complex a)
+%{
+  /* Clang and GCC support complex or _Complex by themselves
+   * (implying double _Complex) as extensions to the C99 standard,
+   * but issue warnings about it. */
+
+  double complex Conj4(double complex a)
   {
     return conj(a);
   }
 
 
-  _Complex Conj5(_Complex a)
+  double _Complex Conj5(double _Complex a)
   {
     return conj(a);
   }
 }
+
+
+complex Conj4(complex a);
+
+
+_Complex Conj5(_Complex a);
 
 
 #endif
--- a/Examples/test-suite/complextest.i
+++ b/Examples/test-suite/complextest.i
@@ -72,15 +72,15 @@
 
 %inline
 {
-  complex Conj(complex a)
+  double complex Conj(complex a)
   {
     return conj(a);
   }
 
 
-  complex float Conjf(float complex a)
+  float complex Conjf(float complex a)
   {
-    return conj(a);
+    return conjf(a);
   }
 }
 
--- a/Examples/test-suite/complextest.i
+++ b/Examples/test-suite/complextest.i
@@ -119,6 +119,18 @@
   {
     return conjf(a);
   }
+
+
+  complex Conj4(complex a)
+  {
+    return conj(a);
+  }
+
+
+  _Complex Conj5(_Complex a)
+  {
+    return conj(a);
+  }
 }
 
 
--- a/Examples/test-suite/complextest.i
+++ b/Examples/test-suite/complextest.i
@@ -68,19 +68,68 @@
 
 
 %{
+#include <complex.h>
 %}
 
 %inline
 {
-  double complex Conj(complex a)
+  complex double Conj(complex double a)
   {
     return conj(a);
   }
 
 
-  float complex Conjf(float complex a)
+  complex float Conjf(complex float a)
   {
     return conjf(a);
+  }
+
+
+  double complex Conj1(double complex a)
+  {
+    return conj(a);
+  }
+
+
+  float complex Conjf1(float complex a)
+  {
+    return conjf(a);
+  }
+
+
+  _Complex double Conj2(_Complex double a)
+  {
+    return conj(a);
+  }
+
+
+  _Complex float Conjf2(_Complex float a)
+  {
+    return conjf(a);
+  }
+
+
+  double _Complex Conj3(double _Complex a)
+  {
+    return conj(a);
+  }
+
+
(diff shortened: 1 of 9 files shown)


wsfulton:
Hi @lpsinger. This pull request is an improvement as the C99 standard type is `_Complex` rather than `complex` as SWIG currently incorrectly implements. The SWIG header ccomplex.i is meant to provide support for the C header complex.h. If a user includes C's complex.h, then the macro `complex` expands to `_Complex`. So, the correct implementation is for SWIG's ccomplex.i to provide the `complex` macro and expand it to `_Complex`. The patch thus also needs modifying in parser.y to change `complex` to `_Complex` and not also accept `complex` for the `Cmp` functions. There are also a few other places where `complex` needs changing to `_Complex' in the source that your patch missed. Could you please make these changes and update your pull request with them?

lpsinger:
@wsfulton, see latest commit. Is that what you have in mind?

wsfulton:
@lpsinger yes the changes are what I had in mind. More or less. So now there is support for `float _Complex` and `double _Complex`, but nothing for `_Complex`. You deleted a number of lines for the plain `complex` type instead of changing it to `_Complex`, so they should be restored.

Additionally, ccomplex.i should have the following (like complex.h):
```
#define complex _Complex
```
Then the type `complex` will be available. Tests should test both _Complex and `complex` (if complex.i or ccomplex.i is included) and just _Complex if there these two .i files are not %include'd.

lpsinger:
I didn't realize that `_Complex` by itself worked. I thought it was only a type modifier. I'll put it back.

lpsinger:
Please take a look again.

ojwb:
Does this also fix #252?

wsfulton:
@lpsinger, thanks. So it seems, the standard does not support `_Complex` by itself. The Travis testing  failures using clang on MacOS indicate this:
```
complextest_wrap.c:2755:3: warning: plain '_Complex' requires a type specifier; assuming '_Complex double'

  _Complex Conj4(_Complex a)
```
gcc also warns when using `-pedantic`
```
warning: ISO C does not support plain 'complex' meaning 'double complex' [-Wpedantic]

    1 |   _Complex Conj4(_Complex a);
```
So, while they both accept this syntax and the changes you put back in enable SWIG to accept this syntax, I don't think we should test it as it will generate compiler warnings and errors. Can you modify the tests accordingly and also, given visual C++ doesn't support C99 (see the Appveyor failures), let's only test this if the compiler supports C99. I suggest taking an approach along these lines:
```swig
%{
#if defined __STDC_VERSION__ && __STDC_VERSION__ >= 199001L
#include <complex.h>
#else
#define _Complex
#define complex
#endif
%}
```
So the tests are hacked to not test `_Complex` if C99 is not supported. And arrange for the test so that SWIG parses plain `_Complex` but the compilers see `double _Complex`, so something like this in the .i file:
```swig
%{
double _Complex Conj4(double _Complex a);
%}
 _Complex Conj4(_Complex a);
```

lpsinger:
> and also, given visual C++ doesn't support C99 (see the Appveyor failures), let's only test this if the compiler supports C99. I suggest taking an approach along these lines:
> 
> ```c++
> %{
> #if defined __STDC_VERSION__ && __STDC_VERSION__ >= 199001L
> #include <complex.h>
> #else
> #define _Complex
> #define complex
> #endif
> %}
> ```

I don't think that MSVC supports either `_Complex` or `complex` as type specifiers. From https://docs.microsoft.com/en-us/cpp/c-runtime-library/complex-math-support:

> The Microsoft C Runtime library (CRT) provides complex math library functions, including all of those required by ISO C99. The compiler does not directly support a complex or _Complex keyword, therefore the Microsoft implementation uses structure types to represent complex numbers.

It goes on to explain that the Microsoft-specific types `_Fcomplex` and `_Dcomplex` are supported in place of the C99 standard `float _Complex` and `double _Complex` types.

lpsinger:
> @lpsinger, thanks. So it seems, the standard does not support `_Complex` by itself. The Travis testing failures using clang on MacOS indicate this:
> 
> ```
> complextest_wrap.c:2755:3: warning: plain '_Complex' requires a type specifier; assuming '_Complex double'
> 
>   _Complex Conj4(_Complex a)
> ```
> 
> gcc also warns when using `-pedantic`
> 
> ```
> warning: ISO C does not support plain 'complex' meaning 'double complex' [-Wpedantic]
> 
>     1 |   _Complex Conj4(_Complex a);
> ```
> 
> So, while they both accept this syntax and the changes you put back in enable SWIG to accept this syntax, I don't think we should test it as it will generate compiler warnings and errors.

I see the warnings when I compile a test program on my Mac, but I don't see any test failures in the macOS Travis jobs for this PR.

wsfulton:
Your latest patch removes the testing of `_Complex` altogether, so it no longer causes a failure. Please test this `_Complex` standalone with SWIG as I mentioned earlier using this kind of approach:
```swig
%{
double _Complex Conj4(double _Complex a);
%}
 _Complex Conj4(_Complex a);
```

You can see the failures on MacOS of your earlier patch here: https://travis-ci.org/github/swig/swig/builds/699403377. We use -Werror on Travis, hence the warnings turn into errors.

Agreed Visual C++ simply does not support C99 nor `_Complex`, hence you need to put in the macros I asked for to skip testing `complex` or `_Complex`. I'm just not interested in testing nor supporting a compiler that does not support the standard.

lpsinger:
> I'm just not interested in testing nor supporting a compiler that does not support the standard.

If we don't want to support the non-C99 bare `_Complex` type, should I simply remove it?

wsfulton:
We could, but we already support bare complex so please just test like I mentioned for `Conj4`.

lpsinger:
@wsfulton, I tried that fix for the GCC and Clang warnings of presenting `_Complex` to swig but `double _Complex` to the compiler, but it is still generating warnings like this because the bare `_Complex` type is used in the generated code:

```
complextest_wrap.c:2953:36: error: ISO C does not support plain 'complex' meaning 'double complex' [-Werror=pedantic]
 2953 | SWIG_AsVal__Complex  (PyObject *o, _Complex* val)
      |                                    ^~~~~~~~
...
```

wsfulton:
Okay. I'll pick it up from here as I have all these compilers installed locally and so it shouldn't be hard to find the best work around.

lpsinger:
Thank you. It has been a *long* time since I have used MSVC, and aside from continuous integration services I don't have it handy.

wsfulton:
@lpsinger, thanks, these are merged in now and I've tweaked the tests to work with non-c99 compilers. See  
d967e4c8b8...8245277ad and 4f18450...638ca8152d.

PR #1774, opened 2020-04-10
Support node12

brgirgis:
These changes are enough to get one of my large projects running with node12. I would be grateful if you could let me know how to proceed further.

PS: better late than never :)


brgirgis:
Just realized there is another PR for this. Feel free to reject.

Issue #1775, opened 2020-04-15
DOH_POOL_SIZE tweaking


yairl:
swig is awesome.

Running it on a large code base is sometimes very slow.
Large in this case means generated swig_wrap is ~700k lines for one module, and we have several similar ones) takes ~7 minutes on a nice, unloaded machine with 32 cores, 256G memory.

Some basic profiling showed that DohCheck() is very expensive; apparently just looking through the pools is now expensive as so much memory is allocated.
Changing the default pool size for 2^14 to 2^22 (16K->4M) made this go away, with processing dropping to <1 minute.

Will be nice if you can either tweak the default (4M isn't a lot these days), or add a argument to the swig binary to control that in runtime.

Highly appreciated, and thanks for swig.

ZackerySpytz:
I agree that the value of DOH_POOL_SIZE should be increased. I've opened GH-1776.

wsfulton:
Implemented by #1776.

PR #1776, opened 2020-04-18
Increase DOH_POOL_SIZE

ZackerySpytz:
Increase DOH_POOL_SIZE to 2^22.

Addresses GH-1775.

### DIFF
--- a/Source/DOH/memory.c
+++ b/Source/DOH/memory.c
@@ -15,7 +15,7 @@
 #include "dohint.h"
 
 #ifndef DOH_POOL_SIZE
-#define DOH_POOL_SIZE         16384
+#define DOH_POOL_SIZE         4194304
 #endif
 
 /* Checks stale DOH object use - will use a lot more memory as pool memory is not re-used. */


ZackerySpytz:
Perhaps 2^22 is too much.

yairl:
Personally, my take is that 4MB should be negligible nowadays, definitely for someone using swig.

An adapting allocator that doubles the pool size each time can be nicer fr a theoretical standpoiny but probably not worth the extra complexity :)

Thanks again.

wsfulton:
I couldn't see any real change in runtime when running our test-suite to just invoke the swig executable using 'make partialcheck-java-test-suite' and 'make partialcheck-python-test-suite'. The tests are fairly small, so that's good news that they didn't change. I also looked at the memory usage and time taken for the slowest running tests in the python test-suite (li_std_vector, li_std_list, std_containers) and there wasn't any real change either in memory usage nor time taken. std_containers_wrap.c is only 42K lines though, so somewhat smaller than the 700K lines mentioned in #1775.

Anyway, seems okay to me given the more typical smaller interface file usage is barely changed.

PR #1777, opened 2020-04-20
Fixed python required version check

tpaviot:
If ```-py3``` is enabled, as a consequence should check for python 3 in the python generated module.

Removed *later* from the exception log, indeed some python 2 releases (e.g. 2.7.15, may 2018) are more recent than python 3 (e.g. 3.1, released in february 2011).

This preliminary check is closer to an ```AssertionError``` than a generic ```RuntimeError```

At last, no need to check for any 2.8 or 2.9 releases since there won't be any, ever.

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -846,9 +846,13 @@ public:
 	Printv(f_shadow_py, "\n", f_shadow_begin, "\n", NIL);
 
       Printv(f_shadow_py, "\nfrom sys import version_info as _swig_python_version_info\n", NULL);
-      Printv(f_shadow_py, "if _swig_python_version_info < (2, 7, 0):\n", NULL);
-      Printv(f_shadow_py, tab4, "raise RuntimeError(\"Python 2.7 or later required\")\n\n", NULL);
-
+      if (py3) {
+      Printv(f_shadow_py, "if _swig_python_version_info < (3, 0):\n", NULL);
+      Printv(f_shadow_py, tab4, "raise AssertionError(\"Python 3.x required\")\n\n", NULL);
+      } else {
+      Printv(f_shadow_py, "if _swig_python_version_info < (2, 7, 0) or _swig_python_version_info >= (3, 0):\n", NULL);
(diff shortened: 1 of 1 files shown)


jschueller:
-1, you can still use bindings built without -py3 with python3

tpaviot:
@jschueller I mostly disagree. I agree that, if you omit ```-py3```, then you don't use any python3 specific feature. But the change I suggest deals with a check performed at runtime, not compile time. At runtime, there's absolutely not any chance that you can run a module built for python 2 (and thus linked against the python2 libraries) using python3. And the opposite is true as well.

jschueller:
main main concern is that this changes the usage of swig: you now have to specify py3 or not according to the python version
this will be annoying to handle in build systems that want to keep python2 compat a bit more

tpaviot:
Yes, the statement  ```or _swig_python_version_info >= (3, 0):``` leads to the behavior you describe. It could be removed. On my side, I would be in favor of something more explicit between py2 and py3.

This is related to an issue I opened at #1779

vadz:
FWIW I agree with @jschueller and, also, what exactly is the benefit of this change? I just don't see it...

ojwb:
As the help output says, `-py3` makes the output Python 3 specific, but as @jschueller says, code generated without `-py3` works with both Python 2 and Python 3 (just without using Python 3-specific features):

```
$ swig -python -help|grep  py3
     -py3            - Generate code with Python 3 specific features and syntax
```

Making the version check for >= 3.0 if `-py3` was specified seems reasonable (I'll push a change to do that shortly), but rejecting Python >= 3.0 if `-py3` wasn't specified would be break a scenario which currently works (and which we intend to work.)

> indeed some python 2 releases (e.g. 2.7.15, may 2018) are more recent than python 3 (e.g. 3.1, released in february 2011)

Here "later" is meant as a version number comparison, not in strictly temporal terms.  It's admittedly potentially imprecise wording, but it's a common way to phrase such things and I don't see anyone being genuinely mislead into thinking that 2.7.15 is later than 3.1.0.

PR #1785, opened 2020-04-30
add example of callback with ref/unref

pedia:


wsfulton:
Hi @pedia. Thanks for the pull request with example. The SWIG examples are intentionally a small set of examples attempting to show more classic use cases and try and have the same/similar example for each target language. I'd say the ref/unref feature is somewhat unusual and so in this case is better suited as a candidate in the test-suite where we do have an explosion of unusual use cases. The test-suite is also much better to make sure something works as intended, whereas examples are not really properly tested as we don't check the output.

We have a ref/unref callback testcase already in Examples/test-suite/director_ref.i. There is however only a Java runtime test of it in Examples/test-suite/java/director_ref_runme.java. If you could instead raise a new pull request to convert the Java example into a director_ref_runme.py file, that would be equally useful. Alternatively submit a new testcase based on your example into the test-suite if you think the existing testcase isn't close enough to your use case. It looks quite close to me though, with the main difference being inheritance of the reference counting class.

Issue #1757, opened 2020-03-10
Doxygen parser misses space after \endlink


HenningJ:
When using `\link ...\endlink` inside a sentence in the doxygen docu, the parser misses/ignores the space after the `\endlink` tag.

```
struct Foo
{
  /// This is a link to \link b another member\endlink inside a sentence.
  int a;
  int b;
};
```
=> `...another member</a>inside...`

mcfarljm:
Fixed in PR # 1788

wsfulton:
#1788 has been merged.

PR #1788, opened 2020-05-02
Fix for missing space after \endlink in doxygen

mcfarljm:
Fix for Issue #1757.

The "endlink" command is processed in `processWordCommands`, which by
default skips space occurring after the command, which is intended for
removing leading space from a command argument.  For "end" commands,
we don't want to do this.  Note that certain end commands such as
"endcode" aren't processed by `processWordCommands` (believe
`addCommandUnique` ends up handling them).

This fix may also improve behavior with other "end" commands, but nothing that is currently in the test suite.

Update usage of `\link` in `doxygen_translate_all_tags.i` to test handling
of space after `\endlink`.

Tweaking some of the usage in `doxygen_misc_constructs.i` to remove what
seems to be an extra space from the input (otherwise we would need to
add an extra space to the expected output).

### DIFF
--- a/Examples/test-suite/doxygen_misc_constructs.i
+++ b/Examples/test-suite/doxygen_misc_constructs.i
@@ -36,7 +36,7 @@
      * @param line line number
      * @param isGetSize if set, for every object location both address and size are returned
      *
-     * @link Connection::getId() @endlink <br>
+     * @link Connection::getId() @endlink<br>
      */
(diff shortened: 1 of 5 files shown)


vadz:
FWIW this looks right to me, thanks. But I don't know if I can merge this myself or if we need to wait for approval from @wsfulton ...

PR #1801, opened 2020-05-25
ruby: use correct wstring encoding

andreas-schwab:
Fixes ["./li_std_wstring_runme.rb:41:in `<main>'"]: FAILED: Li_std_wstring.test_equal_abc(s) -

### DIFF
--- a/Lib/ruby/std_wstring.i
+++ b/Lib/ruby/std_wstring.i
@@ -1,4 +1,6 @@
 %{
+#include <endian.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -15,9 +17,17 @@ extern "C" {
 #ifndef SWIG_RUBY_WSTRING_ENCODING
 
 #if WCHAR_MAX == 0x7fff || WCHAR_MAX == 0xffff
+#if BYTE_ORDER == LITTLE_ENDIAN
 #define SWIG_RUBY_WSTRING_ENCODING "UTF-16LE"
+#elif BYTE_ORDER == BIG_ENDIAN
+#define SWIG_RUBY_WSTRING_ENCODING "UTF-16BE"
+#endif
 #elif WCHAR_MAX == 0x7fffffff || WCHAR_MAX == 0xffffffff
+#if BYTE_ORDER == LITTLE_ENDIAN
 #define SWIG_RUBY_WSTRING_ENCODING "UTF-32LE"
+#elif BYTE_ORDER == BIG_ENDIAN
+#define SWIG_RUBY_WSTRING_ENCODING "UTF-32BE"
+#endif
 #else
 #error unsupported wchar_t size. SWIG_RUBY_WSTRING_ENCODING must be given.
 #endif


vadz:
AppVeyor CI failures seem completely unrelated, I don't know what's going on there.

Out of curiosity, which big endian platform do you use Ruby on?

andreas-schwab:
openSUSE supports ppc64 and s390x.

vadz:
In principle, we could test both of these on Travis too, but I don't know if it's worth it, i.e. whether people actually use SWIG(-generated code) on these platforms. If there are any users there, it would probably be worth it, even if SWIG already has so (too?) many Travis CI jobs, as I wouldn't be surprised if there are bugs in the other backends too.

kkaempf:
> whether people actually use SWIG(-generated code) on these platforms.

Well, SUSE and Redhat offer enterprise distributions on s390x and there's an active community around pp64.

The SWIG testsuite fails without this patch:
```
["./li_std_wstring_runme.rb:41:in `<main>'"]: FAILED: Li_std_wstring.test_equal_abc(s) -
```

vadz:
FYI I've created #1803 to test building Ruby in big endian environment.

vadz:
Returning to this PR, while AppVeyor failures and one of Travis CI ones are bogus, [this one](https://travis-ci.org/github/swig/swig/jobs/690854687) isn't, i.e. this breaks build with clang under macOS. `endian.h` seems to be Linux-specific, so we either need to restrict this to Linux or find a more portable way of testing for it, e.g. use configure to perform the test.

andreas-schwab:
There is AC_C_BIGENDIAN, but that needs to be made available at runtime.

vadz:
Sorry, I don't understand what do you mean by this. Using `AC_C_BIGENDIAN` gives you `WORDS_BIGENDIAN` macro if big endianness is detected. The only problem I know about with this autoconf macro is that it doesn't work when cross-compiling, but this can be overridden by using some command line option or environment variable.

andreas-schwab:
How does std_wstring.i get hold of the WORDS_BIGENDIAN macro?

vadz:
Oops, sorry, I got confused, you're right, we need this in the generated code and not in SWIG code itself, of course (which doesn't even use the same endianness anyhow).

So maybe your patch with `#ifdef __linux` around `endian.h` inclusion is indeed the best we can do. I'd use it to define some `SWIG_BIGENDIAN` macro, which could also be predefined externally in the (unlikely but possible) case that someone is building the generated code on a non-Linux BE system.

wsfulton:
This isn't a portable patch. Is there perhaps a way to get the endian/encoding from the Ruby headers instea?

What does this fix, is there a related test that should be referenced in this pull request?

vadz:
As mentioned in a [previous comment](https://github.com/swig/swig/pull/1801#issuecomment-633572981) the existing test suite fails without this patch on BE architectures -- and I've created #1803 to demonstrate it and you can indeed see the failure in [this build log](https://travis-ci.org/github/swig/swig/builds/691092085).

wsfulton:
Thanks @vadz for clarifications. I've merged #1803 so we can now see the failures. When this patch is portable, I'll consider merging it in. I'd like to see a Windows Ruby test on Appveyor to check the portability. If someone could help do that let please let me know.

vadz:
FWIW Appveyor does have Ruby with DevKit [installed](https://www.appveyor.com/docs/windows-images-software/#ruby), so it should be relatively straightforward to build Ruby extensions there. But I don't plan to do it myself in the near future.

wsfulton:
I've modified this patch to conditionally include `endian.h` in 58ffbe616fca22d815d17099b519097290735c89.

wsfulton:
See commit 6275af60a7586508ccf2a12378a08bfb7c40b26d - s390x testing on Travis now passes.

wsfulton:
@vadz, I had a go at setting up mingw Ruby testing, but there is some work to be done in configure.ac to provide the `-L` linker paths.

Also I concluded that testing visual c++ isn't going to be easy as Ruby provide Windows builds with mingw's gcc and the C extensions need to be compiled with the same compiler that compiled the Ruby interpreter. So I doubt visual c++ is used very much.

vadz:
Oh, yes, definitely, Ruby extensions are built using its DevKit, i.e. in MSYS2 environment. You can't use MSVS for this unless you rebuild Ruby itself with it too and AFAIK nobody does this.

PR #1802, opened 2020-05-25
Try to debug Python failures on AppVeyor

vadz:
Python is not being found any more for some reason, try to understand why.


vadz:
So it's the test `x"$PYSEPARATOR" = x'\\'` which fails now, even though `$PYSEPARATOR` is indeed `\`. Still have no idea what happened (bash version?), but we need to replace this test with something else.

vadz:
Damn, this is embarrassing. I've accidentally committed a change I didn't mean to push which broke this. Reverted now, sorry for the noise.

PR #1803, opened 2020-05-25
Test Ruby backend in big endian environment on Travis CI

vadz:
Check if the tests work correctly in big endian environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -351,6 +351,7 @@ matrix:
       sudo: required
       dist: xenial
     - os: linux
+      arch: s390x
       env: SWIGLANG=ruby CPP11=1
       sudo: required
       dist: xenial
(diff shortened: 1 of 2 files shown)


vadz:
As expected, this fails currently, but could be merged once #1801 is.

wsfulton:
@vadz, I'm not sure why the Travis build was cancelled, so I've restarted it on Travis to see the output with the 3rd and final commit. I'd like to see the effect of the addition of the libpcre3-dev install on build times and see if it works on s390x.

vadz:
I cancelled it because the build which did run already showed the failure in big-endian build and I didn't want to load Travis unnecessarily -- full build takes many hours, and why bother with running it if we already know that it will fail?

PR #1809, opened 2020-06-05
Fix wrapping of virtual comparison operators with directors

### LINKED ISSUE #1642: Directors code yields badly formatted comparison operators
When trying to wrap a C++ class definition which contains a virtual comparison operator for "==", ">=", "<=", or "!=", then the generated directors code  contains additional spaces around each "=" yielding invalid operator names, eg. "operator  =  = ".

Here are the input class definition, the swig command, and the resulting code. The used swig version is current git HEAD (4.0.2 - 97a107ed256b308d).

Input class definition **OperatorEQTest.hpp**:
```cpp
#pragma once

class OperatorEQTest
{
  public:
    virtual ~OperatorEQTest() {};
    virtual bool operator ==(const OperatorEQTest&) const;
};
```
Interface file **OperatorEQTest.i**:
```cpp
%module(directors="1") "OperatorEQTest"
%{
#include "OperatorEQTest.hpp"
%}
%feature("director");
#include "OperatorEQTest.hpp"
```
**swig command**:
```sh
swig  -Wall -ruby -c++ -includeall OperatorEQTest.i
```

The interesting part of the resulting code in **OperatorEQTest_wrap.h**
```cpp
class SwigDirector_OperatorEQTest : public OperatorEQTest, public Swig::Director {
public:
    SwigDirector_OperatorEQTest(VALUE self);
    virtual ~SwigDirector_OperatorEQTest();
    virtual bool operator  =  = (OperatorEQTest const &arg0) const;
};
```
and in **OperatorEQTest_wrap.cxx**
```cpp
[...]
bool SwigDirector_OperatorEQTest::operator  =  = (OperatorEQTest const &arg0) const {
  bool c_result = SwigValueInit< bool >() ;
  VALUE obj0 = Qnil ;
  VALUE SWIGUNUSED result;
[...]
```

I noticed that when I apply the following change the error is gone. However my knowledge about the "historic reason" for adding spaces around "=" is nonexistent, so I cannot say for sure that this solution is valid under all conditions or for satisfying all existing test cases which might expect such spaces in other places.
```diff
diff --git a/Source/Modules/directors.cxx b/Source/Modules/directors.cxx
index a91d5fd9a..c9de1df74 100644
--- a/Source/Modules/directors.cxx
+++ b/Source/Modules/directors.cxx
@@ -177,7 +177,7 @@ String *Swig_method_decl(SwigType *return_base_type, SwigType *decl, const_Strin
 
   // Reformat result to how it has been historically
   Replaceall(result, ",", ", ");
-  Replaceall(result, "=", " = ");
+  //Replaceall(result, "=", " = ");
 
   Delete(args_string);
   Delete(popped_decl);
```

ZackerySpytz:
Closes #1642.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -184,6 +184,7 @@ CPP_TEST_CASES += \
 	director_classes \
 	director_classic \
 	director_constructor \
+	director_comparison_operators \
 	director_conversion_operators \
 	director_default \
 	director_detect \
--- a/Examples/test-suite/director_comparison_operators.i
+++ b/Examples/test-suite/director_comparison_operators.i
@@ -1,0 +1,16 @@
+%module(directors="1") director_comparison_operators
+
+%include "std_string.i"
+%feature("director");
+
+%inline %{
+class Foo {
+public:
+  virtual ~Foo() { }
+  virtual bool operator==(const Foo&) const = 0;
+  virtual bool operator>=(const Foo&) const = 0;
+  virtual bool operator<=(const Foo&) const = 0;
+  virtual bool operator!=(const Foo&) const = 0;
+  virtual std::string test(const char *foo="a=1,b=2") { return foo; }
+};
+%}
--- a/Examples/test-suite/director_comparison_operators.i
+++ b/Examples/test-suite/director_comparison_operators.i
@@ -3,6 +3,13 @@
 %include "std_string.i"
 %feature("director");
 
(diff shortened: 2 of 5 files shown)


ZackerySpytz:
There were some test failures. I'll take a look in a bit.

wsfulton:
The test failures are due to a cygwin packaging issue which seems to be resolved now, so can be ignored.

PR #1812, opened 2020-06-07
Fix the error handling for the PyObject_GetBuffer() calls in pybuffer.i

### LINKED ISSUE #1640: pybuffer.i crashes if PyObject_GetBuffer() fails.
PyBuffer_Release() must not be called if PyObject_GetBuffer() fails, but the macros in pybuffer.i call PyBuffer_Release() before they check the return code from PyObject_GetBuffer().  This will cause these typemaps to crash if the object passed in can't provide a buffer.

see [here](https://reviews.llvm.org/D67793) for an actual example of how this can crash.

ZackerySpytz:
Closes #1640.

### DIFF
--- a/Examples/test-suite/python/python_pybuffer_runme.py
+++ b/Examples/test-suite/python/python_pybuffer_runme.py
@@ -44,3 +44,27 @@ else:
     buf3 = bytearray(b"hello")
     python_pybuffer.title1(buf3)
     check(buf3 == b"Hello")
+
+    try:
+        python_pybuffer.func1(1)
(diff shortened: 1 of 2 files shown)


wsfulton:
There is a testcase called python_pybuffer.i that tests pybuffer.i. Could you modify this please for future regression testing.

ZackerySpytz:
@wsfulton Hi.
> There is a testcase called python_pybuffer.i that tests pybuffer.i.

I'm aware of that, but I originally thought that this patch didn't need tests.

PR #1813, opened 2020-06-08
Display example/test standard output/error with example/test name appended

kwwette:
When running the examples / test suite in parallel (e.g. `make -j2`) it can be difficult to identify the source of error messages from run-me scripts, since they are not guaranteed to appear after the relevant `checking ...` message. This can make it difficult to track down which run-me script is causing the issue, particularly in Travis jobs which take time to re-run.

This PR modifies the examples / test suite Makefile rules to capture the standard output and error from run-me scripts (which are written to temporary files), which are then re-displayed with the name of the example / testcase prepended. If the run-me scripts returns a non-zero exit status, the standard output and error are always output; otherwise they are passed to the value of the Makefile `RUNPIPE` variable, which e.g. might be `>/dev/null` to discard standard output from passing tests.

The additional output capture code is implemented in the Makefile `RUNCAPTURE` macro in `Examples/Makefile.in` (examples) and `Examples/test-suite/common.mk` (test suite). When run-me scripts are invoked, `RUNCAPTURE` is appended to the end of the command-line to pipe the standard output and error to temporary files (`.runcapture_std{out|err}_$$.log`, where `$$` is the process ID for uniqueness), use `sed` to prepend the example / testcase name before re-outputting, deleting the temporary files and returning the run-me script exit status.

I've adapted all languages to use the `RUNCAPTURE` macros, with the exception of R where it appears non-trivial to capture standard output and error in the usual way.

With this PR, I can e.g. now see where the warning message in the Octave OSX build is coming from:

> ...
> checking octave testcase smart_pointer_member (with run test)
> checking octave testcase smart_pointer_multi (with run test)
> smart_pointer_member stderr: /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/octave-help-igwTvm:119: Unknown command `codequoteundirected'.
> smart_pointer_member stderr: warning: print_usage: Texinfo formatting filter exited abnormally
> smart_pointer_member stderr: warning: print_usage: raw Texinfo source of help text follows...
> ...


kwwette:
@wsfulton This should be ready for your approval. After which, I will be able to better diagnose some of the Octave test failures.

wsfulton:
Hi @kwwette, this is quite neat, but I'm afraid that I am not a fan at all of splitting stdout from stderr as that can be even more confusing w.r.t. the order of the output. If you can implement this keeping the default behaviour as is and triggering the feature via an env variable, then I may consider it. I think in all honesty though to keep things simpler, just remove the `SWIGJOBS=-j2` in .travis.yml for debugging and run a reduced set of tests. You can test your own branch on your own repo on Travis easily enough for experimenting/diagnosing. I do this a lot for diagnosing problems that I can't replicate locally.

kwwette:
@wsfulton OK fair enough. I'm afraid I don't have time to improve this further - and as you say, just removing the `SWIGJOBS=-j2` is probably good enough for debugging errors.

PR #1821, opened 2020-06-11
configure: change $PKGCONFIG to $PKG_CONFIG

vapier:
The PKG_CONFIG variable name aligns with the standard upstream
value that the pkg-config project itself uses.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -459,7 +459,7 @@ else
   alllang_default=yes
 fi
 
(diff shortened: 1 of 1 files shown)


vapier:
not sure if people want to maintain support for the old `$PKGCONFIG` name.  doesn't seem like a big deal when it only applies to the configure/compile phase, not the runtime/deploy phase.

PR #1822, opened 2020-06-12
configure: use AC_PATH_TOOL for pkg-config

vapier:
AC_CHECK_PROGS will find the tool using exactly the name given: it
only searches for "pkg-config".  When doing native builds, this is
generally fine.  However, when cross-compiling, this is not ideal
as `pkg-config` is often configured for the build system, not the
system we want to cross-compile for.

Switch to using the AC_PATH_TOOL tool instead.  This will look for
"pkg-config" with a $host- prefix first before falling back to the
plain "pkg-config".  When doing native builds, things should still
behave the same, but now things work better out of the box when we
cross-compile.

For example, `./configure --host=aarch64-linux-gnu` will first look
for "aarch64-linux-gnu-pkg-config" before falling back to the plain
"pkg-config".

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -459,7 +459,7 @@ else
   alllang_default=yes
 fi
 
-AC_CHECK_PROGS([PKG_CONFIG], [pkg-config])
+AC_PATH_TOOL([PKG_CONFIG], [pkg-config])
 
 #--------------------------------------------------------------------
 # Look for Tcl


wsfulton:
Currently we have:
```
AC_PREREQ(2.60)
```
in configure.ac and I was wondering if we need to increase this. Which version of autoconf first provided `AC_PATH_TOOL`?

vapier:
Thanks, that's a good thing to double check.  The good news is that we're all set -- this macro was first introduced in 2.60.
https://lists.gnu.org/archive/html/autotools-announce/2006-06/msg00002.html

PR #1823, opened 2020-06-14
Unwrap director classes only when returning a pointer or reference to an object

treitmayr:
This PR contains the following things:
1. For template instances some post-processing code is added to fix the 'decl' and 'type' attributes of functions. This restores the splitting of the function's return type between 'decl' and 'type', on which other parts in Swig rely on.
The attribute corrections are printed to the console if the commandline option {{--debug-template}} is provided to Swig.
1. Properly count the number of references and pointers in the return type of a function and only generate unwrapping code if this number is 1. The previous check for a trailing ").r." resp. ").p." is not sufficient in situations outlined in swig#1811.
 1. Add a test case which checks the two previous items.

This PR fixes swig#1811.


treitmayr:
Did one of the maintainers get a chance to look at this PR? Does the fix look reasonable, shall the commits be split somewhat, shall things be done in a different way (e.g. this counting of pointer/reference indirections)?

ojwb:
I resolved the minor conflict so the CI should test a merge with current git master.

@wsfulton Could you take a look at this?

It's a pretty clean patch with test coverage, and the changes look plausible to me, but I have no idea if the "apparent SWIG convention" it assumes is actually correct or not.

wsfulton:
I've added this for looking at before we release swig-4.1. The analysis of the problem seems okay. I need to examine the solution.

ojwb:
Thanks.

The CI had 2 bogus failures plus a legitimate failure on ocaml for the new testcase.  I tried asking CI to rerun hoping it would just redo the failed jobs but it really does try to rerun *all* the jobs which is not helpful at all so I cancelled that and the status here now unfortunately reflects that cancelled run.  There was also a run on just the branch (with only one bogus failure plus the ocaml failure) here: 

https://github.com/treitmayr/swig/runs/5366998682?check_suite_focus=true

Ocaml is an experimental target language so allowed to fail, but still not great to be introducing a new failure when it was clean before.

wsfulton:
Which part is required just for non-templates? I'd like a distinction and understanding for the non-template separately and then the template case given the complexity of the patch.

jschueller:
need some help ? I can try to rebase so as to run the tests and see what's up with the template/non-template part

wsfulton:
Hi @jschueller, not to worry, I've figured it all out now and I agree with the analysis and the need for this be fixed.

I'll add in a summary of it all when I merge it in fairly soon and an additional improvement I spotted while analysing this all.

jschueller:
ok, good news!

wsfulton:
I fixed the OCaml problem in 0239ba5536105bec1ac4c2386b10a300e9a483cd.

For the 3 points when the pull request was raised:
1. The "decl" and "type" attribute corrections are good and is the right thing to do. However, It turns out that none of the SWIG code base requires this correction, I think because "decl" is not used and "type" is always modified before presenting it to the target language module.
2. I've put in an improved fix in 2f55379687a394e941a20e2224eb1125d66720cd which also fixes typedefs by resolving the return type when determining whether to unwrap. Note that the more conventional "type" attribute is used rather than "decl", so template fix in 1. is not actually required.
3. Good testcase. I've enhanced it for the typedef case.

treitmayr:
Sorry for not answering right away. As I am now working on a completely different project (at work) not related to anything SWIG, but I just saw that you were able to even improve my PR, so all the better! Thanks for merging!

PR #1824, opened 2020-06-14
Perform proper spacing in director method declarations

treitmayr:
If a director method returns a const pointer, eg. `int *const`, then in its method declaration a space has to be inserted between 'const' and the method name.
This fixes swig#1810.


treitmayr:
There was some error in the Travis CI build during the final cleaning of the directory tree which however does not seem to be related to this PR (at least I do not know that the changes would have cause something like that):
```
configure: error: newly created file is older than distributed files!
Check your system clock
```

wsfulton:
That happens quite often and no-one has tried to find a fix yet, so you ignore that error. I just restarted it again anyway.

Issue #1850, opened 2020-07-30
error "SWIG wrapped code invalid in 32 bit architecture, regenerate code using -DSWIGWORDSIZE32" 


haolongzhangm:
1: on windows with 64 bit , build for x64 ,  so add vcvarsall.bat x64 args, also add CXXFLAGS=-m64 for Ninja

2: so i need enable  -DSWIGWORDSIZE64 for  swig 

then  build will pop error:  error "SWIG wrapped code invalid in 32 bit architecture, regenerate code using -DSWIGWORDSIZE32" 

swig version: https://nchc.dl.sourceforge.net/project/swig/swigwin/swigwin-4.0.2/swigwin-4.0.2.zip 

if I  force add -DSWIGWORDSIZE32 for swig,  build will ok, BUT runtime logic is wrong: exe is bit bit 
file info: _mgb.pyd: PE32+ executable (DLL) (GUI) x86-64, for MS Window,when python  is also a 64bit 
exe, then ptr will be  8 bytes, but when set -DSWIGWORDSIZE32 the wrap.cxx data type is 4 bit , so wrong err

so how to config -DSWIGWORDSIZE64 on windows 64 bit env

haolongzhangm:
any update ?

haolongzhangm:
pls review: https://github.com/swig/swig/pull/1855

Mizux:
you're wrong, on UNIX system std::int64_t are defined as long int, contrary to MSVC thus only here we need to use  `-DSWIGWORDSIZE64` windows take a very conservative approach so you don't need this even in 64bits...

Please take a look at https://github.com/Mizux/cmake-swig/blob/master/doc/swig.md

haolongzhangm:
so uintptr_t size should be 4 or 8 on windows 64?  I think should be 8

BUT at Lib/stdint.i 

 #if defined(SWIGWORDSIZE64)
typedef long int               intptr_t;
typedef unsigned long int      uintptr_t;
 #else
 typedef int                    intptr_t;
 typedef unsigned int           uintptr_t;

both of long int, unsigned long int , int, unsigned int    is 4bit on windows  64

so how to compat this?

Mizux:
good question...
for your information on Archlinux x86_64
```sh
%grepc -rn "typedef.*intptr_t;" /lib/gcc
/lib/gcc/x86_64-pc-linux-gnu/10.1.0/include/stdint-gcc.h:83:typedef __INTPTR_TYPE__ intptr_t;
[...]

PR #1855, opened 2020-08-04
fix windows(64) bit define issue:

haolongzhangm:
long bit wide is 4 at VC and LLVM-clang-cl env,
but ptr should be 8 at 64bit env,
even then long take 8 bit at linux-64/macos/mingw64,
but long long take 8 bit at all 64bit env,

### DIFF
--- a/Lib/stdint.i
+++ b/Lib/stdint.i
@@ -86,8 +86,8 @@ typedef unsigned long long int	uint_fast64_t;
 
 /* Types for `void *' pointers.  */
 #if defined(SWIGWORDSIZE64)
-typedef long int		intptr_t;
-typedef unsigned long int	uintptr_t;
+typedef long long		intptr_t;
+typedef unsigned long long	uintptr_t;
 #else
 typedef int			intptr_t;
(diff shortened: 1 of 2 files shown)


Mizux:
Please close this PR it will break UNIX/Linux world.

On windows simply don't enable it ;)

note: while on unix (x86_64) we have `sizeof(long long)  == sizeof(long int) == 8`, `long long` and `long int` are **TWO** differents types with no implicit cast...

Mizux:
Okk my bad windows 64bits is same as `SWIGWORSIZE32` **EXCEPT** for `[u]intptr_t` where it should use the builtin `__int64` type (see https://github.com/swig/swig/issues/1850#issuecomment-677786849 for details)

ojwb:
`long long` and `long int` are distinct types, but they are implicitly convertible.  However this doesn't look like the right change to me either.

It's not great that the test suite still passes with this change applied though.

ojwb:
This isn't the right change, so closing.  We have #1850 open to track the motivation behind this.

PR #1857, opened 2020-08-12
Raising IndexError in calls to __getitem__/__setitem__ for std::pair in Python

ThomasHauth:
The current implementation for the python std_pair interface does supply the proper size of the pair via the ```__len__``` method:

```
    def __len__(self):
        return 2
```

However, python 3 (tested with pyton 3.8.2 on Ubuntu 20.04) and python 2 (tested with 2.7.18rc1 on Ubuntu 20.04) does not use the sequence size returned by ```__len__``` but calls ```__getitem__``` until and ```IndexError``` exception is raised. Here we run into the problem with the current implementation as it does newer throw this exception:

```
    def __getitem__(self, index): 
        if not (index % 2):
            return self.first
        else:
            return self.second
```

Therefore, very common python assignments used with lists and tuples in python like:

```
full_list = [23, 24]
l1, l2 = full_list
```

`result1, result2 = swigWrappedMethodReturningStdPair()`

or for loops:

```
for it in full_list:
   print (it)
```

Cannot be done with the SWIG-generated pair. In case of the multiple assignments, the error will be that there are too many values to unpack and the for loop will never terminate and result in an endless loop.

According to the Python documentation, _"```IndexError``` should be raised fora value outside the set of indexes for the sequence (after any special interpretation of negative values)"_ (see https://docs.python.org/3/reference/datamodel.html#object.__getitem__)
The same behaviour is expected from the ```__setitem__``` method, which is also modified by this PR.

This PR increases the conformity of the SWIG-generated pair with other python sequences like list and tuple by throwing an ```IndexError``` exception when the third (non-existent) item is attempted to be accessed.

This will enable the above example usage scenarios also for the auto-generated SWIG pair for Python.

### DIFF
--- a/Lib/python/std_pair.i
+++ b/Lib/python/std_pair.i
@@ -189,15 +189,19 @@ SwigPython_std_pair_setitem (PyObject *a, Py_ssize_t b, PyObject *c)
 def __repr__(self):
     return str((self.first, self.second))
 def __getitem__(self, index): 
-    if not (index % 2):
+    if index == 0:
         return self.first
-    else:
+    elif index == 1:
         return self.second
-def __setitem__(self, index, val):
-    if not (index % 2):
-        self.first = val
     else:
-        self.second = val%}
+        raise IndexError
+def __setitem__(self, index, val):
+    if index == 0:
+        self.first = val
+    elif index == 1:
+        self.second = val
+    else:
+        raise IndexError%}
 }
 #endif
 %enddef


wsfulton:
Confirmed this is a problem.

Thanks for the patch @ThomasHauth. It does fix it using SWIG's default flags, however, it doesn't fix it for `-builtin`. The current implementation for `__getitem__`, `__setitem__` as well as `__len__` & `__repr__` in Python code is also sub-optimal because it is implemented in Python, not C. I'd like to see the entire implementation moved to C and then we can use the one implementation for non-builtin and builtin too. The C implementation for builtin can be used as the starting point:

https://github.com/swig/swig/blob/46e17f1458e312c5a5b4213adefffbdab4b05540/Lib/python/std_pair.i#L148-L152

and the conditional compilation for `SWIGPYTHON_BUILTIN` replaced by using %extend along the lines of what is used for the other STL containers. A really robust implementation will need to implement this using slices along the lines of the STL container wrappers, but somewhat simpler given the length is always 2. See

https://github.com/swig/swig/blob/46e17f1458e312c5a5b4213adefffbdab4b05540/Lib/python/pycontainer.swg#L868
https://github.com/swig/swig/blob/46e17f1458e312c5a5b4213adefffbdab4b05540/Lib/python/pycontainer.swg#L880

The patch will need to include a test. I have just committed some basic tests to our `li_std_pair` testcase in a38f30a19475603f0b97ad4d1db9a1e1e2c02b54, you can modify it with a simple one liner to cover this bug.

ThomasHauth:
@wsfulton thanks for your explanations. I will have a look how to unify the builtin/non-builtin variants of the std::pair to python mapping. Concerning the code in pycontainer.swg: Do you think it can be reused for parts of the std::pair implementation or is it better to port the relevant code to stl_pair.i ? Thanks!

wsfulton:
I wouldn't try to share that code. The std::pair implementation should be **much** simpler because it is fixed to 2 elements.

wsfulton:
Hi @ThomasHauth, did you get anywhere with this?

PR #1859, opened 2020-08-13
javascript: replace exceptions with SWIG_exit

### LINKED ISSUE #1858: SWIG host tool cannot be built without C++ exception support
The SWIG core is using exceptions in two places:

Source/Modules/javascript.cxx:      throw "Illegal state.";
Source/Modules/javascript.cxx:      throw "Illegal state.";

When building for Android as a host tool, C++ exception support is disabled by default. These cases are not obviously good uses of exceptions, as they cannot be handled in a useful way. Also, other parser errors in this file are handled with a printf and SWIG_exit(), which looks reasonable for these cases as well.

adelva1984:
When building SWIG for Android, there is no support for C++ exceptions.

In the cases there is "Illegal state", it seems more like an internal
error, so we can replace the throw calls with a debug print and exit
immediately.

Closes #1858

### DIFF
--- a/Source/Modules/javascript.cxx
+++ b/Source/Modules/javascript.cxx
@@ -1575,7 +1575,8 @@ void JSCEmitter::marshalInputArgs(Node *n, ParmList *parms, Wrapper *wrapper, Ma
       Printf(arg, "argv[%d]", i);
       break;
     default:
-      throw "Illegal state.";
+      Printf(stdout, "Illegal state.");
+      SWIG_exit(EXIT_FAILURE);
     }
     tm = emitInputTypemap(n, p, wrapper, arg);
     Delete(arg);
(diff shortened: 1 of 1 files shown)


Source/Modules/javascript.cxx
```diff
@@ -2212,7 +2213,8 @@ void V8Emitter::marshalInputArgs(Node *n, ParmList *parms, Wrapper *wrapper, Mar
       Printf(arg, "args[%d]", i);
       break;
     default:
-      throw "Illegal state.";
+      Printf(stdout, "Illegal state.");
```
wsfulton, line 2216:
```suggestion
      Printf(stderr, "Illegal MarshallingMode.");
```

Source/Modules/javascript.cxx
```diff
@@ -1575,7 +1575,8 @@ void JSCEmitter::marshalInputArgs(Node *n, ParmList *parms, Wrapper *wrapper, Ma
       Printf(arg, "argv[%d]", i);
       break;
     default:
-      throw "Illegal state.";
+      Printf(stdout, "Illegal state.");
```
wsfulton, line 1578:
```suggestion
      Printf(stderr, "Illegal MarshallingMode.");
```

wsfulton (changes requested):
Please modify patches with the 2 suggestions and I'll merge it in. Thanks.

PR #1861, opened 2020-08-13
naming: Add unreachable return to !HAVE_PCRE path

### LINKED ISSUE #1860: naming.c build fails with -Werror=no-return
When building SWIG for Android with an external build system that forces -Werror=no-return on, and HAVE_PCRE is not set, the naming.c file will fail because 'int name_regexmatch_value(..)' returns no value.

This could be fixed by marking SWIG_exit() __attribute__((noreturn)) or C++11 [[noreturn]], but for compiler portability (pre-C++11) it's easier to just make the wrapping function return something, which matches the same !HAVE_PCRE path in Swig_string_regex() in misc.c.

adelva1984:
Android builds all host tools with -Werror=no-return, which generates a
false positive in name_regexmatch_value() if HAVE_PCRE is not present.

Fix this by adding a return code to the !HAVE_PCRE path. This return
will not be reached but will suppress the compiler warning.

If/when SWIG can require C++11 compilers, a better fix would be to make
SWIG_exit() [[noreturn]].

Closes #1860

### DIFF
--- a/Source/Swig/naming.c
+++ b/Source/Swig/naming.c
@@ -1132,6 +1132,7 @@ static int name_regexmatch_value(Node *n, String *pattern, String *s) {
   Swig_error("SWIG", Getline(n),
              "PCRE regex matching is not available in this SWIG build.\n");
   SWIG_exit(EXIT_FAILURE);
+  return 0;
 }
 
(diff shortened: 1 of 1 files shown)


wsfulton:
Thanks!

PR #1868, opened 2020-09-05
Fix C# wchar_t* csvarout to be same as csout

geefr:
Bug and fix as reported on user mailing list - Current csvarout tries to return an IntPtr instead of C# string, causing a compilation error.

I can't find any test cases for this, can add if needed.

As an aside I don't see any wchar.i typemaps for directors, and the string helper class doesn't seem to be used anywhere..

### DIFF
--- a/Examples/test-suite/csharp/li_std_wstring_runme.cs
+++ b/Examples/test-suite/csharp/li_std_wstring_runme.cs
@@ -106,6 +106,13 @@ public class runme
                 check_equal(received, expected);
             }
 
+            foreach (string expected in test_strings)
+            {
+                s.wchar_t_ptr_member = expected;
+                string received = s.wchar_t_ptr_member;
+                check_equal(received, expected);
+            }
+
             /* Not working for Japanese and Russian characters on Windows, okay on Linux
              * Is fixed by adding CharSet=CharSet.Unicode to the DllImport, so change to:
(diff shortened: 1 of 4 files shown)


wsfulton:
There are various tests for `wchar_t`, a lot of them are mixed in with `std::wstring` tests (wstring.i includes wchar.i). However, I don't see one for a `wchar_t*` member variable anywhere. Adding in a li_wchar.i testcase would be fine, but as support is not wide in all the various target languages, you'll need to only test for a subset of languages like it is in https://github.com/swig/swig/blob/975f8fcfdba56294bb190d745cdd449a52e633f4/Examples/test-suite/li_std_wstring.i#L3-L4

There is a director test for wchar in https://github.com/swig/swig/blob/975f8fcfdba56294bb190d745cdd449a52e633f4/Examples/test-suite/director_wstring.i#L43. Please see #1706 for relevant info about C# for this test.

geefr:
Ah thanks, I'll see what I can do testwise here.

(Not actually my bug btw, just picked it up from the list)

geefr:
@wsfulton Okay I tried my best, hope it's not too far from the mark..

After digging around I noticed C# is the only language that has a non-trivial wchar.i. All the others either don't have wchar.i at all, or use the same typemaps as std_wstring.i in some common file.

Did try to make a separate wchar.i test case, but it unravelled quickly, basically turned into a C#-specific test case so though it best to patch into the existing one instead.

Updated the Python 'runme' while I was there, works as expected. Tried to update the Ruby one but couldn't work out the right syntax there.

wsfulton:
Thanks, ruby equivalent added in 71d48228605b0dcf97cc6baf4e4124355e821273.

PR #1887, opened 2020-10-14
Drop deprecated PyEval_CallObject method

jschueller:
see https://docs.python.org/3.9/whatsnew/3.9.html


wsfulton:
Looks okay to me. There is no documentation nor usage within SWIG of `swig_call_defargs`; are you actually using it in somehow?

jschueller:
Indeed, i got it but it turns out it was unrelated to swig, but it doesnt hurt.

PR #1909, opened 2020-12-16
Add and improve Ruby test cases in the context of nesting and namespaces

treitmayr:
This is done in preparation for adding namespace support to the Ruby part of SWIG.
Some existing test cases were reorganized or duplicated for flat/nonflat nesting (e.g. nested_scope). For some a Ruby test script was added.
Finally the ruby/Makefile.in was improved so that for test cases without an explicit test script, the generated wrapper library will just be loaded by the Ruby interpreter to ensure loading works fine.

### DIFF
--- a/Examples/test-suite/ruby/preproc_runme.rb
+++ b/Examples/test-suite/ruby/preproc_runme.rb
@@ -5,23 +5,12 @@
 
 require 'swig_assert'
 
-# helper class for comparing version strings
-class Version
-  attr_reader :array
-  def initialize(s)
-    @array = s.split('.').map { |e| e.to_i }
-  end
-  def <(rhs)
-    a = @array.clone
-    b = rhs.array.clone
-    a << 0 while a.size < b.size
-    b << 0 while b.size < a.size
-    (a <=> b) < 0
-  end
-end
-
+# This extension to the Warning class is intended for suppressing expected
+# Ruby warning messages about invalid or redefined Ruby constants - basically
+# the equivalent of %warnfilter(SWIGWARN_RUBY_WRONG_NAME) but for the moment
+# the wrapper library is loaded by the Ruby interpreter.
+# Note: This only works for Ruby 2.4 and later
 if Object.const_defined?(:Warning) && Warning.respond_to?(:warn)
-  # suppressing warnings like this only works for Ruby 2.4 and later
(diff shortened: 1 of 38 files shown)


treitmayr:
Fixed two test cases for Ruby 1.9.3.

Examples/test-suite/ruby/preproc_runme.rb
```diff
@@ -0,0 +1,51 @@
...
+      if msg =~ /[Aa]lready initialized constant Preproc::A[56]/ ||
+         msg =~ /invalid name .?__GMP_HAVE_/
+        # ignore
+      else
+        super
+      end
+    end
+  end
+  Warning.extend CustomWarningFilter
+end
```
wsfulton, line 37: What is going on in the lines above? I'm just not sure what this is about!

wsfulton (commented):
Thanks @treitmayr, this looks good. Sorry, I couldn't get to the patch earlier, I hope it didn't hold you up with the namespace work. Presumably this is adding support for `%nspace`?

Examples/test-suite/ruby/preproc_runme.rb
treitmayr: The extension to the `Warning` class is intended for suppressing expected Ruby warning messages about invalid or redefined Ruby constants - basically the equivalent of `%warnfilter(SWIGWARN_RUBY_WRONG_NAME)` but for the moment the wrapper library is loaded by the Ruby interpreter. I will add this very description to this section in the test case.

However now that I take a second look at this file I realize that I forgot to remove the `Version` helper class originally used for checking the Ruby version, before replacing it with the much better check for the `Warning` module.

I will commit these two small changes (comment + remove `Version`) in a moment.

treitmayr:
@wsfulton, thank you for checking the PR - see my inline comment and added commit.
You are right, that's the prerequisite to `%nspace` support. I already have a fully working version [here](https://github.com/swig/swig/compare/master...treitmayr:ruby-nspace-support) but it may need some polishing here and there. While my current assignment at work shifted a bit temporarily, I am confident to be able to resume work on it in ~2 months. Side note: My employer Thales Austria sponsors my SWIG contributions.

treitmayr:
Thanks for merging!

PR #1913, opened 2020-12-29
Added windows build instructions for MSYS2

smistad:
As the windows build instructions for swig are almost 15 years old, I propose to add instructions for building the windows binaries using the newer MSYS2 which are simpler than the existing instructions using MSYS(1).

I tested this on a windows 10 64 bit machine using MSYS2 version msys2-x86_64-20201109.


wsfulton:
Thanks I've used this after tweaking slightly in 45c015b6319927fed962967c2a2c7fb816076043.

Issue #1125, opened 2017-10-16
Syntax error in input: static constexpr auto Foo = "bar";


letmaik:
The following leads to a syntax error message from SWIG when including a header file with the following:
```c++
static constexpr auto Foo = "bar";
```
Tested with SWIG 3.0.12.

ojwb:
Looking at `Source/CParse/parser.y` it seems like the only parser rules involving `AUTO` are for functions and declarations initialised by a lambda, so this is a missing C++11 feature.  `storage_class` already allows `static constexpr`.

ojwb:
Copying my remarks over from PR #1930:

I think this should be possible - it's just nobody's implemented it yet.  There are still quite a number of C++11 and newer features SWIG doesn't support (or like `auto` doesn't fully support), and sadly limited developer time.

Probably a useful first step would be to extend the parser to deal with this case, but not try to wrap it.  Parsing `auto` here would be needed as part of fuller support, and would at least mean people could wrap the rest of an interface that has an `auto` variable or constant.

And given we apparently (according to the manual) support:

```
int i;
decltype(i) j;
```

it seems that it should be fairly easy to support:

```
int i;
auto j = i;
```

But apparently `decltype(i+j)` isn't supported, so `auto k = i + j;` would probably be harder.

Unfortunately I have no time to work on any of this - I'm already badly backlogged with SWIG (and other) stuff - but I'll try to review patches if people submit them.

ojwb:
I had a poke at this while I was waiting for something to run, and I think I've basically solved it.  I'll push the fix once I've cleaned it up and added tests and docs.

SWIG should even be able to deduce the type for `auto var = expr;` in the same limited cases that it can for `decltype(expr)`.

ojwb:
I've pushed a branch with my fix and it is currently running through CI testing.

As a bonus, I've extended type deduction to support some expressions involving constant literals, so SWIG can even deduce the type for the case reported here (as `const char*`):

```
static constexpr auto Foo = "bar";
```

PR #1930, opened 2021-01-18
Document lack of support for `auto` for C++ variables (#1125)

SnoopJeDi:
This PR adds a short remark to clarify that SWIG does not support `auto` as a type specifier for variables.

I did my best to capture the remark by @ojwb in #1125 about what _is_ supported, but this could maybe use more contextual information—like why this isn't possible, if it's strictly out of scope—from someone who understands SWIG's innards better than I do.

### DIFF
--- a/Doc/Manual/CPlusPlus11.html
+++ b/Doc/Manual/CPlusPlus11.html
@@ -336,6 +336,10 @@ int i; int j;
 decltype(i+j) k;  // syntax error
 </pre></div>
 
+<p>SWIG does not support <tt>auto</tt> as a type specifier for variables, only
+for specifying the return type of <a href="#CPlusPlus11_lambda_functions_and_expressions">lambdas</a>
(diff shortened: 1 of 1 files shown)


ojwb:
> like why this isn't possible

I think it should be possible - it's just nobody's implemented it yet.  There are still quite a number of C++11 and newer features SWIG doesn't support (or like `auto` doesn't fully support), and sadly limited developer time.

Probably a useful first step would be to extend the parser to deal with this case, but not try to wrap it.  That's needed for fuller support, and would at least mean people could wrap the rest of an interface that has an `auto` variable or constant.

And given we apparently (according to the manual just above where your patch applies) support:

```
int i;
decltype(i) j;
```

it seems that it should be fairly easy to support:

```
int i;
auto j = i;
```

But apparently `decltype(i+j)` isn't supported, so `auto k = i + j;` would probably be harder.

Unfortunately I have no time to work on any of this - I'm already badly backlogged with SWIG (and other) stuff - but I'll try to review patches if people submit them.

Issue #1932, opened 2021-01-21
Java, Windows, MinGW GCC compiler issue


Nashatyrev:
Environment: Windows, MinGW 64-bit GCC compiler, cygwin or msys2 shell.
The original issue: https://github.com/supranational/blst/issues/53

When generating wrapper cpp code we've got the following fragment in our generated wrapper cpp file 
```
/* Fix for jlong on some versions of gcc on Windows */
#if defined(__GNUC__) && !defined(__INTEL_COMPILER)
  typedef long long __int64;
#endif
```
The above code comes from the SWIG [`javahead.swg` include](https://github.com/swig/swig/blob/88896356831a025220b6ebffcb38f0ac08dcddef/Lib/java/javahead.swg#L32-L36)

When trying to compile the generated code with MinGW-64 GCC compiler it fails with: 
```
blst_wrap.cpp:194:21: error: 'long long long' is too long for GCC
  194 |   typedef long long __int64;
```

This is because MinGW GCC has the header included `_mingw.h` which does `#define __int64 long long` which obviously breaks the `typedef` above. Below is the inclusion path: 
```
. ../blst.hpp
.. E:/TDM-GCC-64/lib/gcc/x86_64-w64-mingw32/9.2.0/include/c++/string
... E:/TDM-GCC-64/lib/gcc/x86_64-w64-mingw32/9.2.0/include/c++/bits/char_traits.h
.... E:/TDM-GCC-64/lib/gcc/x86_64-w64-mingw32/9.2.0/include/c++/bits/postypes.h
..... E:/TDM-GCC-64/lib/gcc/x86_64-w64-mingw32/9.2.0/include/c++/cwchar
...... E:/TDM-GCC-64/x86_64-w64-mingw32/include/wchar.h
....... E:/TDM-GCC-64/x86_64-w64-mingw32/include/corecrt.h
........ E:/TDM-GCC-64/x86_64-w64-mingw32/include/_mingw.h
```

May be it makes sense to add `&& !defined(__int64)` to the macro condition? (I'm not a big C specialist though)

Nashatyrev:
One clarification which might matter in this context: we are using TDM-GCC distribution. See [this comment](https://github.com/supranational/blst/issues/53#issuecomment-764966887) for a bit more details

Nashatyrev:
Here is the compiler details: 
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=E:\TDM-GCC-64\bin\gcc.exe
[...]

PR #1934, opened 2021-01-22
Lib/java/javahead.swg: clean up jlong handling.

dot-asm:
As for __int64 definition. __int64 is a non-standard Visual-C-specific
type used in win32/jni_md.h. It is defined by other Win32 compilers in
one way or another, obviously for compatibility. It's more appropriate
to give the compiler a chance to make necessary arrangements instead
of reinventing the wheel. This, giving a chance, can be achieved by
including virtually any standard header. Since jni.h includes stdio.h,
defining __int64 in javahead.swg is redundant. Since doing so actually
triggers compilation errors on MinGW if a system header is included
in the %begin section, it's arguably appropriate to omit it.

As for `#undef _LP64` removal. Undefining a pre-defined macro, which
_LP64 is, is bad style, and if followed by inclusion of systems
headers, it's actually error-prone. Log suggests that it was added
to resolve a warning. I'm inclined to believe that it rather was a
misunderstanding of some kind. Or a bug in warning subsystem of
some particular compiler version, in which case it would have been
more appropriate to advise users to ignore the warning.

### DIFF
--- a/Lib/java/arrays_java.i
+++ b/Lib/java/arrays_java.i
@@ -104,7 +104,7 @@ JAVA_ARRAYS_DECL(int, jint, Int, Int)                 /* int[] */
 JAVA_ARRAYS_DECL(unsigned int, jlong, Long, Uint)     /* unsigned int[] */
 JAVA_ARRAYS_DECL(long, jint, Int, Long)               /* long[] */
 JAVA_ARRAYS_DECL(unsigned long, jlong, Long, Ulong)   /* unsigned long[] */
-JAVA_ARRAYS_DECL(jlong, jlong, Long, Longlong)        /* long long[] */
+JAVA_ARRAYS_DECL(long long, jlong, Long, Longlong)    /* long long[] */
 JAVA_ARRAYS_DECL(float, jfloat, Float, Float)         /* float[] */
 JAVA_ARRAYS_DECL(double, jdouble, Double, Double)     /* double[] */
 
@@ -128,7 +128,7 @@ JAVA_ARRAYS_IMPL(int, jint, Int, Int)                 /* int[] */
 JAVA_ARRAYS_IMPL(unsigned int, jlong, Long, Uint)     /* unsigned int[] */
 JAVA_ARRAYS_IMPL(long, jint, Int, Long)               /* long[] */
 JAVA_ARRAYS_IMPL(unsigned long, jlong, Long, Ulong)   /* unsigned long[] */
-JAVA_ARRAYS_IMPL(jlong, jlong, Long, Longlong)        /* long long[] */
+JAVA_ARRAYS_IMPL(long long, jlong, Long, Longlong)    /* long long[] */
 JAVA_ARRAYS_IMPL(float, jfloat, Float, Float)         /* float[] */
 JAVA_ARRAYS_IMPL(double, jdouble, Double, Double)     /* double[] */
 
--- a/Lib/java/javahead.swg
+++ b/Lib/java/javahead.swg
@@ -30,18 +30,6 @@
 #endif
 
 %insert(runtime) %{
-/* Fix for jlong on some versions of gcc on Windows */
-#if defined(__GNUC__) && !defined(__INTEL_COMPILER)
-  typedef long long __int64;
-#endif
-
-/* Fix for jlong on 64-bit x86 Solaris */
-#if defined(__x86_64)
-# ifdef _LP64
-#   undef _LP64
-# endif
-#endif
-
 #include <jni.h>
 #include <stdlib.h>
 #include <string.h>
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,6 @@
+image:
+- Visual Studio 2019
 platform:
-- x86
 - x64
 
 environment:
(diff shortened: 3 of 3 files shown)


dot-asm:
Additional commit resolves relevant Travis CI failures.

dot-asm:
Re-based. The travis errors are unrelated, two macos job timeouts, in download phase.

wsfulton:
Now that `long long` is widely supported, I'm okay with changes in array_java.i.

I don't understand the motivation for the changes in javahead.swg though. jni.h should not rely on stdlib.h nor string.h to work. I don't know of any documented recommendation for that. Why not leave as is?

wsfulton:
Just to mention, from experience, I'm very wary of messing around with header include orders for something that has worked for twenty odd years unless there is some bug to solve that necessitates it.

dot-asm:
> I don't understand the motivation for the changes in javahead.swg though. jni.h should not rely on stdlib.h nor string.h to work. I don't know of any documented recommendation for that. Why not leave as is?

The motivation is in #1932. The issue is that if one includes **a** system header in `%begin` section, then mingw compilation of the generated wrapper fails. I can work around the problem in my .swg, but **formally** it classifies as swig bug, because it's an inappropriate assumption to make that no header files are included in `%begin` section.

Now, it is possible to pinpoint the specific mingw problem by adjusting `#if`-s, but after having a broader look, I've noted the `_LP64` and figured that it's possible to solve **all** the problems in single stroke by merely rearranging the headers.

As for `_LP64`. As already mentioned, it's a pre-defined macro, just like for example `__x86_64__`. With this in mind, would you condone compilation with `-U__x86_64__` on compiler command line? No. And for the same reason you shouldn't condone undefining `_LP64`. Well, in practice, some[/most?] system headers can "recover" from `-U_LP64` at the command line, but `#undef _LP64` in source is demonstrably error-prone. I seem to fail to find compelling example for Linux, but it's not a problem to find one for Solaris. Take for example
```
#include <stdint.h>
#undef _LP64
#include <stdio.h>
```
If you run `cc -m64 -E a.c | grep size_t\;`, it'll say `typedef unsigned int size_t`. Remove `#undef _LP64` and it will say `typedef unsigned long size_t`...

dot-asm:
> jni.h should not rely on stdlib.h nor string.h to work.

It's a trick question. While it holds true in general, it gets subtle on mingw. Windows jni_md.h relies on `__int64` type, a non-standard type implemented by Visual C as a keyword. "As a keyword" means that you can use `__in64` without including any headers. Now, mingw compiler doesn't implement `__int64` as a keyword, but instead does it in a header. Or in other words, on mingw, jni.h is in fact dependent on **a** header. And it just so happens that stdlib.h or string.h do the job (and apparently did since early 2000).

dot-asm:
> on mingw, jni.h is in fact dependent on **a** header

Actually it's not exactly correct assertion. It's win32/jni_md.h that is 100% dependent on **a** header on mingw, but it doesn't necessarily mean that jni.h is. Because jni.h can include relevant header prior it includes jni_md.h. And it kind of does. It does include stdio.h, and it is sufficient for `__int` to be defined on mingw. This means that one should be able to simply omit all that `#if-this-n-that` jazz without rearranging headers... I can think of possibility that some **earlier** jni.h didn't include anything prior jni_md.h...

dot-asm:
I've harmonized commits with last remark.

dot-asm:
> mingw compilation of the generated wrapper fails

Just in case for reference. `__int64` is defined as `long long` **macro** by mingw. As result attempt to `typedef long long __int64` is expanded as `typedef long long long long`, which obviously fails to compile.

wsfulton:
Okay, I'm happy to accept this change. I'd like to get our mingw testing back up and running on Appveyor first so that it is tested on that platform. I had to disable mingw testing recently in 6358510c57d772ab2ece053109a4c09951b7e036.

dot-asm:
I've re-based and added a throw-away commit to test if mingw is still dead on appveyor...

dot-asm:
mingw works with at least latest image in appveyor.yml, e.g.
```
image:
- Visual Studio 2019
```
As it can be seen it's possible to fire up the Java test on mingw, but it fails one run test, director_thread. Since it's just one among large amount, it's unlikely to be caused by suggested modification:-) At least there were no compilation errors... Should I throw away "throw-away" commits and add the `image` line and uncomment mingw lines commented in 6358510?

dot-asm:
> but it fails one run test, director_thread.

Attempt to reproduce the failure on my VM has failed. Though at the same time it failed few **other** tests, with "swig error : Unrecognized option _wrap.cxx"[!] and with a lot of errors from javac in connection to swig failure, failures to find symbols, class names mostly. Since Windows is not my platform of choice, I'm reluctant to proceed with identifying these problems, as arguably unrelated to the problem in question. I can confirm that it does help the originating #1932. And it was  also [shown](https://ci.appveyor.com/project/swig/swig/builds/38250049) that it doesn't trigger any C/C++ compilation failures or warnings.

wsfulton:
I keep putting off the upgrade to the Visual Studio 2019 images on Appveyor as they keep changing. I guess I better tackle it for mingw. director_thread test failure, oh no! That's going to be horrible to track down.

dot-asm:
I tried 2017 image, mingw doesn't work there. I've tried to specify image on per-entry basis, and it didn't work. Just in case "mingw doesn't work" means that older packages are no longer available in public repositories, which is why there are 404-s all over. I've tried vanilla master build in #1976 and it succeeded. I've re-ran test on this commit and it failed in director_thread again. I had a look at wrapper and can't see how suggested changes would trigger a problem. And since I can't reproduce it on my VM, I'm letting this thing with mingw on appveyor go and won't try anything further.

wsfulton:
Thanks, committed in 3163888c4bbaa7f165e02846ea2983a5f53b4caa.

dot-asm:
Thanks! On tangential note, jdk 15 complains about `sealed` becoming a keyword... Oh! You seem to be already on it:-)

PR #1938, opened 2021-01-30
Fix C++ director wrapper return values not being initialized in C# module

GenuineAster:
Sometimes when returning by value, the C++ wrapper around C# directors didn't initialize return values, causing a compiler warning if they are enabled (and in our case, a compiler error). This change simply adds a SwigValueInit there.


wsfulton:
Can you please provide a testcase for this so we can use it for future regression testing and to check the pull request. Thanks.

PR #1939, opened 2021-01-30
Check for null when using reflection in C# directors

GenuineAster:
this.GetType().GetMethod(...) can return null, which generates a warning when the result is later used. This change adds a null-check around the usage of that result, and should be backwards-compatible with previous C# version too.


wsfulton:
Can you a provide a use case for this scenario please?

GenuineAster:
This just happens every time you use the Swig directors feature in C# wrapper code, in my case when we have nullable enabled but I'm not sure if that is relevant to the issue.

ojwb:
There are no such warnings from running SWIG's testsuite which generates, builds and runs numerous testcases which use directors:

https://github.com/swig/swig/runs/5503414924?check_suite_focus=true

@GenuineAster Can you provide a simple self-contained example demonstrating this warning?

ojwb:
I see this has been auto-closed by the fork being deleted.

A reproducer was never provided, but the code this proposes to change was replaced by #1323 and the new code looks like it has `null` checks, so I think this was addressed in 4.0.0 (and if it wasn't, we can't now apply the proposed patch here anyway).  If anyone can reproduce this, please provide a reproducer and we can reopen.

PR #1944, opened 2021-02-03
Update CSharp.html

DaveInCaz:
global::System seems to be needed inside the typemap examples and also in the examples of generated code. But in these *example* C# programs, that would not be typical usage because the System namespace is almost always imported and then used implicitly. 

Also, it was already not used in the first of the Console... calls in the example. So this is more consistent.

### DIFF
--- a/Doc/Manual/CSharp.html
+++ b/Doc/Manual/CSharp.html
@@ -1857,9 +1857,9 @@ and the following usage from C# after running the code through SWIG:
   Wheel wheel = new Bike(10).getWheel();
   Console.WriteLine("wheel size: " + wheel.size);
   // Simulate a garbage collection
-  global::System.GC.Collect();
-  global::System.GC.WaitForPendingFinalizers();
-  global::System.Console.WriteLine("wheel size: " + wheel.size);
+  GC.Collect();
+  GC.WaitForPendingFinalizers();
+  Console.WriteLine("wheel size: " + wheel.size);
 </pre>
 </div>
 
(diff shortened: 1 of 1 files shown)


wsfulton:
Since SWIG 2, the SWIG typemaps do not assume System namespace has been imported. Hence all typemaps and examples do not assume this. The documentation should not assume System has been imported either for consistency and more robust examples. Hence this change to the documentation is incorrect. As the typemaps are for an example, you are of course welcome to modify the typemaps for your own usage. I think, if anything, we should be adding some missing `global::` namespace to the documentation for more reliable compilation given the current SWIG defaults.

DaveInCaz:
@wsfulton thanks for reviewing this.

It looked like these specific cases in the documentation were meant to be examples of code use _outside_ of SWIG templates.  Example code showing how you might consume SWIG-generated objects.

(There were numerous other places where global::system... was used that I obviously left untouched because they were inside SWIG templates.)

Actually, the `Console.WriteLine` examples were inconsistent (which is what I noticed in the first place). So I thought it was a good idea to make them consistent and do that in the form that typical C# programs would use.

But maybe that was a misunderstanding or misreading of the docs on my part. If so, they still should be made consistent though, should I re-submit to fix the `Console.WriteLine` examples by adding global::System where it is missing?

Dave

ojwb:
> should I re-submit to fix the `Console.WriteLine` examples by adding global::System where it is missing?

Please do.

ojwb:
> should I re-submit to fix the `Console.WriteLine` examples by adding global::System where it is missing?

@DaveInCaz Please do.

(Repeating reply but with a mention)

PR #1945, opened 2021-02-09
Fix typos in attribute2ref() in Lib/typemaps/attribute.swg

### LINKED ISSUE #1872: Possible bug in attribute.swg/attribute2ref
Hi

I'm using the function "attribute2ref" contained inside the file Libs/typemaps/attribute.swg, and I'm calling it providing only the first three arguments (namely, Class, AttributeType and AttributeName).

I got a build error  'no member named ‘AccessorName’

Looking at the code 

```
%define %attribute2ref(Class, AttributeType, AttributeName, AccessorMethod...)
  #if #AccessorMethod != ""
    %attribute_custom(%arg(Class), %arg(AttributeType), AttributeName, AccessorMethod, AccessorMethod, &self_->AccessorMethod(), self_->AccessorMethod() = *val_)
  #else
    %attribute_custom(%arg(Class), %arg(AttributeType), AccessorName, AccessorName, AccessorName, &self_->AccessorName(), self_->AccessorName() = *val_)
  #endif
%enddef

```

It seems that in the call  %attribute_custom after the #else statement "AccessorName" should be indeed replaced with AttributeName.

Is this the case?

Thank you very much

ZackerySpytz:
AccessorName was being used instead of AttributeName.

Closes #1872.

### DIFF
--- a/Examples/test-suite/li_attribute.i
+++ b/Examples/test-suite/li_attribute.i
@@ -95,6 +95,7 @@ struct MyFoo; // %attribute2 does not work with templates
 
 // class/struct attribute with get/set methods using return/pass by reference
 %attribute2(MyClass, MyFoo, Foo, GetFoo, SetFoo);
+%attribute2ref(MyClass, MyFoo, Foo2);
 %inline %{
   struct MyFoo { 
     MyFoo() : x(-1) {}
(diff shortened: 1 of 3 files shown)


wsfulton:
Thanks @ZackerySpytz

Issue #1949, opened 2021-02-21
Invalid warnings and memory error on empty Doxygen strings


geographika:
The following is a comment in RST on a C struct. The word immutable should be in bold when processed in Sphinx so is wrapped in double asterisks. The \ is required to escape to ensure correct output. 

```c++
    int width; ///< \**immutable** image width in pixels
```

When run with the following command parameters:

`swig -python -py3 -doxygen -shadow -o test_wrap.c test.i`

the following line throws an unhandled memory exception when debugging in Visual Studio as the string is empty:

https://github.com/swig/swig/blob/ac8de714af1d7fbd755597b0f51792212f552ef7/Source/Doxygen/doxyparser.cxx#L1089

`Unhandled exception at 0x00007FF84489B7EC (ucrtbased.dll) in swig.exe: An invalid parameter was passed to a function that considers invalid parameters fatal.`

This is thrown in the `xsstring` module with the error `cannot dereference string iterator because it is out of range (e.g. an end iterator)`

The empty string is created here (`endOfWordPos` and `pos` are both 2):

https://github.com/swig/swig/blob/ac8de714af1d7fbd755597b0f51792212f552ef7/Source/Doxygen/doxyparser.cxx#L1181

In addition to the memory error, erroneous errors are output by SWIG to the command line:

```
..\swiginc\class.i(164) : Warning 560: Unknown Doxygen command: .
..\swiginc\class.i(179) : Warning 560: Unknown Doxygen command: .
..\swiginc\class.i(186) : Warning 560: Unknown Doxygen command: .
..\swiginc\class.i(191) : Warning 560: Unknown Doxygen command: .
```

vadz:
I'm not sure why is backslash required, but in any case this shouldn't crash, of course. But I can't reproduce the problem with just
```diff
diff --git a/Examples/test-suite/doxygen_parsing.i b/Examples/test-suite/doxygen_parsing.i
index 40f37a4c2..864fdb884 100644
--- a/Examples/test-suite/doxygen_parsing.i
+++ b/Examples/test-suite/doxygen_parsing.i
@@ -2,6 +2,8 @@

 %inline %{

+int width; ///< \**immutable** image width in pixels
+
 /**
  * The class comment
  */
```
Could you please make a minimal patch to this test case showing it? TIA!

More generally, we really need to run at least some SWIG CI builds using ASAN...

geographika:
@vadz thanks for the feedback. I'll update the test case in the pull request. 
[...]

PR #1950, opened 2021-02-21
Avoid Doxygen string memory errors

geographika:
Fix/workaround for #1949

This adds a guard for empty strings to avoid errors when calling `cmd.begin()`.


vadz:
Thanks!

This is certainly better than crashing, so it should be merged IMO, but I'm not sure if it's normal for `addDoxyCommand()` to be called with an empty command in the first place. If it is, it should probably be mentioned in its comment.

vadz:
Closing in favour of #1952 as discussed in #1949.

Issue #1561, opened 2019-06-20
Option to disable function annotations when -py3 option is used


gieseanw:
At the moment, the `-py3` option unconditionally enables function annotations. I'd like to be able to turn on Python 3, but turn off function annotations. In my workflow, I'm generating stub files automatically via another process. These type hints conflict with the annotations, and the annotations are taking precedence.

Looking at source/Modules/python.cxx, the function `make_pyParmList` sets a boolean 

    bool funcanno = py3 ? true : false;

And this is passed to `make_autodocParmList`

It seems like perhaps another python command line option could be provided that will turn them off. Perhaps it should be handled by the "autodoc" directive in some way?

For now, I'm relegated to post-processing the generated .py files to remove the annotations.

wsfulton:
I actually meant to have function annotations turned off by default when using -py3 for the swig-4.0.0 release and am frustrated now that I didn't get round to it. I suggest a new command line option, `-noannotations` which will suppress the function annotations from being generated when using -py3. In swig-4.1, we can deprecate this option if we hopefully have a proper type hint solution, please see #735.

wsfulton:
@gieseanw, why don't you just drop the `-py3` option? It hardly changes anything else, or is there something from this option that you need?

Issue #735, opened 2016-06-29
[Python] Add support for PEP 484 type hints


rossng:
Python 3.5 and later have support for [type hints/annotations](https://docs.python.org/3/library/typing.html), which help out in an IDE and are very useful for writing reliable code.

It would be great if SWIG could optionally generate type annotations for Python. Even better, support could be added for more complex constructs such as iterables (e.g. `std::vector`).

Beliaar:
Well, there is the -py3 option that adds annotations to the proxy classes.

But they are very useless for IDEs because they use the C/C++ type, and not the actual python type.

dineshgpatel:
IDE like VS code only support auto-completion based on PEP 484 type hints. 
@wsfulton  Is there any plan to support Python3 type hinting in SWIG release 4.0.0 or earlier?

wsfulton:
These type annotations would be a great addition to SWIG. SWIG is an all volunteer project, but I don't know of anyone working on this. I encourage anyone interested in adding this support to discuss it here.

phwuil:
Hi,
As long as there is no such feature, is it possible to avoid the generation of the *non-PEP 484-compatible* annotations produced by `-py3` ? 

For instance, when producing sphinx documentation with the `sphinx_autodoc_typehints` extension, the actual annotations produce error such as :

```
Exception occurred:
  File "/usr/lib64/python3.7/typing.py", line 448, in __init__
    raise SyntaxError(f"Forward reference must be an expression -- got {arg!r}")
  File "<string>", line None
SyntaxError: Forward reference must be an expression -- got 'std::string'
```
(the argument has been type-annotated with `strd::string`)

wsfulton:
Currently the `-py3` option adds in Python3 function annotations that use the C++ type names. The ["autodoc"](http://swig.org/Doc3.0/Python.html#Python_nn67) feature uses the Python type names mostly, but sometimes the C++ type when documenting types. If my understanding of PEP-484 is correct, a different set of types are required (those in the typing module). 

We need a consistent and flexible approach for the 3 different sort of types across docstrings, doxygen sourced comments and Python3 function annotations. Looks like these are the 3 types possible:
1. Pure Python types
2. The C++ types
3. The types from the typing module required fro PEP-484

[...]

Issue #1779, opened 2020-04-20
Move python default version to 3


tpaviot:
After the latest swig version (4.0.1) was released (in August 2019), python2 support was officially dropped down in January 2020. That would make sense that swig keeps consistent with this decision, and move its default python version to 3.

Currently, in order to generate a python wrapper using swig, you have to pass both the ```-python``` flag, as well as ```-py3``` to explicitly request for python 3 specific features. The compiled module ca be used with python 3 only. If only ```-python``` is passed to swig, then the generated python code syntax is python 2. Today, this should be the opposite: the default python code should be 3 if nothing is passed but the ```-python``` flag, and python 2 code should be explicitly requested.

The following suggestion would allow this with only a few changes to the current codebase:

* add another ```-py2``` flag

* make ```-py2``` and ```-py3``` mutually exclusive

* set default py3 value to true by default, to false if ever ```-py2``` is passed

* optional: make either ```-py2``` or ```-py3``` mandatory to prevent swig users from an unexpected API change to the python code when moving from 4.0.1 to any later version.

ojwb:
> Currently, in order to generate a python wrapper using swig, you have to pass both the -python flag, as well as -py3 to explicitly request for a python 3 compliant module. If only -python is passed, then the generated python code is python 2.

That's wrong - see the manual:

>33.12 Python 3 Support
>
>SWIG is able to support Python 3.x. The wrapper code generated by SWIG can be compiled with both Python 2.x or 3.x. Further more, by passing the `-py3` command line option to SWIG, wrapper code with some Python 3 specific features can be generated (see below subsections for details of these features).

tpaviot:
@ojwb you're right, what I wrote is actually wrong I went to fast.

tpaviot:
I made a small change to the issue description, I guess it's more accurate and conform to the documentation.

wsfulton:
Related to #1561. `-py3` does very little apart from add in these function annotations, which are really not very good and should be off by default and only explicitly added in if required. Something to think about for swig-4.1.0.

ojwb:
Something I'd definitely be keen to do here is eliminate the need to run the testsuite with:

```
make check-python-test-suite PY3=y
```

(Or `PY3=1` depending which docs you read - what actually seems to matter is if it's empty or not.)

[...]

Issue #1531, opened 2019-05-06
SWIG 4.0: Python: Lamba try to capture an array element


Mizux:
Using `SWIG 3.0.12`, swig generate one local variable per `PyObject` pointer (ed `obj[0-3]`)
constraint_solver_python_wrap.cc:
```cpp
...
SWIGINTERN PyObject *_wrap_Solver_Decision(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  PyObject *resultobj = 0;
  operations_research::Solver *arg1 = (operations_research::Solver *) 0 ;
  SwigValueWrapper< std::function< void (operations_research::Solver *) > > arg2 ;
  SwigValueWrapper< std::function< void (operations_research::Solver *) > > arg3 ;
  void *argp1 = 0 ;
  int res1 = 0 ;
  PyObject * obj0 = 0 ;
  PyObject * obj1 = 0 ;
  PyObject * obj2 = 0 ;
 ...
 arg1 = reinterpret_cast< operations_research::Solver * >(argp1);
  {
    arg2 = [obj1](operations_research::Solver* s) {
      return PyFunctionSolverToVoid(obj1, s);
    };
  }
```

now, using SWIG 4.0, swig seems to use a `PyObject` pointer array (ed `swig_obj[3]`)
```cpp
...
SWIGINTERN PyObject *_wrap_Solver_Decision(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  PyObject *resultobj = 0;
  operations_research::Solver *arg1 = (operations_research::Solver *) 0 ;
  SwigValueWrapper< std::function< void (operations_research::Solver *) > > arg2 ;
  SwigValueWrapper< std::function< void (operations_research::Solver *) > > arg3 ;
  void *argp1 = 0 ;
  int res1 = 0 ;
  PyObject *swig_obj[3] ;
...
 arg1 = reinterpret_cast< operations_research::Solver * >(argp1);
  {
    arg2 = [swig_obj[1]](operations_research::Solver* s) {
      return PyFunctionSolverToVoid(swig_obj[1], s);
    };
  }
```

But gcc/clang forbid this capture usage
see: https://stackoverflow.com/questions/44485030/lambda-capture-an-array-element-failed

Is there a way to force SWIG to **NOT** use an array ?
note: I'm compiling using C++11 standard and on Linux:
```sh
$ g++ --version
g++ (Debian 7.3.0-15) 7.3.0
```
[...]

PR #1951, opened 2021-02-21
WIP: Add Python property annotations

geographika:
Currently the `-py3` option adds annotations to functions in the Python module. 

You can annotate the getters and setters of a property, however when these are on the DLL/SO it is not possible.
However as per https://stackoverflow.com/a/42463560 there is an alternative syntax of:

`bar: Bar = property(lambda self: Bar(self))`

This pull request adds in the property type using the same approach as for functions e.g.

`numstyles: "int" = property(_mapscript.classObj_numstyles_get, doc=r"""number of styles for class""")`

Note if #1561 is to be implemented then these property annotations should also be disabled. 
Note also the C-types are used as discussed in #735 - and are therefore wrapped in quotes to avoid syntax errors. Ideally the mapped Python types would eventually be used. 

This change solves #1914 - Sphinx/autodoc can use the property annotations to generated API documentation including property types. Sample output image below:

image

If this seems like a useful addition please let me know if I should continue investigating how to add/update tests.


geographika:
The `-py3` flag to add these will mean Python2 output is unaffected. However the property annotation syntax used in this pull request is only available from Python 3.6 onwards (as caught by the failing Travis tests. 

To do:

- disable the property annotations for pre py3.6
- add a test for py3.6 onwards to check the annotations are output

Feedback on if this is a wanted/unwanted feature welcome. Adding an `-annotations` flag to add both property and function type hints rather than the `-py3` flag might be more appropriate?

As the property annotations work almost identically to the function annotations (mainly C-types in quotes) I think either outputting none or both makes sense.

ojwb:
> Feedback on if this is a wanted/unwanted feature welcome.

It seems useful (AIUI the current annotations use the C/C++ type which limits their usefulness, but people are looking at addressing that).

> Adding an -annotations flag to add both property and function type hints rather than the -py3 flag might be more appropriate?

Yes, annotations really should have their own option (see my proposal re `-py3`: https://github.com/swig/swig/issues/1779#issuecomment-1029586192)

> As the property annotations work almost identically to the function annotations (mainly C-types in quotes) I think either outputting none or both makes sense.

I'm not heavily into Python, but that sounds a plausible plan except that from what you say below property annotations don't work with Python < 3.6, and we wouldn't want to force people already generating annotations to choose to between dropping support for Python < 3.6 or dropping annotations completely.

> On further reading of the source code I can't see anywhere that newer Python syntax is used for a point release of Python e.g. something that works in 3.6+ but not earlier, apart from py2/py3, so it looks like if this pull request was added it would need to be enabled by a new command line option.

Is there a trick to generate something that Python < 3.6 can parse but doesn't see as an annotation for the property annotations?

Generally SWIG aims to generate code that works across at least a whole major version rather than require a version of the target language to be specified when swig is run.

If that's not possible, I guess we could have an option for property annotations that generates Python code with a >= 3.6 requirement.

geographika:
Thanks for the feedback @ojwb. 

> It seems useful (AIUI the current annotations use the C/C++ type which limits their usefulness, but people are looking at addressing that).

It would be much nicer to have the Python types, similar to the fine work you did on PHP in #2148. The C/C++ types are still useful for docs generation (rather than IDEs), as the C types are generic enough to be understood across all language bindings. 

Are people actively looking at implementing Python types? Is there a branch/pull request to follow?

> I'm not heavily into Python, but that sounds a plausible plan except that from what you say below property annotations don't work with Python < 3.6, and we wouldn't want to force people already generating annotations to choose to between dropping support for Python < 3.6 or dropping annotations completely.

It sounds like a separate commandline option would be preferable. It does lead to a growth in command options, but it makes the output more flexible, and if good defaults are used then keeps it simple for the common use cases. 

> Is there a trick to generate something that Python < 3.6 can parse but doesn't see as an annotation for the property annotations?

The syntax will throw errors for earlier versions, but there might be a way of dynamically checking for the running Python versions in the output and going something with inspection. It could make for some messy/magic code output though. Would this be preferable to only supporting the feature on Python >= 3.6?

ojwb:
> Are people actively looking at implementing Python types? Is there a branch/pull request to follow?

I don't think so (to either).

> > Is there a trick to generate something that Python < 3.6 can parse but doesn't see as an annotation for the property annotations?
> 
> The syntax will throw errors for earlier versions, but there might be a way of dynamically checking for the running Python versions in the output and going something with inspection. It could make for some messy/magic code output though. Would this be preferable to only supporting the feature on Python >= 3.6?

I think at this point an option to turn it on which gives code which only works with Python >= 3.6 is reasonable - it'll work with all non-EOL Python versions after all, which means it should be usable by a lot of users (and more as time goes on).

wsfulton:
One technical correction is that the annotations support added in this pull request is not termed property annotations, rather variable annotations as described in [PEP 526](https://www.python.org/dev/peps/pep-526/).
 
I've morphed this pull request into something that fits in with the evolving plans outlined in #1531. See commits bf382f01b422c0a67b858e4886562cc09cbba648...bc68a9b034e28788300210c01d6d2d96f0594467 and 3984321c440327701161ddfc03de865f0e9c526a...3159de3e9fcd8b7bf8747711fe240a36d620075f. The latter contains contains your contribution. It should address the concerns regarding Python >= 3.6 as they are easily turned off.

Please follow #735 for future annotations improvements which will include the variable annotations contributed here.

bonqra:
Note that this needs to be disabled for PEP 484 type hints, see [pyright](https://pyright-playground.decorator-factory.su/?gist_id=c3740407fda4b9ba22d945efef2e4706&filename=main.py) and [mypy](https://mypy-play.net/?mypy=latest&python=3.10&gist=c3740407fda4b9ba22d945efef2e4706).

geographika:
Many thanks @wsfulton ! I'll be following #735 and will test the latest master branch on the MapServer SWIG bindings.

PR #1953, opened 2021-02-22
Document CMake build steps on Windows

geographika:
Update docs to include CMake build steps on Windows using Visual Studio, relating to #1040
Temporary output HTML to show formatting at: http://www.geographika.net/swig/Windows.html#Windows_cmake
Comments, suggestions, changes welcome.

### DIFF
--- a/Doc/Manual/Windows.html
+++ b/Doc/Manual/Windows.html
@@ -356,17 +356,6 @@ Note that the Cygwin environment will also allow one to regenerate the autotool 
 These files are generated using the <tt>autogen.sh</tt> script and will only need regenerating in circumstances such as changing the build system.
 </p>
 
-<H4><a name="Windows_cygwin">3.3.1.2 Building swig.exe using Cygwin</a></H4>
-
-
-<p>
-Note that SWIG can also be built using Cygwin.
-However, SWIG will then require the Cygwin DLL when executing. 
-Follow the Unix instructions in the README file in the SWIG root directory.
-Note that the Cygwin environment will also allow one to regenerate the autotool generated files which are supplied with the release distribution. 
-These files are generated using the <tt>autogen.sh</tt> script and will only need regenerating in circumstances such as changing the build system.
-</p>
-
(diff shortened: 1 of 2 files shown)


Doc/Manual/Windows.html
```diff
@@ -364,6 +365,89 @@ <H4><a name="Windows_building_alternatives">3.3.1.3 Building swig.exe alternativ
...
+
+<H4><a name="Windows_building_alternatives">3.3.1.3 Building swig.exe alternatives</a></H4>
+
+
+<p>
+If you don't want to install Cygwin or MinGW, use a different compiler to build 
+SWIG. For example, all the source code files can be added to a Visual C++ project 
+file in order to build swig.exe from the Visual C++ IDE.
+</p>
+
```
wsfulton, line 387: Duplicate of above, should be deleted.

wsfulton (commented):
If nuget is suggested to install pcre, shouldn't the recommendation be to use it to install bison too?

I'd expect chocolatey to be the easiest way to install the two dependencies. If you are able to test that and it works, it would be great to have these two simple ways documented for installing the dependencies, namely, chocolatey and nuget.

Doc/Manual/Windows.html
geographika: Sorry about that - removed

geographika:
> If nuget is suggested to install pcre, shouldn't the recommendation be to use it to install bison too?

Yes that would make sense - I've updated (and tested) so nuget is used for bison, CMake, and pcre. 
 
> I'd expect chocolatey to be the easiest way to install the two dependencies. If you are able to test that and it works, it would be great to have these two simple ways documented for installing the dependencies, namely, chocolatey and nuget.

I couldn't find PCRE on Chocolatey - https://chocolatey.org/packages?q=pcre I tried searches for Perl etc. and nothing relevant came up.

wsfulton:
Looks good, thanks. I'll merge when #1040 is complete.

PR #1956, opened 2021-02-27
Lib/javascript/v8: omit pre-6 node.js pre-processor conditions.

dot-asm:
Node.js v6.0 defines V8_MAJOR_VERSION as 5, hence everything V8_MAJOR_VERSION<5 can be omitted...

This should be considered as initial step for a further clean-ups and amendments. Latest post-#1746 modifications broke pre-12 builds, so that is fixed here, as I plan to dig further into how to handle the warnings...

### DIFF
--- a/Lib/javascript/v8/javascriptcode.swg
+++ b/Lib/javascript/v8/javascriptcode.swg
@@ -445,7 +445,7 @@ fail:
 #if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
   $jsparent_obj->Set(SWIGV8_SYMBOL_NEW("$jsname"), $jsmangledname_obj);
 #else
-  $jsparent_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("$jsname"), $jsmangledname_obj).Check();
+  $jsparent_obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW("$jsname"), $jsmangledname_obj);
 #endif
 
 %}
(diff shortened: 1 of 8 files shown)


dot-asm:
v12 build fails with deprecated interface warning treated as error, and pure v8 does a face-plant. Looking into it...

dot-asm:
Oh! Pure v8 build makes the face-plant, because Ubuntu libv8 package is "stuck" on 3.14.5. This means that one can't remove all those `#if V8_MAJOR_VERSION<4`. This means that pre-6 node would keep working:-) Of course one can still disclaim support for pre-6, but it's likely to keep working as long as one supports libv8 on a popular Linux distro. I'll revert last commit and concentrate on the deprecated stuff (and other warnings)...

nullr0ute:
> Oh! Pure v8 build makes the face-plant, because Ubuntu libv8 package is "stuck" on 3.14.5. This means that one can't remove all those `#if V8_MAJOR_VERSION<4`. This means that pre-6 node would keep working:-) Of course one can still disclaim support for pre-6, but it's likely to keep working as long as one supports libv8 on a popular Linux distro. I'll revert last commit and concentrate on the deprecated stuff (and other warnings)...

Is it getting security updates? Sure newer versions aren't shipped in ubuntu under a different package name? Fedora ships v8 8.4.371.19

dot-asm:
> Is it getting security updates?

It's still the one available in Ubuntu 20, so one has to assume it's considered supported. It might be kept simply because of some arcane dependency...

> Sure newer versions aren't shipped in ubuntu under a different package name?

I couldn't find one. There is libjavascriptcoregtk-4.0 (don't know what kind it is), and of course node.js...

> Fedora ships v8 8.4.371.19

Lucky you:-) Ubuntu appears to be the only supported by Travis...

dot-asm:
Superseded by later commits.

PR #1957, opened 2021-02-28
Lib/javascript/v8: use context-aware initialization.

dot-asm:
Context-aware initialization allows to instantiate add-ons multiple times, most importantly in multiple workers' contexts. Workers made first appearance in v10.5. Condition for performing context-aware initialization, NODE_MODULE_VERSION >= 0x0040, a.k.a. v10+, is just a conservative choice. It was an option for longer than that and formally one can go all the way to supported minimum, v6.x, which is denoted by NODE_MODULE_VERSION 0x0030...

2nd and 3rd commits are not directly related, just minor cleanups.

### DIFF
--- a/Lib/javascript/v8/javascripthelpers.swg
+++ b/Lib/javascript/v8/javascripthelpers.swg
@@ -61,11 +61,11 @@ SWIGRUNTIME void SWIGV8_AddMemberVariable(SWIGV8_FUNCTION_TEMPLATE class_templ, 
  * Registers a class method with given name for a given object.
  */
 SWIGRUNTIME void SWIGV8_AddStaticFunction(SWIGV8_OBJECT obj, const char* symbol,
-  const SwigV8FunctionCallback& _func) {
-#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903) || (SWIG_V8_VERSION < 0x0704)
+  const SwigV8FunctionCallback& _func, v8::Local<v8::Context> context) {
+#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903)
   obj->Set(SWIGV8_SYMBOL_NEW(symbol), SWIGV8_FUNCTEMPLATE_NEW(_func)->GetFunction());
 #else
-  SWIGV8_MAYBE_CHECK(obj->Set(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW(symbol), SWIGV8_FUNCTEMPLATE_NEW(_func)->GetFunction(SWIGV8_CURRENT_CONTEXT()).ToLocalChecked()));
+  SWIGV8_MAYBE_CHECK(obj->Set(context, SWIGV8_SYMBOL_NEW(symbol), SWIGV8_FUNCTEMPLATE_NEW(_func)->GetFunction(context).ToLocalChecked()));
 #endif
 }
 
@@ -73,11 +73,12 @@ SWIGRUNTIME void SWIGV8_AddStaticFunction(SWIGV8_OBJECT obj, const char* symbol,
  * Registers a class method with given name for a given object.
  */
 SWIGRUNTIME void SWIGV8_AddStaticVariable(SWIGV8_OBJECT obj, const char* symbol,
-  SwigV8AccessorGetterCallback getter, SwigV8AccessorSetterCallback setter) {
+  SwigV8AccessorGetterCallback getter, SwigV8AccessorSetterCallback setter,
+  v8::Local<v8::Context> context) {
 #if (V8_MAJOR_VERSION-0) < 5
   obj->SetAccessor(SWIGV8_SYMBOL_NEW(symbol), getter, setter);
 #else
-  SWIGV8_MAYBE_CHECK(obj->SetAccessor(SWIGV8_CURRENT_CONTEXT(), SWIGV8_SYMBOL_NEW(symbol), getter, setter));
+  SWIGV8_MAYBE_CHECK(obj->SetAccessor(context, SWIGV8_SYMBOL_NEW(symbol), getter, setter));
 #endif
 }
 
--- a/Lib/javascript/v8/javascriptinit.swg
+++ b/Lib/javascript/v8/javascriptinit.swg
@@ -5,27 +5,27 @@
 %insert(init) %{
 
 SWIGRUNTIME void
-SWIG_V8_SetModule(void *, swig_module_info *swig_module) {
-  v8::Local<v8::Object> global_obj = SWIGV8_CURRENT_CONTEXT()->Global();
+SWIG_V8_SetModule(v8::Local<v8::Context> context, swig_module_info *swig_module) {
+  v8::Local<v8::Object> global_obj = context->Global();
   v8::Local<v8::External> mod = SWIGV8_EXTERNAL_NEW(swig_module);
   assert(!mod.IsEmpty());
 #if (V8_MAJOR_VERSION-0) < 5
   global_obj->SetHiddenValue(SWIGV8_STRING_NEW("swig_module_info_data"), mod);
 #else
   v8::Local<v8::Private> privateKey = v8::Private::ForApi(v8::Isolate::GetCurrent(), SWIGV8_STRING_NEW("swig_module_info_data"));
-  global_obj->SetPrivate(SWIGV8_CURRENT_CONTEXT(), privateKey, mod);
+  global_obj->SetPrivate(context, privateKey, mod);
 #endif
 }
 
 SWIGRUNTIME swig_module_info *
-SWIG_V8_GetModule(void *) {
-  v8::Local<v8::Object> global_obj = SWIGV8_CURRENT_CONTEXT()->Global();
+SWIG_V8_GetModule(v8::Local<v8::Context> context) {
+  v8::Local<v8::Object> global_obj = context->Global();
 #if (V8_MAJOR_VERSION-0) < 5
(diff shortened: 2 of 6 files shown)


dot-asm:
Travis failure is irrelevant, a Ruby job timeout. Merge conflict resolved.

Lib/javascript/v8/javascripthelpers.swg
```diff
@@ -61,23 +61,24 @@ SWIGRUNTIME void SWIGV8_AddMemberVariable(SWIGV8_FUNCTION_TEMPLATE class_templ,
  * Registers a class method with given name for a given object.
  */
 SWIGRUNTIME void SWIGV8_AddStaticFunction(SWIGV8_OBJECT obj, const char* symbol,
-  const SwigV8FunctionCallback& _func) {
-#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031903) || (SWIG_V8_VERSION < 0x0704)
+  const SwigV8FunctionCallback& _func, v8::Local<v8::Context> context) {
```
chenhengqi, line 64: IMHO, `context` should be passed as reference istead of value.

Lib/javascript/v8/javascripthelpers.swg
```diff
@@ -61,23 +61,24 @@ SWIGRUNTIME void SWIGV8_AddMemberVariable(SWIGV8_FUNCTION_TEMPLATE class_templ,
...
 #endif
 }
 
 /**
  * Registers a class method with given name for a given object.
  */
 SWIGRUNTIME void SWIGV8_AddStaticVariable(SWIGV8_OBJECT obj, const char* symbol,
-  SwigV8AccessorGetterCallback getter, SwigV8AccessorSetterCallback setter) {
+  SwigV8AccessorGetterCallback getter, SwigV8AccessorSetterCallback setter,
+  v8::Local<v8::Context> context) {
```
chenhengqi, line 77: And here.

Lib/javascript/v8/javascripthelpers.swg
dot-asm: v8::Local<T> is a pointer placeholder. In other words it's not v8::Context that is passed by value, but pointer/reference to v8::Context.

Lib/javascript/v8/javascriptinit.swg
```diff
@@ -69,11 +69,17 @@ SWIG_V8_GetModule(void *) {
...
-void SWIGV8_INIT (SWIGV8_OBJECT exports, SWIGV8_OBJECT /*module*/)
+void SWIGV8_INIT (SWIGV8_OBJECT exports, SWIGV8_VALUE /*module*/, v8::Local<v8::Context> context, void*)
 #endif
 {
-  SWIG_InitializeModule(static_cast<void *>(&exports));
+#if NODE_MODULE_VERSION < 0x0040
+  v8::Local<v8::Context> context = SWIGV8_CURRENT_CONTEXT();
+#endif
+
+  SWIG_InitializeModule(static_cast<void *>(&context));
```
dot-asm, line 82: Just in case. @chenhengqi, I guess your remark was triggered by this `&context`. Well, it doesn't really matter from efficiency viewpoint if we keep passing this reference, or dereference it at cast-from-void-* point and pass v8::Local by value. It's more of a readability issue. Rationale was that passing `v8::Local<T>` by value is better aligned with the spirit of v8...

dot-asm:
Travis failure is JSC-related that is being resolved elsewhere.

Lib/javascript/v8/javascriptinit.swg
wsfulton: Given Local<> is just holds a pointer, the performance isn't an issue. Also v8.h mentions Local should be passed by value: https://github.com/v8/v8/blob/d9f3d4290e8dcbec9a82e4d6d07bd7e9528ae8e4/include/v8.h#L175-L178.

Lib/javascript/v8/javascriptinit.swg
```diff
@@ -124,6 +130,10 @@ void SWIGV8_INIT (SWIGV8_OBJECT exports, SWIGV8_OBJECT /*module*/)
 }
 
 #if defined(BUILDING_NODE_EXTENSION)
+#if (NODE_MODULE_VERSION < 0x0040)
```
wsfulton, line 133:
```suggestion
#if (NODE_MODULE_VERSION < 64)
```
Minor point, but from what I can see the convention for defining `NODE_MODULE_VERSION` is not as a hex number but as a decimal in node_version.h and also in the table at https://nodejs.org/en/download/releases/.

wsfulton (commented):
Looks okay to me.

Lib/javascript/v8/javascriptinit.swg
dot-asm: I wish it was legitimate to cast `context` to `void *` **here**. Of course one can probably make adjustments so that SWIG_InitializeModule would accept `Local<Context>`. Should I look into it?

Lib/javascript/v8/javascriptinit.swg
wsfulton: Yeah okay, change all usage in SWIG to decimal.

dot-asm: Done.

Lib/javascript/v8/javascriptinit.swg
wsfulton: Yeah why not if you can.

dot-asm: Note that I wrote "probably." It's not like it's impossible (nothing is), but it won't be few additional lines. SWIG_InitializeModule is common for all languages, and so are structures it uses. Taking https://github.com/swig/swig/blob/master/Lib/swiginit.swg#L121 for example. Using `Local<Context>` as type for `clientdata` would trigger V8-specific way to check if it's initialized. In other words it would be more #ifdef-s, which might be an overkill. After all, it's just **one** language among many...

It's late here, I'm going offline for today, and I have meetings tomorrow... So think about it, and give your verdict, today or tomorrow...

Cheers.

wsfulton: Leave it as is then. Thanks.

chenhengqi: @dot-asm Thanks for the details.

dot-asm: Hmmm... You know what? That reference to L121 means that this line we are commenting on is wrong. In sense that it would be inappropriate to save reference to local variable (if it is to be dereferenced later on). As mentioned, I'm a bit busy today, but I should be able to provide more adequate solution during the day...

dot-asm: It looks like it was assessment double-fault:-) Well, it was late after all. First of all, L121 has nothing to do with what we're talking about. And second faulty assessment was that it would be too disruptive for other languages. I've got caught thinking in multi-module terms, while SWIG is single-module in spirit. (Just in case, "multi-module" means separated headers and possible multiple compile units.) Either way `&context` is gone, so check it out. There also some additional minor cleanups...

Lib/swiginit.swg
```diff
@@ -50,9 +50,12 @@ extern "C" {
 #define SWIGRUNTIME_DEBUG
 #endif
 
+#ifndef SWIG_INIT_CLIENT_DATA_TYPE
+#define SWIG_INIT_CLIENT_DATA_TYPE void *
+#endif
```
wsfulton, line 55: I can't get my head round why this fallback definition is needed as `SWIG_INIT_CLIENT_DATA_TYPE` looks to be always defined (in javascriptinit.swg). So probably this can be removed. If not I'd like to see it defined to the different types in the same place in javascriptinit.swg so that it is clear when the type is different.

dot-asm: It won't be defined in **other** languages, hence that fallback.

(It ought to be late in the day for you too, huh:-)

dot-asm: Should I throw in comment mentioned that it can be overridden in language-specific modules?

wsfulton: Duh, I missed that it was in a general purpose file and not a v8 specific file. Leave it as is, I'll add a comment. It's otherwise okay for merging, which I'll do.

dot-asm: Thanks!

If you feel that I can help with something else related, now or in the future, just tag me. Well, I naturally can't guarantee that I'll have time or that it won't be over my head, but I'll definitely try to help.

Cheers. (Out for today.)

dot-asm: One extend this SWIG_INIT_CLIENT_DATA_TYPE approach to recent modifications to Lib/javascript/jsc/javascriptinit.swg. If you don't have time, just say the word.

dot-asm:
Just in case, a reminder. As already [mentioned](https://github.com/swig/swig/pull/1746#issuecomment-780910946), this is not fully-fledged support for context awareness. But it does work for purely computational tasks, i.e. when there is no shared data, shared among objects, being preserved between calls to C++. Plan is to proceed on as-problem-emerges basis, because it's easier to resolve an actual problem than to speculate what the problem might be.

wsfulton:
I've merged this. If possible please add some testcases to test this in another pull request. Thanks.

wsfulton:
@dot-asm, regarding help should you have some spare time... any comments or improvements for pull requests or issues tagged with javascript will be most helpful.

dot-asm:
> If possible please add some testcases to test this in another pull request.

See if test in #1965 would be acceptable.

> any comments or improvements for pull requests or issues tagged with javascript will be most helpful.

I don't really plan to actively follow the project and seek for problems to solve. But if you feel I can help with something specific, tag my handle when responding on github, and I'll see what I can do. [*You* in the previous sentence doesn't necessarily mean you personally, offer stands for all maintainers.]

PR #1961, opened 2021-03-02
CMake follow-up

xantares:
see #1040


wsfulton:
The changes are all good.

The appveyor cmake builds still need warnings as errors turned on though. The following in particular needs addressing, it seems quite fundamental. I know there a bewildering set of flags required to before cl will behave like a normal compiler, doesn't CMake get this right by default?
```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\xlocale(341): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc [C:\projects\swig\swig.vcxproj]
```

xantares:
what's the option in msvc to turn warnings into error ? I dont find it ?

vadz:
On Thu, 04 Mar 2021 23:46:37 -0800 xantares <notifications@github.com> wrote:

x> what's the option in msvc to turn warnings into error ? I dont find it.

 It's /WX
VZ

wsfulton:
I see the latest change, but that change is manually passing specific flags to the compiler. Is that what every C++ project expects in order to work with Visual C++? Just feels horribly wrong.

xantares:
We could also hardwire these flags in the cmake project

> I see the latest change, but that change is manually passing specific flags to the compiler. Is that what every C++ project expects in order to work with Visual C++? Just feels horribly wrong.

it's a choice to hardwire flags or not

CMakeLists.txt
```diff
@@ -69,6 +69,14 @@ check_include_file_cxx ("boost/shared_ptr.hpp" HAVE_BOOST)
...
 check_function_exists (popen HAVE_POPEN)
 
+if (MSVC)
+  set (CMAKE_C_FLAGS "/WX ${CMAKE_C_FLAGS}")
+  set (CMAKE_CXX_FLAGS "/EHsc /WX ${CMAKE_CXX_FLAGS}")
+elseif (CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+  set (CMAKE_C_FLAGS "-Wall -Wextra ${CMAKE_C_FLAGS}")
+  set (CMAKE_CXX_FLAGS "-Wall -Wextra ${CMAKE_CXX_FLAGS}")
+endif ()
+
```
wsfulton, line 79: The approach that we use in the autotools and I think is common across open source projects is to **not** enable warnings as errors by default. However, we do turn this on in the Travis/Appveyor tests. The thinking is that you want the code to build with warnings that later versions of the compilers may add without breaking the build for users. However, we want to break the build during development if a warning appears. So in summary, move `/WX` into Appveyor.yml.

wsfulton (changes requested):

CMakeLists.txt
xantares: Yes, I prefer that

wsfulton (approved):

wsfulton:
Thanks, committed, see c6af0c5a677223868c8a515501db08defab674e9..167164885dfc80776178fced46a805dc768359a3

Issue #1498, opened 2019-03-12
Shouldn't SWIG_V8_ConvertInstancePtr call SWIG_TypeCast if types are not equal?


murillo128:
I have the current code snippets:

```
%nodefaultctor AudioInput;
%nodefaultdtor AudioInput;
struct  AudioInput {};

%nodefaultctor AudioOutput;
%nodefaultdtor AudioOutput;
struct  AudioOutput {};

struct AudioPipe : 
	public AudioInput,
	public AudioOutput
{
	AudioPipe(int rate);
};


struct AudioDecoderFacade
{
	int Start();
	void AddAudioOuput(AudioOutput* ouput);
	void RemoveAudioOutput(AudioOutput* ouput);
	bool SetIncoming(RTPIncomingMediaStream* incoming);
	int Stop();
};
```

which generates the following code for the `AddAudioOuput` method:

```

static SwigV8ReturnValue _wrap_AudioDecoderFacade_AddAudioOuput(const SwigV8Arguments &args) {
  SWIGV8_HANDLESCOPE();
  
  v8::Handle<v8::Value> jsresult;
  AudioDecoderFacade *arg1 = (AudioDecoderFacade *) 0 ;
  AudioOutput *arg2 = (AudioOutput *) 0 ;
  void *argp1 = 0 ;
  int res1 = 0 ;
  void *argp2 = 0 ;
  int res2 = 0 ;
  
  if(args.Length() != 1) SWIG_exception_fail(SWIG_ERROR, "Illegal number of arguments for _wrap_AudioDecoderFacade_AddAudioOuput.");
  
  res1 = SWIG_ConvertPtr(args.Holder(), &argp1,SWIGTYPE_p_AudioDecoderFacade, 0 |  0 );
  if (!SWIG_IsOK(res1)) {
    SWIG_exception_fail(SWIG_ArgError(res1), "in method '" "AudioDecoderFacade_AddAudioOuput" "', argument " "1"" of type '" "AudioDecoderFacade *""'"); 
  }
  arg1 = (AudioDecoderFacade *)(argp1);
  res2 = SWIG_ConvertPtr(args[0], &argp2,SWIGTYPE_p_AudioOutput, 0 |  0 );
  if (!SWIG_IsOK(res2)) {
    SWIG_exception_fail(SWIG_ArgError(res2), "in method '" "AudioDecoderFacade_AddAudioOuput" "', argument " "2"" of type '" "AudioOutput *""'"); 
  }
[...]

PR #1365, opened 2018-11-26
implement shared_ptr in JavaScript

mislavn:
I want to add shared_ptr functionality to SWIG 4.0, this PR is based on #236.

Currently the JavaScript bindings compile but the destructors are not called and I need to add unit test requested in #236. Any help with this is appreciated.

PR #1963, opened 2021-03-05
Use SWIG_TypeCast in SWIG_V8_ConvertInstancePtr if types don't match 

murillo128:
As per #1498

### DIFF
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -289,8 +289,12 @@ SWIGRUNTIME int SWIG_V8_ConvertInstancePtr(SWIGV8_OBJECT objRef, void **ptr, swi
     if(!type_valid) {
       return SWIG_TypeError;
     }
+    int newmemory = 0;
+    *ptr = SWIG_TypeCast(tc, cdata->swigCObject, &newmemory);
+  } else {
+    *ptr = cdata->swigCObject;
   }
-  *ptr = cdata->swigCObject;
+
   if(flags & SWIG_POINTER_DISOWN) {
     cdata->swigCMemOwn = false;
   }


wsfulton:
This change is also part of #1365.

Lib/javascript/v8/javascriptrun.swg
```diff
@@ -289,8 +289,12 @@ SWIGRUNTIME int SWIG_V8_ConvertInstancePtr(SWIGV8_OBJECT objRef, void **ptr, swi
     if(!type_valid) {
       return SWIG_TypeError;
     }
+    int newmemory = 0;
+    *ptr = SWIG_TypeCast(tc, cdata->swigCObject, &newmemory);
```
dot-asm, line 293: What jumps at me is that it makes you wonder if `newmemory` would require handling. If I "swig" your snippet and look, I see that conversion functions don't use `newmemory` in any way. It's probably used in other languages, but it's irrelevant in the context. And given these specific javascript circumstances I for one would argue that passing `NULL` instead of `&newmemory` would be a slightly better approach. Because **if** something changes and conversion subroutines would **start** using `newmemory`, it will be denoted by SEGV, which will inevitably prompt maintainers to look at this exact spot and figure out proper way to handle it. I mean if it doesn't SEGV, and `newmemory` would actually require some handling, then the problem might slip through the cracks and remain unnoticed for a longer period of time.

Disclaimer. I'm not swig maintainer, and the above comment might be misaligned with maintainers' opinions. But if the suggested approach is found sensible and applied, it would also be appropriate to leave a short "why-NULL" note in commentary.

wsfulton: When #1365 is incorporated, the newmemory flag will be needed. It is used by the smart pointer casts, where you can't simply cast one pointer to another as the objects are actually pointers to smart pointers where casting requires creating a new smart pointer containing the cast pointer. Not sure if that makes sense, but take a look at say some Python wrappers wrapping shared_ptr.

Without the smart pointer support, I agree NULL would be better because this patch does not handle the case of newmemory being set.

murillo128: The shared pointer patch handles the newmemory variable, we may just reuse it here:

https://github.com/swig/swig/pull/1365/files#diff-339ae135626830816428ae1daef3325fa3eb5ab3ebb9a3aba2c10c22fcad3dd9R138

wsfulton:
I used your code in fee5e239ecdc9ecf994f2647ab719a3fc52a86a8 and a follow up commit 6aef217438328f6653be270fa0e2296a5254a89c fixes it for JSC.

PR #1973, opened 2021-03-14
WIP: JavaScript V8 isolate-aware init

dot-asm:
These are **intermediate** commits mostly for review, in essence to ensure that it doesn't go in undesired direction. Undesired by the maintainers that is:-) With this in mind it might make more sense to look at individual commits...

[Just in case, ping @wsfulton].


Lib/javascript/v8/javascriptrun.swg
```diff
@@ -466,7 +477,7 @@ SWIGRUNTIME SWIGV8_VALUE SWIG_V8_NewPointerObj(void *ptr, swig_type_info *info,
 
 #define SWIG_GetInstancePtr(obj, ptr)    SWIG_V8_GetInstancePtr(obj, ptr)
 
-SWIGRUNTIME SwigV8ReturnValue _SWIGV8_wrap_equals(const SwigV8Arguments &args) {
+SWIGRUNTIME SwigV8ReturnValue _wrap_SWIGV8_Proxy_equals(const SwigV8Arguments &args) {
```
wsfulton, line 480: _wrap prefix is reserved for wrapper code, so best left as it was.

Lib/swiginit.swg
```diff
@@ -50,13 +50,29 @@ extern "C" {
 #define SWIGRUNTIME_DEBUG
 #endif
 
+static void SWIG_InitializeTypes(void);
```
wsfulton, line 53: This is a generic file used by all languages and should not have any lanaguage specific code. The Javascript code needs moving to Javascript specific files.

wsfulton (commented):
I've a couple of comments, otherwise I think it is okay.

Lib/javascript/v8/javascriptrun.swg
dot-asm: I've spotted inconsistency in `_SWIGV8_wrap_equals` vs. `_wrap_getCPtr` naming and tried to harmonize the names. Rationale behind switching is that these are methods just like any other. They are kind of hardwired. But sure, let's stick with `_SWIGV8_wrap_equals`. Would it be appropriate to rename `_wrap_getCPtr` as `SWIGV8_wrap_getCPtr`?

Lib/swiginit.swg
dot-asm: Yes, and it's common subroutine and it's the **same** for all languages. If you consider the corresponding https://github.com/swig/swig/pull/1973/commits/26f95e964c6b14acdde979611fad891364d5b7c5 by itself and **disregard** the `static class SWIG_PreInit` thing for a moment, you'll see that the overall result would be 100% equivalent to the original. So splitting per se is perfectly language-neutral. Now, it was also assumed that calling `SWIG_InitializeTypes` separately from static constructor would also be language-neutral. This turned to be wrong assumption, so I'll have to figured it out. For the moment I make static constructor dependent on SWIGJAVASCRIPT macro, but yes, it needs better solution. First I need to understand if the reasons for why it fails for, say, Python are applicable even to Javascript. If it can't happen in Javascript, then I'll simply move the static constructor to Javascript module...

Just in case for reference. Why static constructor? Since static constructor is executed at library load, and we know that library loading is MT-safe, we can use the latter property to avoid race conditions in `SWIG_InitializeModule`. It's attractive, because the approach is platform-neutral, i.e. you don't have to think if it's POSIX or Windows.

Lib/javascript/v8/javascriptrun.swg
wsfulton: This looks bit of a mess in the Javascript module, both v8 and jsc. There are a number of runtime functions (the ones using the `SWIGRUNTIME` macro) which should begin with `Swig` or `SWIG`, not `_wrap`. In fact, all of the functions in the Lib/javascript directory should begin not begin with `_wrap`. I would be more than happy to consider a separate pull request to fix all of these.

Lib/swiginit.swg
wsfulton: Cool. I'd prefer initialisation to be done during the SWIG_InitializationModule if possible as initialization problems during file load are a pig to debug. A singleton function which returns a pointer to a static local instance can be initialized in a thread-safe manner when it is first called for C++11 and later, although in practice most compilers implement it as thread-safe for C++98 too.

dot-asm:
Just in case, last push was a unintended. I'll be responding and addressing feedback later... Sorry about the distraction:-(

Lib/javascript/v8/javascriptrun.swg
dot-asm: I've removed the method renaming part.

Lib/swiginit.swg
dot-asm: Call it occupational hazard, as I do have tendency to minimize reliance on compiler and run-time specifics and prefer solutions that are universal:-) In some degree it's a side effect of dealing with VC compiler, which to this day appears to be reluctant to commit to standards. `__cplusplus` pre-define is an example of that. They do let you specify `/std:c++14,17,latest` on compiler command line, but do they reflect it in `__cplusplus` pre-define? No, it reads `199711L` irregardless. I apologize for only tangentially related rant:-)

I've omitted the static constructor in favour of C++11 mutex. This effectively implies that those who intend to use swig-generated module with node.js worker_threads have to ask compiler for at least C++11 compliance. It's appropriate to inform the user about it. I can add a note to Doc/Manual/Javascript.html, but  one can also add `#warning` in `#else` condition for `<mutex>`. Would it be appropriate?

dot-asm: > one can also add `#warning` in `#else` condition for `<mutex>`

Another possibility is to have an `assert` in `#else` condition for quard_locker that would try to detect if worker_threads are used at run-time and crash in a more controllable manner. This can be done by saving `isolate` value to a global `void *` and crash if it changes from one non-zero value to another. Well, formally speaking it would require `<atomic>`, which is also a C++11 thing, but one can make "best effort" with volatiles...

dot-asm:
Additional commit allows to pass optional data argument specified when creating the callback. It can be considered controversial for two reasons.

1. Use of variadic macro in SWIGV8_FUNCTEMPLATE_NEW. Even though `__VA_ARGS__` was standardized only in C++11, it was supported longer than that, by GCC 3.0, Visual Studio 2005, etc. It's not unreasonable to assume that users who venture into V8 programming do use contemporary compiler.

2. Injecting a local variable, `jdata`, into some `SWIG_New*` macro expansions. `SWIG_New*` are used by common language-neutral typemaps and arranging option for an additional language-specific argument might require more extensive surgery beyond the scope of this particular problem.

Thoughts?

wsfulton:
Hi @dot-asm. I hope this isn't too frusrating, but regarding code generation, this is the philosophy: http://swig.org/Doc4.0/Introduction.html#Introduction_nn13. So I'm afraid `__VA_ARGS__` are only a last choice option. 

Regarding the locking classes you've used, they are clearly not universally available. We usually avoid platform specific implementations by using the target language APIs, which are going to be available on the platforms of interest for that target language and will be much better tested than anything we could come up with independently. I'm not entirely sure what you're trying to achieve in this commit now, so am not best placed to suggest any specifics, other than suggest you aim for portable techniques such as:

1. Use the singleton locking pattern for initialisation that I mentioned earlier or file scope static initialisation (like you proposed earlier)
2. Use some Javascript APIs for locking.

Modifying `SWIG_NewPointerObj` family of functions is a no no, they are common across all languages. Could you perhaps enhance the `swig_type_info::client_data` instead of modifying the function signature. If you provide a full explanation what the extra parameter is for, perhaps I can understand better and find a generic way to incorporate it for all scripting languages.

wsfulton:
Incidentally, I noticed you modified calls to `SetWeak`, which are currently giving hundreds of warnings on Ubuntu 20.04 when using the platform provided node: `sudo apt install nodejs node-gyp` which installs

Node.js: v10.19.0
node-gyp: v6.1.0

```
checking Examples/javascript/class
In file included from /usr/include/nodejs/src/node.h:63,
                 from ../example_wrap.cxx:171:
/usr/include/nodejs/deps/v8/include/v8.h: In instantiation of ‘void v8::PersistentBase<T>::SetWeak(P*, typename v8::WeakCallbackInfo<P>::Callback, v8::WeakCallbackType) [with P = SWIGV8_Proxy; T = v8::Object; typename v8::WeakCallbackInfo<P>::Callback = void (*)(const v8::WeakCallbackInfo<SWIGV8_Proxy>&)]’:
../example_wrap.cxx:1163:112:   required from here
/usr/include/nodejs/deps/v8/include/v8.h:9502:16: warning: cast between incompatible function types from ‘v8::WeakCallbackInfo<SWIGV8_Proxy>::Callback’ {aka ‘void (*)(const v8::WeakCallbackInfo<SWIGV8_Proxy>&)’} to ‘Callback’ {aka ‘void (*)(const v8::WeakCallbackInfo<void>&)’} [-Wcast-function-type]
 9502 |                reinterpret_cast<Callback>(callback), type);
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
I started to get a headache with the overloaded templates, SWIG version macros and build obfuscation that node-gyp does, so went and did something else. If you have any insight, please let me know.

dot-asm:
> Incidentally, I noticed you modified calls to `SetWeak`

I have to disagree with this assessment:-) If you checkout 7b37c8cd022d027c887482f028d00d3503bf5b39, which **predates** anything I've suggested, you'll still get the warnings. I'll have a look, but sounds more like a bug in compiler warning subsystem... Most notably you don't get it with older gcc or clang... But I'll see what can be done...

dot-asm:
> I hope this isn't too frusrating,

No, it's all cool, your project - your rules. That's why the request is marked "WIP" and why I said "[intention is to] ensure that it doesn't go in undesired direction" upfront:-)

> but regarding code generation, this is the philosophy: http://swig.org/Doc4.0/Introduction.html#Introduction_nn13. So I'm afraid `__VA_ARGS__` are only a last choice option.

I understand and appreciate the approach, but at the same time you have to recognize that you can't do better than any particular target language. And for this reason it makes lesser sense to hold all wrapper generators to the very same standard. Here is what I mean. Minimum requirement for building V8 is gcc 4.8, clang 3.5, Visual Studio 2013, and all of them support variadic macros. And hence one can argue that it wouldn't be unreasonable to allow them **specifically** in V8 generator:-)

Of course, you can say that these are **build** requirements, and user might get away with compiling wrapper with older compiler. Well, g++ 3.4 can parse 3.14.5 v8.h, but not 5.0... You can still say that in such case it would be preferable if error is attributed specifically to v8.h, and not to the wrapper. Well, g++ 3.4 can do variadic macros, so do all clangs, so errors won't be misattributed...

But OK, I'll get rid of variadic macros in finalized submission:-)

[I'll address other remarks in separate responses.]

dot-asm:
> Modifying `SWIG_NewPointerObj` family of functions is a no no,

Well, now, when I look closer, I can see that the additional argument injection is not actually uncharted, so the suggestion shouldn't be that controversial. Indeed, quoting Lib/python/pyrun.swg:
```
#define SWIG_NewPointerObj(ptr, type, flags)   SWIG_Python_NewPointerObj(self, ptr, type, flags)
```
I'm obviously referring to the additional `self` argument injected in macro expansion. And one can observe similar pattern in jsc, scilab... It's just that I've added additional argument to the end of the argument list. For obvious reasons, in order to be able to declare it optional. As for the latter, argument declared optional. I expect that it will be possible to remove it in the finalized version, so that if alternative order is preferred, it can be arranged later on.

> they are common across all languages.

Ideally, there should be a common option for language-specific argument. But it's surely a separate topic... And with scilab injecting 3[!] additional arguments, it definitely goes beyond the scope of this request, and can be discussed/resolved later for all relevant generators. Plus V8;-)

> Could you perhaps enhance the `swig_type_info::client_data` instead of modifying the function signature. If you provide a full explanation what the extra parameter is for, perhaps I can understand better and find a generic way to incorporate it for all scripting languages.

No, I can't do it through client_data. The problem is that I can't save Templates to global variables. More specifically not with multiple isolates. So that in the final version the clazz_temp in compartmentalized SWIGV8_ClientData will go unused and I'll use index field to address templates in per-isolate [and per-context?] tables that will be passed through that additional `jsdata` argument. Formally speaking this is not the only way to solve the problem, but I reckon it's most natural, simplest and presumably most efficient one in the circumstances.

dot-asm:
> Regarding the locking classes you've used, they are clearly not universally available.

That's the thing, they **are**, in C++11. And I argue that it **reasonable** to expect that users who aim for worker_threads would be required to use C++11 or better compiler. Note that if they don't, they will still be able to compile and it would work without worker_threads as it did so far. The remaining reasons for the choice should become more apparent from below.

> We usually avoid platform specific implementations

100% with you.

> by using the target language APIs, which are going to be available on the platforms of interest for that target language and will be much better tested than anything we could come up with independently.

Right, C++11 standard has your back.

> I'm not entirely sure what you're trying to achieve in this commit now, so am not best placed to suggest any specifics, other than suggest you aim for portable techniques such as:
> 
> 1. Use the singleton locking pattern for initialisation that I mentioned earlier

I'm aware of Singleton Pattern, and it's not MT-safe by itself, it would still require mutex-ing. There are references to something called Double-Checked Locking with Singleton, but it's a Java-specific pattern that utilizes Java builtin synchronization and atomics. Adopting it for the purposes in question would still require mutex-ing (and atomics) and it would be an overkill.

> or file scope static initialisation (like you proposed earlier)

As mentioned, static constructor turned to be non-language-neutral. And with V8 as sole application, it felt unjustified. I figured it would be more appropriate to keep the solution more localized and focused.

> 2. Use some Javascript APIs for locking.

V8 doesn't provide inter-isolate locking, only within the same isolate. It's used for synchronization of promises and other async-s. Now, node[!] provides MT mutexes, or rather libuv, but I've chosen to keep it node-neutral. It's also more elegant...

dot-asm:
> node[!] provides MT mutexes, or rather libuv, but I've chosen to keep it node-neutral.

Just in case, this does mean that if a node-specific solution is preferred, it can be arranged.

dot-asm:
> Incidentally, I noticed you modified calls to `SetWeak`

Problem is even invariant to 7b37c8c. I mean you can take swig 4.0.1, g++8 or later, and as soon as you pass `-Wextra`, it will complain with node.js v6, v8, v10. So that nobody modified anything, gcc simply got "overzealous."

dot-asm:
There are test cases that are bound to fail with worker_threads, those modifying and checking static variables, which would be shared among all the workers. One can make case for automatic instantiation of **copies** of all[?] non-constant variables, so that each isolate [or context?] gets own copy to work on. But this goes beyond the scope of this request, and hence the question is rather how to suppress tests that are bound to fail? Two possibilities:

1. add a "not to be tested with worker_threads" commentary to problematic runme.js, so that one can grep and skip over;
2. pick just a couple of tests and hard-code them into makefile.in, maybe it would be possible to express it with a glob;

2 is obviously easier to implement, so I'll start with that.

wsfulton:
Regarding threading requiring C++11, I'll buy your argument that worker_threads need a C++11 compiler. You've coded it up as being an optional advantage and documented this which is fine by me. 

> I'm aware of Singleton Pattern, and it's not MT-safe by itself, it would still require mutex-ing. There are references to something called Double-Checked Locking with Singleton, but it's a Java-specific pattern that utilizes Java builtin synchronization and atomics.
 
It is thread-safe by design in C++11 (no need for double checked locking) because local statics are guaranteed to be initialised just once by the C++11 standard. I hence quite like this pattern because if you are using C++11 it is guaranteed to be thread safe and if you are not using C++11, but using a modern compiler you happen go get the thread safety anyway, because the compilers don't make the code thread-unsafe in C++98 mode.

What I don't understand, as there is no use in case in the patch, is how often is `SWIGV8_INIT` called? Is it okay for the initialisation to be repeated, because it looks like you allow that?

The addition of `jsdata` as a way to keep the `SWIG_NewPointerObj` and similar functions unchanged is okay too.

Regarding use of `__VA_ARGS__` in `SWIGV8_FUNCTEMPLATE_NEW`... I see that `FunctionTemplate::New`, that it is wrapping in a Node version neutral way, takes a variable number of arguments, but I can't see anything other than one argument being used. I think `SWIGV8_FUNCTEMPLATE_NEW` is for internal use, so is there any point in changing it from the way it is now? If there is a real need for it to be varargs, then no worries, keep it that way.

In summary, I this patch is okay, but I await responses to above, in particular details on the use case required for it. Perhaps, you can document, how worker_threads should be used when using SWIG and how they interact with the initialisation code in order to clarify.

Lib/swiginit.swg
wsfulton: Yes, it is atrocious that Visual C++ does not respect even getting `__cplusplus` correct.

Lib/javascript/v8/javascriptrun.swg
```diff
@@ -236,8 +240,10 @@ public:
 };
 
 class SWIGV8_ClientData {
+private:
+  v8::Persistent<v8::FunctionTemplate> clazz_templ;
 public:
-  v8::Persistent<v8::FunctionTemplate> class_templ;
+  size_t index;
```
wsfulton, line 246: Is this used for anything? I can see anything reading it.

Lib/javascript/v8/javascriptrun.swg
```diff
@@ -250,13 +256,47 @@ public:
...
+#if (V8_MAJOR_VERSION-0) >= 4
+    if (v8_swig_types != NULL) {
+      v8_swig_types[index+1].Reset(isolate, class_templ);
+      return;
+    }
+#endif
+    if (clazz_templ.IsEmpty()) {
+      SWIGV8_SET_CLASS_TEMPL(clazz_templ, class_templ);
+    }
+  }
```
wsfulton, line 294: Perhaps some consistency in the getter and setter naming would be better here... `GetClassTemplate()` and `SetClassTemplate()` ??

wsfulton:
Oh, you've just rebased the patch. It's hard to follow what is going if you rebase as now I don't know what you have changed since my last comments.

dot-asm:
> Oh, you've just rebased the patch. It's hard to follow what is going if you rebase as now I don't know what you have changed since my last comments.

Yes, but individual commits are incremental. However, I'll provide more detailed replies and point out new potentially controversial changes tomorrow:-)

Lib/javascript/v8/javascriptrun.swg
dot-asm: It is now. As mentioned, as it's taken into multi-isolate environment, we no longer can save templates' handles to global variables. But we can save index within per-isolate table. Or in other words, `index` effectively replaces `class_templ`. And the private `clazz_templ` is a fall-back for legacy V8 versions.

I've spotted apparent inconsistency in `#if V8_MAJOR_VERSION` condition here. I'll ponder a little bit more about it...

Lib/javascript/v8/javascriptrun.swg
dot-asm: Yes, sure. Just in case, the original rationale behind current naming choice is to make modification of callers minor. For example original `->class_templ` becomes `->class_templ()`. It's easier to review, but of course one can use the opportunity and make naming more self-consistent...

dot-asm:
> > [Singeton]

> It is thread-safe by design in C++11 ... because local statics are guaranteed to be initialised just once by the C++11 standard.

But I didn't find any references specifically to singleton in C++11... You ought to refer to `std::call_once`, right? In which case it would explain my confusion:-) Incidentally it's defined in `<mutex>`, the very same one used for `std::lock_guard`, so we're really talking about the same thing. Same in the sense that whatever portability/availability objection one can throw at lock_guard would apply even to call_once.

On a more specific note. If we were to use call_once, we would have to take it to the common-language domain. And given the type of failure I've experienced with the originally suggested static constructor with other languages, it would probably take two call_once's... Either way it would go out of scope of just one language, and it's probably not the right time and place to do something like that.

dot-asm:
> how often is `SWIGV8_INIT` called? Is it okay for the initialisation to be repeated, because it looks like you allow that?

It's called for each context that calls `require` and different contexts can be in different isolates. Each caller is entitled for an export object, so there is something that needs to be done in **each** call. There are things to do just once (globals), there are things to do in each context (instantiate objects), and there are things to do in each isolate (create templates). If you look closely you'll notice that the proposed patch does **everything** [V8-related!] in each call instead of just what's actually needed in each particular case. Yes, it's a fair point, and there are two practical counterarguments to it.

1. It's reckoned that the most common scenario would be to create isolates on a per-worker basis and it would customarily accommodate one context. This means that things to do in each context and in each isolate are more than likely to "coalesce." Which means that there are rather just two things to do, once and per worker. Things that need to be done once are just constant assignments to clientdata structures, indices and pointers to destructors. And as for the unlike scenario when templates will be duplicated in the same isolate, well, it's can't and won't be devastating in overall context.
2. Since the do-once things mentioned above are effectively constant assignments, it is in fact possible to execute them multiple times without side effects. And since they are fast [in comparison to the rest], it's argued that  rearranging the code to accommodate the required alternative pattern simply isn't worth the trouble.

To summarize. The trade-off is between how intrusive modifications are and minuscule loss of performance, if any. And it's the former that is favoured here.

dot-asm:
> `__VA_ARGS__` ... can't see anything other than one argument being used... If there is a real need for it to be varargs, then no worries, keep it that way.

It was reservation for further modifications that were added yesterday. Since it wasn't clear how many I would need, it felt more convenient to make it as flexible as possible. As "dust settles" it would be possible to reconsider it. But if it's acceptable after all, then OK, less work;-)

dot-asm:
Now, the controversial part in the yesterday modifications. Things like
```
#define SWIGV8_HANDLESCOPE() v8::HandleScope scope(isolate)
```
Keyword is that a number of macros assume that there is `isolate` variable declared in the current scope. The reason for the choice is obvious, less typing this time around, but it can be confusing for those who follow. I mean if somebody else implements a new function later on and doesn't declare `isolate`, the error message from the compiler would be confusing. On the other hand there are plenty of examples to follow...

dot-asm:
> There are test cases that are bound to fail with worker_threads, ... the question is how to suppress tests that are bound to fail? Two possibilities:
> 
> 1. add a "not to be tested with worker_threads" commentary to problematic runme.js, so that one can grep and skip over;
> 2. pick just a couple of tests and hard-code them into makefile.in, maybe it would be possible to express it with a glob;
> 
> 2 is obviously easier to implement, so I'll start with that.

Since I've spotted just two failures I actually implemented 1. However! Since it's based solely on empirical observations, it doesn't actually mean that more can't legitimately fail. So there is an uncertainty in test outcome now...

dot-asm:
I think I've addressed all remarks. The remaining question from my side is about implied `isolate` variable. Out for today...

dot-asm:
As for the mess on travis. I simply experimented with a way to turbo-charge test by bypassing node-gyp that takes ~1/2 of execution time. But this unmasked a number of warnings that are -Werror-ed...

dot-asm:
> The remaining question from my side is about implied `isolate` variable.

There is an advantage of having implied `isolate` and it's when developer specifies custom typemap and uses affected macro in it. I mean if macros are modified in 4.1, then developers might find themselves in situation that their custom typemaps would need adjustment...

dot-asm:
In essence as far I'm personally concerned, WIP can be removed.

dot-asm:
> > Incidentally, I noticed you modified calls to `SetWeak`

> ... nobody modified anything, gcc simply got "overzealous."

Just in case, it's mitigated in 9898de5, **and** on related note, one can extend this to suppress more warnings specifically in v8.h. As there **are** more, which are not actually v8-version specific. I mean the original remark turned to be pre-v12-specific, but one can get different warnings even in v14 v8.h...

molllyn1 (approved):

dot-asm:
Difference between previous push is 4b23e3c, which adds commentary about isolate variable. It also adds SWIGV8_OBJECT_INEW macro to be used internally.

dot-asm:
Minor modifications in d5c069d and new fb95710. Just in case for reference, [not that it matters a lot on a grand scale, but] the  d5c069d stands for for ~60% performance improvement of a tight loop in javascript that calls an inexpensive method.

dot-asm:
To summarize. In its core this resolves real-life hard SEGV crashes at either module load by multiple workers or a worker termination times. All "runme" tests but two were complemented with corresponding two-worker tests for node v12 and further. I consider feedback provided so far to be addressed/resolved. If there are more concerns I'm happy to address them.

Lib/javascript/v8/javascriptrun.swg
```diff
@@ -357,32 +378,32 @@ SWIGRUNTIME void SWIGV8_SetPrivateData(SWIGV8_OBJECT obj, void *ptr, swig_type_i
...
-    cdata->handle.MakeWeak(cdata, ((SWIGV8_ClientData*)info->clientdata)->dtor);
+  if(cdata->swigCMemOwn && info->clientdata) {
+    cdata->handle.MakeWeak(cdata, info->clientdata->dtor);
   } else {
     cdata->handle.MakeWeak(cdata, SWIGV8_Proxy_DefaultDtor);
   }
 #elif (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < 0x031918)
-  if(cdata->swigCMemOwn && (SWIGV8_ClientData*)info->clientdata) {
-    cdata->handle.MakeWeak(v8::Isolate::GetCurrent(), cdata, ((SWIGV8_ClientData*)info->clientdata)->dtor);
+  if(cdata->swigCMemOwn && info->clientdata {
```
dot-asm, line 419: Oops! Typo here, missing `)`.

dot-asm:
Additional commit to clean up `object` variable handling in destructors.

As for red X from travis. It now installs newer node-gyp for node v10 and forward, and it doesn't work on xenial.

dot-asm:
As a ping, to summarize once again. In its core this resolves real-life hard SEGV crashes at either module load by multiple workers or a worker termination times.

If it's considered too messy, then please state so. If it would be appropriate to break it down to separate requests, the please state so.

dot-asm:
> As a ping, to summarize once again. In its core this resolves real-life hard SEGV crashes at either module load by multiple workers or a worker termination times.
> 
> If it's considered too messy, then please state so. If it would be appropriate to break it down to separate requests, then please state so.

Lib/javascript/v8/javascriptinit.swg
```diff
@@ -69,8 +76,53 @@ void SWIGV8_INIT (SWIGV8_OBJECT exports_obj, SWIGV8_VALUE /*module*/, v8::Local<
...
+  static std::mutex guard;
+  const std::lock_guard<std::mutex> lock(guard);
+#endif
 
   SWIG_InitializeModule(context);
+
+#if (V8_MAJOR_VERSION-0) < 5
+  // Multiple threads are not supported, everything is saved in globals.
+  v8::Persistent<v8::FunctionTemplate>* v8_swig_types = NULL;
+  v8::Local<v8::Value> v8_swig_types_as_data;
```
ojwb, line 92: SWIG now only supports V8 5.0 and later, so this branch (and the conditional) aren't needed now (probably wasn't the case when you started this patch).

Lib/javascript/v8/javascriptinit.swg
```diff
@@ -69,8 +76,53 @@ void SWIGV8_INIT (SWIGV8_OBJECT exports_obj, SWIGV8_VALUE /*module*/, v8::Local<
...
+  // template duplicates are tolerable. Just in case for reference, each
+  // Node.js Worker thread is executed in own |isolate|, which is the most
+  // common case when SWIGV8_INIT would be called multiple times.
+
+  // Per-context array that holds Persistent<Template>-s
+  size_t sz = sizeof(v8::Persistent<v8::FunctionTemplate>)
+            * (sizeof(swig_types)/sizeof(swig_types[0]) + 1);
+  v8::Local<v8::ArrayBuffer> v8_swig_types_as_data = v8::ArrayBuffer::New(isolate, sz);
+  v8::Persistent<v8::FunctionTemplate>* v8_swig_types =
+#if (V8_MAJOR_VERSION-0) >= 8
```
ojwb, line 115: For consistency with all the existing V8 version checks in SWIG this should really be:

```
#if (SWIG_V8_VERSION >=0x0800)
``

File: Lib/javascript/v8/javascriptinit.swg [hunk 8]
```diff
@@ -120,7 +173,14 @@ void SWIGV8_INIT (SWIGV8_OBJECT exports_obj, SWIGV8_VALUE /*module*/, v8::Local<
 #if (NODE_MODULE_VERSION < 64)
 NODE_MODULE($jsname, $jsname_initialize)
 #else
-NODE_MODULE_CONTEXT_AWARE($jsname, $jsname_initialize)
+#ifndef NODE_GYP_MODULE_NAME
+# define NODE_GYP_MODULE_NAME $jsname
+#endif
+#if defined(__GNUC__) && __GNUC__>=8
+// NODE_MODULE_INIT is not self-consistent in all supported Node.js versions :-(
+# pragma GCC diagnostic ignored "-Wcast-function-type"
```
ojwb, line 181: It would be cleaner to `# pragma GCC diagnostic push` before this, then after the place we want to suppress the warning to repeat the `#if` conditonal with `# pragma GCC diagnostic pop` inside.

Is this needed for clang too?  All non-antique clang versions currently pretend to be GCC 4.2.1 by default.

It would also be good to note in a comment the node version at which this became consistent so that when SWIG raises its minimum supported node version we know whether this workaround is still needed.  Otherwise it'll end up lingering for years after it could have been removed.

Lib/javascript/v8/javascriptrun.swg
ojwb: Why do we always define `clazz_templ` when it seems it's only needed for V8 < 8?

Can't we use a preprocessor conditional to only have the mechanism needed for the V8 version we're actually compiling against?

Lib/javascript/v8/javascriptrun.swg
```diff
@@ -263,43 +297,52 @@ SWIGRUNTIME int SWIG_V8_ConvertPtr(SWIGV8_VALUE valRef, void **ptr, swig_type_in
...
-SWIGRUNTIME SWIGV8_VALUE SWIG_V8_NewPointerObj(void *ptr, swig_type_info *info, int flags) {
+SWIGRUNTIME SWIGV8_VALUE SWIG_V8_NewPointerObj(void *ptr, swig_type_info *info, int flags,
+                                               SWIGV8_VALUE data) {
+  v8::Isolate* isolate = v8::Isolate::GetCurrent();
   SWIGV8_HANDLESCOPE_ESC();
-  
+  v8::Persistent<v8::FunctionTemplate>* v8_swig_types = NULL;
+#if (V8_MAJOR_VERSION-0) >= 5
+  v8::Local<v8::ArrayBuffer> ab = v8::Local<v8::ArrayBuffer>::Cast(data);
+#if (V8_MAJOR_VERSION-0) >= 8
```
ojwb, line 307: `#if (V8_MAJOR_VERSION-0) >= 5` can be assumed now.  The `>= 8` check should be `SWIG_V8_VERSION >= 0x0800` to match existing code.

And similarly for any other uses to avoid me having to note them individually when a `git grep V8_MAJOR_VERSION` will find them for you.

Lib/javascript/jsc/javascriptstrings.swg
```diff
@@ -8,34 +8,32 @@ SWIG_JSC_AsCharPtrAndSize(JSContextRef context, JSValueRef valRef, char** cptr,
   if(JSValueIsString(context, valRef)) {
     JSStringRef js_str = JSValueToStringCopy(context, valRef, NULL);
     size_t len = JSStringGetMaximumUTF8CStringSize(js_str);
-    char* cstr = (char*) %new_array(len, char);
-    /* JSStringGetUTF8CString returns the length including 0-terminator */
-    len = JSStringGetUTF8CString(js_str, cstr, len);
-
-    if(alloc) *alloc = SWIG_NEWOBJ;
+    if(cptr) {
```
ojwb, line 11: This seems unrelated to the "isolate-aware" changes unless I'm missing something.  This is quite a large patch so including unrelated changes (even if they are useful changes) is not helpful for review.

Lib/javascript/v8/javascriptinit.swg
dot-asm: > It would be cleaner to `# pragma GCC diagnostic push` before this,

Rationale behind not making it more elaborate than necessary was that it's **known** to be the very end of the compilation unit.

> Is this needed for clang too?

Not that I could tell.

Lib/javascript/jsc/javascriptstrings.swg
dot-asm: https://github.com/swig/swig/pull/1973#issuecomment-830243478 and https://github.com/swig/swig/pull/1973#issuecomment-846629495. And please be specific.

Lib/javascript/v8/javascriptinit.swg
dot-asm: > It would also be good to note in a comment the node version at which this became consistent so that when SWIG

Commentary says "**all** supported versions." And as far as I can see the problem persists with latest version.

Lib/javascript/v8/javascriptrun.swg
dot-asm: Fixed.

Lib/javascript/v8/javascriptrun.swg
dot-asm: > Why do we always define `clazz_templ` when it seems it's only needed for V8 < 8?

This seems to be wrong assessment. clazz_templ is used whenever `v8_swig_types` is NULL, but it's not like it's NULL when V8<8.

dot-asm:
In its core this resolves real-life hard SEGV crashes at either module load by multiple workers or a worker termination times.

dot-asm:
... and Merry Christmas to you too.

Doc/Manual/Javascript.html
```diff
@@ -89,6 +89,8 @@ <H3><a name="Javascript_running_swig">28.2.1 Running SWIG</a></H3>
 <pre>
 $ swig -c++ -javascript -jsc example.i</pre>
 </div>
+<p>Unlike JavaScriptCore, V8 is a C++ API, and thus, the generated modules
+must be generated and compiled as C++.</p>
```
ojwb, line 93: I think this is potentially misleading, as "must be generated [,,,] as C++" seems to suggest you need to use `swig -c++` which is wrong.  Try the `simple` example:

```
$ cd Examples/javascript/simple
$ make --no-print-directory
make -f ../../Makefile SRCDIR='' CXXSRCS='' \
SRCS='example.c' SWIG_LIB_DIR='../../../Lib' SWIGEXE='../../../swig' \
SWIGOPT='-DV8_VERSION=0x031110' TARGET='example' INTERFACE='example.i' javascript_wrapper
env SWIG_LIB=../../../Lib  ../../../swig -javascript -DV8_VERSION=0x031110 -node -o example_wrap.cxx example.i
make -f ../../Makefile SRCDIR='' CXXSRCS='' \
SRCS='example.c' SWIG_LIB_DIR='../../../Lib' SWIGEXE='../../../swig' \
SWIGOPT='-DV8_VERSION=0x031110' TARGET='example' INTERFACE='example.i' javascript_build
sed -e 's|$srcdir|./|g' binding.gyp.in > binding.gyp
MAKEFLAGS= node-gyp --loglevel=silent configure build 1>>/dev/null
make -f ../../Makefile SRCDIR='' TARGET='example' javascript_run
env NODE_PATH=$PWD:  node runme.js 
GCD of x and y is=21
Global variable Foo=3
Variable Foo changed to=3.1415926
```

Note no `-c++` on the SWIG command line (SWIG did use to use `-c++` here when you started work on this, but I fixed that in d3b17437b12625bd31118f40e06a3bb28b6c5fda).

If "generated" is trying to say you need to call the result `.cxx` or similar, that's not accurate either as you could let SWIG generate `example_wrap.c` and use `-xc++` on the compiler command like (with GCC and clang at least).

The key point here is you need to **compile** the generated wrapper as C++ for V8, which we actually already say below.

Lib/javascript/jsc/javascriptstrings.swg
ojwb: From the linked comments:
> If it's considered too messy, then please state so. If it would be appropriate to break it down to separate requests, the please state so.

OK, I think this part would be better broken out as a separate change as it doesn't seem to be related to "isolate-aware init".  I would struggle to be more specific as I lack information on what's intended, as all I seem to have to go on is the (rather non-specific!) commit message:

```
Lib/javascript/*/javascriptstrings.swg: fix logical errors.

And improve readability.
````

File: Lib/javascript/v8/javascriptinit.swg (hunk 8 above)
Comment (ojwb) [reply Lib/javascript/v8/javascriptinit.swg:174]: > Rationale behind not making it more elaborate than necessary was that it's known to be the very end of the compilation unit.

The `post-init` goes after this (it's just empty by default):

```
$ git diff
diff --git a/Examples/javascript/simple/example.i b/Examples/javascript/simple/example.i
index 24093b9b..35c1ab8d 100644
--- a/Examples/javascript/simple/example.i
+++ b/Examples/javascript/simple/example.i
@@ -1,6 +1,9 @@
 /* File : example.i */
 %module example
 
+%insert("post-init") {
+// THIS GOES LAST
+}
 %inline %{
 extern int    gcd(int x, int y);
 extern double Foo;
$ make -s -C Examples/javascript/simple
GCD of x and y is=21
Global variable Foo=3
Variable Foo changed to=3.1415926
$ tail Examples/javascript/simple/example_wrap.cxx 
#if defined(BUILDING_NODE_EXTENSION)
#if (NODE_MODULE_VERSION < 64)
NODE_MODULE(example, example_initialize)
#else
NODE_MODULE_CONTEXT_AWARE(example, example_initialize)
#endif
#endif

// THIS GOES LAST

```

> Commentary says "**all** supported versions." And as far as I can see the problem persists with latest version.

Oh, I see - then I'd suggest rewording as "NODE_MODULE_INIT is not self-consistent in any supported Node.js version" then - I interpreted it as "it's not the case that: NODE_MODULE_INIT is self-consistent in all supported Node.js versions" (but it is in some)...

File: Doc/Manual/Javascript.html (hunk 11 above)
Comment (dot-asm) [reply Doc/Manual/Javascript.html]: Fixed.

File: Lib/javascript/jsc/javascriptstrings.swg (hunk 10 above)
Comment (dot-asm) [reply Lib/javascript/jsc/javascriptstrings.swg]: Removed the commit in question.

File: Lib/javascript/v8/javascriptinit.swg (hunk 8 above)
Comment (dot-asm) [reply Lib/javascript/v8/javascriptinit.swg:175]: Added push/pop.

File: Doc/Manual/Javascript.html (hunk 11 above)
Comment (ojwb) [reply Doc/Manual/Javascript.html]: Thanks.

Just FYI, I remembered b508598a42b80f9e63f6ebbaf04fdbfee30ea2a2 - before that you would need to run swig with `-c++` when generating wrappers for node (or the undocumented `-c++out` option, or hack around the problem like our testsuite did by running swig with `-v8 -DBUILDING_NODE_EXTENSION=1` instead of `-node`).  So your original change here probably was essentially correct when it was written.

**Comment by mmomtchev** [comment]
@dot-asm I have added a great number of JS run tests in the last few weeks, so you will probably need to rebase and rerun all the tests

Also, why do you shunt `node-gyp` on most platforms and do manual compilation? You seem to be looking for Node.js' original sources instead of the header files as they are being distributed if you install a binary?

**Comment by mmomtchev** [comment]
@dot-asm I will also suggest that in your worker run tests, you require the SWIG shared library in the main thread - maybe eventually as an additional test.

Because, currently, those tests are very random - you spawn two threads, they load/unload the shared library and given that most SWIG run tests are very fast - you will end up with a run test that simply sequentially initializes SWIG twice. Node.js will unload the shared library if the last thread that is using it exits. This is an interesting test in itself, but if you require the shared library in the main thread, you will guarantee that your SWIG init logic will run in both modes - first initialization and subsequent initialization.

In fact, if you always spawn only one thread at a time and you require the SWIG shared library in the worker thread - even the current implementation that is not isolate-aware will work.

**Comment by dot-asm** [comment]
> Also, why do you shunt `node-gyp` on most platforms and do manual compilation?

This is done in the "not to be merged!!!" commit **exclusively** in order to make the tests execute faster. So that I don't have to spend too much time waiting for the results. Again, the last commit is "not to be merged!!!"

> In fact, if you always spawn only one thread at a time and you require the SWIG shared library in the worker thread - even the current implementation that is not isolate-aware will work.

No, it won't. I caught SEGV even in these tests. You should be able to check it yourself by reordering the commits. While the workers are spawned in succession, they are executed in parallel. Create a script with
```
console.log("start");
setTimeout(function() { console.log("stop"); }, 1500);
```
And pass it as argument to the Examples/test-suite/javascript/node_workers.js.

> most SWIG run tests are very fast

While this is a legitimate concern, the tests' outcome appears to be conclusive. I've added `-Wl,-znodelete` to the node-compile script so that the shared library is never unmapped and the tests pass. Apparently JS initialization is not as fast as merely starting workers.

> I will also suggest that in your worker run tests, you require the SWIG shared library in the main thread

Problem with that of course is that I don't know which one to load. Well, one can try to infer the name... Like stripping `_runme.js` suffix?

mmomtchev:
> Problem with that of course is that I don't know which one to load. Well, one can try to infer the name... Like stripping `_runme.js` suffix?

Yes, or simply `require` the `_runme.js`, this should also work - by running the test in the main thread.

dot-asm:
> > Problem with that of course is that I don't know which one to load. Well, one can try to infer the name... Like stripping `_runme.js` suffix?
> 
> Yes, or simply `require` the `_runme.js`, this should also work - by running the test in the main thread.

`require(process.argv[2])` didn't work, so I required `basename(process.argv[2], "_runme.js")`.

dot-asm:
> `require(process.argv[2])` didn't work

I've rearranged the script in question, because the original way, having individual workers "require" the module, arguably reflects the potential reality better. Indeed, users are more likely to have a dispatcher script that spawns workers without knowledge of their inner workings. Again, the original script did catch the addressed problem. But to address the concern I've "required" the module in the exit handler for the 2nd thread, so that both usage patterns get exercised.

PR #1974, opened 2021-03-18
php: Fix char* typecheck typemap to accept Null

### LINKED ISSUE #1655: Multiple signature error; C++ to PHP
I have reduced my app to a simple scenario to demonstrate the problem. I use swig 3.0.12. The wrong signature is chosen.

```
class NameSet
{
};

class FilterSet
{
};

// An example class with two construction signatures.
class Example
{ public:
    Example(const char *, const NameSet * = 0, bool=false);
    Example(FilterSet &, const char * = 0, const NameSet * = 0, bool=false);
};

// Simply stream out the constructor signature.
Example::Example(const char *, const NameSet *, bool)
{
  cout << "Example(const char *, const NameSet *, bool)\n";
}


// Simply stream out the constructor signature.
Example::Example(FilterSet &, const char *, const NameSet *, bool)
{
  cout << "Example(FilterSet &, const char *, const NameSet *, bool)\n";
}
```
C++ main works as expected, printing 

> Example(FilterSet &, const char *, const NameSet *, bool)

```
int main(void)
{
  FilterSet fs;
  NameSet ns;
  Example(fs, nullptr, &ns);
  return 0;
}
```

However, the PHP produces

> PHP Recoverable fatal error:  Object of class FilterSet could not be converted to string in /a/devuser/cjs/wrk/prg/test/swig/swig/php/lib/rhel-7-x86_64/exlib.php on line 113

```
  $fs = new FilterSet;
  $ns = new NameSet;
  $ex = new Example($fs, null, $ns);
```

The reason can be seen in the generated C++ wrapper code. The second argument of the second signature is not a IS_STRING (it is a IS_NULL), so construction jumps to the first signature. See following snippet.

```
  if (argc == 3) {
    int _v = 0;
    {
      {
        void *tmp;
        _v = (SWIG_ConvertPtr(&argv[0], (void**)&tmp, SWIGTYPE_p_FilterSet, 0) >= 0);
      }
    }
    if (!_v) goto check_5;
    {
      _v = (Z_TYPE(argv[1]) == IS_STRING); 
    }
    if (!_v) goto check_5;
    {
      {
        void *tmp;
        _v = (SWIG_ConvertPtr(&argv[2], (void**)&tmp, SWIGTYPE_p_NameSet, 0) >= 0);
      }
    }
    if (!_v) goto check_5;
    _wrap_new_Example__SWIG_4(INTERNAL_FUNCTION_PARAM_PASSTHRU); return;
  }
check_5:
  
  if (argc == 3) {
    _wrap_new_Example__SWIG_0(INTERNAL_FUNCTION_PARAM_PASSTHRU); return;
  }
```

ojwb:
The corresponding in typemap already does.

Fixes #1655, reported by CJSlominski.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -8,6 +8,10 @@ Version 4.1.0 (in progress)
 ===========================
 
 2021-03-18: olly
+	    #1655 [PHP] Fix char* typecheck typemap to accept PHP Null like the
+	    corresponding in typemap does.
+
+2021-03-18: olly
 	    #1900, #1905 [PHP] Fix wrapping of overloaded directed methods with
 	    non-void return.
 
--- a/Examples/test-suite/overload_polymorphic.i
+++ b/Examples/test-suite/overload_polymorphic.i
@@ -23,4 +23,7 @@ class Unknown;
 int test2(Unknown* unknown) { return 0; }
 int test2(Base* base) { return 1; }
 
+int test3(const char*, const Base* = 0, bool = false) { return 0; }
+int test3(Base&, const char* = 0, const Base* = 0, bool = false) { return 1; }
+
 %}
--- a/Examples/test-suite/php/overload_polymorphic_runme.php
+++ b/Examples/test-suite/php/overload_polymorphic_runme.php
@@ -1,0 +1,14 @@
+<?php
+
+require "tests.php";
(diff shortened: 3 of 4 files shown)


ojwb:
CI failure seems an unrelated system issue.

PR #1975, opened 2021-03-18
Improve PHP object creation

### LINKED ISSUE #1527: Segfaults in PHP 7.2 when wrapping resources to objects
Hi, there seems to be a bug in SWIG_SetPointerZval(..., 2) causing segfaults in PHP7.2+, in our project later when releasing the wrapped object from memory.

From PHP sources:
```
/* This function requires 'properties' to contain all props declared in the     
 * class and all props being public. If only a subset is given or the class     
 * has protected members then you need to merge the properties separately by    
 * calling zend_merge_properties(). */                                          
ZEND_API int _object_and_properties_init(zval *arg, zend_class_entry *class_type, HashTable *properties ZEND_FILE_LINE_DC) /* {{{ */
```

This patch helped:
```
<       ALLOC_HASHTABLE(ht);
<       zend_hash_init(ht, 1, NULL, NULL, 0);
<       zend_hash_str_update(ht, "_cPtr", sizeof("_cPtr") - 1, &resource);
<       object_and_properties_init(z, ce, ht);
---
>       object_init_ex(z, ce);
>       zend_update_property(ce, z, "_cPtr", sizeof("_cPtr") - 1, &resource);
```

ojwb:
Reportedly the code we were using in the directorin case gives segfaults
in PHP 7.2 and later, but we've been unable to reproduce these, but the
new approach is also simpler and should be bit faster too.

Fixes #1527

### DIFF
--- a/Lib/php/phprun.swg
+++ b/Lib/php/phprun.swg
@@ -90,15 +90,13 @@ SWIG_SetPointerZval(zval *z, void *ptr, swig_type_info *type, int newobject) {
     } else {
       /*
        * Wrap the resource in an object, the resource will be accessible
-       * via the "_cPtr" member. This is currently only used by
+       * via the "_cPtr" property. This code path is currently only used by
        * directorin typemaps.
        */
-      zval resource;
       zend_class_entry *ce = NULL;
       const char *type_name = type->name+3; /* +3 so: _p_Foo -> Foo */
       size_t type_name_len;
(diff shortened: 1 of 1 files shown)


ojwb:
Rebased onto master and added a CHANGES.CURRENT entry in 71475b0af9677deeaf6fe55c0c5f53fec9f730d2.

Issue #1603, opened 2019-07-30
SWIG4 not support template <template<typename, typename> class Sequence, typename T>


ThibaultLesaunier:
Hi guys, 

All is in the title. SWIG3 works fine with 

template <template<typename, typename> class Sequence, typename T>

and SWIG4 has a problem with it.

Any help appreciate
Thanks

wsfulton:
Can you please provide some valid C++ to demonstrate. Which version of SWIG 3, worked?

ThibaultLesaunier:
Hi,

I'm just declaring a method C++ like : 

template <template<typename, typename> class Sequence, typename T>
XPCFErrorCode declarePropertySequence(const char * name, Sequence<T, ::std::allocator<T>> & wrappedValue, IProperty::AccessSpecifier iotype = IProperty::AccessSpecifier::IProperty_INOUT);

and SWIG has a problem on "template" line.

SWIG-3.0.12 was use before the 4.0, on this file and no problem with it.

Thanks

rokups:
A minimal testcase would be most helpful. You can make `test.i` module and include c++ code that reproduces issue in `%inline %{ /*here*/ %}`.

ThibaultLesaunier:
My code is composed of many Interfaces and Class. How can I test just this part of code without all import xD ?

rokups:
I mean a synthetic testcase that reproduces same issue but isnt code from your project.

ThibaultLesaunier:
**test.i**

```
%inline %{ 
	template <template<typename, typename> class Sequence, typename T>
    XPCFErrorCode declarePropertySequence(const char * name, Sequence<T, ::std::allocator<T>> & wrappedValue, IProperty::AccessSpecifier iotype = IProperty::AccessSpecifier::IProperty_INOUT); 
	%}
```

> swig -csharp -c++ test.i

==> Error : Syntax error input(1)

Same error as my code.

jcsantamaria:
Huh? I get the same error using SWIG 3.0.12 on this simple test case.

wsfulton:
Please put the code snippet within a `%{ %}` block to make sure the C++ compiler compiles it without error. Then change it to a `%inline %{ %}` block. If SWIG gives an error, you then know that SWIG has a bug because the C++ compiler accepted it. If the code doesn't compile with a C++ compiler, please close the ticket.

jcsantamaria:
I use the same code snippet provided by @ThibaultLesaunier.  

[...]

PR #1977, opened 2021-03-18
Fix handling of template template parameters (#1603)

goto40:
This PR aims to solve #1603 :

**Summary**

 * The issue #1603 describes problems with template template parameters.
 * It seems that a erroneous comparison in "Source/Swig/stype.c" possibly prevents the template types to be correctly replaced (see diff).
 * The fix is commented as such in "Source/Swig/stype.c": `POSSIBLE FIX....`. This comment needs to be removed, if we agree on the proposed changes.

**I would like to discuss with the core development team the proposed patch in this PR...**

Note: This PR also introduces some unittests - these tests may not match the test philosophy of the SWIG project (I did not check the tests indicated in the CMakeLists.txt). I could remove these tests once we find consensus about the possible fix (if the proposed fix really solves the problem)...

 * The test "Code_with_simple_templates" is a unittest describing the correct functionality (also passes without the patch described in this PR)
 * The test "Code_with_template_templates" is a unittest describing the problematic case (which does not pass without the patch)

The following example is used in the problematic testcase "Code_with_template_templates":
```
        template<class T>
        struct Container1 { 
            T x;
        };
        template<class U>
        struct Container2 { 
            U x;
        };
        template<class BaseT, class TemplateTemplateT>
        struct TestStruct { 
            TemplateTemplateT<BaseT> x;
        };
        %template(IntTestStruct) TestStruct<int, Container1>;
        %template(FloatTestStruct) TestStruct<float, Container2>;
```

 * The swig system yields an attribute `FloatTestStruct::x` of type `TemplateTemplateT<float>` **without the patch (not ok)**.
 * The swig system yields an attribute `FloatTestStruct::x` of type `Container2<float>` **with the patch (ok)**.

I also did a quick interactive test with a simple example generating java files (which I could rework and share).


wsfulton:
@goto40 excellent, thanks for this contribution.

Regarding testing, we have a rather large test-suite which is documented at http://swig.org/Doc4.0/Extending.html#Extending_test_suite. You'd need to add your test case into a new test interface file. I think template_template_parameters.i is one of the few template template parameters testcases we have. Your commit may even fix the commented problem in that one, I havn't checked yet. If not I suggest adding another similarly named testcase.

I'll check the fix in the coming days. You may be interested to know that the main check I will do is run the test-suite in partialcheck mode for a couple of target languages and examine the output diffs (in the `Examples/test-suite/<lang>` directory), before and after the change.

I don't envisage adding a new test framework. btw, we use the autotools as the primary development platform. CMake has only just been added for building the executable on Windows really; porting the entire test-suite to CMake would be a massive task.

goto40:
Great. I am looking forward to see if it works for you...

goto40:
I added an extra check to see if a '<' is following the matched string.
Maybe there is a better Doh*-Function for that... (see last commit)

goto40:
I just play with the testcase `template_template_parameters`.
 * the partial check seems plausible to me (please also have a look).
 * I just adapt the testcase and the java example (`template_template_parameters_runme.java`): with this I try to produce an error w/o the patch which is resolved by the patch...

Is there a way to run just one test case? I did not figure out how to do it (I use `make check-java-test-suite` to run the non-partial tests)

vadz:
> Is there a way to run just one test case? I did not figure out how to do it (I use `make check-java-test-suite` to run the non-partial tests for java)

You can do `make -C Examples/test-suite/java TESTNAME.cpptest` for this.

goto40:
* I adapted the unittests for template template parameters `template_template_parameters.i`.
 * I added a python runme testcase
 * I adapated the java runme testcase
 * I removed the new unittests (as suggested by first review round)
 * I reworked the comment (and removed `POSSIBLE FIX`)

goto40:
Ready for a next round of review.

wsfulton (commented):
FYI, I made these changes.

PR #1978, opened 2021-03-19
Php8 support

ojwb:


ojwb:
This seems to work very well - for example, I have successfully built Xapian's PHP bindings with it and their testsuite passes (tested using PHP 7.4 and 8.0).

The remaining issue I'm aware of here is the hack I've applied to `Examples/test-suite/director_overload2.i` which is really an existing issue with SWIG's PHP backend, but PHP 7 (7.4 at least) gives a warning, while with PHP 8 this warning has become an error.

I think the fix would be to check the parent class versions of a virtual method for default parameters and apply those for subclass versions of the method.  Not sure if this is already handled elsewhere in SWIG, or if it's specific to SWIG/PHP's current two level wrapping approach.

But given this is an existing problem and the CI passes (which means the patch still works for PHP 7.0-7.4), I'm don't think it makes sense to block merging this on fixing that - instead I'll open a new issue to track it.

I'm going to do a final self-review pass over the patch first, and would welcome others looking over it too.  If interested people can test their SWIG-generated bindings using this branch with both PHP 7.x and PHP 8.0 that would be helpful.

wsfulton:
Nice one @ojwb, I vote to merge this in.

neoclust:
thanks. I added the patch in mageia. It builds OK. I will ask for tests

Issue #1529, opened 2019-05-03
[PHP] Move to a modern output


ferdynator:
PHP has come great lengths in the last years. The currently generated swig output does not meet best-practices, uses [deprecated functions](https://php.net/manual/en/function.dl.php)
Namespaces are not supported, [type-hinting](https://www.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration) and [function return types](https://www.php.net/manual/en/functions.returning-values.php). Since swig dropped support for older PHP versions and the default PHP version is `7` many things could be improved.

I would like to implement (or help implementing) some of these features but don't have the most experience with swig. I've already added a PR #1528 to remove a closing tag, because that is considered bad practice. What would be the best first / next steps to take here?

Some languages (as C#) have `namespace` support which is a feature I would love to implement for PHP. Modern PHP applications use a JAVA-style _1-class-per-file_ setup. Does it make sense to copy from the [java implementation](https://github.com/swig/swig/blob/master/Source/Modules/java.cxx)? There are [some FIXMEs](https://github.com/swig/swig/blob/master/Source/Modules/php.cxx#L15) in the php code. Should they be addressed first? 

It seems like PHP is rarely used and thus the generation is quite out of date too. But with upcoming [JIT](https://wiki.php.net/rfc/jit) PHP is getting more interesting for CPU heavy load. And having swig generate nice PHP modules out of the box would be awesome :)

ojwb:
The answer to most of these is just that nobody's implemented it yet.

For namespace support there's already a ticket open: #69

The use of `dl()` is just there as a fallback to try if the C glue module isn't already loaded.  I've wondered about removing it entirely, but worried it would break things for some people running code with CLI PHP.  I guess if we're going to, now's a fairly good time (doing it for 4.0.0 would have been better though).

I don't think moving to one PHP file per class is helpful though - it makes sense for code humans need to maintain, but for generated code it doesn't really help, and having everything in one file is actually more convenient to deal with.  And getting the work from the Nihal's GSoC project merged would change us to using the PHP C API to create classes instead of generating PHP code to define classes which wrap flat functions, and that would make this point moot.

ferdynator:
Are you talking about #1092 ? What can be done to get this merged? I'll be looking into namespaces later this week. Also I can open a PR for the removal of `dl()`. You can merge it any time it fits then.

ojwb:
[...]

PR #1982, opened 2021-03-26
WIP: Create PHP classes via C API

ojwb:
This work is from nihaln's GSoC project from 2017 - this branch is what we were using to integrate reviewed changes.

I've merged in one PR from nihaln which didn't get merged at the time, merged master to the branch and fixed some of the issues that revealed.

~~There's still a bug (uncovered by two new testcases on master) which I've not fixed yet, and~~ there's quite a lot that really needs tidying up - some things aren't done in clean ways, ~~and pretty much all the old machinery is still present~~.

# Current status as of f2009ef6813108a067b7b046c32554890ce7a615

All testcases pass without any PHP warnings, notices, etc for PHP 7.0-7.4 and 8.0.

## To Do

* Hook up mechanism to specify how to wrap flat functions
* Don't generate (or optionally generate?) `foo_get()` and `foo_set()` methods in favour of just having a `foo` property on the wrapped object to wrap a `foo` public class/struct member
* Improved test coverage by adding more `_runme.php`
* Use object handlers instead of magic `__get`, `__set`, `__isset`?


ojwb:
It looks like we're down to two testcase failures due to the new code not actually checking the type of pointers when resolving overloads (which I'm currently working on fixing), plus a PHP 8 specific issue which I've not investigated but sounds like stricter handling of `protected` in PHP.

There are also some warnings about overloads with different signatures which C++ allows but PHP doesn't seem to like.

I also still need to eliminate the remaining place where we're currently generating PHP wrapper code - this is for flat functions which currently get wrapped in an `abstract class` in lieu of namespacing.  This approach predates PHP gaining proper namespace support, and for modern PHP these should probably instead be in a namespace named after the module.  However, for compatibility we probably also need to support the current structure as an option.

ojwb:
> I also still need to eliminate the remaining place where we're currently generating PHP wrapper code - this is for flat functions which currently get wrapped in an abstract class in lieu of namespacing. This approach predates PHP gaining proper namespace support, and for modern PHP these should probably instead be in a namespace named after the module. However, for compatibility we probably also need to support the current structure as an option.

Now done - the generated `.php` file now only contains the machinery to call `dl()` if the extension isn't already loaded - ideally we should also remove that, but we need to sort out how to run the test suite without it (problem discussed in #1529.)

I'm going to try to completely eliminate the `_cPtr` property and wrapping object pointers as PHP resources next.  Currently we have two different ways to wrap objects, which complicates the typemaps and I found that was getting in the way of adding the missing type checks on pointer casts.

ojwb:
> I'm going to try to completely eliminate the _cPtr property and wrapping object pointers as PHP resources next. Currently we have two different ways to wrap objects, which complicates the typemaps and I found that was getting in the way of adding the missing type checks on pointer casts.

Now done.

This introduces two additional testcase failures:

* import_nomodule: I don't really understand this testcase - @wsfulton why do we aim to support this "violation of the type system"?  For now I've just added PHP to the list of languages which don't support this.
* li_boost_shared_ptr_bits: Generated code fails to compile with:
  ```
  li_boost_shared_ptr_bits_wrap.cxx:5150:23: error: ‘SWIGTYPE_p_HiddenPrivateDestructor’ was not declared in this scope; did you mean ‘SWIGTYPE_HiddenPrivateDestructor_ce’?
   5150 |   SWIG_TypeClientData(SWIGTYPE_p_HiddenPrivateDestructor,SWIGTYPE_HiddenPrivateDestructor_ce);
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                       SWIGTYPE_HiddenPrivateDestructor_ce
  ```
  Not sure what's up here - i guess I'm not quite using the right thing to register the type data against, but it only seems to fail in this one corner case (or perhaps two, if that's why import_nomodule now fails.)  PHP doesn't currently have support for shared_ptr so the testcase failure seems a minor issue, except because the code doesn't compile the testsuite fails.

ojwb:
I've resolved the issue that was causing the code generated for `li_boost_shared_ptr_bits` to fail to compile.

So the big remaining issue here is the two testcase failures (overload_null and overload_polymorphic) due to the new code not actually checking the type of pointers when resolving overloads.  Hopefully that'll be easier to fix now there's only one way we're wrapping classes.

ojwb:
I've now implemented type checking when resolving overloads, and all the tests pass locally with PHP 7.4.

The older travis logs show a PHP 8 specific issue which I've not investigated but sounds like stricter handling of `protected` in PHP.

There are also some warnings about overloads with different signatures which C++ allows but PHP doesn't seem to like.

I also still need to eliminate the remaining place where we're currently generating PHP wrapper code - this is for flat functions which currently get wrapped in an `abstract class` in lieu of namespacing.  This approach predates PHP gaining proper namespace support, and for modern PHP these should probably instead be in a namespace named after the module.  However, for compatibility we probably also need to support the current structure as an option.  I've done some work towards this.

Some of the new code on the branch could really do with cleaning up too, though I don't think I want to block merging on that since merging this would remove a lot of clunky code which has known but hard-to-address issues.

I don't think it's production-ready yet, but I'd definitely encourage people to start testing this on their bindings to help shake out any problems the testsuite doesn't currently have coverage for.

ojwb:
Currently there are test failures for all the tested PHP versions except 7.4:

# 7.0

```
checking php testcase arrays_global (with run test)
Unknown:0: Unknown list entry type (653547009)
```

```
checking php testcase li_carrays (with run test)
Unknown:0: Unknown list entry type (1018514433)
```

# 7.1

```
checking php testcase evil_diamond_prop (with run test)
../../../Examples/Makefile:1079: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault (core dumped)
```

```
checking php testcase li_carrays_cpp (with run test)
Unknown:0: Unknown list entry type (985622785)
```

```
checking php testcase li_carrays (with run test)
Unknown:0: Unknown list entry type (1459972353)
```

# 7.2

```
checking php testcase evil_diamond_prop (with run test)
../../../Examples/Makefile:1079: recipe for target 'php_run' failed
```

```
checking php testcase li_carrays_cpp (with run test)
Unknown:0: Unknown list entry type (285863169)
Warning: Unknown list entry type (285863169) in Unknown on line 0
```

# 7.3

```
checking php testcase evil_diamond_prop (with run test)
../../../Examples/Makefile:1079: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault (core dumped)
```

# 8.0

```
checking php testcase director_nested (with run test)
/home/travis/build/swig/swig/Examples/test-suite/php/director_nested_runme.php:24: Invalid callback A::do_advance, cannot access protected method A::do_advance()
```

```
checking php testcase director_protected (with run test)
/home/travis/build/swig/swig/Examples/test-suite/php/director_protected_runme.php:37: Invalid callback FooBar::ping, cannot access protected method FooBar::ping()
```

```
checking php testcase overload_null (with run test)
../../../Examples/Makefile:1079: recipe for target 'php_run' failed
make[2]: *** [php_run] Segmentation fault (core dumped)
```

ojwb:
`-prefix` is mostly ignored by the new branch (it's only used for the fake class name, and none of the actual wrapped classes get prefixed names.)

Now `-prefix` exists because PHP didn't used to have namespaces, dropping support now is unhelpful for projects using it as they'll likely want to plan a migration to using PHP namespaces (and we also still need to actually implement support for PHP namespaces.)

ojwb:
I've fixed `overload_null` for PHP8.

I've eliminated the `$needNewFlow` mechanism from all `in` typemaps, but I think it really needs to be fully eliminated.

The single remaining use in `directorout` typemaps looks to entirely wrong to me - it essentially seems to paper over an ownership handling bug by turning it into a memory leak.  I wonder if implementing `newmemory` handling would fix this properly, though it appears it worked before for PHP without that.

> There are also some warnings about overloads with different signatures which C++ allows but PHP doesn't seem to like (which are errors with PHP 8.)

These warnings/error have been fixed by my recent changes.

I've also gone through the latest travis logs and started a "current status" in the first comment of the ticket.  It looks like there are probably at most four causes.

ojwb:
Wow, I wasn't expecting it but travis just went green, and appveyor is failing only on two cygwin python builds.

I'm guessing the failures were memory corruption from the wrapping of either global variables or class properties.

I'll disable appveyor on the branch as it doesn't test PHP - then the CI state will actually be useful.

ojwb:
I've made appveyor on the branch just do one python build, which means we'll catch failures to build the new PHP backend code with MSVC, but aren't pointlessly spinning CPUs or having our CI results clouded by unrelated CI failures.  (Obviously before merging this I'll restore the disabled CI builds.)

I've also made the PHP examples fail on any diagnostic as the PHP test-suite already does, and I've enabled the C++11 tests in the CI builds for PHP.

The CI is green, but there are clearly some issues - the most obvious is that one testcase fails under ubsan.

I'm going to hold off merging this for now while I try to sort those out, but I'm definitely targetting this at SWIG 4.1.0.

ojwb:
Pointers are now wrapped as PHP objects instead of resources (as recommended by some more modern PHP internals docs).  This allowed me to easily eliminate more uses of `$needNewFlow`.

Pointer-to-member is still a resource, but not for any good reason.

ojwb:
I've now tracked the director_finalizer failure down to apparently being due to a missing implementation of director-disown for PHP.  I say "apparently" as implementing director-disown for PHP doesn't fully fix the issue.  Looking at the testcase, I don't see how it's meant to work, but it does for Perl5 and Python (testing those with ubsan shows no problems) so I think there must be something about this I'm failing to grasp.  I'll guess I'll try comparing the flow by adding prints to the Perl5 and/or Python versions as well as to PHP.

But anyway I tested git master's director_finalizer with ubsan and PHP 8.0 and I get the same problem, so I don't see this issue as a blocker for merging, though of course I'd really like to resolve it.

ojwb:
For Perl, the director_finalizer testcase seems to rely on `DESTROY` being called twice - first to signal finalisation (with `$final` parameter set to `1`) and then again to actually destroy the object (with `$final` `undef`).  Essentially this is hooking into an existing Perl mechanism.

PHP does have two object handlers which at first seem roughly equivalent to the two `DESTROY` calls, though only one is accessible by writing PHP code (the other is only accessible via the C extension API). However I can't seem to get these to deliver what we want - I think they both fire too late.

The sequence is we create an PHP object of a PHP subclass of a SWIG-wrapped C++ class (with directors enabled), then we director-disown the object (so now the C++ object owns the PHP proxy object via having incremented the reference count), then call to C++ to delete the C++ object, which decrements the reference count on the PHP proxy object.

The testcase wants to be able to run some "finalization" code in response to this, and expects that this code can call a method on the C++ object, so this code needs to happen before the C++ object's destructor completes since then C++ object no longer exists.  Currently we try to achieve this by putting this code in the destructor of the PHP class, but there's a PHP-side reference to the same object (the variable we passed to `deleteFoo()`) so the PHP destructor is never going to get run in the interval it needs to be.

And the other PHP object handler runs after the destructor object handler, which seems to be even more too late.

For Python this seems to work by the director-disown method returning a weak reference to the passed object.  Now PHP does have `WeakReference`, but you have to call `get()` on it to do anything with it so (a) it's not as clean to use as in Python and (b) it doesn't actually seem to help, I suspect because when we call `get()` we presumably create a non-weak reference, which then exists during the call to `deleteFoo()` which we pass it to!

So my conclusion is this testcase isn't really implementable in PHP.

The only solution I can see would be that we invent something specific to SWIG/PHP where (say) a `swig_finalize()` method is called if it's defined on the PHP object when the C++ dtor runs.  I don't think that's a good way to go, at least not without evidence there's actually demand for such a finalizer feature from people using SWIG to wrap C/C++ code for use in PHP.

ojwb:
> Get rid of `.php` wrapper file entirely (there's no longer any automatically-generated actual content)?  What about the `phpcode` pragma?

Now done - the `.php` wrapper is only generated if one of the two pragmas is used which add content to it.

> Locally I see testcases `abstract_signature`, `fvirtual` and `using_extend` fail with errors like `Fatal error: Declaration of abstract_bar::meth($arg1, $arg2) must be compatible with abstract_foo::meth($arg1)` (with PHP 8.0 at least) but they seem to pass cleanly in CI

These seems to have been due to some uncleaned generated files in my source tree, as removing all the generated `.php` files fixed it.

ojwb:
I think the time has come to merge this.

There are still a few "todo" items above, but the code on git master is the same or worse for each of these (it's just that it's more feasible to address them now) and there are also several bugs fixed on the branch.

Once the CI is green I'll do the merge.

My plan is to then start a fresh branch to address these points (working on a branch with the CI pruned back to just PHP + a token msvc build means the CI typically completes in less than an hour rather than taking most of a day.)

PR #1986, opened 2021-04-04
[Python] SWIG_AsCharPtrAndSize: deal with bad_alloc exception

rouault:


ojwb:
Your patch makes all the Python CI builds fail.

These typemaps need to able to compile as both C and C++ (so conditionalise the `try` and `catch` on `#ifdef __cplusplus`, etc).

Also, please don't re-indent existing code around your changes.

rouault:
Thanks for the review.

> These typemaps need to able to compile as both C and C++ (so conditionalise the `try` and `catch` on `#ifdef __cplusplus`, etc).

to avoid this, I've added a ``%new_array_may_be_null()`` macro

> please don't re-indent existing code around your changes.

should be fixed

wsfulton:
All the python tests are still broken.

rouault:
> All the python tests are still broken.

hum, ok, I now manage to run the test suite locally (had a hard time figuring out I had to set CXXFLAGS=-I/usr/include/python3.8) and reproduce the failure. Should be fixed now (hopefully...)

wsfulton:
The current behaviour in SWIG_AsCharPtrAndSize is actually as it should be when using C++. We use `new` all over the place and users who want to handle out of memory are expected to handle it with custom code, most commonly by catching `std::bad_alloc` with the `%except` feature. Where SWIG_AsCharPtrAndSize falls short is using C when malloc fails and returns NULL to indicate out of memory. Also when a SWIG user provides an overloaded new operator that returns NULL for out of memory. Both cases ought to be handled better in this function (and a LOT of other places) in SWIG. In an out of memory situation, a SWIG_MemoryError should be returned (which is mapped to Python's MemoryError). I suggest your patch is modified to replace `%new_array_may_be_null` with `%new_array` and `return SWIG_RuntimeError` is replaced with `return SWIG_MemoryError`.

We don't use nothrow new in SWIG generated code. If we were to do that, it'll be a breaking change, which is fine in a major release and is documented, but it would only make sense if it was done throughout the generated code consistently. That's quite a big job.

rouault:
@wsfulton I've tried to align on your recommendations. I had to introduce specific ``#if defined(__cplusplus)`` to drop a reference on the Python object though.

wsfulton:
Hi, I'm afraid the changes are not what I was trying to suggest. The try catch is something a user will put in using `%exception`, but you've put it in the SWIG library code. Some users compile their code without exception support and so this will break their code.

rouault:
> Hi, I'm afraid the changes are not what I was trying to suggest. The try catch is something a user will put in using `%exception`, but you've put it in the SWIG library code. Some users compile their code without exception support and so this will break their code.

C++ without exception... should we really care about those non-standard uses ? How could I possibly drop the reference on the Python object if not catching a potential exception, before re-raising it ?

vadz:
> How could I possibly drop the reference on the Python object if not catching a potential exception, before re-raising it ?

Just in case this question wasn't rhetorical, this is done using smart pointer class decrementing the ref count in its dtor. This is simpler, more readable and less error-prone than writing explicit try/catch even for C++ users who do use exceptions.

rouault:
> Just in case this question wasn't rhetorical, this is done using smart pointer class decrementing the ref count in its dtor. This is simpler, more readable and less error-prone than writing explicit try/catch even for C++ users who do use exceptions.

no, this wasn't rhetorical, I just forgot about it...  @wsfulton How do you feel about this approach ?

PR #1991, opened 2021-04-12
Eliminate unused Printf args

ojwb:
Code clean-up.

Opening an MR to get CI testing.


ojwb:
3 JS builds failed in CI, but they're already failing on git master so unrelated to this change.

PR #1997, opened 2021-04-20
Testsuite improvements

ojwb:
This enables some testcases that could be run for PHP but weren't being.

While working on this, I noticed some other oddities, such as a double-listed testcase for java, and a few testcases which aren't actually currently referenced for any language so I've also addressed some of those.  There's more, but I don't think I have the time or energy right now for trying to sort out a lot more.

I've tested these locally, but opening a PR so they get tested with more versions without me breaking git master.

I suspect the previously unused `sym` testcase may be redundant with another testcase, but I failed to find one which clearly covered the same case.  If there is one, we could remove it instead.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -418,6 +418,7 @@ CPP_TEST_CASES += \
 	struct_value \
 	swig_exception \
 	symbol_clash \
+	sym \
 	template_arg_replace \
 	template_arg_scope \
 	template_arg_typename \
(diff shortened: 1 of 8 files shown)


ojwb:
The 3 javascript CI failures are already present on master.

ojwb:
Just noticed that `director_nested` is now run twice for PHP.  Will fix before merging.

ojwb:
I folded the two tweaks into the relevant commit and merged to master from the command line - closing.

PR #1998, opened 2021-04-20
swigrun.swg: fix typo

rouault:
Spotted by lintian QA tool used by Debian packaging


ojwb:
Thanks.  There were a lot more instances of "allows to" and "allow to", some in the end user documentation, so I've fixed those plus some other typos I noticed along the way in f198ff0a.

PR #1999, opened 2021-04-20
std::filesystem::path (C++17) <-> pathlib.Path (Python 3.4+) Wrapper

yasamoka:


wsfulton:
Hi @yasamoka. Thanks for this contribution and I apologise it hasn't been commented on earlier. The patch is somewhat out of date now I'm afraid. I appreciate you tried to add in C++17 support to the test-suite, but we have a different way of testing C++17 now. Can you please modify the testcase based on one of the cpp17_xxx.i testcases such [cpp17_hex_floating_literals.i](https://github.com/swig/swig/blob/master/Examples/test-suite/cpp17_hex_floating_literals.i) or [li_std_autoptr.i](https://github.com/swig/swig/blob/master/Examples/test-suite/li_std_auto_ptr.i). The new approach is to use `__cplusplus` to detect c++17 support and to provide an equivalent c++98 solution if the compiler does not support C++17. While this has worked for the current C++17 tests, in your case, there is no simple equivalent, so add in something like the following to the testcase:
```c++
bool is_cplusplus17() {
#if __cplusplus >= 201703L
  return true;
#else
  return false;
#endif
}
and only run the test if this function returns true.

You shouldn't have to modify any of configure.ac nor any Makefiles. Just add a single line into common.mk referencing cpp17_std_filesystem.i.

**Comment by yasamoka** [comment]
Hi @wsfulton, thank you for getting back to me on this.

Shall I rebase to upstream and include only the changes that are related to the implementation and test suite, or shall I create a new pull request and close this one?

Let me know what you prefer. Thanks!

**Comment by wsfulton** [comment]
A rebase is fine in this instance. Thanks.

File: Lib/python/std_filesystem.i [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+}
+
+%typemap(in, fragment="common") std::filesystem::path {
+    if (PyUnicode_Check($input)) {
+        const char* s = PyUnicode_AsUTF8($input);
+        $1 = std::filesystem::path(s);
+    }
+    else if (isPathInstance($input)) {
+        const char * s = pathToStr($input);
+        $1 = std::filesystem::path(s);
```
degasus, line 46: Are you sure that this call will decode UTF-8 on Windows? If I read the specs correct, this is only done for `char8_t`, so only for unsigned char.

Should there be checks in the test-suite which checks for unicode edge cases instead of just foo?

jmarrec:
This got very stale, so I updated the testing and opened a new PR at https://github.com/swig/swig/pull/2639. I suggest closing this one.

ojwb:
Superseded so closing.

PR #2007, opened 2021-04-29
Remove support for $source and $target

### LINKED ISSUE #1984: Finally drop support for $source and $target
I noticed these were being set in the PHP backend but were not actually used by any of the PHP typemaps.  Digging a bit I found they're ancient names for typemap parameters supported only for compatibility with user typemaps from before some change long ago - SWIG has been issuing a deprecation warning for uses of them (with a testcase even!) since before "The great merge" in late 2002, and `CHANGES` says they were officially deprecated in late 2001 (so over 19 years and counting).

I think we should drop support for them in 4.1.0 - they make it harder to maintain a SWIG backend or to write a new one, and if anyone has really ignored a deprecation warning about their custom SWIG typemap for over 18 years then they shouldn't be surprised by it stopping working in a new release series.

They don't seem to be entirely consistently supported - e.g. SWIG/D never sets `$source` while SWIG/Go  only sets it in one case.

I'm also dubious they'll actually still work everywhere in languages SWIG has supported for long after this much time given we don't test they actually work anywhere, only that mentioning them in a dummy typemap triggers a deprecation warning.   E.g. on the gsoc2017-php7-classes-via-c-api branch I see:

```
        Replaceall(tm, "$source", &source);
        Replaceall(tm, "$target", ln);
        Replaceall(tm, "$input", source);
```

I'm pretty sure the `&` on the first is wrong...

ojwb:
These were officially deprecated in 2001, and attempts to use them have
resulted in a warning (including a pointer to what to update them to)
for most if not all of that time.

Fixes #1984


ojwb:
I've added a CHANGES.current entry, squashed and pushed to master as 9ddc9dceb7f3fb6f9f58239de0138b32b7c2dff6.

Issue #1979, opened 2021-03-23
javascript memory management


dot-asm:
While working on #1973, I've attempted to keep `Persistent<Template>`-s in an array allocated with C++ new[]. And trouble was that no matter what I tried, the WeakCallback [that would delete[] the array] was never called. Formally speaking one can't actually complain, because V8 gives "no guarantee as to when or even *if* the callback is invoked." Solution was to allocate a buffer from the Javascript memory heap, a.k.a. `Isolate`, and use it as opaque storage for the C++ array. So that when the "isolate" is teared down, nothing leaks. Because it will be teared down in its entirety. Question is if it's possible to extend this approach even to application code. Of course it's not possible to extend it to **all** applications, but I reckon that it's possible to use it with some, if not most. In essence the limitation would be that application structures may not have pointers to other structures allocated from "isolate," for example no lists would be possible. But note that pure data would work, as well as pointers to objects outside "isolate," even vtables... So how about
```
void* operator new(size_t sz, v8::Isolate* isolate,
                   v8::Local<v8::Uint32Array>& ret)
{
    sz = (sz+3)&-4;
    v8::Local<v8::ArrayBuffer> ab = v8::ArrayBuffer::New(isolate, sz);
    ret = v8::Uint32Array::New(ab, 0, sz/4);
#if (V8_MAJOR_VERSION-0) >= 8
    return ab->GetBackingStore()->Data();
#else
    return ab->GetContents().Data();
#endif
}
```
then `Foo* cobj = new (isolate, jsCptr) Foo(...)` and finally "wire" the `jsCptr` to the target Javascript object's as a hidden property. Thoughts?

As already mentioned, this is unsuitable for some applications, hence it would have to be optional. Most flexible option would be if one could specify it on a per-class basis via a %directive, but having it as command-line option that would affect code generation for **all** classes might be a start too.

I suppose first question would be what is the best place and way to inject the placement new? It might be tempting to add language-specific typemap, something similar to `javaout`, but for javascript, but I can imagine that the approach might be useful in multiple languages, maybe all GC-based...

dot-asm:
[...]

PR #2011, opened 2021-05-01
[javascript][v8] SWIG_V8_VERSION generation method corrected.

nxhack:
When "V8_MAJOR_VERSION" is 4 or higher, this fix is required if the definition of "SWIG_V8_VERSION" is as follows.

|node.js|V8|SWIG_V8_VERSION|
|:---:|:---:|:---:|
|16.0.0|9.0.257.17|0x0900|
|14.16.1|8.4.371.19|0x0804|
|12.22.1|7.8.279.23|0x0708|
|10.24.1|6.8.275.32|0x0608|
|8.17.0|6.2.414.78|0x0602|
|6.17.1|5.1.281.111|0x0501|
|4.9.1|4.5.103.53|0x0405|

Signed-off-by: Hirokazu MORIKAWA <morikw2@gmail.com>

### DIFF
--- a/Lib/javascript/v8/javascriptruntime.swg
+++ b/Lib/javascript/v8/javascriptruntime.swg
@@ -58,7 +58,10 @@
 
 #if defined(V8_MAJOR_VERSION) && defined(V8_MINOR_VERSION)
 #undef SWIG_V8_VERSION
-#define SWIG_V8_VERSION (V8_MAJOR_VERSION * 256 + V8_MINOR_VERSION)
+#define SWIG_V8_VERSION ((V8_MAJOR_VERSION / 10) * 4096 + \
+                         (V8_MAJOR_VERSION % 10) * 256 + \
+                         (V8_MINOR_VERSION / 10) * 16 + \
+                         (V8_MINOR_VERSION % 10))
 #endif
 
 #include <errno.h>


wsfulton:
> When "V8_MAJOR_VERSION" is 4 or higher,

I think you mean 10 or higher.

I guess node 18 is going to define v8 as version 10?

Is it just me, but the SWIG_V8_VERSION definition is not clear and rather crazy. If V8_MAJOR_VERSION is defined (>4), then SWIG_V8_VERSION is a 4 hex number, like 0x04321 otherwise it is a 6 hex number, like 0x034567. Thoughts on 'fixing' or improving this with another version number? This could certainly do with some further clarification for this dual mode version number:

https://github.com/swig/swig/blob/9ddc9dceb7f3fb6f9f58239de0138b32b7c2dff6/Lib/javascript/v8/javascriptruntime.swg#L6-L23

nxhack:
>I think you mean 10 or higher.
I guess node 18 is going to define v8 as version 10?

This is a response to future digit increases.

The versioning rules of V8 are as follows  (https://v8.dev/docs/version-numbers)
```
V8’s version numbering scheme
V8 version numbers are of the form x.y.z.w, where:
* x.y is the Chromium milestone divided by 10 (e.g. M60 → 6.0)
* z is automatically bumped whenever there’s a new LKGR (typically a few times per day)
* w is bumped for manually backmerged patches after a branch point
If w is 0, it’s omitted from the version number. E.g. v5.9.211 (instead of “v5.9.211.0”) gets bumped up to v5.9.211.1 after backmerging a patch.
```
'V8_MINOR_VERSION' takes only one digit, but 'V8_MAJOR_VERSION' can have multiple digits.
(If the minor version goes up 10 times :)

> Is it just me, but the SWIG_V8_VERSION definition is not clear and rather crazy. If V8_MAJOR_VERSION is defined (>4), then SWIG_V8_VERSION is a 4 hex number, like 0x04321 otherwise it is a 6 hex number, like 0x034567. Thoughts on 'fixing' or improving this with another version number? This could certainly do with some further clarification for this dual mode version number:

I'm hoping for changes as well.

Support for older V8 versions should be discontinued. Although it is meant to correct versioning inconsistencies, we should stop supporting older V8 engines that are still vulnerable. I would like to see this change from a security perspective as well.

ojwb:
The definition used to be clear, and what was clearly described by the comment you quote.

Originally, `SWIG_V8_VERSION` was only used like this:

```
#if (V8_MAJOR_VERSION-0) < 4 && (SWIG_V8_VERSION < XXX)
```

The idea was that this was just to handle the ancient v8 versions that didn't set any version macros of their own, and for newer V8 we'd just use the now standard `V8_MAJOR_VERSION` and `V8_MINOR_VERSION` for version-specific code.

Then 113d78a083aa897ffdca4ff6bd9b42e630d16d27 started automatically setting `SWIG_V8_VERSION` based on `V8_MAJOR_VERSION` and `V8_MINOR_VERSION`, but with a different format.

But I agree with nxhack - Node is a fast-moving ecosystem used in situations where you really want to be using a version in security support, and we're at the point where we can reasonably just raise the minimum version we support such that v8 < 4.3.0 is no longer a concern.  Then there's no need for manually specifying `SWIG_V8_VERSION` and we can make it a purely automatically-set value and eliminate this mess.

dot-asm:
Quoting https://github.com/swig/swig/issues/1979#issuecomment-812889599:

> ... It was effectively established that probably the only reason to drag on with early V8 support is that to this day Linux vendors have archaic libv8-3.14.5 in currently supported releases. E.g. Ubuntu 18 is bound to support it till 2023, RHEL7 - till 2024, so is Debian 9 till 2022...

And on additional note. The replacement for libv8 spells [libnode-dev](https://github.com/swig/swig/pull/1973/commits/be228d29149f5e67611da8e04487a9ba88db113f), however it's problematic to switch as it is now, because `Tools/javascript` needs extensive "remodeling" to accommodate newer V8.

ojwb:
Debian 9 had codename "stretch", and in the ""Limitations in security support" chapter from the stretch release notes says:

> 5.2.2. Lack of security support for the ecosystem around libv8 and Node.js
>
> The Node.js platform is built on top of libv8-3.14, which experiences a high volume of security issues, but there are currently no volunteers within the project or the security team sufficiently interested and willing to spend the large amount of time required to stem those incoming issues.
>
> Unfortunately, this means that libv8-3.14, nodejs, and the associated node-* package ecosystem should not currently be used with untrusted content, such as unsanitized data from the Internet.
>
> In addition, these packages will not receive any security updates during the lifetime of the stretch release. 

https://www.debian.org/releases/stretch/i386/release-notes/ch-information.en.html#limited-security-support

(Technically we're now past the "lifetime of the stretch release" and it's into LTS which is handled by a different team, but https://wiki.debian.org/LTS/Stretch explicitly says "Additionally, the chapter Limitations in security support from the stretch release notes also still apply.")

The picture isn't prettier for Ubuntu.  Ubuntu clearly can't rely on Debian for security updates for this package, and the package is in the "universe" section which means any support is from the community, not Canonical (so "bound to support it" is inaccurate as nobody's required to do anything in response to a security bug being found).

Checking the changelog from bionic there's only one Ubuntu-specific change which is fixing a build issue, not security:

http://changelogs.ubuntu.com/changelogs/pool/universe/libv/libv8-3.14/libv8-3.14_3.14.5.8-11ubuntu1/changelog

Indeed the last security fix there is from May 2014 (~7 years ago), and was imported from Debian.

I'm not familiar enough with RHEL to know where to look to see if RHEL 7 actually has useful security support for libv8, but I'm doubtful.

nxhack:
@ojwb @dot-asm Thank you for the detailed information.

It seems to have a mixed political(?) background.

In fact, in order to run the latest node.js on debian or ubuntu, it is standard practice to install package managed by the node.js community (including the latest V8) rather than a distribution.

I'll keep an open eye on it for a while.

ojwb:
I don't think it's political - more that the "backport security fixes to a stable old version" stable-release distro mindset and the "everyone upgrade" nodejs mindset don't fit together well.

But from SWIG's point of view, we have limited development resources and supporting what people actually use should be the priority, especially over older versions which don't have meaningful security support, yet are almost exclusively used in environments where security support matters.

And if someone really wants to use the nodejs package from a stable distro release in a context where the lack of security support is OK, they can use the swig package from that stable distro release to generate code for it.

This is essentially the approach I've taken with the PHP support - once PHP upstream developers stop security support, SWIG also stops supporting it (and the error message specifies the last version of SWIG with support.) So we've not supported PHP5 for 2+ years (which nobody has moaned about) and I could put the effort into adding PHP8 support (which numerous people asked about) and to finishing off the work to eliminate the generated `.php` wrapper (which resolves many long-standing reported bugs which were hard to fix before.)

wsfulton:
I've merged this pull request in as it simply fixes a bug setting SWIG_V8_VERSION for when node v8 version is bumped to 10.x.

The documented supported minimum version of node is now 6.0.0 (v8 version 5.0.71.35 according to https://nodejs.org/en/download/releases/). The minimum version was 0.10 until #1746. master is now successfully testing v8 6 and later including latest (v8 17).

As minimum SWIG_V8_VERSION supported is now 0x0500, the code specifically handling V8_MAJOR_VERSION < 4 like ```#if (V8_MAJOR_VERSION-0) < 4``` can be removed.

PR #2013, opened 2021-05-05
Improve php interface handling

ojwb:


ojwb:
@wsfulton Is there a better approach then using `%feature` for this?

https://github.com/swig/swig/pull/2013/commits/a3ee5bffbcc7a9101a71925211d7a76f1e27d52f

The PHP interface doesn't exist as a type SWIG knows about, so a typemap doesn't seem suitable.  I noticed that each global feature seems to get added to every `Node` though, which seems like quite a lot of bloat.

The obvious alternative seems to be to hardcode this in `php.cxx`, but then the user can't specify additional mappings for interfaces they want to use defined by other PHP extensions.  Looking up by name should still work, but has to be redone for every request.

ojwb:
I thought of an alternative way of doing this via fragments, if we added the ability to check if a named fragment existed - for interface `Iterator`, we'd check for fragment (say) `php:interface:Iterator` and if not known we handle by name lookup in RINIT.

If it is known we pull in that fragment in the usual way.

This fragment would chain a fragment for any required headers, then:

```
#define SWIG_PHP_INTERFACE_Iterator zend_ce_iterator
```

And then we can generate code using `SWIG_PHP_INTERFACE_Iterator`

Not sure this is better though.

Or instead it probably could all be done via the compile-time preprocessor, e.g. generate in MINIT:

```
#ifdef SWIG_PHP_INTERFACE_Iterator
  zend_class_implements(SWIGTYPE_Foo_ce, SWIG_PHP_INTERFACE_Iterator);
#endif
```

And in RINIT:

```
#ifndef SWIG_PHP_INTERFACE_Iterator
  zend_class_entry *swig_interface_Iterator_ce = zend_lookup_class(...);
#endif
  // ...
#ifndef SWIG_PHP_INTERFACE_Iterator
  zend_class_implements(SWIGTYPE_Foo_ce, swig_interface_Iterator_ce);
#endif
```

Then one can just `#define SWIG_PHP_INTERFACE_Bar` in the interface file.

This gets a bit ugly to generate and in the generated code if there are multiple interfaces, but I think it's doable, and is perhaps the simplest option of the above for the SWIG user.

Any better ideas or other thoughts?

ojwb:
In the absence of any feedback, I've gone for doing it all via the compiler's preprocessor.

PR #2016, opened 2021-05-07
Python: Fix py3 varlink leak

### LINKED ISSUE #1985: [Python] varlink leak
hello,

running ASAN checks I found out that the varlink objects allocated for each constant wrapped are never deallocated
in particular swig_varlink_dealloc is never called (using good old printfs to verify that in pyinit.swg):
```
class Mesh
{
public:
  static int N;
...
};
```
swig is run with default options, here is some of the asan output:
```
1: Indirect leak of 7 byte(s) in 1 object(s) allocated from:
1:     #0 0x7f4fe8040459 in __interceptor_malloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:145
1:     #1 0x7f4fe460d380 in SWIG_Python_addvarlink /tmp/build/CMakeFiles/openturns.dir/openturnsPYTHON_wrap.cxx:4174
1:     #2 0x7f4fe4614663 in PyInit__openturns /tmp/build/CMakeFiles/openturns.dir/openturnsPYTHON_wrap.cxx:4490
1:     #3 0x7f4fe7dd9d2f  (/usr/lib/libpython3.9.so.1.0+0x216d2f)
```

I tried master and 4.0.2
To reproduce:

example.i
```
%module example 

%{
#include "foo.h"
%}

%include "foo.h"
```

foo.h
```
#ifndef Mesh_h
#define Mesh_h

class Mesh
{
public:
  static const int N;
  Mesh() {}
};

const int Mesh::N = 42;

#endif
```

then run with adress sanitizer enabled:
```
swig -python -c++ example.i
g++ -O1 -g -fsanitize=address -fno-omit-frame-pointer -fPIC -shared -o _example.so example_wrap.cxx -I/usr/include/python3.8
LD_PRELOAD=$(gcc -print-file-name=libasan.so) python3 -c "import example; print(example.Mesh.N)"
```

it seems it was spotted in
https://github.com/swig/swig/pull/1361/commits/65edf2258f359078e98291fb348bccc88653539a
but it works only for python2, I'm testing with python3
cc @adr26

jschueller:
per https://docs.python.org/3/c-api/module.html#c.PyModule_GetDict

"It is recommended extensions use other PyModule_*() and PyObject_*() functions rather than directly manipulate a module’s __dict__."

Closes #1985

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -3422,7 +3422,7 @@ public:
       Printf(f_init, "\t   return;\n");
       Printf(f_init, "#endif\n");
       Printf(f_init, "\t }\n");
-      Printf(f_init, "\t PyDict_SetItemString(md, \"%s\", globals);\n", global_name);
+      Printf(f_init, "\t PyModule_AddObject(m, \"%s\", globals);\n", global_name);
       Printf(f_init, "\t Py_DECREF(globals);\n");
       if (builtin)
 	Printf(f_init, "\t SwigPyBuiltin_AddPublicSymbol(public_interface, \"%s\");\n", global_name);


adr26:
Please see [reftrace.txt](https://github.com/swig/swig/files/6446301/reftrace.txt) for a full reference trace of the example code.

The initial reference is created by `SWIG_Python_newvarlink()`:
```
 # ChildEBP RetAddr      
00 00e4e328 7b3f9fc1     _example!SWIG_Python_newvarlink+0x29 [C:\Temp\leak\example_wrap.cxx @ 3337] 
01 00e4e330 7b3fa352     _example!SWIG_globals+0x11 [C:\Temp\leak\example_wrap.cxx @ 3364] 
02 00e4e348 79dc2133     _example!PyInit__example+0x82 [C:\Temp\leak\example_wrap.cxx @ 3642] 
03 00e4e370 79dc19fb     python38!_PyImport_LoadDynamicModuleWithSpec+0x233 [d:\a\1\s\python\importdl.c @ 165] 

refcount: 1
```
Then the refcount is incremented from 1->2 when `PyDict_SetItemString()` is called, giving ownership of a reference to `globals` to `cvar`:
```
 # ChildEBP RetAddr      
00 00e4e2e8 79cf32e7     python38!insertdict+0x14 [d:\a\1\s\objects\dictobject.c @ 1035] 
01 00e4e300 79cf6267     python38!PyDict_SetItem+0x87 [d:\a\1\s\objects\dictobject.c @ 1545] 
02 00e4e324 7b3fa388     python38!PyDict_SetItemString+0x37 [d:\a\1\s\objects\dictobject.c @ 3415] 
03 00e4e348 79dc2133     _example!PyInit__example+0xb8 [C:\Temp\leak\example_wrap.cxx @ 3651] 
04 00e4e370 79dc19fb     python38!_PyImport_LoadDynamicModuleWithSpec+0x233 [d:\a\1\s\python\importdl.c @ 165] 

refcount: 2
```
The call to `Py_DECREF(globals)` releases the initial reference, once `cvar` safely has a reference to it:
```
 # ChildEBP RetAddr      
00 00e4e324 7b3fa39e     _example!_Py_DECREF+0x17 [c:\Python38\include\object.h @ 470] 
01 00e4e348 79dc2133     _example!PyInit__example+0xce [C:\Temp\leak\example_wrap.cxx @ 3652] 
02 00e4e370 79dc19fb     python38!_PyImport_LoadDynamicModuleWithSpec+0x233 [d:\a\1\s\python\importdl.c @ 165] 

refcount: 1
```
At the end of the test (after another reference has been leaked), `cvar` is released as part of the cleanup of `_example` (from `python -vv`):
```
# cleanup[3] wiping _example
#   clear[2] __name__
#   clear[2] __package__
#   clear[2] __loader__
#   clear[2] __spec__
#   clear[2] SWIG_PyInstanceMethod_New
#   clear[2] new_Mesh
#   clear[2] delete_Mesh
#   clear[2] Mesh_swigregister
#   clear[2] Mesh_swiginit
#   clear[2] cvar
```
At this point, `cvar`'s reference to `globals` is released, and the refcount is decremented from 2->1:
```
 # ChildEBP RetAddr      
00 (Inline) --------     python38!_Py_DECREF+0x3 [d:\a\1\s\include\object.h @ 470] 
01 (Inline) --------     python38!_Py_XDECREF+0x5 [d:\a\1\s\include\object.h @ 541] 
02 00e4f544 79cf32e7     python38!insertdict+0x2ca [d:\a\1\s\objects\dictobject.c @ 1102] 
03 00e4f55c 79d1f10a     python38!PyDict_SetItem+0x87 [d:\a\1\s\objects\dictobject.c @ 1545] 
04 00e4f594 79dbee46     python38!_PyModule_ClearDict+0x2ea [d:\a\1\s\objects\moduleobject.c @ 629] 
05 (Inline) --------     python38!_PyModule_Clear+0xd [d:\a\1\s\objects\moduleobject.c @ 577] 
06 00e4f5e0 79dd53a1     python38!PyImport_Cleanup+0xb46 [d:\a\1\s\python\import.c @ 575] 
07 00e4f608 79c6cea3     python38!Py_FinalizeEx+0x1e1 [d:\a\1\s\python\pylifecycle.c @ 1246] 
08 (Inline) --------     python38!Py_RunMain+0x10 [d:\a\1\s\modules\main.c @ 646] 
09 00e4f798 79c6cf1e     python38!pymain_main+0x183 [d:\a\1\s\modules\main.c @ 674] 
0a 00e4f7b4 008411ea     python38!Py_Main+0x2e [d:\a\1\s\modules\main.c @ 687] 
0b (Inline) --------     python!invoke_main+0x1c [d:\agent\_work\3\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl @ 90] 
0c 00e4f7fc 7628fa29     python!__scrt_common_main_seh+0xfa [d:\agent\_work\3\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl @ 288] 
0d 00e4f80c 77517a4e     KERNEL32!BaseThreadInitThunk+0x19
0e 00e4f868 77517a1e     ntdll!__RtlUserThreadStart+0x2f
0f 00e4f878 00000000     ntdll!_RtlUserThreadStart+0x1b

refcount: 1
```
If another reference to `globals` hadn't been leaked elsewhere, then this release would balance the initial reference to `globals` that is held after the call to `Py_DECREF(globals)` in `PyInit__example()` and `globals` would be cleaned up...

The cause of that other reference leak needs to be addressed, rather than introducing a countervailing incorrect reference decrement in `PyInit__example()` to match the incorrect reference increment later on which causes the leak (two wrongs don't make a right, especially when it comes to reference counting...)

adr26:
As per my comment on the issue, explicitly adding "`del example.cvar;`" to the end of your Python script resolves this.

The creation of `example.cvar` by SWIG in `example.py` is fully intentional, and the problem is cause by a number of extra references to `example.py`'s `__dict__` at the end of Python's execution, so that `example.cvar` (and other globals in `example.py`) aren't automatically freed.

Given this analysis, I don't think this PR should be merged.

PR #2019, opened 2021-05-11
[Python] Fix memory leaks.

adr26:
Fix memory leaks.


jschueller:
it seems a test fails with the -builtin option

adr26:
Hi @jschueller,

I have debugged this and will push a fix shortly, sorry for the error!

Thanks, Andrew R

Lib/python/pyrun.swg
```diff
@@ -1395,19 +1433,14 @@ SWIG_Python_SetModule(swig_module_info *swig_module) {
 #endif
   PyObject *pointer = PyCapsule_New((void *) swig_module, SWIGPY_CAPSULE_NAME, SWIG_Python_DestroyModule);
   if (pointer && module) {
-    PyModule_AddObject(module, "type_pointer_capsule" SWIG_TYPE_TABLE_NAME, pointer);
+    if (PyModule_AddObject(module, "type_pointer_capsule" SWIG_TYPE_TABLE_NAME, pointer) < 0) {
```
wsfulton, line 1436:
```suggestion
    if (PyModule_AddObject(module, "type_pointer_capsule" SWIG_TYPE_TABLE_NAME, pointer) != 0) {
```

CHANGES.current
```diff
@@ -7,6 +7,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.1.0 (in progress)
 ===========================
 
+2021-05-11: adr26
+            #1985 [Python] Fix memory leaks.
```
wsfulton, line 11: Need some more detail here please.

Lib/python/pyerrors.swg
```diff
@@ -95,8 +95,12 @@ SWIG_Python_RaiseOrModifyTypeError(const char *message)
...
     newvalue = PyString_FromFormat("%s\nAdditional information:\n%s", PyString_AsString(value), message);
 #endif
-    Py_XDECREF(value);
-    PyErr_Restore(type, newvalue, traceback);
+    if (newvalue) {
+      Py_XDECREF(value);
+      PyErr_Restore(type, newvalue, traceback);
+    } else {
+      PyErr_Restore(type, value, traceback);
+    }
```
wsfulton, line 103: I don't see any information about this change and why it has been done.

Lib/python/pyrun.swg
```diff
@@ -1309,9 +1336,11 @@ SWIG_Python_NewPointerObj(PyObject *self, void *ptr, swig_type_info *type, int f
       }
     } else {
       newobj = PyObject_New(SwigPyObject, clientdata->pytype);
+      if (newobj) {
```
wsfulton, line 1339: This should be within the `#ifdef`

Lib/python/pyrun.swg
```diff
@@ -269,54 +273,64 @@ SWIG_Python_ExceptionType(swig_type_info *desc) {
 SWIGRUNTIME SwigPyClientData * 
 SwigPyClientData_New(PyObject* obj)
 {
+  SwigPyClientData* data;
```
wsfulton, line 276: From this point below in this function, there is too much unnecessary whitespace change and unnecessary code refactoring for me to follow what has actually changed in a very core function. Please revert unnecessary changes and if necessary follow up with a whitespace only / no op commit.

wsfulton (changes requested):
Great to have these memory leaks fixed. Sorry, but I just couldn't follow the changes in `SwigPyClientData_New`, see the review comments.

Lib/python/pyerrors.swg
adr26: I was reviewing the reference counting and spotted this as an out-of-memory \(OOM\) problem.

As described in [the Python/C API documentation for exception handling](https://docs.python.org/3/c-api/exceptions.html?highlight=pyerr_restore#c.PyErr_Fetch), `PyErr_Fetch()` and `PyErr_Restore()` operate as a pair, with `PyErr_Fetch()` acquiring a reference to `value` and `traceback` for the caller \(here `SWIG_Python_RaiseOrModifyTypeError()`\), and `PyErr_Restore()` handing references back. These then allow an original, precipitating error message either not to be overwritten by some operation performed between `PyErr_Fetch()` and `PyErr_Restore()`, or—as in this case—allow the original error message to be altered / extended.

Looking at this code, note that `newvalue` takes the return value of `PyUnicode_FromFormat()`/`PyString_FromFormat()` and previously `newvalue` was just unconditionally passed to `PyErr_Restore()`. In the case of OOM, `PyUnicode_FromFormat()`/`PyString_FromFormat()` \(which creates a new string object, and passes ownership of a reference to this string to the caller when an object is created\) can return `NULL` and fail to create a new string. While `PyErr_Restore()` handles being passed a `NULL` pointer for its `pvalue` argument, and so would not immediately crash as a result of this, we can note two consequences:

1. The Python error message has been corrupted, with `NULL` being substituted for `value`.
2. A reference to the string `value` is now leaked.

The loss of debug fidelity from the first issue is unfortunate, though perhaps could be argued as not compelling enough to necessitate a change—especially since it will only occur in OOM scenarios \(though I do think it's worth advocating for the ability to debug well!\).

However, I noted that we can avoid this loss of fidelity since we still have the original, referenced string in `value` and if in the case that no new string is created in `newvalue`, we instead restore `value` as the error message, then the problem of the leaked reference in `value` is also fixed as well. Given this analysis, I therefore thought it was prudent to address this issue as I was doing a sweep of the python object reference counting behaviour to look for other potential issues.

\(There are other OOM issues in the generated Python code which I haven't addressed in this PR and which would require more extensive changes. My intent was to consider those separately, and just fix any "easy" or "obvious" issues I found in this sweep.\)

Lib/python/pyrun.swg
adr26: I had removed a level of indentation, and the github UI is handling this particularly badly. I usually review diffs using Beyond Compare and my local git client \(using TortoiseGit to launch BC\) and as you can see it looks a lot more comprehensible when viewed in BC!

difference shown in Beyond Compare

However, that said I have looked at the code and I think I can rework to avoid needing to remove the indentation, so will do this so that there's a better diff on github - thanks

Lib/python/pyrun.swg
adr26: Thank you!

Lib/python/pyrun.swg
adr26: The [documentation for `PyModule_AddObject`](https://docs.python.org/3/c-api/module.html?highlight=pymodule_addobject#c.PyModule_AddObject) has example usage which uses `< 0` rather than `!= 0` as the test:
```c
Py_INCREF(spam);
if (PyModule_AddObject(module, "spam", spam) < 0) {
    Py_DECREF(module);
    Py_DECREF(spam);
    return NULL;
}
```
Its protocol is described as:

> Return `-1` on error, `0` on success.

I compared this with the usage of `PyType_Ready()` in this file, whose [documentation states](https://docs.python.org/3/c-api/type.html?highlight=pytype_ready#c.PyType_Ready) that it will:

> Return `0` on success, or return `-1` and sets an exception on error.

Note the following use of `< 0` as the test with `PyType_Ready()` elsewhere in this file (on three occasions):
```c
    if (PyType_Ready(&swigpyobject_type) < 0)
      return NULL;
```
It therefore appeared to me to be most consistent to use `< 0` in this case, to match the existing use with `PyType_Ready()`.

Would you like me to:

1. Leave the code as-is
2. Take your change, and update calls to `PyType_Ready()` as well
3. Take your change, and but leave `PyType_Ready()` using `< 0`

?

Lib/python/pyrun.swg
```diff
@@ -279,18 +283,26 @@ SwigPyClientData_New(PyObject* obj)
     /* the newraw method and newargs arguments used to create a new raw instance */
     if (PyClass_Check(obj)) {
       data->newraw = 0;
-      data->newargs = obj;
```
adr26, line 282: Lest I face an accusation of frivolous change(!), I moved this so that each path where we were taking out an extra ref on `obj`, the `Py_INCREF(obj)` was before the operation which passed ownership of the reference to something else.

In this case, the ownership of this ref is being passed to `data->newargs`, but I found it easiest to understand the pattern of reference counting if the pattern of:

increment ref count -> assign ref to new owner

was followed throughout: I hope you will forgive me!

Lib/python/pyrun.swg
```diff
@@ -279,18 +283,26 @@ SwigPyClientData_New(PyObject* obj)
...
       if (data->newraw) {
-	Py_INCREF(data->newraw);
-	data->newargs = PyTuple_New(1);
-	PyTuple_SetItem(data->newargs, 0, obj);
+        data->newargs = PyTuple_New(1);
+        if (data->newargs) {
+          Py_INCREF(obj);
+          PyTuple_SET_ITEM(data->newargs, 0, obj);
+        } else {
+          Py_DECREF(data->newraw);
```
adr26, line 296: I considered the options of:

1. Setting `data->destroy = 0` here and then just invoking `SwigPyClientData_Del()`
2. Having cleanup code at the end of the function and using `goto fail;`, etc.
3. Putting the cleanup inline (as I have done)

If you feel strongly that one of the other options would be preferable, I am happy to update accordingly.

Lib/python/pyrun.swg
```diff
@@ -127,7 +127,11 @@ SWIG_Python_AppendOutput(PyObject* result, PyObject* obj) {
     if (!PyList_Check(result)) {
       PyObject *o2 = result;
       result = PyList_New(1);
-      PyList_SetItem(result, 0, o2);
+      if (result) {
+        PyList_SET_ITEM(result, 0, o2);
+      } else {
+        return o2;
```
adr26, line 133: This is another OOM fix - but as well as giving a sane return value when we fail to append (i.e. the object as-was), we should also call `PyErr_NoMemory()` , so the caller is made aware that the append failed. I'll just push a change to do that...

adr26: Scratch that: `PyList_New` \(and others\) all call through to `_PyObject_GC_Malloc` which has this:
```c
    g = (PyGC_Head *)PyObject_MALLOC(
        sizeof(PyGC_Head) + basicsize);
    if (g == NULL)
        return PyErr_NoMemory();
```
so the exception will already be set up when we get `NULL` back...

Lib/python/pyrun.swg
```diff
@@ -279,18 +283,26 @@ SwigPyClientData_New(PyObject* obj)
...
-	PyTuple_SetItem(data->newargs, 0, obj);
+        data->newargs = PyTuple_New(1);
+        if (data->newargs) {
+          Py_INCREF(obj);
+          PyTuple_SET_ITEM(data->newargs, 0, obj);
+        } else {
+          Py_DECREF(data->newraw);
+          Py_DECREF(data->klass);
+          free(data);
+          PyErr_NoMemory();
```
adr26, line 299: The call to `PyErr_NoMemory` is redundant because `PyTuple_New` will set up the OOM exception, would you like me to strip it out?

Lib/python/pyrun.swg
```diff
@@ -343,6 +355,9 @@ SwigPyObject_get___dict__(PyObject *v, PyObject *SWIGUNUSEDPARM(args))
   if (!sobj->dict)
     sobj->dict = PyDict_New();
 
+  if (!sobj->dict)
```
adr26, line 358: Similarly, this could be solved by `Py_XINCREF()` below, since `PyDict_New` sets up the OOM exception. Would you prefer that as a change?

CHANGES.current
adr26: Please see updated changes for full description of changes.

Lib/python/pyrun.swg
wsfulton: It's not that important, leave it, a consistent approach is good. I just don't think that `<0` correlates as well with the documented protocol and `!=0` generates nice and simply assembler.

Lib/python/pyrun.swg
```diff
@@ -299,10 +311,7 @@ SwigPyClientData_New(PyObject* obj)
       data->destroy = 0;
     }
     if (data->destroy) {
-      int flags;
-      Py_INCREF(data->destroy);
-      flags = PyCFunction_GET_FLAGS(data->destroy);
-      data->delargs = !(flags & (METH_O));
+      data->delargs = !(PyCFunction_GET_FLAGS(data->destroy) & METH_O);
```
wsfulton, line 314: Just a note/observation that PyObject_GetAttrString is documented as returning a new reference, so the removal of the Py_INCREF is okay.

adr26: Thank you for noting this - sorry if it wasn't immediately clear.

Lib/python/pyrun.swg
wsfulton: The cleanup is simple and fine to leave here.

Lib/python/pyrun.swg
wsfulton: Yes, please remove.

Lib/python/pyrun.swg
wsfulton: Yes please.

Lib/python/pyrun.swg
adr26: Thanks for the feedback - you would in general have me at "it generates simpler/better assembler"!

I've made the other changes to remove the unnecessary calls to `PyErr_NoMemory()` but have left this untouched. If you would like me to change to option 2, then I'd be happy to do this as well.

Lib/python/pyrun.swg
```diff
@@ -582,6 +603,7 @@ SwigPyObject_own(PyObject *v, PyObject *args)
       } else {
         SwigPyObject_disown(v,args);
       }
+      Py_DECREF(Py_None);
```
wsfulton, line 602: This had me puzzled until I started looking at all the return values of the above functions. For code clarity, I suggest instead to use
```Py_DECREF(SwigPyObject_acquire(v,args));```
and
```Py_DECREF(SwigPyObject_disown(v,args));```

I started looking around for other problems for the functions that use SWIG_Py_Void and the call to SwigPyObject_append in SWIG_Python_InitShadowInstance is another leak you could fix in this patch in a similar way.

**Comment by wsfulton** [comment]
Okay, thanks, nice patch, this is quite a comprehensive set of memory leak fixes. Just a couple of minor points to address, then I'll merge it in.

File: Lib/python/pyrun.swg [hunk 13]
```diff
@@ -1581,7 +1628,7 @@ SWIG_Python_NonDynamicSetAttr(PyObject *obj, PyObject *name, PyObject *value) {
     } else {
       encoded_name = PyUnicode_AsUTF8String(name);
       if (!encoded_name)
-        return -1;
+        goto done;
```
adr26, line 1631: There is still a reference on `name`. This is cleaned up by the code at `done`, and `res` is still at its default value of `-1`:
```c
  done:
  Py_DECREF(name);
  return res;
```

Lib/python/pyrun.swg
```diff
@@ -1422,8 +1467,10 @@ SWIG_Python_TypeQuery(const char *type)
     descriptor = SWIG_TypeQueryModule(swig_module, swig_module, type);
     if (descriptor) {
       obj = PyCapsule_New((void*) descriptor, NULL, NULL);
-      PyDict_SetItem(cache, key, obj);
-      Py_DECREF(obj);
+      if (obj) {
```
adr26, line 1470: Further OOM fix: `PyCapsule_New` may return `NULL`, in which case `PyDict_SetItem` should not be called with a `NULL` value (it will be dereferenced & crash - or trigger `assert(value)` in the debug build).

Lib/python/pyrun.swg
```diff
@@ -1256,7 +1291,11 @@ SWIG_Python_SetSwigThis(PyObject *inst, PyObject *swig_this)
       dict = PyDict_New();
       *dictptr = dict;
     }
-    return PyDict_SetItem(dict, SWIG_This(), swig_this);
+    if (dict) {
```
adr26, line 1294: Further OOM fix: `PyDict_New` may return `NULL`, in which case `PyDict_SetItem` should not be called with a `NULL` dictionary (it will be dereferenced in `PyDict_Check(op)` & crash).

Lib/python/pyrun.swg
```diff
@@ -1200,12 +1230,17 @@ SWIG_Python_NewShadowInstance(SwigPyClientData *data, PyObject *swig_this)
...
-	  dict = PyDict_New();
-	  *dictptr = dict;
-	  PyDict_SetItem(dict, SWIG_This(), swig_this);
-	}
+        PyObject *dict = *dictptr;
+        if (dict == NULL) {
+          dict = PyDict_New();
+          *dictptr = dict;
+        }
+        if (dict) {
```
adr26, line 1238: Further OOM fix: `PyDict_New` may return `NULL`, in which case `PyDict_SetItem` should not be called with a `NULL` dictionary (it will be dereferenced in `PyDict_Check(op)` & crash).

Lib/python/pyrun.swg
```diff
@@ -501,8 +521,12 @@ SwigPyObject_dealloc(PyObject *v)
       if (data->delargs) {
         /* we need to create a temporary object to carry the destroy operation */
         PyObject *tmp = SwigPyObject_New(sobj->ptr, ty, 0);
-        res = SWIG_Python_CallFunctor(destroy, tmp);
-        Py_DECREF(tmp);
+        if (tmp) {
```
adr26, line 524: Further OOM fix: use standard error path (which calls `PyErr_WriteUnraisable`) if `SwigPyObject_New` returns `NULL`

Lib/python/pyrun.swg
```diff
@@ -394,18 +409,23 @@ SwigPyObject_repr(SwigPyObject *v)
 {
   const char *name = SWIG_TypePrettyName(v->ty);
   PyObject *repr = SWIG_Python_str_FromFormat("<Swig Object of type '%s' at %p>", (name ? name : "unknown"), (void *)v);
-  if (v->next) {
+  if (repr && v->next) {
```
adr26, line 412: Further OOM fixes:

If there is an allocation failure, `SwigPyObject_repr` will return NULL, otherwise it will return a reference to the string containing the full representation.  All object references to intermediate strings allocated during building the representation will be freed, in both cases.

Firstly, handle OOM on building the string for the current SWIG object, to avoid calling `PyUnicode_Concat` or `PyString_ConcatAndDel` below with `repr = NULL`.

Lib/python/pyrun.swg
```diff
@@ -394,18 +409,23 @@ SwigPyObject_repr(SwigPyObject *v)
...
     PyObject *nrep = SwigPyObject_repr((SwigPyObject *)v->next);
+    if (nrep) {
```
adr26, line 414: Next, handle OOM on recursively building the representation for the next SWIG objects, to avoid calling `PyUnicode_Concat` or `PyString_ConcatAndDel` below with `nrep = NULL`.

Lib/python/pyrun.swg
```diff
@@ -394,18 +409,23 @@ SwigPyObject_repr(SwigPyObject *v)
...
+      PyObject *joined = PyUnicode_Concat(repr, nrep);
+      Py_DecRef(repr);
+      Py_DecRef(nrep);
+      repr = joined;
 # else
-    PyString_ConcatAndDel(&repr,nrep);
+      PyString_ConcatAndDel(&repr,nrep);
 # endif
+    } else {
+      Py_DecRef(repr);
```
adr26, line 424: Lastly, ensure that on OOM while recursively building the representation for the next SWIG objects (in which case `SwigPyObject_repr` will return `NULL`) the reference to the previously-allocated string for the current SWIG object is released.

Lib/python/pyrun.swg
```diff
@@ -127,7 +127,12 @@ SWIG_Python_AppendOutput(PyObject* result, PyObject* obj) {
     if (!PyList_Check(result)) {
       PyObject *o2 = result;
       result = PyList_New(1);
-      PyList_SetItem(result, 0, o2);
+      if (result) {
+        PyList_SET_ITEM(result, 0, o2);
+      } else {
+        Py_DECREF(obj);
```
adr26, line 133: The protocol for `SWIG_Python_AppendOutput` is that it always steals a reference to `result` and `obj`, and returns a referenced object. If there is an error, this still holds - but an exception may be raised.

In this further OOM fix, we ensure that a reference to `obj` is always consistently stolen.

adr26:
Hi @wsfulton.

I've followed up on the review comments - plus doing another sweep, found yet more easily-addressable OOM fixes.

Could you look over the changes when you have time, please? Thanks again for the constructive feedback to help make this better.

Andrew R

PR #2020, opened 2021-05-13
Changes to allow swig to work with Octave-6.2

rwf1:


rwf1:
It appears to me that the Travis Cl build tests against Octave versions 4.0.0 and 4.2.2.  Those may still be of value for long term distributions but 5.2 is common at least in recent Fedora releases and 6.2 or a later 6.x release seems likely to be coming in new releases.  The Octave interface had some changes between version 4 and version 5 and even more changes between version 5 and version 6.  So it would seem those versions deserve their own tests.

wsfulton:
I don't know if @kwwette wants to have a look and approve the changes.

If not, I need a change to .travis.yml to test Octave 6.2. The patch ought to be validated before merging in as I am no Octave expert and am unable to tell if it has perhaps made things worse. The only way we can keep language modules working in SWIG which don't have a maintainer is by testing and so we should really be testing newer Octave versions before accepting more Octave changes.

I'd like the module_pass example removed from the patch thought as the examples are meant to simple and easy to understand for users to get going. `-Bsymbolic` is rather obscure, I'm not even convinced anyone should be using it.

rwf1:
The -Bsymbolic flag is used by default if mkoctfile is used to compile the wrapped code.  The module_pass example is just a demo of the need for the patch and is out of place as an example.  Its only potential value is as a test to prevent a regression.  It can be removed without affecting the rest of the changes.  

The changes to the code motivated by module_pass are relevant as mkoctfile is meant to simplify compiling code to work with octave and hence is potentially widely used.  I traced the introduction of the -Bsymbolic flag back to 2004 but do not know why it was introduced or what code might now depend on it.  The patch was intended to make it so that swig code can be compiled with or without -Bsymbolic.

ojwb:
> `-Bsymbolic` is rather obscure, I'm not even convinced anyone should be using it.

It can significantly reduce the size and the time to load a shared object, because it resolves references within the object when it's created, rather than when it's loaded.  It's especially a win for large C++ libraries.  It means you lose the ability to intercept and replace such symbols at dynamic link time, but it's pretty rare to need/want to.

Xapian's configure enables the related `-Bsymbolic-functions` automatically by default if supported (it's like `-Bsymbolic` but only applied to functions, not global variables). It's done this for ~14 years so I don't have current benchmarks to hand, but https://maskray.me/blog/2021-05-16-elf-interposition-and-bsymbolic says:

> As a data point, when building the Linux kernel's x86_64 defconfig with a clang -fPIC built clang, my build is 15% faster if I add -Bsymbolic-functions to libLLVM.so and libclang-cpp.so. I cannot tell performance difference with a mostly statically linked PIE clang.

So there are good reasons to use it, but I'd tend to agree it's probably not something for the examples.

rwf1:
This seems to have stalled.  

It appears that there is a place holder in the config file for a test of swig with octave-6:

`
# octave-6.1 not working
#    - compiler: clang
#      os: osx
#      osx_image: xcode12.2
#      env: SWIGLANG=octave SWIGJOBS=-j2 CPP11=1
`
Would uncomenting these lines enable the test ?

ojwb:
All travis build machines are dual core (https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system), so such high `-j` settings aren't going to help.  For this sort of load probably `-j2` is would be fastest, maybe `-j3` if there's plenty of spare RAM.

If the octave tests can't be made to run faster I'd probably suggest restricting CI testing to Linux as that seems to run faster on travis.

rwf1:
It seems that the OSX runs will time out independent of make options.  One of the java builds just squeaks by with 43 minutes so it is not just the octave builds that are slow under OSX.  The available linux options using ubuntu do not include those with octave-6.  Waiting for travis-ci to offer a newer Ubuntu version or splitting the octave tests into two parts are the two paths I can see.  Suggestions ?

ojwb:
Travis only use LTS versions of Ubuntu, and the next one of those won't be released until April 2022, and probably then won't be supported by travis right away - just waiting likely means no test coverage with Octave 6 for a year or more.

Maybe you can use a flatpak build of Octave 6.2 for CI?  https://wiki.octave.org/Octave_for_GNU/Linux#Distribution_independent

rwf1:
As of June 15 travis-ci.org is no longer doing builds.  It appears that for this to go forward there needs to be a transition from travis-ci.org to travis-ci.com.

ojwb:
> As of June 15 travis-ci.org is no longer doing builds. It appears that for this to go forward there needs to be a transition from travis-ci.org to travis-ci.com.

I don't think I have sufficient permissions on the swig github project to do the migration - probably needs @wsfulton.

rwf1:
Since I modified ci.yml I can no longer push to my fork.  The error message is

To https://github.com/rwf1/swig.git
 ! [remote rejected]     master -> master (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/rwf1/swig.git'

But if I look at the file through the browser I can edit it.  Presumably these restrictions are intended to prevent  abusive behavior but what I see is they create a nuisance for legitimate changes but do not prevent things being done with the intent of creating problems.

vadz:
Sorry, I have no idea what is the real problem here, but I've approved running the workflows for you here just in case it could be helpful.

rwf1:
The approval did allow a run to happen but I still cannot push.  
There is a problem somewhere in how I set up the flatpak as the code that happens to use it is not being run.
So at least I can look into that issue.

rwf1:
Last night I had a run under GHA with octave 6.3 go to completion.  That run never appeared on the dashboard but by clicking on the actions tab I could see it.  It seems to be gone this morning.  The last commits I did were not visible on this page last night so perhaps some overnight synchronization function made the commits visible and deleted the runs.  Hopefully all that is needed is for permission to be granted to the runs.

This situation where all the edits need to be done through the web interface is a real nuisance.  There is some cleanup I could do (the files Tools/octaveFlatpakBin/*oct*-6.2.0 are not used, use of old shell syntax in mkoctfile script, and collaps all the commits for setting up the runs into one commit) but I am not going to take that on through the web interface.

rwf1:
A change made by @vadz  has allowed me to push even though I modified ci.yml.  I have made a couple of changes and now the tests all go to completion.  I still cannot push changes to ci.yml so I am reluctant to take on combining commits.  
This now demonstrates that the patches pass all of the tests.

galtieri:
I tried this branch on Win10. This Swig + Octave-6.3.0-w32 + Coorpop-6.4.2-dev  and the target library was correctly built.

wsfulton:
I've committed the fixes for Octave 6 from this pull request. I've also added testing for Octave 6.4 from the devacom/science ppa rather than go with the overhead and complexity of flatpaks. See [852eab7db33c5cfe073a9c8af983b7482a854075..e0935404a8bb7dae1c984149448ce655a030123e](https://github.com/swig/swig/compare/883b42dc708199db0f05f7fc0e72974b75eac001..e0935404a8bb7dae1c984149448ce655a030123e). Thanks for the patches and testing.

wsfulton:
Looking at #1672, the changes in this pull request are not quite right wrt the `isobject` and `isstruct` renames. The new names were introduced in octave-4.4.0 and so the version detection macro should rename them from this version and later. See wsfulton/swig@f76276df7e49a1154872091416173349ecc7e0d3 for details. I tested this out, but with this change there is a bug elsewhere for versions between 4.4 and 6 that I need help to address. I see this on version 5.2 https://github.com/wsfulton/swig/runs/4441673488?check_suite_focus=true:
```
error: assert (exist ("swigexample") == 3) failed
error: called from
    assert at line 92 column 11
    testme_1 at line 22 column 3
    ../../../../../Examples/octave/module_load/runme.m at line 28 column 1
fatal: caught signal Segmentation fault -- stopping myself...
make[3]: *** [../../Makefile:480: octave_run] Error 139
make[2]: *** [../../../../../Examples/octave/module_load/../example.mk:11: check] Error 2
make[1]: *** [Makefile:216: module_load.actionexample] Error 2
``` 
Help!

wsfulton:
@rwf1 I do need some help here please to fix your patch.

rwf1:
Sorry about not posting.  I tracked down the problem to a change in the response of the octave "exist" builtin.  I have not tracked down why the patches cause that change but suspect it is caused by the changes in builtin.swg.  But looking at object.h in the system include files for python shows that the changes should have no effect on behavior as they are both inline calls to the same routine.  I stalled at this point and have had trouble getting back to this but I have not abandoned it.

rwf1:
I had to build octave for 5.2 from the tar file due to an upgrade of Fedora.  Now what I see is the single change that makes this code work with octave-5.2 is to undo the changes introduced in f76276df7e49 that affects isobject.  Between the point where is_object was deprecated in octave-4.2.2 and where isobject was required in octave-6.0 there are a long list of changed files that reference isobject (comparing 5.2 to 6.2 there are 19 changed files, 1 removed file, and 5 new files).  I have tried to see if I could isolate the change that makes isobject work in 6.2 but fail in 5.2 but there are too many.  My suggestion is to revert f76276df7e49.  Those changes are not needed until octave-6.0.  Until that point the only ill effect I can see from not using them is the message about functions being deprecated at compile time.  I suspect there was a bug in the implementation of isobject that was not corrected until it became required in 6.0 and introducing these changes in earlier versions triggers that bug.  I have tried to see if I could isolate the change but haven't been able to.  Knowing exactly what the problem is would be satisfying but it will not change the workaround which is to not apply the switch to isobject until octave-6.0.

wsfulton:
This was actually resolved in #1672 two days ago. Does that seem okay instead?

PR #2032, opened 2021-06-01
Ruby: Fix warnings in generated code about missing parameter in variadic macro

treitmayr:
The Ruby C API function `rb_funcall` is used in various places in generated code for invoking a Ruby method without parameters. The C function uses a variadic parameter list for the arguments passed to Ruby, therefore in these cases the list of variadic parameters is empty.
As an optimization Ruby may implement the `rb_funcall` function as a macro (e.g. in Ruby 2.6.6) for which however a C/C++ compiler in '-pedantic' mode will issue a warning due to the empty list of arguments for `...` as of C99 and C++11.
In order to prevent such compiler warnings, this PR replaces all such occurrences with a call to `rb_funcallv` (which incidentally is invoked by the `rb_funcall` macro anyway, at least for Ruby 2.6.6).

### DIFF
--- a/Lib/ruby/rubyclasses.swg
+++ b/Lib/ruby/rubyclasses.swg
@@ -207,8 +207,8 @@ namespace swig {
                        (VALUEFUNC(swig_rescue_swallow)), Qnil);
       }
       if (ret == Qnil) {
-        VALUE a = rb_funcall(         _obj, hash_id, 0 );
-        VALUE b = rb_funcall( VALUE(other), hash_id, 0 );
+        VALUE a = rb_funcall2(         _obj, hash_id, 0, 0 );
+        VALUE b = rb_funcall2( VALUE(other), hash_id, 0, 0 );
         res = op_func(a, b);
       } else {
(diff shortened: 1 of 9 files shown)


treitmayr:
It appears that the `rb_funcallv` function is only available starting with Ruby 2.1, therefore I fixed the commit by using the older `rb_funcall2` function (which nowadays is just a synonym for `rb_funcallv`).

treitmayr:
While I still think this PR would be perfectly valid, I now noticed a lot of other warnings when compiling with the `-pedantic` switch, mostly shadowed local variables, ignored qualifies, and unused parameters. So this PR is probably not really worth it, but instead I will prepend the wrapper code with a set of directives for accepting a set of warnings.

wsfulton:
Am happy to merge this in. Any improvements to suppress compiler warnings are welcome. While getting rid of `-pedantic` warnings completely might be too challenging, the cleaner the code the better.

ojwb:
Let's actually merge it then!

PR #2033, opened 2021-06-01
Ruby: Fix missing checks for lower bounds of unsigned long (long) parameters

treitmayr:
The Ruby C API functions which are used for converting a function's parameter from a Ruby value to an `unsigned long` or `unsigned long long` integer parameter, does not check for the lower bound (which of course is 0). Consequently if a value of e.g. `-1` is provided for such a parameter, the C/C++ code will be provided with some positive value (I think it is just the binary equivalent to the signed version).
More discussions on this topic can be found at https://bugs.ruby-lang.org/issues/9089 .

In this PR I want to provide the associated test cases first (still without the fix) in order to show which builds in the CI pipeline are affected, and then add the fix to this PR.

Note: The test is currently limited to Ruby but of course could be extended to other languages if the need arises.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -336,6 +336,7 @@ CPP_TEST_CASES += \
 	newobject1 \
 	newobject3 \
 	null_pointer \
+	numeric_bounds_checking \
 	operator_overload \
 	operator_overload_break \
 	operator_pointer_ref \
--- a/Examples/test-suite/numeric_bounds_checking.i
+++ b/Examples/test-suite/numeric_bounds_checking.i
@@ -1,0 +1,47 @@
+%module numeric_bounds_checking
+
+// Tests the bounds checking for integral parameters passed to wrapped functions.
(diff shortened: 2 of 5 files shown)


treitmayr:
Just saw that there is some minimal bounds checking already being done in `Examples/test-suite/primitive_types.i`. I will evaluate if it makes sense to move the new test cases into this file.

treitmayr:
I just noticed that there is some minimal bounds checking already being done in `Examples/test-suite/primitive_types.i` for `unsigned char`. I will evaluate if it makes sense to move the new test cases into this file.

ojwb:
This seems like a clean change and includes test coverage which passes CI so I think we should consider merging for 4.2.0.  The only slight concern I have is that making a call into Ruby to check if the value is negative seems needlessly inefficient but I don't actually see a way to make that check on C/C++ side.

I think I just tagged this originally in the hope a SWIG/Ruby maintainer would look at it but it seems we no longer have anyone active there.

PR #2039, opened 2021-06-20
Support move-only types in SwigValueWrapper.

LindleyF:
This is useful for writing typemaps that use SwigValueWrapper with move-only types.

I make no claims about whether this is sufficient to handle move-only types in auto-generated code; but it appears to do no harm.

### DIFF
--- a/Lib/swig.swg
+++ b/Lib/swig.swg
@@ -672,7 +672,11 @@ template<typename T> class SwigValueWrapper {
   SwigValueWrapper(const SwigValueWrapper<T>& rhs);
 public:
   SwigValueWrapper() : pointer(0) { }
+#if __cplusplus >= 201103L
+  SwigValueWrapper& operator=(T t) { SwigMovePointer tmp(new T(std::move(t))); pointer = tmp; return *this; }
+#else
   SwigValueWrapper& operator=(const T& t) { SwigMovePointer tmp(new T(t)); pointer = tmp; return *this; }
+#endif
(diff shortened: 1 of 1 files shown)


wsfulton:
Can you please provide an example use case for testing/understanding please.

LindleyF:
I'm working on a pattern for generic, composable typemaps. This already works for most T in absl::StatusOr<T>:

`%typemap(out) STATUSOR_T {
  if ($1.ok()) {
    const auto& Tvalue = $1.value();
    {
      // In order to use the subtype's out typemap, we need to temporarily
      // re-define $1. We shadow $1 so it can have a different type, but in this
      // case we do NOT shadow $result so it will be used as expected.
      SwigValueWrapper<CppType> $1;
      $1 = Tvalue;
      $typemap(out, CppType)
    }
  } else {
    ExceptionFunc($1.status());
  }
}`

I'm not sure if SwigValueWrapper is 100% essential, but it's the easiest thing I've found so far to make both pointer and value T work with the same code. However, T==std::unique_ptr<U> does not currently work. It can be made to work with this PR and the addition of a couple of std::moves to the typemap.

LindleyF:
This is submitted to Google's fork of the repo.

wsfulton:
The change I've committed adds a move assignment operator to SwigValueWrapper. This  addresses your needs, I think, please check. It also improves the wrappers to take advantage of move semantics for any type T used with `SwigValueWrapper<T>` that has move semantics.

PR #2040, opened 2021-06-20
add: docs for swig char** python3

singh-lokendra:
- Python3 has removed PyString_AsString
- a similar functionality is available using PyUnicode_AsUTF8

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -5498,16 +5498,16 @@ wheel size: 135019664
 What has happened here is the garbage collector has collected the <tt>Bike</tt> instance as it doesn't think it is needed any more.
 The proxy instance, <tt>wheel</tt>, contains a reference to memory that was deleted when the <tt>Bike</tt> instance was collected.
 In order to prevent the garbage collector from collecting the <tt>Bike</tt> instance, a reference to the <tt>Bike</tt> must
-be added to the <tt>wheel</tt> instance. 
+be added to the <tt>wheel</tt> instance.
 </p>
-    
+
 <p>
 You can do this by adding the reference when the <tt>getWheel()</tt> method
 is called using one of three approaches:
 </p>
-    
+
 <p>
-The easier, but less optimized, way is to use the <tt>%pythonappend</tt> directive 
+The easier, but less optimized, way is to use the <tt>%pythonappend</tt> directive
 (see <a href="#Python_nn42">Adding additional Python code</a>):
 </p>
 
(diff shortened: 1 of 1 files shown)


Doc/Manual/Python.html
```diff
@@ -4971,6 +4971,60 @@ <H3><a name="Python_nn59">33.9.1 Converting  Python list to a char ** </a></H3>
...
+      $1[i] = 0;
+    } else {
+      PyErr_SetString(PyExc_TypeError, "not a list");
+      SWIG_fail;
+    }
+  }
+
+  // This cleans up the char ** array we malloc'd before the function call
+  %typemap(freearg) char ** {
+    free((char *) $1);
```
ojwb, line 5008: This frees the `char **` array but leaks all the allocations in it made by calling `strdup()`.

Doc/Manual/Python.html
```diff
@@ -4971,6 +4971,60 @@ <H3><a name="Python_nn59">33.9.1 Converting  Python list to a char ** </a></H3>
...
+    if (PyList_Check($input)) {
+      int size = PyList_Size($input);
+      int i = 0;
+      $1 = (char **) malloc((size+1)*sizeof(char *));
+      for (i = 0; i < size; i++) {
+        PyObject *o = PyList_GetItem($input, i);
+        if (PyUnicode_Check(o)) {
+          $1[i] = strdup(PyUnicode_AsUTF8(o));
+        } else {
+          free($1);
```
ojwb, line 4994: This leaks any earlier entries allocated with `strdup()`.

Doc/Manual/Python.html
```diff
@@ -4971,6 +4971,60 @@ <H3><a name="Python_nn59">33.9.1 Converting  Python list to a char ** </a></H3>
...
+<div class="code"><pre>
+  %module argv
+
+  // This tells SWIG to treat char ** as a special case
+  %typemap(in) char ** {
+    /* Check if is a list */
+    if (PyList_Check($input)) {
+      int size = PyList_Size($input);
+      int i = 0;
+      $1 = (char **) malloc((size+1)*sizeof(char *));
```
ojwb, line 4988: Why are you allocating an extra entry here?

ojwb:
Currently your revised examples both leak memory which makes them unhelpful to include in the manual.  The Python 2 versions of these examples don't leak as they use the string data returned by `PyString_AsString()`.

ojwb:
I think this section of the manual (for Python 2 and 3) would probably be better just replaced with essentially "Use `argcargv.i`" and a link to the documentation of that (recently added in 6860e2bc03dc2a46de0989dc938bffe63eb0694b).

It doesn't seem helpful to guide users towards rolling their own version of something we provide a working implementation of.  I suspect this section only exists because `argcargv.i` wasn't documented in the manual until a month ago.

ojwb:
> I think this section of the manual (for Python 2 and 3) would probably be better just replaced with essentially "Use `argcargv.i`" and a link to the documentation of that (recently added in [6860e2b](https://github.com/swig/swig/commit/6860e2bc03dc2a46de0989dc938bffe63eb0694b)).

No update to the patch to fix the memory leaks in well over a year, so I've gone for doing this.

> It doesn't seem helpful to guide users towards rolling their own version of something we provide a working implementation of. I suspect this section only exists because `argcargv.i` wasn't documented in the manual until a month ago.

One difference I can see is that the example here only maps `char **` whereas `argcargv.i` only maps `char **` with a second parameter which has a count of the number of entries.

IME, APIs generally take a length count too, though a `NULL`-terminated array of `char*` is certainly a plausible interface.

I don't see how to `%apply` a multi-argument typemap to a single parameter (it'd work to just throw away `argc` in this case), but we could fairly easily provide additional typemaps for that in `argcargv.i` if there's demand (they could use the existing helper).  The main effort is that there's a separate one for each target language.

PR #2042, opened 2021-06-23
Expose $imfuncname to javaout typemaps.

LindleyF:
The $imclassname is already available; this adds $imfuncname for use in javaout typemaps.


LindleyF:
This is submitted to Google's fork of the repo.

vadz:
I don't see why we _wouldn't_ do this, so I'm going to merge it if there are no objections.

But we do need to document this in `Doc/Manual/Java.html`.

ojwb:
@LindleyF  This seems it would be easy to resolve - just needs a short note adding to the docs.

LindleyF:
Pushed a doc change. Wow, the whole "personal access token" thing is totally buried in the settings and far harder to find than it ought to be.....

ojwb:
Thanks.

@wsfulton OK to merge this?

wsfulton:
I can see this being useful so documenting and exposing this special variable is okay. The implementation is not complete though. Probably needs adding wherever $jnicall is substituted and substituteClassname() is called, will need to have a good look. Examples/test-suite/typemap_variables.i needs enhancing with a testcase too to flush out the various typemaps that need the replacement.

Labelling with 4.1 milestone and I'll complete the implementation for the release if no-one else does.

PR #2045, opened 2021-07-01
Support std::array in Golang

juierror:
Support std::array when using swig with Golang


ojwb:
This really should have test coverage so we can be confident it works and will continue to work - i.e. add `Examples/test-suite/go/cpp11_std_array_runme.go` (which can be modelled on the existing `Examples/test-suite/*/cpp11_std_array_runme.*` for other languages).

ojwb:
Thanks for working on this.

You (and the CI infrastructure) would be better off running the new runme locally to debug it - you can run a single testcase with:

```
cd Examples/test-suite/go
make cpp11_std_array.cpptest
```

(This is a C++ testcase - it's `.ctest` instead for a C testcase.)

Examples/test-suite/go/cpp11_std_array_runme.go
```diff
@@ -0,0 +1,74 @@
...
+
+  ai = cpp11_std_array.NewArrayInt6(valsArrayInt6);
+  cpp11_std_array.arrayInPtr(ai);
+  CompareContainers(ai, vals);
+
+  // Fill
+  ai.Fill(111)
+  vals = [6]int{111, 111, 111, 111, 111, 111}
+  CompareContainers(ai, vals);
+}
```
ojwb, line 74: There's no newline on the last line here, which is better avoided.

juierror:
> Thanks for working on this.
> 
> You (and the CI infrastructure) would be better off running the new runme locally to debug it - you can run a single testcase with:
> 
> ```
> cd Examples/test-suite/go
> make cpp11_std_array.cpptest
> ```
> 
> (This is a C++ testcase - it's `.ctest` instead for a C testcase.)

Thank you very much for point out. I will tests it locally to make sure this work before push it to the branch.

Examples/test-suite/go/cpp11_std_array_runme.go
juierror: Thank you, I will fix this.

ojwb:
Great, thanks.

PR #2062, opened 2021-08-02
GitHub Actions workflow

KrisThielemans:
This is a replacement for (or addition to?) Travis CI.

GitHub Actions is free for Open source projects.

This PR tries to be similar to (and re-use some of) the existing Travis files. I've therefore moved the content of `travis-linux-install.sh` to a new script `CI-linux-install.sh`. I've had to make one change to that file related to Python as I needed to install python on GHA, which apparently wasn't needed on Travis.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,8 @@ name: test SWIG via CI
 
 on:
   push:
-    branches: [ master ]
+    # enable on master and GHA branches. The latter should be removed once merged to SWIG/master.
+    branches: [ master, GHA ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,8 +1,8 @@
-name: test SWIG CI
+name: test SWIG via CI
 
 on:
   push:
-    #branches: [ master ]
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,95 @@
+name: test SWIG CI
+
+on:
+  push:
+    #branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # the agent machine operating systems
+        os: [ubuntu-latest]
+        SWIGJOBS: ["-j2"]
+        CSTD: [""]
+        CC: ["gcc"]
+        CPP11: ["1"]
+        CPP14: [""]
+        CPP17: [""]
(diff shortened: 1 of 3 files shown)


KrisThielemans:
I guess this needs to be merged first before you will see the Actions in the tab here. Results on my branch are at
https://github.com/KrisThielemans/swig/actions

KrisThielemans:
Things to do:
- copy more of the travis matrix once everyone's happy
- if I uncomment [`SWIGJOBS` in the matrix definition](https://github.com/KrisThielemans/swig/blob/9d50ec97d0467093db0dd613f7520134c21729f4/.github/workflows/test.yml#L16) only the last test runs. Maybe someone knows more about GHA than me. Here's the [GHA doc](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows).
- it would be better if language-jobs could depend on an initial build-only job. At the moment, all of them would fail, resulting in wasting a lot  of computing resources.
- `CCache-swig` presumably could be cached between runs, but I don't know enough about it.

vadz:
Thanks Kris!

@wsfulton Is it ok if I merge this to allow GitHub Actions to be used for SWIG? Further enhancements could be done in PRs then.

FWIW I think we could/should also remove `.travis.yml` as it's useless now.

KrisThielemans:
sure, want me to remove the .travis now or you'll take care of it?

vadz:
I can definitely do it, I'd just like to wait to see if William doesn't have any objections to this.

vadz:
I've merged it now, let's see how it goes. I've also created #2067 for removing everything Travis related, but didn't merge it yet because it would be nice to propagate more builds from Travis CI to this workflow, notably (but not only) the macOS ones.

Thanks again for doing this!

KrisThielemans:
makes sense. Seems the Actions are working ok 😄

rwf1:
I have been working on a pull request for the octave interface #2020 .  One rough edge I ran into is associated with a permission issue for a first time contributor.  The problem came when I made a change to ci.yml in my local copy of the fork I created.  Once I changed that file I could no longer push to my fork.  I can edit files in my fork (including ci.yml) by using the web interface but I can no longer edit files on my machine and then push them to the repository.  I understand the desire to review the use of resources by reviewing workflows before allowing runs.  But effectively blocking changes through pushes makes it difficult to edit files and since ci.yml can be edited through the web interface I don't see what limiting pushes accomplishes.

vadz:
I've change the option for running actions for SWIG project to "Require approval for first-time contributors who are new to GitHub" which seems to be the least restrictive one. If this doesn't help with your problem, I really don't know what would, sorry.

rwf1:
Thank you.  That now lets me do pushes on files other than ci.yml.  I still need to use the web based editor for that file but as long as the other files can be handled through pushes that is a minor inconvenience.

PR #2064, opened 2021-08-05
[Python] Fix overload_simple_cast test with 3.10

### LINKED ISSUE #2044: python_overload_simple_cast fails with Python 3.10~b3
Python test `python_overload_simple_cast` starts failing with  Python 3.10~b3, which is part of Fedora rawhide. 

```
checking python testcase python_overload_simple_cast (with run test)
  File "/builddir/build/BUILD/swig-4.0.2/Examples/test-suite/python/python_overload_simple_cast_runme3.py", line 42, in <module>
    if fint(ai) != "fint:int":
  File "/builddir/build/BUILD/swig-4.0.2/Examples/test-suite/python/python_overload_simple_cast.py", line 127, in fint
    return _python_overload_simple_cast.fint(arg1)
TypeError: in method 'fint', argument 1 of type 'int'
make[1]: *** [Makefile:123: python_overload_simple_cast.cpptest] Error 1
```

The test failed with swig 4.0.2 and swig 4.1.0,

jschueller:
Closes #2044

see bpo-37999

### DIFF
--- a/Examples/test-suite/python/python_overload_simple_cast_runme.py
+++ b/Examples/test-suite/python/python_overload_simple_cast_runme.py
@@ -9,6 +9,8 @@ class Ai:
     def __int__(self):
         return self.x
 
+    def __index__(self):
+        return self.x
 
 class Ad:
 


vadz:
Sorry if I'm missing something obvious here (which is perfectly possible because I have no idea about what's going on in this example), but could you please explain why do we need to define `__index__` now?

In any case, if it fixes a crash, it should be merged, of course.

jschueller:
The __index__ method is necessary now to allow a custom class to be converted to int

vadz:
The [documentation](https://docs.python.org/3/reference/datamodel.html#object.__index__) seems to imply that `__int__` or `__float__` should be enough, i.e. that `__index__` is only a fallback, so I still don't really understand this...

jschueller:
Also the doc says "Presence of this method indicates that the numeric object is an integer type", so maybe this property is also looked for.

PR #2065, opened 2021-08-05
[CMake] Add option to enable pcre 

### LINKED ISSUE #2031: CMake/Windows issues regarding PCRE
Hello!

I was looking at these lines here: https://github.com/swig/swig/blob/master/CMakeLists.txt#L76-L81

And I wondered, how exactly this option was to be used? After looking at the CMake documentation for `find_package`, I knew I could just set this flag to an empty string to avoid needing PCRE on Windows builds. However, this is not very optimal.

Did you consider to download the PCRE dependency via the CMake script when `IF(MSVC)` matches? Or using an alternative flag?

`PCRE_REQUIRED_ARG` just feels very confusing really. Nothing broken or anything but I think this could be improved on.

jschueller:
Closes #2031


vadz:
In principle, it's possible to build SWIG without PCRE, so I'm not sure why should it be made mandatory.

Personally I'd consider including PCRE(2) as submodule of SWIG repository and build it if the system version is not found.

jschueller:
Instead I added an option to opt-out pcre to match the autotools behavior.

jschueller:
ping @vadz @IngwiePhoenix

vadz:
Looks good to me and should be merged, I think. I'll do it if there are no objections.

jschueller:
@vadz maybe it could be merged ?

vadz:
I hoped to add a CI workflow using CMake first, in order to test that this doesn't break anything, but unfortunately I just don't think I'm going to have time for this in the near future, so I'm going to just merge this with a small mention of the new option in the docs.

Thanks!

vadz:
This was merged as 42ed181.

jschueller:
thanks!

PR #2074, opened 2021-08-31
avoid indentation warning

ajrh1:
gcc-11 complains about the indentation here.    inserting a newline seems like the simplest fix

modules/swig/perl/obj/etp.swig.cpp:10135:7: warning: this ‘if’ clause does not guard... [-Wmisleading-indentation]
10135 |       if (argvi >= items) EXTEND(sp, argvi+1);  ST(argvi) = d3._sv_; argvi++  ;


vadz:
Thanks!

FWIW this looks good to me and ought to be merged, I think. If there are no objections, I'll do it before the (long past) 4.1 milestone.

ojwb:
@vadz Looks sensible to me - I'd say just get it merged, assuming the generated code with this change looks right.

vadz:
I've merged the latest master into this branch to rerun the CI builds, if/when the Perl one passes, I'm going to cherry pick the original commit to master (i.e. this branch should _not_ be merged any more, as this would result in an unnecessary weird criss-cross merge).

vadz:
There was a failure, but a completely unrelated one, so I'm still merging this.

Thanks again!

PR #2078, opened 2021-09-16
Fix li_carrays.m

luciferlys:
li_carrays.m is fixed by using correct index of array in MATLAB.


vadz:
Are you really sure you want to merge 400+ commits or just 3ef3233e5955e77a91a9faffc1f7d0920d9af2b5? I think it's the latter, but there is also an unrelated change to `Examples/test-suite/matlab/overload_extend_runme.m` even in this commit.

luciferlys:
> Are you really sure you want to merge 400+ commits or just [3ef3233](https://github.com/swig/swig/commit/3ef3233e5955e77a91a9faffc1f7d0920d9af2b5)? I think it's the latter, but there is also an unrelated change to `Examples/test-suite/matlab/overload_extend_runme.m` even in this commit.

Hi, I'd like to only merge [3ef3233](https://github.com/swig/swig/commit/3ef3233e5955e77a91a9faffc1f7d0920d9af2b5). I do forget to mention the change to `Examples/test-suite/matlab/overload_extend_runme.m`. I tried to fix that too but failed anyway, I think keeping the change may be helpful for other developers because at least one possible cause was ruled out. Thanks for your review and hope everything has been explained.

luciferlys:
I found that I made a stupid mistake. I shouldn't make a PR to the main swig. Sorry for any inconvenience caused.

PR #2084, opened 2021-09-30
Miscellaneous CI improvements

vadz:
The goal is to bring the GitHub Actions to parity with the old Travis CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,6 @@ jobs:
       SWIGLANG: ${{ matrix.SWIGLANG }}
       PY3: ${{ matrix.PY3 }}
       SWIG_FEATURES: ${{ matrix.SWIG_FEATURES }}
-      CONFIGOPTS: ${{ matrix.CONFIGOPTS }}
       CSTD: ${{ matrix.CSTD }}
       CPP11: ${{ matrix.CPP11 }}
       CPP14: ${{ matrix.CPP14 }}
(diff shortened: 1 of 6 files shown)


vadz:
Assuming the last build passes, this should be ready to be merged, even if more improvements remain possible, of course.

KrisThielemans (approved):

KrisThielemans:
a number of jobs still queued on Appveyor. https://ci.appveyor.com/project/swig/swig/history. Those should probably be cancelled except the last one. (and use `[appveyor skip']' in future)

vadz:
> a number of jobs still queued on Appveyor. https://ci.appveyor.com/project/swig/swig/history. Those should probably be cancelled

Unfortunately I don't have permissions to cancel these builds on AppVeyor, so the best (and only) thing I can do is tol make a PR to cancel the old builds of PRs on AppVeyor if a new PR version is available in the future to avoid this happening again. In any case they should pass because nothing here really affects AppVeyor builds at all (so I'll also add `skip_commits` to `appveyor.yml`...).

> (and use `[appveyor skip']' in future)

Personally I dislike littering the Git history forever with directives for the CI systems. At least `[skip ci]` on its own has some semantic meaning, but I don't like using directives for AppVeyor (or anything else) specifically at all.

vadz:
Merging this did cancel the outstanding AppVeyor builds for this PR, so at least this problem was easy to fix.

PR #2085, opened 2021-10-03
AppVeyor CI improvements

vadz:
The important commit here is the last one, preventing the queue from overflowing with outdated builds in case of an active PR.


vadz:
> Sounds like a good idea. From what I can make out it will only apply to pull requests and not commits to master?

Yes, exactly.

I've added the comment and will merge it in a moment, thanks.

PR #2092, opened 2021-11-02
Try to fix JSC CI job

vadz:


vadz:
This wasn't a JSC-specific problem, finally, but rather just a missing `apt-get update` in the script -- no idea how did it work without it on Travis, but we need to run it to ensure we use the up-to-date package information, of course, so I'm going to push this commit to master to fix the CI builds.

I think we should also apply the (remaining) commit here, i.e. use 20.04 for the JSC CI build rather than 18.04, but this is not critical, so please feel free to close this PR if you prefer to keep running it on the old system.

wsfulton:
Thanks, committed in 6a56426f8f4887f649d06b83c0becbaa69bc6fe3

PR #2093, opened 2021-11-03
Python: Add option to suppress generation of flattened static class methods

jim-easterbrook:
Following a recent discussion on the swig-user mailing list this PR adds an option to suppress the generation of "flattened" static class methods.
https://sourceforge.net/p/swig/mailman/swig-user/thread/CANGqftAP39z0ORLgzJgY7DoTi%2Bj7J%2BqeRTppgnoP8p_Hqya0kA%40mail.gmail.com/#msg37370221

I'd welcome suggestions for a better name for the option!

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -958,6 +958,7 @@ swig -python -help
 <tr><td>-interface &lt;mod&gt;</td><td>Set low-level C/C++ module name to &lt;mod&gt; (default: module name prefixed by '_')</td></tr>
 <tr><td>-keyword        </td><td>Use keyword arguments</td></tr>
 <tr><td>-nofastunpack   </td><td>Use traditional UnpackTuple method to parse the argument functions</td></tr>
+<tr><td>-noflatstaticmember </td><td>Don't generate Foo_bar for static method Foo::bar</td></tr>
 <tr><td>-noh            </td><td>Don't generate the output header file</td></tr>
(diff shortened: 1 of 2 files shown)


casperdcl (approved):

wsfulton:
To fit in with the Zen of python _There should be one-- and preferably only one --obvious way to do it_, I propose we remove the flattened methods by default and add in an option to add them back in for the not-so-Zen users :). I suggest `-flatmembermethods` for the option name.

I was also considering static member variables. I was thinking of also dropping the flattened static member variables (wrapped into cvar). It would mean they could only be accessed via an instance though. I don't see this being always possible as it is not always possible to instantiate all classes (such as when the constructors are not public) and instantiating one has side effects and may be slow. With `-builtin`, though, it is possible to read and write these static member variables without instantiation using syntax like Spam.bar as well as cvar.Spam_bar in addition to instantiating like Spam().bar. The `-builtin` implementation is via a metaclass, which we could port to non-builtin proxy classes. If that was implemented we could also remove the flattened ones in cvar. Probably a separate patch for this though.

jim-easterbrook:
My concern was not to break existing uses, hence the default is things stay as they are.

I've not looked at static member variables - I don't think anything I've used SWIG for has them.

wsfulton:
We're doing a major release next so this is the time where some non-backwards compatible changes  can be accepted.

ojwb:
#2137 takes the alternative approach William proposes here, so closing this in favour of that.

PR #2095, opened 2021-11-09
Use suffix for static methods added by %extend

vadz:
This shouldn't result in any visible changes for the existing backends, but applying this allows to fix a conflict between the wrappers for static functions generated by C backend and the functions synthesized by `%extend`, as they use the same naming convention -- and the public wrappers should really keep it, because they need to have the same form as the real functions (i.e. those not added by `%extend`), so the only solution is to change the names used for the functions added by `%extend`.

Luckily, AFAICS, there should be no real problem with doing this and, if anything, this makes non-overloaded functions more consistent with the overloaded ones, which already use a similar `__SWIG_n` prefix.

### DIFF
--- a/Source/Modules/lang.cxx
+++ b/Source/Modules/lang.cxx
@@ -1320,16 +1320,18 @@ int Language::staticmemberfunctionHandler(Node *n) {
     Delete(mrename);
     mrename = mangled;
 
-    if (Getattr(n, "sym:overloaded") && code) {
-      Append(cname, Getattr(defaultargs ? defaultargs : n, "sym:overname"));
-    }
+    if (code) {
+      if (Getattr(n, "sym:overloaded")) {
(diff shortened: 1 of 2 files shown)


wsfulton (approved):
In summary, for the additional global function that was generated by SWIG when using %extend:
- Previously `__SWIG_x` would be added to the generated name when a function was overloaded and no additional suffix was added for non-overloaded functions.
- Now `__SWIG` is always added to the generated name even for non-overloaded functions. 

I don't think we have an example in the test-suite that this would fix. Is there a testcase  in #2086 that covers this?

vadz:
There is no test case for this specifically, but reverting this commit would result in many test failures in C test suite, so I think it's fair to say that it does cover it.

wsfulton:
Merged, see 3200a815d4dbd40cdd65e8234b64883c82308a0f

PR #2096, opened 2021-11-09
Allow customizing type mangling in SWIG preprocessor

vadz:
Use the "type" naming format for the types mangled by "#@" and "##@"
preprocessor operators, in order to allow customizing them for a
particular backend.

This isn't used by any backend yet, so this doesn't change anything so
far.

---

This is needed by C backend to ensure that mangling used in the extension methods and the actual macros created by `%attribute` and friends is consistent and I don't see any better way to make these macros include the global prefix used for everything else in the C backend.

As the commit message says, this should have no noticeable effect (including performance-wise) on the existing backends.

### DIFF
--- a/Source/Preprocessor/cpp.c
+++ b/Source/Preprocessor/cpp.c
@@ -929,14 +929,14 @@ static String *expand_macro(String *name, List *args, String *line_file) {
       /* Non-standard mangle expansions.  
          The #@Name is replaced by mangle_arg(Name). */
       if (strstr(Char(ns), "\004")) {
-	String *marg = Swig_string_mangle(arg);
+	String *marg = Swig_name_type(arg);
 	Clear(temp);
 	Printf(temp, "\004%s", aname);
 	Replace(ns, temp, marg, DOH_REPLACE_ID_END);
 	Delete(marg);
       }
       if (strstr(Char(ns), "\005")) {
-	String *marg = Swig_string_mangle(arg);
+	String *marg = Swig_name_type(arg);
 	Clear(temp);
 	Clear(tempa);
 	Printf(temp, "\005%s", aname);
--- a/Source/Swig/naming.c
+++ b/Source/Swig/naming.c
@@ -173,6 +173,33 @@ String *Swig_name_mangle(const_String_or_char_ptr s) {
 }
 
 /* -----------------------------------------------------------------------------
+ * Swig_name_type()
+ *
+ * Returns the name of a type.
+ * ----------------------------------------------------------------------------- */
+
+String *Swig_name_type(const_String_or_char_ptr tname) {
+  String *r, *s;
+  String* f = naming_hash ? Getattr(naming_hash, "type") : NULL;
+
+  /* Don't bother doing anything else if there is no special naming format. */
+  if (f) {
+    s = Copy(f);
+    Replace(s, "%c", tname, DOH_REPLACE_ANY);
+  } else {
+    s = (String*)tname;
+  }
+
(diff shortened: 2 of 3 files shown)


wsfulton:
I'm trying to wrap my head around this! I've never used the non-standard C preprocessor macro expansions in SWIG and am not very familiar with them. From what I can make out `#@` mangles a type using Swig_name_mangle at the moment. This has some sort of documented behaviour in https://github.com/swig/swig/blob/3200a815d4dbd40cdd65e8234b64883c82308a0f/Source/Swig/misc.c#L675 This change, however, exposes a whole lot of internal implementations that is in the SWIG source and is one reason that am not keen on it. It also uses "type" as a lookup in `naming_hash` in naming.c. I can't see this being used in any other language models, nor can I fathom what it is trying to do in the C pull request #2086. Maybe it would be easier to understand the motivation for this change if that was clear. Is it just for `#@` and nothing in the SWIG source?

A second reason I'm reluctant to back this patch is it adds in further undocumented non-standard C preprocessor functionality. I think the non-standard C preprocessor features in SWIG were a mistake and we should be moving away from using them rather than adding to them. Aspirationally, I would like to see SWIG move to a standard / third party C preprocessor to fix various bugs and so this patch goes against any hope of that. I think we should think in terms of SWIG features, not preprocessor extensions, for providing whatever drives the name mangling requirements for the C module and %attribute. @vadz, can you please put some flesh around the `%attribute` problem in the C module?

vadz:
AFAIR this is due to `%mangle(Class)` being used in attributes typemaps (`Lib/typemaps/attribute.swg`). I need to customize this to use different mangling for C and C++ wrappers as the names of the C struct and C++ class are different, in general (e.g. namespaces), and I'm almost sure that this can't be resolved without some deep architectural changes to this module if this customization can't be done.

Concerning getting rid of SWIG preprocessor, it's the first time I hear about it, so I'm rather surprised especially because it doesn't seem realistic at all to me -- there are tons of code, both in SWIG own library and in the user files, using its features that can't be implemented using the standard preprocessor. But concerning this particular one, I think we could make `%mangle` a built-in macro, i.e. instead of defining it using `#@`, allow defining this macro in the code. This would be slightly closer to what you'd like to do, but I don't really see a huge difference in practice.

I also don't think it's totally fair to say that this PR is adding any non-standard preprocessor functionality. It just allows to customize the behaviour of the already existing functionality, and in a relatively minor way. If `#@` hack didn't already exist, I wouldn't have added it, but it does exist.

wsfulton:
> I also don't think it's totally fair to say that this PR is adding any non-standard preprocessor functionality. It just allows to customize the behaviour of the already existing functionality, and in a relatively minor way. If #@ hack didn't already exist, I wouldn't have added it, but it does exist.

What I was trying to say is it is adding in a preprocessor function that mangles names from a hard coded manner, which is sort of documented, to one that involves using internal SWIG code with a new level of complexity around the % format specifiers that it effectively exposes. It's making a #@ a much more generic hack.

Can you just add the %attribute testcases to the ignored tests, ditch this pull request to defer the "deep architectural changes to this module" for now? I've always thought the %attribute implementation as rather a complex abomination around other SWIG features. It could possibly be a dedicated feature instead.

vadz:
OK, fine, I've reverted the change to preprocessor `#@` handling (but still kept `Swig_name_type()` addition which is hopefully uncontroversial and is useful in my code) and disabled the previously passing test that relied on it in c1fcf99ea70146982758f02db6058bd7dc58fdf1 in #2086, so closing this one now.

PR #2097, opened 2021-11-11
Improve cast mode for (unsigned)? long long.

degasus:
Call SWIG_CanCastAsInteger only once for better performance

Avoid falling back to SWIG_AsVal(long) as this method should not provide any gain.

Sadly the fallback changed the behavior in cast mode:
For type long, we try to cast with `PyInt_AsLong` first. This truncates all doubles, so `1.999` will yield `1`.
For type long long, we don't call `PyInt_AsLongLong` for unknown reason. The next code will check for 8LSB precision and round correctly.

So the current behavior for long long:
Doubles which fits in 32bit on windows gets implicitly truncated, larger doubles get a precision check (both fraction and 53bit mantissa).
On linux with 64bit longs, everything which fits in 64bit gets implicitly truncated.

This sounds like a *very* strange behavior to me. Which one do we want overall?
Shall it fail for doubles larger than 53bit?
Shall it fail for fractional doubles?


Examples/test-suite/python/python_overload_simple_cast_runme.py
```diff
@@ -165,8 +165,8 @@ def __float__(self):
 
 
 # unsigned long long
-ullmax = 9223372036854775807  # 0xffffffffffffffff
-ullmaxd = 9007199254740992.0
+ullmax = 0xffffffffffffffff
+ullmaxd = float(2**53 - 1)  # 2**53 does not differ from 2**52+1
```
wsfulton, line 169: While I ponder this patch...

1. I don't understand the comment; my python interpreter does not agree with it either:
```
>>> 2**53
9007199254740992
>>> 2**52+1
4503599627370497
```
2. Why "_Decrease the range of double->longlong casts by one. max_d+1.0 will now fail._" ? 9007199254740992 is the maximum value that does accurately convert to float, see below, but you've changed that to one less.
```
>>> ll=9007199254740991
>>> float(ll)
9007199254740991.0
>>> ll=9007199254740992
>>> float(ll)
9007199254740992.0
>>> ll=9007199254740993
>>> float(ll)
9007199254740992.0
```

degasus: Thanks a lot for your review.

1. is a typo. I've corrected it to `2**53 does not differ from 2**53+1`.
2. I actually wonder about the range as well. I think here are two conflicting definitions of max_d. Shall it be the maximum double which can be converted to an integer, or shall `max_d+1` be the first one which cannot be converted to an integer? I think both statements have a valid point, and I'm of course fine with both. Please tell me which way you like more and I will update my patch.

wsfulton: From https://en.wikipedia.org/wiki/Double-precision_floating-point_format 

> Between 2<sup>52</sup>=4,503,599,627,370,496 and 2<sup>53</sup>=9,007,199,254,740,992 the representable numbers are exactly the integers. For the next range, from 2<sup>53</sup> to 2<sup>54</sup>, everything is multiplied by 2, so the representable numbers are the even ones, etc. Conversely, for the previous range from 2<sup>51</sup> to 2<sup>52</sup>, the spacing is 0.5, etc. 

So I think `2**53` should convert. `2**53+1` should not. Anything between `2**53` and `2**53+1` should not either as it is not a whole number. Because whole numbers should not convert, I don't see that there is a real difference in the two definitions of `max_d`.

degasus: That is exactly the point. `2**53+1` does not exist, the closed two numbers are `2**53` and `2**53+2`. Allowing `2**53` will also convert `float(2**53+1) =: 2**53`. So we have to chose between "`2**53+1` should not" and "`2**53` should", both is not possible.

wsfulton: Okay, agreed then that `2**53-1` is the the highest whole number that can be uniquely represented as a Python float / IEEE-754 C double and be converted into an integral number.

wsfulton:
> Sadly the fallback changed the behavior in cast mode:

Am confused about what fallback this is referring to and whether it is before or after your changes in the patch. Please detail a bit more.

> This truncates all doubles, so 1.999 will yield 1.
I couldn't replicate this before or after your change.

wsfulton:
> This truncates all doubles, so 1.999 will yield 1.

So, yes, there is an inconsistency here that needs addressing for castmode. The best documentation we have for cast mode is in the [CHANGES](https://github.com/swig/swig/blob/master/CHANGES) file dated 12/23/2005
```
	      foo(1.0) // Ok
	      foo(1.3) // Error
```
From what I can see silent rounding down incorrectly occurs for signed types (int, long, long long), but an exception is correctly thrown to prevent silent rounding down for unsigned types (unsigned int, unsigned long, unsigned long long).

degasus:
@wsfulton Sorry for the late respond, I've missed your comment. For the "fallback", I'm talking about `SWIG_AsVal_frag(long long)` calling ``SWIG_AsVal_frag(long)`. My patch here dropped this and inlined the required checks to the `long long` itself.

So indeed, the `long` behavior is still kind of broken and I plan to fix this immediately after this one here is done. As this patch here also provides all edge case tests for this cast mode, I hope to get a more consistent behavior overall.

degasus:
@wsfulton Is anything still missing for this PR? I'm currently waiting for it to get merged.

wsfulton:
Are you able to you address the inconsistency? From what I remember experimenting with your patch, I got quite confused with what had changed for the better and for the worse. If the patch only made improvements, it would be okay, bit it does seem to have undesired changes. Given that this is a really core piece of functionality and my time is short, I'm inclined to only accept it if it only introduced desired changes.

degasus:
@wsfulton Yes, this patch adresses this inconsistency for `(unsigned)? long long`. Based on my experiments, an error of roughly the last 5 bits (LSBs) are ignored (e.g. `1.0 + 1e-15`). Larger errors like `1.3`  produces an error.

wsfulton:
> @wsfulton Yes, this patch adresses this inconsistency for (unsigned)? long long

But it changes behaviour in other types, right?

degasus:
Only the behavior of `long long` and `unsigned long long` are modified. The behavior of `long` is not changed in this pull request. I plan do implement the same there afterwards.

degasus:
Is there any way how I can help to avoid the long idle times?

wsfulton:
Sorry, am trying to do too much generally atm and I found it quite hard to find much consistency (both with and without your patch) when I first looked at it and I'm not intimately familiar with this area. As it is really core functionality, I'd like to go through this with a fine tooth comb and would prefer not to merge incremental changes as it is much easier to test and check for consistent behaviour amongst all the primitive types. This is quite hard to do incrementally given overloading using different primitive types is needed to check castmode. So, could you please bring forward your plan to add in fixes for the remaining primitive types (long, int, short etc). 

From what I've already understood in your patch so far, your approach seems okay.

We need a final detailed summary of what has changed too. If you could fill in that end user documentation in the CHANGES.current file or post it in a single comment in this pull request, it'll certainly speed up my checking of the improvements. I'm afraid the pull request comment was not all that clear and contains unanswered questions.

wsfulton:
Hi @degasus, is this abandoned now, or will you attend to the final requests?

wsfulton:
Final ping for @degasus

degasus:
@wsfulton I'm really sorry for not being able to work on this for almost a year.

I think your last points are very valid, and all behavior changes should be kind of double tested and verified.

I've tried to split up the simplest part as #2494 : The unit tests for the min/max values. This PR contains no behavior changes but an integer overflow (obvious bug, not behavior change). I hope that the test documents the behavior difference on MSVC and GCC and makes it easier to discuss.

However the tests in #2494 still lack the rounding properties, I will try to add them as well. Nethertheless #2494 is an improvement without downsides (I guess so...) and can be merged without the discussion about which rounding behavior we want.

wsfulton:
@degasus, could you please revisit this pull request now that #2494 has been merged.

wsfulton:
Given comments in https://github.com/swig/swig/pull/2494#discussion_r1435866678, the `long long` and `long` behaviour needs should be consistent in converting from a Python `float` and the tests there should be modified to reflect a change to fix the consistency.

I think the check in the `long long` for a size < 2**53 is flawed and should instead be replaced with something that determines if the float is actually an integral value.

PR #2100, opened 2021-11-13
Fix typos

rex4539:


vadz:
@wsfulton FYI I've cancelled AppVeyor build after this merge because it didn't go anywhere after 115 (and a half) hours. Doing this seems to have allowed the other queued builds to proceed.

PR #2102, opened 2021-11-16
Fixed crashes when using embedded Python interpreters.

### LINKED ISSUE #2101: Crash when using embedded Python interpreters
We use SWIG to allow users to interact with our SDK by writing Python scripts from within the UI of our application. Each time a user runs a script, a new Python interpreter is created, the extension module is imported, and then the script is run. All of this happens within our application's main process. This looks roughly like the following:
```c++
#include <Python.h>

extern "C"
{
PyObject* PyInit__test(void);
}

void
RunScript( const char* script )
{
	int importSucceeded = PyImport_AppendInittab( "_test", &PyInit__test ); // Import extension module

	Py_Initialize();
	int result = PyRun_SimpleString( script );
	Py_Finalize();
}
```

Our wrapped code looks like the following:
```c++
// Test.h
struct Test
{
    static const int TEST;
};
```
```c++
// Test.cpp
const int Test::TEST = 5;
```
```
// test.i
%module test
%{
#include "Test.h"
%}
%include "Test.h"
```

If you call `RunScript()` with the following script it runs fine, but if you call it again it crashes:
```python
import test
test.Test.__repr__ = lambda self: "Test"
```

The crash is a null pointer dereference in `SWIG_init()`: right at the end of that function, the static `Test::TEST` variable is loaded into the static `globals` var link object. On the second run, this object has been deallocated, which leads to the crash.

I notice that there are multiple issues pertaining to the globals (#1360, #1985, and #2037, which is is still open). Those issues are treating a related issue: in most cases, the globals object is never deleted--a memory leak. In fact, the only reason our application doesn't crash _every_ time you run a script for the second time is that the globals is leaked. For some reason, `swig_varlink_dealloc` is only called if you set a new `__repr__` on a wrapped class, or do some similar things (e.g., adding a new attribute to the module or to a wrapped class).

There are 2 changes to how the globals object is managed that fix the problem. The first is to modify `SWIG_globals()` to check the refcount of the static `globals` variable. If it's 0, create a new one before returning:
```c
SWIGINTERN PyObject *
SWIG_globals(void) {
  static PyObject *globals = 0;
  if (!globals || globals->refcnt <= 0) {
    globals = SWIG_newvarlink();
  }
  return globals;
}
```
The second fix is simpler: get rid of the `SWIG_globals()` function entirely, and just create a new object in `SWIG_init()` itself:
```c
// Instead of globals = SWIG_globals():
globals = SWIG_newvarlink()
}
```
This way there's no static memory available that can get corrupted and cause issues across multiple interpreters. This seems reasonable: if the globals object is supposed to be deallocated when the module is unloaded anyway, why does it have to be static?

EDIT: this was tested against v4.0.1 on Windows

jsenn:
Fixes #2101. There are 3 related changes made here:
1. Move the SWIG_globals() singleton into pyrun from pyint so it
   is visible to SWIG_Python_DestroyModule(). The static globals
   varlink has been extracted out of the function so that it can
   be set to NULL in SWIG_Python_DestroyModule(), which fixes the
   issue described in #2101. (Now when the second interpreter
   starts up, the Swig_Globals_global pointer will be NULL, so it
   knows it has to create a new one.)
2. Remove a Py_DECREF on the globals varlink. The decrement is now
   performed by DestroyModule(), so there's no need to do it in
   SWIG_init().
3. Fixed similar issue with SWIG_Python_TypeCache().


wsfulton:
See my comments in #2101

jsenn:
@wsfulton thanks for taking a look. I just pushed a different fix per your suggestion in the linked issue. See the updated description for details.

NOTE: rather than commit a totally different fix on top of the previous commit, I force-pushed the new change, so if you've pulled the branch you'll need to do a force-pull.

wsfulton:
@jsenn, had a quick look. Why has so much code seemingly moved form the init section to the runtime section?

jsenn:
all of that stuff is the `varlink` plumbing, and then the `SWIG_globals` function itself. `SWIG_Python_DestroyModule` now needs access to `SWIG_globals` so it can properly free the memory, and it lives in the runtime section, so all that stuff had to move into the runtime section as well.

wsfulton:
I see. The moved code is no longer `extern "C"`. Please restore that, otherwise all looks okay.

wsfulton:
Also can you confirm that the reference count does actually go to zero for globals in SWIG_Python_DestroyModule ?

jsenn:
> I see. The moved code is no longer `extern "C"`. Please restore that, otherwise all looks okay.

Good catch. Added the extern "C" to the new code.

> Also can you confirm that the reference count does actually go to zero for globals in SWIG_Python_DestroyModule ?

The refcount isn't 0 after the decref in SWIG_Python_DestroyModule--presumably because there are things other than the module itself holding references to the globals varlink--but it does eventually get to 0, and swig_varlink_dealloc is always called, so the memory is freed.

wsfulton:
Thanks, merged now.

If you could take a look, there are another two singletons which are probably affected if you search for SWIG_STATIC_POINTER in the code. The first is SwigPyPacked_type() which is used for member function pointers. The second is SwigPyObject_type() and is used quite a lot (worth checking the -builtin option too for this one).

jsenn:
@wsfulton thanks! I'll try building our application using `-builtin`. I can't reproduce the crash when using this branch, but that might be because the other singletons are leaked, as the globals object was initially? I'll take a look and see if they have the same issue.

jsenn:
@wsfulton the other singletons don't seem to cause any problems (maybe because they're type objects rather than regular objects?). However, if you use `-builtin`, you run into an issue where all the attributes of the builtin classes are seemingly removed when the module is destroyed, meaning running any subsequent scripts will fail. I tried factoring the singletons you mentioned out, and treating them similarly to `This` and `globals`, but that didn't fix the issue. Not sure what's going on there, but I filed #2114 to make note of it.

PR #2109, opened 2021-11-29
Create v8::Number instead of v8::Integer when uint64_t is bigger than UINT32_MAX

murillo128:
Currently we are converting uint64_t values to v8:Integer, but the v8::Integer:New takes a uin32_t as input, soo if the value is bigger than UINT32_MAX the value is truncated.

With this PR if the value is bigger than UINT32_MAX it is casted to a double and converted into to a v8::Number instead.

### DIFF
--- a/Examples/test-suite/javascript/integers_runme.js
+++ b/Examples/test-suite/javascript/integers_runme.js
@@ -1,18 +1,24 @@
 var integers = require("integers");
 
-var val = 3902408827
-ret = integers.signed_long_identity(val)
-if (ret != val)
-  throw "Incorrect value: " + ret
+function checkOne(val, signed, typeName) {
+  typeName = (signed ? 'signed_' : 'unsigned_') + typeName
 
-ret = integers.unsigned_long_identity(val)
-if (ret != val)
-  throw "Incorrect value: " + ret
+  var size = integers[typeName + '_size']()
+  if ((!signed && val < 0) || (size < 8))
+    return // out of range, skip test
 
-ret = integers.signed_long_long_identity(val)
-if (ret != val)
-  throw "Incorrect value: " + ret
+  ret = integers[typeName + '_identity'](val)
+  if (ret !== val)
+    throw "Incorrect value: expected " + val + ", got " + ret
+}
 
-ret = integers.unsigned_long_long_identity(val)
-if (ret != val)
-  throw "Incorrect value: " + ret
+function checkAll(val) {
+  checkOne(val, true, 'long')
+  checkOne(val, false, 'long')
+  checkOne(val, true, 'long_long')
+  checkOne(val, false, 'long_long')
+}
+
+checkAll(3902408827)
+checkAll(Number.MAX_SAFE_INTEGER)
+checkAll(Number.MIN_SAFE_INTEGER)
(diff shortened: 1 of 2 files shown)


ojwb:
The change looks OK, but there really should also be a testcase added that fails before this change and passes after.

Also, is there an equivalent situation for the signed case when an `int64_t` is < `INT32_MIN` or > `INT32_MAX`?

ojwb:
I notice we used to have similar conditionals here, but 355ef40bc15f2e18372fa749622ee25e849f8d1d and bcc0b6b6164bc867d503a3609c6784ea13d94c42 removed them.  I think the key difference compared to what we had before those changes is that the change proposed here creates a `v8::Number` in the oversize case.

Unhelpfully neither of the previous changes gives justification or added test coverage.

@wsfulton bcc0b6b6164bc867d503a3609c6784ea13d94c42 was yours - any insights?

I think this is worth considering for 4.1.0.  I'd still be happier if it had test coverage, but then the previous changes in this area didn't.

murillo128:
Forgot about this PR.

Could you point us to an example of an already existing test we could use as base for adding the new tests?

cc/ @mildsunrise

ojwb:
> Could you point us to an example of an already existing test we could use as base for adding the new tests?

Probably expanding `Examples/test-suite/javascript/integers_runme.js` would do the job - check that passing suitable values to the wrappers for these functions returns the same value back to JS:

```
  signed long signed_long_identity(signed long x) { return x; }
  unsigned long unsigned_long_identity(unsigned long x) { return x; }
  signed long long signed_long_long_identity(signed long long x) { return x; }
  unsigned long long unsigned_long_long_identity(unsigned long long x) { return x; }
```

Currently only `3902408827` is tested, which will fit in an unsigned 32-bit integer so the case this patch addresses isn't currently tested.

It looks to me like the first case there is buggy for platforms with 32-bit long:

```
var val = 3902408827
ret = integers.signed_long_identity(val)
if (ret != val)
  throw "Incorrect value: " + ret
```

`3902408827` won't fit in a 32-bit signed integer, so it seems incorrect that it round-trips.  That shows a limitation in this testcase - it can detect truncation OK but doesn't check for the value in C/C++ actually being the same as in the target language, just that it maps back and forth - an evil implementation which negated the value in C/C++ would also pass these tests.  Perhaps a test function which returns the passed in integer as a string would be helpful.  For the purposes of test coverage of this patch I don't think you need to worry about that (but if you want to even better!)

ojwb:
> Forgot about this PR.

On that basis, removing from the 4.1 milestone.

ojwb:
Thanks, restoring the milestone.

I'll leave it to @wsfulton to decide whether to include it - I don't want to waste my efforts merging something only to have it reverted if he's unhappy about it.

wsfulton:
The patch will cast a very large 64-bit number into a double to create a `v8::Number`. However, the range of double (and `v8::Number`) does not cover the full range that 64-bit integer numbers do and hence precision will be lost. In fact the newly added test uses MAX_SAFE_INTEGER which is the end of the range that v8::Number can use and is less than UINT64_MAX. Seems that `v8::BigInt` would be a better choice as this does cover the full 64-bit range. Any reason you didn't use this?

murillo128:
BigInt is not the most dev friendly type in javascript:

> A BigInt value cannot be used with methods in the built-in [Math](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math) object and cannot be mixed with a Number value in operations; they must be coerced to the same type

Also given BigInts have arbitrary precision, there is no way to convert from a BigInt to an `uint64_t` safely which would be weird. 

Regarding the loss of precision, that is exactly how Javascript works. You can assign an integer > MAX_SAFE_INTEGER, but it will lose precision.

wsfulton:
Yuck, not sure this is the best approach, but I suppose typemaps will allow users to tweak this to whatever they want.

PR #2111, opened 2021-12-02
Remove obsolete js support

ojwb:
As noted in https://github.com/swig/swig/pull/2011#issuecomment-983962590 this can all be cleaned up now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,9 +89,6 @@ jobs:
           CPPSTD: c++14
         - SWIGLANG: javascript
           ENGINE: jsc
-        - SWIGLANG: javascript
-          ENGINE: v8
-          os: ubuntu-18.04 # libv8-dev only actually provides v8 in 18.04.
         - SWIGLANG: lua
         - SWIGLANG: lua
           VER: '5.3'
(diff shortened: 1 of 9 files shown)


ojwb:
"CI / ruby 2.6 gcc (pull_request)" failed, but is also failing on master.  "CI / ruby 2.6 gcc (push)" was OK.

No idea what's up but nothing Ruby-related is changed here and the failure isn't new, so I'm going to ignore this.

ojwb:
Oh, and appveyor doesn't test javascript, so no point waiting on that to run.

wsfulton:
Ruby 2.6 often fails unfortunately :(

vadz:
> Ruby 2.6 often fails unfortunately :(

FWIW I think there is a real bug in SWIG-generated code, but I just didn't have time to track it down. Using an ASAN-enabled build of Ruby (or maybe just Valgrind?) should help understanding where does it come from.

kkaempf:
> Ruby 2.6 often fails unfortunately :(

Fails in GH actions or during `make check` ?

kkaempf:
I can't reproduce this crash locally (with rvm + Ruby 2.6.9, gcc 11.2.1) 🤔

wsfulton:
I eventually replicated this problem and have raised it as an issue in #2115. @kkaempf if you got any ideas for a fix, can you please comment on this new issue.

PR #2112, opened 2021-12-02
[php7] Use destructor action if present

### LINKED ISSUE #2108: Regression: object destructors no longer called in -php7 mode
Given the following test.i
```
%module foo;
%inline%{
struct foo { int x; };
struct foo* fooNew(void);
void fooDestructor(struct foo*);
%}


%extend foo 
{
    ~foo() 
    {
        fooDestructor(self);
    }

    %newobject my_new;
    static foo* my_new() {
        return fooNew();
    }
}
```

``./swig -Wall -php7 test.i``, with SWIG master (at 0304bdb82baaf5da94b6273c93e2623f1522a2a0) generates a test_wrap.c file where ~foo() is not called, but instead the foo_free_storage() method calls SWIG_remove() which is just a regular free()
```
/* Garbage Collection Method for class foo */
void foo_free_storage(zend_object *object) {
  swig_object_wrapper *obj = 0;
  if (!object)
    return;
  obj = php_fetch_object(object);
  zend_object_std_dtor(&obj->std);
  if (obj->newobject)
    SWIG_remove((struct foo *)obj->ptr);
}
```

This is a simplified test case of a real world project (https://github.com/MapServer/MapServer/issues/6433). The regression dates back to https://github.com/swig/swig/pull/1982

ojwb:
If there's a destructor, use its action instead of free(ptr)
(for C)/delete ptr (for C++).

Fixes #2108

### DIFF
--- a/Examples/test-suite/php/newobject2_runme.php
+++ b/Examples/test-suite/php/newobject2_runme.php
@@ -1,0 +1,17 @@
+<?php
+
+require "tests.php";
+
+check::equal(fooCount(), 0, "no Foo objects expected");
+$foo = makeFoo();
+check::equal(get_class($foo), "Foo", "static failed");
+check::equal(fooCount(), 1, "1 Foo object expected");
+$bar = makeFoo();
+check::equal(get_class($bar), "Foo", "regular failed");
+check::equal(fooCount(), 2, "2 Foo objects expected");
+$foo = null;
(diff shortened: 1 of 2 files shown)


ojwb:
This only touches PHP-specific code and all the PHP CI checks have run and passed, so merging.

PR #2116, opened 2021-12-06
Add Python 3.11 support: use Py_SET_TYPE()

vstinner:
On Python 3.9 and newer, SwigPyBuiltin_SetMetaType() now calls
Py_SET_TYPE(). Py_TYPE() can no longer be usd as an l-value on
Python 3.11:

* https://docs.python.org/dev/c-api/structures.html#c.Py_SET_TYPE
* https://docs.python.org/dev/whatsnew/3.11.html#c-api-changes

### DIFF
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -414,7 +414,11 @@ SwigPyBuiltin_ThisClosure (PyObject *self, void *SWIGUNUSEDPARM(closure)) {
 SWIGINTERN void
 SwigPyBuiltin_SetMetaType (PyTypeObject *type, PyTypeObject *metatype)
 {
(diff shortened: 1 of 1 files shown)


vstinner:
The Py_SET_TYPE() incompatible change was approved by the Steering Council in Python 3.11: https://github.com/python/steering-council/issues/79

See also the draft PEP 674 "Disallow using macros as l-value": https://www.python.org/dev/peps/pep-0674/

There is also an issue with PyDescr_TYPE() and PyDescr_NAME() in SwigPyStaticVar_new_getset(), but I prefer to propose a separated PR for that.

wsfulton:
Looks good to me, thanks

vstinner:
Nice, thank you ;-)

PR #2122, opened 2021-12-14
[lua] Fix maybe-uninitialized warning in generated code

tytan652:
When compiling OBS Studio (which use SWIG) with `-Wmaybe-uninitialized` with GCC.

I found out that this warning is emitted:
```
obsluaLUA_wrap.c:2537:10: warning: ‘result’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 2537 |   return result;
      |          ^~~~~~
```

And this warning comes from the `SWIG_Lua_MustGetPtr()` which does not initialized `result` if `SWIG_ConvertPtr()` goes wrong. So initializing `result` to 0 should fix the issue.

`make check-lua-test-suite` tests passed succesfully on my system. I did not made further tests.


ojwb:
Thanks.

Looks like `SWIG_Lua_MustGetPtr` is never actually used by SWIG itself, but it is documented in `Lua.html` (with the note "This function is rarely used, and may be deprecated in the future") so user typemaps may be using it and we can't simply drop it.

PR #2126, opened 2021-12-20
Use SWIG_TypeCheckStruct() more?

ojwb:
I noticed that since my refactoring to use PHP's C API to create PHP classes, SWIG/PHP could be using `SWIG_TypeCheckStruct()` instead of `SWIG_TypeCheck()` (the difference is the former works from a `swig_type_info*` instead of the typename as a string and so does pointer comparisons instead of string comparisons.  This change sped up running `make check-php-test-suite` by about 10% (which is quite impressive give that runs the PHP interpreter from scratch for every testcase).

I've already pushed that change for PHP in 69b580dfb6e39b5463213908b443a86265bdb23c.

I thought to check if any other places were using code like `SWIG_TypeCheck(to->name, ty)` when `SWIG_TypeCheckStruct(to, ty)` should do the same but more efficiently and found quite a few.

I changed all of them and pushed a branch to allow CI to run more extensive tests than I can locally.  For reasons I've not attempted to dig into yet[*] octave and python failed with my changes, which suggests to me either there's a bug in these backends, or a bug in the docs describing the difference between `SWIG_TypeCheck` and `SWIG_TypeCheckStruct`.  I've backed out the changes for these and that leaves this PR affecting javascript, r, scilab and one change to generic code.

Also, the javascript backend had code in one place to try `SWIG_TypeCheckStruct(x, ty)` with a fallback to `SWIG_TypeCheck(x->name, ty)`.  I dug into the history but the code seems to have been first added like this.  I ran the test suite with the code changed it to call `std::abort()` if the fallback actually did anything and it still passed, so I've dropped that fallback on this branch.

Does anyone have any thoughts on this?

This change gives a good speed-up (at least if PHP is typical) but it makes me nervous that it broke octave and python.

<hr>

[*] Except that I have tried python locally with the change to the common code reverted, but it still fails.

### DIFF
--- a/Lib/javascript/jsc/javascriptrun.swg
+++ b/Lib/javascript/jsc/javascriptrun.swg
@@ -252,7 +252,7 @@ int SWIG_JSC_ConvertPacked(JSContextRef context, JSValueRef valRef, void *ptr, s
   if (ty) {
     if (to != ty) {
       /* check type cast? */
-      swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
+      swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
       if (!tc) return SWIG_ERROR;
     }
   }
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -180,10 +180,6 @@ SWIGRUNTIME int SWIG_V8_ConvertInstancePtr(SWIGV8_OBJECT objRef, void **ptr, swi
   }
   if(cdata->info != info) {
     swig_cast_info *tc = SWIG_TypeCheckStruct(cdata->info, info);
-    if (!tc && cdata->info->name) {
-      tc = SWIG_TypeCheck(cdata->info->name, info);
-      if (tc) std::abort();
-    }
     bool type_valid = tc != 0;
     if(!type_valid) {
       return SWIG_TypeError;
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -182,6 +182,7 @@ SWIGRUNTIME int SWIG_V8_ConvertInstancePtr(SWIGV8_OBJECT objRef, void **ptr, swi
     swig_cast_info *tc = SWIG_TypeCheckStruct(cdata->info, info);
     if (!tc && cdata->info->name) {
       tc = SWIG_TypeCheck(cdata->info->name, info);
+      if (tc) std::abort();
     }
     bool type_valid = tc != 0;
     if(!type_valid) {
@@ -408,7 +409,7 @@ int SWIGV8_ConvertPacked(SWIGV8_VALUE valRef, void *ptr, size_t sz, swig_type_in
   if (ty) {
     if (to != ty) {
       /* check type cast? */
-      swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
+      swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
       if (!tc) return SWIG_ERROR;
     }
   }
--- a/Lib/octave/octrun.swg
+++ b/Lib/octave/octrun.swg
@@ -600,7 +600,7 @@ SWIGRUNTIME void swig_acquire_ownership_obj(void *vptr, int own);
           return SWIG_OK;
         }
       for (unsigned int j = 0; j < types.size(); ++j) {
-	swig_cast_info *tc = SWIG_TypeCheckStruct(types[j].first, type);
+	swig_cast_info *tc = SWIG_TypeCheck(types[j].first->name, type);
 	if (!tc)
 	  continue;
         if(vptr) {
--- a/Lib/octave/octrun.swg
+++ b/Lib/octave/octrun.swg
@@ -600,7 +600,7 @@ SWIGRUNTIME void swig_acquire_ownership_obj(void *vptr, int own);
           return SWIG_OK;
         }
       for (unsigned int j = 0; j < types.size(); ++j) {
-	swig_cast_info *tc = SWIG_TypeCheck(types[j].first->name, type);
+	swig_cast_info *tc = SWIG_TypeCheckStruct(types[j].first, type);
 	if (!tc)
 	  continue;
         if(vptr) {
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1314,7 +1314,7 @@ SWIG_Python_ConvertPtrAndOwn(PyObject *obj, void **ptr, swig_type_info *ty, int 
         if (ptr) *ptr = vptr;
         break;
       } else {
-        swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
+        swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
         if (!tc) {
           sobj = (SwigPyObject *)sobj->next;
         } else {
@@ -1427,7 +1427,7 @@ SWIG_Python_ConvertPacked(PyObject *obj, void *ptr, size_t sz, swig_type_info *t
   if (ty) {
     if (to != ty) {
       /* check type cast? */
-      swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
+      swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
       if (!tc) return SWIG_ERROR;
     }
   }
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1314,7 +1314,7 @@ SWIG_Python_ConvertPtrAndOwn(PyObject *obj, void **ptr, swig_type_info *ty, int 
         if (ptr) *ptr = vptr;
         break;
       } else {
-        swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
+        swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
         if (!tc) {
           sobj = (SwigPyObject *)sobj->next;
         } else {
@@ -1427,7 +1427,7 @@ SWIG_Python_ConvertPacked(PyObject *obj, void *ptr, size_t sz, swig_type_info *t
   if (ty) {
     if (to != ty) {
       /* check type cast? */
-      swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
+      swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
       if (!tc) return SWIG_ERROR;
     }
   }
--- a/Lib/r/rrun.swg
+++ b/Lib/r/rrun.swg
@@ -329,7 +329,7 @@ SWIG_R_ConvertPtr(SEXP obj, void **ptr, swig_type_info *ty, int flags) {
     if (to == ty) {
       if (ptr) *ptr = vptr;
     } else {
-      swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
+      swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
       int newmemory = 0;
       if (ptr) *ptr = SWIG_TypeCast(tc,vptr,&newmemory);
       assert(!newmemory); /* newmemory handling not yet implemented */
@@ -402,7 +402,7 @@ SWIG_R_ConvertPacked(SEXP obj, void *ptr, size_t sz, swig_type_info *ty) {
   if (ty) {
     if (to != ty) {
       /* check type cast? */
-      swig_cast_info *tc = SWIG_TypeCheck(to->name,ty);
+      swig_cast_info *tc = SWIG_TypeCheckStruct(to,ty);
       if (!tc) return SWIG_ERROR;
     }
   }
--- a/Lib/scilab/scirun.swg
+++ b/Lib/scilab/scirun.swg
@@ -220,7 +220,7 @@ SwigScilabCheckPtr(void *pvApiCtx, int iVar, swig_type_info *descriptor, char *f
     }
 
     if (descriptor) {
-      swig_cast_info *cast = SWIG_TypeCheck(SWIG_TypeName((swig_type_info*)pvTypeinfo), descriptor);
+      swig_cast_info *cast = SWIG_TypeCheckStruct((swig_type_info*)pvTypeinfo, descriptor);
       return (cast != NULL);
     }
     else {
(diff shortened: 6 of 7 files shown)


wsfulton:
I don't suppose you could provide the speed up if you just run `make partialcheck-php-test-suite`? The partialcheck is designed to just execute swig and skip everything after the swig invocation in the test-suite.

ojwb:
This is part of the runtime type system so I'd expect essentially no speed
difference for partial check because that won't actually run the changed
code.

Or have I misunderstood something here?

wsfulton:
Oh sorry, no, my mistake. What should work instead is this:
```
make check-php-test-suite
time make check-php-test-suite SWIGTOOL=true COMPILETOOL=true CC=true CXX=true
```
The latter should skip all the swig and compiler invocations and only invoke the php interpreter. (COMPILETOOL is actually only  necessary for the non-scripting target languages).

ojwb:
That seems to work and takes just under 6 seconds, but I tested it on this branch vs master (which shows essentially no difference because there's no relevant code change) whereas I should have tested 69b580dfb6e39b5463213908b443a86265bdb23c and its parent.

I need to AFK but will try those later.

ojwb:
I ended up testing `git master` (3200a815d4dbd40cdd65e8234b64883c82308a0f) vs that commit after `git revert 69b580dfb6e39b5463213908b443a86265bdb23c`.  A notable change since the original commit is I've added `_runme.php` files for another 10 testcases (taking us from 85 to 95).  There's also a new testcase (`preproc_cpp.i`).

I seem to be getting a lot of noise in the results, perhaps because it's quite a short test now.

I was concerned that using `true` might actually end up starting up subshells, etc (I know GNU `make` has some optimisations, but I'm not sure if this is one), so I tried using `:` instead of `true` and also set `SWIG_LIB_SET` (with your command we'd end up running `env FOO=bar true ./swig` [...] a lot.  But even with that I see a lot of variation - here I'm ignoring the first run (in case that's priming the disk and OS page cache) and timed runs 2 to 4 in each case:

```
After change:

$ time make -s check-php-test-suite SWIGTOOL=: COMPILETOOL=: CC=: CXX=: SWIG_LIB_SET=: >/dev/null 2>&1
real	0m5.530s
user	0m4.218s
sys	0m1.389s
$ time make -s check-php-test-suite SWIGTOOL=: COMPILETOOL=: CC=: CXX=: SWIG_LIB_SET=: >/dev/null 2>&1
real	0m5.266s
user	0m4.081s
sys	0m1.263s
$ time make -s check-php-test-suite SWIGTOOL=: COMPILETOOL=: CC=: CXX=: SWIG_LIB_SET=: >/dev/null 2>&1
real	0m5.486s
user	0m4.270s
sys	0m1.306s


Before change:

$ time make -s check-php-test-suite SWIGTOOL=: COMPILETOOL=: CC=: CXX=: SWIG_LIB_SET=: >/dev/null 2>&1
real	0m5.563s
user	0m4.249s
sys	0m1.401s
$ time make -s check-php-test-suite SWIGTOOL=: COMPILETOOL=: CC=: CXX=: SWIG_LIB_SET=: >/dev/null 2>&1
real	0m5.438s
user	0m4.170s
sys	0m1.349s
$ time make -s check-php-test-suite SWIGTOOL=: COMPILETOOL=: CC=: CXX=: SWIG_LIB_SET=: >/dev/null 2>&1
real	0m5.511s
user	0m4.283s
sys	0m1.316s
```

If I compare the sums of `user` before and after (`real` is probably more prone to varying due to other random load on the machine, and we'd expect the only genuine change to be in user time as the key change here is `strcmp()` vs a pointer comparison so any variation in `sys` seems like it can't be meaningful) I get a 6.8% speed-up, but I'm not confident in the precision of that given how much variation there is between runs.

It also doesn't make sense that this is less than the 10% I'd measured before, as we've stripped out a lot of overhead.  It doesn't seem likely the 10 new runmes could make that much difference to the overall result.

Perhaps we need to find a single testcase which actually exercises SWIG run time type system machinery a lot, and then inspect that as a single process.  Any thoughts on a likely candidate?  If not I can rig up a loop to try each `_runme.php` in turn.

ojwb:
I wondered if we might have addressed some underlying cause of the failures for Python and/or Octave so I reverted 402a81b5b07ab0412a90155d9e67af3e9a7cbf6e and rebased on current git master.

Python still fails the testsuite, but only testcase `mod` (not seeing a note of what failed before so not sure if that's changed) which is a multi-module testcase.  I'm suspecting in the multi-module case we end up with two `swig_type_info` entries for some types and the pointer comparison fails but the string comparison succeeds.

Octabe also fails on `mod` and passes on most, perhaps all of the rest of the testsuite (it's really slow and hasn't finished running yet).

I notice that only Go, Octave and Python have a `mod_runme`, and Go's has an extra explicit conversion step:

```
        c := mod_b.NewC()
        d := mod_b.NewD()
        d.DoSomething(mod_a.SwigcptrA(c.Swigcptr()))
```

So maybe this points to `SWIG_TypeCheckStruct()` being buggy in the multi-module case.

I think this deserves investigating a bit before 4.1.0 - it looks like currently guile, javascript, lua, mzscheme, ocaml and php use `SWIG_TypeCheckStruct` and php switched to it since 4.0.2 so this would be a regression for php at least.  I'll write a `mod_runme.php` and maybe some others if I can.

It's likely that javascript escapes any problems by falling back to `SWIG_TypeCheck` if `SWIG_TypeCheckStruct` returns `false` though.

Setting the milestone for testing and if necessary resolving this last part - to be clear I'm certainly not proposing to switch anything more to using `SWIG_TypeCheckStruct` for this release.

ojwb:
Currently the `mod` testcase won't work for PHP because `B` isn't a known class to SWIG when wrapping `mod_b` so it's ignored as a base class for `C`.  That's clearly a bug.  I haven't checked yet if this worked in SWIG 4.0.2, but if it didn't I think it's best to punt on trying to fix this for SWIG 4.1.0.

I think to fix this we can get the `zend_class_entry*` which is already stored in the `clientdata` of the `swig_type_info` entry.  That should work provided `mod_a` is loaded before `mod_b`.  I'm not sure how to ensure that happens though...

The other approach would be to look up by name, but that's awkward to do.  The problem is that classes are initialised at "minit" (module init), but we can't look up a class by name until "rinit" (request init) time (the PHP API call segfaults at minit).  "rinit" happens on every page load, so it's much better to initialise at module init where possible.

For interfaces we initialise those we can at minit and delay ones we'd have to lookup by name until rinit, but that works because we can add an interface to an already registered PHP class, but it seems subclasses have to be specified when the class is created.  If we have to move a particular class registration to rinit, then any subclasses will also need moving.  It's probably doable, but is likely to be fiddly and harms runtime performance.

ojwb:
`mod` is broken for lua too and also fixed by using `SWIG_TypeCheck()` instead of `SWIG_TypeCheckStruct()`.

@wsfulton Are you happy to also fix these for 4.1.0 for cases which aren't regressions?  They seem small and safe fixes - for lua it's a single line change plus a new `mod_runme.lua` for a regression test and I'd imagine it'll be similar for any other affected languages:

```
--- a/Lib/lua/luarun.swg
+++ b/Lib/lua/luarun.swg
@@ -1789,7 +1789,7 @@ SWIGRUNTIME int  SWIG_Lua_ConvertPtr(lua_State *L,int index,void **ptr,swig_type
     }
     else
     {
-      cast=SWIG_TypeCheckStruct(usr->type,type); /* performs normal type checking */
+      cast=SWIG_TypeCheck(usr->type->name,type); /* performs normal type checking */
       if (cast)
       {
         int newmemory = 0;
```

For PHP this was a regression compared to 4.0.2.  If Javascript is broken that may be too as there's a new use of `SWIG_TypeCheckStruct()` introduced by 6aef217438328f6653be270fa0e2296a5254a89c.  The other target languages which use it have been doing so for a long time.

ojwb:
> guile, javascript, lua, mzscheme, ocaml and php

Guile: Not looked at (I don't have the dev packages installed already)

Javascript: mod_runme.js (which passes) added in ea514c39615f616ced604ecbef4b8f42aee8148e

Lua: #2408

mzscheme: Not looked at (I don't have the dev packages installed already)

Ocaml: mod_runme.ml (which passes) added in 867e49d7c50aded9afdd36223f05b84645636f15 (interesting that `SWIG_TypeCheckStruct()` seems OK for Ocaml...)

PHP: addressed by 747a6a264f56711c7b725cf883ef623e01b65922

wsfulton:
My first instinct is that these are changes to the heart of the all important runtime and not suitable for this stage of release. My other first reaction is given you have a reproducible example, it should surely be easy to debug SWIG_TypeCheckStruct - it's a rather small function and almost identical to SWIG_TypeCheck. Do whatever you need to in PHP given the overall changes you've introduced though, but I'd delay changes in other languages until after 4.1.0.

Sorry, I havn't looked been able to look at this in detail to offer any more informed opinion, I'm not overly familiar with the multiple modules feature. I'll try find some time to get back on this in the next couple of days.

ojwb:
Oh sorry, the notification for your comment only just arrived for some reason and I'd already merged the lua fix.

> it should surely be easy to debug SWIG_TypeCheckStruct - it's a rather small function and almost identical to SWIG_TypeCheck.

I don't entirely understand this stuff, but it seems it fails to work in this case because when it searches based on the `swig_type_info*` for the type, there are two `swig_type_info` entries for this type - one from `mod_a_wrap.cxx`:

```
static swig_type_info _swigt__p_A = {"_p_A", "A *", 0, 0, (void*)&_wrap_class_A, 0};
```

The other is from `mod_b_wrap.cxx`:

```
static swig_type_info _swigt__p_A = {"_p_A", "A *", 0, 0, (void*)0, 0}
```

The two will have different addresses so `SWIG_TypeCheckStruct()` won't match if the object is created with one but searched for with the other, but `SWIG_TypeCheck()` will match because it checks the `name` member (the first member of the struct, so `_p_A` here).

I don't see how this was ever meant to work - maybe it was just an oversight that this optimisation doesn't always work in the multimodule case.

wsfulton:
@ojwb, let's kick this one down the road for 4.2??

ojwb:
@wsfulton That's OK with me I guess, though I'd welcome a second opinion from you on my analysis in my most recent comment here.

It seems to me that `SWIG_TypeCheckStruct()` is actually an attempt at optimising based on a premise which is (at least currently) flawed - if so either it should be dropped entirely (or made opt-in for when you definitely don't care about the multi-module case), or we need to fix things so this premise is actually correct (if that's possible - perhaps the typeinfo could be reconciled on module load so each name only occurs once across currently loaded modules?)

wsfulton:
c3338b1a1686503721583a9e53a2f78a7a88731f introduced `SWIG_TypeCheckStruct` and `SWIG_TypeCheck` and describes a bit about the motivation in Doc/Devel/runtime.txt.

I tweaked Python to use `SWIG_TypeCheckStruct` and fallback to `SWIG_TypeCheck` if that fails and the only time the `SWIG_TypeCheck` fallback worked when `SWIG_TypeCheckStruct` failed was also for the mod multicpptest.

I don't have any motivation to delve into the multi-module support to advise
too much here. For a low effort amount of development, I suggest implementing the fallback
approach in all languages. I think this will improve performance in for
normal use case, with a performance hit on error. This is generally in line
with C++ philosophy and exceptions, where high performance is valued in the normal use case,
and a big performance hit is taken on error throwing the exception.

> @wsfulton That's OK with me I guess, though I'd welcome a second opinion from you on my analysis in my most recent comment here.

> It seems to me that `SWIG_TypeCheckStruct()` is actually an attempt at optimising based on a premise which is (at least currently) flawed - if so either it should be dropped entirely (or made opt-in for when you definitely don't care about the multi-module case), or we need to fix things so this premise is actually correct (if that's possible - perhaps the typeinfo could be reconciled on module load so each name only occurs once across currently loaded modules?)

Yes, so it seems. This suggested fix would be a more robust approach.
A quick fix might be to only use the fallback approach if the module knows that a second module has been loaded.
One thing that should be looked at in more detail is the `-external-runtime`
option. The only example I can find using this option is Examples/lua/dual and
Examples/lua/embed3. I think this is meant to be the proper way to use a
common runtime. It might be worth experimenting with this example.

Probably the best way to measure performance is to write a simple tool that
generates some simple classes and then check the runtime lookups using one of
them. Try generating say, 10, 100, 1000, 10000, ... classes to feed into SWIG
and then check performance looking up an instance of one of these classes
passed into a SWIG wrapper.

ojwb:
> I tweaked Python to use `SWIG_TypeCheckStruct` and fallback to `SWIG_TypeCheck` if that fails and the only time the `SWIG_TypeCheck` fallback worked when `SWIG_TypeCheckStruct` failed was also for the mod multicpptest.

Can you show the code for this?

wsfulton:
[typecheck-debugging.diff.txt](https://github.com/swig/swig/files/13423526/typecheck-debugging.diff.txt)

ojwb:
Thanks for showing the diff.

> I think this will improve performance in for normal use case, with a performance hit on error

The problem I see with this logic is that the situation where `SWIG_TypeCheckStruct()` fails to work but `SWIG_TypeCheck()` does work is not an error case - it's just a particular case that can happen when there are multiple SWIG-wrapped modules loaded.  I think you also need one module to import types from the other, and maybe the load order matters.  While one could argue this is a corner case, it isn't an error case and is probably actually the normal case for some SWIG-generated bindings so making it deliberately slower to gain performance in other cases doesn't seem ideal.

I think the ideal solution would be to resolve the typeinfo for entries like this when loading a module so there's a single entry per type (which have a non-NULL pointer if any module has been loaded which actually defines the type):

```
static swig_type_info _swigt__p_A = {"_p_A", "A *", 0, 0, (void*)0, 0}
```

That should then allow `SWIG_TypeCheckStruct()` to work reliably.

It looks doable - instead of generating struct initialisers containing `&_swigt__p_A`, etc we could generate code which stores the pointer to the resolved equivalent and run that on module load after resolving the typeinfo with any SWIG-generated modules which have already been loaded.  I think it could even interwork with modules generated with existing SWIG versions.

wsfulton:
> I think the ideal solution would be to resolve the typeinfo for entries like this when loading a module ...

Agreed, but sounds non-trivial amount of work. I can't see this being implemented quickly, so am removing it from the 4.2 milestone.

ojwb:
> Agreed, but sounds non-trivial amount of work. I can't see this being implemented quickly, so am removing it from the 4.2 milestone.

Yes, I don't think we should hold up 4.2.0 for this.

PR #2137, opened 2021-12-31
Python: Add option to generate flattened static class methods

jschueller:
Like #2093 except that the option is to enable static methods, the default becoming no static methods, for the zen of Python.

cc @jim-easterbrook @wsfulton @casperdcl

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -2537,7 +2537,7 @@ public:
   /* ------------------------------------------------------------
    * dispatchFunction()
    * ------------------------------------------------------------ */
-  void dispatchFunction(Node *n, String *linkage, int funpack = 0, bool builtin_self = false, bool builtin_ctor = false, bool director_class = false) {
+  void dispatchFunction(Node *n, String *linkage, int funpack = 0, bool builtin_self = false, bool builtin_ctor = false, bool director_class = false, bool use_static_method = false) {
     /* Last node in overloaded chain */
 
     bool add_self = builtin_self && (!builtin_ctor || director_class);
@@ -2641,11 +2641,11 @@ public:
     Printv(f->code, "}\n", NIL);
     Wrapper_print(f, f_wrappers);
     Node *p = Getattr(n, "sym:previousSibling");
-    if (!builtin_self && (flat_static_method || !in_class || !builtin))
+    if (!builtin_self && (use_static_method || !builtin))
       add_method(symname, wname, 0, p);
 
     /* Create a shadow for this function (if enabled and not in a member function) */
-    if (!builtin && (shadow) && (!(shadow & PYSHADOW_MEMBER)) && (flat_static_method || !in_class)) {
+    if (!builtin && (shadow) && (!(shadow & PYSHADOW_MEMBER)) && (use_static_method)) {
       emitFunctionShadowHelper(n, in_class ? f_shadow_stubs : f_shadow, symname, 0);
     }
     DelWrapper(f);
@@ -3312,18 +3312,19 @@ public:
       Wrapper_print(f, f_wrappers);
     }
 
+    bool use_static_method = flat_static_method || !in_class || (Cmp(storage, "friend") == 0);
     /* Now register the function with the interpreter.   */
     if (!Getattr(n, "sym:overloaded")) {
-      if (!builtin_self && (flat_static_method || !in_class || !builtin))
+      if (!builtin_self && (use_static_method || !builtin))
 	add_method(iname, wname, allow_kwargs, n, funpack, num_required, num_arguments);
 
       /* Create a shadow for this function (if enabled and not in a member function) */
-      if (!builtin && (shadow) && (!(shadow & PYSHADOW_MEMBER)) && (flat_static_method || !in_class)) {
+      if (!builtin && (shadow) && (!(shadow & PYSHADOW_MEMBER)) && (use_static_method)) {
 	emitFunctionShadowHelper(n, in_class ? f_shadow_stubs : f_shadow, iname, allow_kwargs);
       }
     } else {
       if (!Getattr(n, "sym:nextSibling")) {
-	dispatchFunction(n, linkage, funpack, builtin_self, builtin_ctor, director_class);
+	dispatchFunction(n, linkage, funpack, builtin_self, builtin_ctor, director_class, use_static_method);
       }
     }
 
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -3312,7 +3312,7 @@ public:
       Wrapper_print(f, f_wrappers);
     }
 
-    bool use_static_method = flat_static_method || !in_class || (Cmp(storage, "friend") == 0);
+    bool use_static_method = flat_static_method || !in_class || constructor || (Cmp(storage, "friend") == 0);
     /* Now register the function with the interpreter.   */
     if (!Getattr(n, "sym:overloaded")) {
       if (!builtin_self && (use_static_method || !builtin))
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -93,6 +93,7 @@ static int castmode = 0;
 static int extranative = 0;
 static int nortti = 0;
 static int relativeimport = 0;
+static int flat_static_method = 0;
 
 /* flags for the make_autodoc function */
 namespace {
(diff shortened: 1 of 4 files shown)


casperdcl (approved):
lgtm

ojwb:
I've closed #2093 in favour of this and copied over the milestone.

wsfulton:
I have added 761099720966c2d6701a1dc1473e918de3571665...2272d00c1a60ea6f3710e4d9dc17746b9df69780 as preparatory work for this patch. It removes the (soon to be) legacy flattened static method syntax from all the examples and testcases. I've added an additional testcase, python_flatstaticmethod, which tests both the flattened static method syntax as well as the normal staticmethod access syntax. I also had to fix two problems around `%callback` and the `kwargs` feature using the normal staticmethod syntax.

In order to accept this patch, the next steps are:
1. Apply this patch onto master HEAD.
2. Revert the `-flatstaticmember` change which made to Examples/Makefile.in in the patch. This restores the testing to the default SWIG options.
3. Please rename `-flatstaticmember` to `-flatstaticmethod`. I prefer this name as C++ class static member methods are wrapped as Python class static methods (`@staticmethod`). 'member' is generic and can include member variables and the option does not apply to member variables, just member methods.
4. Uncomment the following to test the new option https://github.com/swig/swig/blob/2272d00c1a60ea6f3710e4d9dc17746b9df69780/Examples/test-suite/python/Makefile.in#L101
5. Unfortunately the test-suite will not then pass. It highlights bugs in the patch that need fixing. Failed tests (using -builtin option) are the friends and rename_scope testcases. I think the problem is something around `%rename`. I think the logic in the patch for python.cxx will need re-examining.

jschueller:
could you have a look @jim-easterbrook ?

jim-easterbrook:
I'm not sure I can deal with this. I don't really understand how SWIG works and had made what I thought were the minimum changes needed to remove the static method flattened versions. I didn't realise there'd be problems with `%rename`. I'll see if I can get my head round that, and how to run the test suite.

This is really a very low priority for SWIG - it isn't fixing a significant problem.

jschueller:
ok
to run the test-suite, you can run `make check-python-test-suite`, eventually with `SWIG_FEATURES=-builtin` to enable builtin

ojwb:
Sorry, I thought everything wsfulton had asked for had been done, but I see we're actually still at the final step:

> 5. Unfortunately the test-suite will not then pass. It highlights bugs in the patch that need fixing. Failed tests (using -builtin option) are the friends and rename_scope testcases. I think the problem is something around %rename. I think the logic in the patch for python.cxx will need re-examining.

jschueller:
yes, still no luck with that

jim-easterbrook:
I apologise for not making any progress with this. I'll try and find time to look at it in the next week or so.

jschueller:
do you have a branch I can push here ?

jim-easterbrook:
I've forked `jschueller:flatstaticmember` and was going to open a PR when I've got everything working OK.
BTW, I'm running the test suite with Python 2 (the default on my machine) and haven't found out how to run it with Python 3.

jschueller:
for Foo_int, it looks like it comes from line ~-4985:
```
Printf(f_shadow_stubs, "%s = %s\n", symname, subfunc);
```
maybe it should be guarded by an "if flat"

jim-easterbrook:
I've just tried commenting it out - now I get failures in `constructor_rename` and `overload_rename` so I'm in the right area.  (:

jim-easterbrook:
This is all getting too deep into SWIG for me. Given
```
%rename(Foo_int) Foo::Foo(float a, int c, float b=1.0);
class Foo {
public:
  Foo(float a, float b=1.0);
  Foo(float a, int c, float b=1.0);
};
```
SWIG generates functions `_wrap_new_Foo` and `_wrap_new_Foo_int`. `_wrap_new_Foo` is used as `tp_init` for the `Foo` type object, but `_wrap_new_Foo_int` has to be kept in the interface as a static function so you can call `Foo_int` as an alternative constructor.

jim-easterbrook:
Success! Enabling flat static methods for constructor methods does the trick. The python_richcompare test is still failing with -builtin, but that's not because of not having flat static methods.

ojwb:
> BTW, I'm running the test suite with Python 2 (the default on my machine) and haven't found out how to run it with Python 3.

`PY3=1` (or any non-empty value I think) - e.g.:

```
make check-python-test-suite PY3=1
```

Or:

```
make -C Examples/test-suite/python director_basic.cpptest PY3=1
```

And yes, running tests as Python3 really should be the default now, with `PY2=1` or similar to select Python2.  I have a partial patch for that, but I'd like to find a good approach to replace the trick CI currently relies on of setting `PY3=3` and then doing stuff like:

```
Tools/CI-linux-install.sh:                      WITHLANG=$SWIGLANG$PY3=$SWIGLANG$VER
Tools/CI-linux-install.sh:                      $RETRY sudo apt-get install -qq python${PY3}-dev
Tools/CI-linux-install.sh:                      WITHLANG=$SWIGLANG$PY3
appveyor.yml:- if not "%OSVARIANT%"=="" bash -c "exec 0</dev/null && ./autogen.sh && time ./configure CC=%CC% CXX=%CXX% --without-alllang --with-$SWIGLANG$PY3$SWIGWITHLANG --enable-cpp11-testing || cat config.log"
```

Hmm, given that's all in CI-specific code maybe just continuing to set `PY3=3` for CI still is the way to go, but replace the uses outside of CI (and have CI set `PY2=2` for the Python2 builds which is then what actually selects the Python version for the testsuite and examples.

jschueller:
@ojwb this should be ok now

wsfulton:
I'll have another look today if this is ready now.

Source/Modules/python.cxx
```diff
@@ -3307,18 +3312,19 @@ class PYTHON:public Language {
       Wrapper_print(f, f_wrappers);
     }
 
+    bool use_static_method = flat_static_method || !in_class || constructor || (Cmp(storage, "friend") == 0);
```
wsfulton, line 3315: The variable name is confusing and subsequent logic is hard to follow to me. You really ought to be using a flag based on whether the method is a static method, so something like:
```
bool generate_flat_static_method = flat_static_method && Swig_storage_isstatic(n);
```
and then adjust the subsequent logic on this.

Swig_storage_isstatic(n) is simply checking the "storage" attribute contains "static" and is always set for static methods.

Source/Modules/python.cxx
```diff
@@ -2636,11 +2641,11 @@ class PYTHON:public Language {
     Printv(f->code, "}\n", NIL);
     Wrapper_print(f, f_wrappers);
     Node *p = Getattr(n, "sym:previousSibling");
-    if (!builtin_self)
+    if (!builtin_self && (use_static_method || !builtin))
       add_method(symname, wname, 0, p);
 
     /* Create a shadow for this function (if enabled and not in a member function) */
-    if (!builtin && (shadow) && (!(shadow & PYSHADOW_MEMBER))) {
+    if (!builtin && (shadow) && (!(shadow & PYSHADOW_MEMBER)) && (use_static_method)) {
```
wsfulton, line 2648: Brackets are superfluous here (they shouldn't be around `shadow` either).

wsfulton (changes requested):
When using `-builtin` with this patch, the disown_XXXX methods are missing, eg in the apply_signed_char testcase, this is missing:
```
{ "disown_DirectorTest", _wrap_disown_DirectorTest, METH_O, NULL},
```

Reminder to self for next review: check effects when using`%extend`.

jim-easterbrook:
This is all proving to be much more difficult than I anticipated. I'm way out of my comfort zone and don't think I can make any useful further contribution. I'm sorry to have put everyone to so much trouble.

wsfulton:
This is more logical and seems to work:
```
-    bool use_static_method = flat_static_method || !in_class || constructor || (Cmp(storage, "friend") == 0);
+    bool use_static_method = flat_static_method || !Swig_storage_isstatic_custom(n, "staticmemberfunctionHandler:storage");
```
I had to use `Swig_storage_isstatic_custom` instead of `Swig_storage_isstatic` because I overlooked that the parse tree is temporarily modified by `Language::staticmemberfunctionHandler`.

I have some more checking and tweaks to make and all going well will then merge it in.

jschueller:
thanks a lot

PR #2138, opened 2022-01-04
PCRE2

### LINKED ISSUE #2120: SWIG depends on obsolete pcre library
Hi *,

according to https://pcre.org/, the pcre library version that SWIG is currently using is deprecated. Quote (emphasis added):

> current version, PCRE2, released in 2015, is now at version 10.39.
> The older, but still widely deployed PCRE library, originally released in 1997, is at version 8.45. This version of PCRE is now at end of life, and *is no longer being actively maintained*. Version 8.45 is expected to be the final release of the older PCRE library, and new projects should use PCRE2 instead.

This has been reported in Debian as https://bugs.debian.org/999948. Debian is planning to remove the old pcre from Debian for the bookworm release, but swig has a build dependency on it.

I intend to create a pull request to update the pcre dependency, but won't stop anybody else to work on this ;-)

Greetings, Torsten

jschueller:
Closes #2120


vadz (commented):
Thanks for these changes! FWIW everything here looks good to me, but I think you also need to update the package being installed in `Tools/CI-linux-install.sh` and maybe `Tools/mkwindows.sh`, if it's still used.

wsfulton:
Did you see the Github CI failures related to this change, such as:
```
SWIG:1: Error: PCRE capture replacement failed while matching "wx(?!EVT)(.*)" using "m_width" - request for group 1 is greater than the number of captures 0.
SWIG:1: Error: PCRE compilation failed: 'unmatched closing parenthesis' in '(.*?)(nsa)(.*?)\2(.*?)\2(.*?)\2(.*)':1.

SWIG:32: Error: Matching "Foo" against regex "[A-Z]\w*[A-Z]" failed: -51
```

Source/Swig/misc.c
```diff
@@ -1477,12 +1483,12 @@ String *Swig_string_regex(String *s) {
...
   DohDelete(pattern);
   DohDelete(subst);
-  pcre_free(compiled_pat);
+  pcre2_code_free(compiled_pat);
   return res ? res : NewStringEmpty();
 }
 
 String *Swig_pcre_version(void) {
-  return NewStringf("PCRE Version: %s", pcre_version());
+  return NewStringf("PCRE Version: %d.%d", PCRE2_MAJOR, PCRE2_MINOR);
```
ojwb, line 1491: This now reports the PCRE2 version compiled against, whereas before it would report the PCRE version in use at runtime (which may be different to the one compiled against if libpcre2 is a shared library).   The version actually being used is more interesting as bugs are much more likely to be in the library code than inlined by headers.

It looks like the equivalent to `pcre_version()` is something like (untested):

```
    int len = pcre2_config(PCRE2_CONFIG_VERSION, NULL);
    char *buf = malloc(len);
    String *result;
    pcre2_config(PCRE2_CONFIG_VERSION, buf);
    result = NewString(buf);
    free(buf);
    return result;
```

wsfulton:
@jschueller thanks for taking on this change. Now that Travis tests are working, and it looks like you are figuring out the Appveyor changes, can you stop force pushing. In theory pushes to this branch changing just appveyor.yml should not (unnecessarily) do the very compute intensive GHA build because of:
https://github.com/swig/swig/blob/d73f933c4bb5ca1f327870554f75a87372964162/.github/workflows/ci.yml#L3-L8.
This is not working, but I think it is because you are force pushing. If you just add additional commits to just appveyor.yml, just the Appveyor CI should run. In my experience, it can take a lot of commits before configuring appveyor.yml correctly. You could also do this in your own github repo if you don't want us to see every small attempt at getting it right and then push the final working version here. If you need help with Appveyor, I can add it to my queue of work.

wsfulton:
I tested the distribution build with these changes and they work okay and the windows build seems to work with pcre2-10.39.tar.bz2.

jschueller:
yes I wasnt too successful, seems like I have to build pcre from source but didtn manage to build something compatible
you can have a go
```
appveyor-retry appveyor DownloadFile https://github.com/PhilipHazel/pcre2/archive/refs/tags/pcre2-10.39.zip
7z x pcre2-10.39.zip
```

wsfulton:
> yes I wasnt too successful, seems like I have to build pcre from source but didtn manage to build something compatible
you can have a go

Am confused! I successfully got a Windows build. What do you want me to have a go at? BTW, I also checked that your changes to pcre-build.sh works on an Ubuntu box (this script builds pcre2 from source, run `Tools/pcre-build.sh --help` for more info).

jschueller:
I had pcre built with msvc (with cmake) but then it failed to link when used in swig (a cryptic message about crt libs),
maybe you know better which tools/options must be used

jschueller:
I had pcre built with msvc (with cmake) but then it failed to link when used in swig (a cryptic message about crt libs),
maybe you know better which tools/options must be used

wsfulton:
I see. CMake is a secondary build system and not mandatory to keep working. Best left to those who rely on it to fix it.

jschueller:
swig's  cmake is ok, I just couldnt build an usable pcre on appveyor

wsfulton:
I see, but obtaining and using the dependencies are just as important as building SWIG itself. The instructions in Windows.html are meant to be a set of simple get going instructions including obtaining and using PCRE. Maybe those will help inspire what you need to do for PCRE2.

jschueller:
@wsfulton I think I got it right, the only remaining appveyor failures look like in master

wsfulton:
Looking good. I've put in a fix for mingw on master. I missed it going wrong on 1 Jan when the mingw folk re-organised the their packages. Could you please fix the consequent conflict and push again, then hopefully appveyor testing for the pull request will go green. Note that cygwin Appveyor testing on master are marked as `allow_failures`, I don't know what is wrong there.

wsfulton:
All tests passed. Yay! The only thing that really needs doing now is to patch up Windows.html to reflect the kind of changes you have put into Appveyor.yml. It's real shame that nuget does not provide a pcre2 package like it does pcre, otherwise the docs would be much easier to update.

wsfulton:
One last thing, we have Tools/CMake/FindPCRE.cmake and a new file FindPCRE2.cmake. Could you please remove FindPCRE.cmake and move FindPCRE2 into Tools/CMake. Tools/nuget-install.cmd can also be deleted now given since nuget is no longer used and you used appveyor-retry to do the retries.

jschueller:
ok, done

Issue #2051, opened 2021-07-13
Crash at Py_EndInterpreter due to negative reference count 


benjamin-sch:
If Python is embedded in an application and the sub-interpreters are used 
I get a crash at Py_EndInterpreter due to a negative reference count. 
According to the values in the debugger the object has been deleted already.

It is not necessarily the first Py_EndInterpreter call that leads to the crash.
I guess this depends on the amount of objects that are wrapped by SWIG.
Without SWIG I couldn't reproduce this issue.

Example: https://github.com/benjamin-sch/python-swig-negative-ref-cnt-example

Is this a bug in SWIG, in Python or am I using the Python API wrong?

Callstack:
```
 	KernelBase.dll!wil::details::DebugBreak(void)	Unknown	Symbols loaded.
 	python39_d.dll!fatal_error_exit(int status) Line 2201	C	Symbols loaded.
 	python39_d.dll!fatal_error(_iobuf * stream, int header, const char * prefix, const char * msg, int status) Line 2286	C	Symbols loaded.
 	python39_d.dll!_Py_FatalErrorFunc(const char * func, const char * msg) Line 2302	C	Symbols loaded.
 	python39_d.dll!_PyObject_AssertFailed(_object * obj, const char * expr, const char * msg, const char * file, int line, const char * function) Line 2199	C	Symbols loaded.
 	python39_d.dll!_Py_NegativeRefcount(const char * filename, int lineno, _object * op) Line 124	C	Symbols loaded.
>	_SwigModule_d.pyd!_Py_DECREF(const char * filename, int lineno, _object * op) Line 425	C++	Symbols loaded.
 	_SwigModule_d.pyd!_Py_XDECREF(_object * op) Line 497	C++	Symbols loaded.
 	_SwigModule_d.pyd!SwigPyClientData_Del(SwigPyClientData * data) Line 1366	C++	Symbols loaded.
 	_SwigModule_d.pyd!SWIG_Python_DestroyModule(_object * obj) Line 2412	C++	Symbols loaded.
 	python39_d.dll!capsule_dealloc(_object * o) Line 261	C	Symbols loaded.
 	python39_d.dll!_Py_Dealloc(_object * op) Line 2209	C	Symbols loaded.
 	python39_d.dll!_Py_XDECREF(_object * op) Line 497	C	Symbols loaded.
 	python39_d.dll!free_keys_object(_dictkeysobject * keys) Line 598	C	Symbols loaded.
[...]

PR #2141, opened 2022-01-12
Python subinterpreter issues

benjamin-sch:
This should avoid the crash reported in #2051 

1) the first commit counts the "capsules" used. It may be greater 1 if subinterpreters are used. As the InitializeModule seems to initialize the types only once in the case of multiple subinterpreters the deletion of the types should be as well only being done once. Otherwise we get the negative refcnt exception.
2) the second commit always gets the type_pointer from the "capsule" as it may be different after a re-initialization if SWIG is used with python embedded in an application and the application deinitializes/initializes python more than once. Furthermore it might be different if subinterpreters are used as each subinterpreter will add its own "capsule" in SWIG_SetModule.

### DIFF
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1627,19 +1627,19 @@ SWIG_Python_TypeCache(void) {
 
 SWIGRUNTIME swig_module_info *
 SWIG_Python_GetModule(void *SWIGUNUSEDPARM(clientdata)) {
+#ifdef SWIG_LINK_RUNTIME
   static void *type_pointer = (void *)0;
   /* first check if module already created */
   if (!type_pointer) {
-#ifdef SWIG_LINK_RUNTIME
     type_pointer = SWIG_ReturnGlobalTypeList((void *)0);
-#else
-    type_pointer = PyCapsule_Import(SWIGPY_CAPSULE_NAME, 0);
-    if (PyErr_Occurred()) {
-      PyErr_Clear();
-      type_pointer = (void *)0;
-    }
-#endif
   }
+#else
+  void *type_pointer = PyCapsule_Import(SWIGPY_CAPSULE_NAME, 0);
+  if (PyErr_Occurred()) {
+    PyErr_Clear();
+    type_pointer = (void *)0;
+  }
+#endif
   return (swig_module_info *) type_pointer;
 }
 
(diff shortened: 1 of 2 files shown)


Lib/python/pyrun.swg
```diff
@@ -1627,28 +1627,33 @@ SWIG_Python_TypeCache(void) {
...
+
+static int interpreter_counter = 0; // how many (sub-)interpreters are using swig_module's types
+
 SWIGRUNTIME void
 SWIG_Python_DestroyModule(PyObject *obj)
 {
   swig_module_info *swig_module = (swig_module_info *) PyCapsule_GetPointer(obj, SWIGPY_CAPSULE_NAME);
   swig_type_info **types = swig_module->types;
   size_t i;
+  if (--interpreter_counter != 0) // another sub-interpreter nay be still using the swig_module's types
```
ojwb, line 1655: comment typo "nay", should I assume be "may"

benjamin-sch: Thanks. You're right. I've fixed it.

ojwb:
This looks plausible to me and the appveyor failures are unrelated, but someone more active in the python side of things should probably review.

It should probably have a `CHANGES.current` entry.

BenjaminSbj:
> This looks plausible to me and the appveyor failures are unrelated, but someone more active in the python side of things should probably review.
> 
> It should probably have a `CHANGES.current` entry.

Thanks for the hint. I have added an entry to CHANGES.current

benjamin-sch:
> This looks plausible to me and the appveyor failures are unrelated, but someone more active in the python side of things should probably review.
> 
> It should probably have a `CHANGES.current` entry.

Thanks for the hint. I've rebased the PR and added an entry to CHANGES.current

ojwb:
@wsfulton You should probably review this...

ojwb:
@wsfulton wrote on swig-devel:
> If there are any small patches, please tag them with this milestone for consideration for the release.

Doing so...

wsfulton:
I've merged this, please check it as there was a merge conflict I had to resolve and I don't have a way to check this code is working with multiple interpreters. On that note, I've not come across sub-interpreters. Are you using https://docs.python.org/3/c-api/init.html#sub-interpreter-support ? If so, I might add a simple mention of these APIs to your new interpreter_counter code.

PR #2144, opened 2022-01-18
[perl] Require at least Perl 5.8.0

ojwb:
As discussed and agreed in #1629, it's become hard to test with Perl
5.6 or earlier, such old versions are no longer in active use, and
4.1.0 is an appropriate time to make such a change.

I've dropped the compatibility code that was obvious to me, but there's
probably more that can be cleaned up now.


ojwb:
Appveyor failures are unrelated (and appveyor doesn't seem to test Perl anyway) so have merged.

PR #2158, opened 2022-01-22
readme : use markdown extesion and add CI status

romintomasetti:
Improving the main `README`:
* Using the MarkDown extension allows for code editors to show more appropriate previews of the file. Also, it helps GitHub show a nice view of it.
* Using the MarkDown alternate synthax for headings, see https://www.markdownguide.org/basic-syntax/#alternate-syntax
* Using links for the files with `[]()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,7 @@
+[![CI](https://github.com/swig/swig/actions/workflows/ci.yml/badge.svg)](https://github.com/swig/swig/actions/workflows/ci.yml)
+
 SWIG (Simplified Wrapper and Interface Generator)
+=================================================
 
 Version: 4.1.0 (in progress)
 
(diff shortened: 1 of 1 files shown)


wsfulton:
The README file is referenced in various places and copied around in various scripts in the swig/swig repo as well as swig/www repo for the swig.org website, so this patch as it is will break a few things. I don't know how to display markdown files on the swig.org website, it's only got pretty basic features! A link to the README.md file on Github, updated for each release could be a replacement to the plain text file. Not sure that markdown is retro enough for the 1990s look and feel of swig.org though :)

ojwb:
The renaming can be avoided - if it's left as `README` and a symlink `.github/README.md` -> `../README` is added to the repo then github will render it as markdown (at least this works for `.rst` - we use that for the xapian repo - I assume it must work for `.md` too).

The copy on the website would need to either just accept the raw formatting, or be run through something that produces HTML, or turned into a link to the file here.

wsfulton:
I can't see the website any time soon being updated to support markdown, so closing.

Issue #1112, opened 2017-10-04
Docker image for swig


vaibhavsood:
Hi,

Want to check if there is a docker image/Dockerfile for swig which is officially supported? I could not find one under dockerhub official images https://hub.docker.com/explore/

If not, want to check if there is any plan to add an official swig image to dockerhub? https://docs.docker.com/docker-hub/official_repos/#how-do-i-create-a-new-official-repository

ojwb:
AFAIK, there isn't and there are no such plans.

What do you see it being useful for?  SWIG isn't some complex platform or web application - it's a single self-contained tool, and a whole docker image for a single tool doesn't seem a helpful approach.

We also already have more to do than there is developer time for.

wsfulton:
I also don't know of any plans. I'm quite happy to support an official image though if @vaibhavsood or someone can maintain it and ensure the quality is 1st class. The creation and testing of the image would need to be automated in some way. I have an experimental dockerfile I once created for a docker image with gcc-7 to test SWIG, but you probably know more what should be in the dockerfile. 

- Would it make sense to test the image on Travis? (Travis does support Docker images.) 
- What would go on the image, given that SWIG is rather useless without a C/C++ compiler and target languages? SWIG can work with about 20 target languages atm. Would all the languages and compilers be installed or would it be just the tools required to build SWIG?

ojwb:
I guess a "SWIG development" image would have its uses.  However, trying to keep an image with ~20 different languages up to date seems like a lot of on-going work, so it really needs a volunteer who is actually going to commit to doing all that work for a useful length of time, or else we'll quickly end up with an image full of out of date versions of things and a growing pile of support queries about it.

rcoup:
FWIW, I've been using this container image for running specific versions of SWIG. Helps to avoid massive diffs when projects aren't using the same version as you have installed.

Take/use/adapt/whatever 👍 

https://gist.github.com/rcoup/35e11e31d069689c9c09a70cc2fc7a0d

wsfulton:
[...]

PR #2161, opened 2022-01-24
docker : adding development-ready docker image that can be used in VS…

romintomasetti:
# Development and docker

As a side quest, I added a `dockerfile` for developers. It only allows to work with Python for now, but other languages can be added by modifying the requirements (see `requirements/requirements_system.dev.txt`) according to the SWIG wiki.

Additionally, I created a `devcontainer` configuation file for those working in Visual Studio Code. Further information can be found here: https://code.visualstudio.com/docs/remote/containers.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,15 @@
+{
+    # Instructions for building the Docker image
+    "build" : {
+        "dockerfile": "dockerfile",
+        "context": ".."
+    },
+    # VS Code extensions to install
+    "extensions": [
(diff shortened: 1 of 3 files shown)


ojwb:
This would presumably address #1112.

As noted there, it really needs someone to commit to maintaining it, or else we pretty quickly end up with an out of date image and that's eventually more harmful than helpful.

ojwb:
In #1112 William said "I suggest the Docker images should be community maintained", and I tend to agree.

This has been open just under 6 months now and nobody's stepped up as committed to maintaining these images going forwards so I'm afraid I don't think we can accept this patch.

PR #2164, opened 2022-01-25
Constant expression parsing improvements

ojwb:
This greatly improves the situation with parsing `<` and `>` in constant expressions, which is a limitation that has been reported a number of times over the years.  My patch doesn't handle all cases of `<` and `>` but all reported cases I could find here and on sourceforge now work.

Also enables an existing commented out testcase for `%`.

Tests pass for PHP at least.  Assuming no show-stopper issue is found by CI I need to add a `CHANGES` entry still.  Probably deserves a mention in RELEASENOTES too.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -141,6 +141,7 @@ CPP_TEST_CASES += \
 	compactdefaultargs \
 	const_const_2 \
 	constant_directive \
+	constant_expr \
 	constant_pointers \
 	constover \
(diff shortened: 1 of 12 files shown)


ojwb:
Ooops, I had an older attempt at this in the history - given how recently I pushed the branch I've just squashed that away and force pushed.

ojwb:
Added those entries and rebased onto current master as 6f952806e4ca33f866b20d4890f42f82b9b92e3d.

PR #2168, opened 2022-01-26
Add missing defines for PCRE2 (on Windows)

geographika:
Following the updates in #2138 I've been testing the CMake install steps on Windows. 

When trying to build (with static linking to the PCRE2 library so not requiring a pcre.dll to be on the system path), I was getting the following errors:

```
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_config_8 referenced in function Swig_pcre_version [C:\
swig\build\swig.vcxproj]
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_compile_8 referenced in function Swig_string_regex [C:
\swig\build\swig.vcxproj]
naming.obj : error LNK2001: unresolved external symbol __imp_pcre2_compile_8 [C:\swig\build\swig.vcxproj]
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_code_free_8 referenced in function Swig_string_regex [
C:\swig\build\swig.vcxproj]
naming.obj : error LNK2001: unresolved external symbol __imp_pcre2_code_free_8 [C:\swig\build\swig.vcxproj]
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_match_data_create_from_pattern_8 referenced in functio
n Swig_string_regex [C:\swig\build\swig.vcxproj]
naming.obj : error LNK2001: unresolved external symbol __imp_pcre2_match_data_create_from_pattern_8 [C:\swig\build\swig
.vcxproj]
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_match_8 referenced in function Swig_string_regex [C:\s
wig\build\swig.vcxproj]
naming.obj : error LNK2001: unresolved external symbol __imp_pcre2_match_8 [C:\swig\build\swig.vcxproj]
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_match_data_free_8 referenced in function Swig_string_r
egex [C:\swig\build\swig.vcxproj]
naming.obj : error LNK2001: unresolved external symbol __imp_pcre2_match_data_free_8 [C:\swig\build\swig.vcxproj]
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_get_ovector_pointer_8 referenced in function Swig_stri
ng_regex [C:\swig\build\swig.vcxproj]
misc.obj : error LNK2019: unresolved external symbol __imp_pcre2_get_error_message_8 referenced in function Swig_string
_regex [C:\swig\build\swig.vcxproj]
[...]

Issue #2001, opened 2021-04-23
Build warnings on Windows when using Bison <3


geographika:
For info - the following warnings are logged when building with Bison 2.4.1.1 on Windows (the current version available on [nuget](https://www.nuget.org/packages/bison-win32/):

```
D:/GitHub/swig/Source/CParse/parser.c(1972): error C2220: warning treated as error - no 'object' file generated [D:\GitHub\swig\swig.vcxproj]
D:/GitHub/swig/Source/CParse/parser.c(1972): warning C4273: 'malloc': inconsistent dll linkage [D:\GitHub\swig\swig.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\ucrt\corecrt_malloc.h(97): note: see previous definition of 'malloc'
D:/GitHub/swig/Source/CParse/parser.c(1979): warning C4273: 'free': inconsistent dll linkage [D:\GitHub\swig\swig.vcxproj]
  C:\Program Files (x86)\Windows Kits\10\Include\10.0.18362.0\ucrt\corecrt_malloc.h(85): note: see previous definition of 'free'
```

Removing warnings as errors (by removing `/WX` from `"-DCMAKE_CXX_FLAGS="/WX /DPCRE_STATIC"`) still allows swig.exe to be built. 

There are no warnings when building with Bison 3.7.4 (from https://community.chocolatey.org/packages/winflexbison3). 

I'll see how easy it may be to update the nuget package, and can update the CMake build notes.

ojwb:
@geographika Does adding `#include <stdlib.h>` just after `#include <ctype.h> ` in `Source/CParse/parser.y` fix this?

geographika:
@ojwb - I just tried with the latest code and added in the include - unfortunately doesn't appear to resolve the warnings:

```
C:/swig/build/Source/CParse/parser.c(1973,24): warning C4273: 'malloc': inconsistent dll linkage [C:\swig\build\swig.vc
xproj]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\corecrt_malloc.h(101,15): message : see previous defin
ition of 'malloc' [C:\swig\build\swig.vcxproj]
C:/swig/build/Source/CParse/parser.c(1980,19): warning C4273: 'free': inconsistent dll linkage [C:\swig\build\swig.vcxp
roj]
[...]

PR #2169, opened 2022-01-26
Updated and tested Windows CMake build docs

geographika:
Following updates in #2138, I've run through the Windows CMake docs/build steps, and made the following changes:

- Updated (and tested) to Visual Studio 2019
- Updated and tested with Nuget 6.0
- Added some more hyperlinks to dependency docs
- Changed `-DPCRE2_LIBRARY=%PCRE_ROOT%/lib/pcre2-8.lib` to `pcre2-8-static.lib` - @jschueller does this make sense? Otherwise running `swig.exe` throws an error if `pcre.dll` is not on the system path. It also throws the error `LINK : fatal error LNK1181: cannot open input file 'C:\Tools\pcre2\lib\pcre2-8.lib' [C:\swig\build\swig.vcxproj]` unless PCRE2 is built with the `-DBUILD_SHARED_LIBS=ON` flag, which was not in the steps previously. 
- Makes the builds "out-of-source" - all CMake files are created in `/build` subfolders rather than mixed in with the source code
- Adds an install step to the CMake steps
- Adds the `-DCMAKE_CXX_FLAGS="/DPCRE2_STATIC"` to allow the static builds

### DIFF
--- a/Doc/Manual/Windows.html
+++ b/Doc/Manual/Windows.html
@@ -230,72 +230,69 @@ SWIG can also be compiled and run using <a href="https://www.msys2.org/">MSYS2</
 <p>
 SWIG can be built using <a href="https://cmake.org/">CMake</a> and Visual Studio rather than autotools. As with the other approaches to 
 building SWIG the dependencies need to be installed. The steps below are one of a number of ways of installing the dependencies without requiring Cygwin or MinGW.
-For fully working build steps always check the Continuous Integration setups currently detailed in the <a href="https://github.com/swig/swig/blob/master/appveyor.yml">Appveyor YAML file</a>.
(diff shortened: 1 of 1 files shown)


wsfulton:
#2138 needs resolving first. The patch would be better if there weren't any unnecessary whitespace change in it.

I'm just wondering if the bison-2 build failures are with Visual Studio 2019 and not Visual Studio 2015, see #2001? If and only if this is the case, perhaps the Visual Studio version in the docs should remain at Visual Studio 2015 until bison 3 is available on nuget?

geographika:
@wsfulton - #2168 has been closed, as the builds are fine with the CXX flag added. 
The issues in #2001 are warnings rather than build errors, but yes ideally Bison3 (and PCRE2) would be added to Nuget. 

Reference to self - looks like the files can be bundled up for Nuget as per https://digitalhouseblog.wordpress.com/2019/08/22/how-to-make-a-nuget-package-for-c/

PR #2170, opened 2022-01-26
Move %attribute documentation into the manual

### LINKED ISSUE #2166: Move %attribute documentation into the main documentation
We talked about this in #2156, here is a reminder to move the documentation.

cminyard:
It's fairly hidden where it is, put it where users can see it more
easily.

Fixes #2166

The above is the commit header.

I put it into the "Structures and unions" section.  I debated whether to put it in "Swig and C++", but since this will work on C code as well, where I put it seemed more appropriate to me.

When I regenerated the TOC it made some changes in the area around "Instructions for using the Examples with other compilers" for windows, but I removed those.


wsfulton:
The best place for this is Library.html.
```html
<li><a href="SWIG.html#Attributes">Attributes and the %attribute directive</a>
```
The link name should be prefixed with the name of the file to prevent namespace clashes in the single html file, so should be:
```html
<li><a href="SWIG.html#SWIG_Attributes">Attributes and the %attribute directive</a>
```

Just the one .html file should be in the patch, there are always various minor updates in the html which are fixed in batches.

cminyard:
> The best place for this is Library.html.

That doesn't make much sense to me.  What am I missing?  This is about access to elements in structures.  The Library section is about accessing standard C and C++ libraries, and a few other miscellaneous things kind of related to that.  It's not where I would look to find this.

I'll fix the other issues, of course.

wsfulton:
The CHANGES file talks about the attribute.i library when it was added back in 2005. Also from the header:
https://github.com/swig/swig/blob/fa8c89ddf503ceeb917d5c62f8583ee2bc569aef/Lib/attribute.i#L1-L5

An explanation is the .i files in the Lib directory comprise the SWIG Library. Unless you %include these library files, the functionality is non-existent. I'd say attribute.i contains a quirky set of directives that are probably not suitable for the **SWIG Basics** chapter. I'd put it under the **Utility Libraries** section in Library.html.

However, I do see your point of view, so by all means link forward to the Library section with a couple of sentences from somewhere that seems appropriate, such as the **Structure data members** section in SWIG.html.

cminyard:
I added another commit on top of the previous one that addresses all the issues.  It should be ready here.  I'm not sure why it failed CI, but I can't imagine it has anything to do with this change.

If you want a new request instead of having two commits, I can do that, we can close this.

PR #2172, opened 2022-01-26
Improve documentation on %rename

cminyard:
I went through the docs and couldn't find anything about how %rename overrides a previous %rename or doing a %rename("") cancels the rename.  Since those are pretty important operations, document it.

Also add a section on how to override methods in a class using the above.

### DIFF
--- a/Doc/Manual/Contents.html
+++ b/Doc/Manual/Contents.html
@@ -172,6 +172,7 @@
 <li><a href="SWIG.html#SWIG_rename_ignore">Renaming and ignoring declarations</a>
 <ul>
 <li><a href="SWIG.html#SWIG_nn29">Simple renaming of specific identifiers</a>
+<li><a href="SWIG.html#SWIG_renaming_overriding">Overriding methods with <tt>%rename</tt></a>
 <li><a href="SWIG.html#SWIG_advanced_renaming">Advanced renaming support</a>
 <li><a href="SWIG.html#SWIG_limiting_renaming">Limiting global renaming rules</a>
 <li><a href="SWIG.html#SWIG_chosen_unignore">Ignoring everything then wrapping a few selected symbols</a>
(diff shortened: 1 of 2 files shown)


wsfulton:
Thanks. I think the terminology should be _replace_ rather than _override_, as _override_ has a special meaning in C++. Documentation for what is being explained should be simple and I think showing how to replace a destructor and then talking about %delobject etc is off-topic for this section. A simpler example showing how to replace a member method would be better off in this case.

Incidentally, the better approach is to %extend with a replacement method, then have the %ignore, then the method that you want to ignore.

You have a typo `Myclass` instead of `MyClass` and there is a semicolon missing. Best take the example from something that actually runs.

Kind of related, I wonder why we have never put in an %unignore directive. Either of the following should work, but the former does not for the destructor in your example. Hmm, I better investigate.
```
#define %unignore %feature("ignore", "0")
#define %unignore %rename("")
```

cminyard:
> Thanks. I think the terminology should be _replace_ rather than _override_, as _override_ has a special meaning in C++. 

Yes, you are correct.

> Documentation for what is being explained should be simple and I think showing how to replace a destructor and then talking about %delobject etc is off-topic for this section. A simpler example showing how to replace a member method would be better off in this case.

I guess that's fine.  The destructor thing is what I was doing, but a method is just as easy.
 
> Incidentally, the better approach is to %extend with a replacement method, then have the %ignore, then the method that you want to ignore.

Ok.  I will still leave in a comment about this; it wasn't obvious to me that something in a %extend would still be ignored.

> 
> You have a typo `Myclass` instead of `MyClass` and there is a semicolon missing. Best take the example from something that actually runs.

Yeah, I was being lazy.  I'll copy it from working code.

> 
> Kind of related, I wonder why we have never put in an %unignore directive. Either of the following should work, but the former does not for the destructor in your example. Hmm, I better investigate.
> 
> ```
> #define %unignore %feature("ignore", "0")
> #define %unignore %rename("")
> ```

Indeed, that would be orthogonal.   I kind of wondered that, too, at first.

wsfulton:
Done in #2175.

PR #2175, opened 2022-01-27
Improve documentation on %rename

cminyard:
Add documentation on %rename of names replacing previous %renames of
the same name, and how to replace methods in classes using %rename.

I went through the docs and couldn't find anything about how %rename overrides a previous %rename or doing a %rename("") cancels the rename. Since those are pretty important operations, document it.

This fixes issued pointed out in the previous request, #2172.

### DIFF
--- a/Doc/Manual/SWIG.html
+++ b/Doc/Manual/SWIG.html
@@ -47,6 +47,7 @@
 <li><a href="#SWIG_rename_ignore">Renaming and ignoring declarations</a>
 <ul>
 <li><a href="#SWIG_nn29">Simple renaming of specific identifiers</a>
+<li><a href="#SWIG_renaming_replacing">Replacing methods with <tt>%rename</tt></a>
 <li><a href="#SWIG_advanced_renaming">Advanced renaming support</a>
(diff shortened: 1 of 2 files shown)


wsfulton:
Thanks. I moved the section on replacing methods to the C++ chapter as this was all about C++ and the %rename had not yet been introduced in the SWIG chapter. What could be done is to put in a replacement section, where you had the C++ class method replacement, using the same ideas but with a C identifier replacement example. Perhaps something from the **Constants** section already covered, such as replacing a constant from a `#define` macro.

Doc/Manual/SWIG.html
```diff
@@ -1841,6 +1842,24 @@ <H4><a name="SWIG_nn29">5.4.7.1 Simple renaming of specific identifiers</a></H4>
...
+some things in one file and not in another, you could do:
+</p>
+
+<div class="code">
+<pre>
+  %rename(print1) print
+  %include "header1.h" //Anything "print" in here will become "print1"
+  %rename(print2) print
+  %include "header2.h" //Anything "print" in here will become "print2"
+  %rename("") print
```
ojwb, line 1858: Doesn't a `%rename` directive need a terminating `;`?

PR #2178, opened 2022-01-28
Update Windows CMake build docs to use NuGet Bison3

geographika:
I've created a new NuGet package containing Bison 3.7.4 for Windows - https://www.nuget.org/packages/Bison/
This can now be used to avoid any build warnings (as was the case with Bison 2 - see #2001).


ojwb:
Great, thanks.

PR #2182, opened 2022-01-28
Talk about namespaces and unignoring

### LINKED ISSUE #193: %ignore "" and unignoring classes with namespaces does not work
Hello !

When using the ignore everything feature, it works with global namespace classes but not with classes within namespaces.

Attached the data to reproduce it.

``` swig
%module swigrepro
%{
%}

%ignore ""; // ignore everything

%rename("%s") Toto::Bar; // unignore Bar class in Toto namespace
%rename("%s") Toto::Bar::Bar; // unignore Bar constructor in Toto namespace
%rename("%s") Toto::Bar::~Bar; // unignore Bar destructor in Toto namespace
%rename("%s") Toto::Bar::foo; // unignore Bar foo method in Toto namespace

%rename("%s") BarAlone; // unignore BarAlone class
%rename("%s") BarAlone::BarAlone; // unignore BarAlone constructor
%rename("%s") BarAlone::~BarAlone; // unignore BarAlone destructor
%rename("%s") BarAlone::foo; // unignore BarAlone foo method

namespace Toto
{
  class Bar
  {
  public:
      Bar();
      ~Bar();
      double foo();
  };
}

class BarAlone
{
public:
    BarAlone();
    ~BarAlone();
    double foo();
};
```

And the command to get the resulting data is

``` shell
swig -c++ -java -package com.swig.bug -verbose swigrepro.i
```

Thanks,
Julien.

cminyard:
It was not obvious to at least one person that namespaces need to be
unignored just like classes and methods.  Add an explicit reference to
that in the docs.

Also add something to unignore all classes in the example that shows how
to ignore all classes, as that might not be obvious.

Fixes #193

### DIFF
--- a/Doc/Manual/SWIG.html
+++ b/Doc/Manual/SWIG.html
@@ -2253,6 +2253,21 @@ the following approach could be taken:
 </div>
 
 <p>
+If <tt>Star</tt> was in the <tt>Galaxy</tt> namespace, you would need
+to unignore the namespace, too, and add the namespace to all the
+renames:
+</p>
+
+<div class="code">
+<pre>
+%rename("%s") Galaxy;
+%rename("%s") Galaxy::Star;
+%rename("%s") Galaxy::Star::Star;
+...
+</pre>
+</div>
+
+<p>
 Another approach which might be more suitable as it does not require naming all the methods in the
 chosen class is to begin by ignoring just the classes. This does not add an explicit ignore to any
 members of the class, so when the chosen class is unignored, all of its methods will be wrapped.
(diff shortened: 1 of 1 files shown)


Doc/Manual/SWIG.html
```diff
@@ -2263,6 +2278,7 @@ <H4><a name="SWIG_chosen_unignore">5.4.7.5 Ignoring everything then wrapping a f
 %rename($ignore, %$isclass) ""; // Only ignore all classes
```
ojwb, line 2278: This existing example here is actually wrong - it needs to be `"$ignore"` (i.e. in quotes or else you get a parse error).

ojwb: OK - solved - I'd failed to put a name after `%module` in the line above in my `test.i` and SWIG reports the parse error in the line after because that's where the next token is.

PR #2188, opened 2022-02-03
Method calls in expression

ojwb:


ojwb:
CI really passed (the failure is a PHP build failing to upload its cache after passing all tests) so merging.

PR #2189, opened 2022-02-03
[C#] Fix memory leak in directorin typemap for std::string.

### LINKED ISSUE #998: Memory leak with director, csharp, std::string and sending strings to c#
## SWIG.i

```c++
%{
#include "net_invoker.h"
%}

%feature("director") NetInvokerBase;

class NetInvokerBase {
public:
    virtual ~NetInvokerBase() {}
    virtual const NetMethodInfo* GetMethodInfo(std::string tt)
    {
        return NULL;
    }
};

class NetInvoker {
public:
  static void set(NetInvokerBase* invoker) { NetInvoker::invoker = invoker; }
  static void reset() { NetInvoker::invoker = 0; }
private:
  static NetInvokerBase *invoker;
};
```

## C#

```c#
public class RuntimeNetInvoker : Qt.NetCore.NetInvokerBase
{
    public override bool IsValidType(string type)
    {
        return true;
    }

    public override NetMethodInfo GetMethodInfo(string tt)
    {
        return null;
    }
}

NetInvoker.set(new RuntimeNetInvoker());
```

## C++

```c++
std::string netType = "test";
NetInvoker::invoker->GetMethodInfo(netType);
```

This problem goes away if I take out the ```std::string``` parameter to ```GetMethodInfo```. So, this told me the issue is somewhere in the generated c++ code that marshals the ```std::string``` to the c# callback.

Here was the original code that was generated.

```c++
NetMethodInfo const *SwigDirector_NetInvokerBase::GetMethodInfo(std::string tt) {
  NetMethodInfo *c_result = 0 ;
  void * jresult = 0 ;
  char * jtt  ;
  
  if (!swig_callbackGetMethodInfo) {
    return NetInvokerBase::GetMethodInfo(tt);
  } else {
    jtt = SWIG_csharp_string_callback((&tt)->c_str());
    jresult = (void *) swig_callbackGetMethodInfo(jtt);
    c_result = (NetMethodInfo *)jresult; 
  }
  return (NetMethodInfo const *)c_result;
}
```

If I take out the ```SWIG_csharp_string_callback``` call, and passed the string directly to ```swig_callbackGetMethodInfo```, everything works, and the memory leak is gone.

So, changing the generated wrapper to this, resolved it.

```c#
NetMethodInfo const *SwigDirector_NetInvokerBase::GetMethodInfo(std::string tt) {
  NetMethodInfo *c_result = 0 ;
  void * jresult = 0 ;
  char * jtt  ;
  
  if (!swig_callbackGetMethodInfo) {
    return NetInvokerBase::GetMethodInfo(tt);
  } else {
    //jtt = SWIG_csharp_string_callback((&tt)->c_str());
    jresult = (void *) swig_callbackGetMethodInfo((&tt)->c_str());
    c_result = (NetMethodInfo *)jresult; 
  }
  return (NetMethodInfo const *)c_result;
}
```
How do I resolve this? I suspect the issue is somewhere in ```%include "std_string.i"```. for c#.

I am using swig 3.0.8, fresh from the Ubuntu repos.

ojwb:
Fixes #998

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -8,6 +8,10 @@ Version 4.1.0 (in progress)
 ===========================
 
 2022-02-03: olly
+	    [C#] #283 #998 Fix memory leak in directorin typemap for
+	    std::string.
+
+2022-02-03: olly
 	    [Python] #967 Make `self` parameter available to user typemaps.
 
 2022-02-03: teythoon
--- a/Lib/csharp/std_string.i
+++ b/Lib/csharp/std_string.i
@@ -20,7 +20,7 @@ namespace std {
 class string;
 
 // string
-%typemap(ctype) string "char *"
+%typemap(ctype) string "const char *"
 %typemap(imtype) string "string"
 %typemap(cstype) string "string"
 
@@ -42,7 +42,7 @@ class string;
    }
    $result.assign($input); %}
 
-%typemap(directorin) string %{ $input = SWIG_csharp_string_callback($1.c_str()); %}
+%typemap(directorin) string %{ $input = $1.c_str(); %}
 
 %typemap(csin) string "$csinput"
 %typemap(csout, excode=SWIGEXCODE) string {


ojwb:
All the CI jobs for csharp had passed so merged.

wsfulton:
These changes were only made for `std::string` and not `const std::string&`. Is there a reason or is it just an oversight?

wsfulton:
I can't find a C# runtime test in this area in our test-suite and I really think we need one if we're changing this because the marshalling of strings in the CLR is complete black magic, undocumented by Microsoft and not consistent across platforms. The leaks reported in the SWIG issues were only for Linux and not Windows, which makes me wonder if the new solution will actually be okay on Windows (which of course is the primary C# platform).

wsfulton:
I'll look at adding in some C# runtime tests for these directors.

ojwb:
> These changes were only made for `std::string` and not `const std::string&`. Is there a reason or is it just an oversight?

Sorry, I just turned the suggested patch in #998 into a PR (as you'd suggested the submitter do years ago but they never followed through).  No deep reason for not addressing the const ref case.

wsfulton:
`const std::string&` done now in 8e4868af75f4f5fd0c60a41cae92763ac8110e79

wsfulton:
BTW, I verified there is one less memory leakage with the patch on linux using valgrind. I couldn't determine this on Windows: drmemory bombed out. The added examples at least run on Windows, so hopefully all okay.

PR #2190, opened 2022-02-03
Add possibility to use only stable Python ABI

vadz:
This rebases the changes of #1613 on master (this is not just a trivial rebase because the code affected by the original changes has been changed/moved around several times in master), includes the additional changes of #1687 and #1727 that I had to do on top of it and, finally, adds a CI build using the new option.

@wsfulton Please consider merging this. This PR doesn't change anything for the existing use cases and just adds an extremely useful new option. And I've just spent a couple more hours on fixing this yet again, because it's too useful to not have it, but I simply can't keep doing it forever, it's too time/energy-consuming and morale-sapping.

### DIFF
--- a/Lib/python/director.swg
+++ b/Lib/python/director.swg
@@ -14,6 +14,24 @@
 #include <vector>
 #include <map>
 
+#if defined(SWIG_PYTHON_THREADS)
+/*  __THREAD__ is the old macro to activate some thread support */
+# if !defined(__THREAD__)
+#   define __THREAD__ 1
+# endif
+#endif
+
+#ifdef __THREAD__
+#ifndef Py_LIMITED_API
+# include "pythread.h"
+#else
+# if defined(_WIN32)
+#   include <windows.h>
+# else
+#   include <pthread.h>
+# endif
+#endif
+#endif
 
 /*
   Use -DSWIG_PYTHON_DIRECTOR_NO_VTABLE if you don't want to generate a 'virtual
@@ -244,25 +262,89 @@ namespace Swig {
   };
 
 
-#if defined(SWIG_PYTHON_THREADS)
-/*  __THREAD__ is the old macro to activate some thread support */
-# if !defined(__THREAD__)
-#   define __THREAD__ 1
-# endif
-#endif
-
 #ifdef __THREAD__
-# include "pythread.h"
+#ifndef Py_LIMITED_API
+   class Mutex
+   {
+   public:
+       Mutex() {
+           mutex_ = PyThread_allocate_lock();
+       }
+
+       ~Mutex() {
+           PyThread_release_lock(mutex_);
+       }
+
+   private:
+       void Lock() {
+           PyThread_acquire_lock(mutex_, WAIT_LOCK);
+       }
+
+       void Unlock() {
+           PyThread_free_lock(mutex_);
+       }
+
+       PyThread_type_lock mutex_;
+
+       friend class Guard;
+   };
+#elif defined(_WIN32)
+    class Mutex : private CRITICAL_SECTION {
+    public:
+        Mutex() {
+            InitializeCriticalSection(this);
+        }
+
+        ~Mutex() {
+            DeleteCriticalSection(this);
+        }
+
+    private:
+        void Lock() {
+            EnterCriticalSection(this);
+        }
+
+        void Unlock() {
+            LeaveCriticalSection(this);
+        }
+
+        friend class Guard;
+    };
+#else
+    class Mutex {
+    public:
+        Mutex() {
+            pthread_mutex_init(&mutex_, NULL);
+        }
+
+        ~Mutex() {
+            pthread_mutex_destroy(&mutex_);
+        }
+
+    private:
+        void Lock() {
+            pthread_mutex_lock(&mutex_);
+        }
+
+        void Unlock() {
+            pthread_mutex_unlock(&mutex_);
+        }
+
+        friend class Guard;
+
+        pthread_mutex_t mutex_;
+    };
+#endif
   class Guard {
-    PyThread_type_lock &mutex_;
+    Mutex &mutex_;
 
   public:
-    Guard(PyThread_type_lock & mutex) : mutex_(mutex) {
-      PyThread_acquire_lock(mutex_, WAIT_LOCK);
+    Guard(Mutex & mutex) : mutex_(mutex) {
+      mutex_.Lock();
     }
 
     ~Guard() {
-      PyThread_release_lock(mutex_);
+      mutex_.Unlock();
     }
   };
 # define SWIG_GUARD(mutex) Guard _guard(mutex)
@@ -330,7 +412,7 @@ namespace Swig {
     typedef std::map<void *, GCItem_var> swig_ownership_map;
     mutable swig_ownership_map swig_owner;
 #ifdef __THREAD__
-    static PyThread_type_lock swig_mutex_own;
+    static Mutex swig_mutex_own;
 #endif
 
   public:
@@ -382,7 +464,7 @@ namespace Swig {
   };
 
 #ifdef __THREAD__
-  PyThread_type_lock Director::swig_mutex_own = PyThread_allocate_lock();
+  Mutex Director::swig_mutex_own;
 #endif
 }
 
--- a/Lib/python/pybuffer.i
+++ b/Lib/python/pybuffer.i
@@ -12,18 +12,43 @@
  *      }
  */
 
+/* Note that in Py_LIMITED_API case we have no choice, but to use deprecated
+ * functions, as they provides the only way to access buffer data with limited
+ * API, which doesn't include Py_buffer definition. We also disable the
+ * warnings about doing this because they're not useful in our case.
+ */
+
 %define %pybuffer_mutable_binary(TYPEMAP, SIZE)
 %typemap(in) (TYPEMAP, SIZE) {
   int res; Py_ssize_t size = 0; void *buf = 0;
+%#ifndef Py_LIMITED_API
   Py_buffer view;
   res = PyObject_GetBuffer($input, &view, PyBUF_WRITABLE);
+%#else
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic push
+    %#pragma GCC diagnostic ignored "-Wdeprecated"
+    %#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+  %#elif defined(_MSC_VER)
+    %#pragma warning(push)
+    %#pragma warning(disable: 4996)
+  %#endif
+  res = PyObject_AsWriteBuffer($input, &buf, &size);
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic pop
+  %#elif defined(_MSC_VER)
+    %#pragma warning(pop)
+  %#endif
+%#endif
   if (res < 0) {
     PyErr_Clear();
     %argument_fail(res, "(TYPEMAP, SIZE)", $symname, $argnum);
   }
+%#ifndef Py_LIMITED_API
   size = view.len;
   buf = view.buf;
   PyBuffer_Release(&view);
+%#endif
   $1 = ($1_ltype) buf;
   $2 = ($2_ltype) (size/sizeof($*1_type));
 }
@@ -45,14 +70,34 @@
 %define %pybuffer_mutable_string(TYPEMAP)
 %typemap(in) (TYPEMAP) {
   int res; void *buf = 0;
+%#ifndef Py_LIMITED_API
   Py_buffer view;
   res = PyObject_GetBuffer($input, &view, PyBUF_WRITABLE);
+%#else
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic push
+    %#pragma GCC diagnostic ignored "-Wdeprecated"
+    %#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+  %#elif defined(_MSC_VER)
+    %#pragma warning(push)
+    %#pragma warning(disable: 4996)
+  %#endif
+  Py_ssize_t size;
+  res = PyObject_AsWriteBuffer($input, &buf, &size);
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic pop
+  %#elif defined(_MSC_VER)
+    %#pragma warning(pop)
+  %#endif
+%#endif
   if (res < 0) {
     PyErr_Clear();
     %argument_fail(res, "(TYPEMAP)", $symname, $argnum);
   }
+%#ifndef Py_LIMITED_API
   buf = view.buf;
   PyBuffer_Release(&view);
+%#endif
   $1 = ($1_ltype) buf;
 }
 %enddef
@@ -74,15 +119,34 @@
 %define %pybuffer_binary(TYPEMAP, SIZE)
 %typemap(in) (TYPEMAP, SIZE) {
   int res; Py_ssize_t size = 0; const void *buf = 0;
+%#ifndef Py_LIMITED_API
   Py_buffer view;
   res = PyObject_GetBuffer($input, &view, PyBUF_CONTIG_RO);
+%#else
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic push
+    %#pragma GCC diagnostic ignored "-Wdeprecated"
+    %#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+  %#elif defined(_MSC_VER)
+    %#pragma warning(push)
+    %#pragma warning(disable: 4996)
+  %#endif
+  res = PyObject_AsReadBuffer($input, &buf, &size);
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic pop
+  %#elif defined(_MSC_VER)
+    %#pragma warning(pop)
+  %#endif
+%#endif
   if (res < 0) {
     PyErr_Clear();
     %argument_fail(res, "(TYPEMAP, SIZE)", $symname, $argnum);
   }
+%#ifndef Py_LIMITED_API
   size = view.len;
   buf = view.buf;
   PyBuffer_Release(&view);
+%#endif
   $1 = ($1_ltype) buf;
   $2 = ($2_ltype) (size / sizeof($*1_type));
 }
@@ -106,14 +170,34 @@
 %define %pybuffer_string(TYPEMAP)
 %typemap(in) (TYPEMAP) {
   int res; const void *buf = 0;
+%#ifndef Py_LIMITED_API
   Py_buffer view;
   res = PyObject_GetBuffer($input, &view, PyBUF_CONTIG_RO);
+%#else
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic push
+    %#pragma GCC diagnostic ignored "-Wdeprecated"
+    %#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+  %#elif defined(_MSC_VER)
+    %#pragma warning(push)
+    %#pragma warning(disable: 4996)
+  %#endif
+  Py_ssize_t size;
+  res = PyObject_AsReadBuffer($input, &buf, &size);
+  %#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
+    %#pragma GCC diagnostic pop
+  %#elif defined(_MSC_VER)
+    %#pragma warning(pop)
+  %#endif
+%#endif
   if (res < 0) {
     PyErr_Clear();
     %argument_fail(res, "(TYPEMAP)", $symname, $argnum);
   }
+%#ifndef Py_LIMITED_API
   buf = view.buf;
   PyBuffer_Release(&view);
+%#endif
   $1 = ($1_ltype) buf;
 }
 %enddef
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -87,3 +87,17 @@ SWIG_Python_str_FromChar(const char *c)
 #define PyDescr_NAME(x) (((PyDescrObject *)(x))->d_name)
 #define Py_hash_t long
 #endif
+
+#ifdef Py_LIMITED_API
+# define PyTuple_GET_ITEM PyTuple_GetItem
+/* Note that PyTuple_SetItem() has different semantics from PyTuple_SET_ITEM as it decref's the original tuple item, so in general they cannot be used
+  interchangeably. However in SWIG-generated code PyTuple_SET_ITEM is only used with newly initialized tuples without any items and for them this does work. */
+# define PyTuple_SET_ITEM PyTuple_SetItem
+# define PyTuple_GET_SIZE PyTuple_Size
+# define PyCFunction_GET_FLAGS PyCFunction_GetFlags
+# define PyCFunction_GET_FUNCTION PyCFunction_GetFunction
+# define PyCFunction_GET_SELF PyCFunction_GetSelf
+# define PyList_GET_ITEM PyList_GetItem
+# define PyList_SET_ITEM PyList_SetItem
+# define PySliceObject PyObject
+#endif
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -36,7 +36,7 @@
 SWIGINTERN char*
 SWIG_Python_str_AsChar(PyObject *str)
 {
-#if PY_VERSION_HEX >= 0x03030000
+#if PY_VERSION_HEX >= 0x03030000 && !defined(Py_LIMITED_API)
   return (char *)PyUnicode_AsUTF8(str);
 #elif PY_VERSION_HEX >= 0x03000000
   char *newstr = 0;
@@ -57,7 +57,7 @@ SWIG_Python_str_AsChar(PyObject *str)
 #endif
 }
 
-#if PY_VERSION_HEX >= 0x03030000 || PY_VERSION_HEX < 0x03000000
+#if (PY_VERSION_HEX >= 0x03030000 && !defined(Py_LIMITED_API)) || PY_VERSION_HEX < 0x03000000
 #  define SWIG_Python_str_DelForPy3(x)
 #else
 #  define SWIG_Python_str_DelForPy3(x) free( (void*) (x) )
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -70,6 +70,7 @@ static int max_bases = 0;
 static int builtin_bases_needed = 0;
 
 static int py3 = 0;
+static int py3_stable_abi = 0;
 
 /* C++ Support + Shadow Classes */
 
@@ -130,6 +131,7 @@ static const char *usage3 = "\
      -nothreads      - Disable thread support for the entire interface\n\
      -olddefs        - Keep the old method definitions when using -fastproxy\n\
      -py3            - Generate code with Python 3 specific features and syntax\n\
+     -py3-stable-abi - Generate code compatible with Python 3 stable ABI (PEP 384)\n\
      -relativeimport - Use relative Python imports\n\
      -threads        - Add thread support for all the interface\n\
      -O              - Enable the following optimization options:\n\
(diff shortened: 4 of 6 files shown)


vadz:
Damn, I forgot to check the examples again (I should have remembered by now that checking the unit tests is not enough for SWIG...), sorry. Will fix a.s.a.p.

Lib/python/pyhead.swg
```diff
@@ -87,3 +87,17 @@ SWIG_Python_str_FromChar(const char *c)
 #define PyDescr_NAME(x) (((PyDescrObject *)(x))->d_name)
 #define Py_hash_t long
 #endif
+
+#ifdef Py_LIMITED_API
+# define PyTuple_GET_ITEM PyTuple_GetItem
+/* Note that PyTuple_SetItem() has different semantics from PyTuple_SET_ITEM as it decref's the original tuple item, so in general they cannot be used
+  interchangeably. However in SWIG-generated code PyTuple_SET_ITEM is only used with newly initialized tuples without any items and for them this does work. */
+# define PyTuple_SET_ITEM PyTuple_SetItem
```
ojwb, line 95: This seems dangerous since existing user typemap code might use `PyTuple_SET_ITEM` in other ways and so break in likely obscure ways if they try it with the limited API.  There's also a risk of someone using it in a future SWIG version in other ways.

Maybe it should redirect through `SWIG_PyTyple_ADD_ITEM()` or something like that?  (`ADD` to make it clear it works when adding a new item rather than a general set that could replace.)

Source/Modules/python.cxx
```diff
@@ -922,15 +943,15 @@ class PYTHON:public Language {
...
+      String *name = NewString("SWIG_PyInstanceMethod_New");
+      String *line = NewString("");
+      Printf(line, "\t { \"%s\", %s, METH_O, NULL},\n", name, name);
+      Append(methods, line);
       Append(methods_proxydocs, line);
+      Delete(line);
+      Delete(name);
     }
-    Delete(line);
-    Delete(name);
```
ojwb, line 933: What's going on here?

You said "This PR doesn't change anything for the existing use cases" but this seems to change ... something ... to be only done under `fastproxy` outside of the limited ABI case.

Lib/python/pybuffer.i
```diff
@@ -12,18 +12,43 @@
...
+/* Note that in Py_LIMITED_API case we have no choice, but to use deprecated
+ * functions, as they provides the only way to access buffer data with limited
+ * API, which doesn't include Py_buffer definition. We also disable the
+ * warnings about doing this because they're not useful in our case.
+ */
+
 %define %pybuffer_mutable_binary(TYPEMAP, SIZE)
 %typemap(in) (TYPEMAP, SIZE) {
   int res; Py_ssize_t size = 0; void *buf = 0;
+%#ifdef Py_LIMITED_API
```
ojwb, line 24: All the other files seem to have the `!Py_LIMITED_API` case first in conditionals, but in this file it's the other way around...

Lib/python/pybuffer.i
```diff
@@ -12,18 +12,43 @@
  *      }
  */
 
+/* Note that in Py_LIMITED_API case we have no choice, but to use deprecated
+ * functions, as they provides the only way to access buffer data with limited
+ * API, which doesn't include Py_buffer definition. We also disable the
+ * warnings about doing this because they're not useful in our case.
+ */
```
ojwb, line 19: So what's the plan here longer term?

Has someone flagged this situation to the Python developers?  Ideally the non-deprecated buffer mechanism gets added to the stable ABI before the deprecated one gets removed from Python.

Lib/python/director.swg
```diff
@@ -244,25 +258,64 @@ namespace Swig {
...
+
+        void Unlock() {
+            pthread_mutex_unlock(&mutex_);
+        }
+
+        friend class Guard;
+
+        pthread_mutex_t mutex_;
+    };
+# endif
```
ojwb, line 308: > This PR doesn't change anything for the existing use cases

Here we now use pthreads/microsoft NIH rather than Python's thread functions for locking even in the non-limited ABI case.

vadz: Sorry, I forgot about this... Of course, we _had_ used these functions already until the changes of 4aaf48d65055bd68cc55e81f915af39693e9a7a6, so it's not like we had ever had any problems with them, but I can add another preprocessor check to use Python functions when `Py_LIMITED_API` is off.

Lib/python/pybuffer.i
vadz: > So what's the plan here longer term?

Sorry, no idea. I need this code to work today and this is the only way to make it work, to the best of my knowledge.

> Has someone flagged this situation to the Python developers? Ideally the non-deprecated buffer mechanism gets added to the stable ABI before the deprecated one gets removed from Python.

Ideally, yes. For now the deprecated functions still exist even in 3.10, in spite of being deprecated since 3.0.

henryiii: I'd love a way to access buffer data via the stable ABI. Probably very much worth bringing up.

Lib/python/pybuffer.i
vadz: I've reversed the check direction.

Lib/python/pyhead.swg
vadz: This would result in more changes and I'm not sure if it's worth making them. Also, as this only affects the use case which doesn't exist today, I think we could definitely merge this first and replace the calls to `PyTuple_SET_ITEM()` with whatever later.

Source/Modules/python.cxx
vadz: Looks like a mistake in the [original changes](https://github.com/swig/swig/pull/1613/files#diff-c686001fe7d4424a7542f21deceb3f4c8e5dc806ec3edbf9cb36444747456a8b), thanks for noticing it!

Fixed now.

vadz: No, sorry, it's my mistake. `SWIG_PyInstanceMethod_New()` is only used with "fastproxy" and we must not emit it when not using it as it wouldn't even compile with limited API.

I.e. the previously emitted declarations were actually useless before and this fixes it as a (welcome, even if minor) side effect.

So, finally, there is nothing to fix here.

Lib/python/pybuffer.i
vadz: To be clear, this does work today, when using stable ABI. It just uses deprecated (but still perfectly working) functions for doing it.

Lib/python/director.swg
vadz: Done now.

Lib/python/pybuffer.i
ojwb: I am clear on that, but also by merging a new feature we're creating an expectation that we'll support it for some time, so merging a feature which relies on something in the target language API which has been long marked as deprecated is problematic.  It'd be reassuring to know that the Python developers were at least aware of the situation.

vadz: Turns out I actually have great timing because after checking for it again, the new buffer protocol was [just stabilized](https://bugs.python.org/issue45459) _yesterday_. So we'll be able to start using it from 3.11.

Of course, realistically speaking, it will be quite some time before anybody wants to use stable ABI for 3.11 and later only, so we're going to keep the current code for many years ahead. It's ugly but at least we know that we'll have something to switch over to when it's finally removed (if it ever is, some of the discussions seem to indicate that stable ABI functions will never be removed at all).

I think this should answer your concerns.

Source/Modules/python.cxx
ojwb: I've opened #2191 to address this sub-issue more completely - the condition these are actually used under is `!builtin && fastproxy` so I've used that here, for the static equivalent just below, and in `pyinit.swg`.  I did some limited testing locally and it looked good, but CI is much more thorough.  Assuming that can be merged it'll simplify this patch somewhat.  I have a merged version I can push if you want (I did a test merge to see if this actually allowed `-py3-stable-abi -builtin` to work, but it doesn't).

ojwb: #2191 now merged.

Source/Modules/python.cxx
```diff
@@ -396,6 +398,11 @@ class PYTHON:public Language {
 	  py3 = 1;
 	  Preprocessor_define("SWIGPYTHON_PY3", 0);
 	  Swig_mark_arg(i);
+	} else if (strcmp(argv[i], "-py3-stable-abi") == 0) {
+	  py3_stable_abi = 1;
+	  py3 = 1;
```
ojwb, line 403: Should this actually imply `-py3`?

You normally get Python3-compatible wrappers without `-py3` and my understanding was `-py3` mostly turned on things of debatable usefulness (such as annotations which aren't very useful as they use the C/C++ type).  See #1779 where William discusses this a bit.

Unless `-py3` is actually required it seems the user ought to get to choose; if it is required, it'd be good to note something like:

      -py3-stable-abi - Generate code compatible with Python 3 stable ABI (PEP 384) (implies -py3)

Source/Modules/python.cxx
vadz: I'm going to have to rebase and force push now, as this can't be merged any more after this merge...

Source/Modules/python.cxx
vadz: The code using stable API won't compile with Python 2 anyhow, so it doesn't seem to make much sense to use it without `-py3`.

vadz:
Force pushed the rebased version now to resolve conflicts due to #2191 merge.

Source/Modules/python.cxx
ojwb: My point is that `-py3` (at least currently) doesn't just say "generate wrappers which only work with Python3" but enables things like those annotations which aren't wanted by most people.  There's then no way to turn them off again.

`-py3` really needs sorting out, but chaining more options to it if anything makes that a bit harder to do.

Lib/python/pybuffer.i
ojwb: Mostly.

I think we want to support the new stable buffer API **before** we're forced to by the old one being removed.  Unless SWIG starts to get new releases an order of magnitude or so more frequently users will likely end up stuck with the latest SWIG release not working with the latest Python release in this configuration if we're only reactive here.

It doesn't need to block the merge though.

ojwb: It's not entirely helpful to do reply+"resolve conversation" by the way - means I have to keep clicking to see your replies.

AlsoIME the usual protocol is that it's the reviewer who decides whether the matter is resolved or not...

Lib/python/director.swg
ojwb: Thanks.

> Of course, we had used these functions already until the changes of https://github.com/swig/swig/commit/4aaf48d65055bd68cc55e81f915af39693e9a7a6, so it's not like we had ever had any problems with them

Before 4aaf48d65055bd68cc55e81f915af39693e9a7a6 the code generated with `-threads` didn't even compile on Microsoft Windows.  I suppose that's not a problem with the functions as such, but it clearly shows nobody had actually tested with them whereas the code added by that commit has been in use for 14 years.

Source/Modules/python.cxx
vadz: Sure, we could remove it from here, it doesn't matter to me personally as I always use both options together anyhow. It just looks like `-py3-stable-abi` should be a superset of `-py3` but, again, I don't care enough about this to argue.

Lib/python/pybuffer.i
vadz: Sorry, I've resolved this one because I thought that the linked issue answers the question about the future and there is really nothing more to do about this today, as we can't rely on everyone having Python 3.11 just yet.

Once 3.11 is released we could at least add another condition and use the new functions in  `Py_LIMITED_API >= 3.11` case, but it would be nice to add CI tests using 3.11 before doing this so, again, I don't think there is anything that can be done today.

In the worst case, if it's really problematic, let's drop this commit. If all the rest is merged and I just have to maintain this one in my branch, I can live with it.

Lib/python/pyhead.swg
ojwb: Actually, I had a look at the Python C API docs for this, which say (my emphasis):

> Like `PyTuple_SetItem()`, but does no error checking, and **should only be used to fill in brand new tuples**.

So using it on something that isn't a brand new tuple is documented as something not to do, and in the light of that just mapping it like above seems OK - any user typemap broken by this would be dubious code.

wsfulton:
Noted, please bear with me until I get to this.

ojwb:
@vadz See #2209 - the Python part of it is actually with this branch.

vadz:
> @vadz See #2209 - the Python part of it is actually with this branch.

I did see it, but IMNSHO it has nothing to do with this branch because my code does use `::InitializeCriticalSection()` correctly. They probably use some other branch or their own workaround for the problem fixed here -- and such a proliferation of various (sometimes wrong) branches is exactly what you would expect as long as the official version doesn't include stable ABI support.

rglockan:
Hello,
Is this still planned to be part of the 4.1 release at the end of month? The new issue is no longer tagged with the 4.1 milestone.

ojwb:
> Is this still planned to be part of the 4.1 release at the end of month? The new issue is no longer tagged with the 4.1 milestone.

Note that the date on the 4,1 milestone is entirely nominal - when I set the date to the end of this month I just picked a date that was a vaguely plausible distance into the future rather than based on an realistic assessment of what was still to do.  At that point we were well passed the previously set date so it was hard to pick a worse date.

I see the milestone date is actually optional, so I've just removed it.  SWIG developers are volunteers and SWIG releases are more in the "when it's ready" school rather than being date based, and having a date there can clearly give users false expectations.

Whether this patch makes 4.1.0 is really entirely up to William - he's the one who indicated he wanted to review it before it gets merged, and he's also the one who actually makes the releases.

rglockan:
Thanks for the detailed answer. I already feared that something along these lines it would be, but atleast i now have confirmation.

RytoEX:
We use SWIG in OBS Studio to allow users to run Python scripts to control parts of the application. This PR is of great interest to us to provide more flexibility in the versions of Python we can support. We rebased this on master and made a few changes to get this working on Windows and macOS.

Here is a patch of the changes to the PR that were required for us: https://gist.github.com/RytoEX/c9a7a7f65e469138401096ca86ddf89a
* Avoid empty initializer list for PyType_Spec initializations
* Replace SWIG_exit with Exit per https://github.com/swig/swig/commit/55377bdc08ffbe37fbb89566a8754302ae186227

Here is working patch that applies to current master: https://gist.github.com/PatTheMav/0abbd031d7f6f1c7af7a22090437b338

Lib/python/pyhead.swg
```diff
@@ -36,7 +36,7 @@
 SWIGINTERN char*
 SWIG_Python_str_AsChar(PyObject *str)
 {
-#if PY_VERSION_HEX >= 0x03030000
+#if PY_VERSION_HEX >= 0x03030000 && !defined(Py_LIMITED_API)
```
clintonstimpson, line 39: Instead of this #ifdef, and reverting back to a Python 2 API, can you use both `PyUnicode_AsUTF8String()` and `PyBytes_AsString()`?  The ifdef would fall out, and the limited api would be supported.  That seems simple enough to commit/review/merge separately from a Py_LIMITED_API switch.

clintonstimpson:
> We use SWIG in OBS Studio to allow users to run Python scripts to control parts of the application. This PR is of great interest to us to provide more flexibility in the versions of Python we can support. We rebased this on master and made a few changes to get this working on Windows and macOS.
> 
> Here is a patch of the changes to the PR that were required for us: https://gist.github.com/RytoEX/c9a7a7f65e469138401096ca86ddf89a
> 
>     * Avoid empty initializer list for PyType_Spec initializations
> 
>     * Replace SWIG_exit with Exit per [55377bd](https://github.com/swig/swig/commit/55377bdc08ffbe37fbb89566a8754302ae186227)
> 
> 
> Here is working patch that applies to current master: https://gist.github.com/PatTheMav/0abbd031d7f6f1c7af7a22090437b338

Thanks for posting the patch!!  I tried it out and yes, it compiles, and works with simple cases.  However, it breaks things when multiprocessing is used.

encukou:
Is there anything I can do on the CPython stable ABI side to help move this forward?

RytoEX:
> Thanks for posting the patch!! I tried it out and yes, it compiles, and works with simple cases. However, it breaks things when multiprocessing is used.

I don't know what this means. We're currently using a patched SWIG for OBS Studio 28.

okainov:
@wsfulton  @vadz any updates on this one?

ojwb:
FWIW, there has been some progress towards this recently - @clintonstimpson has got 2a0abbb7f6ddab87463d7c9bc97917238e25784b and e683168018558f4e73dc26629c4bbec224d23438 merged.

Lib/python/pyhead.swg
ojwb: Addressed by #2392

wsfulton:
I would like to see stable Python ABI in swig-4.2. If someone would like to fix the conflicts in this pull request that would be appreciated. Alternatively, continue breaking it into multiple smaller pull requests, such as has been happening recently also works.

wsfulton:
I'm getting my head round this stable abi now, so sorry for all the delays.

I've raised #2727 based off this. Conflicts are resolved and a few improvements have been added. 

This is still work in progress. A fix for the string_view tests is needed. I was thinking of replacing the `PyUnicode_AsUTF8AndSize` (and the rest of the typemaps) with SWIG_AsCharPtrAndSize. @ojwb, was there perhaps a reason you didn't use this when you commited std_string_view.i in 29d1f95d0d282fdefd0d476b38946c24fcdd8a0b?

ojwb:
> I was thinking of replacing the `PyUnicode_AsUTF8AndSize` (and the rest of the typemaps) with SWIG_AsCharPtrAndSize. @ojwb, was there perhaps a reason you didn't use this when you commited std_string_view.i in [29d1f95](https://github.com/swig/swig/commit/29d1f95d0d282fdefd0d476b38946c24fcdd8a0b)?

It's lower overhead.  Using `SWIG_AsCharPtrAndSize` ends up creating a new bytes object:

```
    obj = PyUnicode_AsUTF8String(obj);             
    if (!obj)
      return SWIG_TypeError;
    if (alloc)
      *alloc = SWIG_NEWOBJ;
```

Using `PyUnicode_AsUTF8AndSize` we get a pointer to a cached UTF8 representation maintained inside the unicode object:

https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize

Note that `PyUnicode_AsUTF8AndSize` **is** in the stable ABI since Python 3.10.

vadz:
> A fix for the string_view tests is needed.

I could provide one (but not before tomorrow), but I don't see anything better to do than using `PyUnicode_AsUTF8String()` which is indeed less efficient -- but I think it's too early to drop support for Python < 3.10 yet (we could use the current typemap for it, however, of course).

Please let me know if it's worth doing this or if you'd rather do it yourself.

Lib/python/pybuffer.i
ojwb: > Once 3.11 is released we could at least add another condition and use the new functions in `Py_LIMITED_API >= 3.11` case, but it would be nice to add CI tests using 3.11 before doing this so, again, I don't think there is anything that can be done today.

Python 3.11 has now been released and we now have CI for Python 3.11 (and 3.12) so this could now be done.

I don't think it needs to block the merge, but it'd be great to sort this out soon.

ojwb:
> I am not familiar with this detail, but at least the converted UTF-8 version stored in the python string can be accessed via the buffer inside the object - https://docs.python.org/3/c-api/unicode.html?highlight=pyunicode_asutf8andsize#c.PyUnicode_AsUTF8AndSize. This may have been overlooked when @bhy originally wrote it.

`PyUnicode_AsUTF8AndSize()` was added in Python 3.3 so didn't exist back then (Python 3.0 was current at that point).  I don't think there used to be a cached UTF8 version of the string at all.  (And SWIG 4.1.0 dropped support for Python3 < 3.3 so we can just use this for Python 3 in the non-stable ABI case.)

wsfulton:
Switching to using the approach in `SWIG_Python_str_AsChar` didn't go well.
https://github.com/swig/swig/blob/02c45b9145e5ee3d44d0fa097bb9ee2eae2a3345/Lib/python/pyhead.swg#L33-L48
This code is in master but is actually unsafe in the 2nd else block for `Py_LIMITED_API < 0x030A0000` as `chars` returns a pointer to a bytes object that does not exist as its reference count is zero at the end of the function. I switched more code to use this function and tests uncovered undefined behaviour after calling unrelated Python APIs.

@clintonstimpson this code was committed in 1f40d38e56c79d5ba7b2442033297c8095e0cbe8 as part of #2392. I'm not entirely sure how to support the older Py_LIMITED_API without changing the signature of `SWIG_Python_str_AsChar`, or otherwise copying the returned string and resurrecting `SWIG_Python_str_DelForPy3` deleted by @ojwb in 0b9d4eff09.

clintonstimpson:
> @clintonstimpson this code was committed in [1f40d38](https://github.com/swig/swig/commit/1f40d38e56c79d5ba7b2442033297c8095e0cbe8) as part of #2392. I'm not entirely sure how to support the older Py_LIMITED_API without changing the signature of `SWIG_Python_str_AsChar`, or otherwise copying the returned string and resurrecting `SWIG_Python_str_DelForPy3` deleted by @ojwb in [0b9d4ef](https://github.com/swig/swig/commit/0b9d4eff09a9d12435e9688842662994ec2b154d).

The older Py_LIMITED_API is already supported in the #else case by using PyUnicode_AsUTF8String() and PyBytes_AsString.  Only those opting for Py_LIMITED_API and python less than 3.10 will pay for the extra allocation.

clintonstimpson:
> The older Py_LIMITED_API is already supported in the #else case by using PyUnicode_AsUTF8String() and PyBytes_AsString. Only those opting for Py_LIMITED_API and python less than 3.10 will pay for the extra allocation. Are you trying to optimize this further?

I'll also add that it uses the PyBytes_AsString internal buffer, so I imagine any attempt to manage your own allocation and clean up at a later point would give nearly identical performance.  Is there a potential issue with multiple calls to SWIG_Python_str_AsChar() and wanting separate char* instances... I'm not sure if that is a potential issue here.

ojwb:
I don't have a better idea than resurrecting `SWIG_Python_str_DelForPy3`.  Changing the function signature seems worse as it'll require reworking of user typemaps which use `SWIG_Python_str_AsChar` even for users who don't care about Python's limited ABI (or only care about it for Python >= 3.10).  Also it's likely in many cases there are still `SWIG_Python_str_DelForPy3` calls in place from when it was needed.

At least the horror^Woverhead is restricted to people wanting to use the limited ABI while supporting Python < 3.10 and will die again eventually.

wsfulton:
I have implemented a fix for the undefined behaviour in f89dd59d4b82ece899087682fdb86e94d2611513. While it does have an api change, this is the time to do it before a major release. The benefits are considerable compared to the inconvenience of a simple change (much more efficient implementation. I don't think we can rely on calls to SWIG_Python_str_DelForPy3 still being implemented if users did a cleanup for swig-4.1. The requirement to call it is also different (a different range of versions required it). Better fail during compilation than leak and be slow. It's an undocumented function too.

If there is a lot of noise around this, `SWIG_Python_str_AsChar` can always be be put back either by users or into the SWIG code base with perhaps some deprecated warnings. 

`SWIG_Python_str_FromChar` can also be cleaned up, so am going to do that next and then look at switching `SWIG_AsCharPtrAndSize` to use `SWIG_PyUnicode_AsUTF8AndSize` for a more efficient `std::string` implementation.

wsfulton:
I have fixed the `std::string_view` 'in' typemaps for the stable abi in 84f6031aa50443dc0ea9b0632acfd1c11813b90b. 

I can't see how to fix the 'directorout' typemaps, but I think these typemaps are also fundamentally flawed even for non stable abi usage. In particular:
https://github.com/swig/swig/blob/ee4f03117760f8d31c25e74278cff0ddc8e2ef65/Lib/python/std_string_view.i#L86-L88
This is only correct while the PyObject has not been deleted. If you look at the generated code for `result` it is:
```
  swig::SwigVar_PyObject result = PyObject_CallMethodObjArgs(swig_get_self(), (PyObject *) swig_method_name, NULL);
```
The destructor for `swig::SwigVar_PyObject` will result in the `result` being deleted and the string_view being returned with a view into this deleted object.

I can't see how this can be fixed other than copying the string and leaking memory. I would conclude that `std::string_view` cannot be used with directors and the test and typemaps should be deleted. @ojwb, you added these typemaps in, do you see another option?

ojwb:
> do you see another option?

I'll have a look.

ojwb:
> I would conclude that `std::string_view` cannot be used with directors and the test and typemaps should be deleted.

The `directorin` typemaps seem entirely fine to me and I see no argument for removing those.

The `directorout` typemaps have limitations, but I think for most target languages they can be used with care (we really need to clearly document this limitation - I'd overlooked this problem originally so didn't).  Care is also needed in the pure C++ case when implementing a virtual method returning `std::string_view`, though for some target languages there's the additional wrinkle that the garbage collector may move variables in memory - e.g. I believe C# does unless you use `fixed` or `stackalloc`.

E.g. here's an implementation of `cpp17_director_string_view_runme.py` - I store the value I'm going to return in an attribute of the object, so it should remain valid while the object does (or until another call to `B.get_first()`:

```
from cpp17_director_string_view import *

import sys
if sys.version_info[0:1] < (3, 0):
    # string_view.i only supported for Python3
    sys.exit(0)

class B(A):

    def __init__(self, string):
        A.__init__(self, string)

    def get_first(self):
        self._get_first_result = super.get_first() + " world!"
        return self._get_first_result

    def process_text(self, string):
        A.process_text(self, string)
        self.smem = "hello"


b = B("hello")

b.get(0)
if b.get_first() != "hello world!":
    raise RuntimeError("b.get_first(): {}".format(b.get_first()))

if b.call_get_first() != "hello world!":
    raise RuntimeError("b.call_get_first(): {}".format(b.call_get_first()))

b.call_process_func()

if b.smem != "hello":
    raise RuntimeError("smem: {}".format(smem))
```

This is clean with `make cpp17_director_string_view.cpptest RUNTOOL=valgrind` but it's also clean under valgrind if I just `return super.get_first() + " world!"`.  The PHP runme is too.  I'm not arguing there isn't a problem here, but do you have a testcase which actually demonstrates the problem?

ojwb:
Superseded by #2727

wsfulton:
The testcase as committed is undefined behaviour full stop even if valgrind cannot find an issue with it. It may happen to work, but that is within the realms of undefined behaviour. Caching the object in the Python objects as shown seems to be a valid fix and the testcase ought to describe this for future reference. Or, as you suggest document this danger.

I suppose the directorout typemap for `std::string_view` is identically as dangerous as pointer or reference typemap. However, in this case, we issue a warning:
```
Warning 473: Returning a pointer or reference in a director method is not recommended.
```
Adding this warning is a reasonable alternative to removing the the typemaps completely.

wsfulton:
> Caching the object in the Python objects as shown seems to be a valid fix and the testcase ought to describe this for future reference.

I verified that this is a valid fix as the reference count in `result` is two after the call to `SwigDirector_A::get_first()` with this code:
```c++
  swig::SwigVar_PyObject result = PyObject_CallMethodObjArgs(swig_get_self(), (PyObject *) swig_method_name, NULL);
```
and drops to one at the end of the method. Without the fix the count is 1 and drops to zero at the end of the call resulting in undefined behaviour.

ojwb:
To be clear I wasn't suggesting that there wasn't a problem here, but rather hoping for some way to test whether there is one so we can be confident that a caching approach actually works.  Checking the ref count doesn't go to zero serves that purpose nicely.

Warning like with the returned pointer/reference case seems appropriate since returning `std::string_view` is really returning a pointer (just wrapped up in a class, and accompanied by a length).

I started to poke the testcases for this yesterday and noticed the Perl directorout testcase is commented out currently because the return value gets corrupted, but this is still the case if I return a cached value.  I haven't dug into what's going on yet.  It could be unrelated as the Perl std::string_view typemaps attempt to reuse existing std::string macro machinery where possible and I may have reused a bit too much.

I can work on a patch but not sure how soon I'll get it done I'm afraid as I'm unlikely to have much hacking time between now and the end of the year.  Depends a bit when I clear the final paid work tasks with deadlines this week.

wsfulton:
I've changed to leak instead of forcing undefined behaviour for the stable abi in commit 69b2234c41fea6f9fe8600e08fb28318f290face for now. A similar analysis and workarounds may be needed for the other languages.

ojwb:
@wsfulton For C#, we currently have:

```
%typemap(directorout, canthrow=1, warning=SWIGWARN_TYPEMAP_THREAD_UNSAFE_MSG) const string_view &
%{ if (!$input) {
    SWIG_CSharpSetPendingExceptionArgument(SWIG_CSharpArgumentNullException, "null string", 0);
    return $null;
   }
   /* possible thread/reentrant code problem */
   static std::string $1_str;
   $1_str = $input;
   static $*1_ltype $1_strview;
   $1_strview = $1_str;
   $result = &$1_strview; %}
```

but:

```
%typemap(directorout, canthrow=1) string_view
%{ if (!$input) {
    SWIG_CSharpSetPendingExceptionArgument(SWIG_CSharpArgumentNullException, "null string", 0);
    return $null;
   }
   $result = std::string_view($input); %}
```

The former I adapted from the `const std::string&` typemap - it's not a nice approach, but maybe we need to use this approach for both cases here?

That's what we already do for Java but I don't really understand C# at this low level well enough to know.  What's the internal string representation?  Is it UTF-8 for mono but wide characters for microsoft?

ojwb:
Also, I have some string_view testcase improvements for other languages - should I push them to the branch or master (I suspect you'll get minor conflicts) or hold them until you've merged the branch?  Trouble is I'm going away for the holidays...

ojwb:
> I suspect you'll get minor conflicts

Just occurred to me I can just try a merge to see and the only conflict is a pre-existing one on `CHANGES.current` so I'll just push these to master.

wsfulton:
How about I merge this branch to master (probably in 12 hours time, I expect my last commit to go make this pull request go green). I'll then merge to master in the morning and you can pick up the string_view changes there.

wsfulton:
I really want to make a release candidate asap, so am unlikely to do any more dev work. 

> That's what we already do for Java but I don't really understand C# at this low level well enough to know. What's the internal string representation? Is it UTF-8 for mono but wide characters for microsoft?

This confuses me no end, can't say without researching, but it is really hard to get right.

ojwb:
I've addressed what I easily can and pushed to make sure CI is still happy.  I think we can do a bit more before 4.2.0 - see #1567 for details.

PR #2191, opened 2022-02-03
[Python] Improve handling of SWIG_Py*Method_New

ojwb:
The SWIG_PyInstanceMethod_New method is no longer added to wrapped
classes except when it's actually needed, which is when
(!builtin && fastproxy) is true (which it isn't by default).

The SWIG_PyStaticMethod_New method is no longer is now similarly
gated - previously only (fastproxy) was checked.

Finally the C/C++ functions that implement these were always compiled
into the module, but now they're only included if
(!builtin && fastproxy) is true.

Issue noted by vadz in #2190.

### DIFF
--- a/Lib/python/pyinit.swg
+++ b/Lib/python/pyinit.swg
@@ -8,6 +8,8 @@
 %fragment("<stddef.h>"); // For offsetof
 #endif
 
+#if defined SWIGPYTHON_FASTPROXY && !defined SWIGPYTHON_BUILTIN
+
 %insert(runtime) %{
 #ifdef __cplusplus
 extern "C" {
@@ -23,6 +25,8 @@ SWIGINTERN PyObject *SWIG_PyStaticMethod_New(PyObject *SWIGUNUSEDPARM(self), PyO
 }
 #endif
 %}
+
+#endif
 
 %init %{
 
@@ -103,6 +107,8 @@ SWIG_Python_FixMethods(PyMethodDef *methods, const swig_const_info *const_table,
   }
 } 
 
+#if defined SWIGPYTHON_FASTPROXY && !defined SWIGPYTHON_BUILTIN
+
 /* -----------------------------------------------------------------------------
  * Method creation and docstring support functions
  * ----------------------------------------------------------------------------- */
@@ -160,6 +166,8 @@ SWIGINTERN PyObject *SWIG_PyStaticMethod_New(PyObject *SWIGUNUSEDPARM(self), PyO
   }
   return PyStaticMethod_New(func);
 }
+
+#endif
 
 #ifdef __cplusplus
 }
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -452,6 +452,10 @@ public:
       SWIG_exit(EXIT_FAILURE);
     }
 
+    if (fastproxy) {
+      Preprocessor_define("SWIGPYTHON_FASTPROXY", 0);
+    }
+
     if (doxygen)
       doxygenTranslator = new PyDocConverter(doxygen_translator_flags);
 
@@ -617,6 +621,10 @@ public:
 
     if (builtin) {
       Printf(f_runtime, "#define SWIGPYTHON_BUILTIN\n");
+    }
+
+    if (fastproxy) {
+      Printf(f_runtime, "#define SWIGPYTHON_FASTPROXY\n");
     }
 
     Printf(f_runtime, "\n");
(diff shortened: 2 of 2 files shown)


ojwb:
All Python tests have passed in CI so merging.

geographika:
@ojwb - I've been testing the master branch on the MapServer SWIG bindings and run into a `mapscriptPYTHON_wrap.c(47897,59): error C2065: 'SWIG_PyInstanceMethod_New': undeclared identifier` which could be related to this commit?

https://github.com/geographika/python-mapscript/runs/5527964139?check_suite_focus=true#step:6:426

SWIG master branch (broken) output: https://raw.githubusercontent.com/geographika/python-mapscript/daf21f60212a750c64e14250e519ba3e06776184/src/3.8/mapscriptPYTHON_wrap.c

SWIG 4.0.2 (working output): https://raw.githubusercontent.com/geographika/python-mapscript/main/src/3.8/mapscriptPYTHON_wrap.c

The working version includes:

```
/* Method creation and docstring support functions */

SWIGINTERN PyMethodDef *SWIG_PythonGetProxyDoc(const char *name);
SWIGINTERN PyObject *SWIG_PyInstanceMethod_New(PyObject *SWIGUNUSEDPARM(self), PyObject *func);
SWIGINTERN PyObject *SWIG_PyStaticMethod_New(PyObject *SWIGUNUSEDPARM(self), PyObject *func);
```

Looking at the guards I'm not sure why there are any references to `SWIG_PyStaticMethod_New`. The bindings are built without `-fastproxy`. It does use the `-py3` flag though:

https://github.com/MapServer/MapServer/blob/bc665eba612418b91f1424508ed7f2a0f57a71be/mapscript/python/CMakeLists.txt#L14

I'll continue debugging. Feel free to ignore this comment, or I can open it as a new issue.

ojwb:
`SWIG_PyInstanceMethod_New` actually gets wrapped by SWIG, which seems ... odd.

```
static PyMethodDef SwigMethods[] = {
	 { "SWIG_PyInstanceMethod_New", SWIG_PyInstanceMethod_New, METH_O, NULL},
```

Do the interface files in this project refer to `SWIG_PyInstanceMethod_New` directly?

geographika:
There is no use of `SWIG_PyInstanceMethod_New` anywhere in the codebase, but I can't rule out non-standard or legacy use of SWIG features as the bindings have been around for over 15 years. 

From my reading of the commit `SWIG_PyInstanceMethod_New` should either be present 3 times or none, not 2 (which is the cause of the error).

ojwb:
I was wrong above - when it is used it is actually used from Python code:

```
Printf(f_shadow, "_swig_new_instance_method = %s.SWIG_PyInstanceMethod_New\n", module);
```

> From my reading of the commit `SWIG_PyInstanceMethod_New` should either be present 3 times or none, not 2 (which is the cause of the error).

OK, there's:

* the C++ forward declaration `SWIGINTERN PyObject *SWIG_PyInstanceMethod_New(PyObject *SWIGUNUSEDPARM(self), PyObject *func);` which is gated by the SWIG-time conditional `#if defined SWIGPYTHON_FASTPROXY && !defined SWIGPYTHON_BUILTIN`
* the C++ definition, which is gated by the SWIG-time conditional `#if defined SWIGPYTHON_FASTPROXY && !defined SWIGPYTHON_BUILTIN`
* the Python use generated by `Printf(f_shadow, "_swig_new_instance_method = %s.SWIG_PyInstanceMethod_New\n", module);` which is gated by `if (!builtin && fastproxy) {`
* the wrapping of the C++ function in Python by `Printf(line, "\t { \"%s\", %s, METH_O, NULL},\n", name, name);` which is gated by `if (!builtin && fastproxy) {`

Your generated wrappers don't have the first but do have the second and third.  You didn't show the generated `.py` so I can't easily tell if you have the last one.  If you don't have the last one, then having the second and third is wrong; if you do, missing the first is the problem.

Oh except ... looking at generated files from the testsuite, they actually contain the conditional from the second, so that's not being handled at SWIG-time like it should be.  That's at least part of the problem here then - I'll fix that and then let's see where we are.

geographika:
@ojwb - thank you! Error gone when using SWIG with the [b4a92ae](https://github.com/swig/swig/commit/b4a92ae34f6d39825bc0eba056d6983575db4695) update. 
All references to `SWIG_PyInstanceMethod_New` are now absent in the generated C file: https://raw.githubusercontent.com/geographika/python-mapscript/main/src/3.7/mapscriptPYTHON_wrap.c

ojwb:
Great, thanks for confirming.

PR #2194, opened 2022-02-04
Auto-detect non-assignable classes

sethrj:
Classes with references or const data are now marked as 'noassign'. This removes the need for numerous explicit `private: operator=` declarations in SWIG.

### DIFF
--- a/Source/Modules/allocate.cxx
+++ b/Source/Modules/allocate.cxx
@@ -779,6 +779,15 @@ Allocate():
 
       if (Swig_storage_isstatic(n)) {
 	Setattr(n, "cplus:staticbase", inclass);
+      } else if (Cmp(Getattr(n, "kind"), "variable") == 0) {
+        /* Check member variable to determine whether assignment is valid */
+        if (GetFlag(n, "feature:immutable")) {
+          /* Can't assign a class with an immutable member variable */
+	  Setattr(inclass, "allocate:noassign", "1");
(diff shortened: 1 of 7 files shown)


Source/Modules/allocate.cxx
```diff
@@ -779,6 +779,15 @@ Allocate():
 
       if (Swig_storage_isstatic(n)) {
 	Setattr(n, "cplus:staticbase", inclass);
+      } else if (Cmp(Getattr(n, "kind"), "variable") == 0) {
+        /* Check member variable to determine whether assignment is valid */
+        if (GetFlag(n, "feature:immutable")) {
+          /* Can't assign a class with an immutable member variable */
```
ojwb, line 785: I'm not sure if this is quite right.

The documentation doesn't really spell it out, but `%immutable` actually prevents direct assignment being wrapped.  However the variable in C/C++ can still change - consider this:

```
%module example
%{
int a = 42;
void set_a(int v) { a = v; }
%}
%immutable;
int a;
void set_a(int v);
```

Then:

```
>>> import example
>>> example.cvar.a
42
>>> example.set_a(1)
>>> example.cvar.a
1
```

If calling `set_a()` is allowed to change "immutable" `a`, then I find it hard to argue that calling a copy constructor should not be allowed to.

sethrj: Ah, I see. That makes sense; let me remove this clause and see if anything surprising happens. Thanks for the review!

PR #2196, opened 2022-02-06
Enable multiple 'fragment' keywords to be attached to typemaps

sethrj:
@ojwb Here is another SWIG "core" change for the Fortran branch.

This is consistent with the use of `%fragment` and I think more intuitive than the current "alternate syntax" method `%typemap("in", "header", fragment="frag_a,frag_b,frag_c")`. I think it's also less error-prone, since at present all but the last `fragment=` keyword are silently ignored.

The main functional advantage over the extant comma-separated fragments is that the new method supports fragment type specialization for each of the fragments used.

### DIFF
--- a/Examples/test-suite/fragments.i
+++ b/Examples/test-suite/fragments.i
@@ -34,3 +34,86 @@ int foo(int hola)
 }
 
 %}
+
+/* Instantiate multiple fragments at once using multiple keywords */
+typedef int int_infrag3;
+typedef int explicit_frag3;
+
+%fragment("explicit_frag1","header", noblock=1) {
+typedef int explicit_frag1;
+}
+
+%fragment("explicit_frag2","header", noblock=1, noblock=1) {
+typedef explicit_frag1 explicit_frag2;
+}
+
+%fragment("explicit_frag3","header",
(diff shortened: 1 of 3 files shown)


Examples/test-suite/fragments.i
```diff
@@ -34,3 +34,86 @@ int foo(int hola)
...
+
+/* Instantiate multiple fragments at once using multiple keywords */
+typedef int int_infrag3;
+typedef int explicit_frag3;
+
+%fragment("explicit_frag1","header", noblock=1) {
+typedef int explicit_frag1;
+}
+
+%fragment("explicit_frag2","header", noblock=1, noblock=1) {
```
ojwb, line 46: Is the double `noblock=1` intentional?

sethrj: It's been a long time since I wrote this test, but perhaps it's to test that nothing breaks with duplicate non-fragment kwargs? I can delete the duplication.

wsfulton: I remove the additional noblock in d59cbf1c39efd4541275b9868514dd24b9880ba7

sethrj:
Thanks @wsfulton !

PR #2197, opened 2022-02-06
Print rename warnings except anonymous template methods

sethrj:
This set of changes reduces code duplication in `%namewarn` directives, prints the renamed variable when using a regex-based warning, and omits printing the name when the conflicting name is an anonymous template (which won't generate any wrapper code).


sethrj:
The cygwin builds [appear to be failing](https://ci.appveyor.com/project/swig/swig/builds/42470837/job/se8t17d09x2kdosn) due to an unrelated reason:
```
Can't locate threads.pm in @INC (you may need to install the threads module) (@INC contains: /usr/share/automake-1.16 /usr/local/lib/perl5/site_perl/5.32/x86_64-cygwin-threads /usr/local/share/perl5/site_perl/5.32 /usr/lib/perl5/vendor_perl/5.32/x86_64-cygwin-threads /usr/share/perl5/vendor_perl/5.32 /usr/lib/perl5/5.32/x86_64-cygwin-threads /usr/share/perl5/5.32) at /usr/share/automake-1.16/Automake/ChannelDefs.pm line 23.
[116](https://ci.appveyor.com/project/swig/swig/builds/42470837#L116)
```

ojwb:
Yes, a few of the appveyor builds seem to be failing recently.

There's also some flakiness in the Ruby builds on Linux (I think there's an open ticket for it, but I didn't trivially find it just now).

This looks good to me, except a testcase for the regex case would be useful to have.

sethrj:
> This looks good to me, except a testcase for the regex case would be useful to have.

@ojwb I was having trouble getting the regextarget tests to work, and since regex is somewhat optional (SWIG can technically be built without PCRE right?) I've removed that option. Simplifying the rename warnings and ignoring anonymous template matches is a net gain though.

ojwb:
> SWIG can technically be built without PCRE right?

`Doc/Manual/Preface.htm` seems to imply it's a hard dependency:

> PCRE2 needs to be installed on your system to build SWIG, in particular
pcre2-config must be available. If you have PCRE2 headers and libraries but not
pcre2-config itself or, alternatively, wish to override the compiler or linker
flags returned by pcre2-config, you may set PCRE2_LIBS and PCRE2_CFLAGS variables   
to be used instead. And if you don't have PCRE2 at all, the configure script
will provide instructions for obtaining it.

But `configure.ac`'s error message if it isn't says a whole bunch of stuff, but ends with:

> Use configure --without-pcre to disable regular expressions support in SWIG (not recommended).

No idea if that actually option actually works (CI doesn't seem to test it).  I'd tend to argue it'd be better to just make it a hard dependency and then users can rely on it (and we'd have one fewer option to worry about keeping working).  I guess the question is if there's a platform that PCRE2 doesn't support but SWIG currently does.

But anyway it looks like the testsuite already just assumes PCRE was enabled - e.g. I can't see any mechanism to disable trying `rename_pcre_encoder.i`.

ojwb:
I've merged what's here as it does seem a useful improvement, but I'd certainly encourage sorting out the `regextarget` change as that also seems useful.

sethrj:
> No idea if that option actually works (CI doesn't seem to test it).  I'd tend to argue it'd be better to just make it a hard dependency and then users can rely on it (and we'd have one fewer option to worry about keeping working).  I guess the question is if there's a platform that PCRE2 doesn't support but SWIG currently does.

I completely agree it should be a hard dependency, but I recall on my old Fortran branch being directed *not* to rely on the regex features since they are optional.

ojwb:
> I completely agree it should be a hard dependency

#2207

PR #2199, opened 2022-02-07
Mark 'externc' storage for variables

### LINKED ISSUE #2198: `externc` storage is not set for variables inside `extern "C"` blocks
The `storage` attribute for variables is only set to `externc` when the declaration itself is annotated as `extern "C"`, but not when they're embedded in an `extern "C"` block. For example, the following only marks `int1` as `externc`, but it sets storage of `int2` only to `extern`.
```c++
%module extern_c

%inline %{
extern "C" int int1;
extern "C" { extern int int2; }
%}

%{
extern "C" {
  int int1 = 123;
  int int2 = 456;
}
%}
```

The `-debug-module 1` output shows:
```
+++ cdecl - 0x7fd80e931be0 ----------------------------------------
| sym:name     - "int1"
| name         - "int1"
| decl         - ""
| storage      - "externc"
| kind         - "variable"
| type         - "int"
| sym:symtab   - 0x7fd80e800620
| sym:overname - "__SWIG_0"
| 
+++ extern - 0x7fd80e931fe0 ----------------------------------------
| name         - "C"

      +++ cdecl - 0x7fd80e931f00 ----------------------------------------
      | sym:name     - "int2"
      | name         - "int2"
      | decl         - ""
      | storage      - "extern"
      | kind         - "variable"
      | type         - "int"
      | sym:symtab   - 0x7fd80e800620
      | sym:overname - "__SWIG_0"
      | 
```

This doesn't affect anything in current SWIG but is needed for the Fortran `%fortranbindc` feature.

sethrj:
Fixes #2198 . The example code there now is  correctly tagged with `externc` for both declarations:

```
+++ cdecl - 0x7fc2b5931be0 ----------------------------------------
| sym:name     - "int1"
| name         - "int1"
| decl         - ""
| storage      - "externc"
| kind         - "variable"
| type         - "int"
| sym:symtab   - 0x7fc2b5800620
| sym:overname - "__SWIG_0"
| 
+++ extern - 0x7fc2b5931fe0 ----------------------------------------
| name         - "C"

      +++ cdecl - 0x7fc2b5931f00 ----------------------------------------
      | sym:name     - "int2"
      | name         - "int2"
      | decl         - ""
      | storage      - "externc"
      | kind         - "variable"
      | type         - "int"
      | sym:symtab   - 0x7fc2b5800620
      | sym:overname - "__SWIG_0"
```

I don't think this is currently possible to test in mainline SWIG because no languages rely on the feature, but without this change a Fortran module language test fails.

### DIFF
--- a/Examples/test-suite/cpp11_thread_local.i
+++ b/Examples/test-suite/cpp11_thread_local.i
@@ -16,9 +16,9 @@ thread_local static int tsval;
 extern thread_local int etval;
 thread_local extern int teval;
 extern "C" thread_local int ectval;
-extern "C" { thread_local int ectval2; }
+extern "C" { thread_local int ectval2 = 56; }
 extern "C++" thread_local int ecpptval;
-extern "C++" { thread_local int ecpptval2; }
+extern "C++" { thread_local int ecpptval2 = 67; }
 
 thread_local int ThreadLocals::stval = 11;
 thread_local int ThreadLocals::tsval = 22;
@@ -32,8 +32,6 @@ thread_local const int ThreadLocals::tscval99;
 // externs
 thread_local int etval = 33;
 thread_local int teval = 44;
-thread_local int ectval = 55;
-thread_local int ectval2 = 56;
+extern "C" thread_local int ectval = 55;
 thread_local int ecpptval = 66;
-thread_local int ecpptval2 = 67;
 %}
--- a/Examples/test-suite/cpp11_thread_local.i
+++ b/Examples/test-suite/cpp11_thread_local.i
@@ -18,6 +18,7 @@ thread_local extern int teval;
 extern "C" thread_local int ectval;
 extern "C" { thread_local int ectval2; }
 extern "C++" thread_local int ecpptval;
+extern "C++" { thread_local int ecpptval2; }
 
 thread_local int ThreadLocals::stval = 11;
 thread_local int ThreadLocals::tsval = 22;
@@ -34,4 +35,5 @@ thread_local int teval = 44;
 thread_local int ectval = 55;
 thread_local int ectval2 = 56;
 thread_local int ecpptval = 66;
+thread_local int ecpptval2 = 67;
 %}
--- a/Examples/test-suite/cpp11_thread_local.i
+++ b/Examples/test-suite/cpp11_thread_local.i
@@ -32,6 +32,10 @@ thread_local const int ThreadLocals::tscval99;
 // externs
 thread_local int etval = 33;
 thread_local int teval = 44;
-extern "C" thread_local int ectval = 55;
+extern "C" {
+thread_local int ectval = 55;
+}
+extern "C++" {
 thread_local int ecpptval = 66;
+}
 %}
--- a/Examples/test-suite/cpp11_thread_local.i
+++ b/Examples/test-suite/cpp11_thread_local.i
@@ -16,6 +16,7 @@ thread_local static int tsval;
 extern thread_local int etval;
 thread_local extern int teval;
 extern "C" thread_local int ectval;
+extern "C" { thread_local int ectval2; }
 extern "C++" thread_local int ecpptval;
 
 thread_local int ThreadLocals::stval = 11;
@@ -31,5 +32,6 @@ thread_local const int ThreadLocals::tscval99;
 thread_local int etval = 33;
 thread_local int teval = 44;
 thread_local int ectval = 55;
+thread_local int ectval2 = 56;
 thread_local int ecpptval = 66;
 %}
--- a/Examples/test-suite/extern_c.i
+++ b/Examples/test-suite/extern_c.i
@@ -5,6 +5,7 @@ extern "C" {
 void RealFunction(int value);
 typedef void Function1(int value); // Fails
 typedef int Integer1;
+int Integer3;
 }
 typedef void Function2(int value); // Works
 typedef int Integer2;
@@ -27,5 +28,14 @@ Hook2_t hook2;
 
 extern "C" typedef int Integer;
 Integer int1;
+extern "C" int int2;
+extern "C" { extern int int3; }
+extern "C" { int int4 = 789; }
 %}
 
+%{
+extern "C" {
+  int int2 = 123;
+  int int3 = 456;
+}
+%}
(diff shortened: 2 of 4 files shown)


ojwb:
I assume the logic of the current test for adding `externc` is that only functions get name mangled, so `extern "C" int x;` and `int x;` are equivalent.

For example, compiling with with `g++ -S` and `g++ -DEXTERNC -S` gives me identical assembler output:

```
#ifdef EXTERNC
extern "C"
#endif
int xyzzy = 12;
```

So why does this matter to Fortran?

sethrj:
I think the mangling of variables in the default global namespace is an unspecified implementation choice rather than a requirement of the language. From reading the [dcl.link](https://eel.is/c++draft/dcl.link) section of the standard, even though most of the examples are functions, it uses "variables and functions".

Oftentimes GCC makes implementation choices that tend to allow more lenient behavior that MSVC and other unusual compilers choke on. This [godbolt demo](https://godbolt.org/z/E9TjabhMx) suggests that might be the case here too: the assembly identifier for a default (C++ linkage) integer is `|int x|` whereas the C linkage ones are `y` and `z`.

ojwb:
Ah OK, that makes sense.  FWIW, it'll be what the [Itanium C++ ABI](https://itanium-cxx-abi.github.io/cxx-abi/) specifies rather than GCC's implementation choices that determine the mangling GCC uses.

Looking at the parser code, I think there's a case that will still get mishandled:

```
extern "C" thread_local int x;
```

This `storage_class` rule throws away the `"externc"` from `extern_string` for cases such as the above:

```
               | extern_string THREAD_LOCAL { $$ = "thread_local"; }
```

Probably that wants to be something like: `$$ = $1; Append($$, " thread_local");`

sethrj:
@ojwb I'm not having any luck with your suggestion: I'm getting segfaults for the test case
```cxx
extern "C" thread_local int int_tl;
```
as well as warnings when building the lex-generated code about discarding const qualifiers.

Even this attempt fails in the same way:
```lex
           | extern_string THREAD_LOCAL { $$ = $1;
                if ($$) {
                  Append($$, " thread_local");
                } else {
                  $$ = "thread_local";
                }}
```

Any suggestions? Currently the above code generates:
```
+++ cdecl - 0x7ffbbf934580 ----------------------------------------
| sym:name     - "int_tl"
| name         - "int_tl"
| decl         - ""
| storage      - "thread_local"
| kind         - "variable"
| type         - "int"
| sym:symtab   - 0x7ffbbf800620
| sym:overname - "__SWIG_0"
```
and
```cxx
extern "C" { thread_local int int_tl2; }
```
generates 
```
      +++ cdecl - 0x7fa5f3934780 ----------------------------------------
      | sym:name     - "int_tl2"
      | name         - "int_tl2"
      | decl         - ""
      | storage      - "externc"
      | kind         - "variable"
      | type         - "int"
      | sym:symtab   - 0x7fa5f3800620
      | sym:overname - "__SWIG_0"
```
so neither is quite right.

On master, both generate "thread_local" storage and ignore "externc". I'm not certain that "thread_local" even works across the C/C++ language barrier from reading about it (`_Thread_local` seems to be a C equivalent?) so perhaps this is a moot point.

sethrj:
Hmm guess I have to add CXX11 guards around `thread_local`?
```
extern_c_wrap.cxx:329:12: error: ‘thread_local’ does not name a type; did you mean ‘pthread_rwlock_t’?
 extern "C" thread_local int int_tl;
```

ojwb:
C11 has `thread_local` (via a compatibility define):

https://en.cppreference.com/w/c/thread/thread_local

I'd expect the "thread local" feature to work across C and C++.

Generally features requiring a particular C++ standard version are put in a test with a prefix like `cpp11_` (and are then listed in `CPP11_TEST_CASES`, etc in `Examples/test-suite/common.mk`).  Probably putting it in `Examples/test-suite/cpp11_thread_local.i` would make sense, and that even has the non-block cases already:

```
extern "C" thread_local int ectval;
extern "C++" thread_local int ecpptval;
```

I don't actually see anywhere SWIG makes use of `thread_local`, though probably it should still get put into the `storage` value in case there's a need for it in the future.

I think I see why there's a segfault - `Append()` is for a DOH `String*` but the value being appended to was just a C string literal (set by: `$$ = "thread_local";`).  Probably this would work:

```
           | extern_string THREAD_LOCAL {
                if (Equal($1, "extern")) {
                  $$ = "extern thread_local";
                } else {
                  $$ = "externc thread_local";
                }
           }
```

Or turn it into a `String*` but then that either means deliberately leaking it, or else wider changes such that it's always a `String*` so `Delete()` can be called on it.

sethrj:
@ojwb Your suggestion fixed the `extern "C" thread_local int ectval;` case, but in order to fix `extern "C" { thread_local int ectval2; }` I had to add a few lines inside the `extern {` postprocessing (which traverses the daughter nodes).

I also encountered (and fixed) a spurious warning for `extern "C++" {` which is valid.

Presently on master:
```c++
extern "C" thread_local int ectval;
extern "C" { thread_local int ectval2; }
```
both get `storage: "thread_local"`, but now on this branch, both get `storage: "externc thread_local"`.

wsfulton:
Please don't use assert in the Python tests (extern_c_runme.py in the pull request).

sethrj:
@wsfulton Sorry about that -- I'm used to pytest which relies on `assert` for testing.  I've uploaded a fix (including to another assert that snuck in #2197) to #2211.

Can I ask the rationale for avoiding `assert`?

ojwb:
I assume it's that `assert` becomes a no-op if Python optimisation is turned on:

```
$ python3 -c 'assert(0)'
  File "<string>", line 1, in <module>
AssertionError
$ python3 -O -c 'assert(0)'
$ PYTHONOPTIMIZE=x python3 -c 'assert(0)'
$
```

It'd be good if we could somehow "poison" `assert` to prevent new instances being introduced.  I removed a pile of uses only a couple of weeks ago in 25f996a5c8141ba1a9e3ab7b571aa54900f1bd07.  If it's the norm with some python test harnesses, this is likely to be a recurring problem.

sethrj:
> I assume it's that `assert` becomes a no-op if Python optimisation is turned on

That was my first thought too, but since we only run the test cases through the makefile harness we have control over the flags...

ojwb:
We control the command line flags, but we don't currently ensure `PYTHONOPTIMIZE` isn't set in the environment.  We presumably could fairly easily though.

As I said above I'm only assuming optimisation is the reason.  Maybe it's actually more about consistency between testcases, or something else I haven't thought of.

@wsfulton It would be helpful if you could give an explicit rationale for "don't use assert in the Python tests", and document it somewhere more permanent as this is clearly a recurring problem.

PR #2202, opened 2022-02-09
Fix typos

rex4539:


Source/DOH/fio.c
```diff
@@ -19,13 +19,13 @@
 static DOH *encodings = 0;	/* Encoding hash */
 
 /* -----------------------------------------------------------------------------
- * Writen()
+ * Written()
```
ojwb, line 22: This one is not a typo - it's "write n" (as is clear from the description).

rex4539: Reverted.

Doc/Manual/Extending.html
```diff
@@ -2991,7 +2991,7 @@ <H3><a name="Extending_nn38">40.10.7 Low-level code generators</a></H3>
   /* create the wrapper object */
   Wrapper *wrapper = NewWrapper();
 
-  /* create the functions wrappered name */
+  /* create the functions wrapped name */
```
ojwb, line 2994: I think that should really say "create the wrapper function's name" - I'll merge this and then push a fix for that.

PR #2204, opened 2022-02-10
Update python manual scripts to py3 compatibility

sethrj:
This minor infrastructure change allows python3 to execute the manual helper scripts whilst retaining compatibility with Python 2.5.


Doc/Manual/makechap.py
```diff
@@ -11,10 +11,10 @@
 # then change the heading link name to something that does not look like an
 # autogenerated link name.
 ###############################################################################
+from __future__ import print_function
```
ojwb, line 14: Maybe I'm missing something, but it doesn't look to me like this is needed (here or in the other script).  All the calls to `print` after the patch seem to pass a single parenthesised value, which should be compatible with Python2 as-is.

sethrj: Good point, but it's also a little bit deceptive: changing `print("Hello world")` to `print("Hello", "world")` ends up printing a tuple instead of the same string 😄 . If you think it's better to revert the import here I'm happy to oblige—hopefully the next time someone has to change this script, there will be no more python 2 installations in regular use 😅

ojwb: Probably best to see what @wsfulton thinks as he's the main one to use these scripts (and the only one to have modified them since 2005!)

I don't think I've ever run them myself.

sethrj: I guess new target languages don't get added very often 😉 so there's not much need.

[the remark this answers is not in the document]
sethrj: I just realized after pushing but 2.7 still requires the `__future__` to turn `print` from a statement to a function. I've removed it, but thankfully it looks like all calls to `print` take a single argument so on Python 2.7 `print("foo")` is really `print "foo"` which is fine.

sethrj: Done.

ojwb:
Looks good to me, and you've address wsfulton's review comment so merging.

PR #2222, opened 2022-03-02
Python: avoid unused SwigMethods_proxydocs var

jschueller:
Avoids a warning since 3aa302c:
```
...PYTHON_wrap.cxx:4556:20: error: ‘SwigMethods_proxydocs’ defined but not used [-Werror=unused-variable]
 4556 | static PyMethodDef SwigMethods_proxydocs[] = {
```

cc @wsfulton

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -819,7 +819,8 @@ public:
     Printf(f_wrappers, "%s\n", methods);
     Append(methods_proxydocs, "\t { NULL, NULL, 0, NULL }\n");
     Append(methods_proxydocs, "};\n");
-    Printf(f_wrappers, "%s\n", methods_proxydocs);
+    if ((fastproxy && !builtin) || have_fast_proxy_static_member_method_callback)
+      Printf(f_wrappers, "%s\n", methods_proxydocs);
 
     if (builtin) {
       Dump(f_builtins, f_wrappers);


Source/Modules/python.cxx
```diff
@@ -819,7 +819,8 @@ class PYTHON:public Language {
     Printf(f_wrappers, "%s\n", methods);
     Append(methods_proxydocs, "\t { NULL, NULL, 0, NULL }\n");
     Append(methods_proxydocs, "};\n");
-    Printf(f_wrappers, "%s\n", methods_proxydocs);
+    if (fastproxy)
+      Printf(f_wrappers, "%s\n", methods_proxydocs);
```
ojwb, line 823: Looks good, except the only use of this seems to be in `Lib/python/pyinit.swg` inside a conditional:

```
#if defined SWIGPYTHON_FASTPROXY && !defined SWIGPYTHON_BUILTIN
```

So I wonder if the test here should really be `(fastproxy && !builtin)` instead?  `builtin` can be true or false here, as shown by the test just below.

ojwb: Oh, I see you tried that in a previous iteration and it failed:

https://github.com/swig/swig/runs/5394070693?check_suite_focus=true

ojwb: Looks like `((fastproxy && !builtin) || have_fast_proxy_static_member_method_callback)` is the exact condition it's used under, or did you already check that and find it doesn't work?  It's hard to fully see what's been tried with all the force-pushing.

jschueller: you're right

ojwb:
All the python jobs have passed so merged - thanks.

ojwb:
No need for that final rebase and force push - it triggers an extra pointless CI build (which I've manually cancelled this time).  I can just squash the changes down when I merge anyway.

jschueller:
ok

PR #2223, opened 2022-03-03
Fail cleanly on allocation failures

### LINKED ISSUE #1901: swig 4.0.2 assertion failed on OpenABM-Covid19 project
SWIG 4.0.2 hits an assertion failure when generating R bindings for this project:
https://github.com/olegat/OpenABM-Covid19

SWIG 3.0.12 is not affect by hit and works as intended.

## Reproduction Steps
**Requisites**
- SWIG 4.0.2 in `PATH`
- make in `PATH` (optional)

**Steps**
```
git clone https://github.com/olegat/OpenABM-Covid19
cd OpenABM-Covid19
git checkout 860f4d32479e959a1eec0adedf41aee8306ee4a5
make Rswig
```
Alternatively instead of using `make`, you can run the SWIG command line directly:
```
swig -r -package OpenABMCovid19 -Isrc -o src/covid19_wrap_R.c -outdir R src/covid19.i
```

## Expected Result (swig 3.0.12)

`make Rswig` creates these files:
- R/OpenABMCovid19.R
- src/covid19_wrap_R.c

**Output** (using SWIG from Cygwin64 on Windows)
```
swig -r -package OpenABMCovid19 -Isrc -o src/covid19_wrap_R.c -outdir R src/covid19.i
src/model.h:101: Warning 314: 'next' is a R keyword, renaming to '_next'
src/constant.h:15: Warning 314: 'FALSE' is a R keyword, renaming to '_FALSE'
src/constant.h:16: Warning 314: 'TRUE' is a R keyword, renaming to '_TRUE'
src/individual.h:75: Warning 314: 'next' is a R keyword, renaming to '_next'
src/individual.h:80: Warning 314: 'next' is a R keyword, renaming to '_next'
src/individual.h:90: Warning 314: 'next' is a R keyword, renaming to '_next'
sed -i 's/R_registerRoutines/R_useDynamicSymbols(dll,0);R_registerRoutines/' src/covid19_wrap_R.c
sed -i 's/.Call("R_SWIG_debug_getCallbackFunctionData"/.Call("R_SWIG_debug_getCallbackFunctionData", PACKAGE="OpenABMCovid19"/' R/OpenABMCovid19.R
sed -i 's/.Call("R_SWIG_R_pushCallbackFunctionData"/.Call("R_SWIG_R_pushCallbackFunctionData", PACKAGE="OpenABMCovid19"/' R/OpenABMCovid19.R
sed -i 's/ans <- new("_p_char"/#ans <- new("_p_char"/' R/OpenABMCovid19.R
```

## Actual Result (swig 4.0.2)

No source files are generated.

**Output** (using SWIG from Homebrew on macOS)
```
swig -r -package OpenABMCovid19 -Isrc -o src/covid19_wrap_R.c -outdir R src/covid19.i
src/model.h:101: Warning 314: 'next' is a R keyword, renaming to '_next'
src/constant.h:15: Warning 314: 'FALSE' is a R keyword, renaming to '_FALSE'
src/constant.h:16: Warning 314: 'TRUE' is a R keyword, renaming to '_TRUE'
src/individual.h:75: Warning 314: 'next' is a R keyword, renaming to '_next'
src/individual.h:80: Warning 314: 'next' is a R keyword, renaming to '_next'
src/individual.h:90: Warning 314: 'next' is a R keyword, renaming to '_next'
Assertion failed: (0), function DohCopy, file DOH/base.c, line 71.
make: *** [src/covid19_wrap_R.c] Abort trap: 6
```

ojwb:
Previously code in the SWIG tool didn't handle allocation failures
well.  Most places didn't check for NULL return from
malloc()/realloc()/calloc() at all, typically resulting in undefined
behaviour, and some places used assert() to check for a NULL return
(which is a misuse of assert() and such checks disappear if built with
NDEBUG defined leaving us back with undefined behaviour).

All C allocations are now done via wrapper functions (Malloc(),
Realloc() and Calloc()) which emit and error and exit with non-zero
status on failure, so a non-NULL return can be relied upon.

Fixes #1901.

### DIFF
--- a/Source/CParse/parser.y
+++ b/Source/CParse/parser.y
@@ -32,7 +32,7 @@
 
 /* We do this for portability */
 #undef alloca
-#define alloca malloc
+#define alloca Malloc
 
 /* -----------------------------------------------------------------------------
  *                               Externals
--- a/Source/CParse/templ.c
+++ b/Source/CParse/templ.c
@@ -620,7 +620,7 @@ static Node *template_locate(String *name, Parm *tparms, Symtab *tscope) {
 	int parms_len = ParmList_len(parms);
 	int *priorities_row;
 	max_possible_partials = Len(partials);
-	priorities_matrix = (int *)malloc(sizeof(int) * max_possible_partials * parms_len); /* slightly wasteful allocation for max possible matches */
+	priorities_matrix = (int *)Malloc(sizeof(int) * max_possible_partials * parms_len); /* slightly wasteful allocation for max possible matches */
 	priorities_row = priorities_matrix;
 	for (pi = First(partials); pi.item; pi = Next(pi)) {
 	  Parm *p = parms;
@@ -818,7 +818,7 @@ success:
     Printf(stdout, "    chosen template:'%s'\n", Getattr(n, "name"));
   }
   Delete(parms);
-  free(priorities_matrix);
+  Free(priorities_matrix);
   return n;
 }
 
--- a/Source/DOH/doh.h
+++ b/Source/DOH/doh.h
@@ -122,6 +122,10 @@
 #define DohIterator        DOH_NAMESPACE(Iterator)
 #define DohFirst           DOH_NAMESPACE(First)
 #define DohNext            DOH_NAMESPACE(Next)
+#define DohMalloc          DOH_NAMESPACE(Malloc)
+#define DohRealloc         DOH_NAMESPACE(Realloc)
+#define DohCalloc          DOH_NAMESPACE(Calloc)
+#define DohFree            DOH_NAMESPACE(Free)
 #endif
 
 #define DOH_MAJOR_VERSION 0
@@ -163,12 +167,11 @@ typedef struct {
 
 /* Memory management */
 
-#ifndef DohMalloc
-#define DohMalloc malloc
-#endif
-#ifndef DohRealloc
-#define DohRealloc realloc
-#endif
+/* Wrappers around malloc(), realloc() and calloc() which never return NULL. */
+extern void *DohMalloc(size_t size);
+extern void *DohRealloc(void *ptr, size_t size);
+extern void *DohCalloc(size_t n, size_t size);
+
 #ifndef DohFree
 #define DohFree free
 #endif
@@ -440,6 +443,10 @@ extern void DohMemoryDebug(void);
 #define Next               DohNext
 #define Iterator           DohIterator
 #define SortList           DohSortList
+#define Malloc             DohMalloc
+#define Realloc            DohRealloc
+#define Calloc             DohCalloc
+#define Free               DohFree
 #endif
 
 #ifdef NIL
--- a/Source/DOH/file.c
+++ b/Source/DOH/file.c
@@ -291,10 +291,6 @@ DOH *DohNewFile(DOHString *filename, const char *mode, DOHList *newfiles) {
     return 0;
 
   f = (DohFile *) DohMalloc(sizeof(DohFile));
-  if (!f) {
-    fclose(file);
-    return 0;
-  }
   if (newfiles)
     Append(newfiles, filename);
   f->filep = file;
@@ -314,8 +310,6 @@ DOH *DohNewFile(DOHString *filename, const char *mode, DOHList *newfiles) {
 DOH *DohNewFileFromFile(FILE *file) {
   DohFile *f;
   f = (DohFile *) DohMalloc(sizeof(DohFile));
-  if (!f)
-    return 0;
   f->filep = file;
   f->fd = 0;
   f->closeondel = 0;
@@ -331,8 +325,6 @@ DOH *DohNewFileFromFile(FILE *file) {
 DOH *DohNewFileFromFd(int fd) {
   DohFile *f;
   f = (DohFile *) DohMalloc(sizeof(DohFile));
-  if (!f)
-    return 0;
   f->filep = 0;
   f->fd = fd;
   f->closeondel = 0;
--- a/Source/DOH/hash.c
+++ b/Source/DOH/hash.c
@@ -173,10 +173,7 @@ static void resize(Hash *h) {
     p = p + 2;
   }
 
-  table = (HashNode **) DohMalloc(newsize * sizeof(HashNode *));
-  for (i = 0; i < newsize; i++) {
-    table[i] = 0;
-  }
+  table = (HashNode **) DohCalloc(newsize, sizeof(HashNode *));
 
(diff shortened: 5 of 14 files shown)


ojwb:
Possibly we should try to call `SWIG_exit()` instead of `exit()` to at least try to delete any files that were being created.

I haven't tried that for now, mostly because that doesn't currently happen in any of the cases I've changed so my patch doesn't make that aspect worse at all.

It's a bit awkward to do if we carry on with what seems to be the current assumption that DOH is a layer below SWIG which might get used elsewhere - I think we'd need to add a way to register a function with DOH which it would call on out of memory conditions.

Perhaps a better alternative would be to register a cleanup function via `atexit()`.  The awkward part there is that the cleanup function doesn't get told what the exit status is, but we could keep `SWIG_exit()` but now have it set a "clean exit" flag (which defaults false in case exit isn't via `SWIG_exit()`) and the cleanup function can then check that to decide whether to remove files.

wsfulton:
Nice tidyup @ojwb.

How about adding an exit function in DOH, so near the startup, SWIG registers SWIG_exit as the exit function for DOH to call in the event of an error. If the register function has not been called, then default to exit().

ojwb:
I can take a look at that, but I'll get this merged first as it's better or the same in all ways than the current code and touches quite a lot of places and so liable to conflict if left unapplied for long.

I also think most if not all of the calls to `abort();` should change to `SWIG_exit(EXIT_FAILURE);` (with an error message output first in cases where there isn't one already) so that generated files get removed.

The developer docs should really mention "use `Malloc(x)` instead of `malloc(x)`", use `SWIG_exit(x)` instead of `exit(x)`, etc, though I suspect people often don't look at them so I'm also going to look at using `#pragma GCC poison` on identifiers like `malloc` so that patches which try to add new uses will fail CI.

ojwb:
55377bdc08ffbe37fbb89566a8754302ae186227 adds DOH `Exit()` and `SetExitHandler()` functions (and uses them everywhere).

ojwb:
735732d721ea18a34013cea43b1a8951408f9fc9 replaces calls to `abort()` with `Exit(EXIT_FAILURE)` so that output files get removed.

ojwb:
747a51f09518eede4b2c6f5e9af8f95865fa4d7d uses `#pragma GCC poison` to prevent direct uses of `malloc()`, `realloc()`, `calloc()`, `free()` and `exit()` being introduced.  There's an awkward issue with Bison's template which check `defined malloc` which triggers GCC's poisoned id checks, so in `parser.y` `malloc` and `free` are not poisoned.  That's probably not a likely place to see them introduced (there are currently no explicit calls to `Malloc()` or `Free()` in that file), but we could post-process the generated `parser.c` if we wanted full identifier poisoning everywhere.

I think that completes the follow-on work here.  The only other thing I wondered is whether it was worth having our own `Assert(x)` macro which was like the standard `assert(x)` but called `Exit()` on failure so that generated files get removed if an assertion fails, but if that's really a concern then I'd argue we're misusing `assert()`.

wsfulton:
In 55377bdc08ffbe37fbb89566a8754302ae186227, the error count has been replaced by 1. I'm intrigued how it is possible that an int on most platforms will return 0 if the count is 256? How about keeping the error count with a ceiling of whatever the max value that will work if this is a problem?

wsfulton:
Sorry, I only just saw your comment @ojwb:

> Also some error statuses have special meanings e.g.
>     those defined by <sysexits.h>.

I wasn't aware of anything following a convention other than exiting with zero or non-zero. So new behaviour seems reasonable.

ojwb:
The full background is that if you pass an `int` to `exit()` (or `main`'s `return`) but at least on POSIX platforms:

> The value of status may be 0, EXIT_SUCCESS, EXIT_FAILURE, [[CX](javascript:open_code('CX'))] [Option Start]  or any other value, though only the least significant 8 bits (that is, status & 0377) shall be available from [wait()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html) and [waitpid()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/waitpid.html); the full value shall be available from [waitid()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/waitid.html) and in the siginfo_t passed to a signal handler for SIGCHLD.

The shell exit status at least uses the traditional `wait()` value:

```
$ cat exit.c
#include <stdlib.h>
int main() { exit(256); }
$ gcc exit.c
$ ./a.out
$ echo $?
0
```

Seems Microsoft copied this feature as https://docs.microsoft.com/en-us/previous-versions/6wdz5232(v=vs.140) says:

> the low-order byte of status is made available to the host environment or waiting calling process, if one exists, after the process exits

There's actually an additional complication as a process which is killed by a signal traditionally exits with status 128+signal_number:

```
$ cat
^C
$ echo $?
130
```

POSIX says "Implementations are encouraged to choose exit values greater than 256 to indicate programs that terminate by a signal so that the exit status cannot be confused with an exit status generated by a normal termination" but it seems that's not caught on - I guess because it risks breaking assumptions in existing working code.

So in practice values from 0 to 127 seem to be pretty portably but all C++ actually guarantees about `exit()` is:

> If exit_code is 0 or [EXIT_SUCCESS](https://en.cppreference.com/w/cpp/utility/program/EXIT_status), an implementation-defined status indicating successful termination is returned. If exit_code is [EXIT_FAILURE](https://en.cppreference.com/w/cpp/utility/program/EXIT_status), an implementation-defined status indicating unsuccessful termination is returned. In other cases implementation-defined status value is returned.

wsfulton:
Hi @ojwb, I had some problems with the gcc poison macros when building SWIG and so have commented this bit out. See 77853770bdaec814e0e231856c4590f35a8578bb. I think it needs a rethink making sure the poison macros only apply to SWIG code and not external/system/library headers. Sorry, I've just got too much to do to have a go at fixing myself.

ojwb:
Sorry, I didn't really consider this possibility but I can see the issue now.  Unfortunately I don't think there's a way to say "poison but not in system headers" so we'd have to very carefully restructure the code such that the poisoning always happens after any non-SWIG headers have been included just in case a system/external header tries to reference one of these symbols.  I think that's just going to end up brittle in the face of future changes though.

So probably the way to go to help prevent direct calls to these functions being added in future changes is to have a CI build with these pragmas on, since it seems fine with modern glibc.  If that proves problematic too we can rethink or give up the idea, but I think some sort of automatic enforcement of this would be better than relying on reviewers spotting attempts to introduce direct calls to these functions.

Disabling for now definitely makes sense.  I'll take a look.

Issue #2142, opened 2022-01-12
%attribute fails with template classes


keithmballard:
I have class ```NodeAccessorBase``` in the namespace ```mesh``` for which I would like to add an attribute based on the underlying C++ getter method.  So I added an attribute and templated type like:
```
%attribute(mesh::NodeAccessorBase, mesh::GlobalIndex, g, global_index);
%template(NodeAccessor) mesh::NodeAccessorBase<mesh::NodeContainer>;
```
However, I get the following compiler error in `_wrap_NodeAccessor_g_get`:
```
Error C3861 'mesh_NodeAccessorBase_Sl_mesh_NodeContainer_Sg__g_get': identifier not found <swig module>
```

I also tried the following:
```
%template(NodeAccessor) mesh::NodeAccessorBase<mesh::NodeContainer>;
%attribute(NodeAccessor, mesh::GlobalIndex, g, global_index);
```
But the Python class for ```NodeAccessor``` did not contain an attribute ```g```.

Am I missing something, or is this a bug with %attribute for templated classes?  Any help would be greatly appreciated.

Note: I am using the release version of Swig 4.0.2 on Windows 10 with Visual Studio 2022.

ojwb:
`%attrbiute` is somewhat lacking in documentation (nothing in the manual) but from a comment in `Lib/typemaps/attribute.swg` (my emphasis):

> Use %attribute when you have a pair of get/set methods to a **primitive type** like in:

A templated class is not a primitive type!

You probably want:

> The %attributeval is the same as %attribute, but should be used when the type is a class/struct (ie a non-primitive type) and when the get and set methods return/pass by value.

Or `%attributeref` is it's passed/returned by reference.

KrisThielemans:
@owjb's answer is about the return type (second argument), while the error message is about the template class in the first argument. I get the same error in a case like this
```c++
template <class T> struct A
{
  T a() const;
  void a(T&);
};
```
The following worked for me
```swig
%attributeref(A<unsigned int>, unsigned int, a)
```

Although having to do this for every instantiation is annoying, it seems quite hard to solve, so for me an update to the documentation could be sufficient.

I believe I can only get this to work if I use `%attribute` with a full specialisation for 2 reasons:
[...]

PR #2224, opened 2022-03-04
add doc on %attribute and templates

KrisThielemans:
Documents how to use `%attribute` with templates. Addresses #2142.

@ojwb does this look ok? Apologies for the multiple commits. Not a good idea to try to do this via GitHub's edit feature... Please squash-merge if it's fine.

### DIFF
--- a/Doc/Manual/Library.html
+++ b/Doc/Manual/Library.html
@@ -2327,6 +2327,20 @@ shared_ptr which has <tt>%naturalvar</tt> turned on in
 <tt>%shared_ptr</tt>.
 </p>
 
-
+<H4>12.5.3 &percnt;attribute and C++ templates</H4>
+<tt>%attribute</tt> and friends have to be used on fully instantiated templates. For example
+<div class="code">
+<pre>
+%attributeref(A&lt;int&gt;, int, a)
+%inline %{
+  template &lt;class T&gt; struct A
+  {
+    T a() const;
+    void a(T&);
+  };
+%}
+</pre>
+</div>
+This means that it has to be repeated if you want to use different instantiations.
 </body>
 </html>
--- a/Doc/Manual/Library.html
+++ b/Doc/Manual/Library.html
@@ -2346,7 +2346,7 @@ shared_ptr which has <tt>%naturalvar</tt> turned on in
 </pre>
 </div>
 <p>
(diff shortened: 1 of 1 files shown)


wsfulton (changes requested):
Could you please use the SWIG coding style wrt braces and ref placement. Am also confused as you mention instantiating the template but there is no instantiation!

Doc/Manual/Library.html
```diff
@@ -45,6 +45,9 @@ <H1><a name="Library">12 SWIG library</a></H1>
 <ul>
 <li><a href="#Library_nn17">exception.i</a>
 <li><a href="#Library_attributes">attribute.i</a>
+<ul>
+<li><a href="#Library_attribute_templates">&percnt;attribute and C++ templates</a></li>
```
ojwb, line 49: The rest of the manual just uses a literal `%`.

Doc/Manual/Library.html
```diff
@@ -2327,6 +2330,20 @@ <H3><a name="Library_attributes">12.5.2 attribute.i</a></H3>
...
+%inline %{
+  template &lt;class T&gt; struct A
+  {
+    T a() const;
+    void a(T&);
+  };
+%}
+</pre>
+</div>
+This means that it has to be repeated if you want to use different instantiations.
```
ojwb, line 2347: Seems to be missing a `<p>`...`</p>` block.

ojwb:
> Could you please use the SWIG coding style wrt braces and ref placement. Am also confused as you mention instantiating the template but their is no instantiation!

IIUC the point to make is that you need to say `A<int>` not `A<T>` or just `A`.  As you note that isn't really "instantiating".  I couldn't think what the correct term was, but looking at https://en.cppreference.com/w/cpp/language/templates it seems to be "template-id".  Not sure how many people will understand that (I wouldn't have until I went and looked) especially as the audience probably include people who aren't really C++ coders.  I guess we could say something like "template-id (i.e. `A<int>` not `A<T>` or just `A`)" to clarify what's meant for people who don't know the term.

ojwb:
> Apologies for the multiple commits. Not a good idea to try to do this via GitHub's edit feature... Please squash-merge if it's fine.

Yeah, Github's edit feature is really better avoided except for really simple stuff like trivial typos in docs.  Not a big problem though - can squash to merge once it's ready.

And thanks for following through with a patch.

KrisThielemans:
Hopefully ok now. let me know.

PR #2226, opened 2022-03-08
Fix checking of "optimal" typemap attribute

ojwb:
Previously SWIG checked that the typemap action contained ";\n" not
followed by an identifier character, and that it contained no other
`;`, but that incorrectly allows some cases it shouldn't.

Instead check that the action ends with `;\n` and contains no other
`;`, which is simpler and correctly rejects these cases.

### DIFF
--- a/Examples/test-suite/errors/cpp_typemap_out_optimal_bug.i
+++ b/Examples/test-suite/errors/cpp_typemap_out_optimal_bug.i
@@ -1,0 +1,45 @@
+%module x
+
+// Just the following languages tested
+#if defined (SWIGCSHARP) || defined (SWIGD)
+%typemap(out, optimal="1") SWIGTYPE %{
+  $result = new $1_ltype((const $1_ltype &)$1);
+%}
+#elif defined (SWIGJAVA)
+%typemap(out, optimal="1") SWIGTYPE %{
+  *($&1_ltype*)&$result = new $1_ltype((const $1_ltype &)$1);
+%}
+#elif defined (SWIGUTL)
+%typemap(out,noblock="1", optimal="1") SWIGTYPE {
+  %set_output(SWIG_NewPointerObj(%new_copy($1, $ltype), $&descriptor, SWIG_POINTER_OWN | %newpointer_flags));
+}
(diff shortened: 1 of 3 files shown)


ojwb:
@wsfulton Please take a look at this when you get a chance to make sure I'm not missing something.

I spotted this from looking at the code - I was looked at fixing the `DOH_REPLACE_NOQUOTE` bug and noticed that the `DOH_REPLACE_ID_BEGIN`, etc options skipped over the length of the "needle" being searched for, which goes wrong if another match could start within the needle.  It turns out that doesn't matter for any existing uses (not sure if this limitation/feature/bug should be addressed, documented or ignored though).

Anyway, while looking at the existing uses, I saw the code here could misbehave for certain inputs and managed to construct one which did while being valid (if peculiar).

wsfulton (approved):
Looks okay to me. 

One thing about the code style, make beautify will modify this code as the line lengths are shorter than 160. It would be easier on the eye to use nested if statements instead of one if statement with multiple &&.

Regarding the bug you've identified, by all means fix it if you like even if we don't suffer it curently. A comment in the code would be a useful minimum if you don't have the time/inspiration.

ojwb:
> One thing about the code style, make beautify will modify this code as the line lengths are shorter than 160. It would be easier on the eye to use nested if statements instead of one if statement with multiple &&.

Three nested if-s wouldn't work well here, because we want a single `else` if any of them aren't true.  I don't really see how to address that without the code ending up less readable overall.

I think it's mostly less readable in the diff view here because github insists on treating tab as 4 columns (which I complained about years ago and they eventually fixed but seem to have broken again).

I'll look if the replace bug/feature is easy to fix without risk of introducing bugs, and if not document it as a limitation.

ojwb:
I've fixed the replace bug/feature in a0e3825de696231065a9c92cc218a843220597d9.

Potentially we could optimise by checking up front if/how `token` self-overlaps (kind of like Boyer-Moore) but I tend to think longer term we'd be better off just using PCRE for such tasks rather than trying to reinvent the pattern matching wheel.

PR #2227, opened 2022-03-08
Avoid problems due to the use of "register" in Python 2.7 header

vadz:
Compiling SWIG-generated code using Python 2.7 in C++17 mode, which is
the default for recent compilers, such as gcc 11, results in a bunch of
annoying warnings like this:
```
/usr/include/python2.7/unicodeobject.h:534:24: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  534 |     register PyObject *obj,     /* Object */
      |                        ^~~
/usr/include/python2.7/unicodeobject.h:553:24: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  553 |     register PyObject *obj      /* Object */
      |                        ^~~
/usr/include/python2.7/unicodeobject.h:575:29: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  575 |     register const wchar_t *w,  /* wchar_t buffer */
      |                             ^
/usr/include/python2.7/unicodeobject.h:593:23: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  593 |     register wchar_t *w,        /* wchar_t buffer */
      |                       ^
In file included from /usr/include/python2.7/Python.h:97,
                 from ../../interface/swig/build/python/common_wrap.cxx:188:
/usr/include/python2.7/stringobject.h:173:24: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  173 |     register PyObject *obj,     /* string or Unicode object */
      |                        ^~~
/usr/include/python2.7/stringobject.h:174:21: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  174 |     register char **s,          /* pointer to buffer variable */
      |                     ^
/usr/include/python2.7/stringobject.h:175:26: warning: ISO C++17 does not allow ‘register’ storage class specifier [-Wregister]
  175 |     register Py_ssize_t *len    /* pointer to length variable or NULL
      |                          ^~~
```
while using clang with -std=c++17 (or later) results in errors and not
even warnings.

Avoid them by simply defining "register" keyword as nothing, which is
enough to avoid the warnings/errors and shouldn't create any new
problems as SWIG-generated code doesn't use this keyword and any
user-defined code shouldn't use it neither when using C++17.

### DIFF
--- a/Lib/python/pyruntime.swg
+++ b/Lib/python/pyruntime.swg
@@ -11,6 +11,15 @@
 # include <corecrt.h>
 #endif
 
+#if defined(__cplusplus) && __cplusplus >= 201703L
+/* Avoid using "register" keyword removed in C++17 in Python 2.7 headers. */
+# define register
+/* Additionally disable checks for redefined keywords when using MSVC. */
+# if defined(_MSC_VER)
+#  define _ALLOW_KEYWORD_MACROS
+# endif
+#endif
+
 #if defined(_DEBUG) && defined(SWIG_PYTHON_INTERPRETER_NO_DEBUG)
 /* Use debug wrappers with the Python release dll */
 # undef _DEBUG


wsfulton:
-1 from me. 
- register has not been removed from C++, it is reserved for future usage and so this patch as it stands has the potential to break future usage of the register keyword as it applies to all versions of C++ >= 17.
- Even if a user is using register, SWIG [is not a police enforcement operation](http://swig.org/Doc4.0/Introduction.html#Introduction_nn13) and aims not to break user's code (caveat, does not always succeed!) and SWIG ought not to suppress warnings about user's code.
- Supported versions of Python have this fixed.
- There are existing mechanisms for users who need to deal with this such as
  - suppressing the warnings from the commandline:
  - using `-Dregister=`
  - using %begin to add in equivalent to this patch.

@vadz, I appreciate they are annoying, they annoy me too as I have also seen them, but I feel the 'right thing' in your unusual case of mixing old and new technologies is to suppress the warnings in your own interface files.

ojwb:
Another option would be to directly disable the warning with:

```
#pragma GCC diagnostic ignored "-Wregister"
```

Could be only done for `#include <Python.h>` via pushing and popping the diagnostics around that.

vadz:
Thanks Olly, this is indeed a good/better idea, I don't know why I hadn't done it like this.

I can update the PR to push/pop this diagnostic before/after the header inclusion if there is any interest in it.

ojwb:
It seems reasonable to me to try to address this if it can be done cleanly and without risk of fallout.  It also seems sensible to avoid putting effort into Python2 support at this point, but overall addressing this seems just on the OK side of the "keep it working until we can drop it" line.

It could perhaps also be only done `'#ifndef SWIGPYTHON_PY3` since that's defined for SWIG's preprocessor if `-py3` is passed which means if `Python.h` is from Python 2 we'll `#error` after including it anyway.  Hopefully `-py3` should be fully decoupled from enabling optional features in SWIG 4.1.0 so should be increasingly used by Python3-only projects.

wsfulton:
A push/pop around Python.h removes the downsides, so gets my support.

Just please make sure the pragma is only parsed by versions of gcc that understand this pragma and understands the pragma for "-Wregister". I have seen this kind of warning suppression end up breaking builds / causing a build failure.

PR #2230, opened 2022-03-13
[Scilab] Functions must exported in dynamic link library on Windows 

YungLee:
This commit fix two bugs for swig/scilab on Windows :

    Functions must exported in dynamic link library
    builder.sce script file not generated if using '-builder' option


ojwb:
Thanks, but I'm going to just close this in favour of the existing #1853.  Cleaning up the history was a nice thought, but this PR already has 3 merges so it doesn't really achieve that.  It's also very easy for us to squash a PR to a single commit as we merge (there's even a button for it in the github UI).

And all the existing discussion is in #1853.

PR #2233, opened 2022-03-15
[Go] Make -intgosize optional

### LINKED ISSUE #683: Default -intgosize to sizeof(void*)
If we remove the `intgosize` parameter test in `go.cxx`, cgo will complain about not finding `ptrdiff_t` and `size_t`:

```
error: unknown type name 'ptrdiff_t'
 typedef ptrdiff_t intgo;
error: unknown type name 'size_t'
 typedef size_t uintgo;
```

So `stddef.h` should be added into cgo_comment section when it is decided to support default intgosize.

ojwb:
We no longer support Go < 1.2 so we can default intgo to ptrdiff_t
and uintgo to size_t.

Fixes #683

### DIFF
--- a/Doc/Manual/Go.html
+++ b/Doc/Manual/Go.html
@@ -178,10 +178,10 @@ swig -go -help
 <td>-intgosize &lt;s&gt;</td>
 <td>Set the size for the Go type <tt>int</tt>.  This controls the size
   that the C/C++ code expects to see.  The &lt;s&gt; argument should
-  be 32 or 64.  This option is currently required during the
+  be 32 or 64.  This option was required during the
   transition from Go 1.0 to Go 1.1, as the size of <tt>int</tt> on
-  64-bit x86 systems changes between those releases (from 32 bits to
-  64 bits).  In the future the option may become optional, and SWIG
+  64-bit x86 systems changed between those releases (from 32 bits to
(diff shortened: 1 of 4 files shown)


ojwb:
The CI has passed for Go 1.6, 1.8, 1.12 and 1.17.

I left one explicit `-intgosize 32` in `configure.ac` for gccgo 4.7.x since that seemed to be hardwired to 32 whereas later gccgo tested `sizeof(void*)`.  I'm not sure if that means gccgo 4.7.x corresponds to Go 1.0.x, or if it was 32-bit only, or what.  Some input on whether that case can be removed too would be helpful.  If it can't, maybe the gccgo minimum version should just be raised if gccgo 4.7.x is no longer relevant.

ojwb:
Ian acked in #683 so I've pushed the change here with a `CHANGES.current` entry:
> Your pull request looks good to me. Thanks.

KuoKongQingYun:
https://github.com/swig/swig/issues/2348
I have the same problem, I want to remove -intgosize.
I have another train of thought that is using golang's "Build constraints", but I don't have the ability to modify the source of swig.

ojwb:
@KuoKongQingYun This will be fixed in the next SWIG release - meanwhile SWIG git master has the change.

PR #2235, opened 2022-03-17
Default to running tests with Python 3

ojwb:
Specify PY2=1 to use Python 2.

See #1779

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
 
     # By default, the name of the build is the language used and SWIG options, but matrix entries
     # can define the additional "desc" field with any additional information to include in the name.
-    name: ${{ matrix.SWIGLANG || 'none' }}${{ matrix.PY3 }} ${{ matrix.ENGINE}} ${{ matrix.VER }} ${{ matrix.SWIG_FEATURES }} ${{ (matrix.compiler || 'gcc') }}${{ matrix.GCC }} ${{ matrix.CPPSTD }} ${{ matrix.CSTD }} ${{ matrix.desc }} ${{ matrix.continue-on-error && '(can fail)' }}
+    name: ${{ matrix.SWIGLANG || 'none' }}${{ matrix.PY2 }} ${{ matrix.ENGINE}} ${{ matrix.VER }} ${{ matrix.SWIG_FEATURES }} ${{ (matrix.compiler || 'gcc') }}${{ matrix.GCC }} ${{ matrix.CPPSTD }} ${{ matrix.CSTD }} ${{ matrix.desc }} ${{ matrix.continue-on-error && '(can fail)' }}
 
     strategy:
       matrix:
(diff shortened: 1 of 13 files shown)


wsfulton:
```
~/swig/github/swig (master=)$ ./configure --help  | grep python
  --without-python        Disable Python
  --with-python=path      Set location of Python executable
  --without-python3       Disable Python 3.x support
  --with-python3=path     Set location of Python 3.x executable
```
This feels incomplete as the default python is 2 here instead of 3. Is there a plan here? 

The Appveyor checks for the pull request show a failure and so after the merge to master the same failures appear. I think it is something to do with --with-java not working so in the same area. I've incidentally made numerous changes to appveyor.yml in parallel and merged to master, but they are all green testing in my own repo.

To be honest, I don't understand the change on this line and may be related:
https://github.com/swig/swig/blob/06364ae74981795abe35f85e0acd8f37e055be90/appveyor.yml#L162

ojwb:
The way configure works is that it probes for both Python2 and Python3 separately - each can be found or not found separately.  There's not really a "default" there.

Before my change, the testsuite tried to use Python2 (whether it was found or not) unless you passed `PY3=1` in which case it tried to use Python3 (whether it was found or not).  That's where there was a default which I've changed.

Are you suggesting we should change to `--without-python` and to be for Python 3?  That seems needlessly disruptive as these options are likely used in packaging scripts.  I'm pushing a change so configure --help and other messages say "Python 2.x" rather than just "Python" which should avoid any potential confusion here as now you get:

```
$ ./configure --help|grep python
  --without-python        Don't probe for Python 2.x
  --with-python=path      Set location of Python 2.x executable
  --without-python3       Don't probe for Python 3.x
  --with-python3=path     Set location of Python 3.x executable
```

I actually checked the appveyor failures before merging and ignored them as they were clearly entirely unrelated to my change:

```
+ aclocal -I Tools/config
Can't locate threads.pm in @INC (you may need to install the threads module) (@INC contains: /usr/share/automake-1.16 /usr/local/lib/perl5/site_perl/5.32/i686-cygwin-threads-64int /usr/local/share/perl5/site_perl/5.32 /usr/lib/perl5/vendor_perl/5.32/i686-cygwin-threads-64int /usr/share/perl5/vendor_perl/5.32 /usr/lib/perl5/5.32/i686-cygwin-threads-64int /usr/share/perl5/5.32) at /usr/share/automake-1.16/Automake/ChannelDefs.pm line 23.
BEGIN failed--compilation aborted at /usr/share/automake-1.16/Automake/ChannelDefs.pm line 26.
Compilation failed in require at /usr/share/automake-1.16/Automake/Configure_ac.pm line 27.
BEGIN failed--compilation aborted at /usr/share/automake-1.16/Automake/Configure_ac.pm line 27.
Compilation failed in require at /usr/bin/aclocal-1.16 line 38.
BEGIN failed--compilation aborted at /usr/bin/aclocal-1.16 line 38.
cat: config.log: No such file or directory
```

That's a problem with the installed Perl and/or automake which we've been getting them sporadically recently, e.g. same failure here before my change: https://ci.appveyor.com/project/swig/swig/builds/42911976/job/qaraud7p88smh0p2

I assume some of the appveyor build VMs have a bad cygwin configuration so it's pot luck whether that job works or not.

Appveyor seems quite unreliable and is also very slow to complete because jobs aren't run in parallel.  Perhaps we should migrate to using GHA for all the CI builds?

> To be honest, I don't understand the change on this line and may be related:

This part was the simplest way I could see to set `PY3=3` if-and-only-if `PY2` wasn't set to `2` (it's set to `2` further up for Python2 builds):

`x=${PY2:-x3} && PY3=${x:1)`

The first part sets `x` to either `2` (Python 2) or `x3` (Python 3).  The second part sets `PY3` to the substring of x ignoring the first character, so empty (Python 2) or `3` (Python 3).

There may be a simpler way to do this, but I don't really know DOS command file syntax, so I wanted something that I could pull off entirely inside the `bash` invocation.  Feel free to improve if you think you can do better.

wsfulton:
https://ci.appveyor.com/project/swig/swig/builds/42925018 shows that the mingw+java breaks. This passed before the merge for python3. Sure, the cygwin build has been red for a while, but it was on the "allow_failures" section so it didn't break the build overall. The problem is every pull request has failed since this patch was merged in. The problem is now for the java build, it shows:
```
configure: Disabling Java
```
and I'm sure it has something to do with the appveyor.yml line I highlighted. Please prioritise a fix for this.

wsfulton:
So how does the test-suite work now wrt to the -py3 option? I havn't finished sorting -py3 out yet. To be honest, I would have been inclined to minimise disruption and leave everything exactly as is until -py3 was sorted and maybe even leave the PY3 env variable untouched until we dropped Python 2 altogether.

The 'default' for swig -python is still python2 right? So if the default way of running the test-suite is now python3, I'm confused. Maybe this change needs better documenting or I just need to figure what what's changed :)

ojwb:
Oops, wrong ref to this ticket on the first of those two.

I believe the issue here was simply that PY3 was set to 3 regardless of the language, so this was trying to configure `--with-java3`.  I've redone the test using batch file conditionals so `PY3=3` is only set for Python, though I can't test that locally at all.

The handling of `-py3` is that it's passed to swig for python3 is used and not for python2, just like before.

```
$ make -C Examples/test-suite/python director_basic.cpptest 2>/dev/null | grep 'env SWIG_LIB'
env SWIG_LIB=../../../Lib  ../../../swig -python -py3 -c++ -outcurrentdir -I../../../Examples/test-suite -o director_basic_wrap.cxx ./../director_basic.i
$ make PY2=2 -C Examples/test-suite/python director_basic.cpptest 2>/dev/null | grep 'env SWIG_LIB'
env SWIG_LIB=../../../Lib  ../../../swig -python  -c++ -outcurrentdir -I../../../Examples/test-suite -o director_basic_wrap.cxx ./../director_basic.i
```

PR #2237, opened 2022-03-17
Improve type handling for long and long long in csharp

igt-gteubl:
This pull request aims to improve the handling of the `long` and `long long` datatypes (and their variants).
It is related to #568 and #650 and is based on this pull request (which has not been merged yet): https://github.com/swig/swig/pull/650/files

This pull request improves upon the base in such a way, that typemaps for the `long long` datatype (and its variants) are also generated.

Additionally, it fixes a bug in the generated code, that leads to compile errors in certain circumstances:

When `SWIGWORDSIZE64` is set, there is a difference in how `int64_t` is defined between compilers (tested on Linux GCC and MacOS CLANG):
* `int64_t` is `long long` in MacOS
* `int64_t` is `long` in Linux (GCC)

Since generated SWIG code uses the basic types of `long*` and `long long*` to try the casting to `int64_t*`, the code is either not compatible with Linux or MacOS, since `long long*` cannot be cast to `int64_t*` in Linux and vice versa.

This commit introduces a new definition `SWIGINT64ISLONGLONG` which users can set in addition to `SWIGWORDSIZE64` in order to signal that the compiler in use will treat `int64_t` as `long long`.


emmenlau:
This seems very sensible! Anything I can do to move this forward?

emmenlau:
Dear @igt-gteubl and all, thanks for this very good work that is very important for csharp in modern 64bit environments.

I found two small shortcomings in this PR, and therefore continued the work in #2379. Please feel free to pull my changes into this PR and I will happily close mine! I took good care to preserve your commits as original as possible.

Shortcomings in this PR:
 * The 64bit handling does not yet work for `size_t` and `ptrdiff_t`.
 * This PR uses 5 separate commits that build on top of each other. To better understand and review the changes, it's slightly easier to squash them into a single commit.

Both points are addressed in #2379, so only one of the two PRs should be merged.

igt-gteubl:
Hi @emmenlau, 
thanks for building upon my pull request. Maybe together we can get it to be merged in the forseeable future.

I don't really have anything more to contribute on my side (since that's actually the code we're currently using), so if you're ok with it, I'd close this PR and hand over the issue to you.

Is that ok for you?

emmenlau:
Dear @igt-gteubl , yes this is good for me! Thanks a lot for your great work, and lets see if we can get this over the finish line in #2379!

igt-gteubl:
Superseded by #2379.

Issue #1684, opened 2019-12-18
python - swig_runtime_data conflict between builtin and non-builtin modules


eltoder:
It appears that SWIG types built with -builtin and without -builtin share the same swig_runtime_data capsule, because the flag is not encoded in the name. However, the types are not compatible. For example, when using -builtin, SwigPyObject is defined as

```
typedef struct {
  PyObject_HEAD
  void *ptr;
  swig_type_info *ty;
  int own;
  PyObject *next;
#ifdef SWIGPYTHON_BUILTIN
  PyObject *dict;
#endif
} SwigPyObject;
```

and tp_dealloc starts with

```
    SwigPyObject *sobj;                         \
    sobj = (SwigPyObject *)a;                   \
    Py_XDECREF(sobj->dict);                     \
```

If a module using swig without -builtin uses the same type, its type info can be overwritten with this one, which will crash on the Py_XDECREF line.

I suspect there are more incompatibilities between the two modes, so they should use separate capsules.

vadz:
This does look like a problem, thanks. I'm not sure how do you end up with the same types in "builtin" and not "builtin" versions, but it would be better to fix this, of course.

AFAICS it should be as simple as adding `#ifdef SWIGPYTHON_BUILTIN/#else/#endif` around `SWIGPY_CAPSULE_NAME` definition in `Lib/python/pyhead.svg`, but I didn't test it and don't have a test case for it, so it would be nice if you could please make the change, check that it works and make a PR if it does. TIA!

eltoder:
In our case, it was an instantiation of std::vector<double> that was used both by our module and cvxpy installed from pipy.

Yes, I think it should be simple. I'll try to make a PR.

ojwb:
@eltoder Did you create a patch?  I can't seem to see a PR from you.

ojwb:
Thanks for the quick response.

I certainly think something should be done about this - it's bad that two different Python modules created with SWIG can conflict behind the scenes like this just because one was built with `-builtin`.

[...]

PR #2238, opened 2022-03-18
Use different capsule names with and without -builtin

eltoder:
Types generated with and without -builtin are not compatible. Mixing
them in a common type list leads to crashes. Avoid this by using
different capsule names: "type_pointer_capsule" without -builtin and
"type_pointer_capsule_builtin" with.

See #1684

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -847,9 +847,10 @@ swig_and_compile_c =  \
 
 swig_and_compile_multi_cpp = \
 	for f in `cat $(top_srcdir)/$(EXAMPLES)/$(TEST_SUITE)/$*.list` ; do \
+	  swigopt=$$(name=SWIGOPT_$$f; eval echo \$$$$name); \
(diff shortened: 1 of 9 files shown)


wsfulton:
Thanks for the patch @eltoder. Kudos for squeezing this into the test-suite and forcing/overriding when -builtin is used! I would have said the test framework was not designed for this type of flexibility. Although clever, the implementation for the test is not what I'd call particularly maintainable nor easy to follow, which I think is important, so I've put in an alternative, which is not ideal either, but I feel is at least more maintainable. See f2dd436a5b6c71d1c552a9238a33980f2f0bb529 which enabled another improvement in c5f209706ed411099ae128a49a9c1c987e789d0d.

eltoder:
Thank you for merging @wsfulton. Yes, the test suite did not cooperate, but it felt bad to fix something very subtle without any tests. I did not want to spend a lot of time refactoring, since I wasn't sure if there was any interest in merging. I considered an approach similar to f2dd436a5b6c71d1c552a9238a33980f2f0bb529. My approach seemed more general and not too messy, until I realized that I'm limited to sh syntax and cannot use bash's indirect variables. I'm fine with either option. Thank you for polishing the test, hopefully this will not regress in the future.

PR #2239, opened 2022-03-18
C#: Fix CS0034 in Capacity method of std::vector

igt-gteubl:
On platforms where `size_t` maps to the `ulong` type in C#, the generated code for the `Capacity` setter of `std::vector<T>` typemaps generates:

`error CS0034: Operator '<' is ambiguous on operands of type 'int' and 'ulong'`

This relates to this discussion in the dotnet forum: https://github.com/dotnet/csharplang/discussions/2322

To fix this issue, a two-step comparison and a cast was introduced.

### DIFF
--- a/Lib/csharp/std_vector.i
+++ b/Lib/csharp/std_vector.i
@@ -63,7 +63,7 @@
       return (int)capacity();
     }
     set {
-      if (value < size())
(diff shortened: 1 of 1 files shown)


wsfulton:
Looks reasonable to me, thanks.

PR #2241, opened 2022-03-19
[Go] Add missing newlines for goin typemaps

cminyard:
In some cases the \n was missing from goin directives in some places.
Without this change, you will get

}	swig_r = (int)(C....

in the output in some cases, and that won't compile in Go, it will
error with:

  ../gensio/gensio.go:1508:3: expected ';', found swig_r

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,17 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.1.0 (in progress)
 ===========================
 
+2022-04-20: cminyard
+	    Fix an issue where newlines were not properly generated
+	    for godirectorin typemaps.  If you have a virtual function
+	    not assigned to zero, in some cases it won't generate a
+	    newline and you will see errors:
+	      example.go:1508:3: expected ';', found swig_r
+	    when compiling the go code.
+
+	    Also add an example of using goin and godirectorin and add
+	    a test for this situation.
+
 2022-04-11: robinst
 	    #2257 Fix new Ruby 3.2 warning "undefining the allocator of T_DATA
 	    class swig_runtime_data".
--- a/Examples/go/check.list
+++ b/Examples/go/check.list
@@ -12,3 +12,4 @@ reference
 simple
 template
 variables
+goin
--- a/Examples/go/goin/Makefile
+++ b/Examples/go/goin/Makefile
@@ -1,0 +1,18 @@
+TOP        = ../..
+SWIGEXE    = $(TOP)/../swig
+SWIG_LIB_DIR = $(TOP)/../$(TOP_BUILDDIR_TO_TOP_SRCDIR)Lib
+CXXSRCS    =
+TARGET     = example
+INTERFACE  = example.i
+SWIGOPT    =
+
+check: build
+	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' CXXSRCS='$(CXXSRCS)' TARGET='$(TARGET)' INTERFACE='$(INTERFACE)' go_run
+
+build:
+	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' CXXSRCS='$(CXXSRCS)' \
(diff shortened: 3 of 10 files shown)


ojwb:
Can you add a regression test for this?

cminyard:
Well, this was frustrating because it took me a while to figure out exactly why it was failing.

It turns out if you set a virtual function to 0, like "virtual bool asdf(xxx) = 0", it would not fail.  It would only fail if you didn't set it to zero and provided the method.

Anyway, I have an example now that fails to compile without the patch, and I updated the patch header with the new info.  Hopefully the example is useful, there weren't any for goin or godirectorin.

cminyard:
I had kind of forgotten about this, but I looked at it today and fixed the issues, it is now passing all the tests.  Is it possible to get this looked at soon?  I'm going to need it for my software.  Thanks.

wsfulton:
Up to @ianlancetaylor to accept this. While the example might show it working, the correct place for testing this is in the test-suite as the examples are kept to a minimal set of useful user examples.

cminyard:
On Sun, Apr 17, 2022 at 10:04:07AM -0700, William S Fulton wrote:
> Up to @ianlancetaylor to accept this. While the example might show it working, the correct place for testing this is in the test-suite as the examples are kept to a minimal set of useful user examples.

I thought about that, but I put it in examples because there was no
example of this at all.  I guess I'll let @ianlancetaylor say where he
wants this.

ianlancetaylor:
The example seems OK but there also ought to be a test.  Perhaps you can add another case to Examples/test-suite/go_director_inout.i.

ianlancetaylor:
@wsfulton This change looks fine to me.  I'm not sure what the usual next step is.  Should I just click on "Merge pull request"?  Should I do "Create a merge commit" or "Squash and merge"?  Thanks.

cminyard:
Ok, I added a test that exercises this bug.  It's basically the same thing as the example, but the values being passed in are checked.  Since this is a code generation bug, it really doesn't show anything new verses the example, but it's probably good to have other things about this code checked, since it's a fairly complicated example.

I tried to generate a simpler reproducer, but I wasn't successful.  It needs to be something of the nature of the example.  I did leave in part of that, having some diversity in the example (having a default handler in one case) covers a little more territory than assigning all the virtual functions to zero.

ojwb:
> This change looks fine to me. I'm not sure what the usual next step is. Should I just click on "Merge pull request"? Should I do "Create a merge commit" or "Squash and merge"?

@ianlancetaylor It's basically a judgement call.

For a single commit PR, it's probably best overall to rebase it onto master to merge it (if it rebases cleanly).  It's lying about the branch history but it keeps master's history easier to follow.

For a case like this, the branch has a clean history which splits the change into 3 commits which seems fairly logical sub-parts so "Create a merge commit" is probably the best choice.

If the branch history is a complete mess (sometimes there are seemingly arbitrary combinations of changes in different commits or multiple merges from master) then squashing is probably better.

User-visible changes (like this) should generally get an entry added to `CHANGES.current` - you might find it simplest to ignore the github UI for this and instead pull the PR branch locally, add a commit to it to update `CHANGES.current` and then merge that local branch to master.  Pushing master should then auto-close the PR as github can see that the commits in the branch have been merged to master.

However, if you merge by squashing (or rebasing) locally, you'll need to add `Closes #nnnn` to the commit message, or else remember to close the PR manually.

ianlancetaylor:
@cminyard Would you like to add an entry to CHANGES.current to this PR?  Thanks.

(I'm the kind of person who uses git by rote scripts, so I'm not going to try to pulling the PR branch locally.)

cminyard:
On Wed, Apr 20, 2022 at 03:37:58PM -0700, Ian Lance Taylor wrote:
> @cminyard Would you like to add an entry to CHANGES.current to this PR?  Thanks.

Ok, I added an entry in CHANGES.current and pushed it up.

> 
> (I'm the kind of person who uses git by rote scripts, so I'm not going to try to pulling the PR branch locally.)

It took me a while to wrap my head around git, coming from someone who
did a bunch of development on CVS itself.  Now I'm a git expert, and I
just needed to use CVS for something, and it took me a while to figure
it out :).

-corey

ianlancetaylor:
Thanks.

PR #2246, opened 2022-03-23
add all template parameters to java unordered set

charles-huet-ixxi:
When trying to use a `%template(MyEnumSet) std::unordered_set<MyEnum,std::hash<std::underlying_type<MyEnum>::type>>;`, swig errors out with `Too many template parameters. Maximum of 1.`.

This fixes this very specific issue.

Should I do this also for other types such as `std::unordered_map` ?
Should I do this also for other languages if necessary ?

### DIFF
--- a/Lib/java/std_unordered_set.i
+++ b/Lib/java/std_unordered_set.i
@@ -43,7 +43,10 @@
 
 namespace std {
 
-template <class Key>
+template <class Key,
+	   typename _Hash = std::hash<Key>,
+	   typename _Pred = std::equal_to<Key>,
+	   typename _Alloc = std::allocator<Key>>
 class unordered_set {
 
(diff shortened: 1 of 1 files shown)


ojwb:
Yes, it'd be good to fix any other similar omissions for Java and other languages.

At least for `std::unordered_set` it seems other languages which support it use a common version in `Lib/std/std_unordered_set.i` which already has these optional parameters:

```
  template <class _Key,
            class _Hash = std::hash< _Key >,
            class _Compare = std::equal_to< _Key >,
            class _Alloc = allocator< _Key > >
```

It'd be good to have testcases which check these parameters are present too.

Just adding the parameters is probably enough for simple cases, but for full support they should actually be used too - looking at the `Lib/std/std_unordered_set.i` version, there's at least these currently missing from the Java-specific version:

```
    typedef _Hash hasher;
    typedef _Compare key_equal;
    typedef _Alloc allocator_type;
```

PR #2249, opened 2022-03-31
Update Windows CMake builds to use NuGet package for PCRE2

geographika:
Following on from #2178 and #2169 this updates the CMake Windows build instructions to use a newly created PCRE2 NuGet package, which simplifies the build steps.

References: 

+ https://www.nuget.org/packages/PCRE2/
+ https://github.com/geographika/nuget-pcre2

Let me know if it would be useful to add a new GitHub Action to test these builds steps as part of CI.

### DIFF
--- a/.github/workflows/nuget.yml
+++ b/.github/workflows/nuget.yml
@@ -44,4 +44,4 @@ jobs:
       shell: cmd
       working-directory: install2/bin
       run: |
-          swig.exe -help
+          swig.exe -version
--- a/.github/workflows/nuget.yml
+++ b/.github/workflows/nuget.yml
@@ -25,23 +25,30 @@ jobs:
         submodules: recursive
 
     - name: Install Dependencies
+      shell: powershell
       run: |
           nuget install CMake-win64 -Version 3.15.5 -OutputDirectory C:\Tools\CMake
           nuget install Bison -Version 3.7.4 -OutputDirectory C:\Tools\bison
           nuget install PCRE2 -Version 10.39 -OutputDirectory C:\Tools\pcre2
 
     - name: Build
-      shell: cmd
+      shell: powershell
       run: |
-          SET PATH=C:\Tools\CMake\CMake-win64.3.15.5\bin;C:\Tools\bison\Bison.3.7.4\bin;%PATH%
-          SET PCRE_ROOT=C:\Tools\pcre2\PCRE2.10.39.0
-          SET PCRE_PLATFORM=x64
-          cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX="%CD:\=/%/install2" -DCMAKE_C_FLAGS="/DPCRE2_STATIC" ^
-          -DCMAKE_CXX_FLAGS="/DPCRE2_STATIC" -DPCRE2_INCLUDE_DIR=%PCRE_ROOT%/include -DPCRE2_LIBRARY=%PCRE_ROOT%/lib/pcre2-8-static.lib -S . -B build
+          $env:PATH="C:\Tools\CMake\CMake-win64.3.15.5\bin;C:\Tools\bison\Bison.3.7.4\bin;" + $env:PATH
+          $PCRE_ROOT="C:\Tools\pcre2\PCRE2.10.39.0"
+          $PCRE_PLATFORM="x64"
+          $WORKING_DIR=(Get-Location).ToString() -replace "\\","/"
+          cmake -G "Visual Studio 16 2019" -A "x64" `
+          -DCMAKE_INSTALL_PREFIX="$WORKING_DIR/install2" `
+          -DCMAKE_C_FLAGS="/DPCRE2_STATIC" `
+          -DCMAKE_CXX_FLAGS="/DPCRE2_STATIC" `
+          -DPCRE2_INCLUDE_DIR="$PCRE_ROOT/include" `
+          -DPCRE2_LIBRARY="$PCRE_ROOT/lib/pcre2-8-static.lib" `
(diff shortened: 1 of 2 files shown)


wsfulton:
Good stuff. How about updating appveyor.yml so that the cmake build there only depends on nuget and not cygwin. We could I suppose _contaminate_ Github Actions with Windows builds, but seems neater to keep Windows in Appveyor.

ojwb:
@geographika Did you see wsfulton's suggestion above?

geographika:
@ojwb thanks for the reminder. Happy to take care of this, although it will be in a couple of weeks time. Would it make sense to have it as a separate pull request, or should I update this branch?

ojwb:
Great.

Probably better to just update this branch.  Feel free to rebase and force push.

geographika:
@ojwb @wsfulton - the current Appveyor build tests different languages as well as Cygwin, MinGW, x86 and x64. As it is all running and building successfully, I think this should be left as-is. There are also still docs on how to build using Cygwin, MSYS2, and MinGW. 

The docs update in this pull request provides an alternate approach using nuget packages (which personally I found easier), but packages are only available for x64 Windows.

I've added in a GitHub Action, separate from the main `ci.yml` to test this approach, and the steps in the documentation. I tried integrating into Appveyor, but this was tricky with the different matrices, and Appveyor only allows one .yml file (as I understand it), so it could not be added separately.

.github/workflows/nuget.yml
```diff
@@ -0,0 +1,47 @@
...
+        submodules: recursive
+
+    - name: Install Dependencies
+      run: |
+          nuget install CMake-win64 -Version 3.15.5 -OutputDirectory C:\Tools\CMake
+          nuget install Bison -Version 3.7.4 -OutputDirectory C:\Tools\bison
+          nuget install PCRE2 -Version 10.39 -OutputDirectory C:\Tools\pcre2
+
+    - name: Build
+      shell: cmd
```
wsfulton, line 34: I think it's time to move off cmd, it is so horrible, onto the less horrible powershell. I see you used powershell above, so let's just use one shell.

.github/workflows/nuget.yml
```diff
@@ -0,0 +1,47 @@
...
+      run: |
+          SET PATH=C:\Tools\CMake\CMake-win64.3.15.5\bin;C:\Tools\bison\Bison.3.7.4\bin;%PATH%
+          SET PCRE_ROOT=C:\Tools\pcre2\PCRE2.10.39.0
+          SET PCRE_PLATFORM=x64
+          cmake -G "Visual Studio 16 2019" -A x64 -DCMAKE_INSTALL_PREFIX="%CD:\=/%/install2" -DCMAKE_C_FLAGS="/DPCRE2_STATIC" ^
+          -DCMAKE_CXX_FLAGS="/DPCRE2_STATIC" -DPCRE2_INCLUDE_DIR=%PCRE_ROOT%/include -DPCRE2_LIBRARY=%PCRE_ROOT%/lib/pcre2-8-static.lib -S . -B build
+          cmake --build build --config Release --target install
```
wsfulton, line 41: Best put into an Install section

wsfulton (commented):
I'm still digesting this aspect of GHA - I didn't know it was possible to have a second yml file...

https://github.com/swig/swig/actions/runs/2895000855 shows this build taking 1h27. That can't be right?? If these Windows Nuget builds take that long to complete a 2 minute job, then I'm a bit worried that will slow down overall CI. Perhaps it was a bad time of day, have you any other experience of windows on GHA?

.github/workflows/nuget.yml
geographika: Will do

wsfulton:
Hopefully, let's see when you push those suggested changes.

wsfulton:
Ping @geographika, are you going to make those suggested changes?

geographika:
@wsfulton - not forgotten! Will finalise this next week. I wasn't sure though if the docs should also be updated to use PowerShell as well as the GitHub Action. I presume so, as the action is testing the approach in the docs.

wsfulton:
Yeah, good suggestion, makes sense to modernise the docs too.

geographika:
@wsfulton - the Windows docs and GitHubAction both updated to use PowerShell - the GHA tests the docs steps run successfully. 

The GHA runs successfully in 2mins 11 seconds (https://github.com/swig/swig/actions/runs/3134293967/jobs/5088662780). The failing tests are unrelated to this pull request.

.github/workflows/nuget.yml
wsfulton: Could you separate out the Build and Install into separate sections, otherwise looking good.

geographika: I tried a [commit ](https://github.com/swig/swig/pull/2249/commits/730e49d379cad28dcbafe63c1d51b1a9bc68b9ee)with that, but the line `cmake --build build --config Release --target install` builds and installs with the same command. I can see if I can break in to 2 cmake commands?

vadz: I'm a complete CMake newbie, but normally `cmake --build` without any `--target` should just build everything. And then `cmake --build xxx -t install` could be used for installing.

geographika: Ok reading more there is a `cmake --install . --config Release` in newer versions of CMake that could be used. Will test now.

geographika:
@wsfulton - commands split into separate build and install steps.

wsfulton:
Looks good to me, thanks. I'll probably change the names of the Github Actions to better categorise what systems they are running on. Useful to know about this additional GHA functionality.

PR #2252, opened 2022-04-05
Remove bogus parser grammar rule for unary &&

ojwb:
This appears to have been accidentally added in
d8cc75946b236f08f9561a82867997e1021f96ab as part of the changes to
support C++11 rvalue references, presumably to match the existing
correct rule for unary &.

There is a non-standard GCC extension for taking the address of a label
using unary &&:

https://gcc.gnu.org/onlinedocs/gcc/Labels-as-Values.html

However this extension only works in an expression in a function body,
and SWIG doesn't parse the contents of function bodies so this grammar
rule isn't even providing any useful support for this extension.

### DIFF
--- a/Source/CParse/parser.y
+++ b/Source/CParse/parser.y
@@ -6743,10 +6743,6 @@ valexpr        : exprsimple { $$ = $1; }
 		 $$ = $2;
                  $$.val = NewStringf("&%s",$2.val);
 	       }
-               | LAND expr {
-		 $$ = $2;
-                 $$.val = NewStringf("&&%s",$2.val);
-	       }
                | STAR expr {
 		 $$ = $2;
                  $$.val = NewStringf("*%s",$2.val);


ojwb:
Two CI failures are unrelated (but I've set them to retry as it seems you can now retry just failed builds).

@wsfulton OK to merge?  The original commit adding this rule was yours.

ojwb:
In case you also wonder, I've tested using this extension as the default value of a C++ function parameter, and both GCC and clang disallow doing so.

GCC:

```
landtest.c:1:21: error: label ‘p’ referenced outside of any function
    1 | void* f(void* x = &&p) {
      |                     ^
```

clang:
```
landtest.c:1:21: error: use of address-of-label extension outside of a function body
void* f(void* x = &&p) {
                    ^
```

wsfulton (approved):
Yup, looks good to me.

PR #2264, opened 2022-04-21
[Python] Replace tp_print with tp_vectorcall_offset slot from Python 3.8

### LINKED ISSUE #2262: PyHeapTypeObject errors with PyPy
(Originally sent to swig-user mailing list)

Our conda-forge feedstock received an automatic PR to add PyPy support. Unfortunately, I get a lot of errors when compiling the SWIG wrapper like this:
```
$SRC_DIR/build/src/swig/CMakeFiles/_stir.dir/stirPYTHON_wrap.cxx:158287:1:
   error: too many initializers for 'PyHeapTypeObject' {aka '_heaptypeobject'}
```
This is with swig 4.0.2 and pypy 3.8. You can find the log at [Pipelines - Run 20220414.1 logs (azure.com)](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=492070&view=logs&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=b2a8456a-fb11-5506-ca32-5ccd32538dc0&l=6141)

I find the following in the wrapped file (EDIT: generated locally by swig 3.0.12)
```
#if PY_VERSION_HEX < 0x02030000
typedef struct {
  PyTypeObject type;
  PyNumberMethods as_number;
  PyMappingMethods as_mapping;
  PySequenceMethods as_sequence;
  PyBufferProcs as_buffer;
  PyObject *name, *slots;
} PyHeapTypeObject;
#endif

/// stuff

static PyHeapTypeObject SwigPyBuiltin__swig__SwigPyIterator_type = {
  {
    PyVarObject_HEAD_INIT(NULL, 0)
#else
    PyObject_HEAD_INIT(NULL)
    0,                                        /* ob_size */
#endif
    "stir.SwigPyIterator",                    /* tp_name */
    sizeof(SwigPyObject),                     /* tp_basicsize */
    0,                                        /* tp_itemsize */
    _wrap_delete_SwigPyIterator_destructor_closure,               /* tp_dealloc */
    (printfunc) 0,                            /* tp_print */
    (getattrfunc) 0,                          /* tp_getattr */
    /// more stuff
```

I cannot find any reference to `PyHeapTypeObject` but wonder if the `#if PY_VERSION_HEX` statements are failing with PyPy.

KrisThielemans:
The `tp_print` slot is no longer supported and replaced with `tp_vectorcall_offset`. The `(printfunc)` cast that we used caused problems on PyPy.

Fixes #2262

### DIFF
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -211,7 +211,11 @@ SwigPyStaticVar_Type(void) {
       sizeof(PyGetSetDescrObject),              /* tp_basicsize */
       0,                                        /* tp_itemsize */
       (destructor)SwigPyStaticVar_dealloc,      /* tp_dealloc */
-      0,                                        /* tp_print */
+#if PY_VERSION_HEX < 0x030800b4
+      (printfunc)0,                             /*tp_print*/
+#else
+      (Py_ssize_t)0,                            /*tp_vectorcall_offset*/
+#endif
       0,                                        /* tp_getattr */
       0,                                        /* tp_setattr */
       0,                                        /* tp_compare */
@@ -295,7 +299,11 @@ SwigPyObjectType(void) {
       PyType_Type.tp_basicsize,                 /* tp_basicsize */
       0,                                        /* tp_itemsize */
       0,                                        /* tp_dealloc */
-      0,                                        /* tp_print */
+#if PY_VERSION_HEX < 0x030800b4
+      (printfunc)0,                             /*tp_print*/
+#else
+      (Py_ssize_t)0,                            /*tp_vectorcall_offset*/
+#endif
       0,                                        /* tp_getattr */
       0,                                        /* tp_setattr */
       0,                                        /* tp_compare */
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -212,9 +212,9 @@ SwigPyStaticVar_Type(void) {
       0,                                        /* tp_itemsize */
       (destructor)SwigPyStaticVar_dealloc,      /* tp_dealloc */
 #if PY_VERSION_HEX < 0x030800b4
-      (printfunc)0,                             /*tp_print*/
+      (printfunc)0,                             /* tp_print */
 #else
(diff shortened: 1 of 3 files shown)


KrisThielemans:
I found there were some other `tp_print` occurrences, which I've left alone.

we really should have a PyPy GitHub action, but I have no time to look into that. sorry.

KrisThielemans:
Failure in the `[ruby 2.6 gcc](https://github.com/swig/swig/runs/6111228195?check_suite_focus=true#logs) job, which has to be unrelated.

Lib/python/builtin.swg
```diff
@@ -211,7 +211,11 @@ SwigPyStaticVar_Type(void) {
       sizeof(PyGetSetDescrObject),              /* tp_basicsize */
       0,                                        /* tp_itemsize */
       (destructor)SwigPyStaticVar_dealloc,      /* tp_dealloc */
-      0,                                        /* tp_print */
+#if PY_VERSION_HEX < 0x030800b4
+      (printfunc)0,                             /*tp_print*/
+#else
+      (Py_ssize_t)0,                            /*tp_vectorcall_offset*/
+#endif
```
ojwb, line 218: These comments (and those below) are inconsistently formatted compared to the existing comments above and below them which all have spaces around the content (which also seems better as it's easier to read).

ojwb:
The "ruby 2.6 gcc" job is flaky on master too, and indeed seems unlikely to be connected.  It's now possible to re-run only failed jobs, so I've tried that (twice now as the first re-run also failed).

I'm very tempted to mark this job as allowed to fail on master.  It really it would be better if someone who knew Ruby fixed it but Ruby 2.6 went EOL earlier this month though so I'm not sure if we're likely to find anyone who cares enough: https://www.ruby-lang.org/en/news/2022/04/12/ruby-2-6-10-released/

Lib/python/builtin.swg
KrisThielemans: should be fixed now

PR #2275, opened 2022-05-06
Add argcargv.i to Lua

erezgeva:
Found it in 
https://stackoverflow.com/questions/9860362/swig-lua-call-mint-argv-char-argc

Did not check it myself.
The author message
"I have looked into Ruby and ported Ruby code to Lua:"


erezgeva:
Test with: lua 5.1, 5.2, 5.3

PR #2276, opened 2022-05-06
Add argcargv.i to more languages: Perl 5, Tcl, PHP

erezgeva:
Add argcargv.i to Perl 5, Tcl 8, PHP 7

Test them on my Debian 11.3 using amd64 

Versions on swig and languages:
- swig 4.0.2-1
- php 7.4.28-1+deb11u1
- perl 5.32.1-4+deb11u2
- tcl 8.6.11+dfsg-1

Testing also Lua argcargv.i from  [Add argcargv.i to Lua #2275](https://github.com/swig/swig/pull/2275)
With: lua 5.1, 5.2, 5.3

### DIFF
--- a/Examples/test-suite/perl5/argcargvtest_runme.pl
+++ b/Examples/test-suite/perl5/argcargvtest_runme.pl
@@ -1,0 +1,20 @@
+use strict;
+use warnings;
+use Test::More tests => 8;
+BEGIN { use_ok('argcargvtest') }
+require_ok('argcargvtest');
+
+my $largs = ["hi", "hola", "hello"];
+is(argcargvtest::mainc($largs), 3, "test main typemap 1");
+
+my $targs = ["hi", "hola"];
+is(argcargvtest::mainv($targs, 1), "hola", "test main typemap 2");
+
+my $errorVal = 0;
+my $ret = eval qq(argcargvtest::mainv("hello", 1); \$errorVal = 1;);
+is($ret, undef, "test main typemap 3");
+is($errorVal, 0, "test main typemap 4");
+
+is(argcargvtest::initializeApp($largs), undef, "test main typemap 5");
+
+ok(1, "done");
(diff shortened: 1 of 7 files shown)


wsfulton:
Could you please use the existing testcase [argcargv.i](https://github.com/swig/swig/blob/master/Examples/test-suite/argcargvtest.i) and convert one of the runtime tests [argcargvtest_runme.py](https://github.com/swig/swig/blob/master/Examples/test-suite/python/argcargvtest_runme.py) or [argcargvtest_runme.m](https://github.com/swig/swig/blob/master/Examples/test-suite/octave/argcargvtest_runme.m) to each of the newly supported languages. That way, I'll know that the new typemaps work. Just add in new Examples/test-suite/tcl/argcargvtest_runme.tcl etc files and in the Examples/test-suite/tcl directory run `make argcargvtest.ctest` to test locally. Also needs adding to the Examples/test-suite/tcl/Makefile.in file, although I'll probably refactor and move this testcase into Examples/test-suite/common.mk.

erezgeva:
> make argcargvtest.ctest

Hi,

Tried the `make argcargvtest.ctest`
Get 
`error: unknown type name ‘bool’`

Is this test written for C++ only?

I saw the 'bool' used here: `Examples/test-suite/argcargvtest.i`

I though the test need to pass both C and C++.

I get the same results when I run the test in python.

`test-suite/python (master)$ make argcargvtest.ctest`
Fails with the same error on bool.

Using `make argcargvtest.cpptest` pass.

erezgeva:
Add test for Perl5, PHP, and TCL.
Also for Ruby. 
Ruby already have argcargv.i but did not have a test.
Octave have a test but not a argcargv.i.
Only python had them both.

On Pull request [#2275](https://github.com/swig/swig/pull/2275), I add test for Lua.

So we should have now, 6 languages that support argcargv.i and have a test:
Python, Ruby, Lua, Perl5, PHP and TCL.

Erez

erezgeva:
Thanks for adding the 2 pull request.
Thanks for the updates.
And thanks for maintaining this wonderful tool. :-)

PR #2277, opened 2022-05-09
Python: define PY_SSIZE_T_CLEAN

xypron:
It is recommended to always define PY_SSIZE_T_CLEAN before including
Python.h. Cf. https://docs.python.org/3/c-api/intro.html

This avoids errors like:

    SystemError: PY_SSIZE_T_CLEAN macro must be defined for '#' formats

Signed-off-by: Heinrich Schuchardt <heinrich.schuchardt@canonical.com>

### DIFF
--- a/Lib/python/embed.i
+++ b/Lib/python/embed.i
@@ -29,7 +29,9 @@ As far as I know, this module is C++ safe.
 #endif
 
 %wrapper %{
-
+#ifndef PY_SSIZE_T_CLEAN
+#define PY_SSIZE_T_CLEAN
+#endif
 #include <Python.h>
 
 #ifdef __cplusplus
--- a/Lib/python/pyruntime.swg
+++ b/Lib/python/pyruntime.swg
@@ -11,6 +11,9 @@
 # include <corecrt.h>
 #endif
 
+#ifndef PY_SSIZE_T_CLEAN
+#define PY_SSIZE_T_CLEAN
+#endif
 #if defined(_DEBUG) && defined(SWIG_PYTHON_INTERPRETER_NO_DEBUG)
 /* Use debug wrappers with the Python release dll */
 # undef _DEBUG


xypron:
Build failures are due to CI using non-existent ppa:deadsnakes/python3.10. There is a ppa:deadsnakes/python3.10-jammy.

wsfulton:
Have you got a handy example that triggers this error. I don't see it in our CI.

ojwb:
SWIG doesn't use any of the `#` formats in question, so this won't manifest in SWIG itself.

I'm not really sure if it's best for SWIG to define this automatically or not.  It seems doing so risks breaking existing working user interface files with custom typemaps which use `#` formats and `int` with Python 3.9 or older, though such interface files will presumably break with Python 3.10 anyway (or else are carefully conditionalising the type used rather than just defining `PY_SSIZE_T_CLEAN` which seems an odd approach).  SWIG defining does make life a little easier for people with custom typemaps using `#` formats and targetting both Python >= 3.10 and < 3.10.

Anyway if we are going to do this, SWIG 4.1.0 is certainly an appropriate time to as it isn't an entirely compatible change.

wsfulton:
We need to allow users to not have PY_SSIZE_T_CLEAN defined if that is what they need. So I'll merge this with a modification to instead define PY_SSIZE_T_CLEAN only if SWIG_NO_PY_SSIZE_T_CLEAN is not defined.

hramrach:
> We need to allow users to not have PY_SSIZE_T_CLEAN defined if that is what they need. So I'll merge this with a modification to instead define PY_SSIZE_T_CLEAN only if SWIG_NO_PY_SSIZE_T_CLEAN is not defined.

That will still give a redefinition warning when PY_SSIZE_T_CLEAN is defined by the user.

ojwb:
@hramrach If you want to propose a change please open a new PR with it.

Issue #11, opened 2013-01-13
r fails director_nspace_director_name_collision


joequant:
r fails this regression test.  The problem is that R should add the namespace to the internal methods.

ojwb:
`make -C Examples/test-suite/r director_nspace_director_name_collision.cpptest` now passes:

```
if [ -f ./director_nspace_director_name_collision_runme.R ]; then echo "checking r testcase director_nspace_director_name_collision (with run test)" ; else echo "checking r testcase director_nspace_director_name_collision" ; fi;
checking r testcase director_nspace_director_name_collision
make -f ../../../Examples/Makefile SRCDIR='./' CXXSRCS='' SWIG_LIB_DIR='../../../Lib' SWIGEXE='../../../swig' INCLUDES='-I../../../Examples/test-suite' SWIGOPT='-outcurrentdir -I../../../Examples/test-suite' NOLINK=true TARGET='director_nspace_director_name_collision' INTERFACEDIR='../' INTERFACE='director_nspace_director_name_collision.i' r_cpp
make[1]: Entering directory '/home/olly/git/swig/Examples/test-suite/r'
env SWIG_LIB=../../../Lib  ../../../swig -c++ -r -outcurrentdir -I../../../Examples/test-suite -o director_nspace_director_name_collision_wrap.cpp  ./../director_nspace_director_name_collision.i
( PKG_CPPFLAGS="-I. -I./ -I../../../Examples/test-suite"  R CMD SHLIB -o director_nspace_director_name_collision.so director_nspace_director_name_collision_wrap.cpp    > /dev/null )
make[1]: Leaving directory '/home/olly/git/swig/Examples/test-suite/r'
if [ -f ./director_nspace_director_name_collision_runme.R ]; then env LD_LIBRARY_PATH=.:$LD_LIBRARY_PATH PATH=.:"$PATH"  R CMD BATCH --quiet --no-save --no-restore '--args .' ./director_nspace_director_name_collision_runme.R || (cat ./director_nspace_director_name_collision_runme.Rout ; false); else  R CMD BATCH --quiet --no-save --no-restore '--args .' ./director_nspace_director_name_collision.R || (cat ./director_nspace_director_name_collision.Rout ; false); fi
```

`git bisect` shows this was fixed by 6fecb05379da3dd3e7a2dd2711d18b89db3dea28 (though that revision fails for other reasons without 122f61cb0bcedb15db0612f9519de4d396f75efc being applied too).

Issue #10, opened 2013-01-08
Swig compilation error on Centos5.8


patroqueeet:
Dear all,

I'm quite a newbee in compiling applications. While trying to install swig 1.3.40 during "make" execution I ran into

```
g++ -DHAVE_CONFIG_H   -I../Source/Include -I../Source/CParse -I../Source/Include -I../Source/DOH -I../Source/CParse -I../Source/Preprocessor -I../Source/Swig -I../Source/Modules   -g -O2 -Wall -W -ansi -pedantic -MT Modules/allegrocl.o -MD -MP -MF $depbase.Tpo -c -o Modules/allegrocl.o Modules/allegrocl.cxx &&\
        mv -f $depbase.Tpo $depbase.Po
/tmp/ccG9S2WG.s: Assembler messages:
/tmp/ccG9S2WG.s:6369: Error: bad register name `%'
/tmp/ccG9S2WG.s:7280: Error: bad register name `%'
/tmp/ccG9S2WG.s:8259: Error: bad register name `%'
/tmp/ccG9S2WG.s:11930: Error: bad register name `%'
/tmp/ccG9S2WG.s:13747: Error: bad register name `%'
/tmp/ccG9S2WG.s:13751: Error: bad register name `%'
/tmp/ccG9S2WG.s:13884: Error: bad register name `%'
make[2]: *** [Modules/allegrocl.o] Error 1
make[2]: Leaving directory `/tmp/swig/swig-1.3.40/Source'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/tmp/swig/swig-1.3.40/Source'
make: *** [source] Error 2
```

Though I scanned all the usual source, I couldn't find a solution. I would be very glad if you can point me into the right direction.

Best

J.

PR #2284, opened 2022-05-21
[Ruby]  fix signedness

erezgeva:
Fix comparison of integer expressions of different signedness.
For the 2 `getslice()` functions, use a new function that use proper casting:
	`swig::check_index_insert(ptrdiff_t i, size_t size)`
In `RubySequence_Cont`, use standard `size_t` for size_type.

Testing with
`Examples/test-suite/ruby (Ruby-getslice)$ make li_std_vector.cpptest`
`Examples/test-suite/ruby (Ruby-getslice)$ make cpp11_std_array.cpptest`

And adding `-Wall` to both compiling and calling swig.

### DIFF
--- a/Lib/ruby/rubycontainer.swg
+++ b/Lib/ruby/rubycontainer.swg
@@ -101,7 +101,7 @@ namespace swig {
   inline Sequence*
   getslice(const Sequence* self, Difference i, Difference j) {
     typename Sequence::size_type size = self->size();
-    typename Sequence::size_type ii = swig::check_index(i, size, (i == size && j == size));
+    typename Sequence::size_type ii = swig::check_index(i, size, (i == (Difference)size && j == (Difference)size));
     typename Sequence::size_type jj = swig::slice_index(j, size);
 
     if (jj > ii) {
@@ -135,6 +135,8 @@ namespace swig {
       std::advance(sb,ii);
       std::advance(se,jj);
       self->erase(sb,se);
+      sb = self->begin();
+      std::advance(sb,ii);
       self->insert(sb, v.begin(), v.end());
     }
   }
@@ -338,7 +340,7 @@ namespace swig
     typedef T value_type;
     typedef T* pointer;
     typedef int difference_type;
-    typedef int size_type;
+    typedef std::size_t size_type;
     typedef const pointer const_pointer;
     typedef RubySequence_InputIterator<T, reference> iterator;
     typedef RubySequence_InputIterator<T, const_reference> const_iterator;
@@ -849,7 +851,7 @@ namespace swig
           i = len + i;
       }
       Sequence::difference_type j = length + i;
-      if ( j > static_cast<Sequence::difference_type>(len) ) {
+      if ( j > static_cast<Sequence::difference_type>(len) && v.size() > 0 ) {
         swig::resize( $self, j, *(v.begin()) );
       }
 
--- a/Examples/test-suite/ruby/li_std_containers_int_runme.rb
+++ b/Examples/test-suite/ruby/li_std_containers_int_runme.rb
@@ -1,0 +1,203 @@
+#!/usr/bin/env ruby
+
+# Check std::vector and std::list behaves the same as Ruby iterable
+# types (list)
+
+require 'swig_assert'
+require 'li_std_containers_int'
+
+def failed(a, b, msg)
+    # to_a() convert to array
+    # convert to array, so we can call join().
+    # join() join array element to string with a seperator.
+    raise RuntimeError, msg+" ["+a.to_a().join(", ")+"] ["+b.to_a().join(", ")+"]"
+end
+
+def compare_sequences(a, b)
+    # A start, out of range, may produce nil value
+    # As long as vector behave as list, we do not care :-)
+    if a == nil and b == nil then
+        return
+    end
+    if a.size != b.size then
+        failed(a, b, "different sizes")
+    end
+    a.each_index do |i|
+        if a[i] != b[i] then
+            failed(a, b, "elements are different")
+        end
+    end
+end
+
+def compare_containers(pythonlist, swigvector, swiglist)
+    compare_sequences(pythonlist, swigvector)
+    compare_sequences(pythonlist, swiglist)
+end
+
+# Check std::vector and std::list assignment behaves same as Ruby list
+# assignment including exceptions
+def container_insert_step(i, l, newval)
+    ps = (1..6).to_a
+    iv = Li_std_containers_int::Vector_int.new(ps)
+    il = Li_std_containers_int::List_int.new(ps)
+
+    # Ruby slice
+    begin
+        if l == nil then
+            ps[i] = newval
+        else
+            ps[i,l] = newval
+        end
+        ps_error = nil
+    rescue => e
+        ps_error = e
+    end
+
+    begin
+        if l == nil then
+            iv[i] = newval
+        else
+            iv[i,l] = newval
+        end
+        iv_error = nil
+    rescue => e
+        iv_error = e
+    end
+
+    begin
+        if l == nil then
+            il[i] = newval
+        else
+            il[i,l] = newval
+        end
+        il_error = nil
+    rescue => e
+        il_error = e
+    end
+
+    if iv_error != nil and il_error == nil then
+        raise RuntimeError, "ValueError std::vector<> fails while std::list<> pass: " + iv_error.to_s
+    end
+    if iv_error == nil and il_error != nil then
+        raise RuntimeError, "ValueError std::vector<> pass while std::list<> fail: " + il_error.to_s
+    end
+    if ps_error != nil and iv_error == nil then
+        raise RuntimeError, "ValueError C++ wrapper should fail like ruby: " + ps_error.to_s
+    end
+    if ps_error != nil and iv_error != nil and il_error != nil then
+        compare_containers(ps, iv, il)
+    end
+end
+
+# Check std::vector and std::list delete behaves same as Ruby list
+# delete including exceptions
+def container_delete(i, j)
+    ps = (1..6).to_a
+    iv = Li_std_containers_int::Vector_int.new(ps)
+    il = Li_std_containers_int::List_int.new(ps)
+
+    # Ruby slice
+    begin
+        if j == nil then
+            ps[i] = nil
+        else
+            ps[i,(j - i)] = nil
+        end
+        ps_error = nil
+    rescue => e
+        ps_error = e
+    end
+
+    # std::vector<int>
+    begin
+        if j == nil then
+            iv[i] = nil
+        else
+            iv[i,(j - i)] = nil
+        end
+        iv_error = nil
+    rescue => e
+        iv_error = e
+    end
+
+    # std::list<int>
+    begin
+        if j == nil then
+            il[i] = nil
+        else
+            il[i,(j - i)] = nil
+        end
+        il_error = nil
+    rescue => e
+        il_error = e
+    end
+
+    if iv_error != nil and il_error == nil then
+        raise RuntimeError, "ValueError std::vector<> fails while std::list<> pass: " + iv_error.to_s
+    end
+    if iv_error == nil and il_error != nil then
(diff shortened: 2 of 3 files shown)


wsfulton:
I don't know if this is okay without studying Ruby containers and slices in much more detail, in particular behaviour when negative indexes are used because that will dictate the C++ types to use and the appropriate C++ casting. This is quite a big change as it changes the api and I would need to understand why int was chosen in the first place and why size_t is more correct. I expect that a similar change for difference_type to ptrdiff_t from int would be needed.

I thoroughly overhauled Python slicing a few years back and the way I checked it was all good was by comparing against native container's slicing, see [li_std_containers_int_runme.py](https://github.com/swig/swig/blob/1ad77e8c9b955dfbf4c7b48890dcf49278b1e422/Examples/test-suite/python/li_std_containers_int_runme.py). There might be some sort of similar testing in Ruby, but we'd need to find it to make sure the changes okay. Otherwise converting that test to Ruby would be the way forward, but remove the additional Python complexity of slicing steps.

erezgeva:
> slicing

I share your concern.
There is a possibility for negative index.
If you look on `swig::check_index`
It does handle negative index 
`if ( i < 0 ) {`
`      if ((size_t) (-i) <= size)`
`	return (size_t) (i + size);`

So negative index translate to `i + size`.
But negative index  can never be equal to `size`.
If you look on `swig::check_index` end you see
` } else if (insert && ((size_t) i == size)) {`
`      return size;`
`    }`

So only positive index that equal to size is qualified for insert.
I think the insert flag is to ensure both indexes meet the same condition for insert.
Insert done with i >= 0 && (size_t) i == size for both indexes.

I will check [li_std_containers_overload_runme.rb](https://github.com/swig/swig/blob/master/Examples/test-suite/ruby/li_std_containers_overload_runme.rb)
And the generated code of 
`Examples/test-suite/ruby (Ruby-getslice)$ make li_std_containers_overload.cpptest`


### As for Ruby syntax:
See: https://ruby-doc.org/core-2.7.0/Array.html
_Array indexing starts at 0, as in C or Java.
 A negative index is assumed to be relative to the end of the array—that is, an index of -1 indicates the last element of the array, -2 is the next to last element in the array, and so on._

This description match 
`if ( i < 0 ) {`
`      if ((size_t) (-i) <= size)`
`	return (size_t) (i + size);`
And from description you can dictate that negative index can not insert a new element.

erezgeva:
Looking on generated `li_std_containers_overload_wrap.cxx`

The use of `getslice(const Sequence* self, Difference i, Difference j) `

Looking on `SWIGINTERN VALUE std_vector_Sl_X_Sg__slice__SWIG(...)`
It implement slice of std::vector<>with this/self, a start index that can be negative
` if ( i < 0 ) {`
`          i = self->size() + i;`
And length for the number of elements to slice.
Then the length is translate into an end indexm `length + i`.
Now `getslice` is called with a start index, and end index.
And if both start index and end index equal size.
Both `check_index()` and `slice_index()` will return 'size' value.
And `jj == ii == size` for insert

Now the end of `getslice`
`if (jj > ii) {`
...
Return existing elements
`   } else {`
`      return new Sequence();`
Will insert a new single element.

The new `swig::check_index_insert` is aligned with current code and ruby language.
Only non-negative and equal to size can be insert.

erezgeva:
Tried to translate li_std_containers_int to ruby

1. notice that, out of range indexes, works differently for containers.
2. `slice!` is not implemented, so it is not possible to delete a range.
3. Other array methods are not implemented as, `join`. Though we can call `join` after converting to to array with `to_a`.
4. I get 'segmentation fault'. I will try to debug and see if it comes from C++ code, and where.

I think Ruby containers need more work, apart of this change
[seg_fault.txt](https://github.com/swig/swig/files/9104436/seg_fault.txt)
.

erezgeva:
Run the script wit debugger
[seg_falt_gdb.txt](https://github.com/swig/swig/files/9114851/seg_falt_gdb.txt)

I add the script to the pull request.

erezgeva:
I just rebase the commit. No progress yet :-(

erezgeva:
Did some cleaning of the testing.

The bug seems to be in `std_list.i`.
Outside ruby. And probably not related to my changes in the container.
It seems a problem of freeing unallocated memory.
I will try `valgrind` on the test.

erezgeva:
The error is probably in `Lib/ruby/rubycontainer.swg`
`setslice()` line 142. The `self->erase` seems trying to delete an empty object.

ojwb:
> The self->erase seems trying to delete an empty object.

You mean it's called with `sb == se`?  That should be OK - see https://en.cppreference.com/w/cpp/container/list/erase which says:

> if first == last: erasing an empty range is a no-op.

erezgeva:
> > The self->erase seems trying to delete an empty object.
> 
> You mean it's called with `sb == se`? That should be OK - see https://en.cppreference.com/w/cpp/container/list/erase which says:
> 
> > if first == last: erasing an empty range is a no-op.

Not sure. I'll try with debugger and see the values.
The crash however is in a different location, `traits_asptr::asptr` from `Lib/ruby/rubycontainer.swg' line 55.
I suspect there might be a race condition.

erezgeva:
`(gdb) bt`
`#0   tcache_get (tc_idx=<optimized out>)         at malloc.c:2937`
`#1   __GI___libc_malloc (bytes=24)               at malloc.c:3051`
`#2   operator new(unsigned long)  from /lib/x86_64-linux-gnu/libstdc++.so.6`
`#3   swig::traits_asptr_stdseq<std::__cxx11::list<int, std::allocator<int> >, int>::asptr`
`     (obj=93824995849400, seq=0x7fffffffcf80)  at li_std_containers_int_wrap.cxx:4093`
`#4   swig::traits_asptr<std::__cxx11::list<int, std::allocator<int> > >::asptr`
`     (obj=93824995849400, lis=0x7fffffffcf80) at li_std_containers_int_wrap.cxx:4198`
`#5   swig::asptr<std::__cxx11::list<int, std::allocator<int> > >`
`     (obj=93824995849400, vptr=0x7fffffffcf80) at li_std_containers_int_wrap.cxx:2930`
`#6   _wrap_new_list_int__SWIG_1   at li_std_containers_int_wrap.cxx:6742`
`#7   _wrap_new_list_int           at li_std_containers_int_wrap.cxx:7380`
`#8   ?? ()             from libruby-2.7.so.2.7`
`#9   rb_vm_call0 ()    from libruby-2.7.so.2.7`
`#10  rb_vm_call_kw ()  from libruby-2.7.so.2.7`
`#11  rb_funcallv_kw () from libruby-2.7.so.2.7`

`(gdb) frame 3`
`sequence *pseq = new sequence();`
Line 1017 `Lib/ruby/rubycontainer.swg` in
   `<class Seq, class T = typename Seq::value_type > traits_asptr_stdseq::asptr(VALUE, sequence **)`
`(gdb) p *seq`
`$5 = (std::__cxx11::list<int, std::allocator<int> > *) 0x0`

Seems the crash is in the allocation of empty `std::list`.

erezgeva:
The class in the delete function `free_std_list_Sl_int_Sg_`    `std::list< int > *arg1` without the allocator.
But allocation with `argc == 1`, is done with `std::list< int,std::allocator< int > >`.

erezgeva:
Another question.
[`template< class Alloc >struct allocator_traits`  since C++11 in  <memory>, why do we need to define it?](https://en.cppreference.com/w/cpp/header/memory)

Why do we define it in `Lib/ruby/rubycontainer.swg`?
For older C++ or C we can define it. Perhaps the container should be used only with C++ 11 and above?

erezgeva:
Try to remove allocator in `Lib/std/std_list.i`.
Same results.

erezgeva:
I reduce the test to few lines.

list_crash.i
`%module list_crash`
`%include std_list.i`
`%template(list_int) std::list<int>;`

list_crash_runme.rb
`#!/usr/bin/env ruby`
`require 'list_crash'`
`l = List_crash::List_int.new([5, 7])`
`l[-2,2] = [9]`

The expected result is the same as using ruby list.
The list should reduce from two elements `[5, 7]` to a single element `[9]`.

Looking with debugger we end up in `Lib/ruby/rubycontainer.swg` line 138 in `setslice` function:
`      typename Sequence::iterator sb = self->begin();`
`      typename Sequence::iterator se = self->begin();`
`      std::advance(sb,ii);`
`      std::advance(se,jj);`
`      self->erase(sb,se);`
`      self->insert(sb, v.begin(), v.end());`

Where `self` is `std::list<int>` with 2 elements  `[5, 7]`
`ii = 0`
`jj = 2`
`v` is a `std::list<int>` list with the new element `[9]`.

The crash happens inside `insert()`

`    #0  0x00007fffe66b960b in std::__detail::_List_node_base::_M_transfer(std::__detail::_List_node_base*, std::__detail::_List_node_base*) () from /lib/x86_64-linux-gnu/libstdc++.so.6`
`    #1  0x00007fffe6956835 in std::__cxx11::list<int, std::allocator<int> >::_M_transfer (this=0x5555557b64e0, __position=5, __first=9, __last=1) at /usr/include/c++/12/bits/stl_list.h:1989`
`    #2  0x00007fffe69561a2 in std::__cxx11::list<int, std::allocator<int> >::splice (this=0x5555557b64e0, __position=5, __x=...) at /usr/include/c++/12/bits/stl_list.h:1621`
`    #3  0x00007fffe6955360 in std::__cxx11::list<int, std::allocator<int> >::splice (this=0x5555557b64e0, __position=5, __x=empty std::__cxx11::list) at /usr/include/c++/12/bits/stl_list.h:1632`
`    #4  0x00007fffe69550b6 in std::__cxx11::list<int, std::allocator<int> >::insert<std::_List_const_iterator<int>, void> (this=0x5555557b64e0, __position=5, __first=9, __last=1) at /usr/include/c++/12/bits/list.tcc:141`

erezgeva:
Hi [wsfulton](https://github.com/wsfulton)

I manage to reduce test and pin point the crash code.
The crash seems a C++ bug.

I wrote a pure C++ test mimic the test which crash :-)

See [ruby_crash.cpp](https://github.com/swig/swig/pull/2284/files#diff-bee1b015fdf2089049d1711ba6eaf5afb8c01b1a5c7c8788c01fdfcd340c1976)

Erez

erezgeva:
Hi [wsfulton](https://github.com/wsfulton),

Found what was missing in `Lib/python/pycontainer.swg`.
After calling `self->erase(sb,se);`
`sb` need to initialise again with:
`      sb = self->begin();`
`      std::advance(sb,ii);`

The same fix is needed `setslice` in `Lib/r/rcontainer.swg`, `Lib/octave/octcontainer.swg`.
I do not fix them as I do not plan to add testing.

erezgeva:
The error comes from changes in the STL.
They are not related to this PR.

Lib/ruby/rubycontainer.swg
```diff
@@ -101,7 +101,8 @@ namespace swig {
   inline Sequence*
   getslice(const Sequence* self, Difference i, Difference j) {
     typename Sequence::size_type size = self->size();
-    typename Sequence::size_type ii = swig::check_index(i, size, (i == size && j == size));
+    typename Sequence::size_type ii = swig::check_index(i, size,
+      (i == (Difference)size && j == (Difference)size));
```
wsfulton, line 105: Why do you put this on two lines, please put it back. Same for getslice.

erezgeva: Done

wsfulton:
Hi @erezgeva, nice work getting to the bottom of this. I know from improving the Python equivalent that this is not easy and just takes time and effort. The patch looks good to me to merge in for swig-4.2.1 and I'll do that later today. Thanks also for checking for equivalent problems in R and Octave.

erezgeva:
> Hi @erezgeva, nice work getting to the bottom of this. I know from improving the Python equivalent that this is not easy and just takes time and effort. The patch looks good to me to merge in for swig-4.2.1 and I'll do that later today. Thanks also for checking for equivalent problems in R and Octave.

Thanks. My pleasure.

Issue #2314, opened 2022-07-12
Proposal to drop support for Python 3.2


wsfulton:
Currently SWIG supports Python 2.7 as well as Python 3.2 and later. Recently the Github Actions CI has stopped working for Python 3.2 and I don't have the time to investigate why this Python package stopped working. Feb 2016 is when security fixes were last provided for 3.2, which seems far enough back to drop support for in the pending swig-4.1.0 version, especially as this is one of the earlier Python 3 versions.

jschueller:
good idea, I can give a hand

ojwb:
Seems sensible to me - better to direct our finite efforts elsewhere.

PR #2315, opened 2022-07-12
Require py>=3.3

jschueller:
see #2314


ojwb:
#2314 says "drop support for Python 3.2" not "drop support for Python <= 3.2" - I very much doubt @wsfulton is suggesting we also drop Python 2 support as you're attempting to do here.

jschueller:
oh, I thoughts his arguments applied to 2.7 too
I guess there's not much to do then

wsfulton:
Yeah, sorry if not entirely clear, but it is to drop just 3.2 support.

jschueller:
ok

PR #2320, opened 2022-07-18
Drop support for Python 3.2

### LINKED ISSUE #2314: Proposal to drop support for Python 3.2

jschueller:
Closes #2314

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,10 +112,6 @@ jobs:
         - SWIGLANG: python
           PY2: 2
         - SWIGLANG: python
-          VER: '3.2'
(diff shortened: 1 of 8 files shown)


ojwb:
The docs need updating too, and dropping this support means we can drop special cases in various places, so I've pushed a commit with those.  I think I've got them all, but if anyone can see others please say.

I've also rebased onto current master to avoid the `CHANGES.current` update from preventing a clean merge.

PR #2329, opened 2022-07-25
Rework testing of C++14 and C++17 and add C++20

ojwb:
As discussed in #1622.

Automatically enabling C++20 support when the compiler supports it revealed a couple of existing testcases which didn't actually compile as C++20 - the first commit fixes these.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.1.0 (in progress)
 ===========================
 
+2022-07-25: olly
+	    Support for C++14 binary integer literals in preprocessor expressions.
+
 2022-07-20: wsfulton
             [C#, Java] Ensure the order of interfaces generated in proxy interfaces for the
             %interface family of macros is the same as that parsed from the bases in C++.
--- a/Examples/test-suite/cffi/Makefile.in
+++ b/Examples/test-suite/cffi/Makefile.in
@@ -7,6 +7,9 @@ CFFI         = @CFFIBIN@
 SCRIPTSUFFIX = _runme.lisp
 
 HAVE_CXX11   = @HAVE_CXX11@
+HAVE_CXX14   = @HAVE_CXX14@
+HAVE_CXX17   = @HAVE_CXX17@
+HAVE_CXX20   = @HAVE_CXX20@
 srcdir       = @srcdir@
 top_srcdir   = @top_srcdir@
 top_builddir = @top_builddir@
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -91,7 +91,10 @@ CPP_TEST_BROKEN += \
 	template_default_pointer \
 	template_private_assignment \
 	template_expr \
-	$(CPP11_TEST_BROKEN)
+	$(CPP11_TEST_BROKEN) \
+	$(CPP14_TEST_BROKEN) \
+	$(CPP17_TEST_BROKEN) \
+	$(CPP20_TEST_BROKEN)
 
 
 # Broken C test cases. (Can be run individually using: make testcase.ctest)
@@ -632,6 +635,24 @@ CPP11_TEST_BROKEN = \
 #	cpp11_variadic_templates \    # Broken for some languages (such as Java)
 #	cpp11_reference_wrapper \     # No typemaps
 
+# C++14 test cases.
+CPP14_TEST_CASES += \
+
+# Broken C++14 test cases.
+CPP14_TEST_BROKEN = \
+
+# C++17 test cases.
+CPP17_TEST_CASES += \
+
+# Broken C++17 test cases.
+CPP17_TEST_BROKEN = \
+
+# C++20 test cases.
+CPP20_TEST_CASES += \
+
+# Broken C++20 test cases.
+CPP20_TEST_BROKEN = \
+
 # Doxygen support test cases: can only be used with languages supporting
 # Doxygen comment translation (currently Python and Java) and only if not
 # disabled by configure via SKIP_DOXYGEN_TEST_CASES.
@@ -697,6 +718,18 @@ endif
 
 ifeq (1,$(HAVE_CXX11))
 CPP_TEST_CASES += $(CPP11_TEST_CASES)
+endif
+
+ifeq (1,$(HAVE_CXX14))
+CPP_TEST_CASES += $(CPP14_TEST_CASES)
+endif
+
+ifeq (1,$(HAVE_CXX17))
+CPP_TEST_CASES += $(CPP17_TEST_CASES)
+endif
+
+ifeq (1,$(HAVE_CXX20))
+CPP_TEST_CASES += $(CPP20_TEST_CASES)
 endif
 
 # C test cases. (Can be run individually using: make testcase.ctest)
@@ -783,6 +816,9 @@ command_line_define.ctest: SWIGOPT += -DFOO
 C_TEST_CASES := $(filter-out $(FAILING_C_TESTS),$(C_TEST_CASES))
 CPP_TEST_CASES := $(filter-out $(FAILING_CPP_TESTS),$(CPP_TEST_CASES))
 CPP11_TEST_CASES := $(filter-out $(FAILING_CPP_TESTS),$(CPP11_TEST_CASES))
+CPP14_TEST_CASES := $(filter-out $(FAILING_CPP_TESTS),$(CPP14_TEST_CASES))
+CPP17_TEST_CASES := $(filter-out $(FAILING_CPP_TESTS),$(CPP17_TEST_CASES))
+CPP20_TEST_CASES := $(filter-out $(FAILING_CPP_TESTS),$(CPP20_TEST_CASES))
 MULTI_CPP_TEST_CASES := $(filter-out $(FAILING_MULTI_CPP_TESTS),$(MULTI_CPP_TEST_CASES))
 
 
@@ -796,6 +832,9 @@ BROKEN_TEST_CASES = 	$(CPP_TEST_BROKEN:=.cpptest) \
 
 ALL_CLEAN = 		$(CPP_TEST_CASES:=.clean) \
 			$(CPP11_TEST_CASES:=.clean) \
+			$(CPP14_TEST_CASES:=.clean) \
+			$(CPP17_TEST_CASES:=.clean) \
+			$(CPP20_TEST_CASES:=.clean) \
 			$(C_TEST_CASES:=.clean) \
 			$(MULTI_CPP_TEST_CASES:=.clean) \
 			$(CPP_TEST_BROKEN:=.clean) \
@@ -823,6 +862,12 @@ check-c: $(C_TEST_CASES:=.ctest)
 check-cpp: $(CPP_TEST_CASES:=.cpptest)
 
 check-cpp11: $(CPP11_TEST_CASES:=.cpptest)
+
+check-cpp14: $(CPP14_TEST_CASES:=.cpptest)
+
+check-cpp17: $(CPP17_TEST_CASES:=.cpptest)
+
+check-cpp20: $(CPP20_TEST_CASES:=.cpptest)
 
 check-multicpp: $(MULTI_CPP_TEST_CASES:=.multicpptest)
 
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -168,11 +168,6 @@ CPP_TEST_CASES += \
 	cpp_parameters \
 	cpp_static \
 	cpp_typedef \
-	cpp14_binary_integer_literals \
-	cpp17_hex_floating_literals \
-	cpp17_nested_namespaces \
-	cpp17_nspace_nested_namespaces \
-	cpp17_u8_char_literals \
 	curiously_recurring_template_pattern \
 	default_args \
 	default_arg_expressions \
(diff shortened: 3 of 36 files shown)


ojwb:
Some appveyor builds are unhappy.

Looking at the log, the first thing I notice is a lot of warnings:

```
[00:02:08] checking Examples/csharp/callback
[00:02:08] cl : Command line warning D9002 : ignoring unknown option '/std:c++11'
[00:02:08] cl : Command line warning D9002 : ignoring unknown option '/Zc:__cplusplus'
[00:02:09] cl : Command line warning D9002 : ignoring unknown option '/std:c++11'
[00:02:09] cl : Command line warning D9002 : ignoring unknown option '/Zc:__cplusplus'


I think essentially the problem is that `cccl` wrapper is buggy in its handling of `-std:*` because it always adds `/Zc:__cplusplus` to this, but "/Zc:__cplusplus option is only available starting in Visual Studio 2017 version 15.7" (https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170).

I think this bug was latent because `AX_CXX_COMPILE_STDCXX` was buggy in never detecting MSVC as being capable of handling anything other than C++98, but now I've fixed that we can see the cccl problem.

Also, `/std:c+11` isn't valid ("Note that the MSVC compiler does not, and never will, support a C++11, C++03, or C++98 standards version switch" from https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/).  We seem to end up using it due to configure thinking it works:

```
[00:01:31] checking whether C++11 to C++14 testing is enabled... no             
[00:01:31] checking whether cccl supports C++11 features with -std=c++11... yes 
[00:01:31] checking whether C++11 testing is enabled... yes                     
[00:01:31] checking CXXFLAGS to use for testing (PLATCXXFLAGS)... -std=c++11    
```

Another factor that's triggering this now is that SWIG's configure passes `noext` to `AX_CXX_COMPILE_STDCXX`, so it skips the test as to whether the compiler supports the requested C++ standard version by default, because compilers usually enable extensions by default, so if you want no extensions you need to specify an option.  Older `AX_CXX_COMPILER_STDCXX` versions (and the predecessor `_11` etc suffixed macros) didn't take that into account and would happily use the compiler without options even if that actually implemented extensions.

But I think this is really a cccl problem, because `AX_CXX_COMPILE_STDCXX` tests a compile with the option it picks, and it doesn't care about warnings so it thinks `cccl -std=c++11` works and picks it.

I think the answer to this is for cccl to explicitly special-case `-std=c++11` (and `03` and the GNU variants).  I worry though that rejecting them might result in `AX_CXX_COMPILE_STDCXX` not  being able to detect C++11 support in MSVC (at least when it doesn't support C++14) because with `noext` it needs to find an option to use in order to get C++11 without extensions, and Microsoft explicitly refuse to provide `/std:c++11`.  If there's a way to disable extensions, cccl could translate `-std=c++11` into that.  Or I guess it could no-op for `-std=c++11` which isn't strictly correct as you still get Microsoft extensions then.  Or maybe map `-std=c++11` to `/std=c++14`.

I'm not sure what the errors are, e.g.:

```
[00:08:23] checking csharp testcase cpp11_alignment
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(328): warning C4346: 'value_type': dependent name is not a type
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(328): note: prefix with 'typename' to indicate a type
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(337): note: see reference to class template instantiation 'Y<T>' being compiled
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(328): error C2061: syntax error: identifier 'value_type'
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(328): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(335): warning C4346: 'value_type': dependent name is not a type
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(335): note: prefix with 'typename' to indicate a type
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(335): error C2061: syntax error: identifier 'value_type'
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(335): error C2334: unexpected token(s) preceding '{'; skipping apparent function body
[00:08:23] cpp11_alias_nested_template_scoping_wrap.cxx(328): fatal error C1903: unable to recover from previous error(s); stopping compilation
[00:08:23] Internal Compiler Error in C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\amd64\cl.exe.  You will be prompted to send an error report to Microsoft later.
```

Looking at https://en.cppreference.com/w/cpp/compiler_support/11 I'm wondering if this is just that this compiler is too old to support C++11, but if so how did it pass `AX_CXX_COMPILE_STDCXX`'s C++11 compilation test?

@wsfulton Can you help here?

I could submit a patch to swig/cccl to special case `-std=c++11`, etc but I can't really test any of this except by throwing it at CI and seeing what sticks as I just don't use Microsoft Windows (and the more I look at these sort of problems, the more glad I am of that!  Their continued mishandling of `__cplusplus` is particularly astonishing...)

**Comment by ojwb** [comment]
Also, note that while mostly this is refactoring, the first commit fixes some C++20 compatibility problems in testcases, and 777e203b4b6bc8662292a7a4ee77d5c446cf837e fixes missing support for binary literals in preprocessor expressions, which currently always quietly parse as `0` so while such literals are likely uncommon, the failure mode is problematic.  It's also a blocker for adding support for `<=>` as that needs the C++20 test harness support.

Hmm, I wonder if reverting the fix to `AX_CXX_COMPILE_STDCXX` would workaround the `cccl` bugs.  It would be better to fix `AX_CXX_COMPILE_STDCXX` since that makes it work with MSVC used without the cccl wrapper, but I guess that would be no worse than current SWIG git master (and 4.0.2).  I'll try that.

BTW, is `cccl` actually still needed?  The Xapian CI uses MSVC with MSYS and successfully configures and builds SWIG (without using cccl).  I don't think we try to run the SWIG test suite though, so maybe that's the part that needs `cccl`.

**Comment by ojwb** [comment]
Reverting the fix to `AX_CXX_COMPILE_STDCXX` helps, but uncovers more problems, some of which are additional generic C++20 compatibility issues I can reproduce on Linux too.  I've pushed attempted fixes for everything which was failing in the Python build log I looked at.

**Comment by ojwb** [comment]
Another run revealed more C++20 incompatibilities in existing test cases which I've pushed fixes for.

I've cleaned up the history a bit too.

**Comment by ojwb** [comment]
@wsfulton I got this to work by reverting the fix to `AX_CXX_COMPILE_STDCXX` but that just means the problematic interaction between that macro and cccl is once again latent.  It'll surface again once upstream merge my PR to fix `AX_CXX_COMPILE_STDCXX` to actually detect C++11 and up with MSVC and we pick up that new macro version, or once MSVC actually starts to define `__cplusplus` correctly.

I'll submit a PR against cccl, but I do wonder if we still need cccl here.

**Comment by wsfulton** [comment]
@ojwb, I'm struggling to understand all of these observations. 

> @wsfulton I got this to work by reverting the fix to AX_CXX_COMPILE_STDCXX but that just means the problematic interaction between that macro and cccl is once again latent.

Which commit is the fix and which is the revert?

I believe the AX_CXX_COMPILE_STDCXX worked correctly in serial 14 as the older Visual C++ compilers do not properly support C++11 and C++14 as __cplusplus is not set correctly. Previously we did not test C++11 and C++14 with these older compilers and I think that is okay as they are not standards conformant. Was it your intention in serial 15 of AX_CXX_COMPILE_STDCXX to enable Visual C++ to think it was supporting C++11 and C++14, even though they do not define __cplusplus correctly?

> [00:02:08] checking Examples/csharp/callback
[00:02:08] cl : Command line warning D9002 : ignoring unknown option '/std:c++11'
[00:02:08] cl : Command line warning D9002 : ignoring unknown option '/Zc:__cplusplus'

Was this before with your change or before it? Presumably the older compilers only.

**Comment by ojwb** [comment]
> @ojwb, I'm struggling to understand all of these observations.
> 
> > @wsfulton I got this to work by reverting the fix to AX_CXX_COMPILE_STDCXX but that just means the problematic interaction between that macro and cccl is once again latent.
> 
> Which commit is the fix and which is the revert?

I rebased the branch between that and merging and eliminated both the fix and the revert from the history.

However, you can see the fix at https://github.com/autoconf-archive/autoconf-archive/pull/255 (which was merged upstream yesterday).

> I believe the AX_CXX_COMPILE_STDCXX worked correctly in serial 14 as the older Visual C++ compilers do not properly support C++11 and C++14 as __cplusplus is not set correctly.

The problem here is that NO existing version of MSVC sets __cplusplus correctly by default for anything after C++98, but the version of the macro we're using (i.e. serial 14) assumes that if `__cplusplus` is less than the value for a particular standard that means that standard isn't supported.  While that may be true for older MSVC, it would mean we'd never test anything newer than C++98 under MSVC.  These tests are intended to speed up failing for a compiler without the required support - the structure is like this for each C++ version:

```
// If the compiler admits that it is not ready for C++11, why torture it?
// Hopefully, this will speed up the test.

#ifndef __cplusplus

#error "This is not a C++ compiler"

#elif __cplusplus < 201103L  
 
#error "This is not a C++11 compiler"
 
#else

// Do a compile test to actually test for C++11 support

#endif
```

The patch which is applied upstream in serial 15 just disables this attempted speed-up optimisation for MSVC and always tries the feature based compile tests there.  Effectively this reverts to how the older `AX_CXX_COMPILE_STDCXX_*` family of macros behaved as they didn't have this attempted optimisation, so it would restore the behaviour SWIG had before 82fb0540ca253e14f479ffcb4b94b5ee31b30fea ("Replace AX_CXX_COMPILE_STDCXX_11 with AX_CXX_COMPILE_STDCXX") which is what SWIG 4.0.2 had.

So really we do want this fix, but because of cccl's unhelpful (I'd argue buggy) handling we can't have it.

> Previously we did not test C++11 and C++14 with these older compilers and I think that is okay as they are not standards conformant. Was it your intention in serial 15 of AX_CXX_COMPILE_STDCXX to enable Visual C++ to think it was supporting C++11 and C++14, even though they do not define __cplusplus correctly?

No - as above it's to ignore the value of `__cplusplus` under MSVC and to instead rely on the feature based test for C++11 and C++14 support (like we used to).

> > [00:02:08] checking Examples/csharp/callback
> > [00:02:08] cl : Command line warning D9002 : ignoring unknown option '/std:c++11'
> > [00:02:08] cl : Command line warning D9002 : ignoring unknown option '/Zc:__cplusplus'
> 
> Was this before with your change or before it? Presumably the older compilers only.

You seem to have stopped mid-sentence here.

This is with the fix, but as per my earlier comments it's really due to cccl.  I've opened a PR with a suggested fix: https://github.com/swig/cccl/pull/17

However, do we actually still need cccl here at all?  The handling of MSVC by automake has improved over the years and we can now successfully build Xapian with MSVC and MSYS, just by using automake's `compile` wrapper:

```
$env:CXX="c:/projects/xapian/INST/share/automake-1.16/compile cl -nologo"
```

Xapian uses one automake patch that hasn't been merged upstream still, but that's only needed because we also use libtool (it patches `compile` to know that `.lo` is an object file and treat it the same as `.o` and `.obj`) so that's not relevant for SWIG as we don't use libtool.

**Comment by wsfulton** [comment]
Thanks for the explanations, I've got a clearer picture now. 

> The problem here is that NO existing version of MSVC sets __cplusplus correctly by default for anything after C++98, but the version of the macro we're using (i.e. serial 14) assumes that if __cplusplus is less than the value for a particular standard that means that standard isn't supported.  While that may be true for older MSVC, ...

So yes, by default, Visual C++ is not a standards conforming compiler. But this is also true of the latest versions as __cplusplus is not set correctly unless `/Zc:__cplusplus` is passed to it, then it does become a standards conforming compiler (but not by default!). So my view is that serial 14 was correct in detecting C++17 and C++20 for latest Visual C++ (is supported if user passes in `/Zc:__cplusplus /std:c++17` and not conformant if `/std:c++17` is passed in.

As cccl translates `-std=c++17` into `/Zc:__cplusplus /std:c++17` the combination works correctly with serial 14. Then intention was to only test standards conforming compilers. Maybe I've overlooked something, I'll experiment when I boot up Windows, but that's my current thinking.

Note that we use:
```
AX_CXX_COMPILE_STDCXX(17, [noext], [optional])
```
so standards conforming only as we use 'noext'. From the AX_CXX_COMPILE_STDCXX file:
```
# SYNOPSIS
#
#   AX_CXX_COMPILE_STDCXX(VERSION, [ext|noext], [mandatory|optional])

...
#   The second argument, if specified, indicates whether you insist on an
#   extended mode (e.g. -std=gnu++11) or a strict conformance mode (e.g.
#   -std=c++11).  If neither is specified, you get whatever works, with
#   preference for no added switch, and then for an extended mode.
```
Do your changes only apply to ext or also when noext is used?

wsfulton:
AX_CXX_COMPILE_STDCXX has been updated to serial 25 in f66ea491e40c56c3dec40d8835699ee74d8936de. I'm working on releasing cccl with various updates including https://github.com/swig/cccl/pull/17. Also of note is dealing with newer versions of autoconf which no longer just returns the compiler name, but the compiler name and some sort of preferred `-std=` setting, see 0cc821269c7b688ffe26133cf19a6a85491a1ed2.

PR #2331, opened 2022-07-26
[Tcl] Fix handling of long long on 32-bit platforms

ojwb:
Fixes https://sourceforge.net/p/swig/bugs/977/ reported by Vera and
Kjell Wooding.


hackalog:
Hahah. Was surprised to see this in my inbox, but glad to see a fix finally made it in.

Issue #1009, opened 2017-06-22
Support Python 3 stable API


vadz:
It would be great if SWIG could generate code using only [stable Python 3 API](https://docs.python.org/3/c-api/stable.html). This isn't trivial to do, i.e. just defining `Py_LIMITED_API` right now doesn't work at all as I wrote at the [old SF issue](https://sourceforge.net/p/swig/feature-requests/78/), but it would be worth spending more effort on this IMO as otherwise distributing binary Python3 modules is really painful because you need to build them for each 3.x version separately (at least with Python 2, you need to support 2 versions at worst, and 90% of people probably only need to support 2.7).

mromberg:
It seems Py_LIMITED_API is documented:

https://www.python.org/dev/peps/pep-0384/#id7

  It would be interesting to see what the script on that page:

http://svn.python.org/projects/python/branches/pep-0384/Tools/scripts/abitype.py

  thinks is wrong with some of the swig generated code.   I have no idea how complete that script is.

  One problem I can foresee is that swig would not only have to generate code that works for Py_LIMITED_API but it would need to also generate code that works for python2.   My guess is that the same code would not work for both.   So... yet another command line python option for swig would have to be done.

   Still, it might be nice to have the swig code for python3 to start heading in this direction.  It will probably have to be done a little bit at a time.

vadz:
Here is a more detailed list of problems that appear when trying to compile
SWIG-generated code with `-DPy_LIMITED_API`, in order from the simplest to
most difficult to solve ("NA" below means "not available in stable ABI"):

- `PyCFunction_GET_XXX()` macros NA, but they can be just replaced with the
  corresponding functions. We could keep macros for non-stable ABI, but the
  tiny gain from doing it is probably not worth it.
- `fastproxy` requires `PyInstanceMethod_New()` which is NA. I don't know
  how useful this option is, ideal (from simplicity point of view) would be to
  just drop it, but we could also keep it but make all the related code
  conditional on `Py_LIMITED_API` being not defined.
- `fastinit` requires `SWIG_Python_InitShadowInstance()` which uses tuple
  functions which are NA. Same analysis as for `fastproxy` above.
- `PyTypeObject` is now opaque, yet we use it *extensively*. There are a
  couple of dozens places where just `tp_name` is used and I don't know how to
  access it using just stable API. In most places, `tp_name` is used only for
  diagnostic purposes, but it would be still a pity to lose it in the error
  messages. And in a couple of places it is used for the type checks, although
[...]

PR #2344, opened 2022-08-30
Use PyObject instead of PySliceObject throughout with Python 3.2+

clintonstimpson:
Note: this also eliminates many compile errors when attempting to define Py_LIMITED_API.

See also #2190 where a new #ifdef/#endif section was created to solve this same problem.
Here, we leverage an already existing ifdef in pycontainer.swg.

### DIFF
--- a/Lib/python/pycontainer.swg
+++ b/Lib/python/pycontainer.swg
@@ -15,9 +15,9 @@
 #include <iostream>
 
 #if PY_VERSION_HEX >= 0x03020000
-# define SWIGPY_SLICE_ARG(obj) ((PyObject*) (obj))
+# define SWIGPY_SLICEOBJECT PyObject
 #else
-# define SWIGPY_SLICE_ARG(obj) ((PySliceObject*) (obj))
+# define SWIGPY_SLICEOBJECT PySliceObject
 #endif
 %}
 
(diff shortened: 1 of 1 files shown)


clintonstimpson:
@ojwb @wsfulton , can you review?

ojwb:
I don't understand the patch.  A quick look at the headers for Python 3.2 and later shows they still have `PySliceObject` and crudely redefining it like this just seems like it's going to break things - e.g. consider code in a user typemap which uses `PySliceObject`...

clintonstimpson:
Good point you brought up about a user typemap.  This will need reworked.

ojwb:
Ah, so the type passed to various functions changed from `PySliceObject*` to `PyObject*` in Python 3.2.

It looks like uses of `SWIGPY_SLICE_ARG(slice)` can now just be replaced by `slice` and this macro removed as the `slice` argument to the various functions where this is used is now always the appropriate type already.

clintonstimpson:
Good idea about removing the SWIGPY_SLICE_ARG macro.  I've pushed an updated commit for that and I'm curious how the CI checks will do.

clintonstimpson:
linking #1009

ojwb:
Looks OK to me now.

@wsfulton Thoughts?

Issue #2114, opened 2021-12-03
Python: using the -builtin option doesn't work when using multiple embedded interpreters


jsenn:
This is similar to #2101. We use SWIG to allow users to interact with our SDK by writing Python scripts from within the UI of our application. Each time a user runs a script, a new Python interpreter is created, the extension module is imported, and then the script is run. All of this happens within our application's main process. This looks roughly like the following:

```c++
#include <Python.h>

extern "C"
{
PyObject* PyInit__test(void);
}

void
RunScript( const char* script )
{
	int importSucceeded = PyImport_AppendInittab( "_test", &PyInit__test ); // Import extension module

	Py_Initialize();
	int result = PyRun_SimpleString( script );
	Py_Finalize();
}
```
Our wrapped code looks like the following:
```c++
// Test.h
class Test
{
    static int Get1();
};
```
```c++
// Test.cpp
#include Test.h
int Test::Get1() { return 1; }
```
```
// test.i
%module test
%{
#include "Test.h"
%}

%include "Test.h"
```

Having run swig with the `-builtin` flag, if you call `RunScript()` with the script `import test; print(test.Test.Get1())` once it succeeds, but if you do it again, you'll get the following output:
```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
AttributeError: type object 'test.Test' has no attribute 'Get1'
```
For some reason the `Get1` function is no longer visible to the interpreter. Note that the same thing happens with non-static member functions.

I wasn't able to track this one down, but @wsfulton [suggested](https://github.com/swig/swig/pull/2102#issuecomment-984447459) in #2102 there may be issues with `SwigPyPacked_type`, which is involved in holding member function pointers, so that may be a good place to look.

wsfulton:
For the record, I replicated this by using your first block of code in a file cpprun.cxx with the addition of:
```c++
int main() {
  RunScript("import test; print(test.Test.Get1())");
  RunScript("import test; print(test.Test.Get1())");
  return 0;
}
```
and building
```bash
env SWIG_LIB=$HOME/swig/swig-4.3.1/Lib  $HOME/swig/swig-4.3.1/swig -python -builtin -c++  -o test_wrap.cxx test.i
g++ -g -c -fPIC -I.  -I/usr/include -std=c++20  test_wrap.cxx  Test.cpp   -I/usr/include/python3.12 -I/usr/include/python3.12 # -DSWIG_REFCNT_DEBUG
[...]

Issue #277, opened 2014-12-16
For safety, malloc result should be checked for NULL


ekeithkw:
Code generated for Java interfacing should check for NULL results from malloc, and throw an appropriate exception.  For example:

typedef struct _CharInfo {
    xCharInfo   metrics;    /\* info preformatted for Queries _/
    char       *bits;   /_ pointer to glyph image */
}           CharInfoRec;

.. generates a setter for 'bits' as:

SWIGEXPORT void JNICALL Java_com_attachmate_rx_xserver_dix_font_jni_RXFontProxyJNI_CharInfoRec_1bits_1set(JNIEnv *jenv, jclass jcls, jlong jarg1, jobject jarg1_, jstring jarg2) {
  struct _CharInfo *arg1 = (struct _CharInfo *) 0 ;
  char *arg2 = (char *) 0 ;

  (void)jenv;
  (void)jcls;
  (void)jarg1_;
  arg1 = _(struct _CharInfo *_)&jarg1; 
  arg2 = 0;
  if (jarg2) {
    arg2 = (char _)(_jenv)->GetStringUTFChars(jenv, jarg2, 0);
    if (!arg2) return ;
  }
  {
    free(arg1->bits);
    if (arg2) {
      arg1->bits = (char _) malloc(strlen((const char *)arg2)+1); // <-- possible NULL result
      strcpy((char *)arg1->bits, (const char *)arg2);
    } else {
      arg1->bits = 0;
    }
  }
  if (arg2) (_jenv)->ReleaseStringUTFChars(jenv, jarg2, (const char *)arg2);
}

The current code would probably get a segmentation fault and terminate the JVM.

ojwb:
The proposed patch isn't suitable as the file in question is `Lib/swig.swg` so is used for languages other than Java too - `SWIG_JavaThrowException` won't work there.

This probably needs some sort of "signal out of memory" mechanism that typemaps can use.

ojwb:
It's really not great that we just let the `NULL` pointer get used as that's undefined behaviour, but unless the attempted allocation was large it seems likely that attempting to handle it by throwing an exception in the target language will fail due to lack of memory too.

I don't really have any better ideas though, but any mechanism to signal out of memory should minimise doing things that need to allocate memory.

PR #2350, opened 2022-09-02
Use more portable PyType_FromSpec with Python 3.4+

clintonstimpson:
Also supports building with Py_LIMITED_API.

### DIFF
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -404,8 +404,8 @@ SwigPyStaticVar_new_getset(PyTypeObject *type, PyGetSetDef *getset) {
   return descr;
 }
 
-SWIGINTERN void
-SwigPyBuiltin_InitBases(PyTypeObject *type, PyTypeObject **bases) {
+SWIGINTERN PyObject*
+SwigPyBuiltin_InitBases(PyTypeObject **bases) {
   Py_ssize_t base_count = 0;
   PyTypeObject **b;
   PyObject *tuple;
@@ -415,8 +415,6 @@ SwigPyBuiltin_InitBases(PyTypeObject *type, PyTypeObject **bases) {
     bases[0] = SwigPyObject_type();
     bases[1] = NULL;
   }
-  type->tp_base = bases[0];
-  SWIG_Py_INCREF((PyObject *)bases[0]);
   for (b = bases; *b != NULL; ++b)
     ++base_count;
   tuple = PyTuple_New(base_count);
@@ -424,7 +422,7 @@ SwigPyBuiltin_InitBases(PyTypeObject *type, PyTypeObject **bases) {
     SWIG_Py_INCREF((PyObject *)bases[i]);
     PyTuple_SET_ITEM(tuple, i, (PyObject *)bases[i]);
   }
-  type->tp_bases = tuple;
+  return tuple;
 }
 
 SWIGINTERN PyObject *
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -404,7 +404,7 @@ SwigPyStaticVar_new_getset(PyTypeObject *type, PyGetSetDef *getset) {
   return descr;
 }
 
-SWIGINTERN PyObject*
+SWIGINTERN PyObject *
 SwigPyBuiltin_InitBases(PyTypeObject **bases) {
   Py_ssize_t base_count = 0;
   PyTypeObject **b;
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1064,7 +1064,7 @@ SwigPyObject_TypeOnce(void) {
     "SwigPyObject",
     sizeof(SwigPyObject),
     0,
-    Py_TPFLAGS_DEFAULT,
+    Py_TPFLAGS_DEFAULT|Py_TPFLAGS_BASETYPE,
     slots
   };
   return (PyTypeObject *)PyType_FromSpec(&spec);
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -335,10 +335,14 @@ swig_varlink_setattr(PyObject *o, char *n, PyObject *p) {
   return res;
 }
 
+#if PY_VERSION_HEX >= 0x03090000 || (!defined(SWIGPYTHON_BUILTIN) && PY_VERSION_HEX >= 0x03040000)
+#define SWIG_HEAPTYPES
+#endif
+
 SWIGINTERN PyTypeObject*
 swig_varlink_type(void) {
   static char varlink__doc__[] = "Swig var link object";
-#ifndef Py_LIMITED_API
+#ifndef SWIG_HEAPTYPES
   static PyTypeObject varlink_type;
   static int type_init = 0;
   if (!type_init) {
@@ -907,7 +911,7 @@ swigobject_methods[] = {
 SWIGRUNTIME PyTypeObject*
 SwigPyObject_TypeOnce(void) {
   static char swigobject_doc[] = "Swig object carries a C/C++ instance pointer";
-#ifndef Py_LIMITED_API
+#ifndef SWIG_HEAPTYPES
   static PyNumberMethods SwigPyObject_as_number = {
     (binaryfunc)0, /*nb_add*/
     (binaryfunc)0, /*nb_subtract*/
@@ -1173,7 +1177,7 @@ SwigPyPacked_dealloc(PyObject *v)
 SWIGRUNTIME PyTypeObject*
 SwigPyPacked_TypeOnce(void) {
   static char swigpacked_doc[] = "Swig object carries a C/C++ instance pointer";
-#ifndef Py_LIMITED_API
+#ifndef SWIG_HEAPTYPES
   static PyTypeObject swigpypacked_type;
   static int type_init = 0;
   if (!type_init) {
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -1156,11 +1156,11 @@ SwigPyPacked_Check(PyObject *op) {
   tp_name = PyObject_GetAttrString((PyObject *)op_type, "__name__");
   if (!tp_name)
     return 0;
-  cmp = PyUnicode_CompareWithASCIIString(tp_name, "SwigPyPacked");
+  cmp = PyUnicode_CompareWithASCIIString(tp_name, SWIG_name ".SwigPyPacked");
   SWIG_Py_DECREF(tp_name);
   return cmp == 0;
 #else
-  return (strcmp(op_type->tp_name, "SwigPyPacked") == 0);
+  return (strcmp(op_type->tp_name, SWIG_name ".SwigPyPacked") == 0);
 #endif
 }
 
@@ -1188,7 +1188,7 @@ SwigPyPacked_TypeOnce(void) {
       PyObject_HEAD_INIT(NULL)
       0,                                    /* ob_size */
 #endif
-      "SwigPyPacked",                       /* tp_name */
+      SWIG_name ".SwigPyPacked",            /* tp_name */
       sizeof(SwigPyPacked),                 /* tp_basicsize */
       0,                                    /* tp_itemsize */
       (destructor)SwigPyPacked_dealloc,     /* tp_dealloc */
@@ -1281,7 +1281,7 @@ SwigPyPacked_TypeOnce(void) {
     { 0, NULL }
   };
   PyType_Spec spec = {
-    "SwigPyPacked",
+    SWIG_name ".SwigPyPacked",
     sizeof(SwigPyPacked),
     0,
     Py_TPFLAGS_DEFAULT,
--- a/Lib/python/pyruntime.swg
+++ b/Lib/python/pyruntime.swg
@@ -30,6 +30,7 @@
 #else
 # include <Python.h>
 #endif
+#include <structmember.h>
 
 #if __GNUC__ >= 7
 #pragma GCC diagnostic pop
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -4370,9 +4370,9 @@ public:
     Printf(f, "  }\n");
     Printv(f, "#endif\n", NIL);
     Printf(f, "};\n\n");
-    Printf(f, "static PyTypeObject* create_%s_type(PyTypeObject* type, PyTypeObject** bases, PyObject* dict) {\n", templ);
-    Printv(f, "  PyObject* tuple_bases;\n", NIL);
-    Printf(f, "  PyTypeObject* pytype = (PyTypeObject*)&%s_type;\n", templ);
+    Printf(f, "static PyTypeObject *create_%s_type(PyTypeObject *type, PyTypeObject **bases, PyObject *dict) {\n", templ);
+    Printv(f, "  PyObject *tuple_bases;\n", NIL);
+    Printf(f, "  PyTypeObject *pytype = (PyTypeObject *)&%s_type;\n", templ);
     Printf(f, "  pytype->tp_dict = dict;\n");
     Printv(f, "  SwigPyBuiltin_SetMetaType(pytype, type);\n", NIL);
     static String *tp_new = NewString("PyType_GenericNew");
@@ -4390,9 +4390,9 @@ public:
 
     Printv(f, "#else\n", NIL);
 
-    Printf(f, "static PyTypeObject* create_%s_type(PyTypeObject* type, PyTypeObject** bases, PyObject* dict) {\n", templ);
+    Printf(f, "static PyTypeObject *create_%s_type(PyTypeObject *type, PyTypeObject **bases, PyObject *dict) {\n", templ);
     Printf(f, "  PyMemberDef members[] = {\n");
-    Printf(f, "  {(char*)\"__dictoffset__\", T_PYSSIZET, %s, READONLY, NULL},\n", getSlot(n, "feature:python:tp_dictoffset", tp_dictoffset_default));
+    Printf(f, "  {(char *)\"__dictoffset__\", T_PYSSIZET, %s, READONLY, NULL},\n", getSlot(n, "feature:python:tp_dictoffset", tp_dictoffset_default));
     Printf(f, "  {NULL, 0, 0, 0, NULL}};\n");
     Printf(f, "  PyType_Slot slots[] = {\n");
     printSlot2(f, getSlot(n, "feature:python:tp_init", tp_init), "tp_init", "initproc");
@@ -4486,8 +4486,8 @@ public:
     Printf(f, "  %s,\n", getSlot(n, "feature:python:tp_flags", tp_flags_py3), "tp_flags");
     Printf(f, "  slots\n");
     Printf(f, "  };\n");
-    Printv(f, "  PyObject* tuple_bases = SwigPyBuiltin_InitBases(bases);\n", NIL);
-    Printf(f, "  PyTypeObject* pytype = (PyTypeObject*)PyType_FromSpecWithBases(&spec, tuple_bases);\n");
+    Printv(f, "  PyObject *tuple_bases = SwigPyBuiltin_InitBases(bases);\n", NIL);
+    Printf(f, "  PyTypeObject *pytype = (PyTypeObject*)PyType_FromSpecWithBases(&spec, tuple_bases);\n");
     Printf(f, "  PyDict_Merge(pytype->tp_dict, dict, 1);\n");
     Printv(f, "  SwigPyBuiltin_SetMetaType(pytype, type);\n", NIL);
     Printf(f, "  PyType_Modified(pytype);\n");
@@ -4524,7 +4524,7 @@ public:
     Printv(f_init, "    }\n", NIL);
     Printf(f_init, "    SwigPyBuiltin_%s_clientdata.pytype = builtin_pytype;\n", mname);
     if (GetFlag(n, "feature:implicitconv")) {
-      Printf(f_init, "    SwigPyBuiltin_%s_clientdata.klass = (PyObject*)builtin_pytype;\n", mname);
+      Printf(f_init, "    SwigPyBuiltin_%s_clientdata.klass = (PyObject *)builtin_pytype;\n", mname);
     }
     Printv(f_init, "    SWIG_Py_INCREF((PyObject *)builtin_pytype);\n", NIL);
     Printf(f_init, "    PyModule_AddObject(m, \"%s\", (PyObject *)builtin_pytype);\n", symname);
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -4379,7 +4379,7 @@ public:
     Printf(f, "  pytype->tp_new = %s;\n", getSlot(n, "feature:python:tp_new", tp_new));
     Printv(f, "  tuple_bases = SwigPyBuiltin_InitBases(bases);\n", NIL);
     Printv(f, "  pytype->tp_base = bases[0];\n", NIL);
-    Printv(f, "  Py_INCREF(pytype->tp_base);\n", NIL);
+    Printv(f, "  SWIG_Py_INCREF(pytype->tp_base);\n", NIL);
     Printv(f, "  pytype->tp_bases = tuple_bases;\n", NIL);
     Printv(f, "  if (PyType_Ready(pytype) < 0) {\n", NIL);
     Printf(f, "    PyErr_SetString(PyExc_TypeError, \"Could not create type '%s'.\");\n", symname);
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -58,6 +58,7 @@ static String *builtin_tp_init = 0;
 static String *builtin_methods = 0;
 static String *builtin_default_unref = 0;
 static String *builtin_closures_code = 0;
+static String *f_varlinks = 0;
 
 static String *methods;
 static String *methods_proxydocs;
@@ -555,6 +556,7 @@ public:
     class_members = NewHash();
     builtin_methods = NewString("");
     builtin_default_unref = NewString("delete $self;");
+    f_varlinks = NewString("");
 
     if (builtin) {
       f_builtins = NewString("");
@@ -836,6 +838,8 @@ public:
 
     initialize_threads(f_init);
 
+    Dump(f_varlinks, f_init);
+
     Printf(f_init, "#if PY_VERSION_HEX >= 0x03000000\n");
     Printf(f_init, "  return m;\n");
     Printf(f_init, "#else\n");
@@ -915,6 +919,7 @@ public:
     Delete(f_directors_h);
     Delete(f_runtime);
     Delete(f_begin);
+    Delete(f_varlinks);
 
     return SWIG_OK;
   }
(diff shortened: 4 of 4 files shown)


wsfulton:
CI has seg faults with this patch.

clintonstimpson:
I'm able to reproduce the segfaults seen in CI.

So, I thought I'd try running with a debug build of python (3.10) so I could see more of what is happening.
I'm hitting an assert:
python3: Objects/typeobject.c:3813: _PyType_Lookup: Assertion `_PyType_HasFeature(type, Py_TPFLAGS_VALID_VERSION_TAG)' failed.
I also checked the swig/Examples/python/reference example on the master branch with my debug build of python and it runs without issues.

I don't have experience with this type of problem.  Any ideas?

ojwb:
@clintonstimpson Those segfaults need addressing.  Seems to be with `-builtin`, e.g. https://github.com/swig/swig/actions/runs/3012183221/jobs/4840023292

clintonstimpson:
Yeah, I've been able to reproduce the segfaults.  With a debug build of Python, it looks like the issue comes from a static type being incorrectly marked as a heap type.
I'm still looking at this (slowly), and have gone down the path of more fully using PyType_FromSpec, even with -builtin.  Maybe folks will like using -builtin with the python limited api - if that's possible.
With that, I have a handful of tests left to fix and hopefully it'll work.
One thing I uncovered is the li_std_[w]stream tests define "cin" after the basic_stream type is defined, but before basic_stream.  With dynamic types defined at runtime, this segfaults.  I'm not sure how to fix that such that cin is defined after basic_stream is initialized, but maybe it can be fixed separately.

clintonstimpson:
I've pushed another set of changes ... it doesn't yet pass the test suite with -builtin.  But, I get a different set of failures.
Also, this set of changes makes progress towards support for Py_LIMITED_API with the -builtin option.

Can I get feedback/help on 2 things?
1. General feedback on the code changes and approach being taken.  I'm new to SWIG contributions.
2. Help on the li_std_stream.cpptest and li_std_wstream.cpptest failures.  These code changes cause a segfault when one attempts to use non-ready types.  Before this code change, using the non-ready types did not segfault.  I assume the swig generated code needs to re-order initialization of "cin", "cout" streams.

There's also cpp_static.cpptest, struct_value.cpptest, li_std_vector_back_reference.cpptest and python_nondynamic.cpptest which fail with -builtin.  I'm not yet sure why those are failing.

Thanks!

clintonstimpson:
@ojwb, any idea why the generated code seems a bit out of order?  It seems I'm blocked by this when I'm trying to use a dynamically created PyTypeObject.
```
  /* type 'std::basic_string< char >' */
  builtin_pytype = (PyTypeObject *)&SwigPyBuiltin__std__basic_stringT_char_t_type;
  ...
  if (PyType_Ready(builtin_pytype) < 0) {
  ...

  SWIG_addvarlink(globals, "cin", Swig_var_cin_get, Swig_var_cin_set);
  PyDict_SetItemString(md, "cin", PyObject_GetAttrString(globals, "cin"));
  SwigPyBuiltin_AddPublicSymbol(public_interface, "cin");
  SWIG_addvarlink(globals, "cout", Swig_var_cout_get, Swig_var_cout_set);
  PyDict_SetItemString(md, "cout", PyObject_GetAttrString(globals, "cout"));
  SwigPyBuiltin_AddPublicSymbol(public_interface, "cout");

  /* type 'std::basic_istream< char >' */
  builtin_pytype = (PyTypeObject *)&SwigPyBuiltin__std__basic_istreamT_char_t_type;
  ...
  if (PyType_Ready(builtin_pytype) < 0) {
  ...
```

I'm hoping to have `SWIG_addvarlink` for `"cin"` and `"cout"` called after the `PyTypeObject` is created/initialized for `std::basic_istream<char>`.

ojwb:
If you need to enforce an ordering, you probably need to write the varlinks that need to come later to a temporary `String *f_varlinks` (the first argument to `Printf`, etc can be a `String *` instead of a `File *` so it's pretty easy to change that), then `Dump(f_varlinks, f_init);` late on in processing.

clintonstimpson:
I'm glad to see CI passing.
Along with the CI testing, I tested python 3.10 vs 3.8 and 3.7 with -builtin.  I think CI only covered Python 3.8 with -builtin.
Python 3.9 added a way to set `dictoffset` which was needed for some tests.

I've split this pull request into 2 commits, one for changes to initialization to allow for dynamic types, and the second for actually creating dynamic types.

ojwb:
@clintonstimpson `python_runtime_data.multicpptest` seems to be failing on appveyor now.

clintonstimpson:
@ojwb , the `python_runtime_data.multicpptest` is now fixed.  It was previously failing due to not having a null terminated array.

Lib/python/pyruntime.swg
```diff
@@ -30,6 +30,7 @@
 #else
 # include <Python.h>
 #endif
+#include "structmember.h"
```
ojwb, line 33: Seems to be a header from the Python API, so that should be: `#include <structmember.h>`

Source/Modules/python.cxx
```diff
@@ -4345,6 +4352,146 @@ class PYTHON:public Language {
...
+    Printf(f, "  return pytype;\n");
+    Printf(f, "}\n\n");
+
+    Printv(f, "#else\n", NIL);
+
+    Printf(f, "static PyTypeObject* create_%s_type(PyTypeObject* type, PyTypeObject** bases, PyObject* dict) {\n", templ); 
+    Printf(f, "  PyMemberDef members[] = {\n");
+    Printf(f, "  {(char*)\"__dictoffset__\", T_PYSSIZET, %s, READONLY, NULL},\n", getSlot(n, "feature:python:tp_dictoffset", tp_dictoffset_default));
+    //Printf(f, "  {(char*)\"__weaklistoffset__\", T_PYSSIZET, %s, READONLY, NULL},\n", getSlot(n, "feature:python:tp_weaklistoffset"));
+    //Printf(f, "  {(char*)\"__vectorcalloffset__\", T_PYSSIZET, %s, READONLY, NULL},\n", getSlot(n, "feature:python:tp_vectorcall_offset"));
```
ojwb, line 4379: All these commented out lines of code don't seem helpful - if they need resolving, it'd be best to actually resolve them, and failing that at least document with code comments why they haven't been resolved and any implications of that.  If there's nothing to resolve it doesn't seem useful to include them.

Source/Modules/python.cxx
```diff
@@ -4345,6 +4352,146 @@ class PYTHON:public Language {
...
+    printSlot2(f, getSlot(n, "feature:python:nb_inplace_and"), "nb_inplace_and", "binaryfunc");
+    printSlot2(f, getSlot(n, "feature:python:nb_inplace_xor"), "nb_inplace_xor", "binaryfunc");
+    printSlot2(f, getSlot(n, "feature:python:nb_inplace_or"), "nb_inplace_or", "binaryfunc");
+    printSlot2(f, getSlot(n, "feature:python:nb_floor_divide"), "nb_floor_divide", "binaryfunc");
+    printSlot2(f, getSlot(n, "feature:python:nb_divide"), "nb_true_divide", "binaryfunc");
+    printSlot2(f, getSlot(n, "feature:python:nb_inplace_floor_divide"), "nb_inplace_floor_divide", "binaryfunc");
+    printSlot2(f, getSlot(n, "feature:python:nb_inplace_divide"), "nb_inplace_true_divide", "binaryfunc");
+    printSlot2(f, getSlot(n, "feature:python:nb_index"), "nb_index", "unaryfunc");
+    Printv(f, "#if PY_VERSION_HEX >= 0x03050000\n", NIL);
+    Printv(f, "#endif\n", NIL);
```
ojwb, line 4456: This seems to generate an empty `#if`!

Lib/python/pyruntime.swg
clintonstimpson: Fixed.

Source/Modules/python.cxx
clintonstimpson: Right, they aren't helpful.  I've removed them because they only make sense if the C-struct has those members.  The only member in the C-struct right now is a dictionary.

Source/Modules/python.cxx
clintonstimpson: I've removed it.

clintonstimpson:
FYI, the last push simply reverts a white space change to avoid a potential merge conflict with another branch.

clintonstimpson:
@ojwb @wsfulton, any thoughts on this one?  It makes some pretty fundamental changes to support the Limited API, and also some progress toward supporting -builtin with the limited api.  Another possible gain here is supporting newer versions of python with fewer ifdefs.

ojwb:
It certainly looks plausible to me, but I was hoping someone more familiar with Python's C API would merge it, or at least OK it.

encukou:
Hi! I help maintain C API in CPython.
I'll take a break from code over the holidays, but I'll be happy to review this in January.

Lib/python/builtin.swg
```diff
@@ -403,16 +403,14 @@ SwigPyBuiltin_InitBases (PyTypeObject *type, PyTypeObject **bases) {
     bases[0] = SwigPyObject_type();
     bases[1] = NULL;
   }
-  type->tp_base = bases[0];
-  Py_INCREF((PyObject *)bases[0]);
   for (b = bases; *b != NULL; ++b)
     ++base_count;
   tuple = PyTuple_New(base_count);
```
encukou, line 408: Like almost any Python API, `PyTuple_New` can fail. But it seems SWIG generally doesn't handle errors, is that right?

Source/Modules/python.cxx
```diff
@@ -4345,6 +4352,133 @@ class PYTHON:public Language {
...
+    printSlot2(f, getSlot(n, "feature:python:sq_inplace_repeat"), "sq_inplace_repeat", "ssizeargfunc");
+
+    Printf(f, "  {Py_tp_members, members},\n");
+    Printf(f, "  { 0, NULL}\n");
+    Printf(f, "  };\n");
+    Printf(f, "PyType_Spec spec = {\n");
+    Printf(f, "  %s,\n", quoted_symname);
+    Printf(f, "  sizeof(SwigPyObject),\n");
+    Printf(f, "  0,\n");
+    Printf(f, "  %s,\n", getSlot(n, "feature:python:tp_flags", tp_flags_py3), "tp_flags");
```
encukou, line 4468: Here too you might need  `Py_TPFLAGS_IMMUTABLETYPE` to make the *type object* immutable.

Source/Modules/python.cxx
```diff
@@ -4112,6 +4118,7 @@ class PYTHON:public Language {
     String *tp_as_mapping = NewStringf("&%s_type.as_mapping", templ);
     String *tp_as_buffer = NewStringf("&%s_type.as_buffer", templ);
```
encukou, line 4119: The `tp_as_*` shouldn't be needed for heap types, they can go in the `#ifndef`.

Source/Modules/python.cxx
```diff
@@ -4345,6 +4352,133 @@ class PYTHON:public Language {
...
+    Printf(f, "  };\n");
+    Printf(f, "PyType_Spec spec = {\n");
+    Printf(f, "  %s,\n", quoted_symname);
+    Printf(f, "  sizeof(SwigPyObject),\n");
+    Printf(f, "  0,\n");
+    Printf(f, "  %s,\n", getSlot(n, "feature:python:tp_flags", tp_flags_py3), "tp_flags");
+    Printf(f, "  slots\n");
+    Printf(f, "  };\n");
+    Printv(f, "  PyObject* tuple_bases = SwigPyBuiltin_InitBases(bases);\n", NIL);
+    Printf(f, "  PyTypeObject* pytype = (PyTypeObject*)PyType_FromSpecWithBases(&spec, tuple_bases);\n");
```
encukou, line 4472: And here I'm also not sure about the expected lifetime of the new type.
Does `SWIGINTERN` mean it's only created once per process?

Lib/python/pyrun.swg
```diff
@@ -335,9 +335,19 @@ swig_varlink_setattr(PyObject *o, char *n, PyObject *p) {
   return res;
 }
 
+#if PY_VERSION_HEX >= 0x03090000 || (!defined(SWIGPYTHON_BUILTIN) && PY_VERSION_HEX >= 0x03040000)
```
encukou, line 338: This might need to be a user-level setting, since e.g. a custom `tp_traverse` function would need to be adjusted for heap types.

encukou (commented):
Here's what I see.
I don't know much about SWIG, so I'd welcome some clarifications.

Lib/python/pyrun.swg
```diff
@@ -402,12 +412,38 @@ swig_varlink_type(void) {
...
+      { 0, NULL }
+    };
+    PyType_Spec spec = {
+      "swigvarlink",
+      sizeof(swig_varlinkobject),
+      0,
+      0,
+      slots
+    };
+    py3varlink_type = (PyTypeObject*)PyType_FromSpec(&spec);
```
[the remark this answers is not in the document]
ojwb: I don't think there's a formal policy, but I'm fairly sure we have similar cases elsewhere.  Probably the worst consequence is potentially showing if you run a leak checker which runs at process exit.  If it's easy to release maybe it's worth doing, but it doesn't seem worth adding complexity to track (to me at least).

Lib/python/builtin.swg
ojwb: The generated wrappers ideally should handle errors, but there are certainly cases where currently they don't.  A particular case that comes to mind is there's not much checking for `malloc()` failing, which really isn't great but it's also a situation where it's often hard to gracefully recover - recovery likely means further calls to the target language API which generally will end up trying to allocate memory, which is likely to also fail unless the original failed allocation was large (and calls to `malloc()` in the generated wrapper code will generally be for structures, etc and so small).  E.g. see #277.

It looks like there's already missing checking here so good to highlight but could reasonably be addressed separately.

Source/Modules/python.cxx
ojwb: It's a macro which is essentially defined to `static` (possibly plus compiler-specific stuff to suppress unused warnings) if not already defined.  There seems to be no documentation of why it exists beyond `/* internal SWIG method */` and the VCS history doesn't help, but I guess it's useful to be able to override it in some situation.  Or maybe someone just saw the "define if not already defined" pattern for existing macros which are useful to override and copied it for one which isn't.

So you'll get one per SWIG-wrapped module loaded rather than one per process (unless you pre-defined `SWIGINTERN` to something to change that).

ojwb:
> The easiest way to get a new interpreter is with a [Py_Finalize](https://docs.python.org/3/c-api/init.html#c.Py_FinalizeEx)/[Py_Initialize](https://docs.python.org/3/c-api/init.html#c.Py_Initialize) “power cycle”. Reusing the old type can lead to subtle bugs.

Perhaps this is the cause of #2114.

Lib/python/pyrun.swg
```diff
@@ -335,9 +335,19 @@ swig_varlink_setattr(PyObject *o, char *n, PyObject *p) {
   return res;
 }
 
+#if PY_VERSION_HEX >= 0x03090000 || (!defined(SWIGPYTHON_BUILTIN) && PY_VERSION_HEX >= 0x03040000)
+#define SWIG_HEAPTYPES
+#endif
+
+#ifdef SWIG_HEAPTYPES
+static PyTypeObject* py3varlink_type = NULL;
```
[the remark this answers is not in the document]
clintonstimpson: Yeah, it could be moved inside.

Lib/python/pyrun.swg
```diff
@@ -402,12 +412,38 @@ swig_varlink_type(void) {
...
+      { Py_tp_setattr, (void*)swig_varlink_setattr },
+      { Py_tp_str, (void*)swig_varlink_str },
+      { Py_tp_doc, (void*)varlink__doc__ },
+      { 0, NULL }
+    };
+    PyType_Spec spec = {
+      "swigvarlink",
+      sizeof(swig_varlinkobject),
+      0,
+      0,
```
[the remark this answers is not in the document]
clintonstimpson: Thanks for pointing that out!

Source/Modules/python.cxx
clintonstimpson: Ok.  Whether this class was immutable was something I considered during these changes.  It seems there is an opportunity to simplify the code by not having it immutable so it can store static data, but I figured it might be better to keep this change set minimal.

Lib/python/pyrun.swg
clintonstimpson: Isn't tp_traverse intended for C structures?  Here, the only data stored by the C struct is a dict, and that struct is not customizable by the user.

Source/Modules/python.cxx
clintonstimpson: The lifetime of this type depends on the users script.  Anytime an c object is constructed in python, this type is used.

wsfulton:
Couple of high level questions, before I try to digest this
1. What's the purpose of SWIG_HEAPTYPES? We've got a huge matrix of options, which I'd like to see reduced, not increased!
2. A simple overview of what `PyType_FromSpec` replaces would be helpful. Is the motivation for the merge request just as the title implies - more portable? In what way, or is it just Py_Limited_API friendly? Just point me at some docs if need be. I've got my head deep in some other SWIG internals and C++ partial specialization corner cases atm, so sorry, I havn't done much research on this area.

clintonstimpson:
The first objective here is to enable compiling with `Py_LIMITED_API`.
A second objective is to be more portable starting with Python 3.9.  Portability is gained when using `PyType_FromSpec()`, because no code changes are necessary for defining types when attempting to add support for Python 3.10.  Before this merge request, adding support for Python 3.10 requires adding `#ifdefs` when defining the static `PyTypeObject`.
I'm choosing 3.9 because some degree of support for -builtin and `Py_LIMITED_API` seems to be enabled at that point.
Hopefully in the long run, the matrix of of options will be reduced.

I guess this merge request could include going back to the static type definitions, and remove ifdefs, which would no longer be needed.

Generally speaking, `PyType_FromSpec()` is replacing the static declarations of `PyTypeObject`.

wsfulton:
Can you please resolve the conflicts given the stable ABI changes that have just been merged into master. I'm sorry that we've just run out of time for this to make it into the swig-4.2 release.

clintonstimpson:
> Can you please resolve the conflicts given the stable ABI changes that have just been merged into master. I'm sorry that we've just run out of time for this to make it into the swig-4.2 release.

Sure thing, I've just rebased this pull request.

ojwb:
@wsfulton This seems to be waiting on you to review - the submitter rebased and resolved the conflicts as you requested.

wsfulton:
So sorry, I should have got to this earlier. I'm afraid another rebase is required as there are a number of conflicts, and I'm a bit short on time to spend  resolving them. I think they are mostly down to further `Py_LIMITED_API` improvements that have gone into the code base since you last rebased. It might even be the case that there is very little in this patch adding in `Py_LIMITED_API` fixes.

Some questions:

1. #include <structmember.h> ... this header is deprecated in 3.12. I think you need to add a macro around this inclusion so that it is only included for 3.11 and earlier versions.
2. Do you have references for when PyType_FromSpec was first made available and when it became 'production ready'. I don't really follow your choice for using it builtin types for 3.4 and later but only 3.9 and later for non-builtin types.
3. If I look at the generated code for there seems to be a double `Py_INCREF`. Can you check this please, for example, the Examples/python/class using `-builtin` has:
```c++
  SwigPyBuiltin__Shape_clientdata.pytype = builtin_pytype;
  Py_INCREF(builtin_pytype);
  Py_INCREF(builtin_pytype);
```
Possibly related to the removed `PyINCREF` in `SwigPyBuiltin_InitBases`?? It's late here, I'm not sure!!

Changes required please:
1. Please don't begin static/global variables with py, that's reserved for Python. Instead begin them with `Swig_`.
2. Please use code style `PyObject *`, `char *` instead of `PyObject*`, `char*` etc where you've made changes.

wsfulton:
@clintonstimpson any chance you can respond in the next couple of days as there is a chance we can get this into swig-4.3. I can probably figure out the rebase if you can respond to the other queries.

clintonstimpson:
@wsfulton.  Ok, I'd be interested to see how the rebase works and see what is already included in the code base, now that you mention more `Py_LIMITED_API` improvements have gone in.
To answer your questions:

1. Ok, I wasn't aware <structmember.h> was deprecated in 3.12.  A macro sounds good, and hopefully there is a replacement in 3.12.
2. PyType_FromSpec was first made available in Python 3.2, the first version to introduce the `Py_LIMITED_API`.  The logic is actually to use PyType_FromSpec if using Python 3.4 and greater, or use PyType_FromSpec if using -builtin with Python 3.9 or greater.  There is no reversing of that addition with a newer version of Python.  It has been a while, but my memort is that Python 3.9 introduced the __dictoffset__ and Py_tp_methods needed with the -builtin option.
3. Yeah, the double `Py_INCREF` looks like a bug.

encukou:
Don't worry about `structmember.h`, see the [code](https://github.com/python/cpython/blob/78aeb38f7d880a340295214abc4f7e77ffdad509/Include/structmember.h#L4):
```
 * This header is deprecated: new code should not use stuff from here.
 * New definitions are in descrobject.h.
 *
 * However, there's nothing wrong with old code continuing to use it,
 * and there's not much maintenance overhead in maintaining a few aliases.
 * So, don't be too eager to convert old code.
```
or the [whatsnew docs](https://docs.python.org/3/whatsnew/3.12.html):
> Your old code can stay unchanged, unless the extra include and non-namespaced macros bother you greatly.

ojwb:
Stopping including `structmember.h` still seems useful as some of the things it defines are rather common generic names which could collide with an API being wrapped (`READONLY`  is probably the most problematic).  That could probably happen as a separate change after 4.3.0 though.

wsfulton:
I've added a couple of polishing off commits ending in 54d7ebebced7246e526bbee26417710973fb4a48. There is a crash in the last test which needs looking at by someone here please, but I'm signing out for the day now.

clintonstimpson:
> I've added a couple of polishing off commits ending in [54d7ebe](https://github.com/swig/swig/commit/54d7ebebced7246e526bbee26417710973fb4a48). There is a crash in the last test in Github CI which needs looking at by someone here please, but I'm signing out for the day now.

Thanks for running CI on it.  I debugged it a bit and found that `PyType_FromSpecWithBases` is failing and it is failing because the base types are not "ready" with `PyType_Ready` returning false.

The first commit is the one which re-arranged code to set up bases.  If that got dropped during the rebase, that might be why.  Hmm..  It doesn't look the same as the commit which you mentioned was cherry-picked, and it also doesn't look it got dropped in the rebase.

I can keep looking at this, but will probably not be quick.

wsfulton:
Looking better, thanks. I see
```
/home/runner/work/swig/swig/build/build/Examples/test-suite/python/../../../../../Examples/test-suite/python/member_pointer_runme.py:26: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute
  memberPtr = cvar.perimetervar
```
in the builds such as https://github.com/swig/swig/actions/runs/11076298285/job/30779132295. No such warning beforehand, so can you see what's needed to remove this?

wsfulton:
> 2. PyType_FromSpec was first made available in Python 3.2, the first version to introduce the Py_LIMITED_API. The logic is actually to use PyType_FromSpec if using Python 3.4 and greater, or use PyType_FromSpec if using -builtin with Python 3.9 or greater. There is no reversing of that addition with a newer version of Python. It has been a while, but my memory is that Python 3.9 introduced the `__dictoffset__` and `Py_tp_` methods needed with the -builtin option.

There are three failures in the test-suite if using `SWIG_HEAPTYPES` with python-3.8 and lower, so the patch requiring `-builtin` and 3.9 and higher seems to be correct.

The test-suite results are identical with or without defining `SWIG_HEAPTYPES` with python-3.3, so the patch requiring 3.4 and higher (non-builtin) could be changed to use 3.3 and higher instead of 3.4 and higher. Any reason why you chose 3.4 as the minimum instead of 3.3? I didn't test 3.2 as 3.3 is the minimum supported version by SWIG.

clintonstimpson:
I don't recall a good reason to require 3.4 or higher.  Moving that down to 3.3 sounds good to me.

clintonstimpson:
> Looking better, thanks. I see
> 
> ```
> /home/runner/work/swig/swig/build/build/Examples/test-suite/python/../../../../../Examples/test-suite/python/member_pointer_runme.py:26: DeprecationWarning: builtin type SwigPyPacked has no __module__ attribute
>   memberPtr = cvar.perimetervar
> ```
> 
> in the builds such as https://github.com/swig/swig/actions/runs/11076298285/job/30779132295. No such warning beforehand, so can you see what's needed to remove this?

Yeah, the name would need to be changed from `"SwigPyObject"` to `"<modulename>.SwigPyObject"` before calling `PyType_FromSpec`.  When the module name is included, the `__module__` attribute gets filled in.

clintonstimpson:
> Yeah, the name would need to be changed from `"SwigPyObject"` to `"<modulename>.SwigPyObject"` before calling `PyType_FromSpec`. When the module name is included, the `__module__` attribute gets filled in.
> 
> I thought about including that in this pull request, but ran into issues with it. I'm not remember what those issues were.
> 
> I assume you're also aware of #2881 which is related to that.

Ok, I think I'm seeing why it didn't work for me before.  There are other occurrences of the string `"SwigPyPacked"` which I failed to update at the same time.  A quick modification on to change the type name to `"_member_pointer.SwigPyPacked"` and other occurrences fixes the warning and also keeps the member_pointer test passing.

wsfulton:
swig/py3-typefromspec is now clintonstimpson/py3-typefromspec plus 16c22f661e3b17598a2982076c72a67d2038879b and 98e3b6c44ee0b8ac67483bfff0e193e9162f82b9. I can see seg faults locally with python-3.9 which is the oldest version that the heap types plus builtin is turned on for. However, there are seg faults as you can see in CI https://github.com/swig/swig/actions/runs/11082374271 using python>3.9 and builtin. 

The non-builtin looks 100% to me now. It's just builtin that needs attention. We had earmarked a releasing a new version of SWIG today - I had in mind a release candidate. The builtin changes are technically major, so if we include them, they do need to go into the release candidate. I don't mind delaying by a couple of days, but needless to say, the current problems will need fixing quickly.

wsfulton:
Looking at the seg fault for the constructor_copy_non_const testcase in gdb shows just a python interpreter stack trace, so hard to debug. I ran it through valgrind and got a clue:
```
==69729== Invalid read of size 4
==69729==    at 0x503040: PyObject_GC_Del (in /usr/bin/python3.9)
==69729==    by 0x5FB46D: ??? (in /usr/bin/python3.9)
==69729==    by 0x43A67C: ??? (in /usr/bin/python3.9)
==69729==    by 0x575AA96: SwigPyBuiltin__CCDerived_type_create (constructor_copy_non_const_wrap.cxx:6128)
==69729==    by 0x575E6BB: PyInit__constructor_copy_non_const (constructor_copy_non_const_wrap.cxx:9608)
==69729==    by 0x611731: ??? (in /usr/bin/python3.9)
==69729==    by 0x5201E8: ??? (in /usr/bin/python3.9)
==69729==    by 0x514673: _PyEval_EvalFrameDefault (in /usr/bin/python3.9)
```
but nothing has jumped out at me yet as I don't have a debug build of this python interpreter to hand.

wsfulton:
While heap types work for builtin and python>=3.12, it isn't clear what the problem is for earlier versions. I think it is safer to not try and use heap types for builtin until the problems are understood. I've just committed 0d70a7ba0ec5445b8f42f31c5b1a9121430837f4 to only use heap types for non-builtin and expect this to pass CI and I'd be comfortable merging this to master for the release. I'd have preferred builtin and heap types also worked though.

03a001a5ce4b310145f688214d7342f6fd309203 reverts the changes to fix the `__module__` attribute missing for SwigPyPacked. More work is also needed in this area to fix these warnings as this breaks `Py_LIMITED_API` builds, see for example https://github.com/swig/swig/actions/runs/11082374271/job/30795306293.

In summary there are two things to fix:
1. builtin and heap types
2. `__module__` attribute missing warnings (issue #2881)

clintonstimpson:
Thanks @wsfulton .  I agree it'll be good to wait till those 2 remaining issues are understood before enabling built-in with heap types.

clintonstimpson:
Do you want to use other issues and pull requests to capture remaining work, and close this one?

wsfulton:
Yes, let's create a new issue to track heap types and Py_LIMITED_API for builtin. I came across https://doc.qt.io/qtforpython-6/developer/limited_api.html which will be a useful reference. That mentions the `__module__` deprecation warning that we have struggled with and so it all seems related.

Issue #2357, opened 2022-09-09
Passing Umlauts from string (C#) to std::string (C++) fails on Windows


emmenlau:
I've just started my first work with swig and find it absolutely awesome! Great work!

Now I've found a small bottleneck. I create in C# a `string` like this: `string vSendString = "Hello World, äöü, ß, 是.";`. When I call a C++ method with this as the argument, in C++ I receive a `std::string` with slightly different contents, so the comparison fails:
```
if (aString != "Hello World, äöü, ß, 是.") {
    throw(std::runtime_error("The received string '" + aString + "' does not match the expected string 'Hello World, äöü, ß, 是.'"));
}
```

I'm using VSCode to edit both source code files, and I think the source code should hold these strings correctly(?) Also the compiler should therefore see them to be the same(?)

Am I doing something wrong?

I found the following discussion but that seems not such a great solution, is it required? https://stackoverflow.com/a/25966756/7200132

wsfulton:
You'll have to work out what unicode you are using and marshall appropriately. C++ std::wstring rather than std::string is normally used for Unicode strings and unfortunately very platform specific.  #1233 has some Unicode improvements for std::wstring which will be in the 4.1.0 release. Maybe you could utilise the std::wstring typemaps if your std::string contains Unicode.

emmenlau:
Thanks a lot @wsfulton for your help! Admittedly I did not put much thought into the encoding because my simple test was starting from literal characters typed into the file.

We are using UTF8 on all platforms. We need to support Linux, Windows and Mac. Thanks to the very nice http://utf8everywhere.org/ page I was silently assuming that UTF8 is becoming more of a standard, but I know `wstring` and UTF16 are also commonly used.

In the past I used something like the following to transfer strings between C# and C++ on all platforms (Linux, Windows, Mac):
```
        /// <summary>
        /// Encode a string into an UTF8 byte array. The UTF8-encoded byte
        /// array is compatible with native code, and has the same
        /// representation as a std::string (with char character type).
        /// </summary>
        internal static byte[] utf8Encode(string aMessage)
        {
            return System.Text.Encoding.UTF8.GetBytes(aMessage);
        }

        /// <summary>
        /// Decode the first <paramref name="aMessageLength"/> bytes of an UTF8
        /// byte array into a string. The UTF8-encoded byte array is compatible
        /// with native code, and has the same representation as a std::string
        /// (with char character type).
        /// </summary>
[...]

PR #2364, opened 2022-09-13
Lib/csharp/csharphead.swg: Marshall strings as UTF-8

emmenlau:
This PR is only a DRAFT to discuss ideas to solve #2357

### DIFF
--- a/Lib/csharp/csharphead.swg
+++ b/Lib/csharp/csharphead.swg
@@ -13,7 +13,7 @@
 
 #if !defined(SWIG_CSHARP_NO_EXCEPTION_HELPER)
 %insert(runtime) %{
-/* Support for throwing C# exceptions from C/C++. There are two types: 
+/* Support for throwing C# exceptions from C/C++. There are two types:
  * Exceptions that take a message and ArgumentExceptions that take a message and a parameter name. */
 typedef enum {
   SWIG_CSharpApplicationException,
@@ -87,19 +87,19 @@ static void SWIGUNUSED SWIG_CSharpSetPendingExceptionArgument(SWIG_CSharpExcepti
 
 %insert(runtime) %{
 #ifdef __cplusplus
-extern "C" 
+extern "C"
 #endif
 SWIGEXPORT void SWIGSTDCALL SWIGRegisterExceptionCallbacks_$module(
                                                 SWIG_CSharpExceptionCallback_t applicationCallback,
                                                 SWIG_CSharpExceptionCallback_t arithmeticCallback,
-                                                SWIG_CSharpExceptionCallback_t divideByZeroCallback, 
-                                                SWIG_CSharpExceptionCallback_t indexOutOfRangeCallback, 
+                                                SWIG_CSharpExceptionCallback_t divideByZeroCallback,
+                                                SWIG_CSharpExceptionCallback_t indexOutOfRangeCallback,
                                                 SWIG_CSharpExceptionCallback_t invalidCastCallback,
                                                 SWIG_CSharpExceptionCallback_t invalidOperationCallback,
                                                 SWIG_CSharpExceptionCallback_t ioCallback,
                                                 SWIG_CSharpExceptionCallback_t nullReferenceCallback,
-                                                SWIG_CSharpExceptionCallback_t outOfMemoryCallback, 
-                                                SWIG_CSharpExceptionCallback_t overflowCallback, 
+                                                SWIG_CSharpExceptionCallback_t outOfMemoryCallback,
+                                                SWIG_CSharpExceptionCallback_t overflowCallback,
                                                 SWIG_CSharpExceptionCallback_t systemCallback) {
   SWIG_csharp_exceptions[SWIG_CSharpApplicationException].callback = applicationCallback;
   SWIG_csharp_exceptions[SWIG_CSharpArithmeticException].callback = arithmeticCallback;
@@ -115,7 +115,7 @@ SWIGEXPORT void SWIGSTDCALL SWIGRegisterExceptionCallbacks_$module(
 }
 
 #ifdef __cplusplus
-extern "C" 
+extern "C"
 #endif
 SWIGEXPORT void SWIGSTDCALL SWIGRegisterExceptionArgumentCallbacks_$module(
                                                 SWIG_CSharpExceptionArgumentCallback_t argumentCallback,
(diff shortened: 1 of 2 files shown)


stevebaxter:
Nice solution! For this to work properly in all cases, you also need to change the `SWIGStringHelper` delegate:

```
%insert(runtime) %{
/* Callback for returning strings to C# without leaking memory */
typedef char * (SWIGSTDCALL* SWIG_CSharpStringHelperCallback)(const char *);
static SWIG_CSharpStringHelperCallback SWIG_csharp_string_callback = NULL;
%}

%pragma(csharp) imclasscode=%{
  protected class SWIGStringHelper {

    [return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]
    public delegate string SWIGStringDelegate([global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)] string message);

    static SWIGStringDelegate stringDelegate = new SWIGStringDelegate(CreateString);

    [global::System.Runtime.InteropServices.DllImport("$dllimport", EntryPoint="SWIGRegisterStringCallback_$module")]
    public static extern void SWIGRegisterStringCallback_$module(SWIGStringDelegate stringDelegate);

    static string CreateString(string cString) {
      return cString;
    }

    static SWIGStringHelper() {
      SWIGRegisterStringCallback_$module(stringDelegate);
    }
  }

  static protected SWIGStringHelper swigStringHelper = new SWIGStringHelper();
%}
```

If you don't do this then you will get corrupted text when the default code page (the "language for non-unicode applications") is set to Japanese and you're passing UTF-8 Japanese text.

It would be really good to get this PR merged. Hopefully everyone is on .NET 4.8 now.

For anyone that is looking to do this without using a custom build of SWIG, we did the following:

1. Created our own versions of `std_string.i` and `csharphead.i` with these changes, but without the `#ifdef`s and without the section `SWIG_CSHARP_NO_IMCLASS_STATIC_CONSTRUCTOR`
2. Included our versions of these files in our root .i file
3. Passed `-DSWIG_CSHARP_NO_STRING_HELPER=1 -DSWIG_CSHARP_NO_EXCEPTION_HELPER=1` to `swig` to disable the default implementations and use our own

erezgeva:
Though I hate white spaces at the end of line.
Removing them in lines you did not change make the patch harder for reading.
Please do not do that.

erezgeva:
I did not find explanation for the change.
Why do we need to change?

stevebaxter:
This adds support for interoperability of C++ UTF-8 strings with C#, preserving the encoding correctly.

emmenlau:
Dear @erezgeva the motivation is in https://github.com/swig/swig/issues/2357, is that helpful?

erezgeva:
> Dear @erezgeva the motivation is in #2357, is that helpful?

I meant on the solution, but I probably did not explained myself properly.
Sorry.

erezgeva:
Could you, please add a test that verify that the fix works.

Perhaps implement Examples/test-suite/unicode_strings.i
And add it a test for unicode string with umlaut.

julian-smith-artifex-com:
I came across this PR as part of working on MuPDF's C# bindings.

I've managed to get things to work simply by copying just the new `std_string.i` code into the top of my `.i` file (and i've also added a third typemap for `const char*` because MuPDF's API uses `const char*` as well as `std::string`).

So i have this at the top of my SWIG input `.i` file:

```
// This ensures that our code below overrides whatever is defined
// in std_string.i and any later `%include "std_string.i"` is
// ignored.
%include "std_string.i"

// See https://github.com/swig/swig/pull/2364. We also add typemaps
// for `const char*`.

%{
#include <string>
%}

namespace std
{
    %typemap(imtype,
             inattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             outattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             directorinattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             directoroutattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]"
             ) string "string"


    %typemap(imtype,
             inattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             outattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             directorinattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             directoroutattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]"
             ) const string & "string"

    %typemap(imtype,
             inattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             outattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             directorinattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
             directoroutattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]"
             ) const char* "string"
}

```

* This seems to work on Windows for passing Unicode strings between C++ and C# in both directions.
* I haven't tried using Unicode strings with Directors.
* Until the PR is merged into SWIG, pasting the above code into the `.i` file seems to be simpler than creating new versions of `std_string.i` and `csharphead.swg` and using `-DSWIG_CSHARP_NO_STRING_HELPER =1 -DSWIG_CSHARP_NO_EXCEPTION_HELPER=1`. But it's entirely possible that i've missed something important here.
* Unicode strings already work fine with Mono on Linux, i.e. without the above changes.
* So far things work for me without the changes to `csharphead.swg`. I'm not sure what they are for - is it to create C++ exception objects from C#?

alexzouiten:
Note that you can specify multiple types in a typemap to avoid repetition, e.g. like so :

```
%typemap(imtype,
    inattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
    outattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
    directorinattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
    directoroutattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]"
) std::string, const std::string &, const char* "string"
```

julian-smith-artifex-com:
> Note that you can specify multiple types in a typemap to avoid repetition, e.g. like so :
> 
> ```
> %typemap(imtype,
>     inattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
>     outattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
>     directorinattributes="[global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]",
>     directoroutattributes="[return: global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPUTF8Str)]"
> ) std::string, const std::string &, const char* "string"
> ```

Thanks for this, it works so i've modified my mupdf code accordingly.

I'm using `SWIG Version 4.3.1`; interestingly things go wrong if i preserve the original code's use of name `"string"`:

`%typemap(...) std::string, const std::string & "string", const char* "string"`
Swig fails with `Error: Syntax error in input(1).`

Other variations also fail:

`%typemap(...) std::string, const std::string &, const char* string`
Swig fails with `Error: Syntax error in input(1).`

`%typemap(...) std::string, const std::string & string, const char* "string"`
SWIG parses ok but tests fail at runtime, presumably the `const std::string` conversion isn't happening.

julian-smith-artifex-com:
> Note that you can specify multiple types in a typemap to avoid repetition, e.g. like so :

Unfortunately using multiple types in the typemap breaks `swig -csharp`, for example:
```
platform\csharp\mupdfcpp_swig.i(25) : Error: Syntax error in input(1).
```
So i've had to revert to using three separate typemaps.

Issue #2370, opened 2022-09-15
Build of med is failing with SWIG 4.1.0


jplesnik:
I am testing rebuild of SWIG dependencies against version 4.1.0 in Fedora rawhide/38.

The build of `med-4.1.0` is failing with this version.
```
/builddir/build/BUILD/med-4.1.0/redhat-linux-build/python/med/medequivalence_modulePYTHON_wrap.cxx:5410:5: error: 'std_vector_Sl_med_float_Sg__operator_Sa__Se_' was not declared in this scope; did you mean 'std_vector_Sl_med_float_Sg__operator_Sa___SWIG'?
 5410 |     std_vector_Sl_med_float_Sg__operator_Sa__Se_(&v,value);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     std_vector_Sl_med_float_Sg__operator_Sa___SWIG
/builddir/build/BUILD/med-4.1.0/redhat-linux-build/python/med/medequivalence_modulePYTHON_wrap.cxx: In function 'std::vector<double, std::allocator<double> > std_vector_Sl_med_float_Sg__operator_Ss___SWIG(const std::vector<double, std::allocator<double> >*, const std::vector<double, std::allocator<double> >&)':
/builddir/build/BUILD/med-4.1.0/redhat-linux-build/python/med/medequivalence_modulePYTHON_wrap.cxx:5426:5: error: 'std_vector_Sl_med_float_Sg__operator_Ss__Se_' was not declared in this scope; did you mean 'std_vector_Sl_med_float_Sg__operator_Ss___SWIG'?
 5426 |     std_vector_Sl_med_float_Sg__operator_Ss__Se_(&v,value);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     std_vector_Sl_med_float_Sg__operator_Ss___SWIG
```
Complete build log is [here](https://download.copr.fedorainfracloud.org/results/jplesnik/swig-rebuild/fedora-rawhide-x86_64/04830441-med/builder-live.log.gz). Other information about the build can be found [here](https://copr.fedorainfracloud.org/coprs/jplesnik/swig-rebuild/build/4830441/).

jschueller:
seems to come from 26bf86322b18040e2f2156c34a47b44c927d1399, any clue what happens @vadz ?

vadz:
[...]

Issue #2368, opened 2022-09-15
Build of player is failing with SWIG 4.1.0


jplesnik:
I am testing rebuild of SWIG dependencies against version 4.1.0 in Fedora rawhide/38.

The build of `player-3.1.0` is failing with this version.
```
/builddir/build/BUILD/player-release-3-1-0/redhat-linux-build/client_libs/libplayerc/bindings/python/playercPYTHON_wrap.c: In function 'python_playerc_blackboard_create':
/builddir/build/BUILD/player-release-3-1-0/redhat-linux-build/client_libs/libplayerc/bindings/python/playercPYTHON_wrap.c:4859:33: error: 'playerc_blackboard___python_on_blackboard_event__' undeclared (first use in this function); did you mean 'playerc_blackboard___python_on_blackboard_event____SWIG'?
 4859 |   device->on_blackboard_event = playerc_blackboard___python_on_blackboard_event__;
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                 playerc_blackboard___python_on_blackboard_event____SWIG
```
Complete build log is [here](https://download.copr.fedorainfracloud.org/results/jplesnik/swig-rebuild/fedora-rawhide-x86_64/04830452-player/builder-live.log.gz). Other information about the build can be found [here](https://copr.fedorainfracloud.org/coprs/jplesnik/swig-rebuild/build/4830370/).

If you want to look at source code of used `player`, please check [github](https://github.com/playerproject/player/tree/release-3-1-0).

jschueller:
seems similar to https://github.com/swig/swig/issues/2370

ojwb:
@jplesnik Please can you retest with b9fa8c2 or newer.

Issue #2366, opened 2022-09-14
Build of libsolv with SWIG 4.1.0


jplesnik:
The build of libsolv is failing with SWIG 4.1.0.

```
+ /usr/bin/python3 -c 'import solv'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/builddir/build/BUILDROOT/libsolv-0.7.22-1.fc38.x86_64/usr/lib64/python3.11/site-packages/solv.py", line 12, in <module>
    import _solv
ImportError: /builddir/build/BUILDROOT/libsolv-0.7.22-1.fc38.x86_64/usr/lib64/python3.11/site-packages/_solv.so: undefined symbol: XSolvable___eq__
error: Bad exit status from /var/tmp/rpm-tmp.El0FDg (%check)
    Bad exit status from /var/tmp/rpm-tmp.El0FDg (%check)
```
Complete build log is [here](https://download.copr.fedorainfracloud.org/results/jplesnik/swig-rebuild/fedora-rawhide-x86_64/04830436-libsolv/builder-live.log.gz). 

If you want to look at source code of used libsolv, please check [libsolv github](https://github.com/openSUSE/libsolv).

jschueller:
definetely 26bf86322b18040e2f2156c34a47b44c927d1399, cc @vadz

jschueller:
```
curl -L https://github.com/openSUSE/libsolv/archive/refs/tags/0.7.22.tar.gz | tar xz
cd libsolv-0.7.22
cmake -DENABLE_PYTHON=ON -DCMAKE_INSTALL_PREFIX=/usr .
make install -j8
python3 -c "import solv"
```

jschueller:
for this on it seems libsolv relies on the old swig-generated name, with this patch it goes to another similar error:
```
diff --git a/bindings/solv.i b/bindings/solv.i
index 1d35bf6..b4ab667 100644
--- a/bindings/solv.i
+++ b/bindings/solv.i
@@ -3261,7 +3261,7 @@ returnself(matchsolvable)
   %rename("!=") __ne__;
 #endif
   bool __ne__(XSolvable *s) {
-    return !XSolvable___eq__($self, s);
+    return !XSolvable___eq__SWIG($self, s);
   }
 #if defined(SWIGPYTHON)
   int __hash__() {
```

vadz:
[...]

PR #2371, opened 2022-09-17
Make method wrappers suffix optional and disabled by default

vadz:
Unfortunately the changes of 26bf86322 (Use SWIG-specific for non-overloaded synthesized functions too, 2021-11-09) did break some existing code bases using SWIG as they hardcoded the old wrapper function names.

So turn this off by default and add a global variable allowing to enable this, which can be done for a specific language only. This is ugly but, unfortunately, there is no way to use the Language object from the C function Swig_MethodToFunction(), so the only alternative would be to add another parameter to it, but it already has 6 of them, so it wouldn't really be that much better.

See #2366, #2368, #2370.

----

@wsfulton This fixes the breakage due to my commit above and needs to be applied before 4.1 to fix the regressions due to it.

### DIFF
--- a/Source/Modules/main.cxx
+++ b/Source/Modules/main.cxx
@@ -37,6 +37,9 @@ int Verbose = 0;
 int AddExtern = 0;
 int NoExcept = 0;
 int SwigRuntime = 0;		// 0 = no option, 1 = -runtime, 2 = -noruntime
+extern "C" {
+  int UseWrapperSuffix = 0;	// If 1, append suffix to non-overloaded functions too.
+}
 
 /* Suppress warning messages for private inheritance, preprocessor evaluation etc...
    WARN_PP_EVALUATION                           202
--- a/Source/Modules/lang.cxx
+++ b/Source/Modules/lang.cxx
@@ -77,6 +77,9 @@ static Hash *classhash;
 extern int GenerateDefault;
 extern int ForceExtern;
 extern int AddExtern;
+extern "C" {
+  extern int UseWrapperSuffix;
+}
 
 /* import modes */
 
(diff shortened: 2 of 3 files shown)


Source/Modules/main.cxx
```diff
@@ -37,6 +37,7 @@ int Verbose = 0;
 int AddExtern = 0;
 int NoExcept = 0;
 int SwigRuntime = 0;		// 0 = no option, 1 = -runtime, 2 = -noruntime
+int UseWrapperSuffix = 0;	// If 1, append suffix to non-overloaded functions too.
```
jschueller, line 40: is there an option to actually enable it ?

vadz: Yes, in [C backend](https://github.com/swig/swig/pull/2086).

jschueller: oh, thanks

ojwb:
> some existing code bases [...] hardcoded the old wrapper function names

We could say they shouldn't be hardcoding implementation details which are meant to be internal to SWIG.

The original change avoids potential name clashes, doesn't it?  If so it seems a desirable change, and maybe forcing existing bindings which assumed too much about SWIG internals to update is a price worth paying.

Or have I misunderstood and that part was only relevant for the C target language?

vadz:
> > some existing code bases [...] hardcoded the old wrapper function names
> 
> We could say they shouldn't be hardcoding implementation details which are meant to be internal to SWIG.

We could say it, but it seems better to avoid breaking the existing code, especially considering that, from what I've seen, these hardcoded names appear inside macros hairy enough that people would be almost certainly very unwilling to dig into them...
 
> The original change avoids potential name clashes, doesn't it? If so it seems a desirable change, and maybe forcing existing bindings which assumed too much about SWIG internals to update is a price worth paying.
> 
> Or have I misunderstood and that part was only relevant for the C target language?

I think it's only relevant to C language, as the name clash is between the wrapper function and the original function -- but these names are not used in the same files except for C (and probably Objective-C too). I'm sure that it is needed for C, however, which is why I'd like to have it in master, as at least like this I can avoid having differences between the C branch in its for the core SWIG files.

ojwb:
> I think it's only relevant to C language, as the name clash is between the wrapper function and the original function

OK - merging then.

Issue #2353, opened 2022-09-06
Failing octave test


jplesnik:
I tried to rebuild SWIG 4.1.0 on Fedora Rawhide, but octave test is failing with error:

```
error: swig_ref/horzcat method failed
error: called from
    horzcat
    runme at line 50 column 3
make[3]: *** [../../Makefile:473: octave_run] Error 1
make[2]: *** [../example.mk:11: check] Error 2
make[1]: *** [Makefile:215: operator.actionexample] Error 2
```

Fedora rawhide contains octave-7.2.0
Fedora build and logs could by found [here](https://koji.fedoraproject.org/koji/taskinfo?taskID=91700112)
I enabled only octave tests for the build. Other tests, which we run on Fedora, are passing.

ojwb:
There are fixes in 4.1.0 for "Octave 4.4 - 6.0" and "Octave 6 up to and including 6.4".  I know little about Octave, but I suspect Octave 7 support needs work.

jschueller:
cc @kwwette

wsfulton:
@friedrichatgc would you be able to look at this? The failing test was introduced by your #1672 pull request. BTW, why was the subdirectory given such an odd name with a leading at sign: `@swig_ref` ?

wsfulton:
@jplesnik if it isn't hard, could you run the octave test-suite and perhaps suppress all the new noisy annoying gcc warnings? These ones: `-Wmisleading-indentation -Wdeprecated` and the `_FORTIFY_SOURCE requires compiling with optimization` ? Hopefully, it is just this one test that is broken.

friedrichatgc:
@wsfulton The horzcat.m function extends the octave class named "swig_ref" (SWIG introduces this name) by the member function horzcat.
From the octave docu [https://docs.octave.org/v4.0.1/Creating-a-Class.html#Creating-a-Class]
"All object oriented classes in Octave must be located in a directory that is the name of the class prepended with the ‘@’ symbol."
That's why this subdirectory must be named exactly like this.

I'm out for the next week and don't know when I find time to look at this. But I will try.

jplesnik:
> @jplesnik if it isn't hard, could you run the octave test-suite and perhaps suppress all the new noisy annoying gcc warnings? These ones: `-Wmisleading-indentation -Wdeprecated` and the `_FORTIFY_SOURCE requires compiling with optimization` ? Hopefully, it is just this one test that is broken.

[...]

PR #2374, opened 2022-09-18
Fix octave operator horzcat test

friedrichatgc:
Fixed and reenable the octave horzcat operator test for Octave 7.1.

Added documentation about extending Octave operators not known by C++ using standard Octave object orientated programming methods.


friedrichatgc:
This PR should fix #2353.

wsfulton:
Thanks, that looks perfect.

PR #2379, opened 2022-09-22
Further improve type handling for long and long long in csharp

emmenlau:
This pull request aims to improve the handling of the `long` and `long long` datatypes (and their variants) in csharp. It is based on and continues the work of #2237, and related to #568 and #650.

This pull request improves upon the base in such a way, that typemaps for the `long long` datatype (and its variants) are also generated.

Additionally, it fixes a bug in the generated code, that leads to compile errors in certain circumstances:

When `SWIGWORDSIZE64` is set, there is a difference in how `int64_t` is defined between compilers (tested on Linux GCC and MacOS CLANG):
 * `int64_t` is `long long` in MacOS
 * `int64_t` is `long` in Linux (GCC)

Since generated SWIG code uses the basic types of `long*` and `long long*` to try the casting to `int64_t*`, the code is either not compatible with Linux or MacOS, since `long long*` cannot be cast to `int64_t*` in Linux and vice versa.

This commit introduces a new definition `SWIGINT64ISLONGLONG` which users should set in addition to `SWIGWORDSIZE64`, when the compiler treats `int64_t` as `long long`.

**Important:**
To make good use of this code, users will need to *manually* define `SWIGWORDSIZE64` and `SWIGINT64ISLONGLONG` as needed. An example on how to do that in cmake is here:
```
    # Start with empty SWIG_EXTRA_COMPILE_OPTIONS:
    set(SWIG_EXTRA_COMPILE_OPTIONS)
    if(UNIX)
        # On Unix, 'long' is 'int64_t' in contrast to Windows where it is
        # 'int32_t' (see also
        # https://learn.microsoft.com/en-us/cpp/cpp/fundamental-types-cpp#sizes-of-built-in-types)
        list(APPEND SWIG_EXTRA_COMPILE_OPTIONS "-DSWIGWORDSIZE64")
    endif()
    if(APPLE)
        # On Apple, 'int64_t' is 'long long', in contrast to Linux where it is
        # 'long':
        list(APPEND SWIG_EXTRA_COMPILE_OPTIONS "-DSWIGINT64ISLONGLONG")
    endif()

    set_property(SOURCE "MySwigInterface.i" PROPERTY COMPILE_OPTIONS
        ${SWIG_EXTRA_COMPILE_OPTIONS})
```


emmenlau:
I have tested this PR on Linux (Ubuntu 22.04 with clang 14), MacOS (with XCode 13), and Windows with Visual Studio 17.3.3 and ClangCl 15 compiler frontend. Without this PR, on at least one of these platforms `size_t` was not correctly used as a 64bit unsigned integer. With this PR, identical csharp code is generated on all platforms and the C++ seems to sensibly use 64bit unsigned integer.

emmenlau:
Dear core developers, my changes trigger one error in the CI, in https://github.com/swig/swig/actions/runs/3103774133/jobs/5027472861:
```
../../../../../Examples/test-suite/csharp/sizet_runme.cs(9,15): error CS0266: Cannot implicitly convert type `ulong' to `uint'. An explicit conversion exists (are you missing a cast?)
../../../../../Examples/test-suite/csharp/sizet_runme.cs(10,15): error CS0266: Cannot implicitly convert type `ulong' to `uint'. An explicit conversion exists (are you missing a cast?)
../../../../../Examples/test-suite/csharp/sizet_runme.cs(11,15): error CS0266: Cannot implicitly convert type `ulong' to `uint'. An explicit conversion exists (are you missing a cast?)
../../../../../Examples/test-suite/csharp/sizet_runme.cs(12,15): error CS0266: Cannot implicitly convert type `ulong' to `uint'. An explicit conversion exists (are you missing a cast?)
```
This is very relevant and I'd like to address it. But I think the problem is not in the code, but rather the required change in the build system. SWIG should now define `SWIGWORDSIZE64` and `SWIGINT64ISLONGLONG` as needed. Is that a reasonable way forward? Help would be appreciated!

emmenlau:
Dear @wsfulton , I think I need your help to continue. In my humble opinion this PR here is a very important change for csharp, because without it, I am basically unable to use many 64bit types, with different problems on different platforms. All this is resolved by this PR (thanks mostly to @igt-gteubl and others).

But what is a clean way for SWIG to define `SWIGWORDSIZE64` and `SWIGINT64ISLONGLONG`? The setting depends on the c++ compiler on the target platform (be it MSVC, Apple or Linux/Unix).

Ideally for users, the swig generator would detect the setting automatically and things "just work"? But if the user wants to generate for a *different* target platform than the current, it could be helpful if there is a way to override the setting, for example with compiler defines or a command line switch?

What do you think?

ojwb:
> Ideally for users, the swig generator would detect the setting automatically and things "just work"? But if the user wants to generate for a different target platform than the current, it could be helpful if there is a way to override the setting, for example with compiler defines or a command line switch?

The aim is that the generated code is portable without the use of special switches.

emmenlau:
> The aim is that the generated code is portable without the use of special switches.

That sounds interesting, but I'm not sure I fully understand. Currently in master, without this PR, users must already set one define `SWIGWORDSIZE64`, that applies at swig generation time, and changes how the code is generated. This PR adds another define `SWIGINT64ISLONGLONG` on top of that. Both defines are necessary to correctly support 32 vs 64bit and Linux vs MacOS.

Can you say how you would prefer to change all of this?

As far as I know, the define `SWIGINT64ISLONGLONG` is only required on the C++ side, not the C# side. So it may be very well possible to generate C++ code that automatically sets `SWIGINT64ISLONGLONG` based on the compiler define `__APPLE__`. Is that what you are saying?

As far as I know, the other define `SWIGWORDSIZE64` applies also to the generated C# code. I guess there is no way to set that without knowing the target architecture that the user is generating code for. Or at least I'm not knowledgeable enough with C# to see a way to do that.

ojwb:
I haven't looked at the details of this issue since wsfulton is the SWIG/C# maintainer and I really don't know much about SWIG/C# or C# itself.

But the general approach SWIG takes it to generate portable code if at all possible.  If it isn't possible, we try to at least generate code where the difference can be selected at wrapper C/C++ compile time so the generated wrapper source code can be distributed and successfully built on arbitrary platforms (potentially with some `-DSWIGXXX` options passed to the compiler).

There are some cases where that's hard to do - `SWIGWORDSIZE64` is one of those but as `swigarch.i` says:

_Use only in extreme cases, when no arch. independent code can be generated_

(The whole `SWIGWORDSIZE32` vs `SWIGWORDSIZE64` dichotomy is also fundamentally flawed, but that's literally a separate issue: #1850)

So if you can make the generated wrapper use C/C++ compile time conditionals like `#ifdef SWIGINT64ISLONGLONG` that's preferable to generating different wrapper code based on SWIG-time `#ifdef SWIGINT64ISLONGLONG` conditionals.

> So it may be very well possible to generate C++ code that automatically sets SWIGINT64ISLONGLONG based on the compiler define __APPLE__. Is that what you are saying?

I'm dubious about trying to set `SWIGINT64ISLONGLONG` automatically with preprocessor checks as getting it right in every case seems hard to do (inferring based on checking `__APPLE__` seems particularly brittle and liable to break with future Apple platforms which will presumably eventually have 128 bit or larger word sizes).  Probably best to leave that to the SWIG-using project's build system (autoconf/cmake/whatever) which can actually probe for what the underlying type of `int64_t` is on the current platform.

emmenlau:
Thanks a lot @ojwb , I'll incorporate your feedback into this PR!

One side note: I think it would be nice if `SWIGINT64ISLONGLONG` will respect the setting *if* it is already set, but otherwise include a few common defaults to automatically set it with preprocessor checks. Since only apple is different, this will probably cover 98% of the cases, and in the rare other cases, users can override the setting from their build system. Is that good?

ojwb:
Trying to determine the setting with the limited preprocessor macros available just seems to be creating some brittle code for us to have to maintain going forwards.

`__APPLE__` only says "this is a apple platform", but you're trying to use it to infer something about what one particular typedef is.  That inference happens to be correct for a particular apple platform with the default compiler toolchain right now.

It's also clearly insufficient as any platform with a 32-bit `long` can't be using `typedef long int64_t;`...

igt-gteubl:
I don't know how relevant this information is for this issue, but in our projects, we detect if we must set `SWIGINT64ISLONGLONG` by running a compiler pre-check in our build tool (in this case CMake).
```
#include <stdint.h>
int main() {
  const int64_t* a = 0;
  const long long* b = a;
  // suppress 'unused' warning that will trigger an error in case of -Werror
  (void) b;
  return 0;
}
```
If it compiles, we set the flag, if not, we don't.

As a side-note, we do something similar for `SWIGWORDSIZE64`, because when building for several targets, even determining this needs some special handling.

I'm not sure if this logic can be automated - maybe the generated code can do that on it's own at compile-time (via a macro or so), but for us it was way easier to just pass that flag in, based on what the compiler actually wants from us.

emmenlau:
> `__APPLE__` only says "this is a apple platform", but you're trying to use it to infer something about what one particular typedef is. That inference happens to be correct for a particular apple platform with the default compiler toolchain right now.
>
> It's also clearly insufficient as any platform with a 32-bit `long` can't be using `typedef long int64_t;`...

This criticism seems a bit unjustified. Do you assume that preprocessor checks could not cover this case?

But it seems you have a better idea how to solve this. That is actually great, I did not mean to enforce anything!

emmenlau:
@igt-gteubl that sounds like a very reasonable way forward!

But I guess that ideally this would need to run at the time the downstream project is built, the project that consumes the SWIG-generated code. One possible way to do that would be to require all users to add something like this check to their own build systems. But that seems quite an effort, and it may not be easily discoverable.

I'm not sure if there are better ways. I guess if SWIG would want to support that without further user involvement, a possible way would be to integrate this into a SWIG cmake package configuration. Then at least users of cmake could import the configuration and would not need extra steps, the check would be executed automatically. However this idea would require users to have cmake, and be aware that they need to import something from SWIG. And since cmake brings already a nice SWIG integration with the `UseSWIG` module, it may be very worthwhile to work with upstream cmake to support this out of the box.

Not sure if I'm on the right track with any of this, just my 2 cents...

wsfulton:
long long is 64-bit on all systems and int is 32-bit - nice table here: https://www.ccoderun.ca/programming/2017-03-10_sizeof/. 
The default is to treat long as 32-bit, noting:

%typemap(imtype) long,               const long &               "int"
%typemap(imtype) unsigned long,      const unsigned long &      "uint"

So it is treated as int (32-bit) on the C# side. These integer types are blittable, see the table at https://www.mono-project.com/docs/advanced/pinvoke/#blittable-types. Currently on the C side, the types we use are:

%typemap(ctype) long,               const long &               "long"
%typemap(ctype) unsigned long,      const unsigned long &      "unsigned long"

So even when on systems where long is 64-bits, long is truncated to 32-bit. What I propose is to fix the anomoly above in the default case to:

%typemap(ctype) long,               const long &               "int"
%typemap(ctype) unsigned long,      const unsigned long &      "unsigned int"

so that both the marshalling between the C type and .NET type is exactly 32-bit. Checking on Linux 64-bit where long is 64-bit, the marshalling is equally broken with numbers > 32-bit, before and after the change.

Then users can choose to treat long as 64-bit by defining SWIGWORDSIZE64:
```swig
#if defined(SWIGWORDSIZE64)
%apply long long { long };
%apply unsigned long long { unsigned long };
%apply const long long & { const long & };
%apply const unsigned long long & { const unsigned long & };
%include <swigarch.i>
#endif
```
This is very similar to what I've prepared for Java for #646, but havn't pushed it yet.

I can't see a perfect solution other than to avoid the long type. Above at least provides some choice and compromise for those that need to use long. And yes, it does unfortunately provide the option to use different code on different platforms if that's what a user really wants, but might be necessary. Using the default option of 32-bits is probably the best recommendation and to use numbers < 32-bit in size. 

Some additional improvements could be:

1. When SWIGWORDSIZE64 is not defined (the default), the typemaps could be enhanced to do a runtime truncation check that a number does not exceed 32-bits when long is 64-bits passing the number from C to .NET.
2. When SWIGWORDSIZE64 is defined, the typemaps could be enhanced to do a runtime truncation check a number does not exceed 32-bits when long is 32-bits passing the number from .NET to C.

wsfulton:
fb91d1fa25ba00f93b9870d84ec3cbc0739650d7 addresses this issue.

PR #2391, opened 2022-10-05
[Python] Remove deprecated and apparently useless defarg.swg

### LINKED ISSUE #2390: unused swig_call_defargs ?
seems swig_call_defargs is not used anywhere, should it be removed (python/defarg.swg) ?

ojwb:
The only documentation is in the file itself and describes a Python wrapper around the C function defined here, but that digging though the git history this Python wrapper doesn't seem to have ever actually been generated by SWIG.

This file was also marked as deprecated in 2005.

Fixes #2390


wsfulton:
Seems reasonable to me.

PR #2392, opened 2022-10-07
Use more portable PyUnicode_AsUTF8String.

clintonstimpson:
This supports the use of Py_LIMITED_API

### DIFF
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -34,7 +34,14 @@ SWIGINTERN char*
 SWIG_Python_str_AsChar(PyObject *str)
 {
 #if PY_VERSION_HEX >= 0x03030000
-  return (char *)PyUnicode_AsUTF8(str);
+# if !defined(Py_LIMITED_API) || Py_LIMITED_API+0 >= 0x030A0000
+  return (char*)PyUnicode_AsUTF8AndSize(str, NULL);
+# else
+  PyObject* bytes = PyUnicode_AsUTF8String(str);
+  char* chars = PyBytes_AsString(bytes);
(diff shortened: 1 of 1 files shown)


wsfulton:
Any info on performance difference? I havn't looked in detail, but superficially it looks like an extra string copy.

ojwb:
At the very least it requires an extra PyObject creation and destruction, but it does seem like it probably does copy the string (or else there would need to be reference counting of the buffer holding the cached internal UTF-8 representation).

`PyUnicode_AsUTF8AndSize()` is likely better from a performance point of view, as it's documented to "Return a pointer to the UTF-8 encoding of the Unicode object" (like `PyUnicode_AsUTF8()` is documented to).

It's only "Part of the Stable ABI since version 3.10" so use would need to be conditional on "<want_stable_abi> || <python version >= 3.10>".

clintonstimpson:
@ojwb @wsfulton, I've changed this to use `PyUnicode_AsUTF8AndSize()` as suggested.
However, it still uses `PyUnicode_AsUTF8String/PyBytes_AsString` if one sets `Py_LIMITED_API` or sets `Py_LIMITED_API < 3.10`, so only those who wanted limited api support with older versions of python might have a performance penalty.  I haven't looked into how much of a penalty that would be.

ojwb:
@clintonstimpson Thanks.  This deserves a `CHANGES` entry, but I'm assuming there are some more `Py_LIMITED_API`-related changes to come so one can be added once we're fully there.

clintonstimpson:
@ojwb  Thanks for merging this.  And yes, there are more changes to make.  Some of those changes will depend on #2350 being merged.

Issue #886, opened 2017-02-02
Exporting overloaded operators from C++ to C# is very troublesome.


Neutrino-Sunset:
I'm trying to consume a C++ library from a Dotnet Framework module, so I'm using SWIG to generate a C API (which I compile into the C++ library itself), and C# wrapper class (which I compile into my Dotnet Framework client module). Several of the types in the C++ library declare operator overloads but the SWIG documentation states that it handles overloaded operators so I didn't expect any problems. Unfortunately though the process of consuming these operator overloads from the C# module was not as effortless as I had expected.

The following C++ class definition contains one overloaded operator declared as a member function and another declared as a friend function.

*Class1.h*

    class Class1
    {
    public:
       bool operator<(const Class1& rhs)const;
       friend bool operator==(const Class1& lhs, const Class1& rhs);
    };

SWIG's default behaviour is to warn that "Can't wrap 'operator`<whatever>`' unless renamed to a valid identifier." Rename directives can be added to the SWIG config file to cause these operators to be picked up. E.g.

*TestLib1.swg*

    %module TestLib1
    %{
    #include "Class1.h"
    %}
    %rename(LessThan) Class1::operator<(const Class1&)const;
    %rename(Class1_Equals) operator==(const Class1&, const Class1&);

This results in the member `operator<` being declared in Class1 as `Class1::LessThan` and friend `operator==` being declared as function `TestLib1.Class1_Equals`.

These functions both work, however for types that are predominantly manipulated via operators (e.g. user defined numeric types) the need to fall back to using function syntax in the client code is an unacceptable inconvenience.

Given that the operator functions themselves have been generated it would seem to be a simple process to augment the generated C# class to use them. I have managed to achieve this with minimal impact on the generated C# class files via the use of partial classes. To achieve this I modify the generated C# class file so that the class is declared partial, then add the overloaded operator definitions in a separate file like so...

*Class1Operators.cs*

    public partial class Class1
    {
       public static bool operator<(Class1 lhs, Class1 rhs)
       {
          return lhs.Equals(rhs);
       }
    
       public static bool operator==(Class1 lhs, Class1 rhs)
       {
          return TestLib1.Class1_Equals(lhs, rhs);
       }
    }

[...]

PR #2393, opened 2022-10-10
C#: Added equality operator for instances of `std::vector`

emmenlau:
This PR is a draft to better discuss issue #886. It adds an equality (and inequality) operator to instances of `std::vector` if their contained items are comparable. This is by default enabled for fundamental types and `std::string`, and can be enabled for further types by the user with the help of macro `SWIG_STD_VECTOR_ENHANCED(x)`.


wsfulton:
Can you rewrite to use less modern C# features. The test results are showing lots of:
```
error CS1644: Feature `pattern matching' cannot be used because it is not part of the C# 7.0 language specification
```

Can you show it working by enhancing one of the existing std::vector testcases?

PR #2394, opened 2022-10-10
Support boolean values in preprocessor expressions

emmenlau:
I think the SWIG preprocessor does currently not support statements including boolean values, for example:
```
#if false
```

I've added support for this, to the best of my knowledge, based on the preprocessor code for numeric values. In the code I assume that the char-pointer returned by `Char(Scanner_text(scan))` is zero-terminated, otherwise its unsafe to use `strcmp()`.

Is this a reasonable PR? Please guide me if the code should be improved.

### DIFF
--- a/Source/Preprocessor/expr.c
+++ b/Source/Preprocessor/expr.c
@@ -320,7 +320,17 @@ int Preprocessor_expr(DOH *s, int *error) {
 	*error = 1;
 	return 0;
       }
-      if ((token == SWIG_TOKEN_INT) || (token == SWIG_TOKEN_UINT) || (token == SWIG_TOKEN_LONG) || (token == SWIG_TOKEN_ULONG)) {
+      if (token == SWIG_TOKEN_BOOL) {
+	/* A boolean value.  Reduce EXPR_TOP to an EXPR_VALUE */
+	char *c = Char(Scanner_text(scan));
+	if (strcmp(c, "true") == 0) {
+	  stack[sp].value = (long) 1;
+	} else {
+	  stack[sp].value = (long) 0;
+	}
+	stack[sp].svalue = 0;
+	stack[sp].op = EXPR_VALUE;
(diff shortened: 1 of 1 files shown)


Source/Preprocessor/expr.c
```diff
@@ -320,7 +320,17 @@ int Preprocessor_expr(DOH *s, int *error) {
 	*error = 1;
 	return 0;
       }
-      if ((token == SWIG_TOKEN_INT) || (token == SWIG_TOKEN_UINT) || (token == SWIG_TOKEN_LONG) || (token == SWIG_TOKEN_ULONG)) {
+      if (token == SWIG_TOKEN_BOOL) {
+	/* A boolean value.  Reduce EXPR_TOP to an EXPR_VALUE */
+	char *c = Char(Scanner_text(scan));
+	if (strcmp(c, "true") == 0) {
```
wsfulton, line 326:
```suggestion
	String *c = Scanner_text(scan);
	if (Strcmp(c, "true") == 0) {
```

wsfulton (commented):
Looks okay to me with this minor change, but @ojwb knows this area best.

ojwb:
Seems OK to me but does indeed need test coverage.

That should probably include a check that this isn't activated for C++ - e.g. this should not error without `-c++` because in C `true` is not special here, so it just gets replaced with `0` like any other undefined identifier:

```
#if true
#error C++ only
#endif
```

Looks to me like that is already the case with this patch, but test coverage ensures that and catches any future change which would break that.

ojwb:
Needs the feedback addressing and feature tests adding.

wsfulton:
I've earmarked this for the swig-4.2 release. @emmenlau could you address these simple change requests please.

wsfulton:
Squash merged in 9ed26fcf6680646b6c48ee7172a119b002c06f64.

PR #2398, opened 2022-10-13
[lua] Fix C++ memory leaks on SWIG_fail

ojwb:
SWIG_fail in lua leads to a call to lua_error() which calls longjmp() which means the destructors of any live function-local C++ objects won't get run.

To avoid this happening, we now wrap almost everything in the function in a block, and end that right before lua_error() at which point those destructors will get called.

I don't see how to add a regression test for this within SWIG's testsuite, but it's observable with valgrind.

@wsfulton I'd like to merge this for SWIG 4.1.0 if possible.

### DIFF
--- a/Source/Modules/lua.cxx
+++ b/Source/Modules/lua.cxx
@@ -556,6 +556,12 @@ public:
        this line adds this into the wrapper code
        NEW LANGUAGE NOTE:END *********************************************** */
     Printv(f->def, "static int ", wname, "(lua_State* L) {", NIL);
+    // SWIG_fail in lua leads to a call to lua_error() which calls longjmp()
+    // which means the destructors of any live function-local C++ objects won't
+    // get run.  To avoid this happening, we wrap almost everything in the
+    // function in a block, and end that right before lua_error() at which
+    // point those destructors will get called.
+    if (CPlusPlus) Append(f->def, "\n{");
 
     /* NEW LANGUAGE NOTE:***********************************************
        this prints the list of args, eg for a C fn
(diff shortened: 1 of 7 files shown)


ojwb:
> I don't see how to add a regression test for this within SWIG's testsuite, but it's observable with valgrind.

In the fresh light of morning it occurs to me that we can test this with an object on the stack whose decrements increments `dtors_called` counter.  I'll give that a go.

I can probably take a look at the equivalent issue in Go.

Source/Modules/lua.cxx
```diff
@@ -767,9 +773,19 @@ class LUA:public Language {
...
     // add the failure cleanup code:
     Printv(f->code, "\nif(0) SWIG_fail;\n", NIL);
-    Printv(f->code, "\nfail:\n", NIL);
-    Printv(f->code, "$cleanup", "lua_error(L);\n", NIL);
-    Printv(f->code, "return SWIG_arg;\n", NIL);
+    Printv(f->code, "\nfail:\n", "$cleanup", NIL);
+    if (CPlusPlus) {
+      if (!Strchr(cleanup, ';') && !Strchr(cleanup, '}')) {
+	// If there's no cleanup code we need a dummy statement after the label.
+	Append(f->code, ";\n}\n");
```
ojwb, line 780: Actually, I think this extra `;` needs to be conditional only on `CPlusPlus` - otherwise setting the cleanup code to e.g. `// This is a dummy comment; it should still work!` would result in an uncompilable wrapper...

ojwb:
Now with a regression test.  I found and fixed the same issue in R.

I've had a go at Go and it's much more complex as it needs refactoring to `goto fail:` rather than directly calling `_swig_gopanic()`, so I think that makes more sense as a separate PR (which may not make 4.1.0).

Issue #1879, opened 2020-10-02
SWIG_fail for Java?


nrbrook:
When writing typemaps for Python and others you can use the `SWIG_fail` macro to jump to cleanup code. This doesn't exist when wrapping for java so allocations can't be cleaned up if the typemap needs to return early.

Example:

```
%typemap(in,noblock=1) int * {
	$1 = malloc(sizeof(int));
	if($1 == NULL) {
#ifdef SWIGJAVA
		return $null; // doesn't run cleanup code, leaks arg1 if arg2 allocation fails
#else
                SWIG_fail; // jumps to cleanup code
#endif
	}
}
%typemap(freearg,noblock=1) int * {
	free($1);
}


void afunc(int *varA, int *varB);
```

ojwb:
#2402 covers this issue, so closing this in favour of that.

PR #2400, opened 2022-10-13
Fix issues with exception_memory_leak testcase

ojwb:
The out typemap uses a function name which doesn't match the name of the function we want it to apply to, so this testcase wasn't actually triggering an exception so wasn't actually testing anything!

With that fixed, the testcase fails to compile for PHP due to use of SWIG_exception_fail() (which not all target languages implement), and with that fixed, the _runme.php needs a try ... catch adding to handle the raised exception.

### DIFF
--- a/Examples/test-suite/exception_memory_leak.i
+++ b/Examples/test-suite/exception_memory_leak.i
@@ -1,6 +1,7 @@
 %module exception_memory_leak
 
 %include <std_string.i>
+%include <exception.i>
 
 %typemap(in) Foo* foo
 {
@@ -11,10 +12,14 @@
   Foo::inc_freearg_count();
   delete $1;
 }
-%typemap(out) Foo* verify_no_memory_leak
+%typemap(out) Foo* trigger_internal_swig_exception
 {
-  if ($1 == NULL)
-    SWIG_exception_fail(SWIG_RuntimeError, "Let's see how the bindings manage this exception!");
+  if ($1 == NULL) {
+    SWIG_exception(SWIG_RuntimeError, "Let's see how the bindings manage this exception!");
+#ifdef SWIG_fail
+    SWIG_fail;
+#endif
+  }
   $1 = NULL;
 }
 
--- a/Examples/test-suite/php/exception_memory_leak_runme.php
+++ b/Examples/test-suite/php/exception_memory_leak_runme.php
@@ -18,6 +18,10 @@ check::equal(Foo::get_count(), 2, "Should have 2 Foo objects");
 check::equal(Foo::get_freearg_count(), 1, "freearg should have been used once");
 
 // SWIG exception triggered and handled.
-trigger_internal_swig_exception("null", $b);
+try {
+    trigger_internal_swig_exception("null", $b);
+    check::fail("Expected exception not thrown");
+} catch (Exception $e) {
+}
 check::equal(Foo::get_count(), 2, "Should have 2 Foo objects");
 check::equal(Foo::get_freearg_count(), 2, "freearg should have been used twice");


Examples/test-suite/exception_memory_leak.i
```diff
@@ -11,10 +12,14 @@
...
+%typemap(out) Foo* trigger_internal_swig_exception
 {
-  if ($1 == NULL)
-    SWIG_exception_fail(SWIG_RuntimeError, "Let's see how the bindings manage this exception!");
+  if ($1 == NULL) {
+    SWIG_exception(SWIG_RuntimeError, "Let's see how the bindings manage this exception!");
+#ifdef SWIG_fail
+    SWIG_fail;
+#endif
+  }
```
ojwb, line 22: Needing to do this seems clumsy and makes it harder for users to write typemap code which works across target languages.

Looking at the docs, `SWIG_exception(CODE, MESSAGE)` seems the best defined interface.  Some (but not all) target languages also provide `SWIG_exception_fail(CODE, MESSAGE)` (e.g. PHP and Java don't seem to).  We can't just use `SWIG_exception(x, y); SWIG_fail;` here instead because e.g. Java doesn't seem to provide `SWIG_fail` (#1879 - not clear to me if that's a bug, missing feature, or "wontfix").

Looking at the implementations of `SWIG_exception()` for each target language in `exception.i`, some call a target language function which may or may not return (I'd have to check docs for each to know).

PHP and Lua do `SWIG_fail;` so we could just `#define SWIG_exception_fail(C,M) SWIG_exception(C,M)` for these languages.

Java and CSharp do `return $null;` afterwards, which I guess is close to `SWIG_fail;` (except no `$cleanup` happens?)

Ocaml does `CAMLreturn0;`.

@wsfulton Thoughts?

vadz: FWIW in my code I had to do the following (this was a few years ago, maybe things have improved since then?):
```cpp
#if defined(SWIGC)
  #define MY_exception(msg) SWIG_exception(SWIG_RuntimeError, msg)
  #define SWIG_exception_noret(code, msg) SWIG_exception(code, msg)
  #define MY_after_exception() return $null
#elif defined(SWIGCSHARP)
  #define MY_exception(msg) MY_SetPendingException(msg)
  #define SWIG_exception_noret(code, msg) SWIG_CSharpException(code, msg)
  #define MY_after_exception() return $null
#elif defined(SWIGJAVA)
  #define MY_exception_arglist JNIEnv *jenv
  #define MY_exception_args jenv

  #define MY_exception(msg) MY_SetPendingException(jenv, msg)
  #define SWIG_exception_noret(code, msg) SWIG_JavaException(jenv, code, msg)
  #define MY_after_exception() return $null
#elif defined(SWIGPYTHON) || defined(SWIGJAVASCRIPT)
  #define MY_exception(msg) SWIG_Error(SWIG_RuntimeError, msg)
  #define SWIG_exception_noret(code, msg) SWIG_Error(code, msg)
  #define MY_after_exception() goto fail
#else
  #error "Need to define exception handling macros for this language"
#endif

%header {
%#include "my/exception.h"
static void MY_ExceptionHandler(MY_exception_arglist)
{
  try {
    throw;
  }
  catch ( MY::Exception& e ) {
    MY_exception(e.GetErrorMessage());
  }
  catch ( std::exception& e ) {
    SWIG_exception_noret(SWIG_RuntimeError, e.what());
  }
  catch ( ... ) {
    SWIG_exception_noret(SWIG_RuntimeError,
        "Unexpected unknown exception in C++ code, please report as a bug");
  }
}
}

%exception {
  try {
    $action
  } catch ( ... ) {
    MY_ExceptionHandler(MY_exception_args);
    MY_after_exception();
  }
}
```
which is pretty ugly and it would be nice if SWIG could provide some macros working for all languages (this also requires `MY_SetPendingException()` function to be defined in C#/Java).

ojwb: OK, so it's not just me.

If I understand your example, `SWIG_fail;` is problematic because you've factored out the exception handling into a separate function which doesn't have a `fail:` label (and being able to factor out like that can make a huge difference to code size - we do similar in Xapian's bindings, but we instead have the helper return a `SWIG_XXXError` code and a `std::string` for the message and call `SWIG_exception()` in the `%exception` block itself).

I think the part I'm very unclear on here is exactly what API we should be providing here.

The current documentation of `SWIG_exception()` seems silent on what happens afterwards.  For some languages it does `SWIG_fail`, for some it does `return $null;` or `CAMLreturn0;`, for a few it seems like it just continues to execute whatever is afterwards.

It seems unhelpful for it to just continue (PHP used to just continue and that required special handling in various typemaps that I was able to eliminate by making it `SWIG_fail;` - I looked at some of the other implementations and had concluded PHP was just buggy there, but I can't find any documentation to actually back that up).  I'm not sure though if there are languages where it's not possible (or very hard) to actually bail out.

Having `SWIG_exception()` defined to fail would make `SWIG_exception_fail()` redundant and it could just be aliases for compatibility.

But your example above illustrates that a variant which works in a helper function is useful to have too.

Probably this deserves its own issue opening - I'll do that in the likely event it isn't resolved before I get this PR merged.

vadz: > OK, so it's not just me.

Yes, sorry, I should have said that confirming this was the main point of my comment. IMO it's a problem that anybody implementing customized exception handling is going to have.

> If I understand your example, `SWIG_fail;` is problematic because you've factored out the exception handling into a separate function which doesn't have a `fail:` label

Yes, exactly.

> and being able to factor out like that can make a huge difference to code size

Absolutely.

> I think the part I'm very unclear on here is exactly what API we should be providing here.

Unfortunately me neither, I'd have to re-review all the languages again (looking at this code history I see that it was initially written 11+ years ago, so at least some things must have changed since then). But it seems clear that we need to

1. Keep `SWIG_exception()` for compatibility because changing it any way is bound to break the existing code.
2. Add a new family of macros including something like `SWIG_raise_exception()` and `SWIG_exception_return()`.

And this indeed should be discussed in a separate issue (just in case you were not aware of it, you can extract your comment into one using "Reference in new issue").

ojwb: Thanks, I wasn't aware of that.  I used it but it doesn't actually see to add much over just cutting and pasting the text into a new issue.  In particular it only took the text of a single comment, and the pop up to edit it was rather small and I couldn't resize it so it wasn't very convenient to edit the text.

PR #2401, opened 2022-10-13
Define PY_SSIZE_T_CLEAN only when not defined yet

hramrach:
Users of swig 4.0.2 and python 3.10 will likely define the PY_SSIZE_T_CLEAN macro by hand when relevant because their bindings won't build otherwise.

Unconditionally defining PY_SSIZE_T_CLEAN in swig 4.10 will lead to macro redefinition warning.

Signed-off-by: Michal Suchanek <msuchanek@suse.de>

### DIFF
--- a/Lib/python/embed.i
+++ b/Lib/python/embed.i
@@ -29,7 +29,7 @@ As far as I know, this module is C++ safe.
 #endif
 
 %wrapper %{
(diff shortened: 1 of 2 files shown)


ojwb:
Setting milestone as it's a clean change which improves code that's new in 4.1.0.

wsfulton:
Yes, this can be merged as soon as the tests pass. I'm going to bed now!

PR #2412, opened 2022-10-21
CMake: Avoid globbing tarball parser.c

### LINKED ISSUE #2409: Beta 4.1.0, Windows CMake installation fails at building: '/Foswig.dir\Debug\/Source/CParse/parser.c.obj' not allowed with multiple source files [C:\Dev\Software\swig-build\build\swig.vcxproj]
The compilation was done with VS2022 instead of VS2019 as it's described in the documentation

jschueller:
Closes #2409

Also fix a minor warning that's only seen in release build (default with cmake)


wsfulton:
It will be more maintainable to keep to the globbing and remove parser.c from the list. Can you make that change? If possible today as it is meant to be the date we release 4.1.0.

PR #2413, opened 2022-10-21
[python] Prevent a potential multithreading crash

### LINKED ISSUE #2396: Potential Python crash with `-threads` option and Python multithreading
With the code that SWIG generates for Python multithreading (i.e. with the `-threads` option), there's a chance of a SIGABRT from `std::terminate`, due to an exception passing through a destructor.

For reference, this is one of two classes that gets used for multithreading support (from `Lib/python/pythreads.swg`):

```
class SWIG_Python_Thread_Allow {
  bool status;
  PyThreadState *save;
public:
  void end() { if (status) { PyEval_RestoreThread(save); status = false; }}
  SWIG_Python_Thread_Allow() : status(true), save(PyEval_SaveThread()) {}
  ~SWIG_Python_Thread_Allow() { end(); }
};
```

Given a Python application that has a main thread and several daemon threads running, if the main thread happens to exit, the other threads will be expected to exit as well, since they are daemon threads. If one of these threads happens to be in a C++ function call during this time, and `PyEval_SaveThread()` has been called, then when coming out of the C++ function call, the above class's `end()` function will be called, which will then call `PyEval_RestoreThread(save)`.  `PyEval_RestoreThread` then sees that this thread is supposed to exit and then calls `pthread_exit()`.

At this point, as part of a graceful thread exit, `pthread_exit()` triggers a forced unwinding, so that any cleanup functions that may be attached are executed. The way this works is by using a special exception (`abi::__forced_unwind` or the implementation-specific `__cxxabiv1::__forced_unwind`) so that it will go up the stack. It's expected that any try-catch block in the stack either doesn't catch this exception or rethrows it. A sample stack frame at this point is as follows:

```
#0  _Unwind_ForcedUnwind_Phase2 (exc=0x7fffeffffd70, context=0x7fffefffe050, frames_p=0x7fffefffdf58) at ../../../src/libgcc/unwind.inc:163
#1  0x00007ffff5c2cf4e in _Unwind_ForcedUnwind (exc=0x7fffeffffd70, stop=stop@entry=0x7ffff7fb2ab0 <unwind_stop>, stop_argument=0x7fffefffef10) at ../../../src/libgcc/unwind.inc:217
#2  0x00007ffff7fb2c30 in __GI___pthread_unwind (buf=<optimized out>) at unwind.c:121
#3  0x00007ffff7faa18c in __do_cancel () at pthreadP.h:310
#4  __pthread_exit (value=value@entry=0x0) at pthread_exit.c:28
#5  0x0000000000645df5 in PyThread_exit_thread () at ../Python/thread_pthread.h:373
#6  0x00000000004262ae in take_gil (tstate=0xc5c360) at ../Python/ceval_gil.h:224
#7  0x00000000005327b2 in PyEval_RestoreThread (tstate=0xc5c360) at ../Python/ceval.c:467
#8  0x00007ffff656a44c in SWIG_Python_Thread_Allow::end (this=0x7fffefffe2c0) at swsscommon_wrap.cpp:996
#9  SWIG_Python_Thread_Allow::end (this=0x7fffefffe2c0) at swsscommon_wrap.cpp:996
#10 _wrap_SonicV2Connector_Native_get (args=<optimized out>, kwargs=<optimized out>) at swsscommon_wrap.cpp:25848
#11 0x000000000053f350 in cfunction_call (func=<built-in method SonicV2Connector_Native_get of module object at remote 0x7ffff666a8b0>, args=<optimized out>, kwargs=<optimized out>) at ../Objects/methodobject.c:539
#12 0x000000000051d89b in _PyObject_MakeTpCall (tstate=0xc5c360, callable=<built-in method SonicV2Connector_Native_get of module object at remote 0x7ffff666a8b0>, args=<optimized out>, nargs=<optimized out>,
    keywords=<optimized out>) at ../Objects/call.c:191
```

When the unwinder reaches the stack frame for the `end()` function (or the one above that, I'm not entirely sure). Since this object is on the stack, it needs to be deallocated, so it calls the destructor, which then calls `end()`. Because `status` at this point is still `true`, it calls `PyEval_RestoreThread(save)` which goes through the same code path again and starts another forced unwinding, as seen in this stack trace:

```
#0  __cxxabiv1::__gxx_personality_v0 (version=<optimized out>, actions=10, exception_class=0, ue_header=0x7fffeffffd70, context=0x7fffefffe040) at ../../../../src/libstdc++-v3/libsupc++/eh_personality.cc:509
#1  0x00007ffff5c2c8a4 in _Unwind_ForcedUnwind_Phase2 (exc=0x7fffeffffd70, context=0x7fffefffe040, frames_p=0x7fffefffdf48) at ../../../src/libgcc/unwind.inc:182
#2  0x00007ffff5c2cf4e in _Unwind_ForcedUnwind (exc=0x7fffeffffd70, stop=stop@entry=0x7ffff7fb2ab0 <unwind_stop>, stop_argument=0x7fffefffef10) at ../../../src/libgcc/unwind.inc:217
#3  0x00007ffff7fb2c30 in __GI___pthread_unwind (buf=<optimized out>) at unwind.c:121
#4  0x00007ffff7faa18c in __do_cancel () at pthreadP.h:310
#5  __pthread_exit (value=value@entry=0x0) at pthread_exit.c:28
#6  0x0000000000645df5 in PyThread_exit_thread () at ../Python/thread_pthread.h:373
#7  0x00000000004262ae in take_gil (tstate=0xc5c360) at ../Python/ceval_gil.h:224
#8  0x00000000005327b2 in PyEval_RestoreThread (tstate=0xc5c360) at ../Python/ceval.c:467
#9  0x00007ffff64e223f in SWIG_Python_Thread_Allow::end (this=this@entry=0x7fffefffe2c0) at swsscommon_wrap.cpp:996
#10 0x00007ffff6509bae in SWIG_Python_Thread_Allow::end (this=0x7fffefffe2c0) at swsscommon_wrap.cpp:996
#11 SWIG_Python_Thread_Allow::~SWIG_Python_Thread_Allow (this=0x7fffefffe2c0, __in_chrg=<optimized out>) at swsscommon_wrap.cpp:998
#12 _wrap_SonicV2Connector_Native_get (args=<optimized out>, kwargs=<optimized out>) at swsscommon_wrap.cpp:25846
#13 0x000000000053f350 in cfunction_call (func=<built-in method SonicV2Connector_Native_get of module object at remote 0x7ffff666a8b0>, args=<optimized out>, kwargs=<optimized out>) at ../Objects/methodobject.c:539
#14 0x000000000051d89b in _PyObject_MakeTpCall (tstate=0xc5c360, callable=<built-in method SonicV2Connector_Native_get of module object at remote 0x7ffff666a8b0>, args=<optimized out>, nargs=<optimized out>,
    keywords=<optimized out>) at ../Objects/call.c:191
```

This time, the unwinder goes through the destructor, which is a problem. Because C++11 expects destructors to not propagate any exceptions up to the caller, and the forced unwinder is using `abi::__forced_unwind` exception, that exception isn't handled within the destructor, and thus `std::terminate` gets called.

The frequency that this error could happen depends on a number of factors (hardware, Python code flow, C++ code timing), so it's likely on the lower side, but I was able to repro this crash pretty reliably, as described in sonic-net/sonic-buildimage#12255 (note that in that case, `-threads` was not being used, but there was somewhat similar code being used to do the multithreading).

As for a fix, a quick fix I see in the code above would be to change the definition of `end()` to be `if (status) { status = false; PyEval_RestoreThread(save); }`. This, I think, will make sure that when the forced unwinder calls the destructor, it won't call `PyEval_RestoreThread(save)` via `end()` again. This appeared to work, but I don't know if there's other issues that it causes, such as race conditions.

saiarcot895:
If there are daemon thread(s) running, and the main thread exits, there is a chance of a SIGABRT happening if one of the daemon threads happens to be returning from a C call back to Python code. This is partly because of how the unwinder works (it uses a special exception `abi::__forced_unwind` to unwind the stack). If this exception happens to pass through the destructor of a C++ class (specifically, the `SWIG_Python_Thread_Allow` class here), then that is considered an error, and the unwinder will call `std::terminate`, resulting in a SIGABRT.

More details about this issue are in #2396.

Fixes #2396.

### DIFF
--- a/Lib/python/pythreads.swg
+++ b/Lib/python/pythreads.swg
@@ -28,7 +28,7 @@
          bool status;
          PyThreadState *save;
        public:
-         void end() { if (status) { PyEval_RestoreThread(save); status = false; }}
+         void end() { if (status) { status = false; PyEval_RestoreThread(save); }}
          SWIG_Python_Thread_Allow() : status(true), save(PyEval_SaveThread()) {}
          ~SWIG_Python_Thread_Allow() { end(); }
        };


saiarcot895:
Could someone review this?

wsfulton:
Seems reasonable to me. Unless there are any other thoughts, I will merge in a few days once I've finished some other work I'm doing.

PR #2428, opened 2022-11-04
Support python limited api when checking type of a PyObject

clintonstimpson:


clintonstimpson:
Note, I did `make check-python-examples` and `make check-python-test-suite` with the `#ifdef Py_LIMITED_API` temporarily replaced with `#if 1` to ensure tests still pass.

I'm assuming there is a bit of a performance overhead some folks would want to avoid, which is why this has an ifdef.
A previous experiment disabling the string compares and relying only on PyObjectType pointer comparisons resulted in a few tests failing.  So it looks like the pointer comparison succeeds most of the time, and the string compare is there as a fallback for less common cases.

ojwb:
I think this is essentially superseded by #2727.  The code currently on the branch there seems roughly equivalent, except it lacks an initial check for `Py_TYPE(op)` against `target_tp`/`SwigPyPacked_TypeOnce()` in the `Py_LIMITED_API` cases.  Is that something we want?

PR #2429, opened 2022-11-04
Use methods on tuple and list supported by the python limited api

clintonstimpson:


ojwb:
According to comments in the Python headers, what we're currently using are "Macros and static inline functions, trading safety for speed", so it seems the proposed change has performance implications which are imposed on all users, not just those who build the module to use the stable ABI.

We could instead achieve this by conditionally (on the stable ABI being enabled) defining replacements which forward to the stable ABI functions for cases like these.

clintonstimpson:
Yeah, I saw that #2190 uses conditionals, but it seemed there was still the concern of it affecting user's code.  Here's another attempt using conditionals, and also not affecting user's code.

ojwb:
If you mean the concern I raised about it there, I later concluded that was unfounded (see https://github.com/swig/swig/pull/2190#discussion_r799111767).  Calling `PyTuple_SetItem()` instead of `PyTuple_SET_ITEM()` should be safe for uses in user typemaps which follow that "should" in the Python C API docs (the other way round wouldn't be though).

wsfulton:
> If you mean the concern I raised about it there, I later concluded that was unfounded (see [#2190 (comment)](https://github.com/swig/swig/pull/2190#discussion_r799111767)). Calling `PyTuple_SetItem()` instead of `PyTuple_SET_ITEM()` should be safe for uses in user typemaps which follow that "should" in the Python C API docs (the other way round wouldn't be though).

Agreed. Given this has been merged to master now, am closing as resolved.

PR #2430, opened 2022-11-05
more tests for R

AndLLA:
Hallo,
I've ported to R a couple of test cases from python, around the "template" topic.
They all pass in the current implementation :smile: 

Thanks


wsfulton:
Thanks!

PR #2435, opened 2022-11-09
Fixing compile problem in std_array.i.

emmenlau:
This small PR fixes a build problem when using `std::array` in C#. It seems the current master version is broken.

### DIFF
--- a/Lib/csharp/std_array.i
+++ b/Lib/csharp/std_array.i
@@ -26,7 +26,7 @@
     foreach ($typemap(cstype, T) elem in c) {
       if (i >= end)
         break;
-      this[i++] = elem;
+      this[i++] = element;
     }
   }
 


ojwb:
Please run the testsuite locally before opening a PR.

Your change breaks the cpp11_std_array testcase:

```
cpp11_std_array/ArrayInt6.cs(70,19): error CS0103: The name `element' does not exist in the current context
Compilation failed: 1 error(s), 0 warnings
make[2]: *** [cpp11_std_array/../../../../Examples/Makefile:250: csharp_compile] Error 1
make[1]: *** [Makefile:67: cpp11_std_array.cpptest] Error 2
```

And indeed `element` doesn't seem to exist, whereas `elem` is the loop variable in the `foreach` just above the code you changed, so your change seems incorrect.

Can you provide a reproducer for the compile problem you are getting?

emmenlau:
> Please run the testsuite locally before opening a PR.

Please stop commanding people what to do.

emmenlau:
I would very much like to run the test suite. I enabled all cmake options I could find that may help to do that. But it seems I must have missed something.

wsfulton:
@emmenlau, cmake is not the developer's build system. You need to use the classic autotools. See https://swig.org/Doc4.1/Extending.html#Extending_test_suite for the test-suite.

emmenlau:
Thanks @wsfulton , I'll try to switch some of our SWIG integration builds to autotools. Since we support mostly Windows, we usually don't use autotools but it should be easy to add a few Linux build and test machines.

PR #2445, opened 2022-11-20
additional R tests

AndLLA:
Hallo,
I've translated few test cases from the python implementation.

they all pass under the current implementation

Thanks


wsfulton:
Nice, thanks. 

BTW, would you be able to help with an R test that I couldn't figure out how to implement if you do some more R test improvements? d8a028601247e96c15bbaac08100cce0401f186e adds unittest_sequence into unittest.R. It doesn't test that two sequences are equal very well. For example,
```
unittest_sequence(v[[1]], list(FALSE, TRUE, TRUE, TRUE, FALSE)) # Does not actually fail if no match, arg, needs fixing
```
If you modify the expected boolean values, it does not fail.

PR #2448, opened 2022-11-24
Rtest3

AndLLA:
Hallo,
this patch should make `unittest_sequence` more strict,
however note that:
 - I had to replace `list` with `c` because vectors in `R` are not exactly lists, therefore comparing vectors (returned by C++) with R lists complicates things a lot
 - I added `as.integer` because by default numbers are `numeric` and I guess we want to check the exact type returned by C++ which is `integer`
 - the first check fails because `make_vector_int` returns a vector of strings (containing numbers), not integers (in the previous version this test would pass because R does an implicit conversion)

I hope I got it right,

Thanks


ojwb:
R tests are failing in CI

AndLLA:
Hallo,
this is a "feature" because the `unittest_sequence` is now more strict and detects a problem with the first function:
```
v = make_vector_int()
```
instead of a vector of `int` it returns a vector of strings (representing the integer numbers).

the motivation behind this pull request is a comment by @wsfulton in a previous pull request: https://github.com/swig/swig/pull/2445#issuecomment-1325668914

Thanks

ojwb:
We simply can't merge a PR that will break CI.

As part of this PR, either the testcase this breaks needs fixing (best) or disabling (not ideal, but if I follow it's only passing currently because of a bug in this helper...)

AndLLA:
Hallo,
I disabled the test in the code, indicating why it fails.
I'm not sure how to fix the underlying bug in swig, 
I guess we need the expert :smile:  !!!!

Thanks

AndLLA:
Hallo,
Actually the fix was trivial, 
now it works :100: 

I hope this is what was intended 

Thanks

ojwb:
Thanks.  It does seem strange that the function is called `make_vector_int` but returns `std::vector<std::string>` but maybe it's trying to test that implicit conversion between string and integer works in this situation so I'm not really sure.  This testcase only has a "runme" for R, so that's not helpful in deciding.

@wsfulton You added this testcase in d8a028601247e96c15bbaac08100cce0401f186e so I think you need to decide if this is the right fix.

wsfulton:
Thanks for spotting the bug in my testcase! I'll test this out in my next round of merges, but it seems good after a quick look! Thanks.

Issue #2431, opened 2022-11-07
Building 4.1.0 with cmake on Mac 


snarkhunter:
After building 4.1.0 with cmake 3.23.4 on macOS 11.7, swig fails on a simple input file.  The input file is just
```
%module dummy
void f();
```

Running `swig -python -c++ -o dummy.cxx dummy.i`  yields
`/Users/langer/stow/swig-4.1.0/share/swig/4.1.0/swig.swg:431: Error: Syntax error in input(1).`

If swig is built with configure and make, it works properly.  It also worked properly with cmake 3.23.4 on macOS 12.6.

jschueller:
can you try master branch ?

jschueller:
bison 2.3 is pretty old, maybe try a more recent version ?

snarkhunter:
Yikes.  macOS ships with bison 2.3.  I installed 3.8.2 from MacPorts, but get the same error.  CMake does confirm that it's using 3.8.2.
 -- Steve

From: Julien Schueller ***@***.***>
Reply-To: swig/swig ***@***.***>
Date: Thursday, November 10, 2022 at 9:51 AM
To: swig/swig ***@***.***>
Cc: "Langer, Stephen A. (Fed)" ***@***.***>, Author ***@***.***>
Subject: Re: [swig/swig] Building 4.1.0 with cmake on Mac (Issue #2431)


bison 2.3 is pretty old, maybe try a more recent version ?

jschueller:
I just fired a job on macos, everything is ok: https://github.com/jschueller/swig-mac, maybe something was wrong with that bison setup ?

jschueller:
what the minimal bison version version swig requires ?

snarkhunter:
I'm still getting the LESSEQUALGREATER error mentioned above.  I updated the master branch from github and ran cmake in a clean build directory.

jschueller:
please post full cmake log then & make log (make VERBOSE=1)

snarkhunter:
I started again with a completely new clone from github, and it all worked.  Could the LESSEQUALGREATER error from cmake have been caused by something leftover from building with configure/make? 

I tried to reproduce the sequence of steps I had used before in the new clone, but now it won't build with configure.   The new clone has no configure file.  I generated one with autoconf, but it complains about missing config.guess and config.sub.

 -- Steve

jschueller:
yes, definetely, it is generated in the source dir
try "autoreconf -vfi"

[...]

PR #2452, opened 2022-11-28
CMake: Require bison 3.x

jschueller:
Lets require bison 3.x, this is old enough anyways and might avoid https://github.com/swig/swig/issues/2431


wsfulton:
I don't think this is necessary given #2431 is now resolved and we don't have a minimum version of Bison required.

jschueller:
fine with me

PR #2454, opened 2022-11-28
Fix push/pop mismatch

berolinux:
Without this, perlhead.swg does
`#pragma GCC diagnostic push` only if `__GNUC__ >= 10` and `defined(__cplusplus)`, but does `#pragma GCC diagnostic pop` if `__GNUC__ >= 10` without the check for __cplusplus.

This results in a `#pragma GCC diagnostic pop` without a prior push in the `__GNUC__ >= 10` and not `__cplusplus` case, triggering compiler warnings.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,9 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.2.0 (in progress)
 ===========================
 
+2022-11-29: bero
+            Fix mismatch between #pragma GCC diagnostic push and pop statements
+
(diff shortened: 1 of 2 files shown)


berolinux:
Patch updated to also add the missing `#pragma GCC diagnostic push` statement instead of just fixing the mismatch between conditions

berolinux:
Failed CI checks are unrelated to the patch. (The patch only touches perlhead.swg, but failures are entirely unrelated to perl)

wsfulton:
Thanks, I'll add to 4.1.1 release due out tomorrow too.

PR #2455, opened 2022-11-30
[js] Turn on C++ output for node too

ojwb:
Nodejs is like V8 and needs C++ output enabled when wrapping C code.

The testsuite was masking this bug by using SWIG options `-v8 -DBUILDING_NODE_EXTENSION=1` rather than `-node` when testing with nodejs, while the javascript examples currently all seem to all get processed with -c++.


ojwb:
CI failure seems to be because I used `elseif...` which the GNU make manual documents but doesn't mention it's kind of new (3.81 in 2006).  I'm a bit surprised to see so old a GNU make on an Ubuntu image in CI, but we don't appear to use this feature elsewhere in SWIG and it's easy to avoid.

ojwb:
> while the javascript examples currently all seem to all get processed with -c++.

I still need (or at least ought) to address this part.

ojwb:
I sorted out the Javascript examples back in December.  I see there were rapid failures for the jsc jobs, but the logs have expired so I can't see what they actually were now.

I've rebased and pushed.

ojwb:
Weird - jsc is still failing in CI but now works for me locally...

ojwb:
Rebased, resolved issues with the examples changes, and pushed to git master as d3b17437b12625bd31118f40e06a3bb28b6c5fda and the commits preceding it.

PR #2458, opened 2022-12-05
[OCaml] Runtime tests for rename_rstrip_encoder and using2

ZackerySpytz:
Add runtime tests for rename_rstrip_encoder and using2.

They are based on the runtime tests that already exist for other languages.

### DIFF
--- a/Examples/test-suite/ocaml/rename_rstrip_encoder_runme.ml
+++ b/Examples/test-suite/ocaml/rename_rstrip_encoder_runme.ml
@@ -1,0 +1,5 @@
+open Swig
+open Rename_rstrip_encoder
+
+let s = new_SomeThing '()
+let a = new_AnotherThing '()
--- a/Examples/test-suite/ocaml/using2_runme.ml
+++ b/Examples/test-suite/ocaml/using2_runme.ml
@@ -1,0 +1,4 @@
+open Swig
+open Using2
+
+let _ = assert (_spam '(37) as int = 37)


Examples/test-suite/ocaml/rename_rstrip_encoder_runme.ml
```diff
@@ -0,0 +1,5 @@
+open Swig
+open Rename_rstrip_encoder
+
+let s = new_SomeThing '()
+let a = new_AnotherThing '()
```
ojwb, line 5: `Examples/test-suite/python/rename_rstrip_encoder_runme.py` also checks that the method is correctly renamed (it contains `Cls` twice, but only the suffix `Cls` should be removed):

```
from rename_rstrip_encoder import *

s = SomeThing()
a = AnotherThing()
a.DoClsX()
```

It seems useful to test this here too - it does seem this is correctly wrapped for ocaml so it should just be a matter of writing a suitable method call (not sure what the syntax is for that).  

```
$ rgrep DoClsX .
./rename_rstrip_encoder.ml:external _AnotherThing_DoClsX_f : c_obj list -> c_obj list = "_wrap_AnotherThing_DoClsXrename_rstrip_encoder" ;;
./rename_rstrip_encoder.ml:let _AnotherThing_DoClsX arg = match _AnotherThing_DoClsX_f ((fnhelper arg)) with
./rename_rstrip_encoder.ml:	      "DoClsX", _AnotherThing_DoClsX ;
```

wsfulton: b7f64c4da55ed6faacbbfa665825e5d1326aa8ee contains the requested addition to the testcase.

PR #2475, opened 2023-01-05
Fix for #1854: Using a single definition for the import dll in c#

### LINKED ISSUE #1854: C# improved generated PINVOKE code
The `modulePINVOKE.cs` generated file is repeating the same `global::System.Runtime.InteropServices.DllImport` all over the place, using the exact same dllname as first argument.
I know it's not recommended to edit this file, but when using SWIG in a cross-platform project (mainly Xamarin), I'm not sure there is another choice. For instance, the _dllname_ is different on windows/linux/android and on iOS. You have to use a framework for iOS, so the _dllname_ has to be `module.framework/module` instead of just `module`.

It would be a lot easier if the name of the dll is stored in a static variable at the start of `modulePINVOKE.cs` and used in the file (besides being cleaner). That way, one could easily add some `#ifdef` directive just once in the file (and add it again after each file generation).

christophe-calmejane:
Hi,

this is a proposal to fix #1854 by using a single definition for the `dllimport` variable so it's easier to edit the file manually if needed.

### DIFF
--- a/Lib/csharp/csharphead.swg
+++ b/Lib/csharp/csharphead.swg
@@ -149,7 +149,7 @@ SWIGEXPORT void SWIGSTDCALL SWIGRegisterExceptionArgumentCallbacks_$module(
     static ExceptionArgumentDelegate argumentNullDelegate = new ExceptionArgumentDelegate(SetPendingArgumentNullException);
     static ExceptionArgumentDelegate argumentOutOfRangeDelegate = new ExceptionArgumentDelegate(SetPendingArgumentOutOfRangeException);
 
-    [global::System.Runtime.InteropServices.DllImport("$dllimport", EntryPoint="SWIGRegisterExceptionCallbacks_$module")]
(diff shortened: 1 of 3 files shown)


christophe-calmejane:
Hi, any news about merging this one?
I don't know why some checks failed, I don't think it's related to my changes.
Thanks!

ojwb:
I was just try to keep the bug tracker usable by tagging tickets appropriately.  You need @wsfulton to review this.

wsfulton:
I'm afraid I cannot accept this patch as first and foremost:
* SWIG already has a mechanism to modify the C# dll name using the `-dllimport` command line option.

Additionally:
* This patch will break code for users that rely on `$dllimport` being replaced by the name of the dll, as documented in CSharp.html.
* The replacement variable name is only available in the PINVOKE file and not in every .cs file where it is currently possible to use `$dllimport`.

PR #2478, opened 2023-01-10
Lib/csharp: Better standardized std_vector.i and std_array.i

emmenlau:
This is a minor change that should mostly affect the whitespaces. I compared `std_vector.i` and `std_array.i` for csharp and found that the two files are closely related, yet different. Therefore I tried to unify and standardize the two files against each other, to help with maintenance. Is that reasonable?

### DIFF
--- a/Lib/csharp/std_array.i
+++ b/Lib/csharp/std_array.i
@@ -6,37 +6,47 @@
  * The C# wrapper is made to look and feel like a C# System.Collections.Generic.IReadOnlyList<> collection.
  * ----------------------------------------------------------------------------- */
 
-%{
-#include <algorithm>
-#include <array>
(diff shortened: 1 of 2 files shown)


wsfulton:
Addressed all the issues in 20f7309. Note that I removed the added constructor in this pull request as it broke the test-suite with:
```
../../../../../Examples/test-suite/csharp/cpp11_std_array_runme.cs(11,16): error CS0121: The call is ambiguous between the following methods or properties: `cpp11_std_arrayNamespace.ArrayInt6.ArrayInt6(System.Collections.ICollection)' and `cpp11_std_arrayNamespace.ArrayInt6.ArrayInt6(System.Collections.Generic.IEnumerable<int>)'
```

PR #2479, opened 2023-01-10
Lib/csharp: Resolve a few warnings about unused parameters

emmenlau:
This is a minor PR. There are some methods that lead to generated code where a parameter is unused. When compiled with pedantic compiler settings, this leads to a warning or error.
This PR declares the additional parameter `self` as optional (or unused).


wsfulton:
`$self` should be used instead of `self`. I'll correct and merge shortly.

PR #2492, opened 2023-01-26
[Python] Avoid unused parameter self warning

jschueller:
Avoids lots of warnings on regular functions (with gcc / -Wall -Wextra), like:
```
/home/schueller/projects/swig-cmake-example/build/CMakeFiles/openturns.dir/openturnsPYTHON_wrap.cxx:4632:47: warning: unused parameter 'self' [-Wunused-parameter]
SWIGINTERN PyObject *_wrap_Mesh_run(PyObject *self, PyObject *args) {
```

there is an instruction that uses the "self" argument only with the -builtin swig flag, so the condition must be inverted:
```
res1 = SWIG_ConvertPtr(self, &argp1,SWIGTYPE_p_OT__Mesh, 0 |  0 );
```

to reproduce: https://github.com/jschueller/swig-cmake-example (configure with cmake -DCMAKE_CXX_FLAGS="-Wall -Wextra")

fixes https://github.com/swig/swig/commit/7cb2f46e06fef59a3fe2fb5283bc31326a091f8a

cc @ojwb

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -2814,7 +2814,7 @@ public:
       Printv(f->def, linkage, wrap_return, wname, "(PyObject *self, PyObject *args, PyObject *kwargs) {", NIL);
     }
 
-    if (builtin) {
+    if (!builtin) {
       /* Avoid warning if the self parameter is not used. */
       Append(f->code, "(void)self;\n");
     }


ojwb:
It'd be much more helpful to have a reproducer that just uses my existing built SWIG source tree, and didn't require me to have to check out and build another SWIG tree (while dealing with cmake's deficiencies).

I don't think the condition is inverted as the point of the original PR (#801) was to address unused warnings with `-builtin`, so the patch here doesn't seem right.  I could believe we want to unconditionally emit `(void)self;` though.

jschueller:
fair enough, it can be reproduced on the class example:
```
$ cd Examples/python/class
$ swig -c++ -python example.i
$ g++ example_wrap.cxx -Wall -Wextra -I /usr/include/python3.10/ -c
example_wrap.cxx: In function 'PyObject* _wrap_delete_Shape(PyObject*, PyObject*)':
example_wrap.cxx:3307:51: warning: unused parameter 'self' [-Wunused-parameter]
 3307 | SWIGINTERN PyObject *_wrap_delete_Shape(PyObject *self, PyObject *args) {
```

ojwb:
Thanks - rereading the original PR more thoroughly it seems `-builtin` was only mentioned in a comment by wsfulton:

> From what I can make out, you are trying to squash these warnings for -builtin where self is sometimes used in the wrapper function and sometimes not

I don't think that's actually right, but led to me getting the condition reversed in my change.

Just trying all the examples and testsuite with `CXXFLAGS='-Werror=unused' SWIGOPT=-builtin` with your change applied - I'm seeing some errors (promoted from warnings), but none for `self` so far.

ojwb:
That looked good so merged, and added `CHANGES.current` entry in 1ee9af51.

I think this should probably go in 4.1.2 so I've set the milestone and assigned @wsfulton.  Doesn't seem to be a way to reopen though...

PR #2493, opened 2023-01-26
[Python] Always define SwigMethods_proxydocs

jschueller:
This reverts #2222 as now with 4.1 I get:
error: use of undeclared identifier 'SwigMethods_proxydocs'


jschueller:
nvm

PR #2494, opened 2023-01-26
python/cast_mode: Fix integer overflow of LONG_MAX+1.

degasus:
LONG_MAX cannot be represented as double if sizeof(long)=8, so it is usually rounded to LONG_MAX+1. So SWIG_CanCastAsInteger returned true for LONG_MAX+1 and the next cast ended up in an integer overflow.

Worse for unsigned long: It is called by smaller unsigned integers with range checks after the convertion to unsigned long. As the integer overflow yields a 0x0, this is valid for all small unsigned integers. Smaller signed integers will always fail this range check after the integer overflow.

This fixes the tests on my machine:
```
  as_l(lmaxd + LSB)
  as_ll(llmaxd + LSB)
  as_ul(ulmax + LSB)
  as_ull(ullmax + LSB)
```

Wrong old results before this commit:
```
  as_l(float(2**63)) := -0x8000000000000000
  as_ll(float(2**63)) := -0x8000000000000000
  as_uc(float(2**64)) := 0x0
  as_us(float(2**64)) := 0x0
  as_ui(float(2**64)) := 0x0
  as_ul(float(2**64)) := 0x0
  as_ull(float(2**64)) := 0x0
```

Now all of them throw an exception.

Also add tests for the min/max values of this convertions including negative tests for min-1/max+1.

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,17 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.2.0 (in progress)
 ===========================
 
+2023-XX-XX: degasus
+	    [Python] #2494 Fix integer overflow / undefined behavior in the python cast_mode on
+	    sizeof(long)==8 platforms for the implicit conversion of the edge cases:
+	      long(double(LONG_MAX + 1))
+	      long long(double(LONG_MAX + 1))
+	      unsigned char(double(ULONG_MAX + 1))
+	      unsigned short(double(ULONG_MAX + 1))
+	      unsigned int(double(ULONG_MAX + 1))
+	      unsigned long(double(ULONG_MAX + 1))
+	      unsigned long long(double(ULONG_MAX + 1))
+
 2023-12-03: olly
 	    [Ocaml] Remove -suffix command line option which has emitted a
 	    deprecation warning since SWIG 3.0.4 - if you want to specify
--- a/Examples/test-suite/overload_simple.i
+++ b/Examples/test-suite/overload_simple.i
@@ -166,10 +166,17 @@ char *fid(double, double) {
 %}
 
 %inline %{
-unsigned long long ull() { return 0ULL; }
-unsigned long long ull(unsigned long long ull) { return ull; }
-long long ll() { return 0LL; }
-long long ll(long long ull) { return ull; }
+int sizeof_long() { return sizeof(long); }
+
+unsigned long as_ul() { return 0UL; }
+unsigned long as_ul(unsigned long ul) { return ul; }
+long as_l() { return 0L; }
+long as_l(long l) { return l; }
+
+unsigned long long as_ull() { return 0ULL; }
+unsigned long long as_ull(unsigned long long ull) { return ull; }
+long long as_ll() { return 0LL; }
+long long as_ll(long long ll) { return ll; }
 %}
 
 %include cmalloc.i
--- a/Examples/test-suite/python/python_overload_simple_cast_runme.py
+++ b/Examples/test-suite/python/python_overload_simple_cast_runme.py
@@ -164,34 +164,129 @@ if s.type != "void *":
     raise RuntimeError("Spam(void *)")
 
 
+# nextafter: ++ and -- operators for float, off by one LSB
+# nextafter was released in Python 3.9
+from sys import hexversion
+if hexversion >= 0x3090000:
+    from math import inf, nextafter
+else:
+    # workaround: try to load nextafter from numpy if available
+    try:
+        from numpy import nextafter
+    except:
+        # else just disable this tests
+        def nextafter(x, y):
+            return None
+
+    # math.inf was added in Python 3.5
+    inf = float('inf')
+
+def exceptMatch(fun, arg, res, msg):
+    if arg is None or res is None:
+        # nextafter is missing, so skipping this test
+        return
+    if fun(arg) != res:
+        raise RuntimeError(msg)
+
+def exceptTypeError(fun, arg, msg):
+    if arg is None:
+        # nextafter is missing, so skipping this test
+        return
+    try:
+        fun(arg)
+        raise RuntimeError(msg)
+    except TypeError:
+        pass
+
+# x86_64: long is 32bit on MSVC but 64bit on *nix
+assert sizeof_long() in [4, 8]
+
+# unsigned long
+ulmax = 2**32 - 1
+ulmin = 0
+ulmaxd = float(2**32 - 1)
+ulmind = 0.0
+if sizeof_long() == 8:
+    ulmax = 2**64 - 1
+    ulmaxd = nextafter(float(2**64), 0.0)
+
+exceptMatch(as_ul, ulmin, ulmin, "as_ul(ulmin)")
+exceptMatch(as_ul, ulmax, ulmax, "as_ul(ulmax)")
+exceptMatch(as_ul, ulmind, ulmind, "as_ul(ulmind)")
+exceptMatch(as_ul, ulmaxd, ulmaxd, "as_ul(ulmaxd)")
+
+exceptTypeError(as_ul, ulmin - 1, "as_ul(ulmin - 1)")
+exceptTypeError(as_ul, ulmax + 1, "as_ul(ulmax + 1)")
+exceptTypeError(as_ul, nextafter(ulmind, -inf), "as_ul(ulmind - LSB)")
+exceptTypeError(as_ul, nextafter(ulmaxd, inf), "as_ul(ulmaxd + LSB)")
+
+# long
+lmax = 2**31 - 1
+lmin = -2**31
+lmaxd = float(2**31 - 1)
+lmind = float(-2**31)
+lmaxd_v = lmaxd # expected value after the cast
+lmind_v = lmind
+if hexversion < 0x30a0000:
+    # PyLong_AsLong(float) truncated the input before 3.10
+    lmaxd = nextafter(float(2**31), 0.0)
+    lmind = nextafter(float(-2**31 - 1), 0.0)
+    lmaxd_v = float(2**31 - 1)
+    lmind_v = float(-2**31)
+if sizeof_long() == 8:
+    lmax = 2**63 - 1
+    lmin = -2**63
+    lmaxd = nextafter(float(2**63), 0.0)
+    lmind = float(-2**63)
+    lmaxd_v = lmaxd
+    lmind_v = lmind
+
+exceptMatch(as_l, lmin, lmin, "as_l(lmin)")
+exceptMatch(as_l, lmax, lmax, "as_l(lmax)")
+exceptMatch(as_l, lmind, lmind_v, "as_l(lmind)")
+exceptMatch(as_l, lmaxd, lmaxd_v, "as_l(lmaxd)")
+
+exceptTypeError(as_l, lmin - 1, "as_l(lmin - 1)")
+exceptTypeError(as_l, lmax + 1, "as_l(lmax + 1)")
+exceptTypeError(as_l, nextafter(lmind, -inf), "as_l(lmind - LSB)")
+exceptTypeError(as_l, nextafter(lmaxd, inf), "as_l(lmaxd + LSB)")
+
 # unsigned long long
-ullmax = 9223372036854775807  # 0xffffffffffffffff
-ullmaxd = 9007199254740992.0
+ullmax = 2**64 - 1
 ullmin = 0
+ullmaxd = float(2**53) # 64 bit double significand
 ullmind = 0.0
-if ull(ullmin) != ullmin:
-    raise RuntimeError("ull(ullmin)")
-if ull(ullmax) != ullmax:
-    raise RuntimeError("ull(ullmax)")
-if ull(ullmind) != ullmind:
-    raise RuntimeError("ull(ullmind)")
-if ull(ullmaxd) != ullmaxd:
-    raise RuntimeError("ull(ullmaxd)")
+if sizeof_long() == 8:
+    ullmaxd = nextafter(float(2**64), 0.0)
+
+exceptMatch(as_ull, ullmin, ullmin, "as_ull(ullmin)")
+exceptMatch(as_ull, ullmax, ullmax, "as_ull(ullmax)")
+exceptMatch(as_ull, ullmind, ullmind, "as_ull(ullmind)")
+exceptMatch(as_ull, ullmaxd, ullmaxd, "as_ull(ullmaxd)")
+
+exceptTypeError(as_ull, ullmin - 1, "as_ull(ullmin - 1)")
+exceptTypeError(as_ull, ullmax + 1, "as_ull(ullmax + 1)")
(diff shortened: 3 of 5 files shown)


Examples/test-suite/python/python_overload_simple_cast_runme.py
```diff
@@ -164,34 +164,120 @@ def __float__(self):
...
+if as_l(lmaxd) != lmaxd:
+    raise RuntimeError("as_l(lmaxd)")
+
+exceptTypeError(as_l, lmin - 1, "as_l(lmin - 1)")
+exceptTypeError(as_l, lmax + 1, "as_l(lmax + 1)")
+exceptTypeError(as_l, nextafter(lmind, -inf), "as_l(lmind - LSB)")
+exceptTypeError(as_l, nextafter(lmaxd, inf), "as_l(lmaxd + LSB)")
+
 # unsigned long long
-ullmax = 9223372036854775807  # 0xffffffffffffffff
```
degasus, line 168: Note: This does NOT match. The old value `9223372036854775807 == 2**63-1` conflicts with the comment.

Lib/python/pyprimtypes.swg
```diff
@@ -104,7 +104,9 @@ SWIG_AsVal_dec(long)(PyObject *obj, long* val)
     if (!dispatch) {
       double d;
       int res = SWIG_AddCast(SWIG_AsVal(double)(obj,&d));
-      if (SWIG_IsOK(res) && SWIG_CanCastAsInteger(&d, LONG_MIN, LONG_MAX)) {
+      // largest double not larger than LONG_MAX
+      const double long_max = sizeof(long) == 8 ? 0x7ffffffffffffc00LL : LONG_MAX;
```
degasus, line 108: I really hate hardcoding this two constants. Howver I was not able to generate this constant in C with less than 10 lines of code without any a-prioi-knowledge. I anyone is aware of a method to convert a 64 bit integer > 2**53 to double with a given rounding behavior, I will happily update this patch.

How to generate:
  C: nextafter(uint128_t(LONG_MAX)+1, 0)
  Python: hex(int(nextafter(2**63, 0)))

Examples/test-suite/overload_simple.i
```diff
@@ -166,10 +166,17 @@ char *fid(double, double) {
...
-unsigned long long ull() { return 0ULL; }
-unsigned long long ull(unsigned long long ull) { return ull; }
-long long ll() { return 0LL; }
-long long ll(long long ull) { return ull; }
+int sizeof_long() { return sizeof(long); }
+
+unsigned long as_ul() { return 0UL; }
+unsigned long as_ul(unsigned long ul) { return ul; }
+long as_l() { return 0L; }
+long as_l(long l) { return l; }
```
degasus, line 174: Just `l` was not allowed because of https://www.flake8rules.com/rules/E743.html , so I've renamed it to as_l.

Examples/test-suite/python/python_overload_simple_cast_runme.py
```diff
@@ -164,34 +164,120 @@ def __float__(self):
...
+    raise RuntimeError("as_ul(ulmax)")
+if as_ul(ulmind) != ulmind:
+    raise RuntimeError("as_ul(ulmind)")
+if as_ul(ulmaxd) != ulmaxd:
+    raise RuntimeError("as_ul(ulmaxd)")
+
+exceptTypeError(as_ul, ulmin - 1, "as_ul(ulmin - 1)")
+exceptTypeError(as_ul, ulmax + 1, "as_ul(ulmax + 1)")
+exceptTypeError(as_ul, nextafter(ulmind, -inf), "as_ul(ulmind - LSB)")
+exceptTypeError(as_ul, nextafter(ulmind, inf), "as_ul(ulmaxd + LSB)")
```
degasus, line 210: ulmaxd instead of ulmind, also with as_ull

degasus:
@wsfulton This patch passed the CI, after rebase on top of your commits, it now fails. As master has failed the CI as well, I doubt there is a need for me to invastigate further.

By the way, is there something you want me to do here? It fixes an obvious bug without touching the behavior elsewhere. And it increases the test coverage.

Lib/python/pyprimtypes.swg
ojwb: This also seems to assume IEEE double, which is pretty much ubiquitous in current platforms but not actually required by the language standards.

It does seem fiddly to calculate though.

CHANGES.current
```diff
@@ -7,6 +7,17 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.2.0 (in progress)
 ===========================
 
+2023-XX-XX: degasus
+            Fix integer overflow / undefined behavior in the python cast_mode on sizeof(long)==8
+            plattforms for the implicit convertion of the edge cases:
```
ojwb, line 12: Typo: "plattforms"

ojwb:
BTW, please don't rebase and push your PRs so much - every time you do it kicks off 100 or so builds, which is a lot of CI time to be burning when a patch is really just waiting for someone to get to it.  It can also delay CI runs for things people are actively working on since GHA limits us to fewer concurrent jobs than we kick off per push.

Lib/python/pyprimtypes.swg
ojwb: If we have `uint128_t` (or if `sizeof(long) == 4` and we have `uint64_t`) and we're building as C99 or have POSIX functions then we could just use `nextafter((uint128_t)LONG_MAX+1, 0)` (or `nextafter((uint64_t)LONG_MAX+1, 0)`) - checking on godbolt it seems GCC and MSVC both evaluate this at compile time (though clang doesn't seem to currently).  We could also check `FLT_RADIX` and `DBL_MANT_DIG` match IEEE and use a precomputed value for common platforms too.

Many platforms lack `uint128_t` though, and I'd guess a platform not using IEEE floating point is probably especially unlikely to.

Some other assumptions would be easy to fix though, e.g.:

```
llmaxd = float(2**53) # 64 bit double significand
```

We could add C helpers to return `FLT_RADIX` and `DBL_MANT_DIG` (from `<float.h>`) and replace this with something like:

```
llmaxd = float(flt_radix()**dbl_mant_dig()) # 64 bit double significand
```

I'm not sure if people would actually be trying to use SWIG on a platform with non-IEEE floating point though.

@wsfulton Thoughts?

degasus:
As this is a bug fix, what do you think about adding a milestone to it?

If there is anything I can do to help on the decision about hard coding the IEEE float behavior, please let me know.

CHANGES.current
ojwb: Also "convertion" (should be "conversion").

And the convention is to "tag" the entry with the language it applies to like so:

```
2023-12-01: saiarcot895
            [Python] #2413 Prevent potential multi-threading crash; gracefully exit running
            daemon threads on main thread exit.
```

Examples/test-suite/overload_simple.i
```diff
@@ -166,10 +166,21 @@ char *fid(double, double) {
...
 %inline %{
-unsigned long long ull() { return 0ULL; }
-unsigned long long ull(unsigned long long ull) { return ull; }
-long long ll() { return 0LL; }
-long long ll(long long ull) { return ull; }
+#include <errno.h>
+#include <math.h>
+
+int sizeof_long() { return sizeof(long); }
+double getnextafter(double from, double to) { double ret = nextafter(from, to); errno = 0; return ret; }
```
ojwb, line 173: `nextafter()` is C99 so this testcase may fail if using a C90 compiler which is still the baseline we aim to support.

I don't think we have a "C99-only testcase" mechanism, but `nextafter()` is also C++11 and we do have a "C++11-only testcase" mechanism so this could be made a C++11 testcase (rename to add a `cpp11_` prefix and move it to `CPP11_TEST_CASES`).  Or just drop this helper and skip the parts of the `_runme.py` that requires it if using Python < 3.9.

degasus: I've dropped this C helpers and added an optional fallback to numpy instead. If both is missing, the tests will be skipped. I hope the coverage is still good enough.
Our behavior changed with Python >= 3.10, so having at least 3.9 should still test all required cases.

CHANGES.current
degasus: Done.
Nitpicking: Which convention should I follow for the indentation? 12 spaces or one tab + 4 spaces?

And as usual, sorry for my bad English ;-)

Lib/python/pyprimtypes.swg
degasus: I sadly still have no better option than hardcoding this constants in the generated C code. And yes, this is hardcoded for IEEE floats only. Do you want me to add some macros to check if the predefined macros in float.h matches our expections and #error else?

However I skipped your suggestion for generating `llmaxd`. This is within the test, and for the cast_mode feature, the test is currently more a documentation of the behavior than checking for the designed behavior. I fear that importing the size of floats from C makes this just more unreadable.

degasus:
> I've added it to the 4.2 milestone for consideration (but note that's not a guarantee it'll get merged for 4.2.0).

Thanks a lot. Obviously no guarantee for anything, but this makes it less likely to just be forgotten :-)

ojwb:
> Note: The failing test also fails on master.

It's marked as "can fail" (it's due to a bug in upstream ruby which has been fixed and we're just waiting for that fix to percolate through to the ruby 3.3 development builds we're using in CI).

It shows as green in https://github.com/swig/swig/actions but github is stupid and still treats that as a failing test in PRs.

CHANGES.current
ojwb: SWIG coding style expects a tab to move to the next multiple of 8 column so they're equivalent.

Examples/test-suite/python/python_overload_simple_cast_runme.py
wsfulton: Agreed, `9223372036854775807 == 0x7fffffffffffffff`

Examples/test-suite/overload_simple.i
wsfulton: Hard to tell from the diffs displayed here, but `ull()` is now `as_ull()` and `ull()` is now `as_ll()`.

Lib/python/pyprimtypes.swg
wsfulton: Agreed that hard coding is not ideal, but given it is not easily calculated portably, I've put in a comment on how to calculate it as the next best thing. See eac27f7d3ecfe0c120f202eb1a1da081060884f6.

Examples/test-suite/python/python_overload_simple_cast_runme.py
```diff
@@ -164,34 +164,129 @@ def __float__(self):
...
+lmaxd = float(2**31 - 1)
+lmind = float(-2**31)
+lmaxd_v = lmaxd # expected value after the cast
+lmind_v = lmind
+if hexversion < 0x30a0000:
+    # PyLong_AsLong(float) truncated the input before 3.10
+    lmaxd = nextafter(float(2**31), 0.0)
+    lmind = nextafter(float(-2**31 - 1), 0.0)
+    lmaxd_v = float(2**31 - 1)
+    lmind_v = float(-2**31)
```
wsfulton, line 235: I ran the tests on 32-bit Python < 3.10 to see that this change is required, but I didn't figure out why. Do you have a link to the exact release note that fixes this truncation problem?

wsfulton:
Thanks for this pull request @degasus . It took a while digesting and checking it, it must have taken you even longer to figure out the problem and fix. There's just one comment on the 32-bit Python if you can take a look.

Examples/test-suite/python/python_overload_simple_cast_runme.py
```diff
@@ -164,34 +164,129 @@ def __float__(self):
...
+exceptTypeError(as_l, lmax + 1, "as_l(lmax + 1)")
+exceptTypeError(as_l, nextafter(lmind, -inf), "as_l(lmind - LSB)")
+exceptTypeError(as_l, nextafter(lmaxd, inf), "as_l(lmaxd + LSB)")
+
 # unsigned long long
-ullmax = 9223372036854775807  # 0xffffffffffffffff
-ullmaxd = 9007199254740992.0
+ullmax = 2**64 - 1
 ullmin = 0
+ullmaxd = float(2**53) # 64 bit double significand
```
wsfulton, line 257: Also, why test 2**53 for `unsigned long long`, this surely should be `nextafter(float(2**64), 0.0)`, that is, the same as 64-bit `unsigned long` as `sizeof(unsigned long) == sizeof(unsigned long long)` on 64-bit platforms.

Examples/test-suite/python/python_overload_simple_cast_runme.py
degasus: https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong
Before 3.10, `__int__` was used and it truncated any float. Since 3.10, it only tries to use `__index__`, which fails for float. I think we might want to call it a bug to use `PyLong_AsLong` for a float, but this defines the current behavior: https://github.com/swig/swig/blob/master/Lib/python/pyprimtypes.swg#L97

Examples/test-suite/python/python_overload_simple_cast_runme.py
degasus: I think this behavior is implemented on purpose [here](https://github.com/swig/swig/blob/master/Lib/python/pyprimtypes.swg#L291) or [here](https://github.com/swig/swig/blob/master/Lib/perl5/perlprimtypes.swg#L232). By the way, this reminds me to double check the behavior on perl if it has the same bug as fixed here...

cast_mode (sometimes) check for how many LSBs the number is off, and it (sometimes) check for 53b precision. In my opinion, there are reasons to do so and reasons not to do so. I just dislike the "sometimes" based on plattforms. If you have an opinion on which behavior you like more, I will be very happily rewrite #2097 to actually get the same behavior on all plattforms. But I think this should not be considered for the current release.

degasus:
> Thanks for this pull request @degasus . It took a while digesting and checking it, it must have taken you even longer to figure out the problem and fix. There's just one comment on the 32-bit Python if you can take a look.

Thanks a lot for the review, and marry x-mas. What a honor to get your attention at this time!

Figuring out the problem was actually easy, clang provided a warning that the constant LONG_MAX casted to double does not exist. Figuring out the behavior on all plattforms however took a while ;-)

Examples/test-suite/python/python_overload_simple_cast_runme.py
degasus: ..... Sorry for missing to check perl ....  https://github.com/swig/swig/blob/master/Lib/perl5/perlprimtypes.swg#L90 and https://github.com/swig/swig/blob/master/Lib/perl5/perlprimtypes.swg#L157

wsfulton: > rewrite https://github.com/swig/swig/pull/2097 to actually get the same behavior on all plattforms.

Yes, so let's make consistent and resolve in the rewrite of #2097 and agreed it is not going to make it into swig-4.2.0 due out in a few days.

PR #2499, opened 2023-02-07
[Java] Added usage of nullptr instead of NULL

MyroslavaStopets:
Hi,
we are using SonarQube to analyze our C++ and Java source code. Unfortunately, SonarQube is emitting thousand of code smell warnings in the source code generated by Swig. For example:

"nullptr" should be used to denote the null pointer
According to [C++ Core Guidelines ES.47](https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#es47-use-nullptr-rather-than-0-or-null) it's better to use nullptr, which is unambiguous.

Can you please review this PR -- it will reduce some of the numerous amounts of SonarQube code smell warnings we receive in the Swig wrapper code.

Thank You

### DIFF
--- a/Lib/java/director.swg
+++ b/Lib/java/director.swg
@@ -60,11 +60,11 @@ namespace Swig {
   /* Java object wrapper */
   class JObjectWrapper {
   public:
-    JObjectWrapper() : jthis_(NULL), weak_global_(true) {
+    JObjectWrapper() : jthis_(SWIG_NULLPTR), weak_global_(true) {
     }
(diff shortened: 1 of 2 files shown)


wsfulton:
We use `0` in thousands of places instead of `NULL`. I expect there will be many more places to fix. With the addition of `SWIG_NULLPTR` in this pull request at least it facilitates further changes/pull requests to fix elsewhere in all language modules.

PR #2512, opened 2023-03-14
Update octave files to work with Octave v8.1.0

alatina:


opoplawski:
Seems to work great on Fedora.  Thanks.

wsfulton:
I can't see an Ubuntu ppa package for testing Octave 8.1 on Github Actions, which is ideally what we need to test this new version of Octave. I hope someone can provide that, but in the meantime, as the patch looks sensible and a second person has verified it working, I'll accept this patch.

ojwb:
I've added a new CI job using Octave 8.4 which passes, so we now have the missing coverage for this change.

I noticed there are concerning warnings with Octave 10.3 (tested locally) - it looks like Octave's C++ API has changed.  I've opened a new ticket for that: #3282.  @alatina Could take a look?

alatina:
Hi, 
Sure! I'll take a look in the coming few days

PR #2519, opened 2023-03-29
Fix CanCastAsInteger if errno is set.

degasus:
This method checks if the range of the input variable is fine.

However if the errno variable was already set, it fails even for valid inputs.

This fixes some random failures in the python castmode.


ojwb:
Thanks, merged.

Interestingly (at least according to the Linux man pages) this error handling is "just nonsense":

> SUSv2 and POSIX.1-2001 contain text about overflow (which might set errno to ERANGE, or raise an FE_OVERFLOW exception).   In  practice,  the result  cannot  overflow on any current machine, so this error-handling stuff is just nonsense.

PR #2524, opened 2023-03-31
detect Scilab 2023.0.0 and fix various issues

### LINKED ISSUE #894: Scilab: Use of ":" with typemap (int* IN, int IN_SIZE)
For an ":" input, one gets IN_SIZE=1,
IMHO it should be IN_SIZE=-1, so that within a
wrapped function the usage of ":" can be detected.

davidcl:
Scilab version numbering changed in the last release and it failed to be detected by `./configure`. This PR also fix #894 by implementing the requested change to handle `:` as regular integer vector extraction.


configure.ac
```diff
@@ -2423,10 +2418,10 @@ else
       # Set Scilab startup options depending on version
       AC_MSG_CHECKING(for Scilab startup options)
       SCILABOPT="-nwni -nb"
-      if test $SCILAB_VERSION_NO_DOTS -ge 540; then
+      if test $(echo -e '$SCILAB_VERSION\n5.3.3' |sort -n |head -n1) != 5.3.3; then
```
ojwb, line 2421: This trick doesn't actually work correctly - e.g. 5.3.10 compare as less than 5.3.3 (`-n` sorts the bit that looks like a floating point number numerically, but the second `.` and everything after appears to just get resolved with a string comparison):

```
$ printf '%s\n' 5.3.3 5.3.4 5.3.10 | sort -n
5.3.10
5.3.3
5.3.4
```

The existing "NO_DOTS" comparison is also broken though.

ojwb: That's not really suitable as there are platforms using other `sort` implementations.

Did you switch from the `SCILAB_VERSION_NO_DOTS` approach because it fails to handle multi-digit version components correctly?

If so, zero padding them to 3 each isn't hard:

```
SCILAB_VERSION_NO_DOTS=`printf '%03d%03d%03d' "$SCILAB_MAJOR_VERSION" "$SCILAB_MINOR_VERSION" "$SCILAB_MAINTENANCE_VERSION"`
```

E.g. 5.3.10 would then give `005003010`

ojwb: Oh I see the motivation was that the major version is now `2003`.  The above would work since `2003.0.0` gives (`2003000000`) but versions above 2147.483.648 potentially overflow a 32-bit int.

If that's a concern, using `%d%02d%02d` as the format would work if `SCILAB_MINOR_VERSION` and `$SCILAB_MAINTENANCE_VERSION` never go above 99.

Otherwise we could make the format `%04d%03d%03d` and use string comparisons like:

```
SCILAB_VERSION_NO_DOTS=`printf '%04d%03d%03d' "$SCILAB_MAJOR_VERSION" "$SCILAB_MINOR_VERSION" "$SCILAB_MAINTENANCE_VERSION"`
if test "$SCILAB_VERSION_NO_DOTS" '>' 0005003002; then 
```

Slightly awkward is you need to remember to quote `>` (otherwise it's interpreted as file redirection) and also `>=` doesn't work so you need to use `>` and compare against a version one lower.

Alternatively `expr` supports `>=` which does a string compare if the arguments can't be converted to integers:

```
if expr X"$SCILAB_VERSION_NO_DOTS" '>=' X0005003003 >/dev/null; then
```

davidcl: Thanks for your help, I used the `%04d%02d%02d` format to normalize the version and detect the same versions as before using the same `test` expressions.
```
SCILAB_VERSION_NO_DOTS=$(printf '%04d%03d%03d' "$SCILAB_MAJOR_VERSION" "$SCILAB_MINOR_VERSION" "$SCILAB_MAINTENANCE_VERSION")
```

Lib/scilab/scirun.swg
```diff
@@ -58,29 +57,29 @@ static void SWIG_Scilab_SetFuncName(char *funcName) {
...
 /* Argument management functions */
 
-#if SWIG_SCILAB_VERSION >= 540
-#define SWIG_CheckInputArgument(pvApiCtx, minInputArgument, maxInputArgument) CheckInputArgument(pvApiCtx, minInputArgument, maxInputArgument)
-#define SWIG_CheckInputArgumentAtLeast(pvApiCtx, minInputArgument) CheckInputArgumentAtLeast(pvApiCtx, minInputArgument)
-#define SWIG_CheckOutputArgument(pvApiCtx, minOutputArgument, maxOutputArgument) CheckOutputArgument(pvApiCtx, minOutputArgument, maxOutputArgument)
-#define SWIG_NbInputArgument(pvApiCtx) nbInputArgument(pvApiCtx)
-#define SWIG_AssignOutputArgument(pvApiCtx, outputArgumentPos, argumentPos) AssignOutputVariable(pvApiCtx, outputArgumentPos) = argumentPos
-#else
+#if SCI_VERSION_MAJOR < 6
```
ojwb, line 71: This has changed the version number being compared against as well as how the check was done - is that intended?

davidcl: good catch, I will change this `#if` to match the same version as before.

.github/workflows/ci.yml
```diff
@@ -189,6 +189,8 @@ jobs:
           VER: '5.5.2'
         - SWIGLANG: scilab
           os: ubuntu-18.04 # scilab 6.0
+        - SWIGLANG: scilab
+          os: ubuntu-22.04 # scilab 6.1
```
ojwb: 22.04 actually has scilab 6.1.1 (from the CI output and also from looking at the package repos, see output below) while the existing job below this addition is run on ubuntu-20.04 which has scilab 6.1.0.

It doesn't seem useful to test both 6.1.0 and 6.1.1, or at least not useful enough when we already have so many CI jobs that the time and resources to run them all are becoming concerning.  If the "0.0.1" difference is actually more significant here than I'm naively expecting, please speak up.

I've push a change to drop the 6.1.0 job, resurrect the 6.0 job (github stopped supporting their ubuntu-18.04 image recently so all the 18.04 jobs started to fail and I had to disable them) and add a 2023.0.0 build, since there seems to be a binary build on scilab.org for it and really we should have a job to confirm this new support works.

---

Scilab versions packaged for Ubuntu are:

```
$ grep 'LTS' /usr/share/distro-info/ubuntu.csv
6.06 LTS,Dapper Drake,dapper,2005-10-12,2006-06-01,2009-07-14,2011-06-01
8.04 LTS,Hardy Heron,hardy,2007-10-18,2008-04-24,2011-05-12,2013-05-09
10.04 LTS,Lucid Lynx,lucid,2009-10-29,2010-04-29,2013-05-09,2015-04-30
12.04 LTS,Precise Pangolin,precise,2011-10-13,2012-04-26,2017-04-28,2017-04-28,2019-04-28
14.04 LTS,Trusty Tahr,trusty,2013-10-17,2014-04-17,2019-04-25,2019-04-25,2024-04-25
16.04 LTS,Xenial Xerus,xenial,2015-10-22,2016-04-21,2021-04-21,2021-04-21,2026-04-23
18.04 LTS,Bionic Beaver,bionic,2017-10-19,2018-04-26,2023-05-31,2023-05-31,2028-04-26
20.04 LTS,Focal Fossa,focal,2019-10-17,2020-04-23,2025-05-29,2025-05-29,2030-04-23
22.04 LTS,Jammy Jellyfish,jammy,2021-10-14,2022-04-21,2027-06-01,2027-06-01,2032-04-21
$ rmadison -uubuntu scilab
 scilab | 5.5.0-2                | trusty/universe          | source, all
 scilab | 5.5.2-2ubuntu3         | xenial/universe          | source, all
 scilab | 6.0.1-1ubuntu1         | bionic/universe          | source, all
 scilab | 6.0.1-7ubuntu1~18.04   | bionic-security/universe | source, all
 scilab | 6.0.1-7ubuntu1~18.04.2 | bionic-updates/universe  | source, all
 scilab | 6.1.0+dfsg1-1ubuntu3   | focal/universe           | source, all
 scilab | 6.1.0+dfsg1-1ubuntu3.1 | focal-updates/universe   | source, all
 scilab | 6.1.1+dfsg2-3ubuntu1   | jammy/universe           | source, all
 scilab | 6.1.1+dfsg2-4          | kinetic/universe         | source, all
 scilab | 6.1.1+dfsg2-5          | lunar/universe           | source, all
```

ojwb: The 2023.0.0 binary dowload currently fails with HTTP error 403 (Forbidden).  I've reported this to scilab:

https://gitlab.com/scilab/scilab/-/issues/17024

ojwb: I've merged my change to resurrect the scilab 6.0 CI job and the fix for #894 (which really shouldn't be mixed up with these unrelated changes).

I'm a bit hesitant to merge the 2023.0.0 support without the CI job working since it seems likely the 403 issue will get fixed pretty quickly, so let's give them a day or to and see if it's working.  Don't worry about rebuilding the branch as I have it pretty much ready to go locally.

davidcl: actually the scilab binary file as a xz file not a gz file. As this can be good to distinguish this CI download, I think we can also ship a gz file for SWIG.

PR #2526, opened 2023-04-10
implement peeking beyond the end of the current macro expansion

### LINKED ISSUE #2525: SWIG preprocessor: problem when expanding macros with arguments to macros w/o arguments
Consider this:

`attr.i`
```
#define __attribute__(x)
%include "attr.h"
```

`attr.h`
```cpp
#define my_bad_attr __attribute__
#define my_good_attr(x) __attribute__(x)

int bad my_bad_attr((used));
int good my_good_attr((used));
```

Launch:
```
swig -E -javascript -node -c++ attr.i 
```

Result:

```
%endoffile
%endoffile
%includefile(maininput="attr.i") "attr.i" %beginfile

%includefile "attr.h" %beginfile



int bad __attribute__((used));
int good ;
%endoffile
%endoffile
```

mmomtchev:
Resolves #2525


ojwb:
This can easily go into an infinite loop, e.g.:

```
#define X(Y,Z) X(Y##Z,Z##Y)
X(Y,Z)
```

SWIG 4.1.0 handles this the same way as GCC and clang:

```
$ gcc -E infloop-2526.h 
# 0 "infloop-2526.h"
# 0 "<built-in>"
# 0 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 0 "<command-line>" 2
# 1 "infloop-2526.h"

X(YZ,ZY)
$ clang-16 -E infloop-2526.h
# 1 "infloop-2526.h"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 374 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "infloop-2526.h" 2

X(YZ,ZY)
$ swig -E -python infloop-2526.h|grep ^X
X(YZ,ZY)
```

But testing with a build of your patched version doesn't finish (presumably it'll eventually run out of memory):

```
$ ./preinst-swig -E -python infloop-2526.h|grep ^X
```

I think to solve this we really need to make SWIG's preprocessor handle expanding macros more like a compiler's preprocessor does.

mmomtchev:
@ojwb `Preprocessor_replace` must work statement by statement and repeat the macro expansion until there are macros, painting the expanded macros at each step and not expanding macros expanded in previous steps - but still allowing multiple expansions of the same macro during one step.
I don't see any other option, I will completely remove the recursive expansion in this case.

mmomtchev:
@ojwb after trying very hard to make the multiple passes work, I settled on an alternative approach - keep the current anti-recursion exit - which is very hard to do properly without recursion - and instead treat the macro arguments arising from the expansion as a special case. I added the possibility to peek beyond the end of the current recursion expansion.

Examples/test-suite/preproc_predefined.h
```diff
@@ -25,3 +25,14 @@
...
+/*
+ * This should not compile in when this bug is present
+ * https://github.com/swig/swig/issues/2525
+ */
+
+#define my_bad_attr __attribute__
+#define my_good_attr(x) __attribute__(x)
+
+int bad my_bad_attr((used));
+int good my_good_attr((used));
```
ojwb, line 38: You've added this to a testcase which is specifically testing SWIG's predefined macros, which this is not related to at all.  Where you had it before in `preproc_defined.i` seemed more suitable.

Source/Preprocessor/cpp.c
```diff
@@ -1280,11 +1281,23 @@ static DOH *Preprocessor_replace(DOH *s) {
       Replaceall(fn, "\\", "\\\\");
       Printf(ns, "\"%s\"", fn);
       Delete(fn);
-    } else if (Getattr(symbols, id)) {
-      DOH *e;
+    } else if (m = Getattr(symbols, id)) {
       /* Yes.  There is a macro here */
+      /* If it expects arguments, they must from from `next` */
```
ojwb, line 1286: typo: from from

Source/Preprocessor/cpp.c
```diff
@@ -1009,7 +1010,7 @@ static String *expand_macro(String *name, List *args, String *line_file) {
   /* Expand this macro even further */
   Setattr(macro, kpp_expanded, "1");
 
-  e = Preprocessor_replace(ns);
+  e = Preprocessor_replace(ns, line_file);
```
ojwb, line 1013: I'm confused by this part, and wonder if you meant the new parameter passed here to be `next`?

The new second parameter to `Preprocessor_replace()` is called `next` and otherwise the new `next` parameter to `expand_macro()` doesn't seem to get used anywhere...

ojwb:
CI is still failing on your testsuite changes:

```
checking python testcase preproc_expr
  File "/home/runner/work/swig/swig/build/build/Examples/test-suite/python/../../../../../Examples/test-suite/python/preproc_defined_runme.py", line 6, in <module>
    if preproc_defined.y != 8:
       ^^^^^^^^^^^^^^^^^
AttributeError: module 'preproc_defined' has no attribute 'y'
```

Please make sure the tests pass locally before pushing.  It'll save you a lot of time, and it'll save burning a lot of machine time running builds (each push runs **well over 100 builds** of SWIG for different target language versions, which in various configurations).

Examples/test-suite/preproc_defined.i
```diff
@@ -123,3 +123,6 @@ void another_macro_checking(void) {
 #if 0
 # wobble wobble
 #endif
+
+#define __attribute__(x)
+%import <preproc_defined.h>
```
ojwb, line 128: You shouldn't need to add a separate header - just put its contents here wrapped in `%inline` like so:

```
%inline %{
// put header contents here
%}
```

I suspect that'll fix your testsuite problem too - `%import` parses the file and gathers type information but doesn't actually wrap anything in the file.

Source/Preprocessor/cpp.c
```diff
@@ -1280,11 +1280,23 @@ static DOH *Preprocessor_replace(DOH *s) {
...
+      DOH *e;
+      int macro_additional_lines = 0;
       /* See if the macro expects arguments */
-      e = expand_macro(id, 0, s);
+      if (Getattr(m, kpp_args) && line_file) {
+        /* Yep.  We need to go find the arguments and do a substitution */
+        int line = Getline(line_file);
+        args = find_args(line_file, 1, id);
+        macro_additional_lines = Getline(line_file) - line;
+        assert(macro_additional_lines >= 0);
```
ojwb, line 1295: I don't understand what this assertion is trying to achieve.

If you're wanting to assert that we found parameters, `assert(args);` would achieve that more directly (and probably more successfully - it looks to me like if we don't find parameters then `macro_additional_lines` will be zero anyway).

mmomtchev: I think it is somekind of leftover protection for a bug in the macro parsing - maybe something that was not clear if it was fixed or not. I got it from line 1212 and since the original author kept it, I decided to keep it too. It is not about `args` being non-zero which is, first, is a valid situation that is handled and, second, it cannot arise from calling `find_args`. Maybe remove both of them?

Examples/test-suite/preproc_defined.i
```diff
@@ -123,3 +123,30 @@ void another_macro_checking(void) {
...
+/*
+ * The behaviour of Self-Referential Macros is defined
+ * https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Self-Referential-Macros.html
+ */
+int y = 0;
+
+#define x (4 + y)
+#define y (2 * x)
+
+int z = y;
```
ojwb, line 148: This testcase fails because this isn't valid C code:

```
#define x (4 + y)
#define y (2 * x)

int z = y;
$ gcc -Wall -W -c t.c
t.c:4:11: error: initializer element is not constant
    4 | #define y (2 * x)
      |           ^
t.c:6:9: note: in expansion of macro ‘y’
    6 | int z = y;
      |         ^
```

The line expands to `int z = (2 * (4 + y));` and `y` isn't constant.

I'll see if I can come up with something that works without having to switch this to be a C++ testcase, but seriously you really need to be making sure the tests pass **locally** before pushing changes or else you're wasting everybody's time (including your own).

PR #2527, opened 2023-04-12
Fix "undefining the allocator of T_DATA" error seen in Ruby 3.2

wildmaples:
See comment: https://github.com/swig/swig/issues/2257#issuecomment-1499135189

Ruby 3.2 issue: https://bugs.ruby-lang.org/issues/18007 
Similar to https://github.com/robinst/swig/commit/9b5d37fd174331fa2b7113fe968fcf0570de43bf

### DIFF
--- a/Lib/ruby/rubyrun.swg
+++ b/Lib/ruby/rubyrun.swg
@@ -157,7 +157,8 @@ SWIG_Ruby_define_class(swig_type_info *type)
     _cSWIG_Pointer = rb_define_class_under(_mSWIG, "Pointer", rb_cObject);
     rb_undef_method(CLASS_OF(_cSWIG_Pointer), "new");
   }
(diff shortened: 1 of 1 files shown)


ojwb:
Please don't make whitespace changes in the same commit as functional changes.  It makes `git blame` much less useful.

You need to declare the variable at the top of the function (or perhaps better just eliminate it and just do `rb_undef_alloc_func(rb_define_class_under(_mSWIG, klass_name, _cSWIG_Pointer));`

We really should have a regression test for this - does it trigger when running SWIG's existing testsuite?

ojwb:
> We really should have a regression test for this - does it trigger when running SWIG's existing testsuite?

I'm working on adding a Ruby 3.2 CI job, and can now answer this myself - yes it does:

```
../../../../../Examples/test-suite/ruby/voidtest_runme.rb:20: warning: undefining the allocator of T_DATA class SWIG::TYPE_p_void
```

PR #2533, opened 2023-04-17
[Go] Add argc and argv multi-argument to go.

erezgeva:
Hi,

Add Argc Argv to go language.
Make argc and argv test works in C.

Erez

### DIFF
--- a/Examples/test-suite/argcargvtest.i
+++ b/Examples/test-suite/argcargvtest.i
@@ -1,6 +1,6 @@
 %module argcargvtest
 
-#if !defined(SWIGCSHARP) && !defined(SWIGD) && !defined(SWIGGO) && !defined(SWIGGUILE) && !defined(SWIGJAVA) && !defined(SWIGJAVASCRIPT) && !defined(SWIGMZSCHEME) && !defined(SWIGOCAML) && !defined(SWIGR) && !defined(SWIGSCILAB)
+#if !defined(SWIGCSHARP) && !defined(SWIGD) && !defined(SWIGGUILE) && !defined(SWIGJAVA) && !defined(SWIGJAVASCRIPT) && !defined(SWIGMZSCHEME) && !defined(SWIGOCAML) && !defined(SWIGR) && !defined(SWIGSCILAB)
 %include <argcargv.i>
 
 %apply (int ARGC, char **ARGV) { (size_t argc, const char **argv) }
@@ -18,7 +18,11 @@ const char* mainv(size_t argc, const char **argv, int idx)
   return argv[idx];
 }   
 
+#ifdef __cplusplus
 void initializeApp(size_t argc, const char **argv, bool setPGid = true, bool isMakeline = false)
+#else
+void initializeApp(size_t argc, const char **argv)
+#endif
 {
 }
 
--- a/Examples/test-suite/go/argcargvtest_runme.go
+++ b/Examples/test-suite/go/argcargvtest_runme.go
@@ -1,0 +1,19 @@
+package main
+
+import wrap "swigtests/argcargvtest"
+
+func main() {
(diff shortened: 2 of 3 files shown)


Examples/test-suite/argcargvtest.i
```diff
@@ -1,24 +1,28 @@
...
 {
   return argv[idx];
-}   
+}
 
+#ifdef __cplusplus
 void initializeApp(size_t argc, const char **argv, bool setPGid = true, bool isMakeline = false)
+#else
+void initializeApp(size_t argc, const char **argv)
+#endif
```
ojwb, line 25: `argcargvtest` is a C++ testcase (it's in `CPP_TEST_CASES` in `Examples/test-suite/common.mk`) so I don't understand this part of the change.

Examples/test-suite/argcargvtest.i
```diff
@@ -1,24 +1,28 @@
...
+#if !defined(SWIGCSHARP) && !defined(SWIGD) && !defined(SWIGGUILE) && !defined(SWIGJAVA) && !defined(SWIGJAVASCRIPT) && !defined(SWIGMZSCHEME) && !defined(SWIGOCAML) && !defined(SWIGR) && !defined(SWIGSCILAB)
 %include <argcargv.i>
 
 %apply (int ARGC, char **ARGV) { (size_t argc, const char **argv) }
 #endif
 
 %inline %{
 
-int mainc(size_t argc, const char **argv) 
+int mainc(size_t argc, const char **argv)
```
ojwb, line 11: Don't mix whitespace cleanup and functional changes - it makes `git blame` much less useful.

erezgeva: I revert

erezgeva:
All checks should pass. Seems a problem in the CI, `http://azure.archive.ubuntu.com` fails.

Examples/test-suite/argcargvtest.i
erezgeva: Currently we have 8 languages that support `argcargv`.
There are: "`go lua octave perl5 php python ruby tcl`".
All pass `make argcargvtest.ctest`.
We can add `argcargvtest` to `C_TEST_CASES`.

ojwb: I didn't add this testcase, but it includes default parameters which are a C++ feature, so it needs C++ to work.  Sure you can hack out that part, but you then aren't testing whatever those parameters are there for.

> It will allow us to confirm when we change the argcargv.i for languages supporting C.

We support wrapping C and C++ for all target languages though.

erezgeva: Yes, you are right `initializeApp` aims C++.
Writing 2 `argcargvtest_runme.X` scripts or condition them, seems like overhead.
It is much more easy to make `initializeApp` redundant in C.

Testing C does have value.
For example:
Define variables not on start of block
`{`
`   for(int i = 0; ...`
Or using C++ only features like references
`{`
'    int &len = arg1`
Would fail C compilation.

So when we add or change `argcargv.i` we can test with `make argcargvtest.ctest`.
That is what the test suit case is about, a way to check our code as we write it.
CI is just a a second floor.

I do not see the benefit of adding the test to `C_TEST_CASES`,
C++ cover the important parts, and have more code to test.
I do not think we need to inflate the CI. It is already full.

erezgeva: Just for the record, I did write an initial  `go/argcargv.i` for go, and than wrote `argcargvtest_runme.go`.
But it did not work properly, and I used the testing script to debug and fix the `argcargv.i` file.
`make argcargvtest.cpptest` is truly part of development, and I think ensuring it support C is part of the development too, I do not think I need to "hack" the test just to ensure C works.

Issue #2534, opened 2023-04-18
Variable is emitted twice during a unit test (silent error)


mmomtchev:
Launch:
```
swig/Examples/test-suite/javascript$ ENGINE=v8 make allprotected.cpptest
```

Then check `allprotected_wrap.cxx`:
```cpp
SWIGV8_AddStaticVariable(_exports_PublicBase_obj, "staticConstMemberVariable", _wrap_PublicBase_staticConstMemberVariable, JS_veto_set_variable, context);
SWIGV8_AddStaticVariable(_exports_PublicBase_obj, "staticConstMemberVariable", _wrap_PublicBase_staticConstMemberVariable, JS_veto_set_variable, context);
```

The `PublicBase::staticConstMemberVariable` is emitted twice. Due to the way V8 handles the registration, the second simply overwrites the first one - but this is not valid for all languages.

mmomtchev:
The two instances are emitted from here:
https://github.com/swig/swig/blob/f030df8786bc3e707bee9079d093368de3d8bc57/Source/Modules/javascript.cxx#L408

The first one happens in `Language::variableHandler(n)` - where the constants are processed - and the second one - on the next line - `emitter->exitVariable(n)` where the variables are processed

mmomtchev:
I think that `JSEmitter::emitConstant` should not enter the variable at all, the variable has already been entered and will be exited - it should simply modify whatever there is to modify, with the rest being emitted in `JSEmitter::exitVariable`

mmomtchev:
In fact `JSEmitter::emitConstant` is used for enums and some other cases - and it should stay like this.

But the `variableHandler` should not lead to it. These two cannot be part of one another - they must be two separate mutually exclusive handlers.

mmomtchev:
Ok, I have two proposals:
* either a major refactor of all the JavaScript backends - moving the emitting of variables to the `variableWrapper` handler - instead of the current `exitVariable` - which is probably the right thing to do
  - this way `Language` will call either `constantWrapper` or `variableWrapper`
* or, simply setting an attribute flag when emitting a symbol as a constant so that it does not get reemitted as a variable - which is the practical solution

PR #2535, opened 2023-04-18
do not emit static const variables twice

mmomtchev:
It is an ugly solution, but it is low-risk and it works

#2534

### DIFF
--- a/Examples/test-suite/javascript/smart_pointer_const_runme.js
+++ b/Examples/test-suite/javascript/smart_pointer_const_runme.js
@@ -1,0 +1,11 @@
+var smart_pointer_const = require('smart_pointer_const');
+
+var f = new smart_pointer_const.Foo;
+var b = new smart_pointer_const.Bar(f);
+
+b.x = 3;
+if (b.getx() != 3) throw new Error('Failed');
+
+var fp = b.__deref__();
+fp.x = 4;
+if (b.getx() != 4) throw new Error('Failed');
--- a/Source/Modules/javascript.cxx
+++ b/Source/Modules/javascript.cxx
@@ -1707,6 +1707,14 @@ int JSCEmitter::enterVariable(Node *n) {
 }
 
 int JSCEmitter::exitVariable(Node *n) {
+  // Due to special handling of C++ "static const" member variables
+  // (refer to the comment in lang.cxx:Language::staticmembervariableHandler)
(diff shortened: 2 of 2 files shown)


mmomtchev:
@ojwb @wsfulton what do you think of this PR, the problem it resolves is purely cosmetic since the V8 API accepts that a symbols is registered twice.

It is a real problem only with the NAPI backend where it has been solved separately (in the NAPI PR)?

ojwb:
Yeah, the whole extra level of abstraction in `javascript.cxx` doesn't seem helpful.

As for you workaround, I was going to leave it to William to decide whether he was happy with it.

If we're going for it, I would suggest naming the attribute used with a `js_` or `javascript_` prefix for clarity and reduced risk of it colliding with future generic attributes.

mmomtchev:
@ojwb this is ready

ojwb:
> this is ready

As I already noted, I'd like wsfulton's thoughts on this.  He may see a cleaner way to solve it short of a major refactor of `javascript.cxx`.

wsfulton:
We use `:` rather than `_` as a namespace separator in the attributes with the prefix the name of the file usually. So use "javascript:constant". BTW, the "hasconst" might be the existing attribute you can use rather than creating a new one. That's if I understand correctly that you want to detect the "static const" or "constexpr" variables.

mmomtchev:
The problem is that the way the JS backend is structured. There is `enterVariable`, `variableHandler` and `exitVariable`. There is also a `constantHandler` which is called only for constants. The declaration is emitted in `exitVariable`. Constants can be created in two different ways - by a chain of

  `enterVariable` -> `constantHandler` -> `exitVariable`

when the syntax analysis detects a constant or also by a chain of

`enterVariable` -> `variableHandler` -> `enterVariable` -> `constantHandler` -> `exitVariable` -> `exitVariable`

when `variableHandler` switches a variable to a constant - which happens for class members that are `static const` and `static constexpr`. So I need a special flag to detect this second call of `exitVariable`. Otherwise I end up with two identical declarations. For JSC and V8 this is not a problem - if you look at the code you will see two consecutive statements creating the value, the second one overwriting the first one. But for the NAPI generator, it is a problem, because it creates a naming collision. The NAPI generator has its own solution. The question is whether this is needed for V8/JSC.

Issue #1567, opened 2019-06-25
Add support for std::string_view


barreyra:
New c++17 feature.

fflexo:
Superficially this is pretty easy to get the data exposed via SWIG in both directions.

The tricky bit is keeping the semantics right - if you used a `string_view` it's probably because you didn't want to make a copy. For input argument this is easy enough. For return values it's basically not possible to get something that feels like the target language's String class and has 0 copies in most languages.

So I think in the first instance I'd probably aim to support just two cases of function arguments - value and const reference because those are easy to make exactly right.

barreyra:
I agree the return value is tricky. The safest solution here is to make a copy of the value, unless we can be smart about the object's lifetime.

fflexo:
I've been experimenting with this a little, but not got much useful to push yet as I've not actually decided which approach to run with. We can solve this in essentially 3 ways:

1. Hack something that builds on the existing `std::string` typemaps. This turns out to fairly rapidly get quite messy though - I ended up making a private sub-class of `std::string_view` that let the existing `std::string` typemaps work and then deliberately using object slicing to get a real, plain `std::string_view`. It's "ok" because we wanted that to happen, but it's rather more hacky than I'd hoped
2. Write something that's substantially similar to the `std::string` typemaps, including per-language support. There are quite a lot of languages though and I only know 5 or 6 well enough to be confident of doing a decent job.
3. Refactor `std::string` code to make each language give a nicer, pure C++ abstraction around reading/creating strings. This feels pretty intrusive though.

For the output case making a copy, but adding a warning is probably the most useful, least bad option.

vadz:
My 2c: I'd try to refactor the common part of `string` and `string_view` typemaps in a macro that would be used to define them for both classes. Of course, there would also be an additional part for all the stuff in `string` not present in `string_view`, but it looks like the separation could be done purely mechanically and so have a decent chance for not breaking something.

As for the output direction, I'd just make a copy. A warning would be annoying and it's not like you could do anything about it anyhow (other than defining a function just for SWIG returning a `string` instead of `string_view` but what would we gain from forcing people to do this?).

DeeeeLAN:
Any updates or workarounds for this? I have a library I am trying to use swig with and it uses string_view quite a bit.

fflexo:
[...]

PR #2540, opened 2023-04-20
Initial support for std::string_view

ojwb:
Implemented for PHP and Lua so far.

See #1567

### DIFF
--- a/Lib/java/std_string_view.i
+++ b/Lib/java/std_string_view.i
@@ -34,11 +34,14 @@ class string_view;
    }
    const char *$1_pstr = (const char *)jenv->GetStringUTFChars($input, 0);
    if (!$1_pstr) return $null;
-   $1 = std::string_view($1_pstr);
-   /* FIXME: std::string_view requires the string data to remain valid while the string_view is in use. */
-   /* jenv->ReleaseStringUTFChars($input, $1_pstr); */ %}
+   $1 = std::string_view($1_pstr); %}
 
-%typemap(directorout) string_view
+/* std::string_view requires the string data to remain valid while the
+ * string_view is in use. */
+%typemap(freearg) string_view
+%{ jenv->ReleaseStringUTFChars($input, $1_pstr); %}
+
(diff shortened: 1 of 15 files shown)


Lib/java/std_string_view.i
```diff
@@ -0,0 +1,138 @@
...
+   $1_strview = $1_str;
+   $result = &$1_str;
+   jenv->ReleaseStringUTFChars($input, $1_pstr); %}
+
+%typemap(directorin,descriptor="Ljava/lang/String;") const string_view &
+%{ $input = jenv->NewStringUTF(std::string($1).c_str());
+   Swig::LocalRefGuard $1_refguard(jenv, $input); %}
+
+%typemap(out) const string_view &
+%{ $result = jenv->NewStringUTF(std::string(*$1).c_str()); %}
```
wsfulton, line 124: More efficient as it avoids a string copy:
```suggestion
%{ $result = jenv->NewStringUTF($1->data()); %}
```
There are others in the pull request.

wsfulton (changes requested):
Looks like you've copied and tweaked the std_string.i files slightly, which is the most sensible thing to do. Just the efficiency comment to address, otherwise looks useful and worth merging.

Lib/java/std_string_view.i
wsfulton: I've since seen that string_view::data() is not guaranteed to be null terminated, so the copy make sense in this case.

ojwb: Indeed - it's unfortunate it's needed, but there doesn't seem to be a `NewStringUTFWithLength()` variant provided by Java.

ojwb:
I was going to wait until I'd got all the languages I need working, but it probably makes more sense to merge what I have so far.

I should probably write something in the manual about it though.  I'll do that, give it a final check it over and merge soon.

wsfulton:
Yes, at least add it to the list of library files in Library.html. A new section along the lines of **12.4.1 std::string** would be nice to have.

wsfulton (approved):

Issue #2542, opened 2023-04-20
Renaming is mostly broken in JavaScript


mmomtchev:
`constructor_rename` and `overload_rename` produce invalid code that compiles but the tests pass because there is no `_runme.js`.

`rename_cameltest` works because of a coincidence:
```
$ ENGINE=node make rename_camel.cpptest
```
```
$ grep _wrap_hello rename_camel_wrap.cxx
```
```cpp
static SwigV8ReturnValue _wrap_hello(v8::Local<v8::Name> property, const SwigV8PropertyCallbackInfo &info) {
static SwigV8ReturnValue _wrap_hello(const SwigV8Arguments &args) {
  if(args.Length() != 0) SWIG_exception_fail(SWIG_ERROR, "Illegal number of arguments for _wrap_hello.");
SWIGV8_AddStaticVariable(exports_obj, "HELLO", _wrap_hello, JS_veto_set_variable, context);
SWIGV8_AddStaticFunction(exports_obj, "hello", _wrap_hello, context);
```

There is a name collision on `wrap_hello` which works through overloading because a getter and a method have slightly different call signatures.

mmomtchev:
Maybe if tests like this one are supposed to work, `Swig_name_wrapper` should probably take renaming into account? Or, renaming should also create a new full `name` instead of changing only `sym:name` which might not be unique, this is something that should be valid for all languages?

ojwb:
Re: the rename_camel part, I think the problem is in this code calling `Swig_name_wrapper()`:

```
int JSEmitter::emitConstant(Node *n) {  
  // HACK: somehow it happened under Mac OS X that before everything started  
  // a lot of SWIG internal constants were emitted  
  // This didn't happen on other platforms yet...  
  // we ignore those premature definitions  
  if (!State::IsSet(state.globals(HAS_TEMPLATES))) {  
    return SWIG_ERROR;  
  }  
  
  Wrapper *wrapper = NewWrapper();  
  SwigType *type = Getattr(n, "type");  
  String *name = Getattr(n, "name");  
  String *iname = Getattr(n, "sym:name");  
  String *wname = Swig_name_wrapper(name);  
  String *rawval = Getattr(n, "rawval");  
  String *value = rawval ? rawval : Getattr(n, "value");  
  
  // HACK: forcing usage of cppvalue for v8 (which turned out to fix typedef_struct.i, et. al)  
[...]

PR #2546, opened 2023-04-21
use the renamed name when creating a constant

mmomtchev:
Related to #2542


ojwb:
Ideally we'd have a regression test, but AIUI this doesn't actually cause breakage because the node API is C++ so the wrapper is always C++ and we just end up overloading the wrapper function name:

```
static SwigV8ReturnValue _wrap_hello(v8::Local<v8::Name> property, const SwigV8PropertyCallbackInfo &info) {
```
vs
```
static SwigV8ReturnValue _wrap_hello(const SwigV8Arguments &args) {
```

I don't see a trick to build a testcase from.

mmomtchev:
The `napi` branch has a test that failed without this

PR #2548, opened 2023-04-23
Allow using snprintf() instead of sprintf() in wrappers

### LINKED ISSUE #2502: sprintf deprecation on OSX
Hey,

I'm seeing generated code containing `sprintf` with deprecation errors on OSX.
I don't know much about this function but it says to us snprintf.

https://developer.apple.com/forums/thread/714675

```
error: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  
Due to security concerns inherent in the design of sprintf(3), it is highly recommended that 
you use snprintf(3) instead. [-Werror,-Wdeprecated-declarations]
```

I can compile with `-Wno-deprecated-declarations` but I'd rather not.

Suggestions welcome.

ojwb:
We aim to produce code that works with C90 or C++98 so we can't assume snprintf() is available, but it almost always is (even on systems from before it was standardised) so having a way to use it is helpful.

Enable this automatically if the compiler claims conformance with at least C90 or C++98 and check SWIG_HAVE_SNPRINTF to allow turning on manually, but disable is SWIG_NO_SNPRINTF if defined.

The fallback is to call sprintf() without a buffer size check - checking after the call is really shutting the stable door after the horse has bolted, and most of our uses either have a fixed maximum possible size or dynamically allocate a buffer that's large enough.

Fixes #2502 (sprintf deprecation warnings on macos)

### DIFF
--- a/Lib/csharp/csharp.swg
+++ b/Lib/csharp/csharp.swg
@@ -579,7 +579,7 @@ SWIGINTERN const char * SWIG_UnpackData(const char *c, void *ptr, size_t sz) {
                  unsigned long, 
                  unsigned short
 %{ char error_msg[256];
-   sprintf(error_msg, "C++ $1_type exception thrown, value: %d", $1);
+   SWIG_snprintf(error_msg, sizeof(error_msg), "C++ $1_type exception thrown, value: %d", $1);
    SWIG_CSharpSetPendingException(SWIG_CSharpApplicationException, error_msg);
    return $null; %}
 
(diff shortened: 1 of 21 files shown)


ojwb:
Oops, was tidying up my merged branches and accidentally removed this one too.

wsfulton (approved):
Looks okay to me, although I can't work out if the CI all passed. Perhaps you can add an entry to CHANGES.current to trigger another run.

ojwb:
CI did all pass.

I then accidentally deleted the branch while tidying up my merged branches, and reinstating triggered another CI run which I manually cancelled to not waste resources.

I'll just merge it rather than burning a load of extra hours of CPU time.

PR #2551, opened 2023-04-23
tentative fix for unit testing

### LINKED ISSUE #2550: Different handling between `typedef struct` and `typedef` / `struct` when using a namespace
```swig
%module typedefs

namespace nspace1 {
  typedef struct _AffineMatrix {
      int a;
  } AffineMatrix;

  struct _CacheView {
      int a;
  };
  typedef struct _CacheView CacheView;
}

using nspace1::AffineMatrix;
using nspace1::CacheView;

void fn1(AffineMatrix a);
void fn2(CacheView a);
```

Launch (language doesn't matter, it is a parser problem):
```
./swig -node -javascript -c++ typedef.i 

typedef.i:14: Warning 315: Nothing known about 'AffineMatrix'.
typedef.i:14: Warning 315: Nothing known about 'AffineMatrix'.
```

mmomtchev:
Fixes #2550

### DIFF
--- a/Examples/test-suite/javascript/struct_typedef_namespace_runme.js
+++ b/Examples/test-suite/javascript/struct_typedef_namespace_runme.js
@@ -1,0 +1,7 @@
+var struct_typedef_namespace = require("struct_typedef_namespace");
+
+var b = new struct_typedef_namespace.AffineMatrix();
+b.x = b.y = b.z = 1;
+
+if (struct_typedef_namespace.fn1(b) != b.x)
+  throw new Error('failed');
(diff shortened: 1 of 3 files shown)


mmomtchev:
The problem with that one - besides the fact that it is a commit in 20 years old field-tested code - is that I do not understand why `sym:symtab` is not systematically set on the current node - which seems the logical thing to do. Was the original author simply too smart/lazy or is there a good reason to not always do it.

wsfulton:
A leading underscore followed by a capital letter is an illegal identifier in C++, can you correct this in the test case please. Please add struct_typedef_namespace to Examples/test-suite/common.mk so that the test is then run in CI.

The typedefs (using -debug-typedef) look okay. Taking a deeper look next.

PR #2552, opened 2023-04-24
Add more argc and argv multi-argument typemaps.

erezgeva:
Add argc and argv multi-argument typemap to
 Java, Guile, <del>MzScheme,</del> JavaScript, Scilab, <del>D language,</del> and C#. 

### Regarding the other languages:

<del>
D language can convert string array directly to C char pointer pointer.
Which means the developers do no need an argc argv typemap for D.
I did saw issue of converting C char pointer back to D string.
Seems it miss the null termination.
We may need a fix on convert strings from C back to D.</del><br/><br/>

<del>
C sharp can convert string array directly to C char pointer pointer.
No point to use an argc argv typemap.
However it seems we need some changes to convert string array to C char pointer pointer.
In the current state, the swig tries to use SWIGTYPE_p_p_char, which seems in the wrong direction.</del><br/><br/>

I wrote R language argc and argv typemap.
However initializeApp() fails as the rtypecheck need migration or fix in Source/Modules/r.cxx

<del>
As for OCaml, I simply do not understand how to pass a string array from OCaml through the wrapper.
I might miss it, or OCaml use only one dimension of arrays and lists?</del>

### DIFF
--- a/Examples/test-suite/argcargvtest.i
+++ b/Examples/test-suite/argcargvtest.i
@@ -1,9 +1,14 @@
 %module argcargvtest
 
-#if !defined(SWIGCSHARP) && !defined(SWIGD) && !defined(SWIGGUILE) && !defined(SWIGJAVA) && !defined(SWIGJAVASCRIPT) && !defined(SWIGMZSCHEME) && !defined(SWIGOCAML) && !defined(SWIGR) && !defined(SWIGSCILAB)
+#if !defined(SWIGD) && !defined(SWIGMZSCHEME) && !defined(SWIGOCAML) && !defined(SWIGR)
+#define SWIGCSHARP_ARGCARGV_PARAM_IDX "0"
 %include <argcargv.i>
 
+#ifndef SWIGCSHARP
 %apply (int ARGC, char **ARGV) { (size_t argc, const char **argv) }
+#else
+%apply (char **ARGV) { const char **argv }
+#endif
 #endif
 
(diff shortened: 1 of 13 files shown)


erezgeva:
The latest JavaScript commits on master by  Olly Betts <olly@survex.com>.
Seems breaking some of the Javascripts CI.
I re-based my commits before them.

erezgeva:
[Add li_constraints runme test](https://github.com/swig/swig/pull/2567/commits/d77f2877740a2e3b6e284a4efc9392289f47b748)
For: D, C#, go, Java, Lua, Perl, PHP, python, Ruby, TCL, Octave.

As for: Javascripts, R, scilab do not support the check typemap.
@ojwb and @wsfulton Do we want to add check typemap to these 3 languages?

As for guile, it does support the check typemap,
 but I simply do not know understand guile Lisp.  I did manage TCL (which inherit some of Lisp syntax), but guile is a different level of Lisp.

erezgeva:
Remove all languages that are candidate to future remove.
Leave only: Java, Guile, JavaScript, Scilab and C#.

erezgeva:
The error in R is not related to this PR.
As it does not contain any change related to R.

erezgeva:
@ojwb, @wsfulton
This patch was refined.
And now it includes: Java, Guile, JavaScript, Scilab and C#.
Officially supported languages only!
I is ready for review :-)

PR #2556, opened 2023-04-25
[java] Suppress removal warnings for finalize()

ojwb:
#1240 suppressed "deprecation" warnings for `finalize()` methods, but these changes to "removal" warnings with a newer JDK version

@wsfulton OK?

Without this I'm finding running the testsuite to be unusably noisy with the JDK version I have:

```
$ java -version
openjdk version "18.0.2-ea" 2022-07-19
OpenJDK Runtime Environment (build 18.0.2-ea+9-Debian-2)
OpenJDK 64-Bit Server VM (build 18.0.2-ea+9-Debian-2, mixed mode, sharing)
```

I could just keep this locally, but that would be a bit of a pain to do.

### DIFF
--- a/Examples/test-suite/java_director.i
+++ b/Examples/test-suite/java_director.i
@@ -7,7 +7,7 @@
 %module(directors="1") java_director
 
 %typemap(javafinalize) SWIGTYPE %{
-  @SuppressWarnings("deprecation")
+  @SuppressWarnings({"deprecation", "removal"})
   protected void finalize() {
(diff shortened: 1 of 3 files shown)


ojwb:
CI failures are all for non-java languages and look to be due to github connectivity problems.

ojwb:
No response and this is continuing to annoy me so I'm going to merge it.  The alternative is I give up running the Java testsuite locally.  It's not great to be suppressing these but the same logic that led to #1240 being merged seems to apply here too.

PR #2562, opened 2023-04-27
support check typemaps in JS

### LINKED ISSUE #2561: Checking typemaps `%typemap(check)` seem to be absent from JS
There is no `multimap` example in `Examples/javascript` and this does not work - the check is not included at all:

```swig
%module check_typemap

%include exception.i
%include typemaps.i

%inline %{

void circle(double x, double y) {
  double a = x*x + y*y;
  if (a > 1.0) {
    printf("Bad points %g, %g\n", x,y);
  } else {
    printf("Good points %g, %g\n", x,y);
  }
}

%}

%typemap(check) (double cx, double cy) {
   double a = $1*$1 + $2*$2;
   if (a > 1.0) {
	SWIG_exception(SWIG_ValueError,"$1_name and $2_name must be in unit circle");
   }
}
```

Only Ruby has a unit test with a `check` typemap

mmomtchev:
Fixes #2561

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -537,6 +537,7 @@ CPP_TEST_CASES += \
 	typedef_typedef \
 	typemap_arrays \
 	typemap_array_qualifiers \
+	typemap_check \
 	typemap_delete \
 	typemap_directorout \
 	typemap_documentation \
@@ -812,7 +813,6 @@ C_TEST_CASES += \
 	struct_initialization \
 	typedef_classforward_same_name \
 	typedef_struct \
-	typemap_check \
 	typemap_subst \
 	union_parameter \
 	unions \
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -812,6 +812,7 @@ C_TEST_CASES += \
 	struct_initialization \
 	typedef_classforward_same_name \
 	typedef_struct \
+	typemap_check \
 	typemap_subst \
 	union_parameter \
 	unions \
--- a/Examples/test-suite/javascript/typemap_check_runme.js
+++ b/Examples/test-suite/javascript/typemap_check_runme.js
@@ -1,0 +1,9 @@
+const check_typemap = require('check_typemap');
+
+try {
+  check_typemap.circle(100, 100);
+  throw new Error('Didn\'t throw');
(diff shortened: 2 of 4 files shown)


mmomtchev:
@ojwb are all the weird errors due to the fact that I made a `ctest` instead of a `cpptest`? Excluding JSC of course.

ojwb:
> are all the weird errors due to the fact that I made a `ctest` instead of a `cpptest`? Excluding JSC of course.

I don't know what's causing them from looking at the logs, but I don't see why this being a `ctest` would be a problem.

ojwb:
At least one of your problems here is that `SWIG_exception_fail()` isn't a cross-target-language interface (part of #2400).

wsfulton:
It's hard to write tests for typemaps that work for all target languages as typemaps are meant to be target language specific. I suggest this kind of platform neutral trick:
```swig
%{
static bool checked_flag = false;
%}

%typemap(check) (int x, int y) %{
if ($1 == $2 && $1 == 123) {
  checked_flag = true;
}
%}

%catches(const char *) check_the_check_typemap;

%inline %{
void check_the_check_typemap(int x, int y) {
  if (!checked_flag)
    throw "Missing check typemap";
}
%}
```
I suggest using int instead of double as comparisons with double and == is not reliable.

The Javascript runtime test is then simply:
```javascript
typemap_check.check_the_check_typemap(123, 123);
```
If any other values are passed or the check typemap is missing, the thrown C++ exception will be translated into a language specific exception and then fail. The runtime test can be easily copied to all supported languages. I can do that post merge if you like.

erezgeva:
Hi,

I wrote li_constraints_runme for several languages. #2567
I notice that Javascripts, R, scilab do not support the check typemap.
Once you fix the support for Javascripts I can add `li_constraints_runme.js` to test them.

erezgeva:
Hi,

`Lib/constraints.i` already test the check type map.
So I only took @mmomtchev patch for Source/Modules/javascript.cxx.
Add it to #2567 and wrote a new 'Examples/test-suite/javascript/li_constraints_runme.js' to use the already existing test `Examples/test-suite/li_constraints.i`.
The test is very simple and very effective. And do not breaks other languages. 
#2567 pass all checks, very easily.

Once  #2567 is merged, you can close this one.

PR #2564, opened 2023-04-29
Add argc and argv multi-argument typemap to C sharp.

erezgeva:
We need it for: uniformity, protection on C/C++ code
 and proper mapping of string array directly to char pointer pointer.

Note:
I see an error on Windows CI.
As I do not use Windows, I can not debug it.

### DIFF
--- a/Examples/test-suite/argcargvtest.i
+++ b/Examples/test-suite/argcargvtest.i
@@ -1,9 +1,14 @@
 %module argcargvtest
 
-#if !defined(SWIGCSHARP) && !defined(SWIGD) && !defined(SWIGGUILE) && !defined(SWIGJAVA) && !defined(SWIGJAVASCRIPT) && !defined(SWIGMZSCHEME) && !defined(SWIGOCAML) && !defined(SWIGR) && !defined(SWIGSCILAB)
+#if !defined(SWIGD) && !defined(SWIGGUILE) && !defined(SWIGJAVA) && !defined(SWIGJAVASCRIPT) && !defined(SWIGMZSCHEME) && !defined(SWIGOCAML) && !defined(SWIGR) && !defined(SWIGSCILAB)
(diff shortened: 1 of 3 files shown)


erezgeva:
I see a problem with marshalling the string array on visual studio/windows.
I tried to add marshalling with `inattributes` and `ArraySubType` of [`LPStr`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype)

This error seems to persist:
`Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.`
`   at argcargvtestNamespace.argcargvtestPINVOKE.mainc(String[] jarg1)`
`   at argcargvtestNamespace.argcargvtest.mainc(String[] argc) in c:\projects\swig\Examples\test-suite\csharp\argcargvtest\argcargvtest.cs:line 15
   at argcargvtest_runme.Main() in c:\projects\swig\Examples\test-suite\csharp\argcargvtest_runme.cs:line 10`

Any Help is welcomed.

erezgeva:
I try to follow:
https://learn.microsoft.com/en-us/dotnet/framework/interop/marshalling-different-types-of-arrays

And try to add `CallingConvention = CallingConvention.Cdecl` after `DllImport`, yet the same exception.

Perhaps there is a bug in the Windows Visual Studio?
I did see we use 3 versions of Visual Studio, which really rise the question.

erezgeva:
The marshaling must use [SizeParamIndex](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshalasattribute.sizeparamindex)
And pass a parameter with the actual array size.

erezgeva:
Move to commit to [#2552](https://github.com/swig/swig/pull/2552)

Issue #2538, opened 2023-04-19
Remove SWIG/D - it doesn't work with current D releases


ojwb:
The CI for D has been disabled since the switch to GHA.

I had a poke at it to try to get it running.  The conclusions aren't good.

We use pre-built D packages from https://downloads.dlang.org/releases/2.x/ - the older builds here don't work on the Ubuntu releases which GHA still supports (due to depending on libcurl3, which 20.04 and newer releases don't have).  I did a kind of binary chop and found the oldest build which installs on ubuntu-20.04 is D 2.080.0.  However this and every newer version I've tried (2.081.0, 2.081.2, 2.082.1, 2.084.1, 2.086.1 and 2.103.0 - currently the most recent there) don't like the code SWIG generates:

```
+ make -j2 check-d-examples 'CFLAGS=-Werror -fdiagnostics-show-option -std=gnu89 -Wno-long-long -Wreturn-type -Wmissing-field-initializers -Wdeclaration-after-statement' 'CXXFLAGS=-Werror -fdiagnostics-show-option -std=c++98 -Wno-long-long -Wreturn-type -Wmissing-field-initializers'
checking Examples/d/callback
checking Examples/d/class
example.d(81): Error: mutable method `example.Callback.run` is not callable using a `const` `this`
example.d(63): Error: template instance `example.Callback.swigIsMethodOverridden!(void delegate(), void function(), run)` error instantiating
make[3]: *** [../../../Makefile:322: d_compile] Error 1
make[2]: *** [../example.mk:48: build] Error 2
make[2]: Target 'check' not remade because of errors.
make[1]: *** [Makefile:215: callback.actionexample] Error 2
checking Examples/d/constants
checking Examples/d/enum
checking Examples/d/extend
checking Examples/d/funcptr
example.d(105): Error: mutable method `example.Employee.getTitle` is not callable using a `const` `this`
example.d(71): Error: template instance `example.Employee.swigIsMethodOverridden!(string delegate(), string function(), getTitle)` error instantiating
example.d(105): Error: mutable method `example.Employee.getName` is not callable using a `const` `this`
example.d(76): Error: template instance `example.Employee.swigIsMethodOverridden!(string delegate(), string function(), getName)` error instantiating
example.d(200): Error: mutable method `example.Employee.getTitle` is not callable using a `const` `this`
example.d(182): Error: template instance `example.Manager.swigIsMethodOverridden!(string delegate(), string function(), getTitle)` error instantiating
[...]

PR #2567, opened 2023-04-30
Add 'can throw' flag to D language exceptions.

erezgeva:
Merge it with C# flag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,15 +50,14 @@ jobs:
         - SWIGLANG: ""
           compiler: clang
         - SWIGLANG: csharp
-        # D support can't be enabled because dmd 2.066 fails to build anything
-        # under Ubuntu 18.04 due to its standard library (libphobos2.a) not
-        # being compiled with -FPIC, but system gcc using -fpie by default,
-        # resulting in linking errors for any output. And later versions, such
-        # as 2.086.1, are not supported and result in errors in SWIG test suite.
-        #
-        # - SWIGLANG: d
-        #   VER: '2.066.0'
-        #   os: ubuntu-18.04 # This dlang version doesn't work under 20.04.
+        - SWIGLANG: d
+          VER: 'ldc'
+        - SWIGLANG: d
+          VER: 'gdc'
+        - SWIGLANG: d
+          VER: 'gdmd'
+        - SWIGLANG: d
+          VER: '2.103.1'
         - SWIGLANG: go
           VER: '1.6'
           CSTD: gnu11
--- a/.gitignore
+++ b/.gitignore
@@ -68,6 +68,7 @@ CCache/config.log
 CCache/config.status
 CCache/config_win32.h
 Examples/Makefile
+Examples/d/example.mk
 Examples/guile/Makefile
 Examples/test-suite/*/Makefile
 Examples/xml/Makefile
--- a/Examples/Makefile.in
+++ b/Examples/Makefile.in
@@ -280,7 +280,7 @@ csharp_clean:
 
 DLIBPREFIX = @DLIBPREFIX@
 
-SWIGD = $(SWIG) -d -d2
+SWIGD = $(SWIG) -d
 DCOMPILER = @D2COMPILER@
 
 D_RUNME = ./$(RUNME)
@@ -324,7 +324,7 @@ d_run:
 # -----------------------------------------------------------------
 
 d_version:
-	($(DCOMPILER) --version 2> /dev/null || $(DCOMPILER)) | head -n 3
+	($(DCOMPILER) --version 2> /dev/null) | head -n 3
 
 # -----------------------------------------------------------------
 # Clean the D examples
--- a/Examples/d/example.mk.in
+++ b/Examples/d/example.mk.in
@@ -14,7 +14,12 @@ EXTRA_CXXFLAGS =
 EXTRA_LDFLAGS  =
 TARGET         = example_wrap
 SWIGOPT        = -outcurrentdir
+DCOMP_TYPE    = @D2COMPILER_TYPE@
+ifeq ($(DCOMP_TYPE),GCC)
+DFLAGS         = -orunme
+else
 DFLAGS         = -ofrunme
+endif
 
 ifeq (,$(SRCDIR))
 DSRCS          = *.d
--- a/Examples/d/extend/d2/runme.d
+++ b/Examples/d/extend/d2/runme.d
@@ -11,7 +11,7 @@ public:
     super( name );
   }
 
-  override string getPosition() const {
+  override string getPosition() {
     return "CEO";
   }
 
--- a/Examples/d/extend/example.h
+++ b/Examples/d/extend/example.h
@@ -13,7 +13,7 @@ public:
 	Employee(const char* n): name(n) {}
 	virtual std::string getTitle() { return getPosition() + " " + getName(); }
 	virtual std::string getName() { return name; }
-	virtual std::string getPosition() const { return "Employee"; }
+	virtual std::string getPosition() { return "Employee"; }
 	virtual ~Employee() { printf("~Employee() @ %p\n", (void *)this); }
 };
 
@@ -21,7 +21,7 @@ public:
 class Manager: public Employee {
 public:
 	Manager(const char* n): Employee(n) {}
-	virtual std::string getPosition() const { return "Manager"; }
+	virtual std::string getPosition() { return "Manager"; }
 };
 
 
--- a/Examples/test-suite/d/Makefile.in
+++ b/Examples/test-suite/d/Makefile.in
@@ -11,6 +11,12 @@ HAVE_CXX20   = @HAVE_CXX20@
 srcdir       = @srcdir@
 top_srcdir   = ../@top_srcdir@
 top_builddir = ../@top_builddir@
+DCOMP_TYPE  = @D2COMPILER_TYPE@
+ifeq ($(DCOMP_TYPE),GCC)
+DCOMP_OUT   = -o
+else
+DCOMP_OUT   = -of
+endif
 
 CPP_TEST_CASES = \
 	d_nativepointers \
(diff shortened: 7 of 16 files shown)


ojwb:
This seems plausible, but really should have a regression testcase.

Also, how are you testing SWIG/D?  SWIG doesn't seem to work with any vaguely recent D releases and will probably get removed or at least demoted to experimental status in 4.2.0 unless somebody updates it (see #2538).

erezgeva:
I do not insist and did not use D yet.
I simply add argcargv.i to D and saw it uses the same can throw exception flag.

I use Debian 11
`$ cat /etc/debian_version `
`11.7`

With LLVM D Compiler
Version 1.24.0-2

As for testing.
I can translate the C# exception test to D language.
If you can point to the relevant testing.

ojwb:
I'd certainly be happier if the CI used Debian, but the OS we're forced to use isn't the problem here.

The libcurl3 issue is that the older binary D releases (ones old enough that SWIG might work with them) were built for old Ubuntu versions which had libcurl3, which was removed after bionic:

```
$ rmadison -uubuntu libcurl3
 libcurl3 | 7.35.0-1ubuntu2    | trusty          | amd64, arm64, armhf, i386, powerpc, ppc64el
 libcurl3 | 7.35.0-1ubuntu2.20 | trusty-security | amd64, arm64, armhf, i386, powerpc, ppc64el
 libcurl3 | 7.35.0-1ubuntu2.20 | trusty-updates  | amd64, arm64, armhf, i386, powerpc, ppc64el
 libcurl3 | 7.47.0-1ubuntu2    | xenial          | amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 libcurl3 | 7.47.0-1ubuntu2.19 | xenial-security | amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 libcurl3 | 7.47.0-1ubuntu2.19 | xenial-updates  | amd64, arm64, armhf, i386, powerpc, ppc64el, s390x
 libcurl3 | 7.58.0-2ubuntu2    | bionic/universe | amd64, arm64, armhf, i386, ppc64el, s390x
```

If you run that command (on a Debian machine) without `-uubuntu` it shows nothing as even Debian oldstable had transitioned from libcurl3 to libcurl4.

Newer upstream D binary releases don't have this issue (presumably they picked up a dependency on libcurl4 instead), but SWIG/D doesn't work with them, and GHA no longer supports any of the Ubuntu versions with libcurl3.

However, even without the libcurl3 issue, it's clearly a problem that SWIG/D doesn't work with any D release made for many years (somewhere between 5 and 9 years it seems).

Your change is to `constraints.i` which is tested by `Examples/test-suite/li_constraints.i`.  We used to have CI for SWIG/D and not much has changed in this area since, so presumably this testcase successfully processes and builds given a suitably ancient D compiler, in which case you'd probably need to add a `Examples/test-suite/d/li_constraints_runme.d` such that this testcase fails without your patch and passes with.

However, this all requires being able to actually successfully run the SWIG/D testsuite.  Without that a regression test is useless as it doesn't demonstrate that the patch fixes the issue, and it doesn't detect if a future proposed change would reintroduce the issue.

I kicked off a CI run hacked to just try ldc and gdc for Ubuntu 20.04 and 22.04 (the two versions GHA currently supports) but they fail too (looks like the same errors as with the upstream compiler): https://github.com/swig/swig/actions/runs/4856429855

erezgeva:
Fair point.
I mean the CI should first work.

BYW On my Debain I use ldc, not sure if gcc works.
https://packages.debian.org/source/bullseye/ldc

As for libcurl. My Debain have both libcurl3-gnutls and 
libcurl4. And ldc does not looks like it depends any  libcurl.

I know that GHA runners are Ubuntu, but we can use Docker containers.
Not sure if D language is that important, but as similar issue might occur with a more important  language. I think we should support working with containers.
We can also test other issues that might occur on Fedora or other Linux distributions. I do not suggest inflate the CI, just a though of flexibility.

I do plan to run my project with several Docker containers, so once I manage do so in my project, I'd be happy to assist.

Mean while, I will add `Examples/test-suite/d/li_constraints_runme.1.d` and `Examples/test-suite/d/li_constraints_runme.2.d`.

I might have a second thought about the D language,
 but I think we can drop D language version 1 (D1).

According to Wikipedia. 
https://en.wikipedia.org/wiki/D_(programming_language)
D1 is discontinued from 2012.

erezgeva:
I tried 
`make check-d-examples` and
`make check-d-test-suite` on my Debian.

I see errors not related to libcurl, but
`Error: mutable method example.Employee.getTitle is not callable using a const this`
And
`Error: cannot implicitly override base class method access_change.DerivedInt.DerivedInt.WasProtected2 with access_change.BottomInt.BottomInt.WasProtected2; add override attribute`

I suggest open a new PR with:
1.  Removing D1 as it is obsolete
2. Fix the `check-d-examples` and `check-d-test-suite` on my Debian.
3. Try to add 1 working CI on GHA with LLVM D2 hopefully on Ubuntu.

We might drop old compilers and obsolete versions.
But I think we can support new D2 language compilers.

erezgeva:
Hi,

I search
`$ find -name  'li_constraints*'`
`./li_constraints.i`

There is no such test, yet.
Do you want me to write a new one?
Seems we need to add to all languages.
Not just D language.

ojwb:
> BYW On my Debain I use ldc, not sure if gcc works. https://packages.debian.org/source/bullseye/ldc

I've no idea if gdc works/should work, but given I was hacking up a CI test of ldc it was trivial extra work to also get it test gdc.

> As for libcurl. My Debain have both libcurl3-gnutls and libcurl4. And ldc does not looks like it depends any libcurl.

The upstream binary builds old enough that SWIG supports them won't install because they depend on the libcurl3 package - different packages with a similar names aren't going to help (unless they have `Provides: libcurl3` or similar, which they don't).

I really wouldn't fixate on libcurl.  That just prevents us testing with D versions old enough that SWIG currently supports them.  I would say the critical problem here is that SWIG doesn't support any vaguely recent D version (which is better discussed in #2538 rather than this ticket).

> I know that GHA runners are Ubuntu, but we can use Docker containers. Not sure if D language is that important, but as similar issue might occur with a more important language. I think we should support working with containers. We can also test other issues that might occur on Fedora or other Linux distributions. I do not suggest inflate the CI, just a though of flexibility.

This would probably provide a way to install a version of D that SWIG can work with (by running in a docker of an unsupported Ubuntu version), but that complicates the CI machinery and doesn't actually address the problem of not supporting current D version.

> Mean while, I will add `Examples/test-suite/d/li_constraints_runme.1.d` and `Examples/test-suite/d/li_constraints_runme.2.d`.
> 
> I might have a second thought about the D language, but I think we can drop D language version 1 (D1).
> 
> According to Wikipedia. https://en.wikipedia.org/wiki/D_(programming_language) D1 is discontinued from 2012.

Good point, I've noted that in #2538.  I certainly wouldn't worry about adding a new runme for D1.

> I tried `make check-d-examples` and `make check-d-test-suite` on my Debian.
> 
> I see errors not related to libcurl, but `Error: mutable method example.Employee.getTitle is not callable using a const this` And `Error: cannot implicitly override base class method access_change.DerivedInt.DerivedInt.WasProtected2 with access_change.BottomInt.BottomInt.WasProtected2; add override attribute`

Yes - those are what you get with current D releases.  That needs fixing if we're to continue to have D support in SWIG.

I've know basically nothing about D, so I've no idea what's going wrong.

> I suggest open a new PR with:
> 
>     1. Removing D1 as it is obsolete

I'd suggest getting wsfulton to agree first as he can be resistant to dropping support for obsolete versions and I don't want to get into commit revert wars.  Discuss in #2538

>     2. Fix the `check-d-examples` and `check-d-test-suite` on my Debian.
> 
>     3. Try to add 1 working CI on GHA with LLVM D2 hopefully on Ubuntu.

That'd be great, but these also belong in #2538.

> I search `$ find -name 'li_constraints*'` `./li_constraints.i`
> 
> There is no such test, yet. Do you want me to write a new one? Seems we need to add to all languages. Not just D language.

Yes, there is no existing `_runme` for this.  It's the only testcase that uses `constraints.i`, which is the file you're patching, so if the only existing candidate for a regression test.

Ideally we'd have a `_runme` for every target language for almost every testcase (a few do all the testing they need just by being processed and compiled (e.g. some of the `preproc_*.i` tests which test preprocessor features are set up to `#error` if the test fails) and some are specific to a particular target language or a particular feature that isn't supported by all target languages.

But it's not feasible to require someone adding a new testcase to add a complete set of `_runme` files.  It's already hard enough to get people to add testcases; if they need to learn N different new languages to do it then nobody will want to.

I'm surprised there's no `_runme` for this.  Looking at the history it's over 20 years old, which is well before we really got our act together with testing so I guess that's why not.

Adding missing `_runme` files is certainly useful.

erezgeva:
[I add Add li_constraints test](https://github.com/swig/swig/commit/d77f2877740a2e3b6e284a4efc9392289f47b748)
For: D, C#, go, Java, Lua, Perl, PHP, python, Ruby, TCL, Octave.

Javascripts, R, scilab do not support the check typemap.
Javascripts is mention in #2562 and #2561

I can add li_constraints_runme for JS that actually test the check typemap.

erezgeva:
The error in R, ruby ares not related to this PR.
As it have D language patches.

erezgeva:
Found my bug:
I had `const String *pn = Getattr(n, "name");`
Instead of `const String *pn = Getattr(e, "name");`

So I compare the wrong name.

erezgeva:
I rebase on #2570.
So we can prove the patch works.

ojwb:
Just leave this alone until we have #2570 merged.  It's not usefully reviewable now anyway, as the rebase means it now shows all the changes from #2570 as well.

erezgeva:
> Just leave this alone until we have #2570 merged. It's not usefully reviewable now anyway, as the rebase means it now shows all the changes from #2570 as well.

Sure, I was just getting ןmpatient.
But I should respect your time and efforts.

erezgeva:
@ojwb
Thanks for your quick merge and excellent work.

PR #2572, opened 2023-05-06
Javascript check typemap fix

erezgeva:
Take the Javascript fix from #2562
Add li_constraints.i  test to  Javascript, C#, go, Java, Lua, Perl, PHP, python, Ruby, TCL and Octave.

Thanks @mmomtchev for the Javascript fix.

### DIFF
--- a/Examples/test-suite/go/li_constraints_runme.go
+++ b/Examples/test-suite/go/li_constraints_runme.go
@@ -1,0 +1,70 @@
+package main
+
+import wrap "swigtests/li_constraints"
+import "fmt";
+
+func check_double(except bool, f func (float64), val float64, name string) {
+    actual := true
+    proper := false
+	func() {
+		defer func() {
+            r := recover();
+			if r != nil {
+                actual = false
+                proper = fmt.Sprintf("%s", r) == fmt.Sprintf("Expected a %s value.", name)
+			}
+		}()
+        f(val);
+	}()
+    if actual {
+      if !except {
+        panic(fmt.Sprintf("function '%s' with %d should perform an exception", name, val));
+      }
+    } else {
+      if except {
+        panic(fmt.Sprintf("function '%s' with %d should not perform an exception", name, val));
+      } else if !proper {
+        panic(fmt.Sprintf("function '%s' with %d should perform a proper exception", name, val));
+      }
+    }
+}
+
+func main() {
+    check_double(true, wrap.Test_nonnegative, 10, "non-negative");
+    check_double(true, wrap.Test_nonnegative, 0, "non-negative");
+    check_double(false, wrap.Test_nonnegative, -10, "non-negative");
+
+    check_double(false, wrap.Test_nonpositive, 10, "non-positive");
+    check_double(true, wrap.Test_nonpositive, 0, "non-positive");
+    check_double(true, wrap.Test_nonpositive, -10, "non-positive");
+
+    check_double(true, wrap.Test_positive, 10, "positive");
+    check_double(false, wrap.Test_positive, 0, "positive");
+    check_double(false, wrap.Test_positive, -10, "positive");
+
+    check_double(false, wrap.Test_negative, 10, "negative");
+    check_double(false, wrap.Test_negative, 0, "negative");
+    check_double(true, wrap.Test_negative, -10, "negative");
+
+    check_double(true, wrap.Test_nonzero, 10, "nonzero");
+    check_double(false, wrap.Test_nonzero, 0, "nonzero");
+    check_double(true, wrap.Test_nonzero, -10, "nonzero");
+
+    have_exception := false
+	func() {
+		defer func() {
+            r := recover()
+			if r != nil {
+                have_exception = "Received a NULL pointer." == fmt.Sprintf("%s", r)
+			}
+		}()
+        // The NULL value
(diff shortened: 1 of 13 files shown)


erezgeva:
The error in R is not related to this PR.
As it does not contain any change related to R.

Examples/test-suite/perl5/li_constraints_runme.pl
```diff
@@ -0,0 +1,44 @@
...
+check_double(0, "negative", 10);
+check_double(0, "negative", 0);
+check_double(1, "negative", -10);
+
+check_double(1, "nonzero", 10);
+check_double(0, "nonzero", 0);
+check_double(1, "nonzero", -10);
+
+# Pass null value
+my $ret = eval "li_constraints::test_nonnull(undef); 1";
```
ojwb, line 40: `eval` with a block would be better Perl I think as then the code is only parsed once before the script starts., i.e.:

```
my $ret = eval { li_constraints::test_nonnull(undef); 1 };
```

Similarly above.

ojwb: Oh, except the above one generates the code so that can't just be a block.

Examples/test-suite/go/li_constraints_runme.go
```diff
@@ -0,0 +1,61 @@
...
+		}()
+        // The NULL value
+        wrap.Test_nonnull(uintptr(0));
+	}()
+    if !have_exception {
+        panic("test_nonnull should perform exception with 'null' value")
+    }
+    /* The pointer value is not actually used,
+     * We just want it to be a non-null */
+    wrap.Test_nonnull(uintptr(111));
```
ojwb, line 60: Does this end up doing something like `(void*)111`?  If so, that's undefined behaviour isn't it?

I'd suggest adding a helper method to `li_constraints.i` which returns a valid non-NULL `void*` pointer, e.g.:

````
void* get_nonnull() {
    static int i;
    return &i;
}
```

That should then allow testing this for all the target languages, e.g. here it would be something like:

```
wrap.Test_nonnull(wrap.Get_nonnull());
```

Examples/test-suite/perl5/li_constraints_runme.pl
erezgeva: It works on my machine.
What do you prefer?

Examples/test-suite/go/li_constraints_runme.go
erezgeva: Sound a good continue.
I was only seeking to write as many tests as I can without change li_constraints.i.
Only bug fixes.
But your idea should work with most other languages, so it will actually complete the test with other languages.
Only go language can cast to void.
Yes `uintptr(111)` equal `(void*)111`. 
As all pointers from go are translate to `uintptr` in our wrapper.

erezgeva:
Do you want to add 
`void* get_null() {`
`    return NULL;`
`}`

For TCL?

Examples/test-suite/perl5/li_constraints_runme.pl
ojwb: It works, but I think the block form is better when you can use it (as in this case).

Examples/test-suite/go/li_constraints_runme.go
ojwb: Probably new commits is simpler for review.  I can squash as I merge if appropriate.

ojwb:
> Do you want to add `void* get_null() {` ` return NULL;` `}`
> 
> For TCL?

Doc/Manual/Tcl.html says: "_The NULL pointer is represented by the string <tt>NULL</tt>._"

So I think you should be able to just use `"NULL"` for that in Tcl.  `git grep NULL Examples/test-suite/tcl` shows several existing testcases do, for example:

```
Examples/test-suite/tcl/null_pointer_runme.tcl:if {$null != "NULL"} {
Examples/test-suite/tcl/null_pointer_runme.tcl:if {[llength [info commands "NULL"]] != 0} {
```

erezgeva:
> > Do you want to add `void* get_null() {` ` return NULL;` `}`
> > For TCL?
> 
> Doc/Manual/Tcl.html says: "_The NULL pointer is represented by the string NULL._"
> 
> So I think you should be able to just use `"NULL"` for that in Tcl. `git grep NULL Examples/test-suite/tcl` shows several existing testcases do, for example:
> 
> ```
> Examples/test-suite/tcl/null_pointer_runme.tcl:if {$null != "NULL"} {
> Examples/test-suite/tcl/null_pointer_runme.tcl:if {[llength [info commands "NULL"]] != 0} {
> ```

Surprisingly, it work also as Input.
I update the script.

ojwb:
> Surprisingly, it work also as Input. I update the script.

It's handled by `SWIG_Tcl_ConvertPtrFromString()`:

```
    if (strcmp(c,"NULL") == 0)
      return (flags & SWIG_POINTER_NO_NULL) ? SWIG_NullReferenceError : SWIG_OK;
```

Examples/test-suite/perl5/li_constraints_runme.pl
erezgeva: I replace the string evaluation with function referee or lambada functions in all scripts.

PR #2575, opened 2023-05-11
Fix handling of template in array size

### LINKED ISSUE #2486: swig crash when attempting to parse header with std::numeric_limits template
I get an error and a crash when trying to generate a wrapper for the following C++ code:

```
#pragma once

#include <limits>

typedef struct {
   unsigned char myarray[std::numeric_limits<unsigned char>::max()];
} array_t;
```

The error is as follows:
> Bad template type passed to SwigType_remember: a(std::numeric_limits< unsigned char >::max()).unsigned char
> swig: Swig/typesys.c:1706: SwigType_remember_clientdata: Assertion `0' failed.
> Aborted (core dumped)

ojwb:
This was being rejected by SWIG because the type string contains '<' not followed by '('.  That check seems too strict given types can contain constant expressions which can contain '<' characters in several ways, so let's just try removing the check and see what happens.

Fixes #2486

### DIFF
--- a/Examples/test-suite/cpp11_constexpr.i
+++ b/Examples/test-suite/cpp11_constexpr.i
@@ -45,9 +45,7 @@ struct ConstExpressions {
   constexpr friend bool operator==(ConstExpressions,ConstExpressions) { return true; }
   friend constexpr bool operator!=(ConstExpressions,ConstExpressions) { return false; }
 };
-%}
 
-%{
 int Array10[AAA];
 int Array20[BBB];
 int Array30[CCC()];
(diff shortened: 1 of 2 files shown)


ojwb:
@wsfulton CI passed and this fixes the case reported here (which I've added to the testsuite).  Does removing this check seem OK to you?

ojwb:
Appveyor decided to test both the branch and PR, so I cancelled the branch run as it's just redundant.

ojwb:
It occurred to me that we really should have test coverage for an expression used in a type which contains `<(` and `)>` so I've added one.

ojwb:
I just happened upon this commit while looking for something else entirely:

```
commit 0341258af7003121cdddda78c868ea2d497f8364
Author: William S Fulton <wsf@fultondesigns.co.uk>
Date:   Tue Nov 22 08:30:39 2022 +0000

    Fix seg fault handling template parameter expressions containing '<='
    
    Recent commits ensure types are correctly stored in SwigType *. In
    particular template parameters are enclosed within '<(' and ')>'.
    Now we can confidently handle template parameters as really being
    delimited as such to fix an infinite loop handling template expressions
    containing '<' or '>'. The previous implementation only assumed
    template parameters were delimited by '<' and '>'.
    
    Issue #1037
```

Based on this commit message it seems removing this check is now correct (but probably wouldn't have been before 0341258af7003121cdddda78c868ea2d497f8364) so I'll merge this change.

PR #2583, opened 2023-05-15
Add argc and argv multi-argument typemap to MzScheme.

erezgeva:


Lib/mzscheme/argcargv.i
```diff
@@ -0,0 +1,46 @@
...
+      /* All elements are strings! */
+      $1 = (i == len);
+    }
+  }
+}
+
+%typemap(freearg) (int ARGC, char **ARGV) {
+  if ($2 != NULL) {
+    SWIG_free((void *)$2);
+  }
```
ojwb, line 45: You don't need a `NULL`-check before calling `free()` or `delete`.  The ISO C and C++ standards guarantee that it's safe to call these on a NULL pointer, so it's not necessary for the calling code to also check.

I cleaned up all the cases I could find last year in 7ec2f89fe241e3aabd988d3aa6fe17e4878516dc so let's not introduce more.  (I removed the ones added by your other argcargv.i changes in fcdd63dba592fcfeb2fb60314e49eb5f2d185e15.)

ojwb:
```
 argcargvtest_wrap.cxx:1342:58: error: deprecated conversion from string constant to ‘char*’ [-Werror=write-strings]
 1342 | static char *_wrap_const_SWIGCSHARP_ARGCARGV_PARAM_IDX = "0";
      |                                                          ^~~
```

It looks like you're being tripped up by your own changes here, as SWIG's trying to wrap this as a constant.

Something seems clearly buggy with the mzscheme backend as this needs to be a `const char *` not `char *`, but I think this also is a sign that we need a different approach to having the user define `SWIGCSHARP_ARGCARGV_PARAM_IDX` to a value.

The current approach also seems limiting - I don't think you could generate C# wrappers for two argc/argv like functions in the same API if they have different argc parameter indexes.

I think I need to poke at this to try to understand it better.

Lib/mzscheme/argcargv.i
erezgeva: Sure.
I just followed other examples.

There are other you can clean:
Lib/octave/argcargv.i
Lib/go/argcargv.i
Lib/perl5/argcargv.i
Lib/php/argcargv.i
Lib/tcl/argcargv.i

Some you can blame me -:)

erezgeva:
> SWIGCSHARP_ARGCARGV_PARAM_IDX

I prefer it to be a number too.
But we need to insert it in a typemap.
We can try:
`#define SWIGCSHARP_ARGCARGV_PARAM_IDX 0`

And in `Lib/csharp/argcargv.i`
#define ARGCARGV_PARAM_IDX  "WIGCSHARP_ARGCARGV_PARAM_IDX"

`%typemap(imtype,..SizeParamIndex="ARGCARGV_PARAM_IDX"..`

In C/C++ precompiler I use this trick:
`#define stringify(s) #s`
`#define ARGCARGV_PARAM_IDX  stringify(SWIGCSHARP_ARGCARGV_PARAM_IDX)`
Perhaps in work in Swig as well.

Do you want me to test it?

erezgeva:
> ```
>  argcargvtest_wrap.cxx:1342:58: error: deprecated conversion from string constant to ‘char*’ [-Werror=write-strings]
>  1342 | static char *_wrap_const_SWIGCSHARP_ARGCARGV_PARAM_IDX = "0";
>       |                                                          ^~~
> ```
> 
> It looks like you're being tripped up by your own changes here, as SWIG's trying to wrap this as a constant.
> 
> Something seems clearly buggy with the mzscheme backend as this needs to be a `const char *` not `char *`, but I think this also is a sign that we need a different approach to having the user define `SWIGCSHARP_ARGCARGV_PARAM_IDX` to a value.
> 
> The current approach also seems limiting - I don't think you could generate C# wrappers for two argc/argv like functions in the same API if they have different argc parameter indexes.
> 
> I think I need to poke at this to try to understand it better.


As for using different argc parameter index.
Yes, as typemap `imtype` do not have `inattributes` parameters.
Once you set the typemap, you can not set it again.

Since ARGC ARGV is not that popular, I do not think we need to change the `imtype`  type map. User that wish to use  argc parameter in different indexes can simply define multiple ARGC ARGV typemaps like:
                                            
`%typemap(csin) char **ARGV3 "$csinput"`
`%typemap(cstype) char **ARGV3 "string[]"`
`%typemap(imtype, inattributes="[global::System.Runtime.InteropServices.In,global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPArray, ArraySubType=global::System.Runtime.InteropServices.UnmanagedType.LPStr, SizeParamIndex=3)]") char **ARGV3 "string[]"`

I think developers are intelligent, and they only need the "recipe".
It is perhaps a good remark we can add for developers.

erezgeva:
Rebase on #2602.
`SWIGCSHARP_ARGCARGV_PARAM_IDX` using number is better.

PR #2585, opened 2023-05-15
[R] Add argc and argv multi-argument typemap.

erezgeva:
Need a fix on rtypecheck.

Currently under Debug.

### DIFF
--- a/Examples/test-suite/argcargvtest.i
+++ b/Examples/test-suite/argcargvtest.i
@@ -1,6 +1,6 @@
 %module argcargvtest
 
-#if !defined(SWIGOCAML) && !defined(SWIGR)
+#ifndef SWIGOCAML
 %include <argcargv.i>
 
 %apply (int ARGC, char **ARGV) { (size_t argc, const char **argv) }
--- a/Examples/test-suite/r/argcargvtest_runme.R
+++ b/Examples/test-suite/r/argcargvtest_runme.R
@@ -1,0 +1,46 @@
+clargs <- commandArgs(trailing=TRUE)
+source(file.path(clargs[1], "unittest.R"))
+dyn.load(paste("argcargvtest", .Platform$dynlib.ext, sep=""))
+source("argcargvtest.R")
+cacheMetaData(1)
+
+largs = c("hi", "hola", "hello")
+unittest(3, mainc(largs))
+
+targs = c("hi", "hola")
+unittest("hi", mainv(targs, 0))
+unittest("hola", mainv(targs, 1))
+
+# R convert the string to a string vector with a single string.
+# So instead of exception we simply get null
+unittest(NULL, mainv("hello", 1))
+
+tryCatch({
+    mainv(c(1, 2, 3), 1)
+    # force an error if the previous line doesn't raise an exception
+    stop("Test Failure A")
+}, error = function(e) {
+    if (e$message == "Test Failure A") {
+      # Raise the error again to cause a failed test
+      stop(e)
+    }
+    message("PASS")
+})
+
+initializeApp(largs)
+initializeApp(largs, TRUE)
+
+# Check that an empty array works.
+empty_args = c()
+unittest(0, mainc(empty_args))
+
+# check dispatcher with empty array.
+initializeApp(empty_args)
+initializeApp(empty_args, FALSE, TRUE)
+
+# Check that empty strings are handled.
+empty_string = c("hello", "", "world")
+unittest(3, mainc(empty_string))
+unittest("hello", mainv(empty_string, 0))
+unittest("", mainv(empty_string, 1))
+unittest("world", mainv(empty_string, 2))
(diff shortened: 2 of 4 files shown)


ojwb:
A couple of notes for when you get this working:

I spotted a bug in the octave `argcargv.i` implementation which needs to copy the string data but wasn't doing so and as a result every entry in the argv passed to C/C++ was actually the final string from the octave array.  The testcase was only testing the value of a last entry so missed this bug.  I've modified all the existing `argcargvtest_runme.*` to test the first entry too - it would be good to copy that extra coverage over for R.

Some of the implementations accepted an empty target language array whereas others explicitly rejected it.  It seems valid for the array to be empty (resulting in `argc == 0` and `argv[0] == NULL` on the C/C++ side) so I've fixed the target languages which didn't allow this _(*)_, and added test coverage for this for all languages.  It ought to work and have test coverage for R too.

---
_(*) Looking at 4ba7e68c286e38b0ae8a4db5028e7ec4b5172538 I see I didn't finish this for D, and the testcase change for it is commented out - I'll push a fix for that soon._

erezgeva:
> A couple of notes for when you get this working:
> 
> I spotted a bug in the octave `argcargv.i` implementation which needs to copy the string data but wasn't doing so and as a result every entry in the argv passed to C/C++ was actually the final string from the octave array. The testcase was only testing the value of a last entry so missed this bug. I've modified all the existing `argcargvtest_runme.*` to test the first entry too - it would be good to copy that extra coverage over for R.
> 
> Some of the implementations accepted an empty target language array whereas others explicitly rejected it. It seems valid for the array to be empty (resulting in `argc == 0` and `argv[0] == NULL` on the C/C++ side) so I've fixed the target languages which didn't allow this _(*)_, and added test coverage for this for all languages. It ought to work and have test coverage for R too.
> 
> _(*) Looking at [4ba7e68](https://github.com/swig/swig/commit/4ba7e68c286e38b0ae8a4db5028e7ec4b5172538) I see I didn't finish this for D, and the testcase change for it is commented out - I'll push a fix for that soon._

@ojwb 
Thanks for catching these bugs and adding more testings :-)
As for R, it have an issue with the `rtypecheck` typemap.
It does not use it properly, as a result the dispatch of `initializeApp()` always fail.

I though fixing the using of `rtypecheck` typemap in R.
Once it is works properly, I will update the R `runme` to follow other `argcargvtest_runme` tests.
And ensure R `argcargv.i` works the same as other languages.

erezgeva:
I think I found the missing `rtypecheck` in r`.cxx`.
Once the test pass, I will create proper patches.

R also use `NULL` for empty array like D lang.
Probably this idea is more spread ...

erezgeva:
@wsfulton
I saw this message in ``Source/Modules/r.cxx":
`Below should be migrated into rtypecheck typemaps`

I add `%typemap(rtypecheck)`, so perhaps we now can?

erezgeva:
@ojwb

Ready for merge.

OCaml does not have ARGC ARGV typemap,
There is probably a good reason to remove this language.
There is no way to pass string array from OCaml to Swig wrapper.

ojwb:
Gah, I'm in the middle of merging this and you've just updated it.

If you're still working on a branch, please don't say it's "Ready to merge".

ojwb:
Rebasing at this point is actively unhelpful - please stop changing things here unless there's actually something which should hard block the merge.  Improvements and minor fixes we can apply fixes afterwards.

erezgeva:
> Rebasing at this point is actively unhelpful - please stop changing things here unless there's actually something which should hard block the merge. Improvements and minor fixes we can apply fixes afterwards.

OK, I stop.
I think it is in a good shape.

ojwb:
OK, merged.

I did a `git merge --squash` of the PR branch on top of my merge and included that in the push, so I think all your changes are included, but it needed significant manual resolving of conflicts so probably worth you checking.

I noticed while merging that we weren't `NULL` terminating `argv` in the empty array case for R (which your refactor also fixed) and for C# at all, so I've fixed those and added test coverage to check we get this right for all the target languages in 8a73e25fe2a1bd1bbffb052d47891238287aea7d.

erezgeva:
> OK, merged.
> 
> I did a `git merge --squash` of the PR branch on top of my merge and included that in the push, so I think all your changes are included, but it needed significant manual resolving of conflicts so probably worth you checking.
> 
> I noticed while merging that we weren't `NULL` terminating `argv` in the empty array case for R (which your refactor also fixed) and for C# at all, so I've fixed those and added test coverage to check we get this right for all the target languages in [8a73e25](https://github.com/swig/swig/commit/8a73e25fe2a1bd1bbffb052d47891238287aea7d).

I compared `Lib/r/argcargv.i`. 
The only different is I removed the `noblock` attribute from `%typemap(in)`, as C forbids defining variables in a middle of a block. (We run tests in C++ only).

As for `NULL` terminating, I probably missed it. Nice catch!

Thanks for merging, enhancing, finding bugs and your time.
Erez

ojwb:
> The only different is I removed the `noblock` attribute from `%typemap(in)`, as C forbids defining variables in a middle of a block. (We run tests in C++ only).

Oops, pushed a fix for that.

It would probably also fail for C++ if the typemap applied twice to the same function due to declaring a variable which is already declared in that scope.

PR #2588, opened 2023-05-16
Alias stdint.i types in std for C++

### LINKED ISSUE #2532: stdint.i should probably alias types in std namespace for C++
Or at least that's what I believe is happening. 
When compiling a very large generated CXX file for Python, I'm getting the following error:

> file_wrap.cxx:5641:22: error: redefinition of ‘struct swig::traits<unsigned char>’
>  5641 |   template <> struct traits< unsigned char > {
>       |                      ^~~~~~~~~~~~~~~~~~~~~~~
> file_wrap.cxx:3672:25: note: previous definition of ‘struct swig::traits<unsigned char>’
>  3672 |     template <>  struct traits< std::uint8_t > {
>       |                         ^~~~~~~~~~~~~~~~~~~~~~

Looking at the generated file at the above positions, I see both code examples:

```cpp

namespace swig {
  template <> struct traits< unsigned char > {
    typedef value_category category;
    static const char* type_name() { return"unsigned char"; }
  };
  template <>  struct traits_asval< unsigned char > {
    typedef unsigned char value_type;
    static int asval(PyObject *obj, value_type *val) {
      return SWIG_AsVal_unsigned_SS_char (obj, val);
    }
  };
  template <>  struct traits_from< traits_asval > {
    typedef unsigned char value_type;
    static PyObject *from(const value_type& val) {
      return SWIG_From_unsigned_SS_char  (val);
    }
  };
}
```
and

```cpp
  namespace swig {
    template <>  struct traits< std::uint8_t > {
      typedef pointer_category category;
      static const char* type_name() { return"std::uint8_t"; }
    };
  }

```

After deleting the first code block from the file (as a whole - including all these `< unsigned char >` templates), the code compiles to an `.o` file.

ojwb:
This is consistent with <stdint.h> vs <cstdint> in a C++ compiler as <stdint.h> is permitted to additionally provide the std:: versions.

Fixes #2532


ojwb:
It seems including both `<stdint.h>` and `<cstdint>` (in either order) doesn't result in both `std::uint8_t` and `uint8_t` being available for all compilers, so this needs more work.

We could provide a separate `.i` for C++ use, but that seems to create a bit of a gotcha for C++ users who spot `stdint.i`.  Will ponder.  I'd set this to be a draft, but I don't see how to.

PR #2589, opened 2023-05-16
WIP: Skip declarations SWIG can't parse

ojwb:
There are still many valid C++ constructs which SWIG can't parse (typically from newer C++ standards) which currently halt SWIG with a parse error.

We've had quite a lot of user reports where this happens but the user doesn't actually want to wrap the problematic declaration, so let's try getting SWIG to emit a warning and skip to the next declaration.

### DIFF
--- a/Examples/test-suite/errors/c_missing_rbrace.stderr
+++ b/Examples/test-suite/errors/c_missing_rbrace.stderr
@@ -1,2 +1,1 @@
 c_missing_rbrace.i:3: Error: Missing '}'. Reached end of input.
-c_missing_rbrace.i:3: Error: Syntax error in input(1).
--- a/Examples/test-suite/errors/c_spaceship.stderr
+++ b/Examples/test-suite/errors/c_spaceship.stderr
@@ -1,1 +1,2 @@
-c_spaceship.i:3: Error: Syntax error in input(1).
+c_spaceship.i:3: Warning 999: Attempting to skip declaration SWIG can't parse
+c_spaceship.i:4: Error: Missing semicolon (';'). Reached end of input.
--- a/Examples/test-suite/errors/cpp_invalid_exponents1.stderr
+++ b/Examples/test-suite/errors/cpp_invalid_exponents1.stderr
@@ -1,2 +1,2 @@
 cpp_invalid_exponents1.i:3: Error: Exponent does not have any digits
-cpp_invalid_exponents1.i:3: Error: Syntax error in input(1).
+cpp_invalid_exponents1.i:3: Warning 999: Attempting to skip declaration SWIG can't parse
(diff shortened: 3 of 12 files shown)


ojwb:
I'm interested in people's thoughts on this.

I think my main concerns are that:

(a) `skip_decl()` may not resynchronise us at a sensible place
(b) this change may make SWIG's output more confusing when the input isn't valid C/C++ code, e.g. from the testsuite:

```
int foo(vector<int);
```

currently fails with the vague but correct:

```
cpp_missing_rtemplate.i:4: Error: Syntax error in input(1).
````

and with this patch fails with:

```
cpp_missing_rtemplate.i:4: Warning 999: Attempting to skip declaration SWIG can't parse
cpp_missing_rtemplate.i:5: Error: Missing semicolon (';'). Reached end of input.
```

That statement has multiple problems, but not a missing semicolon.  I think what's happening here is that the parser has consumed to EOF already before the parse fails, and maybe this case could be addressed, but I think it illustrates the point.  The counterpoint is that the patch may arguably improve output for some other cases, and of course also the main point of it - that it stops deficiencies in SWIG's parser being blockers for wrapping some unmodified headers (especially those using modern C++ features).

Possibly some more targetted error recovery instead of this rather global approach is the answer (e.g. #2593 uses error recovery to support parsing `decltype` with arbitrary expressions - maybe we could do more similar things).

**Comment by wsfulton** [comment]
I think the patch as it stands will silently swallow errors, which if I've got this right, will need addressing.

As you ask, I would rather handle errors in a more targetted fashion rather than globally. That feels more sensible and less likely to result in odd behaviour.

**Comment by ojwb** [comment]
It shouldn't *silently* swallow as you should always get the new warning instead unless the attempt to resync emits an error (which also isn't silent).  If you explicitly ask SWIG to suppress that warning then it'll be silent, but then you asked for exactly that!  (Ideally you'd be able to suppress known OK instances of the warning but still see new ones, which I don't really see a good way to achieve - in general specifying a symbol to `%warnfilter` probably won't work here as we're failing to parse the declaration which would add that symbol.)

It's also really more of an experimental sketch of a patch at this point too, as I wanted to see if it was implementable and to be able to try that on real world examples from bug reports so don't read too much into exact details of the implementation.  

I think I agree that we should try more targetted error recovery first though.  I certainly understand the nuances of bison error recovery (and SWIG's scanner API) better thanks to trying this!

**Comment by ojwb** [comment]
I'm going to close this unmerged.  Maybe it's worth a revisit at some point, but there's seems to be a fairly inherent tension between skipping parser failures due to shortcomings of our parser and not skipping genuine parse errors in an interface file.

Something tangential to this I'm going to look into further is that this patch avoids a "Syntax error in input(1)." after an error like "Missing '}'. Reached end of input." - the first error here is the relevant one, the second is just not useful  Here's an example from the diff:

```
c_missing_rbrace.i:3: Error: Missing '}'. Reached end of input.
c_missing_rbrace.i:3: Error: Syntax error in input(1).
```

ojwb:
> Something tangential to this I'm going to look into further is that this patch avoids a "Syntax error in input(1)."

df5f5497b8565131024a8dcadc91d72409e04902 addresses that.

Issue #1943, opened 2021-02-02
valarray suport


cesarep:

cesarep:
After a couple tries i managed to successfully interface valarrays with Python without vectors:
```
%typemap(out) valarray<double>{
  $result = PyList_New($1.size());
  for (size_t i = 0; i < $1.size(); i++) {
    PyObject *o = PyFloat_FromDouble((double) $1[i]);
    PyList_SetItem($result, i, o);
  }
}

%typecheck(SWIG_TYPECHECK_DOUBLE_ARRAY) valarray<double> {
  $1 = PySequence_Check($input) ? 1 : 0;
}

%typemap(in) valarray<double> (valarray<double> temp){
  if (!PySequence_Check($input)) {
    PyErr_SetString(PyExc_TypeError, "Expecting a sequence");
    SWIG_fail;
  }
  size_t l = PySequence_Length($input);
  temp.resize(l);
  for (size_t i = 0; i < l; i++) {
    PyObject *o = PySequence_GetItem($input, i);
    if (PyNumber_Check(o)) {
      temp[i] = (double) PyFloat_AsDouble(o);
    } else {
      PyErr_SetString(PyExc_ValueError, "Sequence elements must be numbers");      
      SWIG_fail;
    }
  }
  $1 = temp;
}
```
Though i'm not sure about performance nor any other factors.

wsfulton:
A patch with a new std_valarray.i and test-suite test to test it is welcome for future releases.

PR #2591, opened 2023-05-16
Specifying __cplusplus

smihaila:
`Source/Modules/main.cxx` updates:
1. Added `STRINGIFY_1()` & `STRINGIFY()` macros,
2. Used `STRINGIFY()` macro in two places, for `Preprocessor_define((DOH *) "__cplusplus " STRINGIFY(__cplusplus), 0)`. 

This makes the SWIG's built-in pre-processor expand / translate the value of the `__cplusplus` pre-processor symbol to the expected (long) integer numeric value, as set by the C++ compiler tool-chain that is used in conjunction with `swig.exe -c++` option. For example:
- `199711L` (C++98 compiler conformance level)
- `201402L` (C++14 compiler conformance level)
- `201703L` (C++17 compiler conformance level)
- `202002L` (C++20 compiler conformance level)

Without this change, the value of `__cplusplus` pre-processor symbol is set to ... its name, that is, exactly `__cplusplus`.
Any C++ logic conditional on either `#ifdef __cplusplus` or `#if defined (__cplusplus)`, succeeds without this change, but C++ logic that is conditional on specific _numeric ranges_ i.e via `#if (__cplusplus >= 202002L)` will pose problems.

### DIFF
--- a/Source/Modules/main.cxx
+++ b/Source/Modules/main.cxx
@@ -53,6 +53,12 @@ extern "C" {
  */
 #define EXTRA_WARNINGS "309,403,405,512,321,322"
 
+/* Indirect stringification. Doing two levels allows the argument to be a macro itself.
+ * Example: STRINGIFY(MY_INT_VAL) expands to "10", when we have #define MY_INT_VAL 10
+ */
+#define STRINGIFY_1(x, ...) #x
+#define STRINGIFY(x, ...) STRINGIFY_1(x)
+
 extern "C" {
   extern String *ModuleName;
   extern int ignore_nested_classes;
@@ -498,7 +504,7 @@ static void getoptions(int argc, char *argv[]) {
 	Swig_mark_arg(i);
       } else if (strcmp(argv[i], "-c++") == 0) {
 	CPlusPlus = 1;
-	Preprocessor_define((DOH *) "__cplusplus __cplusplus", 0);
+	Preprocessor_define((DOH *) "__cplusplus " STRINGIFY(__cplusplus), 0);
 	Swig_cparse_cplusplus(1);
 	Swig_mark_arg(i);
       } else if (strcmp(argv[i], "-c++out") == 0) {
@@ -957,7 +963,7 @@ int SWIG_main(int argc, char *argv[], const TargetLanguageModule *tlm) {
 
   // Define the __cplusplus symbol
   if (CPlusPlus)
-    Preprocessor_define((DOH *) "__cplusplus __cplusplus", 0);
+    Preprocessor_define((DOH *) "__cplusplus " STRINGIFY(__cplusplus), 0);
 
   // Parse language dependent options
   lang->main(argc, argv);


ojwb:
This makes SWIG define `__cplusplus` to the value defined by the compiler used to build SWIG, which is pretty much arbitrary since SWIG still aims to be buildable with a C++98 compiler.  It also works against reproducibility - if you rebuild SWIG with a newer C++ compiler (or different options) it should behave the same way.

The question of what SWIG should set for the value of `__cplusplus` seems a reasonable one to raise, but this PR is not the right answer.

For a C++ compiler, this value reflects the language standard which that compiler supports.  SWIG strictly speaking doesn't support all C++98 features (e.g. `valarray` (#1943) was the first I found in a quick trawl of the bug list) but the value for C++98 might be a more helpful default (since SWIG is pretty close to full C++98 support).  Support for C++11 and later has more significant gaps, so defining a higher value by default seems more misleading.

It would be useful for users to be able to control the value of `__cplusplus` if header(s) being wrapped test its value and the newer C++ features they're trying to test for are actually supported by SWIG.

The simple way to do that currently is just to redefine `__cplusplus` before parsing the header(s) to wrap, e.g.

```
// Set __cplusplus to the value for C++17 for foo.h
#undef __cplusplus
#define __cplusplus 201703L
%include "foo.h"
```

It might also be useful to support a `-std=c++NN` command line option (which would just control what `__cplusplus` gets set to rather than actually trying to enable and disable support for features), though having as much as possible in the interface file seems cleaner to me.

Related, SWIG should also predefine the appropriate [feature test macros](https://en.cppreference.com/w/cpp/utility/feature_test). Looking at the list, most are for C++17 and later (I only see one C++14 one) so there probably aren't many we support yet, but e.g. we now support the spaceship operator so should predefine ` __cpp_lib_three_way_comparison`.

smihaila:
Hi,

Yup, I was aware of the `#undef __cplusplus` + `#define __cplusplus 202002L` trick, but I've personally found it cumbersome to use, as one would have to do that before any new `%include` directive, or having to dedicate a .i file which **all** the other .i files **must** include. So, the opportunities for inconsistent values for a pre-processor symbol as essential as `__cplusplus`, can manifest :-(.

Introducing a new `-std=c++NN` option would be the Cadillac solution to this problem.
In the meantime, simply having that symbol set to be literally `__cplusplus` is equally bad - does not solve anything, other then responding as expected to a `#ifdef` / `#if defined (...)`.

It feels so unfortunate that `SWIG` cannot meet my project's expectations, and I get a lot of "friction" with some C++ code base that is heavy into patterns like this:
```c++
class Foo
{
public:
[... some other declarations that are not pre-proc conditioned ...]

#if (__cplusplus < 202002L)
  /** @nodoc Starting from C++20, not required anymore */
  bool operator!=(Foo* pObj) const { return !operator==(pObj); }
  bool operator!=(const Foo& obj) const { return !operator==(obj); }
#else  // C++20
  bool operator==(const Foo& rhs) const { return operator==(static_cast<Foo*>(rhs)); }
#endif
```

Thank you for reviewing this PR.

ojwb:
Your particular example seems to work as-is - `__cplusplus` will get expanded to itself, then treated as `0` by preprocessor expression evaluation, so we take the pre-C++20 path and SWIG sees both operators, which is probably what you want here since I think SWIG doesn't support that part of C++20 yet.

Put your snippet in `test.i` and then `swig -c++ -python -E test.i` output ends:

```
%includefile(maininput="test.i") "test.i" %beginfile
class Foo
{
public:
[... some other declarations that are not pre-proc conditioned ...]


  /** @nodoc Starting from C++20, not required anymore */
  bool operator!=(Foo* pObj) const { return !operator==(pObj); }
  bool operator!=(const Foo& obj) const { return !operator==(obj); }



%endoffile
```

But as I said above, changing `__cplusplus` to be pre-defined to `199711L` seems reasonable.  Defining it in SWIG based on what the compiler used to build SWIG defined it as just isn't though.

smihaila:
Unfortunately, the sample C++ code shown above, which is in case is in the context of a `C++` to `C#` SWIG-ing workflow, will trigger compiler errors with the generated `<SwigInterfaceName>Cs_wrap.h` wrapper/glue code. That's because that auto-generated .h file will then be compiled by msvc++ 2022 edition, which in our project will have the `/std:c++20` explicitly set.

So, the outcome will be this:
- SWIG  will see the two non-equality operators, and generate the `<SwigInterfaceName>`Cs_wrap.h code accordingly,
- But then when the actual compile time comes, via a standard C++ compiler toolchain that enabled C++20 dialect, then those two non-equality operator will not actually be seen in the `Foo` class declaration.

The only two workarounds that I could foresee for this lack-of-consistency problem are:
1. Add 2 x `%ignore` SWIG directives, to explicitly ignore those 2 operator methods from the to-be-swigged C++ class,
or
2. Do the `#undefine __cplusplus` + `#define __cplusplus whateverValue` trick.

And I happen to dislike both of those 2 possible mitigations.
I think I'll probably let this change in our own private code, and possibly with a `swig_vs2022.exe` special file naming for the SWIG resulting build artifacts, so that we can make it SWIG and produce the `<SwigInterfaceName>Cs_wrap.h` files the way we want.

Thanks again.

wsfulton:
As more and more code handles different c++ standards by using the `__cplusplus` macro I think this problem will be more common in the future, so let's find a good solution. 

I also think we should not use the `-std` family of options as I can't see us properly supporting these exact standard values fully. I agree with @ojwb's reasoning in this regard as well as we need to meet a need to modify the value of `__cplusplus` when parsing header files. In addition to the #undef and then redefine trick shown, I propose we support the equivalent gcc/clang command line options to do this, that is `-U`, see https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html. You can wreak havoc with your c++ compiler using:
```
g++ -U__cplusplus -D__cplusplus=209912L ...
```
We can provide the same mad ability for SWIG! The `-U` option suppresses the 'redefined' warning that would have been emitted by gcc:
```
<command-line>: warning: "__cplusplus" redefined
<built-in>: note: this is the location of the previous definition
```
If we implement `-U` we can then get it to suppress SWIG equivalent error:
```
:1: Error: Macro '__cplusplus' redefined,
:1: Error: previous definition of '__cplusplus'.
```
Won't be hard to implement either, just call Preprocessor_undef() in the implementation.

smihaila:
Yes, a new `-U<preprocSymbolNameToUndefine>` would be an excellent idea, and very simple to implement.
Using the `-D__cplusplus=someUnsignedLongValue` was the very first thing I tried to get my SWIG <=> standard C++ compiler toolchain interoperability issue fixed (albeit with disappointing result).

Another aspect to add: In the grand scheme of things, tampering with a critical pre-processor symbol such as `__cplusplus` i.e. considering it a mutable "property", is conceptually wrong. It should be considered a read-only "property" in the context of a compiler toolchain. But hey, since SWIG is not technically considered integral part of a standard compiler toolchain, it is probably okay to violate such conceptual rule.

Thank you all for your time.

We can close this PR, unless more people would like to chime in here with more ideas and comments.

ojwb:
> As more and more code handles different c++ standards by using the `__cplusplus` macro

I'd hope more and more code will use the `__cpp_`-prefixed feature test macros instead, since those are much more granular and so should be more accurate than assuming what's supported based on a single version (it's very common for compilers to support parts of newer standards, and there are also cases where `__cplusplus` reports a standard that the compiler doesn't support every last part of - in both cases the `__cpp_`-prefixed macros can accurately tell you what's supported).

While implementing `-U` is easy and probably has other uses, I wonder if for `__cplusplus` we should actually make our default defining condition on it not already being defined, so you could just use `swig -D__cplusplus=...` to specify the value.

ojwb:
I just had a go at implementing `-U`.  It is easy, except there's a wrinkle in that many of the predefined macros aren't set up until after the option processing loop.  So `-U` works on these:

* `SWIG`
* `__STDC__`
* `SWIG_VERSION`
* Anything you manually defined with `-D` earlier on the command line

However it doesn't work on other pre-defined macros, such as:

* `__cplusplus`
* `SWIGPYTHON` etc

That could be fixed (e.g. store a list of the `-D` and `-U` options and apply them after all predefined macros are handled) but it seems like quite a lot of complexity to handle a case that nobody reasonable would use.

It seems to me that the only valid use cases for `-U` are to undefine `__cplusplus` so you can manually specify its value, and undefining something manually defined with `-D` (which can be useful e.g. if you are using SWIG for multiple target languages and have standard SWIG command line options to use including `-DFOO` which you want to override for a particular target language by not defining `FOO`).

The `__cplusplus` case seems better handled if our default defining of it is skipped if it's set manually, so you only need to `-D__cplusplus=...`.  If we do that, we could just say that using `-U` on predefined macros isn't supported at all, and move the few that are predefined before option parsing to just after, so that the result of `-U` and/or `-D` on a predefined macro is consistent.

ojwb:
> That could be fixed (e.g. store a list of the `-D` and `-U` options and apply them after all predefined macros are handled) but it seems like quite a lot of complexity to handle a case that nobody reasonable would use.

I should probably have noted why the more obvious solution of moving the predefined macros to all get set before option parsing doesn't work - it's that a number of them are set conditional on command line options (e.g. `-c++` -> `__cplusplus`, `-python` -> `SWIGPYTHON`).

I've created #2594 which addresses the original situation (you can now specify `-D__cplusplus=202002L`) and also implements `-U`, as well as cleaning up various other things I noticed along the way.

By default SWIG now defines `__cplusplus` to the value for C++98 rather than literally `__cplusplus`.

(Also, `__STDC__` is now defined to `1` rather than an empty value as that's what compilers do.)

wsfulton:
Hi @ojwb, regarding;

> While implementing -U is easy and probably has other uses, I wonder if for __cplusplus we should actually make our default defining condition on it not already being defined, so you could just use swig -D__cplusplus=... to specify the value.

gcc requires predefined macros to be undefined with `-U` before redefining them. I think it important SWIG copies this behaviour for consistency. This will also add consistency to the equivalent code in a .i file:
```swig
#undef __cplusplus

#define __cplusplus 202002L
```
It also provides consistency mixing command line and code to define and redefine macros, such as using -U on the command line and only a `#define` in an interface file. 

It also ensures, this macro define/redefine in a  `%inline` also matches gcc behaviour as the code in %inline gets parsed by both SWIG and C++ compiler.

wsfulton:
I wonder if we should also mirror gcc's handling for redefined macros, gcc issues a warning and SWIG stops with an error. Generally, SWiG issues a warning instead of an error, so SWIG's approach at the moment is a bit od.
```swig
%inline %{
// inconsistent warning/error between gcc and SWIG
#define ABC 1
#define ABC 2
%}
```

ojwb:
> gcc requires predefined macros to be undefined with -U before redefining them.

It doesn't seem to be required to do so, you just get a warning:

```
$ echo __cplusplus|g++ -D__cplusplus=1 -xc++ -E -
# 0 "<stdin>"
# 0 "<built-in>"
# 0 "<command-line>"
<command-line>: warning: "__cplusplus" redefined
<built-in>: note: this is the location of the previous definition
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 0 "<command-line>" 2
# 1 "<stdin>"
1
```

I see your "like a compiler" argument, but by that argument we should support `-std=c++20` to set `__cplusplus` (which you didn't seem keen to do).  Without that it just seems needlessly annoying to have to use `-U__cplusplus -D__cplusplus=202002L` when that's the only way to control the value `__cplusplus` from the command line.  If we make the redefinition error a warning, you can skip the `-U`, but only if you're happy to have a warning which doesn't really help.

It's already a bit annoying to have to remember the numeric YYYYMM value corresponding to the standard version you want, so maybe we should just implement `-std=c++NN`, but clearly document that all it does is set `__cplusplus` to the corresponding value.  Maybe it should also imply `-c++`, though it seems to not imply `-xc++` for GCC and clang (GCC seems to quietly ignore it without `-c++`; clang emits `error: invalid argument '-std=c++11' not allowed with 'C'`).

We could similarly set `__STDC_VERSION__` based on `-std=cNN` (for NN >= 99; it shouldn't be defined for C89/90).  https://codesearch.debian.net/search?q=__STDC_VERSION__&literal=1 shows checks in headers based on the value of that.

wsfulton:
Thanks for implementing this @ojwb. Just one thing missing, could you add the option to the output of `swig -help`, so add the new options into main.cxx and SWIG.html, thanks. A little section in the docs on exactly what `-std=` does in SWIG is needed and how to change the value of `__cplusplus` from one that is not set by the standard ones (-std=c++98, -std=c++11 etc) might be useful.

wsfulton:
Strange that this re-opened pull request does not appear under the swig-4.2 milestone.

PR #2602, opened 2023-05-20
Improve c# ARGC ARGV

erezgeva:
Change SWIGCSHARP_ARGCARGV_PARAM_IDX from string to number.

And
Remove more redundant NULL checks before free()

### DIFF
--- a/Lib/csharp/argcargv.i
+++ b/Lib/csharp/argcargv.i
@@ -2,17 +2,62 @@
  * SWIG library containing argc and argv multi-argument typemaps
  * ------------------------------------------------------------ */
 
-/*
- * Follow main(int argc, char *argv[]) function
- * Note: SWIGCSHARP_ARGCARGV_PARAM_IDX set the 'argc' parameter location.
- * The C# Marshaling use the argc parameter to determine the array size.
- * See: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshalasattribute.sizeparamindex
- */
+%typemap(cstype) (int ARGC, char **ARGV) "string[]"
+%typemap(imtype) (int ARGC, char **ARGV) "global::System.IntPtr"
+%typemap(ctype) (int ARGC, char **ARGV) "void*"
+%typemap(csin) (int ARGC, char **ARGV) "$modulePINVOKE.Swig_strings_array_to_c($csinput.Length, $csinput)"
+%pragma(csharp) imclasscode=%{
+  [global::System.Runtime.InteropServices.DllImport("$module", EntryPoint="C_Swig_strings_array_to_c")]
+  public static extern global::System.IntPtr Swig_strings_array_to_c(int len, [global::System.Runtime.InteropServices.In,global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPArray, ArraySubType=global::System.Runtime.InteropServices.UnmanagedType.LPStr, SizeParamIndex=0)] string[] array);
+%}
+%insert(wrapper) %{
+typedef struct { int len; char* array[1]; } c_csstrings_array;
+static void* C_Swig_strings_array_free(c_csstrings_array *arr) {
+  if (arr != SWIG_NULLPTR) {
+    int i;
+    for(i = 0; i < arr->len; i++) {
+      free(arr->array[i]);
+    }
+    free(arr);
+  }
+  return SWIG_NULLPTR;
+}
+SWIGEXPORT void* SWIGSTDCALL C_Swig_strings_array_to_c(int len, void *array) {
+  int i;
+  size_t alen, slen;
+  char *p, **ptr;
+  c_csstrings_array *ret;
+  SWIG_contract_assert(SWIG_NULLPTR, len >= 1, "Strings array can not be empty.");
+  alen = sizeof(c_csstrings_array) + sizeof(char *) * (len - 1);
+  ret = (c_csstrings_array *)malloc(alen);
+  if (ret == SWIG_NULLPTR) {
+    SWIG_CSharpSetPendingException(SWIG_CSharpOutOfMemoryException, "fail to duplicate array.");
+    return SWIG_NULLPTR;
+  }
+  memset(ret, 0, alen);
+  ret->len = len;
+  ptr = (char **)array;
+  for(i = 0; i < len; i++) {
+    slen = strlen(ptr[i]) + 1;
+    SWIG_contract_assert(C_Swig_strings_array_free(ret), slen > 1, "String can not be empty");
+    p = (char*)malloc(slen);
+    if (p == SWIG_NULLPTR) {
+      SWIG_CSharpSetPendingException(SWIG_CSharpOutOfMemoryException, "fail to alloc a string.");
+      return C_Swig_strings_array_free(ret);
+    }
+    memcpy(p, ptr[i], slen);
+    ret->array[i] = p;
+  }
+  return ret;
+}
+%}
 
-#ifndef SWIGCSHARP_ARGCARGV_PARAM_IDX
-#define SWIGCSHARP_ARGCARGV_PARAM_IDX "0"
-#endif
+%typemap(in, canthrow=1) (int ARGC, char **ARGV) {
+  c_csstrings_array *arr = (c_csstrings_array*)$input;
+  if (arr != SWIG_NULLPTR) {
+    $1 = ($1_ltype)arr->len;
+    $2 = ($2_ltype)arr->array;
+  }
+}
 
-%typemap(csin) char **ARGV "$csinput"
-%typemap(cstype) char **ARGV "string[]"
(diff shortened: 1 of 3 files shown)


erezgeva:
Follow @ojwb suggestion.

Lib/csharp/argcargv.i
```diff
@@ -10,9 +10,17 @@
...
+#define SWIGCSHARP_ARGCARGV_PARAM_IDX 0
 #endif SWIGCSHARP_ARGCARGV_PARAM_IDX
+/* This typemap can be defined once
+ * In case you need multiple `argc` indexes,
+ * you can define more typemaps in your project, simply copy these typemaps,
+ * name them as 'ARGV<n>' and use 'SizeParamIndex=n'.
+ */
+
+#define stringify(s) #s
+#define stringifyVal(a) stringify(a)
```
wsfulton, line 22: What's the purpose of introducing these macros? The naming convention pollutes the global namespace!

ojwb:
This still seems unsatisfactory - the way you apply these typemaps for SWIG/C# and wrapped API in C# don't match the other languages, applying it multiple times requires manually duplicating the typemaps, and you get an extra constant wrapped thanks to that `#define`:

```
./argcargvtest/argcargvtest.cs:  public static readonly string SWIGCSHARP_ARGCARGV_PARAM_IDX = argcargvtestPINVOKE.SWIGCSHARP_ARGCARGV_PARAM_IDX_get();
```

I think what we're ideally after is that calling it from C# looks like:

```
    string[] largs = {"hi", "hola", "hello"};
    if (argcargvtest.mainc(largs) != 3)
        throw new Exception("bad main typemap");

    string[] targs = {"hi", "hola"};
    if (!argcargvtest.mainv(targs, 1).Equals("hola"))
        throw new Exception("bad main typemap");
```

Exactly how that's best plumbed through isn't clear to me, but as an example of how it could work in `argcargvtest.cs` the generated forwarding functions could look like this, with the current PINVOKE code (but ideally with the `SizeParamIndex` value automatically substituted - perhaps the `imcode` typemap could support `$argnum` like some other typemaps already do):

```
  public static int mainc(string[] argv) {
    int ret = argcargvtestPINVOKE.mainc((uint)argv.Length, argv);
    return ret;
  }

  public static string mainv(string[] argv, int idx) {
    string ret = argcargvtestPINVOKE.mainv((uint)argv.Length, argv, idx);
    return ret;
  }
```

@wsfulton Can you see a way to make this work?

Lib/csharp/argcargv.i
ojwb: They're used to stringize a numeric macro value (it's a classic C preprocessor trick for that - if you just apply `#X` to parameter `X` you get `"X"`).

But I really think we want to avoid needing `SWIGCSHARP_ARGCARGV_PARAM_IDX` (or if it's really needed, to have it supplied in a different way such that it doesn't also get wrapped, and ideally so its value can vary between applications without having to copy and modify these typemaps).

ojwb:
For now I've cherry-picked the first commit as c4bb1e6f0d88091070248e760fa7d619fb0467a0 - thanks for that.

Lib/csharp/argcargv.i
erezgeva: They convert the `SWIGCSHARP_ARGCARGV_PARAM_IDX` from number to string.
The pre-processing is done in 2 steps:
`stringifyVal(SWIGCSHARP_ARGCARGV_PARAM_IDX)` -> 
`stringify(0)` -> `"0"`

I am open for betters options.
We do no need the macros for any other purpose.

erezgeva:
> This still seems unsatisfactory - the way you apply these typemaps for SWIG/C# and wrapped API in C# don't match the other languages, applying it multiple times requires manually duplicating the typemaps, and you get an extra constant wrapped thanks to that `#define`:
> 
> ```
> ./argcargvtest/argcargvtest.cs:  public static readonly string SWIGCSHARP_ARGCARGV_PARAM_IDX = argcargvtestPINVOKE.SWIGCSHARP_ARGCARGV_PARAM_IDX_get();
> ```
> 
> I think what we're ideally after is that calling it from C# looks like:
> 
> ```
>     string[] largs = {"hi", "hola", "hello"};
>     if (argcargvtest.mainc(largs) != 3)
>         throw new Exception("bad main typemap");
> 
>     string[] targs = {"hi", "hola"};
>     if (!argcargvtest.mainv(targs, 1).Equals("hola"))
>         throw new Exception("bad main typemap");
> ```
> 
> Exactly how that's best plumbed through isn't clear to me, but as an example of how it could work in `argcargvtest.cs` the generated forwarding functions could look like this, with the current PINVOKE code (but ideally with the `SizeParamIndex` value automatically substituted - perhaps the `imcode` typemap could support `$argnum` like some other typemaps already do):
> 
> ```
>   public static int mainc(string[] argv) {
>     int ret = argcargvtestPINVOKE.mainc((uint)argv.Length, argv);
>     return ret;
>   }
> 
>   public static string mainv(string[] argv, int idx) {
>     string ret = argcargvtestPINVOKE.mainv((uint)argv.Length, argv, idx);
>     return ret;
>   }
> ```
> 
> @wsfulton Can you see a way to make this work?

I did tried this way first. And saw I can not pull it out.
But perhaps, I miss it.
I am open for suggestions.

ojwb:
I wonder if we should back out the C# `argcargv.i` support for now.  I really think we should aim to provide the equivalent interface that we do for other target languages, and it seems that's going to need more work (and perhaps some additional typemap-related features).  Meanwhile we don't really want to make a release with the current C# `argcargv.i` support as then if we later manage to sort this out we'd have to break compatibility for anyone who'd started using it.

erezgeva:
> I wonder if we should back out the C# `argcargv.i` support for now. I really think we should aim to provide the equivalent interface that we do for other target languages, and it seems that's going to need more work (and perhaps some additional typemap-related features). Meanwhile we don't really want to make a release with the current C# `argcargv.i` support as then if we later manage to sort this out we'd have to break compatibility for anyone who'd started using it.

I gave an idea. I will try something similar to go.

erezgeva:
I send the array with the array length to a generic native function that convert the string array to a C `c_csstrings_array`.
Then I only need to pass the pointer to the `c_csstrings_array` to the wrapped function.

So we have the array length always on index 0 for C# marshaling.
And no limitation on the user.
The cost is reasonable, copying all the strings.
As I assume the C# remove the marshaled memory once we are back from native to C#.

erezgeva:
AppVeyor seems stuck as usual.
And I do not have windows to test the improved C# ARGC ARGV.
I'll try later.

erezgeva:
I see this warning
argcargvtest_wrap.cxx(417): warning C4312: 'type cast': conversion from 'unsigned long' to 'c_csstrings_array *' of greater size

It happens since C# `global::System.IntPtr` is convert to `unsigned long`.
Do you know why?

I mean `global::System.IntPtr` make sense.
https://learn.microsoft.com/en-us/cpp/cppcx/platform-intptr-value-class

But I would expect it to translate to `void*` or `int*`.
But it is out of scope of the patch.

erezgeva:
> 

Forgot the `%typemap(ctype)`

ojwb:
> AppVeyor seems stuck as usual.

It's not stuck now at least.

I think it was probably just backlogged.  It runs 15 jobs, but serially so it takes much longer to complete than GHA's 100+ jobs (GHA allows 20 parallel jobs for the free runners) so frequently ends up several PRs/pushes behind.  I sometimes manually cancelled runs that are not useful if they're delaying a run I want to see the results of, but that's not ideal as the cancelled runs show as failed in the CI results.

erezgeva:
> > AppVeyor seems stuck as usual.
> 
> It's not stuck now at least.
> 
> I think it was probably just backlogged. It runs 15 jobs, but serially so it takes much longer to complete than GHA's 100+ jobs (GHA allows 20 parallel jobs for the free runners) so frequently ends up several PRs/pushes behind. I sometimes manually cancelled runs that are not useful if they're delaying a run I want to see the results of, but that's not ideal as the cancelled runs show as failed in the CI results.

Make sense.
I do my best, to wait for current running appveyor to finish, before I push a new one. And skip it on debugging (unless I use C#).
So I have only 1 appveyor job running.

erezgeva:
@ojwb , @wsfulton 
The improved  C# typemap work with the same syntax like the rest of languages.

ojwb:
Wonderful!

I renamed the helper stuff to use prefixes on the symbols which are more consistent with existing helpers, and reduce the risk of collisions with the API being wrapped.

I also put the helpers in a fragment which seems neater.  Now they only appear in the wrapper if actually used (just including `argcargv.i` but not using it won't emit them, though that was hardly a big issue).

ojwb:
You really want wsfulton for C# stuff, but I think he'd want changing something like the encoding of strings in the wrapped API to be done in a coordinated way across all the typemaps for strings, not just changed for `argcargv.i`.

#2357 is probably a better place to discuss this rather than this closed PR.

erezgeva:
> You really want wsfulton for C# stuff, but I think he'd want changing something like the encoding of strings in the wrapped API to be done in a coordinated way across all the typemaps for strings, not just changed for `argcargv.i`.
> 
> #2357 is probably a better place to discuss this rather than this closed PR.

Already mention both the argcargv and using `UnmanagedType.LPUTF8Str`.
I will issue a new PR for that.
Only I only use ASCII for my application.
As they are low level cli and libraries, which can stick to English and ASCII.

PR #2603, opened 2023-05-22
[D] ARGC ARGV: D empty array translate to null in C.

erezgeva:
It seems D pass array in a very simple way.
Empty arrays are null.
<del>And all arrays have the `str` properties, regardless of their type.
I do not see any point to do more checks on the matter.</del>
D compiler ensure we pass string array.
No further check is needed in `argcargv.i`.

Unharm side effect:
`mainc(null) == 0` pass.
But  `int[] empty_ints; mainc(empty_ints) ;` fail compilation.

Just pass to C.
The only exception is on memory allocation, obviously.

### DIFF
--- a/Examples/test-suite/d/argcargvtest_runme.2.d
+++ b/Examples/test-suite/d/argcargvtest_runme.2.d
@@ -18,9 +18,9 @@ void main() {
 
   initializeApp(largs);
 
-  // An empty array doesn't seem to be valid in D so we can't test that here.
-  // string[] empty_args;
-  // enforce(mainc(empty_args) == 0, "calling mainc failed");
+  // Check that an empty array works.
+  string[] empty_args;
+  enforce(mainc(empty_args) == 0, "calling mainc failed");
 
   // Check that empty strings are handled.
   auto empty_string = ["hello", "", "world"];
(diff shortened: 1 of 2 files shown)


ojwb:
Oh!  I worked up a fix for this too yesterday (as I mentioned I would in https://github.com/swig/swig/pull/2585#issuecomment-1556539103).  I just didn't get round to actually pushing it.  Mine is rather simpler, though perhaps I missed something.  I'll compare them in a bit.

ojwb:
> As long as the idea that empty array is null :-)

Yeah, that seems a reasonable behaviour.  I read up a little and actually an empty array's `.array` can be `NULL` or not depending how it's created, so I added tests that cover both cases.

I went with my simpler version which just uses the standard code path for the empty case.  We'll only access `.array` if `.len` says there are elements, in which case `.array` must be set.

erezgeva:
Looks cool :-)

You can also create empty array with `"".split(":")` from `std.array`.

ojwb:
> You can also create empty array with `"".split(":")` from `std.array`.

OK, but adding `if (len == 0) printf("%p\n", $input.array);` to the d typemap and trying that it gives a `NULL` `.array` so probably doesn't add anything to the test coverage.

PR #2608, opened 2023-05-23
Add 'char_binary' test to Ruby, PHP, Octave

erezgeva:


erezgeva:
@ojwb
These testing are ready for merge.

ojwb:
I don't think I'm the right person to review most of this.

ojwb:
I had another look, and the new runmes don't seem entirely equivalent - the octave and php ones don't seem to test this part:

```
swig_assert(t.strlen(pc) == 4, binding, 'bad multi-arg typemap')
swig_assert(t.ustrlen(pc) == 4, binding, 'bad multi-arg typemap')
```

They also all seem to lack an equivalent to this part from the existing python one:

```
cvar.var_pchar = pc
if cvar.var_pchar != "hola":
    raise RuntimeError("bad pointer case {}".format(cvar.var_pchar))

cvar.var_namet = pc
# if cvar.var_namet != "hola\0":
if cvar.var_namet != "hola":
    raise RuntimeError("bad pointer case")
```

(Not sure what the commented out alternative test there is about - the other languages which test this part all seem to assume the answer won't have a trailing NUL.  It seems to have been there since the runme was first added.)

Are these missing parts testing features which are not working for these languages?

erezgeva:
> sure what the commented out alternative test there is about - the other languages which test this part all seem to assume the answer won't have a trailing NUL. It seems to have been there since the runme was first added.)
> 
> Are these missing parts testing features which are not working for these languages?

These parts did not work.
We can add comments on non working parts.
I follow other languages runme which omit failing parts without comments.
Do you want me to add comments? Or perhaps you should?

ojwb:
I think it's better to include the parts that don't currently work but comment them out, with a note as to why if you have some idea - otherwise it's much harder to know what's not working, so we're less likely to fill in the gaps and make the testcases equivalent across languages.  That's certainly not universally done currently, though I think many of the omissions are probably due to adding test coverage for a new feature or a regression test by extending an existing runme, but only for one language.  That's not entirely ideal in terms of having equivalent testcases across target languages, but at least it's generally not hiding a bug/missing feature for a particular target language.

> Seems that Python string can not contain null.

```
$ python
Python 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0] on linux
>>> "abc\0def"
'abc\x00def'
>>> 
```

I think it's just that the typemaps for `char*` treat NUL as a string terminator (which seems reasonable as that's what you'd typically expect for a `char*`, and that's what happens for other languages which test this).  So I think that commented out line must just be a left over from the original development of this testcase - I'll remove it.

erezgeva:
Looking on python test.
And I understand why it had
`cvar.var_pchar != "hola0":`

This string is suppose to be length 5 with null as part of the string.
Look on the definition of `pc`.
This test is simply wrong.
As `pc` behave as `char *` and **NOT** "STRING LENGTH" that this test suppose to test.
In other languages. the `char *` is a type the languages can not understand unless you use the proper typemap, which is **NOT** "STRING LENGTH".

I think this part should be removed.

ojwb:
I didn't write this testcase, and sadly (like so many of our testcases) it lacks many comments as to what it's trying to test, but I think the idea is that it is testing is that you can build a value in a buffer creating using the `%array_functions` wrapper and then assign it to a `char*` variable wrapped using the default typemaps.

There is (as you say) no length information transferred, so treating the zero byte we manually insert into `pc` as a terminating nul is the most sensible expected behaviour.  The whole situation is certainly a corner case, and unlikely to be something used exactly like this in a real SWIG interface file, but it serves to test the various pieces too.

Issue #2597, opened 2023-05-18
Java (char *STRING, size_t LENGTH)


erezgeva:
Hi,
@wsfulton

Looking on [java.swg](https://github.com/swig/swig/blob/master/Lib/java/java.swg)

And on their 2 tests.
[char_binary.i](https://github.com/swig/swig/blob/master/Examples/test-suite/char_binary.i)
[director_binary_string.i](https://github.com/swig/swig/blob/master/Examples/test-suite/director_binary_string.i)

The typemap is called **'STRNG'** **NOT** bytes .

We can add another  typemap for byte memory block.
Named ` (void *BYTES, size_t LENGTH)`

And fix java to use string for string typemap.

I want to explain a bit
C can have
- Null terminated string, which usually used with `char*`
- Sized string with `chat*` and `size_t length`
- bytes with `void*` or `byte*` or `ubytes*` or `uint8_t*` and a length

Many of SWIG target languages, including Java.
Use a string type that comes with length.
So it is a good practice to support the `STRING` type map with length, in cases we do not wish or can not depends on null termination of strings.
For example, we may use a string like:
`"first sentence\0second sentence\0last sentence\0\0"` 

On the other hand bytes make sense too.
But with proper name!

erezgeva:
JNI uses a modified UTF8
https://stackoverflow.com/questions/32205446/getting-true-utf-8-characters-in-java-jni

The character `0`, is been escaped. 
So we can use a '0' termination instead of length.

Any how, JNI challenges should not break basic logic.
Lead to confusing tests, or misleading term as "binary strings".

ojwb:
> Java JNI support UTF8 with 'GetStringUTFChars' .

Yes, but that's for converting data from Java to UTF8.  A `(char *STRING, size_t LENGTH)` typemap needs to convert in the other direction.

> JNI uses a modified UTF8

JNI does, but C/C++ generally doesn't so embedded zero bytes ideally should be handled.  Currently that seems to be largely ignored by SWIG (the same issue affects SWIG/Tcl too as Tcl also encodes nul as an overlong UTF-8 sequence).

> Any how, JNI challenges should not break basic logic.
> Lead to confusing tests, or misleading term as "binary strings".

I think you're misunderstanding the point I was trying to make there, which was really just that it's not as easy as you'd hope to handle converting a string specified by pointer and length to pass to Java.

erezgeva:
[...]

PR #2609, opened 2023-05-23
Fix Java string length typemap.

erezgeva:
Follow #2597.
Fix Java `char *STRING, size_t LENGTH`.
The current typemap, make sense as shared memory block,
 between C/C++ and wrapped language.
i.e. `void *BYTES, size_t LENGTH`.

But **NOT** for strings, as string require mapping, it does not make sense to share them with same memory between C/C++ and wrapped language.
Fix  `director_binary_string` to skip shared memory.

The `director_binary_string` testing uses callback from C++ to Java.
D lang and go, can call external functions from C/C++.
And probably C#
Go support `extern "C"`, for example see `Lib/go/argcargv.i`.
D support both `extern "C"` and extern "C++"`.
As for C# [FAQ: 2.2 How do I call .NET assembly from native Visual C++?](https://social.msdn.microsoft.com/Forums/en-US/adce905c-2bd3-4276-be53-c644dd928dfa/faq-22-how-do-i-call-net-assembly-from-native-visual-c?forum=vcgeneral)


erezgeva:
I see example for C calliing C# here:
[reverse pinvoke](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke#invoking-managed-code-from-unmanaged-code)

erezgeva:
I see `cdata.i` uses (STRING, size_t LENGTH) for data.
It goes deeper than I though.
Perhaps Java uses `bytes` for data and strings.
But than `go` uses `String` for both .
This is inconsistent.

I though we like typemap behave as similar as possible!

ojwb:
> I see `cdata.i` uses (STRING, size_t LENGTH) for data. It goes deeper than I though. Perhaps Java uses `bytes` for data and strings. But than `go` and `python` uses `String` for both . This is inconsistent.

I think some inconsistencies are going to be inevitable because different programming languages have different approaches to what a string is.  In some it's conceptually a container of Unicode characters, in others it's just bytes and any meaning to those is mostly just imposed by the programmer, in others they can have multiple representations, etc.

> I thought we like typemap behave as similar as possible! Only 2 tests in 3 languages. No wonder why so few tests.

Typically a new testcase doesn't get a full set of `*_runme.*` added.  It's not really realistic to expect people to write testcases in 17 languages (sadly some patch submitters are reluctant to even write a testcase in one language and we end up with quite a lot of stalled PRs which are just lacking a corresponding testcase addition).  New target languages aren't expected to implement a full set of `*_runme.*` either (Extending.html says "The test-suite must also include at least twenty wide-ranging runtime tests.  The most mature languages have a few hundred runtime tests.")

Also `cdata.i` has been around over 20 years and we didn't used to have such good test coverage back then.

erezgeva:
> > I see `cdata.i` uses (STRING, size_t LENGTH) for data. It goes deeper than I though. Perhaps Java uses `bytes` for data and strings. But than `go` and `python` uses `String` for both . This is inconsistent.
> 
> I think some inconsistencies are going to be inevitable because different programming languages have different approaches to what a string is. In some it's conceptually a container of Unicode characters, in others it's just bytes and any meaning to those is mostly just imposed by the programmer, in others they can have multiple representations, etc.

Some may. But this is not a different approach, this is an implementation issue.
Most languages today support Unicode, some uses UTF-16, some UTF-8.
Some like Perl provides conversion functions, but avoid from strict definition.
Yes, most languages use dynamic types, which means you do not really need to define the type you use.
But Java with C#, go and D and do have static types, using a defined static type 'byte' for a typename specifically called STRING, seems to me more than mare inconsistency.
And yet I open an issue first and ask specifically before I submit the patch!
As I understand the problem was implementation not by definition.

> 
> > I thought we like typemap behave as similar as possible! Only 2 tests in 3 languages. No wonder why so few tests.
> 
> Typically a new testcase doesn't get a full set of `*_runme.*` added. It's not really realistic to expect people to write testcases in 17 languages (sadly some patch submitters are reluctant to even write a testcase in one language and we end up with quite a lot of stalled PRs which are just lacking a corresponding testcase addition). New target languages aren't expected to implement a full set of `*_runme.*` either (Extending.html says "The test-suite must also include at least twenty wide-ranging runtime tests. The most mature languages have a few hundred runtime tests.")

I did not express any expectation like that.
And yet, I think that a test case that is implemented by so few languages, is like a red flag.
It might be written so long ago and being missed.
But it might be a bad test, no one wants to touch.

> 
> Also `cdata.i` has been around over 20 years and we didn't used to have such good test coverage back then.

I did found a test that did not have any runme.
And wrote runme for almost all languages.
The `li_constraints.i`.
I can recognise a good test.

erezgeva:
Hi @wsfulton,

Can you review or assign someone else.

wsfulton:
@erezgeva, I put some review comments in #2597, sorry, probably should have gone into this ticket instead.

erezgeva:
For next step. the PR is on draft.

wsfulton:
Does not look ready for swig-4.2 which is about to be released, so am removing the swig-4.2 milestone.

erezgeva:
Hi @wsfulton @ojwb,

I am sorry that I did pass it this issue before.
Looking in Lib/typemaps/strings.swg
I find 2 typemaps
`(Char *STRING, size_t LENGTH)` which exclude the null termination.
And `(Char *STRING, size_t SIZE)` which does include the null termination in size.

I find that `(Char *STRING, size_t SIZE)` was only used to define `(void *, size_t)` in Lib/typemaps/cdata.swg.

I think we can remove `(Char *STRING, size_t SIZE)`.
And I think `LENGTH` in `(Char *STRING, size_t LENGTH)`  should include the null termination, as it defy the purpose.

Looking on "12.3.2 Passing binary data" in Doc/Manual/Library.html
I find no mention on the length argument.

What do you think?

erezgeva:
Hi  @wsfulton,

I add director_binary_string_runme.go.
And I notice that the test tries to create Callback object and crashed as the class is abstract.
However the same success in Java, C# and D.

I thought perhaps we should remove this test from all.

What do you think?

Erez

Lib/java/java.swg
```diff
@@ -1421,10 +1421,24 @@ SWIG_PROXY_CONSTRUCTOR(true, true, SWIGTYPE)
 %apply SWIGTYPE & { SWIGTYPE (CLASS::*const&) }
 
 /* String & length */
+%pragma(java) jniclasscode=%{
```
wsfulton, line 1424: This pragma has reappeared again, did you overlook my previous comment:


             Please apply this patch to your branch:
[bytes.patch.txt](https://github.com/swig/swig/files/13362083/bytes.patch.txt)

_Originally posted by @wsfulton in https://github.com/swig/swig/issues/2597#issuecomment-1812004120_

wsfulton: And similar removal of the pragma is needed for C#

erezgeva: > And similar removal of the pragma is needed for C#

The [bytes.patch.txt](https://github.com/swig/swig/files/13362083/bytes.patch.txt) patch  add a '\0' to `li_cdata_cpp_runme.*` and `li_cdata_c_runme.*`.
I did that to all languages that support including Java.

I did not see any patch/remark regarding the `%pragma` in java and C#.
How can I add functions to Java/C# without using `%pragma`?

erezgeva: After changing cdate  back to byte array in  Java, I revert `li_cdata_cpp_runme.java` and `li_cdata_c_runme.java`  back, so I  rebased your patch into my single commit.
The `%pragma` was not duplicate. Simply in C# and Java, the cdata uses byte array while STRING uses string type. So each typemap need a separate C#/Java `%pragma`, there are no duplications.

Erez

wsfulton: Oh dear, I was actually referring to an additional patch that I didn't provide, so I'm not surpised you're confused. It's this one: 
[remove-java-pragma.patch.txt](https://github.com/swig/swig/files/14364706/remove-java-pragma.patch.txt)
I wrote this a while ago, there is info in the commit message, but the idea is to remove the pragma which always generates a Java method which normally is not used. Can you see if you can incorporate this and an equivalent set of changes for C#?

erezgeva: > Oh dear, I was actually referring to an additional patch that I didn't provide, so I'm not surpised you're confused. It's this one: [remove-java-pragma.patch.txt](https://github.com/swig/swig/files/14364706/remove-java-pragma.patch.txt) I wrote this a while ago, there is info in the commit message, but the idea is to remove the pragma which always generates a Java method which normally is not used. Can you see if you can incorporate this and an equivalent set of changes for C#?

I'll try.
If I fail, or have more issues, I will flag for help.

Thanks
   Erez

erezgeva: In the case of C#.

I think although C# share some similarity with Java.
The C# to native is differ from JNI.

In our case we need to transfer the pointer and the length in a single object.

In `Lib/typemaps/cdata.swg` and `Lib/csharp/csharp.swg`.
The pragma is declaring the prototype of a C function called from C#.
There is a way to allocate C memory in C#, but I think this is a more simple approach.
[Marshalling Classes, Structures, and Unions](https://learn.microsoft.com/en-us/dotnet/framework/interop/marshalling-classes-structures-and-unions)

Anyhow, I think it is does require at least a C# functions in PINVOKE.

Yet, I am open for suggestions
    Erez

Lib/typemaps/cdata.swg
```diff
@@ -1,30 +1,246 @@
...
- * This library file contains macros for manipulating raw C data as strings.
+ * This library file contains macros for manipulating raw C data.
  * ----------------------------------------------------------------------------- */
 
+#ifdef SWIGGO
+#define swig_cdata_nelems_type int
+#else
+#define swig_cdata_nelems_type size_t
+#endif
+
```
wsfulton, line 12: You've made lots of changes to make everything consistent, except for this Go. Why is that?

wsfulton:
I've done a high level review, but need some more info before going into a deeper review.

Could you write an entry into CHANGES.current which details for users what you have done and what the incompatibilities are and how to upgrade (if needed - I havn't figured that out).

Are all the various multi-arg typemaps now documented in Library.html under the "C string handling section"? If not, please add in.

Some other observations with my cursory look, if you can implement these too please.

- Can the appropriate sections in the Java.html docs please be enhanced rather than deleted. If not, please explain.
- The typemaps directory does not traditionally contain language
specific typemaps as this directory is for UTL languages and specifically has
language neutral code. However, you have added language specific typemaps in
cdata.swg. Can you remove these please and put them in the appropriate
language folder. So all the UTL compliant languages should use the typemaps
folder, all the other languages should have their own version in their own
folders.

wsfulton (changes requested):

Lib/typemaps/cdata.swg
erezgeva: Go is very sensitive with types.
`size_t` is defined as `int64`, probably to support large sizes (as `int` might be 32 on 32 bits CPUs).
`Lib/go/go.swg` line 233
`da5e02f81c (Ian Lance Taylor 2010-06-15 19:40:34 +0000 233) %typemap(gotype) size_t, const size_t & %{int64%}`

That will require in Go to use `Cdata(m, int64(7))`
 in `Examples/test-suite/go/li_cdata_runme.go`.

As string are not that long, I though it is an over kill.
My purpose is to make the "testing" consistent as possible.
As the testing represent the user-developer API.

erezgeva: Now Go uses `int64`.
And all cdata uses `size_t`.

erezgeva:
> I've done a high level review, but need some more info before going into a deeper review.

Yap, the PR grow in unplanned way.

> Could you write an entry into CHANGES.current which details for users what you have done and what the incompatibilities are and how to upgrade (if needed - I havn't figured that out).

I wrote something.

> Are all the various multi-arg typemaps now documented in Library.html under the "C string handling section"? If not, please add in.

Cdata did not use STRING-LENGTH typemap, but STRING-SIZE typemap which is not documented.  
I simply renamed it to BYTES-LENGTH and as it now uses `void*`, it no longer need all the different char types.  
That is the reason why I rebased cdata files.  
We can add a mention on the BYTES-LENGTH typemap in the Cdata section.  
We would need to add a new test for the BYTES-LENGTH typemap directly.

> * appropriate sections in the Java.html docs please be enhanced rather than deleted. If not, please explain

The section I removed, is the reason I started this PR.  
Strings are strings and data is passed with byte array.  
We can add two new sections on STRING-LENGTH and cdata-BYTES-LENGTH. 

> * The typemaps directory does not traditionally contain language
specific typemaps as this directory is for UTL languages and specifically has
language neutral code

I would not call it language specific typemep, only different using, of the same typemap.
And in groups, static type languages versa the rest.  
How do you think we should handle it?  
Add a section to each of the static type languages?  
Perhaps describe the Cdata typemap and add a note with reference to the static type languages?

CHANGES.current
```diff
@@ -7,6 +7,22 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
...
+            Add new typemap BYTES-LENGTH to use with cdata.
+            Add BYTES-LENGTH typemap to most languages.
+            Rebase cdata typemap to improve coherent among different languages.
+            Add li_cdata test to most languages.
+            Languages with static type use byte array for cdata.
+            While dynamic type languages retain string.
+            incompatibilities:
+            * STRING-LENGTH: Java uses String
+            * cdata: Go uses byte array and int64 for size
+
```
wsfulton, line 25: This style of docs should be more exact. 'STRING-LENGTH' style is not something used anywhere in SWIG. It needs to be (I think): `(const char *STRING, int LENGTH)` multi-argument typemaps, or maybe? `(const char *STRING, size_t LENGTH)`.

For the incompatibilities, clear, full examples of what users need to change is required.

wsfulton: and what is needed to restore the old functionality if a user needs it.

wsfulton:
How do I review your last set of changes based on my review comments? You've rebased and now I can't see what has changed since I last looked.

erezgeva:
> ow do I review your last set of changes based on my review comments? You've rebased and now I can't see what has changed since I last looked.

Only the last commit.
But you are correct, I should not rebase during review.

Sorry,
    Erez

Lib/mzscheme/cdata.i
```diff
@@ -0,0 +1,10 @@
+/* -----------------------------------------------------------------------------
+ * cdata.i
+ *
+ * SWIG library file containing macros for manipulating raw C data.
+ *
+ * TODO:
+ *  Use generic cdata bu deleting this file and implementing
```
ojwb, line 7: Typo, presumably should be "by"?  Repeated below.

erezgeva: Yap

CHANGES.current
erezgeva: Can you look on my update?
What do you think?

And what about using `Cdata(m, int64(7))` in go?
Do you think we need to use `int` for go?

erezgeva:
Hi @wsfulton,

> > I've done a high level review, but need some more info before going into a deeper review.
> 
> Yap, the PR grow in unplanned way.
> 
> > Could you write an entry into CHANGES.current which details for users what you have done and what the incompatibilities are and how to upgrade (if needed - I havn't figured that out).
> 
> I wrote something.
> 
> > Are all the various multi-arg typemaps now documented in Library.html under the "C string handling section"? If not, please add in.
> 
> Cdata did not use STRING-LENGTH typemap, but STRING-SIZE typemap which is not documented. I simply renamed it to BYTES-LENGTH and as it now uses `void*`, it no longer need all the different char types. That is the reason why I rebased cdata files. We can add a mention on the BYTES-LENGTH typemap in the Cdata section. We would need to add a new test for the BYTES-LENGTH typemap directly.
> 
> > * appropriate sections in the Java.html docs please be enhanced rather than deleted. If not, please explain
> 
> The section I removed, is the reason I started this PR. Strings are strings and data is passed with byte array. We can add two new sections on STRING-LENGTH and cdata-BYTES-LENGTH.
> 

I rewrite the section. So Java uses string, like other languages.

> > * The typemaps directory does not traditionally contain language
> >   specific typemaps as this directory is for UTL languages and specifically has
> >   language neutral code
> 
> I would not call it language specific typemep, only different using, of the same typemap. And in groups, static type languages versa the rest. How do you think we should handle it? Add a section to each of the static type languages? Perhaps describe the Cdata typemap and add a note with reference to the static type languages?

I restore the section, add cdata section to Java, C#, Go, D  and add links from cdata section in Library.

Doc/Manual/Library.html
```diff
@@ -719,19 +719,26 @@ <H3><a name="Library_nn7">12.2.5 cdata.i</a></H3>
...
 binary data structures---for instance, if you needed to extract data from a buffer.
-The target language must support strings with embedded binary data
-in order for this to work.
+Target dynamic typed language must support strings with embedded binary data
+in order for benefit from this module, and should support a 'pack' and 'unpack'
+functions to convert string to array of numbers and via versa.
+Most static typed languages uses string with defined encoding,
+ which means the content of the String will be different from the original raw C data.
+Yet most dynamic typed language do not support arrays of bytes and
+ do not have a defined encoding of strings, which make them proper for holding raw data.
```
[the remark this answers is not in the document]
erezgeva: The section is updated.

wsfulton:
Hi @erezgeva, I'm back onto this patch now. It sure has grown! Looking at C# at the moment, I get build errors in every build, like:
```
./csharp_featuresPINVOKE.cs(197,235): error CS0117: `System.Runtime.InteropServices.UnmanagedType' does not contain a definition for `LPUTF8Str'
```
This is in a lot of extra generated code when not needed in C# by default. I can try make it only generated when needed, but I think you should instead try marshall strings with a length in the same way that plain strings are marshalled. Changing this is going to go down a rabbit hole of trying to make non-standard unicode work across multiple operating systems and it is no fun at all and rather hard. Any reason why you didn't use the same approach as plain char * string marshalling?

erezgeva:
> LPUTF8Str

`LPUTF8Str` was added to [.NET Core 1.1](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype?view=netcore-1.1) and [.NET Framework 4.7](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedtype?view=netframework-4.7) in year 2017.
Which means VC 2015 should have catch it.

Why did the tests in AppVeyor miss it?
AppVeyor does use VC 2013, VC 2015 and VC 2017.

As for using `LPUTF8Str`
As C# is by first a Microsoft product, we should follow [Microsoft Visual Studio Product Lifecycle](https://learn.microsoft.com/en-us/visualstudio/productinfo/vs-servicing) of 10 years.
I will replace `LPUTF8Str` with `LPStr.`
And a note we can replace to `LPUTF8Str` in the future :-)

wsfulton:
Okay LPStr seems okay and everything else works for me now, so that's good. Having reviewed this, the only language specific change I would to see is in C#:
- Group the C# marshalling into a wrapper class in csharphead.swg instead of having disparate functions - model it on SWIGStringHelper, maybe call it SWIGStringWithLengthHelper
- Can you also please try match the line spacing and whitespace spacing as per elsewhere when you move these lines of code

Then lastly, I think you can merge the three Lib/cdata*.swg files into Lib/cdata.swg and just include this one file in all the other language modules. It's unusual to split support for a library file across three files. The order of typemaps etc will be a bit different, but I think it'll be okay. Hopefully you can %include the one Lib/cdata.swg file at the end of all the language specific support files, including the top level Lib/cdata.i.

Lib/typemaps/strings.swg
```diff
@@ -427,70 +426,27 @@
...
-}
-/* old 'int' form */
-%typemap(in) (Char *STRING, int SIZE) = (Char *STRING, size_t SIZE);
-%typemap(freearg) (Char *STRING, int SIZE) = (Char *STRING, size_t SIZE);
 
+%typemap(in) (const Char *STRING, size_t LENGTH) = (Char *STRING, size_t LENGTH);
+%typemap(freearg) (const Char *STRING, size_t LENGTH) = (Char *STRING, size_t LENGTH);
+%typemap(directorin) (const Char *STRING, size_t LENGTH) = (Char *STRING, size_t LENGTH);
 
-/* reverse order versions */
```
wsfulton, line 455: The reverse order versions seem to have been lost. Was that intentional? They'd need putting back in.

erezgeva: OK.
It was not tested, not documented and not implemented in any of the other  languages.
I will put it back and add it to all languages.

wsfulton:
Hi @erezgeva, restoration of string & length typemaps looks fine. It is more than a restoration as it adds these into additional languages. That's handy even if it is expanding the scope of this pull request and new typemaps should have testing. Maybe add those at a later date in a separate pull request. I'm more keen to wrap this up as per [my last comment](https://github.com/swig/swig/pull/2609#issuecomment-2043578218), not sure if you noticed those given you have since requested a review but didn't address them.

erezgeva:
> Lib/cdata*.swg

I somehow missed it, I on your comments.
Though, I am not sure what scope you wish for the length string reverse type map.
Do you want testing and documentation, than I agree, it is better a separate PR.

erezgeva:
> Lib/cdata*.swg

As for the 3 `Lib/typemaps/cdata*.swg`
I share you uneasy.

The problem is:
`typemaps/cdata_struct.swg` define the `SWIGCDATA` structure so it need to be in the start,
We can rename it as `typemaps/cdata_begin.swg`.
While `typemaps/cdata.swg` need to be last, we can rename it to 
`typemaps/cdata_end.swg`.
And `typemaps/cdata_apply.swg` is used partly.
The main `cdata.i` use "%typemap(X) (<new name>) = (<name)"
We can move its content to `typemaps/cdata.swg` and use a macro to defer.

What do you think?

wsfulton:
Okay, try cdata_begin.swg and cdata_end.swg then.

For the length and string reverse order typemaps, please defer to a separate patch request, it should indeed include docs and tests. I actually have a large change to your proposed docs here and it would be much easier for me if you held off making more doc changes until after I have merged this to master, then you can model the doc changes more on what I've done.

erezgeva:
> cdata_begin.swg

Sure. Done.

erezgeva:
@wsfulton,

Anything I can help?

wsfulton:
Thanks, I've merged it in now after making a few small tweaks. See ba9b0a35ab62f0d3cfbb4f7109569d86a00ec53c.

I can see what you are trying to achieve in the html documentation added to C#, D, Go and Java. However, I found it confusing using a mix of C and target language syntax. So instead I've deleted this and enhanced the main documentation in Library.html to cover these 4 statically typed languages.

erezgeva:
> Thanks, I've merged it in now after making a few small tweaks. See [ba9b0a3](https://github.com/swig/swig/commit/ba9b0a35ab62f0d3cfbb4f7109569d86a00ec53c).
> 
> I can see what you are trying to achieve in the html documentation added to C#, D, Go and Java. However, I found it confusing using a mix of C and target language syntax. So instead I've deleted this and enhanced the main documentation in Library.html to cover these 4 statically typed languages.

Great, thanks.
Also for rewrite the documents.

Erez

PR #2617, opened 2023-05-26
add a batch of JS run tests

mmomtchev:


ojwb:
At least some of the CI failures in other languages seem genuine (e.g. lua seems to fail because of the new `typeof` addition to the keyword renaming test).

Examples/test-suite/javascript/extend_template_runme.js
```diff
@@ -0,0 +1,10 @@
+var extend_template = require("extend_template");
+
+f = new extend_template.Foo_0();
+if (f.test1(37) != 37) {
+    throw new Error;
+}
+
+if (f.test2(42) != 42) {
+    throw new Error;
+}
```
ojwb, line 10: This file and the one below are missing a newline at the end of the last line, which is better avoided as it tends to generate whitespace noise in future changes (since some editors unconditionally add the missing newline for you).  It can also be unhelpful if processing the sources with text processing tools since you can get the final line without a newline attached to the start on a different line.  I fixed it this time in f8f47723a86ce8a8ac442b66eca73f755270ca9b,

PR #2620, opened 2023-05-29
fix and test member pointers in javascript

### LINKED ISSUE #2618: Packed objects seem to be broken in SWIG V8
Create a simple run test for `member_pointer`:

`member_pointer_runme.js`:
```js
var member_pointer = require("member_pointer");
area_pt = member_pointer.areapt();
```

In fact, in `SWIGV8_NewPackedObj`, there is a creation of a V8 object using `v8::Object::New` followed by an access of its internal fields. AFAIK, these objects do not have internal fields and this function does not work at all.

mmomtchev:
Class member pointers were broken in both V8 and JSC:
 - `NewPackedObj` did not properly create the object
 - The data was not copied but referenced a variable allocated on the stack (JSC even deleted it after using it!)

Fixes #2618

### DIFF
--- a/Lib/javascript/v8/javascriptrun.swg
+++ b/Lib/javascript/v8/javascriptrun.swg
@@ -438,6 +438,7 @@ int SWIGV8_ConvertPacked(SWIGV8_VALUE valRef, void *ptr, size_t sz, swig_type_in
 
 SWIGRUNTIME void _wrap_SwigV8PackedData_delete(const v8::WeakCallbackInfo<SwigV8PackedData> &data) {
   SwigV8PackedData *cdata = data.GetParameter();
+  cdata->handle.Reset();
   delete cdata;
 }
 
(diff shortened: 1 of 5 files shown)


Lib/javascript/v8/javascriptrun.swg
```diff
@@ -370,9 +373,15 @@ fail:
...
-  SwigV8PackedData(void *data, size_t size, swig_type_info *type): data(data), size(size), type(type) {};
+  SwigV8PackedData(void *data, size_t size, swig_type_info *type): data(nullptr), size(size), type(type) {
+    this->data = malloc(size);
+    if (this->data != nullptr)
+      memcpy(this->data, data, size);
+  };
 
   ~SwigV8PackedData() {
+    if (this->data)
+      free(this->data);
```
ojwb, line 384: There's no need to NULL-check before calling `free()` - ANSI/ISO C requires `free()` accept `NULL`.

Issue #2625, opened 2023-05-31
std::map for type with no default constructor


christophe-calmejane:
Hi,

currently, this doesn't compile
```swig
%module test

%include <std_map.i>

%inline %{
struct NoDefaultConstructorStruct final
{
	int value{0};
	NoDefaultConstructorStruct(int v) : value(v) {}
};
using NoDefaultConstructorStructMap = std::map<int, NoDefaultConstructorStruct>;
%}

%template(NoDefaultConstructorStructMap) std::map<int, NoDefaultConstructorStruct>;
```

Because there is no default constructor for the struct, the typemaps for std::map generates code in wrap.cxx file using c++11 code that do not support this.
Using c++17 `insert_or_assign` in std::map typemaps solves that issue.

Does it make sense to require c++17 for the generated wrap.cxx file? (in this case I may be able to make a PR fixing the typemap)
Is there a mechanism to select the c++ version based on the user's project c++ version?
Is there a way, on a per project basis, to override the default typemaps for std::map and replace `operator[]` with `insert_or_assign`?

Thanks,
Chris

ojwb:
> Does it make sense to require c++17 for the generated wrap.cxx file?

SWIG aims to generate C++98/C90 code, unless wrapping C/C++ language features that inherently require a newer standards version (e.g. `std::unordered_set` is a C++11 feature).  Even then it's sometimes many versions between a compiler implementing some of a new standard version and implementing all of it so best to limit the newer language features used to those actually required.

I'd think in this case you could use something like this construct:

```
#if __cplusplus >= 201703L
      self->insert_or_assign(key, x);
#elif X_IS_DEFAULT_ASSIGNABLE
      (*self)[key] = x;
#else
      if (self->find(key) != self->end()) self->erase(key);
      self->insert(std::pair(key, x));
#endif
```

`X_IS_DEFAULT_ASSIGNABLE` really needs to be hooked up somehow if this is to go in the default typemaps as the fallback is less efficient when we don't need to use it.  Maybe there's a better fallback version, but I'm not seeing one for C++98.  In C++11 one could gdo:

```
     auto r = self->emplace(key, x);
      if (!r.second) *(r.first) = x;
```

However, it seems there are further compilation errors due to other code so more work would be needed:
```
test_wrap.cxx:4360:29: error: no matching function for call to ‘std::pair<int, NoDefaultConstructorStruct>::pair()’
 4360 |           value_type *vp = (new std::pair<T,U>());
```

[...]

Issue #920, opened 2017-02-26
Remove Racket (was MZScheme) support?


DemiMarie:
MZScheme is now known as Racket, and has (by default) a precise GC.  This requires that all C variables that point to Scheme garbage-collected values must be registered with the GC whenever a GC may occur.

ojwb:
I don't think we've had an active maintainer for SWIG's mzscheme backend for some time, but if you want something fixing, patches with suitable test coverage to give a maintainer without specific knowledge of the language are welcome.  If you're also interested in taking on maintainership, that would be great.

ojwb:
If Racket defaults to precise GC and precise GC requires something that SWIG isn't currently doing, does that mean SWIG-generated Racket bindings have broken behaviour if GC happens?

If so, it seems that's been the case for at least 5 years now...

DemiMarie:
Racket is now based on Chez Scheme, so I doubt the original C API is even present now.

ojwb:
Tests are currently passing in CI using the racket package from Ubuntu 20.04, which seems to be version 7.2+dfsg1-2ubuntu3 (but may just happen not to trigger the precise GC issue).

https://en.wikipedia.org/wiki/Racket_(programming_language) says that racket 8.0 is based on Chez Scheme.  If that means a new and incompatible C API then sadly I suspect that's the nail in the coffin for SWIG's current racket support limping along.

DemiMarie:
I’d be fine with just dropping support entirely.  A missing GC root will typically lead to hard-to-debug memory corruption, and Racket 8.0 would need an entirely new backend based on Racket’s FFI.

ojwb:
At least according to https://en.wikipedia.org/wiki/Racket_features the precise GC has been the default since 2007.

I'm going to convert this ticket to a proposal to demote racket (mostly still known as mzscheme in the SWIG code) to "experimental" status in SWIG 4.1.0 because:

* Since 2007 Racket has defaulted to a precise GC which requires that all C variables that point to Scheme garbage-collected values must be registered with the GC whenever a GC may occur (which SWIG apparently doesn't do).  This will typically lead to hard-to-debug memory corruption.

* The latest release series (Racket 8.0) is based on different Scheme implementation and doesn't support the C API which SWIG uses.

[...]

PR #2631, opened 2023-06-06
Fix for #2625 - Cannot bind std::map for type without a default constructor

christophe-calmejane:
Using c++17 insert_or_assign for std::map when available.


Lib/csharp/std_map.i
```diff
@@ -236,7 +236,11 @@
       }
 
       void setitem(const key_type& key, const mapped_type& x) {
+%#if __cpp_lib_map_try_emplace >= 201411L
```
ojwb, line 239: Testing like this can trigger compiler warnings when the macro isn't defined, e.g.:

```
$ printf '#if __cpp_lib_map_try_emplace >= 201411L\n#endif\n'|g++ -xc++ -std=c++98 -Wundef -c -
<stdin>:1:5: warning: "__cpp_lib_map_try_emplace" is not defined, evaluates to 0 [-Wundef]
```

There isn't an earlier version for `__cpp_lib_map_try_emplace` so I'd suggest just checking if it's defined (i.e. `#ifdef __cpp_lib_map_try_emplace`).

ojwb:
It'd be good to have a testcase ensuring that this change actually fixes the case its intended to which would catch if it doesn't for an existing or new target languages and would help ensure it doesn't get broken by future changes.

If it's a C++17 testcase then it'll only get run with a C++17 or better compiler.

christophe-calmejane:
I thought about that but was not sure how the swig test cases can run a test with the -std=c++17 flag passed to the compiler (it's not enough to have a c++17 compiler, the file must be compiled with the correct flag.

(as for the fix itself, it does solve the non-default-constructible issue as my project now compiles fine :D)

Lib/csharp/std_map.i
christophe-calmejane: Interesting, this code is exactly how cppreference instruct how to test for library features.

I guess you are right, there is no older version of that api so only testing for the macro being defined is enough (we could even test both actually, but I'll leave this for the next version of the api)

ojwb: If the example you're looking at is this one from https://en.cppreference.com/w/cpp/feature_test:

```
#if __cpp_constexpr >= 201304                // Check for a specific version of a feature
#  define CONSTEXPR constexpr
#else
#  define CONSTEXPR inline
#endif
```

then the reason for that is that there are multiple possible different values of `__cpp_constexpr` corresponding to various new `constexpr` features - the first was `200704L` but this is showing testing for a later version.  That's why I particularly noted that there isn't an earlier version of `__cpp_lib_map_try_emplace`.

A more similar example to ours is:

```
#if __cpp_binary_literals                    // Check for the presence of a feature
    unsigned mask1 = 0b11000000;
    unsigned mask2 = 0b00000111;
#else
    unsigned mask1 = 0xC0;
    unsigned mask2 = 0x07;
#endif
```

That also emits warnings with `-Wundef` with GCC and clang though.  I don't really see why you'd prefer `#if` to `'#ifdef` for that one; I can see that checking defined-ness as well as the value could be viewed as needlessly uglier if you control the warning flags in use, but for a context like SWIG's generated code or public library headers you'd want to use something like this to check for a specific version of a feature:

```
#if defined __cpp_constexpr && __cpp_constexpr >= 201304
```

christophe-calmejane: Indeed, but that first exemple would also trigger the `undefined warning` you are talking about, as it would not be defined anyway for c++98 ;)
I've always used a compiler more recent (with latest c++ features) than the code I produce, so I never tested for a compiler that don't support it (as it wouldn't compile my code anyway).

In general, I think the most robust way of handling lib features test is having both tests, even if there isn't (at the moment) more than one version. That way it's ready for a new version and the reader is aware this was tested for the defined version.
eg.
```
#ifdef __cpp_lib_map_try_emplace
#  if __cpp_lib_map_try_emplace >= 201411L
#  endif
#else
#endif
```

But in our case here, I replaced the #if with #ifdef and pushed it. Thanks for noticing this ;)

ojwb:
> I thought about that but was not sure how the swig test cases can run a test with the -std=c++17 flag passed to the compiler (it's not enough to have a c++17 compiler, the file must be compiled with the correct flag.

If you add a testcase to `CPP17_TEST_CASES` then it'll only be tested with a suitable compiler+options.  The testcase should be named with a `cpp17_` prefix (that's just a convention to help humans though).

christophe-calmejane:
Sorry for the delay.

I created a test and checked it's failing without my fix and passing after.
I tested from the `Examples/test-suite/perl` folder with success. But I then tried from the python folder and it completely fails to compile the wrap_cxx file (not even where my fix is). How can I declare that my fix only work for some languages (I know for sure it works for `perl` and `csharp` at least).

Other question, how do I configure the `csharp` test-suite, I cannot find how to run the test-suite for csharp in the documentation (only found how to run a single example).

ojwb:
You can add known-to-fail testcases to `FAILING_C_TESTS` or `FAILING_CPP_TESTS` in `Examples/test-suite/python/Makefile.in` etc (beware the existing setting of it there is conditional on Python 2, so you'll want to `FAILING_CPP_TESTS += ...` after that or something to that effect.

It'd be better if we can work out why it doesn't work for Python though, especially as it sounds like it's an existing problem.

`make check-csharp-test-suite` should run the testsuite for csharp (similarly for any target language).  This requires that configure found the requisite tools, etc.  The CI jobs use `Tools/CI-linux-install.sh` to install dependencies, which will probably tell you the exact packages to install if you're using Debian or a Debian-derived distro as the CI Linux builds use Ubuntu, and should provide some clues otherwise.

christophe-calmejane:
Thanks Olly, I'll take a look at the linux code, I thought it would be easier on a windows computer for C# :)
Maybe someday I'll try to setup a docker image if I find the time.

Regarding failing python test (and maybe other languages that I couldn't test yet), it wouldn't compile for the time being anyway for any language, as this PR is there to fix it. So maybe it's best to first merge the PR for fixed languages, and then fix the remaining ones (except if it's an easy fix ofc), as this is not a regression anyway.

This is the file I intended to push:
```swig
%module cpp17_map_no_default_ctor
// Tests c++17 insert_or_assign for std::map with no default constructor

%include <std_map.i>

%inline %{
struct NoDefaultConstructorStruct final
{
        int value{0};
        NoDefaultConstructorStruct(int v) : value(v) {}
};
using NoDefaultConstructorStructMap = std::map<int, NoDefaultConstructorStruct>;
%}

%template(NoDefaultConstructorStructMap) std::map<int, NoDefaultConstructorStruct>;
```

christophe-calmejane:
I tried all languages except mzscheme (couldn't get it properly detected by configure), here are the langages that passed/failed (I'll disable the test for these, as it's better than master which fails for all):

**PASSED**
- ocaml
- perl5
- r
- lua
- tcl
- d
- c#
- scilab
- go

**FAILED**
- javascript
- java
- python
- ruby
- guile

ojwb:
SWIG doesn't work with recent mzscheme (see #920).

You're missing a few target languages there though - there's also at least: go, octave, php, scilab

ojwb:
The initial compile error with python is from:

```
cpp17_map_no_default_ctor_wrap.cxx:4360:29: error: no matching function for call to ‘std::pair<int, NoDefaultConstructorStruct>::pair()’
 4360 |           value_type *vp = (new std::pair<T,U>());
```

`std::pair<T,U>` isn't going to be default constructible unless both `T` and `U` are default constructible.

Looks like this code comes from `Lib/python/std_pair.i` (via a long chain of "required from" starting at `In instantiation of ‘static int swig::traits_asptr<std::pair<_T1, _T2> >::get_pair(PyObject*, PyObject*, std::pair<_T1, _T2>**) [with T = int; U = NoDefaultConstructorStruct; PyObject = _object]’`).

christophe-calmejane:
Hi,
is there anything blocking this merge?

ojwb:
> is there anything blocking this merge?

Mostly my available time for hacking on SWIG.

> std::pair not working correctly (outside the scope of this fix)

I can see that this is already an improvement over the existing situation, but this PR as-is doesn't actually fix the original problem in (something like) 5 out of 16 target languages, which doesn't really seem ideal.

Looking at Python this seems to come from this code in `Lib/python/std_pair.i`:

```
          value_type *vp = %new_instance(std::pair<T,U>);
          T *pfirst = &(vp->first);  
          int res1 = swig::asval((PyObject*)first, pfirst);  
          if (!SWIG_IsOK(res1)) {  
            %delete(vp);
            return res1;  
          }  
          U *psecond = &(vp->second);  
          int res2 = swig::asval((PyObject*)second, psecond);  
          if (!SWIG_IsOK(res2)) {  
            %delete(vp);
            return res2;  
          }  
          *val = vp;  
```

We can use `swig::asptr` instead to avoid needing to be able to default construct `T` and `U`, e.g.:

```
diff --git a/Lib/python/std_pair.i b/Lib/python/std_pair.i
index cf463cb8..5c7d880b 100644
--- a/Lib/python/std_pair.i
+++ b/Lib/python/std_pair.i
@@ -65,20 +65,26 @@
 			  std::pair<T,U> **val) 
       {
 	if (val) {
-	  value_type *vp = %new_instance(std::pair<T,U>);
-	  T *pfirst = &(vp->first);
-	  int res1 = swig::asval((PyObject*)first, pfirst);
+	  T *pfirst = 0;
+	  int res1 = swig::asptr((PyObject*)first, &pfirst);
 	  if (!SWIG_IsOK(res1)) {
-	    %delete(vp);
 	    return res1;
 	  }
-	  U *psecond = &(vp->second);
-	  int res2 = swig::asval((PyObject*)second, psecond);
+	  U *psecond = 0;
+	  int res2 = swig::asptr((PyObject*)second, &psecond);
 	  if (!SWIG_IsOK(res2)) {
-	    %delete(vp);
+	    if (SWIG_IsNewObj(res1)) %delete(pfirst);
 	    return res2;
 	  }
-	  *val = vp;
+	  *val = new std::pair<T,U>(*pfirst, *psecond);
+	  if (SWIG_IsNewObj(res1)) {
+	    %delete(pfirst);
+	    res1 = SWIG_DelNewMask(res1);
+	  }
+	  if (SWIG_IsNewObj(res2)) {
+	    %delete(psecond);
+	    res2 = SWIG_DelNewMask(res2);
+	  }
 	  return SWIG_AddNewMask(res1 > res2 ? res1 : res2);
 	} else {
 	  T *pfirst = 0;
```

And that's enough to make your new testcase pass for Python.

I'm not sure in what circumstances my new version actually ends up allocating new T and U objects for `pfirst` and `psecond`, and whether to avoid that it's worthwhile trying to use the existing approach to avoid that when T and U are default constructible.  @wsfulton Thoughts?

Examples/test-suite/cpp17_map_no_default_ctor.i
```diff
@@ -0,0 +1,15 @@
...
+%inline %{
+struct NoDefaultConstructorStruct final
+{
+	int value{0};
+	NoDefaultConstructorStruct(int v) : value(v) {}
+};
+using NoDefaultConstructorStructMap = std::map<int, NoDefaultConstructorStruct>;
+%}
+
+%template(NoDefaultConstructorStructMap) std::map<int, NoDefaultConstructorStruct>;
```
ojwb, line 15: Github unhelpfully doesn't show it, but I noticed looking locally that this new file seems to have DOS line endings.

christophe-calmejane: Very interesting as all my editors are configured to use LF only :)
I don't remember how I created this file, I'm pretty sure it was a copy of another existing one. I'll check this and commit a fix.

Regarding python, as I have almost no knowledge of it, what do you want me to do? Should I just push a new commit with your suggested change to std::pair, or do you want to do it separately as it affects pair and not map?

ojwb: It affects `std::map` too because `Lib/python/std_map.i` uses `std::pair`.  However I'm not sure what the full consequences of the patch I showed above are so I'm hoping wsfulton will get a chance to comment.

Examples/test-suite/common.mk
```diff
@@ -662,6 +662,7 @@ CPP17_TEST_CASES += \
 	cpp17_nspace_nested_namespaces \
 	cpp17_string_view \
 	cpp17_u8_char_literals \
+	cpp17_map_no_default_ctor \
```
wsfulton, line 665: Not in alphabetical order.

wsfulton:
I've added the swig-4.2 milestone so this doesn't get overlooked. @ojwb, your solution might work and probably can be expanded to the other problematic languages, I need to look. Unfortunately, it's not a quick analysis.

christophe-calmejane:
@wsfulton Maybe we can create a new ticket specifically for that (if it takes too long) since it's currently failing with or without this PR (so this one doesn't get blocked)? I guess it mainly depends on your available time :)

wsfulton:
@ojwb, I looked at your std_pair.i patch which seemed okay. However the test test-suite fails. The inout.i testcase for:
```
  inline void AddOne1p(std::pair<double, double>* p) {
```
because the new code is looking for a `double*` when a number is actually passed in. So the patch will only work for non-primitive types. Not entirely sure what the solution should be. Will give it some thought.

wsfulton:
I'm not really sure how to fix this in the UTL used by Python and Ruby without spending a lot more time on this. Maybe some sort of redesign is needed using more type traits and std::is_default_constructible.

Issue #2612, opened 2023-05-25
Making our parser easier to extend and maintain


ojwb:
I've extended the parser recently to support various newer C++ features, and have come to the conclusion that there are some things we could do to make it easier to make such extensions and also easier to maintain.

My key proposal is essentially that we require using bison to generate it, require bison >= 3.5, and not be too afraid to raise that requirement if some newer bison version provides a feature which would make the parser easier to extend and/or maintain.

We distribute the generated `parser.c` so users building from a SWIG source distribution don't need to install bison(/yacc) so the version required really just needs to be one that's reliably available on development machines of people working on SWIG itself and in CI.  Looking at the bison versions in Ubuntu LTS, focal/20.04 (oldest image GHA now has) is bison 3.5.1.  Debian stable (buster) has 3.7.5.

I'm unclear if we currently aim to support non-bison yacc implementations.  We use `%expect` which is a bison extension so a pure-POSIX yacc wouldn't work.  FWIW, I quickly tried `byacc` which implements a few bison extensions including `%expect` and got a parser which appeared to work, but while this is intellectually interesting the real benefits from it seem minimal.  I'd much rather have an easier to extend and maintain parser.

Here are some examples of things we can make use of with my proposed requirement:

* bison supports named references, which would make many of our rules a lot more readable - in rules like this I find I have to repeatedly count across the rule to work out what say `$6` is, and that makes it much harder to read and understand a rule:
  ```
                  | FEATURE LPAREN idstring COMMA stringnum featattr RPAREN declarator cpp_const SEMI {
                    String *val = Len($5) ? $5 : 0;
                    new_feature($3, val, $6, $8.id, $8.type, $8.parms, $9.qualifier);
                    $$ = 0;
                    scanner_clear_rename();
                  }
  ```
  with named references that could be:
  ```
                  | FEATURE LPAREN idstring COMMA stringnum featattr RPAREN declarator cpp_const SEMI {
                    String *val = Len($stringnum) ? $stringnum : 0;
                    new_feature($idstring, val, $featattr, $declarator.id, $declarator.type, $declarator.parms, $cpp_const.qualifier);
                    $$ = 0;
                    scanner_clear_rename();
                  }
  ```
[...]

PR #2633, opened 2023-06-09
Require bison

ojwb:
First batch of changes for #2612.


ojwb:
Some of the appveyor builds are trying to use a really old bison - it looks like this is just whatever happens to be installed on the older images and comes first in a `PATH` search.  I've psuhed a temporary commit to do a `find` and report any other bison executables that might be a better bet.

ojwb:
All merged now.

PR #2635, opened 2023-06-12
Remove long deprecated features

ojwb:
These features were all deprecated in 1.3.26 (October 9, 2005) or before (many long before), so all more than 17 years and 3 new major versions ago which seems more than enough time for users to have stopped using them, especially as most emit a deprecation warning if used.


ojwb:
Maybe we should still document some of these in the manual as removed features (some never really had any documentation anyway), but I'm inclined to think they were all deprecated so long ago that realistically nobody is going to have an interface file using them and want to look in the manual to find out what they did, so trying to keep documentation for them around is just cruft that makes it harder to read the manual and find documentation as to what you actually should be using.

There's also some contorted handling of renames in the parser (`yyrename`, etc) which hopefully can be simplified once we no longer need to support `%name` with its "rename the next thing" semantics, but that's going to need some careful untangling and I've not attempted to change that in this patch.

Examples/test-suite/errors/pp_deprecated.i
```diff
@@ -1,9 +0,0 @@
-%module xxx
-
-
-%extern ext;
-
-#warning Print this warning
-
-#error This is an error
```
wsfulton, line 8: This seemed to be the only test of #error, so should be added in somewhere else if the entire testcase is removed.

wsfulton (commented):
> Maybe we should still document some of these in the manual as removed features

I would just list these in the CHANGES.current entry as removed for the directives that are mentioned in the html files.

ojwb:
> I would just list these in the CHANGES.current entry as removed for the directives that are mentioned in the html files.

Oh yes, that's a good approach.

Examples/test-suite/errors/pp_deprecated.i
ojwb: Good spot, I'll add one.

(I did investigate and conclude that there wasn't an alternative deprecated feature which could be used for the case this is nominally testing - it needs one for which the warning is emitted at preprocessing time, and none of the remaining ones are.)

ojwb: Done: 365a540fdbfe741ca50066f68b24c67ee7f56322

I've included coverage we didn't have before for `#error` and `#warning` in the inactive branch of `#if` (like we already had for unknown preprocessor directives).

PR #2636, opened 2023-06-13
Type scoping fix

### LINKED ISSUE #2476: Wrong type selection
SWIG 4.1.1 selects incorrect type during Instance::get() processing from the sample code below:
```
#pragma once

#include "utils.h"

namespace classes
{
	class Instance
	{
	public:
		typedef utils::Object<int> Object;
		typedef utils::Container<Object> Container;

		Instance() {}
		~Instance() {}
		
		Container get()
		{
			return Container(new Object(5));
		}
	};
}
```
```
sample_wrap.cxx: In function ‘PyObject* _wrap_Instance_get(PyObject*, PyObject*)’:
sample_wrap.cxx:3189:53: error: type/value mismatch at argument 1 in template parameter list for ‘template<class T> class utils::Container’
   SwigValueWrapper< utils::Container< utils::Object > > result;
```
classes::Instance::Object should be used instead of utils::Object.
The rest of the sample (utils.h):
```
#pragma once

namespace utils
{
	template <typename T>
	class Object
	{
	private:
		T value;
	public:
		Object(T param)
		{
			value = param;
		}
		~Object() {}
		
		T get_value()
		{
			return value;
		}
	};
	
	template <typename T>
	class Container
	{
	private:
		T * object;
	public:
		Container(T * param)
		{
			object = param;
		}
		Container(const Container & container) = delete;
		Container(Container && container)
		{
			object = container.object;
			container.object = NULL;
		}
		~Container()
		{
			if (object)
			{
				delete object;
				object = NULL;
			}
		}

		T * get()
		{
			return object;
		}
	};
}
```

ojwb:
Patch from Kevin Smyth in https://sourceforge.net/p/swig/bugs/1063/

Seems to also fix #2476

### DIFF
--- a/Source/Swig/symbol.c
+++ b/Source/Swig/symbol.c
@@ -2037,11 +2037,12 @@ SwigType *Swig_symbol_template_deftype(const SwigType *type, Symtab *tscope) {
 	Append(tprefix, "<(");
 	expandedparms = Swig_symbol_template_defargs(tparms, tnargs, tscope, tsdecl);
 	p = expandedparms;
+	Symtab *tscope_orig = tscope;
 	tscope = tsdecl;
 	while (p) {
 	  SwigType *ptype = Getattr(p, "type");
 	  SwigType *ttr = ptype ? ptype : Getattr(p, "value");
-	  SwigType *ttf = Swig_symbol_type_qualify(ttr, tscope);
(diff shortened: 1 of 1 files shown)


ojwb:
Breaks stuff for Python and I've no real idea about the code this is changing so closing.

PR #2639, opened 2023-06-16
 std::filesystem::path (C++17) <-> pathlib.Path (Python 3.4+) Wrapper

jmarrec:
Supersedes #1999 by @yasamoka . I pulled your changes (kept you as author of course), and update the testing part only so far


Examples/test-suite/python/cpp17_std_filesystem_runme.py
```diff
@@ -0,0 +1,62 @@
...
+# Similarly, each of the input typemaps should know what to do
+# with a std::filesystem::path instance.
+#
+check(pathToStr(pathPtr) == "foo")
+check(pathConstRefToStr(pathPtr) == "foo")
+check(pathPtrToStr(pathPtr) == "foo")
+
+specialPath = pathlib.Path("/家/屋")
+roundTriped = roundTrip(specialPath)
+check(specialPath == roundTriped)
```
jmarrec, line 62: Added a roundtrip test.

Lib/python/std_filesystem.i
```diff
@@ -0,0 +1,85 @@
...
+
+%typemap(in, fragment="common") std::filesystem::path {
+    if (PyUnicode_Check($input)) {
+        const char* s = PyUnicode_AsUTF8($input);
+        $1 = std::filesystem::path(s);
+    }
+    else if (isPathInstance($input)) {
+        const char * s = pathToStr($input);
+        $1 = std::filesystem::path(s);
+    }
```
jmarrec, line 42: https://github.com/swig/swig/pull/1999#discussion_r902562620

> Are you sure that this call will decode UTF-8 on Windows? If I read the specs correct, this is only done for `char8_t`, so only for unsigned char.
> 
> Should there be checks in the test-suite which checks for unicode edge cases instead of just foo?

Lib/python/std_filesystem.i
```diff
@@ -0,0 +1,85 @@
...
+        int res = SWIG_ConvertPtr($input, &argp, $descriptor, $disown |  0 );
+        if (!SWIG_IsOK(res)) {
+            %argument_fail(res, "$type", $symname, $argnum);
+        }
+        $1 = %reinterpret_cast(argp, $ltype);
+    }
+}
+
+%typemap(out, fragment="common") std::filesystem::path {
+    const std::string& s = $1.string();
```
jmarrec, line 74: I think it is actually here that we need to do something different.

Am I allowed to do an an `if constexpr` with type_traits here since this is C++17 only?

```
if constexpr (std::is_same_v< typename fs::path::value_type, wchar_t>) {
        // use generic_wstring() and PyUnicode_FromString
} else {
       // use generic_string() and PyUnicode_FromWideChar or somehing
}
```

The alternative is to use macros and assume WIN32 => wide string, otherwise string

jmarrec: I've actually tested the rountripping of paths with special char on windows and it appears to work.

jmarrec: The mingw WITHPYTHON=python is failing, all others windows CI builds are ok though... https://ci.appveyor.com/project/swig/swig/builds/47323349/job/8ru8l4v46ipumj8y#L1079

jmarrec: going to try to catch it in d29d8947d962224be0b1e1f3fd4a64003909d6f2

jmarrec:
```
checking python testcase cpp17_director_string_view (with run test)
  File "C:\projects\swig\Examples\test-suite\python\cpp17_std_filesystem_runme.py", line 65, in <module>
    assert specialPath == roundTriped, "'{p}' != '{p2}'".format(p=specialPath, p2=roundTriped)
AssertionError: '/\u5bb6/\u5c4b' != '/家/屋'
make[1]: *** [Makefile:133: cpp17_std_filesystem.cpptest] Error 1
```

ojwb: > Am I allowed to do an an `if constexpr` with type_traits here since this is C++17 only?

That seems reasonable as these typemaps require a C++17 compiler for `std::filesystem::path` anyway.

Lib/python/std_filesystem.i
```diff
@@ -0,0 +1,85 @@
+/*
+  filesystem::path
+*/
+%{
+#if __cplusplus >= 201703L
+#include <filesystem>
+#include <Python.h>
+#endif // __cplusplus >= 201703L
+%}
```
ojwb, line 9: This shouldn't need to be conditional now.

Also, `#include <Python.h>` should always get added to the wrapper by `pyruntime.swg` (which is included implicitly) so individual macro files shouldn't need to explicitly include it too.

Lib/python/std_filesystem.i
```diff
@@ -0,0 +1,85 @@
...
+
+%fragment("common", "header") {
```
ojwb, line 11: The fragment name here ("common") seems too generic.  We don't seem to really have a consistent naming convention for fragments containing helper functions, but something like `"SWIG_std_filesystem"` would at least be unlikely to collide with an unrelated fragment in a user interface file.

Lib/python/std_filesystem.i
```diff
@@ -0,0 +1,85 @@
...
+    PyObject * importPathCls() {
```
ojwb, line 12: These helpers should probably all be `SWIGINTERN`, e.g.:

```
SWIGINTERN PyObject * importPathCls() {
```
```

File: Lib/python/std_filesystem.i [hunk 7]
```diff
@@ -0,0 +1,85 @@
...
+        PyObject * cls = importPathCls();
+        bool is_instance =  PyObject_IsInstance(obj, cls);
+        Py_DECREF(cls);
+        return is_instance;
+    }
+
+    const char * pathToStr(PyObject * obj) {
+        PyObject * str_obj = PyObject_Str(obj);     // New reference
+        const char * s = PyUnicode_AsUTF8(str_obj); // This stores the UTF-8 representation buffer within str_obj
+        // Py_DECREF(str_obj);                      // So we don't decref here
```
ojwb, line 29: So does this mean we end up leaking `str_obj`?

Maybe this helper would be better just hand-inlined at each of the two uses, or else split into two variant helpers which respectively construct and return a `std::filesystem::path` and `std::filesystem::path*` and so can do the `Py_DECREF()`.

Examples/test-suite/python/cpp17_std_filesystem_runme.py
```diff
@@ -0,0 +1,67 @@
...
+# Similarly, each of the input typemaps should know what to do
+# with a std::filesystem::path instance.
+#
+check(pathToStr(pathPtr), "foo")
+check(pathConstRefToStr(pathPtr), "foo")
+check(pathPtrToStr(pathPtr), "foo")
+
+specialPath = pathlib.Path("/家/屋")
+roundTriped = roundTrip(specialPath)
+assert specialPath == roundTriped, "'{p}' != '{p2}'".format(p=specialPath, p2=roundTriped)
```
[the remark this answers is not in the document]
ojwb: I'm guessing the `==` isn't checking equality as you're hoping for.  Maybe see what `repr(specialPath)` and `repr(roundTriped)` report?

There's also a typo in the variable name - should be `roundTripped` (double `p`).

Lib/python/std_filesystem.i
jmarrec: Just trying to understand so this sticks for next time... So SWIGINTERN means `static` right? [1] 
So that gives it internal linkage, and could be inlined. And we want static not inline because it's not used outside of this TU. Am I following this correctly?

---

[1] and `__attribute__ ((__unused__))` but let's not care

Examples/test-suite/python/cpp17_std_filesystem_runme.py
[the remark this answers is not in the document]
jmarrec: Corrected the typo, adding more diagnostics to include repr for the variables too in addition to str (which is what the format call is doing) in a1bf4a46c6faefcb6ffdb2c72db9f2d929689e55. Let's say what MinGW reports...

Lib/python/std_filesystem.i
ojwb: From `Lib/swiglabels.swg`:

```
/* internal SWIG method */
#ifndef SWIGINTERN
# define SWIGINTERN static SWIGUNUSED
#endif
```

So it takes care of the adding an `unused` attribute for you (at least for compilers where SWIG knows how to, but it should for ones where `__attribute__ ((__unused__))` works.

The key thing is we don't want these functions to be visible outside the shared object as they aren't needed there.
The load time for a shared object depends on the number of external symbols (as they need relocating).  There's also a small runtime overhead to calling an external symbol even from within the object that defines it since unless you link with special options it's required to be "interposable" (this is how LD_PRELOAD libraries such as libeatmydata work).

Also we'd then have to worry about what happens when more than one SWIG-wrapped module gets loaded, and potentially calls to these functions from one module get routed to the definition in another, which means updating the code of these functions would need extra care to maintain compatibility with how older generated code used them.

I suggested `SWIGINTERN` rather than `SWIGINTERNINLINE` because if we're going to suggest the compiler inline these then I'm unclear why we've gone to the trouble of pulling them out into a helper functions rather than just hand inlining the code into the typemap (which would make avoiding that leak much simpler).

A side issue (sorry if I noted this already): these helpers should really be named such that they're not going to collide with code being wrapped (e.g. consider what happens if the library I'm wrapping has a `pathToStr` function).  We generally use a `SWIG_` prefix, but it's probably better to make that `SWIG_std_filesytem_` here (e.g. `SWIG_std_filesystem_pathToStr`) so we don't have to worry about it potentially colliding with helpers from other SWIG-supplied macros either.

Examples/test-suite/python/cpp17_std_filesystem_runme.py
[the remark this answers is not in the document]
jmarrec: Pfff Appveyor for mingw / cygwin failed to run. https://ci.appveyor.com/project/swig/swig/builds/47427816/job/edx4ht7x0eu0nq56#L8

> fatal: unable to access 'https://github.com/swig/swig.git/': Failed to connect to github.com port 443 after 21065 ms: Couldn't connect to server.


 Triggering a rerun...

Lib/python/std_filesystem.i
jmarrec: Thanks for the explanation, this is helpful.
I have renamed the fragments, I didn't know I needed to also rename the helpers within the fragments themselves.

ojwb:
If you're likely to be triggering many more CI runs, can you temporarily disable the builds for everything except Python as those jobs aren't really doing anything useful here.  You'll also see the useful results sooner that way.

jmarrec:
@ojwb Seems like with the string v wstring constexpr thingy, appveyor is hapy. The only workflow that failed has nothing to do with this PR: https://github.com/swig/swig/actions/runs/5441910076/jobs/9896444051?pr=2639#step:2:61

>   Error: fatal: unable to access 'https://github.com/swig/swig/': The requested URL returned error: 429

So no, not planning on triggering any CI runs, I consider this PR ready, unless you find something else during another review pass.

wsfulton:
Thanks @jmarrec. I've made merged this in after making a few mods, please take a look at  e3e92986ada49ce869a9dbbfc8b9073097ee2458...bd093173afcd54e545352761e9143156bce179f6. 

If you want the `std::filesystem::path const *` typemaps how you had them, then add this into your interface file:
```swig
%apply const std::filesystem::path & {std::filesystem::path const *};
```

PR #2643, opened 2023-06-23
Eliminate T_ERROR

ojwb:
This was only being used in one very specific case where we failed to parse a type string, which seems like it really is just an internal error so try handling it as such.


ojwb:
@wsfulton Thoughts?

I'm not seeing how we could ever get `T_ERROR` as the type code without a broken type string, which seems like it's an internal consistency error to me.  Digging back in the history it looks like it's been like this since before the start of version control.

It does seem weird that we have these two warnings that probably can't actually be triggered.  I wonder if they should actually be hooked up to cases where we don't know the type of the value (something like `$N.val == NULL` maybe).  I tried this example to see what happens and currently it wraps `x`, and seems to assume `foo()` returns `int` (which I suppose is reasonable for C at least):

```
%module x
%constant x = foo();
```

Maybe there's a better example where we don't know the type and it isn't useful to assume it's `int` though.

ojwb:
Poking git history harder, it actually looks like the last place that set `T_ERROR` aside from the type-string parser error case was removed in 2004 by ad39b124c8f4f9faeb6ae6da37dd1de4100dbd6f.

Source/CParse/parser.y
```diff
@@ -2043,66 +2043,50 @@ clear_directive : CLEAR tm_list SEMI {
...
-		     add_symbols($$);
-		     Delete(type);
-		   } else {
-		     Swig_warning(WARN_PARSE_UNSUPPORTED_VALUE,cparse_file,cparse_line,"Unsupported constant value (ignored)\n");
-		     $$ = 0;
-		   }
-
+		 SwigType *type = NewSwigType($4.type);
+		 $$ = new_node("constant");
+		 Setattr($$,"name",$2);
```
wsfulton, line 2048: Suggest whitespace fixes (spaces after the comma) given other whitespace indentation change on this line and others.

wsfulton (changes requested):
I like this change - malformed SwigType are an internal error and should never occur.

PR #2644, opened 2023-06-24
Use non-throwing new for %new_copy etc

ojwb:
This makes the behaviour for C and C++ consistent since now NULL gets returned which gets mapped to Null, None, etc in the target language.

Fixes https://sourceforge.net/p/swig/bugs/1372/

### DIFF
--- a/Lib/swigfragments.swg
+++ b/Lib/swigfragments.swg
@@ -92,3 +92,7 @@
 %fragment("<memory>", "header") %{
 #include <memory>
 %}
+
+%fragment("<new>", "header") %{
+#include <new>
+%}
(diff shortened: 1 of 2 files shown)


wsfulton:
This is a major API breakage for everyone who expects `std::bad_alloc` to be thrown and has special exception handling in place to catch this exception. I've put this kind of code in place in a various places.

Alas, the control and handling of out of memory isn't a good story in the generated wrappers. If there are any changes made in this area, it should be part of a comprehensive documented plan for consistency and filling gaps and options for handling non-throwing and throwing new.

ojwb:
> I've put this kind of code in place in a various places.

It'd be useful to at least document how to do that then.

wsfulton:
The kind of things I meant is using %exception to catch bad_alloc to handle out of memory in an application specific way by catching the exception, logging it and then mapping it to a particular target language exception. Similarly for throws typemaps which also handle exceptions.

PR #2645, opened 2023-06-28
Fix directed method returning class without default ctor

ojwb:


ojwb:
I wrote in the issue:
> In CI, the testcase I added for this also fails for csharp, d and java, but for a different reason - these all seem to default construct a static object of the type to point a reference at.

However the CI logs have expired so I've rebased and force-pushed the branch (I cherry-picked and merged the comment fix commit a while ago).

wsfulton:
Sorry, I added the swig-4.2 milestone because I thought this was ready for review, but with the CI testing showing failures, I got that wrong, so removing.

ojwb:
It'd be nice to resolve and I'll try to have another poke, but indeed doesn't need to be a blocker for 4.2.0.

erezgeva:
Although Lua director works a bit different.
You can add Lua to the list

PR #2648, opened 2023-07-04
Lib/javascript/napi: omit unnecessary [and some leaky] indirections.

dot-asm:


ojwb:
@mmomtchev This looks like it's changes to the NAPI support you recently contributed - please can you review?

dot-asm:
> I don't see any functional changes

It doesn't leak `$jsclassidx`-es in case the module is unloaded (which node.js is allowed to do).

> Maybe just add a comment saying this.

Done.

mmomtchev:
There are a few more leaks in the property descriptors - one because of a deprecated API call and another one because of the manual suppression of property descriptors. The only way that Node.js can unload a previously loaded module is when a `worker_thread` that was the only user quits. There are still a couple of things missing - including locking - before the modules can be freely unloaded and reloaded, I am still waiting for the multithreaded tests to add them.

wsfulton:
@mmomtchev, what's the conclusion for the merge request, should it be merged? And if so, what should the entry into the CHANGES file be?

Lib/javascript/napi/javascriptcode.swg
```diff
@@ -758,12 +758,10 @@ void $jsmangledname_inst::GetMembers(
   // jsnapi_registerclass
   Napi::Function $jsmangledname_ctor = $jsmangledname_inst::GetClass(env);
   $jsparent.Set("$jsname", $jsmangledname_ctor);
-  if (SWIGTYPE_$jsmangledtype->clientdata == nullptr) {
-    SWIGTYPE_$jsmangledtype->clientdata = new size_t($jsclassidx);
-  }
-  Napi::FunctionReference *$jsmangledname_ctor_ref = new Napi::FunctionReference();
-  *$jsmangledname_ctor_ref = Napi::Persistent($jsmangledname_ctor);
-  env.GetInstanceData<EnvInstanceData>()->ctor[$jsclassidx] = $jsmangledname_ctor_ref;
+  // Almost, but not quite, entirely unlike random bias of 42 to disambiguate
```
wsfulton, line 761: Why 42. Surely initialise to something sensible like zero.

Lib/javascript/napi/javascriptruntime.swg
```diff
@@ -31,6 +31,7 @@
 #include <stdlib.h>
 #include <assert.h>
 #include <map>
+#include <vector>
```
wsfulton, line 34: Please don't add STL types as a mandatory dependency for all users. `<map>` should not be in there too! We keep the dependencies imposed on users to an absolute minimum.

Lib/javascript/napi/javascriptcode.swg
mmomtchev: I don't really know. It does solve some of the leaks, but it abuses a pointer type to store a number.

Alas, it does not really solve the module leaking memory when it is unloaded. I am using a deprecated form of the call used to to assign getters and setters because the recommended form does not support CRTP classes. I spent some time trying to find a solution and I couldn't do it. The recommended form expects the getter and the setter function to be passed as template arguments.

I can only guess that this is the reason that the Node.js team has left this deprecated form as it is.

Alas, clean unloading of binary modules in Node.js is difficult and most modules do not support it. It is a rare situation - it requires that the module is loaded by a `worker_thread` and then the `worker_thread` quits. In this case Node.js will unload the shared library. Me, personally, I recommend that all my users load all my binary modules from the main thread before referencing them in a `worker_thread`.

Anyway, I do not think that this is the most pressing issue at the moment - but I do not see any good reason neither to merge it or to not merge it.

PR #2649, opened 2023-07-05
Fix code generation for OCaml 5.0.0

jamesjer:
The Fedora Linux distribution is about to rebuild all OCaml packages with OCaml 5.0.0.  We encountered two problems with swig-generated code:
- The token `value` appears in included C++ headers; changing it to `caml_value_t` leads to compilation errors
- The name `caml_local_roots` has been replaced with an uppercase macro: `CAML_LOCAL_ROOTS`.
This commit addresses the first issue by reversing the `value` to `caml_value_t` replacement.  Instead of trying to replace `value` in all of the OCaml headers, replace `caml_value_t` in the generated code instead.  The fix for the second issue should work for all versions of OCaml.

I am no OCaml expert, but I can report that this patch leads to a successful build of the Fedora swig package with OCaml 5.0.0.

### DIFF
--- a/Lib/ocaml/ocamlrundec.swg
+++ b/Lib/ocaml/ocamlrundec.swg
@@ -14,7 +14,7 @@ SWIGEXT {
 #else
 #define SWIGEXT 
 #endif
-#define value caml_value_t
+#define caml_value_t value
 #define CAML_VALUE caml_value_t
 #define CAML_NAME_SPACE
 #include <caml/alloc.h>
@@ -80,11 +80,15 @@ SWIGEXT {
                                                  /* Also an l-value. */
 #endif
 
+#ifndef CAML_LOCAL_ROOTS
+#define CAML_LOCAL_ROOTS caml_local_roots
+#endif
+
 #ifdef CAMLreturn0
 #undef CAMLreturn0
 #endif
 #define CAMLreturn0 do{ \
-  caml_local_roots = caml__frame; \
+  CAML_LOCAL_ROOTS = caml__frame; \
   return; \
 }while (0)
 
@@ -93,12 +97,12 @@ SWIGEXT {
 #endif
 #define CAMLreturn(result) do{ \
   caml_value_t caml__temp_result = (result); \
-  caml_local_roots = caml__frame; \
+  CAML_LOCAL_ROOTS = caml__frame; \
   return (caml__temp_result); \
 }while(0)
 
 #define CAMLreturn_type(result) do{ \
-  caml_local_roots = caml__frame; \
+  CAML_LOCAL_ROOTS = caml__frame; \
   return result; \
 }while(0)
 


wsfulton:
Hi @jamesjer I'm afraid I broke master overnight which is why half the tests are showing as failed. If you rebase on top of master now, the testing should go green.

However, we lost our ocaml CI testing recently in f030df8786bc3e707bee9079d093368de3d8bc57. Hopefully with your patch here it can be resurrected on the latest Ubuntu.

@ZackerySpytz could you please review the patch and approve?

jamesjer:
No problem.  I have rebased as requested.

ojwb:
> Hopefully with your patch here it can be resurrected on the latest Ubuntu.

I rebased my branch from trying to re-enable the ocaml CI a few months ago then cherry-picked this patch onto it, but it fails (and in different ways with ubuntu 20.04 and 22.04):

https://github.com/swig/swig/actions/runs/5469228302

Running examples and tests locally works a bit better (so perhaps the CI problem is due to an out-of-tree build?  Or maybe I'm using a newer ocaml version, I have 4.13.1-4) but there is an issue related to the `value` change it seems:

```
checking ocaml testcase bools
bools_wrap.c:1725:19: error: ‘bool value(bool*)’ redeclared as different kind of entity
 1725 | bool value(bool* b) {
      |                   ^
In file included from /usr/lib/ocaml/caml/alloc.h:24,
                 from bools_wrap.c:765:
/usr/lib/ocaml/caml/mlvalues.h:60:16: note: previous declaration ‘typedef intnat value’
   60 | typedef intnat value;
      |                ^~~~~
make[2]: *** [../../../Examples/Makefile:979: ocaml_static_cpp] Error 2
make[1]: *** [Makefile:78: bools.cpptest] Error 2
```

`value` is really an unhelpfully generic name to typedef in a public header, but we can at least just change it in this testcase.  I'll push that change shortly.

There's another failure though the cause of which isn't immediately clear to me:

```
checking ocaml testcase cpp11_function_objects
cpp11_function_objects_wrap.c:1841:34: error: expected ‘)’ before ‘self’
 1841 |     SwigDirector_Test(CAML_VALUE self);
      |                      ~           ^~~~
cpp11_function_objects_wrap.c: In function ‘value _wrap_new_Testcpp11_function_objects(value)’:
cpp11_function_objects_wrap.c:1943:48: error: no matching function for call to ‘SwigDirector_Test::SwigDirector_Test(value&)’
 1943 |     result = (Test *)new SwigDirector_Test(arg1);
      |                                                ^
cpp11_function_objects_wrap.c:1838:7: note: candidate: ‘constexpr SwigDirector_Test::SwigDirector_Test(const SwigDirector_Test&)’
 1838 | class SwigDirector_Test : public Test, public Swig::Director {
      |       ^~~~~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:1838:7: note:   no known conversion for argument 1 from ‘value’ {aka ‘long int’} to ‘const SwigDirector_Test&’
cpp11_function_objects_wrap.c: At global scope:
cpp11_function_objects_wrap.c:2116:37: error: expected constructor, destructor, or type conversion before ‘(’ token
 2116 | SwigDirector_Test::SwigDirector_Test(CAML_VALUE self): Test(), Swig::Director(self) {
      |                                     ^
cpp11_function_objects_wrap.c: In member function ‘virtual void SwigDirector_Test::operator()(int, int)’:
cpp11_function_objects_wrap.c:2124:19: error: expected ‘;’ before ‘swig_result’
 2124 |   SWIG_CAMLlocal2(swig_result, args);
      |                   ^~~~~~~~~~~
cpp11_function_objects_wrap.c:790:16: note: in definition of macro ‘SWIG_CAMLlocal2’
  790 |   caml_value_t x = 0, y = 0; \
      |                ^
In file included from cpp11_function_objects_wrap.c:769:
cpp11_function_objects_wrap.c:2124:19: error: ‘swig_result’ was not declared in this scope; did you mean ‘swig_self’?
 2124 |   SWIG_CAMLlocal2(swig_result, args);
      |                   ^~~~~~~~~~~
/usr/lib/ocaml/caml/memory.h:361:36: note: in definition of macro ‘CAMLxparam2’
  361 |     (caml__roots_##x.tables [0] = &x), \
      |                                    ^
cpp11_function_objects_wrap.c:2124:3: note: in expansion of macro ‘SWIG_CAMLlocal2’
 2124 |   SWIG_CAMLlocal2(swig_result, args);
      |   ^~~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:2124:32: error: ‘args’ was not declared in this scope
 2124 |   SWIG_CAMLlocal2(swig_result, args);
      |                                ^~~~
/usr/lib/ocaml/caml/memory.h:362:36: note: in definition of macro ‘CAMLxparam2’
  362 |     (caml__roots_##x.tables [1] = &y), \
      |                                    ^
cpp11_function_objects_wrap.c:2124:3: note: in expansion of macro ‘SWIG_CAMLlocal2’
 2124 |   SWIG_CAMLlocal2(swig_result, args);
      |   ^~~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:812:16: error: expected ‘;’ before ‘caml__temp_val’
  812 |   caml_value_t caml__temp_val = (val); \
      |                ^~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:2135:3: note: in expansion of macro ‘SWIG_Store_field’
 2135 |   SWIG_Store_field(swig_result,0,args);
      |   ^~~~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:809:43: error: expected primary-expression before ‘)’ token
  809 | #define SWIG_Field(x, i) (((caml_value_t *)(x)) [i])           /* Also an l-value. */
      |                                           ^
cpp11_function_objects_wrap.c:813:17: note: in expansion of macro ‘SWIG_Field’
  813 |   caml_modify (&SWIG_Field ((block), caml__temp_offset), caml__temp_val); \
      |                 ^~~~~~~~~~
cpp11_function_objects_wrap.c:2135:3: note: in expansion of macro ‘SWIG_Store_field’
 2135 |   SWIG_Store_field(swig_result,0,args);
      |   ^~~~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:813:58: error: ‘caml__temp_val’ was not declared in this scope; did you mean ‘caml_ptr_val’?
  813 |   caml_modify (&SWIG_Field ((block), caml__temp_offset), caml__temp_val); \
      |                                                          ^~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:2135:3: note: in expansion of macro ‘SWIG_Store_field’
 2135 |   SWIG_Store_field(swig_result,0,args);
      |   ^~~~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:763:22: error: ‘value’ does not name a type
  763 | #define caml_value_t value
      |                      ^~~~~
cpp11_function_objects_wrap.c:764:20: note: in expansion of macro ‘caml_value_t’
  764 | #define CAML_VALUE caml_value_t
      |                    ^~~~~~~~~~~~
cpp11_function_objects_wrap.c:2138:16: note: in expansion of macro ‘CAML_VALUE’
 2138 |   static const CAML_VALUE *swig_ocaml_func_val = NULL;
      |                ^~~~~~~~~~
cpp11_function_objects_wrap.c:2139:8: error: ‘swig_ocaml_func_val’ was not declared in this scope
 2139 |   if (!swig_ocaml_func_val) {
      |        ^~~~~~~~~~~~~~~~~~~
cpp11_function_objects_wrap.c:2142:33: error: ‘swig_ocaml_func_val’ was not declared in this scope
 2142 |   swig_result = caml_callback3(*swig_ocaml_func_val,swig_get_self(),caml_copy_string("operator ()"),args);
      |                                 ^~~~~~~~~~~~~~~~~~~
make[2]: *** [../../../Examples/Makefile:979: ocaml_static_cpp] Error 2
make[1]: *** [Makefile:78: cpp11_function_objects.cpptest] Error 2
```

ojwb:
> perhaps the CI problem is due to an out-of-tree build?

A local out-of-tree build also works, so it's not that.

Also, without the patch from this PR testcase `cpp11_function_objects` passes, so the change here doesn't seem to be good to merge as-is.

@jamesjer You should be able to run just that one testcase to reproduce the failure with:

```
make -C Examples/testsuite/ocaml cpp11_function_objects.cpptest
```

There are also some warnings about ocaml code from `make check-ocaml-examples` which are there with git master already but if the fix is obvious to someone who knows ocaml it'd be good to eliminate these:

```
checking Examples/ocaml/std_vector
SWIG:1: Warning 524: Experimental target language. Target language OCaml specified by -ocaml is an experimental language. Please read about SWIG experimental languages, https://swig.org/Doc4.0/Introduction.html#Introduction_experimental_status.
File "../../../../../Examples/ocaml/std_vector/runme.ml", line 26, characters 6-35:
26 |       v -> push_back ((x to float)) ;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 10 [non-unit-statement]: this expression should have type unit.
File "../../../../../Examples/ocaml/std_vector/runme.ml", line 33, characters 2-29:
33 |   u -> push_back ((i to int))
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 10 [non-unit-statement]: this expression should have type unit.
checking Examples/ocaml/stl
SWIG:1: Warning 524: Experimental target language. Target language OCaml specified by -ocaml is an experimental language. Please read about SWIG experimental languages, https://swig.org/Doc4.0/Introduction.html#Introduction_experimental_status.
File "../../../../../Examples/ocaml/stl/runme.ml", line 10, characters 42-62:
10 |     let str = (Sys.argv.(i)) to string in v -> push_back (str)
                                               ^^^^^^^^^^^^^^^^^^^^
Warning 10 [non-unit-statement]: this expression should have type unit.
```

ojwb:
It occurred to me to look if `value` is used in `cpp11_function_objects.i` and it is, and that's the problem.  Renaming it similarly to `bools.i` fixes this testcase too.  Will push a fix for that shortly.

With those two commits, the examples and testsuite both pass locally for me.  The examples give the warnings above, but they're already present on git master.  We also lack CI coverage for ocaml, but that's also just as true on git master.  So it doesn't seem either should block merging this, though it'd be really helpful for the long-term viability of SWIG/Ocaml to address them (CI especially).

@jamesjer One other thing I would really like to see is a clarification about `value` vs `caml_value_t` vs `CAML_VALUE`, ideally added as a comment where we define them:

```
#define caml_value_t value
#define CAML_VALUE caml_value_t
```

Looking at the ocaml headers I have locally (from Debian's package version 4.13.1-4) only `value` is used in the headers, so what are the other two?

My initial guess was older/newer names for this type, but searching the web, the only hits for `caml_value_t` at least seem to be related to SWIG.

ojwb:
ocaml CI passes on ubuntu-22.04 if I drop `$SWIGJOBS` from when running tests, so this seems to be a parallelism issue.  I suspect it's generating a file with a fixed pathname and parallel jobs collide.  I'm not seeing where exactly (possibly it's internal to the ocaml tools) and I haven't managed to trigger it locally, but I have a more self-contained change which fixes this outside of CI too by adding `.NOPARALLEL:` to `Examples/ocaml/Makefile.in`.

The failure on ubuntu-20.04 remains and seems to be due to a C/C++ compiler warning in ocaml-generated C code.  I've tried adding `-Wno-error=narrowing` which stops this warning being fatal but there are further failures.  Given ocaml is an experimental target language just having the one CI job seems fine to me, especially as it's using the more recent OS and ocaml versions.

ojwb:
> OCaml itself only typedefs `value` even in the latest version. It has never done anything different.

Oh, so these are indeed entirely SWIG inventions!

Digging in git I found 9ac1568443c4e48310ba4cdbc9428ccad65e1ef3 re `caml_value_t`:

> I've moved completely away from the "value" type to "caml_value_t".  Although
> a bit painful, this will make things easier for me.  I seem to be seeing
> quite a few things that use the word value as a function, type, etc, and
> wanted to get something that doesn't collide.

The point that `value` is liable to collide is valid, but using a macro to rename it in the ocaml headers seems a bit dubious.  Really this would be better addressed by ocaml upstream.  (Also, if we do try to do this, using something prefixed by `caml_` is a bad idea as that's someone else's namespace - a `swig_`-prefixed name would be better.)

I didn't find a clear explanation for `CAML_VALUE` but it seems to be a different attempt to do the same thing, and again is treading on someone else's namespace.

One thing I'm not quite seeing is how the `value` macro was causing problems in C++ headers as there's `#undef value` at the end of `Lib/ocaml/ocamlrun.swg` - is that just being done too late and should instead be right after including ocaml headers to work as intended?

If we stop doing that renaming (as proposed here) we should probably clean up uses of both `caml_value_t` and `CAML_VALUE`, though that is probably better done as a separate change.

I lean towards thinking stopping trying to do this renaming as proposed is a reasonable change, though it does risk breaking existing user interface files by introducing a clash with ocaml's `value` typedef, and which may be hard to fix from the user interface file.

rwmjones:
`value` is a typedef so you can't undefine it.  However unless something includes `<caml/mlvalues.h>` then they won't have it so I'm not sure how it would affect pure C++ code.

I agree that `CAML_VALUE` & `caml_value_t` appear to have no benefit for swig (maybe in OCaml though, but basically every OCaml extension and the whole OCaml compiler uses `value` so good luck changing that).

ojwb:
> `value` is a typedef so you can't undefine it.

SWIG can undefine its macro `value` though.

The trick SWIG is currently using is to essentially do:

```
#define value caml_value_t
#include <caml/mlvalues.h>
/* And any other ocaml headers */
#undef value
```

So `typedef intnat value;` in caml/mlvalues.h actually does `typedef intnat caml_value_t;` instead. and uses of `value` in the ocaml headers instead use that typedef.  Then in the rest of SWIG's generated `_wrap.c`/`_wrap.cxx` it uses `caml_value_t` instead of `value`, and there's no `value` typedef to collide with anything.

> However unless something includes `<caml/mlvalues.h>` then they won't have it so I'm not sure how it would affect pure C++ code.

The generated `_wrap.cxx` will include the ocaml headers and the headers for the library being wrapped in the same compilation unit, e.g. using git master and looking at the "class" example, which wraps code in header `example.h`:

```
$ cd Examples/ocaml/class/
$ make
[...]
$ grep 'define value\|caml/mlvalues\|undef value\|example\.h' example_wrap.cxx 
#define value caml_value_t
#include <caml/mlvalues.h>
#undef value
#include "example.h"
$
```

So currently `example.h` can use `value` in ways it couldn't if this value macro hack is removed as this PR proposes to do - that's why I needed to change `bools.i` (5a11b14bf02271d1fc3d568e8b4561c760171485) and `cpp11_function_objects.i` (eecee6a556b68132f5d8151c31e68b08233ae59b) to get the testsuite to pass with the changes in this PR.  Clearly this could also affect user interface files, but they may not have the luxury of being able to simply rename a use of `value` in the API being wrapped which clashes with ocaml's `typedef` because that API is often an existing third-party library.

This `value` typedef renaming is quite a nasty little hack, but I'm not clear why it is causing problems for you.  I wonder if that `#undef value` is just happening a bit too late, and if so moving it earlier would be a change with less risk of breaking existing user interface files.

(The `CAML_LOCAL_ROOTS` change seems unproblematic.)

ojwb:
I've been trying to get CI testing for ocaml working again.  Doing so, it seems clear that the value typedef hack is problematic.

It works for uses of the `value` typedef in the ocaml headers, and uses in SWIG-generated code we can replace with our renamed version.  However it doesn't work for uses of "value" in macros defined in the ocaml headers for use in code using the ocaml C API.  The previous SWIG/ocaml maintainer attempted to solve this by trying to have SWIG provide tweaked copies to replace all such macros, which have inevitably diverged from those defined by ocaml; also SWIG lacks replacements for more recently introduced macros.

It's far from ideal that ocaml's C API uses a common term like "value" as a typedef as it can clash with uses of "value" as an identifier in the C/C++ API being wrapped, but that's not a problem that SWIG created or (it seems) really is able to solve.

So I've removed that hack and cleaned up around it.

CI is now passing with ocaml 4.13.1 and locally it works with 4.14.1, but there are some problems that ought to be fixed but require more ocaml knowledge than I have.  See below for details.

My branch is a clear improvement over the current situation, and I'm going to merge it for 4.2.1 since ocaml is an flagged as an "experimental" language and our documentation for that clearly says "No backwards compatibility is guaranteed as the module is effectively 'in development'."

Here's the list of problems that require ocaml skills.

# Testing with ocaml 5

Based on the changes proposed here I think this should fix ocaml 5 support too, but I don't have an easy way to test with this version.  Please can someone confirm?

Better still, if someone can add a CI job using ocaml5 we can be confident support won't be accidentally broken.  The CI runs on Ubuntu 20.04 or 22.04.

# SWIG Examples give ocaml "non-unit-statement" warning

```
make -C Examples/ocaml/std_vector
```

emits "non-unit-statement" warnings which I have no idea how to solve:

```
File "../../../../Examples/ocaml/std_vector/runme.ml", line 26, characters 6-35:
26 |       v -> push_back ((x to float)) ;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 10 [non-unit-statement]: this expression should have type unit.
File "../../../../Examples/ocaml/std_vector/runme.ml", line 33, characters 2-29:
33 |   u -> push_back ((i to int))
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning 10 [non-unit-statement]: this expression should have type unit.
```

There's the same warning from `make -C Examples/ocaml/stl`.  This is just a warning so doesn't cause CI to fail, but it really should be addressed.

# Testsuite fails with parallel make

`make -j2 check-ocaml-test-suite` randomly fails with:

```
File "swig.ml", line 1:
Error: Could not find the .cmi file for interface swig.mli.
```

I haven't investigated but it smells like there's a fixed name being used for a generated file, and as a result one parallel job can stomp on this file for another parallel job.  For now I've just disabled parallel make for ocaml in CI.

# Fails with ocaml 4.08.1

Doesn't seem to be a SWIG problem, but rather ocaml relying on deprecated C semantics for empty parameter lists, e.g.:

```
2024-02-22T21:40:55.3883786Z  8949 | extern value caml_get_public_method();
2024-02-22T21:40:55.3883899Z       |              ^~~~~~~~~~~~~~~~~~~~~~
2024-02-22T21:40:55.3884042Z In file included from /tmp/camlobja00dca.c:6:
2024-02-22T21:40:55.3884557Z /usr/lib/ocaml/caml/mlvalues.h:219:18: note: previous declaration ‘value caml_get_public_method(value, value)’
2024-02-22T21:40:55.3884837Z   219 | CAMLextern value caml_get_public_method (value obj, value tag);
2024-02-22T21:40:55.3884951Z       |                  ^~~~~~~~~~~~~~~~~~~~~~
2024-02-22T21:40:55.3885395Z /tmp/camlobja00dca.c:9131:14: error: conflicting declaration of C function ‘value caml_set_oo_id()’
2024-02-22T21:40:55.3885510Z  9131 | extern value caml_set_oo_id();
2024-02-22T21:40:55.3885609Z       |              ^~~~~~~~~~~~~~
```

For now I've simply documented the oldest version we know works (4.13.1) but if ocaml 4.08 is still relevant it would be helpful to have CI for this.

# Generated code for some testcases passes `NULL` to `strcpy()`

Testcases `arrays`, `memberin_extend_c` and `sizeof_pointer` generate code tries to `strcpy` to a variable that has value `NULL`, as gcc -Wnonnull reports:                                                                

```
warning: argument 1 null where non-null expected [-Wnonnull] 
```

None of these have a `_runme.ml` so the pass despite this.  You can run these individually using for example:

```
make -C Examples/test-suite/ocaml arrays.ctest
```

# Known failing testcases

This are marked as known to fail so the testsuite passes, but they ought to be resolved and from the errors it looks like there a very small number of common causes.

See `FAILING_CPP_TESTS` and `FAILING_C_TESTS` in `Examples/test-suite/ocaml/Makefile.in`.

You can run an individual test with e.g. (for a C test it needs to be `.ctest` instead):

```
make -C Examples/test-suite/ocaml allprotected.cpptest
```

You can run also all the tests marked as failing and report any which now pass with:

```
make -s -C Examples/test-suite/ocaml check-failing
```

rwmjones:
Non-unit statement [`unit` is like `void` in C] is a warning that the statement returns some value and the value is being ignored.  You either have to disable the warning on the compiler command line (`-w -10` IIRC), or much better add an explicit `ignore (...)` around the statement in the code.  eg:

```
$ echo '4+5; ()' > test.ml
$ ocamlopt test.ml
File "test.ml", line 1, characters 0-3:
1 | 4+5; ()
    ^^^
Warning 10 [non-unit-statement]: this expression should have type unit.
$ echo 'ignore (4+5); ()' > test.ml
$ ocamlopt test.ml
```

Parallel test suite fails / Could not find the .cmi file

OCaml Makefile dependencies are subtle.  In particular when you have only an `*.ml` file then the compiler generates both the `*.cmi` (compiled interface) and `*.cmx` (compiled object file).  If you have a `*.mli` file then that has to generate the `*.cmi` file first before you try to compile the `*.ml` file (otherwise two parallel instances will try to overwrite each other or run in the wrong order).

Naive attempts to hand-write Makefile rules won't capture this so you won't have a full set of dependencies.  You probably need to use the included `ocamldep` program to generate them rather than trying to do it yourself.

OCaml 4.07 is still used by RHEL 8, so that's still somewhat relevant.  OCaml 4.08 was released in 2019 so is also still used by some Linux distros.  (This may not be a problem unless you're planning to backport swig to these older distros though)

ojwb:
> add an explicit `ignore (...)` around the statement in the code.

Great - that's just what I was after.  Fixed locally, will push shortly.

> Naive attempts to hand-write Makefile rules won't capture this so you won't have a full set of dependencies. You probably need to use the included `ocamldep` program to generate them rather than trying to do it yourself.

Thanks.  I think that's more than I want to get into, but I've noted this in a comment by where we disable parallel make for ocaml in CI.

> OCaml 4.07 is still used by RHEL 8, so that's still somewhat relevant. OCaml 4.08 was released in 2019 so is also still used by some Linux distros. (This may not be a problem unless you're planning to backport swig to these older distros though)

As SWIG upstream we don't backport SWIG to anything, though we are mindful that packagers may want to.  I did have a few tries at getting it to work, though I was hampered by not having local access to 4.08 so each try required a push and wait for CI to run.  There seem to be two problems (though maybe more hidden by these):

* GCC `-Wnarrowing` warnings from a byte array initialiser - these looks harmless to me and just due to char being `signed` on x86-64, but they're promoted to errors because `-Werror` is in effect.  Using GCC from the distro rather than GCC 13 didn't seem to help, and I couldn't manage to figure out how to pass `-Wno-error=narrowing`.

* Errors due to using C prototypes without an empty parameter list.  Again, using GCC from the distro rather than GCC 13 didn't seem to help, and neither did lowering the `-std=cNN` setting.

Currently SWIG/ocaml is marked as "experimental" and I think there are more critical problems to solve with the aim of moving it back out of that status, but if someone cares about and has local access to ocaml 4.08 (or 4.07 probably) it'd likely not be very hard to resolve, especially if they are familiar with ocaml usage.

SWIG/ocaml is at least in better shape now than for any recent SWIG release.

ojwb:
The changes proposed here have been merged (7454f40827874d3fa126f80a11c925c9dbdb3f12) along with additional changes.  We don't have confirmation that ocaml 5 is now working, but it's been more than two weeks so I've opened an issue detailing the remaining problems I noted above (#2833) and am closing this.

PR #2653, opened 2023-07-07
guard against applying in typemaps twice

### LINKED ISSUE #2652: [JS] napi generator inserts `in` typemaps twice when using `numinputs=0`
```
ENGINE=napi make ignore_parameter.cpptest
grep -C 1 "hello" ignore_parameter_wrap.cxx
```

mmomtchev:
Fix #2652
Add a check for applying `numinputs=0` twice for all generators since this is a special case in `emit.cxx:emit_attach_parmmaps()`

### DIFF
--- a/Examples/test-suite/ignore_parameter.i
+++ b/Examples/test-suite/ignore_parameter.i
@@ -6,7 +6,7 @@
   static const char* hi = "hello";
   $1 = const_cast<char *>(hi);
   unique++;
-  if (unique != 1) SWIG_Error(SWIG_ERROR, "in typemap applied more than once");
+  if (unique != 1) SWIG_exception_fail(SWIG_ERROR, "in typemap applied more than once");
 }
 %typemap(in,numinputs=0) int bb "$1 = 101; called_argout = 0;"
 %typemap(in,numinputs=0) double ccc "$1 = 8.8;"
--- a/Examples/test-suite/ignore_parameter.i
+++ b/Examples/test-suite/ignore_parameter.i
@@ -2,7 +2,12 @@
 
 %module ignore_parameter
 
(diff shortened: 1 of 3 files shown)


ojwb:
It'd be nicer if we could have this fail at compile-time if our "canary" typemap is substituted twice since that would also catch such an error with target languages which don't have a _runme for this testcase - just over half don't currently.

I tried catching it by having a variable which would end up defined multiple times in 4b1a44f88119649f0675699c68d34fa9f460e794, but that doesn't fail for napi as one of the substitutions is inside its own block and so we end up with:

```
  int if_multiply_defined_then_in_typemap_substituted_more_than_once;
  arg1 = const_cast<char *>("hello");

  {
    int if_multiply_defined_then_in_typemap_substituted_more_than_once;
    arg1 = const_cast<char *>("hello");

  }
```

Looks like using a label will still give an error in this case, so trying that.

ojwb:
> Looks like using a label will still give an error in this case, so trying that.

The good news is that we don't seem to have this problem for any of the target languages without a runme, and that the label trick does fail at compile time for NAPI (with your fix reverted):

```
../../ignore_parameter_wrap.cxx: In function ‘Napi::Value _wrap_jaguar(const Napi::CallbackInfo&)’:
../../ignore_parameter_wrap.cxx:1946:3: error: duplicate label ‘if_multiply_defined_then_in_typemap_substituted_more_than_once’
 1946 |   if_multiply_defined_then_in_typemap_substituted_more_than_once:
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Minor bad news is that the unused label triggers an error with node (and a second error with napi) as these seem to build with `-Werror` on.  I'll add `SWIGUNUSED` to suppress that warning and merge this shortly.

PR #2450, opened 2022-11-25
MonoAOT fixes

Hazardu:
DCO to be confirmed.

PR #2658, opened 2023-07-31
In case Unity with il2cpp, AOT.MonoPInvokeCallback is needed.

laoleshisui:
SWIG_CSHARP_NO_EXCEPTION_HELPER is ok when only C, but not when CXX is enabled.
SWIGStringHelper has this problem too, but defining SWIG_CSHARP_NO_STRING_HELPER is fine when both C and C++.


ojwb:
Your proposed change breaks the csharp tests in CI with lots of:

```
examplePINVOKE.cs(55,6): error CS0246: The type or namespace name `AOT' could not be found. Are you missing an assembly reference?
```

ojwb:
There's also #1262 and #2450 which appear related (though admittedly I know almost nothing about C#).

laoleshisui:
There is a new change commit of [csharp.cxx] to support direcotor

laoleshisui:
In this way, it supports only one implementation in C# sadly.

ojwb:
@laoleshisui This is still breaking C# jobs in CI.

laoleshisui:
@ojwb I know little about c# too, and have no idea why 'The type or namespace name `AOT' could not be found'. Ask for help. :)

wsfulton:
I'm afraid I don't see why SWIG_CSHARP_NO_EXCEPTION_HELPER is causing any specific problems for C code and not C++ code. The problem here is not clear and the solution breaks the C# SWIG module so am closing. If you can provide an example of the problem we may be able to help. If your issue is to do with AOT, then please take a look at the AOT issues #2450 and #1262.

PR #2665, opened 2023-08-16
Fix missing-field-initializers warning with Py3.12

shadchin:
https://docs.python.org/3.12/c-api/typeobj.html?highlight=tp_watched#c.PyTypeObject.tp_watched

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -4203,6 +4203,9 @@ public:
     Printv(f, "#if (PY_VERSION_HEX >= 0x03080000) && (PY_VERSION_HEX < 0x03090000)\n", NIL);
     printSlot(f, getSlot(), "tp_print");
     Printv(f, "#endif\n", NIL);
+    Printv(f, "#if PY_VERSION_HEX >= 0x030C0000\n", NIL);
+    printSlot(f, getSlot(), "tp_watched");
+    Printv(f, "#endif\n", NIL);
 
     Printv(f, "#ifdef COUNT_ALLOCS\n", NIL);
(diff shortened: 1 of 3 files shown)


Source/Modules/python.cxx
```diff
@@ -4203,6 +4203,9 @@ class PYTHON:public Language {
     Printv(f, "#if (PY_VERSION_HEX >= 0x03080000) && (PY_VERSION_HEX < 0x03090000)\n", NIL);
     printSlot(f, getSlot(), "tp_print");
     Printv(f, "#endif\n", NIL);
+    Printv(f, "#if PY_VERSION_HEX >= 0x030C0000\n", NIL);
+    printSlot(f, getSlot(), "tp_watched");
+    Printv(f, "#endif\n", NIL);
```
ojwb, line 4208: It looks like you've based this off the handling for `tp_print` but most slots look for a feature set on the node, e.g.:

```
printSlot(f, getSlot(n, "feature:python:tp_vectorcall"), "tp_vectorcall", "vectorcallfunc");
```

It also looks to me like `getSlot()` always returns `"0"`, so I'm not sure I really understand what's intended for the few existing slots which use it.  Someone who understands the existing code should review.

ojwb: Oh I just saw wsfulton has set himself as a reviewer.

ojwb:
@wsfulton Did you see my query above?

Source/Modules/python.cxx
wsfulton: Good spot, I missed that. Now added in the missing slot feature support in f64ce2cdb76a5a4ce0749320d2c5fcbd38b9ddc5.

PR #2681, opened 2023-09-28
Enable nullable reference types in C# (Recreated PR)

roetlich:
This was originally in: https://github.com/swig/swig/pull/1713

In C# 8 nullable reference types are [disabled in generated files](https://github.com/dotnet/roslyn/blob/70e158ba6c2c99bd3c3fc0754af0dbf82a6d353d/docs/features/nullable-reference-types.md#generated-code) by default, and generators should enable them by inserting `#nullable enable` or `#nullable restore` into each file.

I added the CLI option `-enablenullable` to insert `#nullable enable` into all generated C# files after the banner. This allows people to use nullable references types without breaking backward compatibility.

### DIFF
--- a/Doc/Manual/CSharp.html
+++ b/Doc/Manual/CSharp.html
@@ -111,6 +111,11 @@ swig -csharp -help
 </tr>
 
 <tr>
+<td>-enablenullable</td>
+<td>Enable <a href="https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references">nullable reference types</a> in the C# code. This inserts the <tt>#nullable enable</tt> directive into all C# files, because otherwise <a href="https://github.com/dotnet/roslyn/blob/70e158ba6c2c99bd3c3fc0754af0dbf82a6d353d/docs/features/nullable-reference-types.md#generated-code"> nullable is disabled for all generated files.</a></td>
+</tr>
+
+<tr>
 <td>-namespace &lt;nm&gt;</td>
 <td>Generate wrappers into C# namespace &lt;nm&gt;</td>
 </tr>
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -46,6 +46,7 @@ class CSHARP:public Language {
   bool global_variable_flag;	// Flag for when wrapping a global variable
   bool old_variable_names;	// Flag for old style variable names in the intermediary class
   bool generate_property_declaration_flag;	// Flag for generating properties
+  bool enable_nullable_flag;
 
   String *imclass_name;		// intermediary class name
   String *module_class_name;	// module class name
@@ -122,6 +123,7 @@ public:
       global_variable_flag(false),
       old_variable_names(false),
       generate_property_declaration_flag(false),
+      enable_nullable_flag(false),
       imclass_name(NULL),
       module_class_name(NULL),
       imclass_class_code(NULL),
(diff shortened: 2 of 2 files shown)


ojwb:
Copied metadata settings from the ticket this replaces.

ojwb:
wsfulton's comment from the original PR still seems relevant:
> The main concern I have in this patch is that it generates code in a very inflexible way. I'd prefer an approach where a user could add in this code with more flexibility.

How about something like adding a `csbegin` section with which one could achieve the effect of the patch here with:

```
%insert(csbegin) "#nullable enable";
```

This addresses the original concern since the code inserted isn't hard-coded and you could just as easily insert some variation on this, or use it to insert something else for a different purpose (so long as it works to insert it at that point in every generated `.cs` file.  You also get the full feature set of `%insert` so you can insert text from a file, or use SWIG macros in the text to insert.

This approach would also mean the interface file is more self-contained - with the patch here you'd need to know to process the interface file with a particular command line option.

roetlich:
Hi @ojwb, this sounds like a better solution, if it works. But I wouldn't know how to implement that. 

@adj123, would this fit your use case?

adj123:
> Hi @ojwb, this sounds like a better solution, if it works. But I wouldn't know how to implement that.
> 
> @adj123, would this fit your use case?

It makes it possible to explicitly enable nullable annotations in each file, so yes, it does work for me.

I could imagine though that if I were someone on the SWIG maintenance team who had an interest in the elegance of its user-facing design, I would probably have objections to this. In that I can't really think of any time you would ever want to inject a block of code into every C# file when the decision of how code elements are distributed across files is an implementation detail in control of SWIG - all that comes to my mind is preprocessor directives that affect how the file is compiled. But in [MSDN's list of directives](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) the only one I see that would be useful is `#nullable enable` (anything else can be specified in the project file).

But hey that's my 2c about something that doesn't affect me personally very much so if the SWIG maintainers disagree I wouldn't try to stand in the way, so long as it enables the feature I need.

adj123:
I do agree it would be nicer if the `#nullable enable` was configured to be added via a command in a SWIG interface file rather than a command line option though

ojwb:
`Doc/Manual/Extending.html` documents adding a new section, but I think the tl;dr is you just call e.g. `Swig_register_filebyname("csbegin", f_csbegin);` (and declare the variable with `File *f_csbegin;`) in `csharp.cxx` to register the new section, and then you can `Dump(f_csbegin, f);` where you're currently conditionally writing a fixed string.

> I could imagine though that if I were someone on the SWIG maintenance team who had an interest in the elegance of its user-facing design, I would probably have objections to this. In that I can't really think of any time you would ever want to inject a block of code into every C# file when the decision of how code elements are distributed across files is an implementation detail in control of SWIG - all that comes to my mind is preprocessor directives that affect how the file is compiled. But in [MSDN's list of directives](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) the only one I see that would be useful is #nullable enable (anything else can be specified in the project file).

It wasn't me that highlighted the inflexibility, but the approach I suggested is at least consistent with how SWIG supports adding code to other places in generated files, and this sort of consistency in user-facing design seems more elegant than adding single-purpose options.

See the older PR for a motivation from wsfulton for specifying a conditionalised setting of nullable, but from your link I'd think warning suppressions might be usefully set like this.  Also new directives may be added in future versions of C#.

wsfulton:
Support has now been added. Use `%module` and `csbegin` as follows:
```swig
%module(csbegin="#nullable enable") mymodule
```
This generic technique can be used to add common comments, import statements and preprocessor directives into all generated .cs files.

I didn't implement it using `%insert` as the insert sections generally target one section in one particular file and a `%insert` is no good if used after the first struct/class as each struct/class generates a complete .cs file.

PR #2690, opened 2023-10-12
Use single characters instead of strings where possible

AtariDreams:
This makes the code clearer and faster

### DIFF
--- a/Examples/lua/lua.c
+++ b/Examples/lua/lua.c
@@ -291,7 +291,7 @@ static void manual_input (void) {
     }
   }
   lua_settop(L, 0);  /* clear stack */
-  fputs("\n", stdout);
+  fputc('\n', stdout);
   progname = oldprogname;
 }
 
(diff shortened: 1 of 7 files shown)


ojwb:
Not saying it's a bad change, though with modern compilers `fputs("\n", f);` actually generates the exact same assembler as `fputc('\n', f);` so that at least won't be any faster: https://godbolt.org/z/MsEvcd5an

AtariDreams:
> Not saying it's a bad change, though with modern compilers `fputs("\n", f);` actually generates the exact same assembler as `fputc('\n', f);` so that at least won't be any faster: https://godbolt.org/z/MsEvcd5an

Alright.

ojwb:
Please save up these changes and push as one batch.  Each time you push it triggers over 100 CI jobs which is a lot of unnecessary load on the CI system.  I've cancelled the backlog you've created twice now, which eats into the time I can spend actually reviewing changes.

PR #2691, opened 2023-10-12
Remove redundant initialization code

AtariDreams:
Some of this code is redundant and taken care of by the initializer.

### DIFF
--- a/Source/Modules/lang.cxx
+++ b/Source/Modules/lang.cxx
@@ -363,10 +363,12 @@ Language::Language():
 none_comparison(NewString("$arg != 0")),
 director_ctor_code(NewString("")),
 director_prot_ctor_code(0),
+director_multiple_inheritance(1),
 symtabs(NewHash()),
 overloading(0),
 multiinput(0),
-cplus_runtime(0) {
+cplus_runtime(0),
+doxygenTranslator(NULL) {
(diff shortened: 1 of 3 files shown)


Source/Modules/lang.cxx
```diff
@@ -363,7 +363,7 @@ Language::Language():
 none_comparison(NewString("$arg != 0")),
 director_ctor_code(NewString("")),
 director_prot_ctor_code(0),
-symtabs(NewHash()),
```
wsfulton, line 366: May as well also add in doxygenTranslator() to complete the list. Coding convention is one line per initializer please.

AtariDreams: @wsfulton Done!

Source/Modules/lang.cxx
```diff
@@ -378,12 +380,8 @@ cplus_runtime(0) {
      Default director 'protected' constructor code, disabled by
      default. Each language that needs it, has to define it.
    */
-  director_prot_ctor_code = 0;
```
ojwb, line 381: The comment just above is about this member variable and is just confusing with this line of code removed.  Probably best relocated to where the member is defined in `swigmod.h`.

AtariDreams: Done! @ojwb

PR #2697, opened 2023-10-19
Lib/csharp: Better standardized std_vector.i and std_array.i

emmenlau:
This PR is only a minor change. I compared the sources of std_vector.i and std_array.i, and have copied items that could be implemented more concise between the two. Let me know if this is useful?

### DIFF
--- a/Lib/csharp/std_array.i
+++ b/Lib/csharp/std_array.i
@@ -24,6 +24,17 @@
     }
   }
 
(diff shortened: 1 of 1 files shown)


wsfulton:
The IsEmpty and constructor changes introduces a divergence between std_vector.i and std_array.i. The cpp11_std_array_runme.cs test now fails. Maybe you are using an old version?? It might be a good idea if you ran the C# test-suite before submitting changes as each push to the git repo uses a **lot** of compute resources.

emmenlau:
Closing as not relevant.

PR #2698, opened 2023-10-19
Differences in C# long handling

emmenlau:
The current upstream swig features better support for long integral types in the C# bindings, thanks to https://github.com/swig/swig/commit/fb91d1fa25ba00f93b9870d84ec3cbc0739650d7. However sadly, my C++/C# binding does not build neither on Linux now Windows with current upstream alone. On the other hand, it *does* work with the previous changes from https://github.com/swig/swig/pull/2379.

Therefore I would like to better understand the differences between the two implementations, and see if my code needs to be fixed or if current upstream swig could use some improvement.

This PR is a continuation of the previous PR https://github.com/swig/swig/pull/2379. I have unified the changes that where first proposed in https://github.com/swig/swig/pull/2237 with the latest upstream code. However, where there have been differences, I preferred the older, unmerged PR, over the current latest upstream swig. This allows me to successfully build my larger C++/C# binding on Linux and Windows successfully.

I found that some definitions difficult to understand in the current latest upstream code. This PR is meant to provide the means for discussion.

It contains:
* Integrate https://github.com/swig/swig/pull/650/files
* Introduce INT64_IS_LONG_LONG
* Additional 'long' vs. 'long long' fixes
* Better standardize long type support with upstream

As an example of the differences in code that I get with current upstream swig: The following method should return the size of an array, which I think is correctly returned as `long` on a 64bit platform, but current upstream swig uses `int` instead:
```
SWIGEXPORT unsigned int SWIGSTDCALL CSharp_MyNameSpace_UInt64Size2Array_size(void * jarg1) {
  unsigned int jresult ;
  std::array< std::uint64_t,2 > *arg1 = (std::array< std::uint64_t,2 > *) 0 ;
  std::array< unsigned long,2 >::size_type result;
  
  arg1 = (std::array< std::uint64_t,2 > *)jarg1; 
  {
    try {
      result = ((std::array< std::uint64_t,2 > const *)arg1)->size();
    } catch(const std::exception& vEx) {
      {
        SWIG_CSharpException(SWIG_RuntimeError, vEx.what()); return 0; 
      };
    } catch(...) {
      {
        SWIG_CSharpException(SWIG_UnknownError, "Unknown exception"); return 0; 
      };
    }
  }
  jresult = (unsigned int)result; 
  return jresult;
}
```

Also, the C++ type `size_t` is mapped to `int` which is not what I would expect on a 64bit platform.

### DIFF
--- a/Lib/csharp/csharp.swg
+++ b/Lib/csharp/csharp.swg
@@ -82,8 +82,8 @@ SWIGINTERN const char * SWIG_UnpackData(const char *c, void *ptr, size_t sz) {
 %typemap(ctype) unsigned short,     const unsigned short &     "unsigned short"
 %typemap(ctype) int,                const int &                "int"
 %typemap(ctype) unsigned int,       const unsigned int &       "unsigned int"
-%typemap(ctype) long,               const long &               "int"
-%typemap(ctype) unsigned long,      const unsigned long &      "unsigned int"
+%typemap(ctype) long,               const long &               "long"
+%typemap(ctype) unsigned long,      const unsigned long &      "unsigned long"
 %typemap(ctype) long long,          const long long &          "long long"
 %typemap(ctype) unsigned long long, const unsigned long long & "unsigned long long"
 %typemap(ctype) float,              const float &              "float"
@@ -98,8 +98,13 @@ SWIGINTERN const char * SWIG_UnpackData(const char *c, void *ptr, size_t sz) {
 %typemap(imtype) unsigned short,     const unsigned short &     "ushort"
 %typemap(imtype) int,                const int &                "int"
 %typemap(imtype) unsigned int,       const unsigned int &       "uint"
+#if defined(SWIGWORDSIZE64)
+%typemap(imtype) long,               const long &               "long"
+%typemap(imtype) unsigned long,      const unsigned long &      "ulong"
+#else
 %typemap(imtype) long,               const long &               "int"
 %typemap(imtype) unsigned long,      const unsigned long &      "uint"
+#endif
 %typemap(imtype) long long,          const long long &          "long"
 %typemap(imtype) unsigned long long, const unsigned long long & "ulong"
 %typemap(imtype) float,              const float &              "float"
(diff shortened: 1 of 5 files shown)


Lib/csharp/csharp.swg
```diff
@@ -82,8 +82,8 @@ SWIGINTERN const char * SWIG_UnpackData(const char *c, void *ptr, size_t sz) {
 %typemap(ctype) unsigned short,     const unsigned short &     "unsigned short"
 %typemap(ctype) int,                const int &                "int"
 %typemap(ctype) unsigned int,       const unsigned int &       "unsigned int"
-%typemap(ctype) long,               const long &               "int"
-%typemap(ctype) unsigned long,      const unsigned long &      "unsigned int"
+%typemap(ctype) long,               const long &               "long"
+%typemap(ctype) unsigned long,      const unsigned long &      "unsigned long"
```
emmenlau, line 86: Why are the long types mapped to name "int" instead of "long"? Is that sensible?

wsfulton: C int is 32-bits and so now it matches the C# 32-bit int. It is purposefully marshalled as 32-bits on both sides by default.

emmenlau: Maybe I'm confused now, but the current swig master is the top version, which maps `long` to `int`. Is that correct?

wsfulton: That's right. The default is to support C long as 32-bit. C long is sometimes 64-vit and sometimes 32-bit, whereas C int is always 32-bit. By using C int, it ensures the types used on both the unmanaged and managed sides are always 32-bit. SWIGWORDSIZE64 uses int instead of long long which is always 64-bit. Please see the updated CSharp.html docs in  fb91d1fa25ba00f93b9870d84ec3cbc0739650d7 and if not clear, improvements are welcome.

emmenlau: I'm super sorry but I'm still confused what is the correct way to work with `size_t` and `ptrdiff_t` in current swig master. I have in my interface file:
```
#if defined(SWIGWORDSIZE64)
%apply unsigned long { size_t };
%apply const unsigned long& { const size_t& };
%apply long { ptrdiff_t };
%apply const long& { const ptrdiff_t& };
#else
%apply unsigned long long { size_t };
%apply const unsigned long long& { const size_t& };
%apply long long { ptrdiff_t };
%apply const long long& { const ptrdiff_t& };
#endif

// Include common stl types:
%include "stdint.i"
%include "std_string.i"
%include "std_shared_ptr.i"
%include "bda/StdContainers.i"

[...]
```

I'm under the impression that this generates an interface now where `size`t is mapped to `unsigned long long` even on platforms where `SWIGWORDSIZE64` is defined. I would expect it to use `unsigned long` for `size_t`, as the `%apply` directive says. I tested that `SWIGWORDSIZE64` is defined in the interface file, so this is not the cause of the problem. Why the `unsigned long long` instead of `unsigned long`?

The final code still works because on Linux, `unsigned long long` and `unsigned long` are both 64bit unsigned integral types. But the whole thing is confusing for me, sorry for that. Can you kindly help?

emmenlau:
Closing in favor of discussion under https://github.com/swig/swig/issues/2702

PR #2727, opened 2023-12-05
Python stable abi

wsfulton:
Adds support for the Python stable abi.

Updates and supercedes #2190.

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -65,6 +65,7 @@
 <li><a href="#Python_nn37">Overhead and code bloat</a>
 <li><a href="#Python_nn38">Typemaps</a>
 <li><a href="#Python_nn39">Miscellaneous</a>
+<li><a href="#Python_stable_abi">Stable ABI</a>
 </ul>
 <li><a href="#Python_nn40">Common customization features</a>
 <ul>
@@ -964,7 +965,6 @@ swig -python -help
 <tr><td>-nothreads      </td><td>Disable thread support for the entire interface</td></tr>
 <tr><td>-olddefs        </td><td>Keep the old method definitions when using -fastproxy</td></tr>
 <tr><td>-relativeimport </td><td>Use relative Python imports</td></tr>
-<tr><td>-stable-abi     </td><td>Generate code compatible with the Python stable ABI</td></tr>
 <tr><td>-threads        </td><td>Add thread support for all the interface</td></tr>
 <tr><td>-O              </td><td>Enable the following optimization options: -fastdispatch -fastproxy -fvirtual</td></tr>
 
@@ -3319,6 +3319,34 @@ director feature for reentrant, recursive or threaded member
 methods that return const references.
 </p>
 
+<H3><a name="Python_stable_abi">33.5.8 Stable ABI</a></H3>
+
+
+<p>
+By default, the version of Python used to compile the wrappers needs to be the same as that used during runtime.
+Alternatvely, the <a href="https://docs.python.org/3/c-api/stable.html">Python Stable ABI</a> enables a single compiled binary to be used by different versions of Python.
+This is enabled by defining <tt>Py_LIMITED_API</tt> during the compilation of the C/C++ wrapper code and setting this macro to a particular minimum version of Python that one wants to support.
+</p>
+
+<p>
+SWIG supports the stable ABI, but only version 3.4 of Python and later is supported.
+There are two recommended approaches for using SWIG and the stable ABI and both require setting the <tt>Py_LIMITED_API</tt> macro to be set to <tt>0x03040000</tt> as a minimum value (Python 3.4).
+Either set this using <tt>%begin</tt> by adding the following into your interface file so that this macro appears at the beginning of the generated C/C++ code:
+</p>
+
+<div class="code"><pre>
+%begin %{
+#define Py_LIMITED_API 0x03040000
+%}
+</pre></div>
+
+<p>
+or simply define the macro using your C/C++ compiler's <tt>-D</tt> command line option, for example, <tt>-DPy_LIMITED_API=0x0304000</tt>.
+</p>
+
+<p>
+<b>Compatibility Note:</b> Support for the stable ABI was added in SWIG-4.2.0.
+</p>
 
 <H2><a name="Python_nn40">33.6 Common customization features</a></H2>
 
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -964,6 +964,7 @@ swig -python -help
 <tr><td>-nothreads      </td><td>Disable thread support for the entire interface</td></tr>
 <tr><td>-olddefs        </td><td>Keep the old method definitions when using -fastproxy</td></tr>
 <tr><td>-relativeimport </td><td>Use relative Python imports</td></tr>
+<tr><td>-stable-abi     </td><td>Generate code compatible with the Python stable ABI</td></tr>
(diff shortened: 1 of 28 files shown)


Doc/Manual/Python.html
```diff
@@ -964,6 +964,7 @@ <H3><a name="Python_commandline">33.2.9 Additional Python commandline options</a
 <tr><td>-nothreads      </td><td>Disable thread support for the entire interface</td></tr>
 <tr><td>-olddefs        </td><td>Keep the old method definitions when using -fastproxy</td></tr>
 <tr><td>-relativeimport </td><td>Use relative Python imports</td></tr>
+<tr><td>-stable-abi     </td><td>Generate code compatible with the Python stable ABI</td></tr>
```
ojwb, line 967: I don't really like SWIG's "runallthewordstogether" command line option naming scheme, but `-stableabi` would be more consistent with it.

Source/Modules/python.cxx
```diff
@@ -626,6 +641,10 @@ class PYTHON:public Language {
       Printf(f_runtime, "#define SWIGPYTHON_FASTPROXY\n");
     }
 
+    if (stable_abi) {
+      Printf(f_runtime, "#define Py_LIMITED_API 0x03040000\n");
+    }
```
ojwb, line 646: It'd be helpful if the user could override this value (at least it will become helpful once SWIG can make use of `PyUnicode_AsUTF8AndSize()` being added to the stable ABI in 3.10 and the non-deprecated buffer API being added in 3.11).

Given `-stable-abi` doesn't actually affect SWIG's code-gen, maybe we should just generate a wrapper file that supports both (which is pretty much what we already seem to do) and leave defining `Py_LIMITED_API` or not and the choice of its value to the user - they could define it on the compiler command line, or in their interface file via something like:

```
%pythonbegin %{
#define Py_LIMITED_API 0x030A0000
%}
```
```

Comment (vadz) [reply Source/Modules/python.cxx:646]: What about generating

```cpp
#ifndef Py_LIMITED_API
#define Py_LIMITED_API 0x03040000
#endif
```

? This would still work by default out of the box but would allow overriding it with a later value by people knowing what they're doing.

wsfulton: We could do this, the only downside is that we can't easily test for different versions of Py_LIMITED_API. Some users may even want to a workable way to build for different abi versions easily by specifying command line options.

So how about the following two ways of turning on the stable ABI as follows:

Either of:
1. Via the command line, for example to specify minimum of 3.4:
`-limitedapi=0x03040000`. The minimum version can be omitted as follows:
`-limitedapi`, whereupon the the minimum version of Python 3 that SWIG
 currently supports will be specified. Syntax follows that of the only other optional value that I can find (`-D`). Note that the command line name is changed to more accurately reflect what the value it is for. The difference between 'stable abi' and 'limited api' is somewhat blurred to me.
2. Via %begin, for example to specify minimum of 3.4:
```
%begin %{
#define Py_LIMITED_API 0x03040000
%}
```

How come we are currently specifying 0x03040000 as the minimum and not just 3 or 0x03020000 (the first version that the stable abi was introduced) or 0x03030000 (the minimum version that SWIG currently supports for non-stable abi)?

vadz: > We could do this, the only downside is that we can't easily test for different versions of Py_LIMITED_API.

Why can't we? It's as simple as enabling stable ABI for different Python versions. Whether we should do this, and if it's worth it, is another question. IMO no because stable ABI is most (only?) useful with the lowest supported version anyhow.

> Some users may even want to a workable way to build for different abi versions easily by specifying command line options.

I don't see any need for this in practice as per above but why not.
 
> So how about the following two ways of turning on the stable ABI as follows:
> 
> Either of:
> 
>     1. Via the command line, for example to specify minimum of 3.4:
>        `-limitedapi=0x03040000`. The minimum version can be omitted as follows:
>        `-limitedapi`, whereupon the the minimum version of Python 3 that SWIG
>        currently supports will be specified.

The latter part _really_ doesn't make any sense to me, this is just not how/what stable ABI is for. Not specifying anything should default to the lowest supported version.

> Syntax follows that of the only other optional value that I can find (`-D`). Note that the command line name is changed to more accurately reflect what the value it is for. The difference between 'stable abi' and 'limited api' is somewhat blurred to me.

I prefer "stable ABI" because it's more clear, but the idea is that only limited API is included in stable ABI.

>     2. Via %begin, for example to specify minimum of 3.4:
> 
> ```
> %begin %{
> #define Py_LIMITED_API 0x03040000
> %}
> ```

Again, why not, but specifying `-DPy_LIMITED_API=0x03040000` in the compiler flags still looks much simpler for me and this is what I'd personally prefer to use. Why deal with the idiosyncratic SWIG flags when you can just use the well known (and stable since 50+ years) compiler `-D` switch?

> How come we are currently specifying 0x03040000 as the minimum and not just 3 or 0x03020000 (the first version that the stable abi was introduced) or 0x03030000 (the minimum version that SWIG currently supports for non-stable abi)?

I think 3.4 is the first version in which stable ABI is actually usable, it didn't include some absolutely required parts before then. I also think it would be perfectly fine to bump up the minimum SWIG supported version to 3.4.

ojwb: Adding a SWIG command line option to specify a `Py_LIMITED_API` value to hard-code into the generated file doesn't seem a helpful approach to me either.  SWIG generates a Python C/C++ wrapper with C/C++ preprocessor conditionals such that it can build with `Py_LIMITED_API` undefined or with it defined to a value of the builders choice.

wsfulton: I worked out that 3.4 minimum is needed for the limited api as the wrappers require `PyType_GetSlot`.

wsfulton: Okay, so the message I'm getting is that using `-D` when compiling the wrappers can define the version for `Py_LIMITED_API` and providing ways for SWIG to define this is not a good idea. Given the `-stable-abi` just defines `Py_LIMITED_API` and doesn't affect code generation in any other way, it seems that there is no point to the `-stable-abi` option at all and so I'll drop it altogether. The documented ways to turn on the stable abi will then be to define `Py_LIMITED_API` during C/C++ code compilation using either `%begin` or `-D` option passed to the C/C++ compiler. 

All round this is much simpler and less complicated for SWIG. I can also modify the test-suite to test for different versions of the abi by adding support for `EXTRA_CXXFLAGS` in Github CI. We need to test different abi version as there is a major change as @ojwb points out in our code base for version 3.10 (for `PyUnicode_AsUTF8AndSize()` for more efficient string handling) which needs testing.

vadz: Thanks, I hadn't realized this but if SWIG can indeed generate the same code with and without `-stable-abi` and just compile it differently depending on whether `Py_LIMITED_API` is defined, it can indeed simplify things.

I think chances of someone defining ``Py_LIMITED_API` by accident are nil, practically speaking, so the only drawback I see in doing it like this is that it won't be very discoverable (e.g. you won't see it in `swig -help` output), but provided it's mentioned in the manual, it's probably not a great deal.

wsfulton: Implemented in ef0bbcd6ed1a4f4356b885168082219b121adb11.

wsfulton:
Also related: #2428.

jschueller:
the cli option `-stable-abi` is not mentionned everywhere, should it be removed ?

ojwb:
> the cli option `-stable-abi` is not mentionned anywhere, should it be removed ?

Thanks - removed in ab1f9f076c9f480882796687af2365ebb54b05ed.

PR #2729, opened 2023-12-12
[Tcl] Use `Tcl_GetString()` instead of `Tcl_GetStringFromObj(…, NULL)`

chrstphrchvz:
I notice that there is currently no usage of `Tcl_GetString()` in SWIG, but I am not aware if it that was intentional.

Future Tcl introduces the `Tcl_Size` type for use in C APIs. In Tcl 8.7, it is defined as `int` for binary compatibility, while in Tcl 9.0, it is defined to `ptrdiff_t` to support ≥32-bit sized things.

`Tcl_GetStringFromObj()` will expect its second parameter to have type `Tcl_Size *`. There is currently a compatibility approach in place which allows `int *` to be passed instead, but disallows `Tcl_GetStringFromObj(…, 0)`, and makes `Tcl_GetStringFromObj(…, NULL)` non-portable. The alternative to using `Tcl_GetString()` to avoid this problem would be to use casts, i.e. `Tcl_GetStringFromObj(…, (Tcl_Size *)NULL)` (see 6c46ff3800350b5da241557ca9d95222348e1790).

### DIFF
--- a/Doc/Manual/Tcl.html
+++ b/Doc/Manual/Tcl.html
@@ -2780,7 +2780,7 @@ used as a <tt>char **</tt> object.
   }
   $1 = (char **) malloc((nitems+1)*sizeof(char *));
   for (i = 0; i &lt; nitems; i++) {
-    $1[i] = Tcl_GetStringFromObj(listobjv[i], 0);
+    $1[i] = Tcl_GetString(listobjv[i]);
   }
   $1[i] = 0;
 }
--- a/Examples/tcl/multimap/example.i
+++ b/Examples/tcl/multimap/example.i
@@ -25,7 +25,7 @@ extern int    gcd(int x, int y);
   }
   $2 = (char **) malloc(($1+1)*sizeof(char *));
   for (i = 0; i < $1; i++) {
-    $2[i] = Tcl_GetStringFromObj(listobjv[i],0);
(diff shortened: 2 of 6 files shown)


ojwb:
> I notice that there is currently no usage of `Tcl_GetString()` in SWIG, but I am not aware if it that was intentional.

I doubt there's anyone still involved in SWIG who could really answer that.

My uninformed guess would be that perhaps this dates to before `Tcl_GetString()` was added (SWIG's Tcl support dates back to at least Tcl7.3). I couldn't find when exactly it was added but assuming stub slots are allocated in ascending order it's 340 compared to `Tcl_GetStringFromObj()` which is 41.  However there's a changelog entry from 2003 which mentions adding a call to `Tcl_GetString()` so it's been around long enough we can just rely on it.

Will merge shortly.

chrstphrchvz:
> My uninformed guess would be that perhaps this dates to before Tcl_GetString() was added (SWIG's Tcl support dates back to at least Tcl7.3). 

That makes sense; `Tcl_GetString()` was introduced in Tcl 8.1.

ojwb:
Thanks - SWIG 4.1.0 dropped support for Tcl < 8.4 so there's no need to worry about older versions now.

PR #2730, opened 2023-12-13
Rename SWIG’s `Tcl_GetBoolFromObj()`

chrstphrchvz:
Proposed on behalf of @PaulObermeier.

Tcl 8.7 introduces a function with the same name, as part of [TIP 618](https://core.tcl-lang.org/tips/doc/trunk/tip/618.md).


ojwb:
Thanks.  I've adjusted the new name to be `SWIG_Tcl_GetBoolFromObj()` to follow the pattern of existing SWIG helper functions that are specific to Tcl.

PR #2736, opened 2023-12-21
[perl,ruby] Emit stringview directorout warning

ojwb:
See #1567

### DIFF
--- a/Lib/typemaps/ptrtypes.swg
+++ b/Lib/typemaps/ptrtypes.swg
@@ -65,7 +65,7 @@
 #if defined(SWIG_DIRECTOR_TYPEMAPS)
 /* directorout */
 
-%define %ptr_directorout_typemap(asptr_meth,frag,Type...)
+%define %ptr_directorout_typemap_with_warning(asptr_meth, frag, Warning, Type...)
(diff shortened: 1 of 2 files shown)


ojwb:
@wsfulton This adds the warning for Perl and Ruby too, but I wonder if there's a better approach to implementing this?  The `dummy=1` parameter this adds for other uses of these typemaps seems ugly, but `warning=Warning` results in an empty warning message (hmm, perhaps `warning=` should be changed to not emit a warning at all?)

ojwb:
Gah, I meant to push the PHP fix to master not this branch - I've move it there and force pushed to remove.

The intended change here breaks several other languages so clearly can't be merged as-is.  Sorry but the maze of macros that is the UTL isn't something to be hacking at in a rush and I've also pretty much run out of time when I can look at this this year, having already spent more time than I really had spare this week on SWIG ("mid to end of December" really doesn't seem a good choice for a release deadline - it's especially problematic in the Southern hemisphere as it's Summer as well as xmas and the new year).

@wsfulton If you can see a better way to achieve this, please go for it.  Otherwise I suggest we just don't have this warning for Perl and Ruby for 4.2.0.

Issue #2628, opened 2023-05-31
C/C++ standards versions required for compiling SWIG itself


ojwb:
I know we aim to support compiling generated wrappers as C90/C++98 - to be clear this is **NOT** about changing that.

Is that requirement also meant to apply to compiling SWIG itself?  I'd vaguely assumed so, but after a `git bisect` run I somehow ended up with `-ansi -pedantic` having been automatically added to my `CFLAGS` and `CXXFLAGS` even after resetting to HEAD and accidentally discovered there are multiple places where this isn't currently the case, e.g.:

```
CParse/templ.c: In function ‘Swig_cparse_template_expand’:
CParse/templ.c:488:11: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  488 |           Parm *parm = ParmList_nth_parm(templateparms, index);
      |           ^~~~
CParse/templ.c:576:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  576 |             Node *tynode = Swig_symbol_clookup(s, 0);
      |             ^~~~
```

The first was introduced by 817c700ced95b2fc615e1619c89fcec21f66af55, the second by 674abaddbf854e2374f115ef08e59fa79702e19a, both changes authored by @wsfulton (not pointing fingers here - my point is that if wsfulton is not keeping to strict C90 that makes me wonder if that's actually an aim).

There are some in parser.y too (one introduced by me, another by wsfulton).

It looks like these are all since the last release.

On the C++ side, there's also a use of `nullptr` in `javascript.cxx` introduced very recently.

If it is genuinely important to non-hypothetical users to be able to compile SWIG itself as C90/C++98 we should fix these and add flags to CI to enforce (ideally to developer builds too so you don't have to wait for CI to fail to find transgressions).

I think my view is that requiring declarations before code is not helpful and adds extra friction during development.  Especially given the mix of C and C++ code in SWIG , it's hard to always remember that you're meant to do last-century-style declaration at the top of the block while also thinking about actually getting everything else right about the code.

Having to split declaration and assignment in C code introduces scope for bugs as uses of the unassigned variable can accidentally be introduced someone changing the code.

Mixed declarations and code are supported by many C compilers as an extension before C99 (GCC, clang and MSVC all allow them without even a warning with `-std=c90`/`/std:c90` - I tested MSVC 19 using godbolt.org).

There are several ways to formally legitimise declarations after code.  We could:

[...]

PR #2742, opened 2023-12-31
Improve Error Handling and Prevent Path Traversal at DohNewFile() in file.c

brandnewblueberry:
This Pull Request introduces several important updates to the DohNewFile function in order to enhance its security and robustness. These changes aim to prevent potential path traversal vulnerabilities and address memory leak issues.

Details of Changes:
Error Handling in File Opening:

Updated the function to return NULL instead of 0 when fopen fails. This makes the error handling more consistent with C standards for functions returning pointers.
Memory Leak Prevention:

Added memory cleanup steps before returning NULL when fopen fails or if subsequent memory allocations (DohMalloc) fail. This ensures that there is no memory leak if fopen or memory allocations are unsuccessful.
Potential Path Traversal Prevention:

Although not directly implemented in this PR, a recommendation is made to integrate a isValidFilename function. This function should validate the input filename to mitigate potential path traversal vulnerabilities. The proposed validation includes checks against relative paths, absolute paths, and other unsafe patterns.
Reason for Changes:
The original implementation of DohNewFile lacked proper error handling for file opening failures and potential memory leaks. It also did not account for security vulnerabilities like path traversal, which are critical in file handling functions.
By returning NULL on failure, the function now clearly indicates error states in a manner consistent with C pointer handling conventions.
The added memory cleanup steps prevent resource leaks, which is crucial for long-running applications.
The recommendation for filename validation is to prompt a review and implementation of security best practices, especially for functions dealing with external inputs and file system operations.
Additional Notes:
The implementation of isValidFilename function is not included in this PR but is highly recommended for future updates.
Further testing and review are advised to ensure the changes align well with the rest of the codebase.
Please review the changes and consider merging them into the main branch. Your feedback and suggestions for further improvements are welcome.

### DIFF
--- a/Source/DOH/file.c
+++ b/Source/DOH/file.c
@@ -286,20 +286,35 @@ DOH *DohNewFile(DOHString *filename, const char *mode, DOHList *newfiles) {
   char *filen;
 
   filen = Char(filename);
+
   file = fopen(filen, mode);
-  if (!file)
-    return 0;
+  if (!file) {
+    return NULL;
+  }
 
   f = (DohFile *) DohMalloc(sizeof(DohFile));
+  if (!f) {
+    fclose(file);
+    return NULL;
+  }
+
   if (newfiles)
     Append(newfiles, filename);
+
   f->filep = file;
   f->fd = 0;
   f->closeondel = 1;
   obj = DohObjMalloc(&DohFileType, f);
+  if (!obj) {
+    DohFree(f);
+    fclose(file);
+    return NULL;
+  }
+
   open_files_list_add(f);
   return obj;
 }
+
 
 /* -----------------------------------------------------------------------------
  * NewFileFromFile()


wsfulton:
The patch looks fine; it improves error handling, thanks.

With regards to path traversal vulnerabilities, you'll need to explain more. Note that SWIG should be able to open any file without restriction, so `fopen` provides the necessary error handling for SWIG. SWIG is not a web server, so path traversal vulnerabilities don't really make sense to me anyway.

ojwb:
> Updated the function to return NULL instead of 0 when fopen fails. This makes the error handling more consistent with C standards for functions returning pointers.

In a function with a pointer return type, `return 0;` is the same as `return NULL;` (in some C/C++ implementations `NULL` is actually effectively implemented by `#define NULL 0`).  Using `NULL` here makes no functional difference so "makes the error handling more consistent with C standards for functions returning pointers" doesn't really hold water.  `NULL` is arguably clearer to a human reading the code though.

Source/DOH/file.c
```diff
@@ -286,21 +286,36 @@ DOH *DohNewFile(DOHString *filename, const char *mode, DOHList *newfiles) {
...
-    return 0;
+  if (!file) {
+    return NULL;
+  }
 
   f = (DohFile *) DohMalloc(sizeof(DohFile));
+  if (!f) {
+    fclose(file);
+    return NULL;
+  }
```
ojwb, line 299: `DohMalloc()` prints an error and exits if memory allocation fails so it's not necessary or useful to check like this.

Source/DOH/file.c
```diff
@@ -286,21 +286,36 @@ DOH *DohNewFile(DOHString *filename, const char *mode, DOHList *newfiles) {
...
+
   f->filep = file;
   f->fd = 0;
   f->closeondel = 1;
   obj = DohObjMalloc(&DohFileType, f);
+  if (!obj) {
+    DohFree(f);
+    fclose(file);
+    return NULL;
+  }
```
ojwb, line 312: Same applies to `DohObjMalloc()` (this exit-on-allocation-failure strategy was implemented in e38847f7e161c022f794de6bf28d5f7da5cba3c1 which debuted in SWIG 4.1.0 - maybe you looked at the code of an older release?)

ojwb:
I think the only change here which is actually useful is to use `NULL` instead of `0`.  Using a literal `0` for `NULL` is very widespread in the SWIG code base, probably a reflection of its age.

Rather than changing just a single instance I think using `NULL` consistently would be better.  GCC has `-Wzero-as-null-pointer-constant` which will "[w]arn when a literal '0' is used as null pointer" which should allow us to find all such uses fairly painlessly.  We might even be able to have a CI build which uses `-Werror=zero-as-null-pointer-constant` to prevent regressions on this.

However this would change a lot of code which will tend to make existing patches fail to apply cleanly and existing branches not merge cleanly.  @wsfulton Thoughts?

ojwb:
> I think using NULL consistently would be better

Or perhaps `nullptr` in C++ code if the conclusion of #2628 is that we require C++11 to build SWIG itself (SWIG 4.2.0 actually requires a C++11 compiler to build since there's a use of `nullptr` as I noted in #2628 which didn't get resolved).

ojwb:
> GCC has `-Wzero-as-null-pointer-constant`

Oh except that only works for C++ code.

As above the added checks for allocation failing are not useful as our allocation functions print an error and exit on failure.  The only useful change here is to change a **single** instance of `0` as a pointer constant to `NULL` and the SWIG codebase is rife with using `0` as a pointer constant - especially the older and lower level stuff like DOH.

I'll also explicitly note that there are no code changes corresponding to the "Prevent Path Traversal" mentioned in the title.

So overall I think it's best to just close this.

wsfulton:
> So overall I think it's best to just close this.

Agreed, thanks.

While we could create a coding standard to use NULL for pointers and make the changes everywhere it doesn't offer much value and indeed widespread changes add unnecessary challenges. But if someone really wants to spend extra time on this kind of improvements, I wouldn'tt object.

PR #2752, opened 2024-01-09
Improved support for Tcl9

PaulObermeier:
Changes done as explained in chapters **Init stubs** and **64-bit data support** at the Tcl9 porting site:
https://core.tcl-lang.org/tcl/wiki?name=Migrating%20C%20extensions%20to%20Tcl%209

Replaced int with Tcl_Size where appropriate.
Replaced INT_MAX with TCL_SIZE_MAX.
Changed SWIG_TCL_STUB_VERSION from "8.4" to "8.4-". 
Corrected calls of load command in some examples.

### DIFF
--- a/Examples/tcl/multimap/example.c
+++ b/Examples/tcl/multimap/example.c
@@ -3,6 +3,12 @@
 #include <stdlib.h>
 #include <ctype.h>
 
+#include <tcl.h>
+
+#ifndef TCL_SIZE_MAX
+typedef int Tcl_Size;
+#endif
+
 /* Compute the greatest common divisor of positive integers */
 int gcd(int x, int y) {
   int g;
@@ -15,7 +21,7 @@ int gcd(int x, int y) {
   return g;
 }
 
-int gcdmain(int argc, char *argv[]) {
+int gcdmain(Tcl_Size argc, char *argv[]) {
   int x,y;
   if (argc != 3) {
     printf("usage: gcd x y\n");
@@ -27,7 +33,7 @@ int gcdmain(int argc, char *argv[]) {
   return 0;
 }
 
-int count(char *bytes, int len, char c) {
+int count(char *bytes, Tcl_Size len, char c) {
   int i;
   int count = 0;
   for (i = 0; i < len; i++) {
@@ -36,7 +42,7 @@ int count(char *bytes, int len, char c) {
   return count;
 }
 
-void capitalize(char *str, int len) {
+void capitalize(char *str, Tcl_Size len) {
   int i;
   for (i = 0; i < len; i++) {
     str[i] = (char)toupper(str[i]);
--- a/Examples/tcl/multimap/example.c
+++ b/Examples/tcl/multimap/example.c
@@ -3,12 +3,6 @@
 #include <stdlib.h>
 #include <ctype.h>
 
-#include <tcl.h>
-
-#ifndef TCL_SIZE_MAX
-typedef int Tcl_Size;
-#endif
-
 /* Compute the greatest common divisor of positive integers */
 int gcd(int x, int y) {
   int g;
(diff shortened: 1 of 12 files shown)


Examples/tcl/multimap/example.c
```diff
@@ -3,6 +3,12 @@
 #include <stdlib.h>
 #include <ctype.h>
 
+#include <tcl.h>
+
+#ifndef TCL_SIZE_MAX
+typedef int Tcl_Size;
+#endif
+
```
ojwb, line 11: This part doesn't seem right - these changes should really only affect the generated code, but here the API being being wrapped is effectively being changed.

Lib/tcl/tclwstrings.swg
```diff
@@ -44,11 +44,11 @@ SWIGINTERNINLINE Tcl_Obj *
...
 {
   Tcl_Obj *res = NULL;
-  if (size < INT_MAX) {
+  if (size < TCL_SIZE_MAX) {
     Tcl_Encoding encoding = NULL;
     char *src = (char *) carray;
-    int srcLen = (int)(size*sizeof(wchar_t));
-    int dstLen = (int)(size*sizeof(Tcl_UniChar));
+    Tcl_Size srcLen = (Tcl_Size)(size*sizeof(wchar_t));
+    Tcl_Size dstLen = (Tcl_Size)(size*sizeof(Tcl_UniChar));
```
ojwb, line 51: This is an pre-existing problem and something to deal with separately, but I notice these multiplications can overflow.  Perhaps the check above should be:

```
if (size < TCL_SIZE_MAX / sizeof(wchar_t) && size < TCL_SIZE_MAX / sizeof(Tcl_UniChar)) {
```

Examples/tcl/multimap/example.c
PaulObermeier: Here I was too smart in replacing int's with Tcl_Size. 
Should be OK now.

Lib/tcl/tclinit.swg
```diff
@@ -24,7 +24,7 @@ SWIGEXPORT int SWIG_init(Tcl_Interp *);
 
 /* Compatibility version for TCL stubs */
 #ifndef SWIG_TCL_STUBS_VERSION
-#define SWIG_TCL_STUBS_VERSION "8.4"
+#define SWIG_TCL_STUBS_VERSION "8.4-"
```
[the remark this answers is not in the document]
PaulObermeier: This feature is documented with Tcl command "package vsatisfies":
https://www.tcl.tk/man/tcl9.0/TclCmd/package.html
I will create a ticket at Tcl Core to update the documentation for related C functions.

ojwb:
I've merged as the changes all seem sensible and don't regress Tcl 8 support.

However, testing with Tcl 9.0 (at least with the tcl9.0-dev 9.0.0~b1~rc1+dfsg-1 package currently in experimental) I see multiple failures (mostly seg faults but also other problems) from both of:

```
make check-tcl-examples
make check-tcl-test-suite
```

So it seems there's more work to do for Tcl 9 support.

It'd be useful if someone who understands the changes in Tcl 9 could work through the failures from the above commands.

PaulObermeier:
Using the official 9.0.b1 Tcl/Tk release (https://sourceforge.net/projects/tcl/files/Tcl/9.0b1/), all examples and tests succeed (except boost and pcre, which I do not have installed).

ojwb:
Great.  It'd be good to have CI coverage with tcl 9 to make sure we don't regress on that, but that'll be much simpler once we can just install a suitable package on an ubuntu version GHA supports using apt and I didn't find a PPA with it in yet in a quick search.  The pace of change in SWIG/Tcl is hardly frenetic so I think for now we can just be mindful of this if any Tcl patches get merged.

PR #2762, opened 2024-01-15
Replace nullptr with NULL (javascript.cxx)

PaulObermeier:
When replacing nullptr by NULL, SWIG 4.2.0 can still be compiled with gcc 4.9.X (ex. Debian Jessie).


wsfulton:
Thanks, committed in 472e9dd1548f48f4b658b4f7c242b0aacacbd811.

PR #2771, opened 2024-01-23
Allow passing NULL value to sq_ass_item

jim-easterbrook:
The `sq_ass_item` slot calls `PySequence_SetItem`, which can be used to delete an element by passing a `NULL` value. Calling `Py_INCREF` on a `NULL` value causes a segfault. This commit adds a `NULL` check similar to that in other closure functions.

### DIFF
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -540,8 +540,10 @@ SwigPyBuiltin_ternaryfunc_closure(SwigPyWrapperFunction wrapper, PyObject *a, Py
   assert(tuple);
   Py_INCREF(b);
   PyTuple_SET_ITEM(tuple, 0, b);
-  Py_INCREF(c);
-  PyTuple_SET_ITEM(tuple, 1, c);
+  if (c) {
+    Py_INCREF(c);
+    PyTuple_SET_ITEM(tuple, 1, c);
+  }
   result = wrapper(a, tuple);
   Py_DECREF(tuple);
   return result;
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -656,8 +656,10 @@ SwigPyBuiltin_ssizeobjargproc_closure(SwigPyWrapperFunction wrapper, PyObject *a
   tuple = PyTuple_New(2);
   assert(tuple);
   PyTuple_SET_ITEM(tuple, 0, _PyLong_FromSsize_t(b));
-  Py_INCREF(c);
-  PyTuple_SET_ITEM(tuple, 1, c);
+  if (c) {
+    Py_INCREF(c);
+    PyTuple_SET_ITEM(tuple, 1, c);
+  }
   resultobj = wrapper(a, tuple);
   result = resultobj ? 0 : -1;
   Py_XDECREF(resultobj);


jim-easterbrook:
The AppVeyor check is failing on `template_default_class_parms_typedef.cpptest` - I don't see how this could be a result of the small change in this PR.

ojwb:
Yes, that's was bogus failure.

It'd be good to add test coverage for the bug you're fixing though, so we can be confident the fix works for all Python versions we have CI for, and so we catch if a future change regresses your fix.

wsfulton:
I restarted the bogus failed job a few days ago to give you a green build.

Is there a simple way to replicate this bug?

jim-easterbrook:
Minimal SWIG interface file `segfault.i`:
```
%module segfault

%feature("python:slot", "sq_item", functype="ssizeargfunc")
    Foo::__getitem__;
%feature("python:slot", "sq_ass_item", functype="ssizeobjargproc")
    Foo::__setitem__;
%typemap(default) PyObject* value {$1 = NULL;}

%inline %{
class Foo {
public:
    int __getitem__(int idx) {
        return 123;
    }
    void __setitem__(int idx, PyObject* value) {
    }
};
%}
```

Corresponding `setup.py` file:
```
from distutils.core import setup, Extension

segfault_module = Extension(
    '_segfault', sources=['segfault_wrap.cxx'])

setup (name = 'segfault',
       version = '0.0',
       author      = "Jim Easterbrook",
       description = "Segfault demonstration",
       ext_modules = [segfault_module],
       py_modules = ["segfault"],
       )
```

Build with:
```
$ swig -c++ -python -builtin segfault.i 
$ python3 setup.py build_ext --inplace
```

Interactive Python session (after building with SWIG 4.2.0):
```
$ python3
Python 3.6.15 (default, Sep 23 2021, 15:41:43) [GCC] on linux
>>> import segfault
>>> f = segfault.Foo()
>>> f[0] = 123
>>> del f[0]
```

jim-easterbrook:
I've just noticed that `SwigPyBuiltin_ssizeobjargproc_closure` used to use the `NULL`-safe `Py_XINCREF` so this bug would have been introduced then.
https://github.com/swig/swig/commit/d30effd88688d262181ef13073cb8deb08553145

ojwb:
So this was a regression introduced in 4.0.0.

Were the other calls which used to be `Py_XINCREF()` safe to change?

jim-easterbrook:
> Were the other calls which used to be `Py_XINCREF()` safe to change?

Good question! I just went through all the closure functions in `Lib/python/builtin.swg` and found that the `ternaryfunc` used for the `tp_call` slot also needs a `NULL` test.

Test module:
```
%module segfault

%feature("python:slot", "sq_item", functype="ssizeargfunc")
    Foo::__getitem__;
%feature("python:slot", "sq_ass_item", functype="ssizeobjargproc")
    Foo::__setitem__;
%feature("python:slot", "tp_call", functype="ternaryfunc")
    Foo::__call__;
%typemap(default) PyObject* value, PyObject* kw {$1 = NULL;}

%inline %{

class Foo {
public:
    int __getitem__(int idx) {
        return 123;
    }
    void __setitem__(int idx, PyObject* value) {
    }
    void __call__(PyObject* args, PyObject* kw) {
    }
};

%}
```

Interactive run:
```
$ python
Python 3.6.15 (default, Sep 23 2021, 15:41:43) [GCC] on linux
>>> import segfault
>>> f = segfault.Foo()
>>> f(123, fred=456)
>>> f(123)
```

wsfulton:
I'll review this one next.

PR #2784, opened 2024-01-30
Fix 1126 related

jim-easterbrook:


jim-easterbrook:
Not a clean PR - will resubmit

Issue #1126, opened 2017-10-17
Python %extend with default args, fastunpack, modernargs and builtin


jim-easterbrook:
I'm having a problem using default arguments when adding methods to a C structure and swigging with optimisation and -builtin. Here's a simple example:
```
/* File: test.i */
%module test
struct foo {};
%extend foo {
  void no_arg() {}
  void one_req(int *required) {}
  void one_opt(int *optional = NULL) {}
  void two_arg(int *required, int *optional = NULL) {}
};
```
With `swig -python -py3 -builtin test.i` the wrapper contains the following:
```
SWIGINTERN PyMethodDef SwigPyBuiltin__foo_methods[] = {
  { "no_arg", (PyCFunction) _wrap_foo_no_arg, METH_VARARGS, (char*) "" },
  { "one_req", (PyCFunction) _wrap_foo_one_req, METH_VARARGS, (char*) "" },
  { "one_opt", (PyCFunction) _wrap_foo_one_opt, METH_VARARGS, (char*) "" },
  { "two_arg", (PyCFunction) _wrap_foo_two_arg, METH_VARARGS, (char*) "" },
  { NULL, NULL, 0, NULL } /* Sentinel */
};
```
With `swig -python -py3 -builtin -fastunpack -modernargs test.i` this changes to
```
SWIGINTERN PyMethodDef SwigPyBuiltin__foo_methods[] = {
  { "no_arg", (PyCFunction) _wrap_foo_no_arg, METH_NOARGS, (char*) "" },
  { "one_req", (PyCFunction) _wrap_foo_one_req, METH_O, (char*) "" },
  { "one_opt", (PyCFunction) _wrap_foo_one_opt, METH_O, (char*) "" },
  { "two_arg", (PyCFunction) _wrap_foo_two_arg, METH_VARARGS, (char*) "" },
  { NULL, NULL, 0, NULL } /* Sentinel */
};
```
This means the `one_opt` method's argument is now required, rather than optional. Dropping any one of the `-builtin -fastunpack -modernargs` flags makes all the methods `METH_VARARGS` again. I don't have this problem with functions that are not part of a `%extend` class.

I found issue #607 which appears to be similar, but doesn't mention `%extend`.

wsfulton:
Can you try with swig-3.0.12 if you aren't already. There were some fixes in this area in 3.0.6.

jim-easterbrook:
[...]

PR #2785, opened 2024-01-30
Change logic relating to setting python:argcount

jim-easterbrook:
This fixes a problem I've been having that appears to be related to bug #1126. The `python:argcount` value is used to select `METH_NOARGS`, `METH_O`, or `METH_VARARGS`. I've changed the logic to ensure that `METH_VARARGS` is always used if there are any optional parameters.


jim-easterbrook:
OK, the test suite is failing because single arguments are being labelled as  `METH_VARARGS` but don't get unwrapped with `SWIG_Python_UnpackTuple`. This is obviously harder than I thought, so I'll close this PR.

PR #2787, opened 2024-01-31
CI: Add c99/c++98 job

jschueller:
Makes sure swig can still be compiled on old distros.
See #2628

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,6 +222,9 @@ jobs:
           CPPSTD: c++11
           CSTD: gnu11
         - SWIGLANG: python
+          CPPSTD: c++98
+          CSTD: gnu99
+        - SWIGLANG: python
           CPPSTD: c++11
         - SWIGLANG: r
           CPPSTD: c++11


ojwb:
#2628 is still waiting for @wsfulton to respond though.  I'm getting mixed signals so far - he bumped the milestone rather than addressing the points I'd noted and as a result 4.2.0 requires C++11, but since then some of those points have been addressed.

Is there really still an older distro in support with a C++ compiler which isn't C++11-capable (even if not enabling C++11 support by default)?  E.g. Ubuntu trusty from 2014 is still in extended support for just under 3 months (https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions) and even that has GCC 4.8.2 which seems to have complete C++11 support aside from one minor thing in type traits (https://en.cppreference.com/w/cpp/compiler_support/11).

We're already using the `AX_CXX_COMPILE_STDCXX` macro so we could easily enable use of `-std=c++11` or similar for such compilers.

jschueller:
yes, I think requiring c++11 is very reasonable

PR #2789, opened 2024-01-31
CMake: Bison>=3.5 is required

jschueller:


wsfulton:
Our cmake build still builds so looks okay.

PR #2790, opened 2024-02-01
[Python] Fix handling of "default" typemap

### LINKED ISSUE #2786: Python: default typemap vs overloaded function
At first I thought this was related to #1126 and I attempted to modify the fix for that but now I think it might be something else.

Test module:
```
%module default_param

%typemap(default) bool optional {$1 = true;}

%inline %{
class Foo {
public:
  void overloaded(bool def=false) {}
  void default_param(bool optional) {}
};
%}
```

setup.py:
```
from distutils.core import setup, Extension

default_param_module = Extension(
    '_default_param', sources=['default_param_wrap.cxx'])

setup (name = 'default_param',
       ext_modules = [default_param_module],
       py_modules = ["default_param"],
       )
```

Build:
```
swig -c++ -python -builtin default_param.i
python3 setup.py build_ext --inplace
```

Test:
```
$ python
Python 3.6.15 (default, Sep 23 2021, 15:41:43) [GCC] on linux
>>> import default_param
>>> f = default_param.Foo()
>>> f.overloaded(True)
>>> f.overloaded()
>>> f.default_param(True)
>>> f.default_param()
  File "<stdin>", line 1, in <module>
TypeError: default_param() takes exactly one argument (0 given)
```

However it works as expected if built with `-nofastunpack`. It also works as expected if I edit `default_param_wrap.cxx` and change the entry in `SwigPyBuiltin__Foo_methods` from `METH_O` to `METH_VARARGS`.

ojwb:
Fix handling of "default" typemap applied to method which takes a single argument when "-builtin" option is used.

Fixes #2786

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -4830,6 +4830,35 @@ public:
 	int funpack = fastunpack && !Getattr(n, "sym:overloaded");
 	String *pyflags = NewString("METH_STATIC|");
 	int argcount = Getattr(n, "python:argcount") ? atoi(Char(Getattr(n, "python:argcount"))) : 2;
+	if (argcount == 1) {
+	  // Check if the parameter has a default typemap applied.
+	  ParmList *plist = CopyParmList(Getattr(n, "parms"));
+	  Swig_typemap_attach_parms("default", plist, NULL);
+
+	  Parm *p;
+	  Parm *pnext;
+
+	  for (p = plist; p; p = pnext) {
+	    pnext = nextSibling(p);
+	    String *tm = Getattr(p, "tmap:in");
+	    if (tm) {
+	      Parm *in_next = Getattr(p, "tmap:in:next");
+	      if (in_next)
+		pnext = in_next;
+	      if (checkAttribute(p, "tmap:in:numinputs", "0")) {
+		continue;
+	      }
+	    }
+
+	    if (Getattr(p, "tmap:default")) {
+	      // Set argcount to 2 to trigger use of METH_VARARGS below.
+	      argcount = 2;
+	    }
+	    break;
+	  }
+	  Delete(plist);
+	}
+
 	if (funpack && argcount == 0)
 	  Append(pyflags, "METH_NOARGS");
 	else if (funpack && argcount == 1)


ojwb:
I'm blaming SWIG's test harness for the first round of CI failures - you can set `SWIG_FEATURES=-builtin` when running the whole testsuite, but for a single testcase that doesn't do anything and you seem to have to use `SWIGOPT=-builtin` instead.  I find the lack of consistency (even extending to whether names have a `_`) very frustrating.

The latest CI failures are my fault - I carefully tested locally but failed to include the changes to `python.cxx` in the commit.

jim-easterbrook:
That's working for my real-life example where I found this problem. Thank you.

ojwb:
Thanks for confirming.

I'm thinking about pushing this down to where `python:argcount` gets set as that seems the better place to handle it (and as a bonus would also eliminate having two copies of the code).  I need to check where else `python:argcount` is used just in case somewhere is actually relying on the current arguably wrong value.

The underlying cause seems to essentially be that `python:argcount` is set if and only if `emit_num_arguments(parms) == emit_num_required(parms)`, and the latter is documented to behave in a way which is unhelpful here:

```
 * emit_num_required()
 *
 * Computes the number of required arguments.  This function is safe for
 * use with multi-argument typemaps and knows how to skip over everything
 * properly. Note that parameters with default values are counted unless
 * the compact default args option is on.
```

We could probably expose a version where whether default arguments are counted is controlled by a parameter instead.

ojwb:
> That behaviour of `emit_num_required` is very odd.

It is, but I'm fearful to try to just change this as this function is called by most target language backends, and also from generic code.

Exposing an alternative version in parallel would at least allow people to check through uses one at a time (or maybe one target language at a time) and adjust any which are broken by the current semantics - maybe that would eventually allow eliminating the current version.

jim-easterbrook:
> The underlying cause seems to essentially be that `python:argcount` is set if and only if `emit_num_arguments(parms) == emit_num_required(parms)`, and the latter is documented to behave in a way which is unhelpful here:
> 
> ```
> /* -----------------------------------------------------------------------------
>  * emit_num_required()
>  *
>  * Computes the number of required arguments.  This function is safe for
>  * use with multi-argument typemaps and knows how to skip over everything
>  * properly. Note that parameters with default values are counted unless
>  * the compact default args option is on.
>  * ----------------------------------------------------------------------------- */
> ```
> 
> We could probably expose a version where whether default arguments are counted is controlled by a parameter instead.

See also the (19 years old) discussion of `ParmList_is_compactdefargs` in `Source/Swig/deprecate.c`.
```
  Proposed solution:

     1. "compactdefargs" should be a feature set on function/method nodes
        instead of ParmList objects.  For example, if you have a function,
        you would check the function node to see if the parameters are
        to be handled in this way.


  Difficulties:

     1. This is used by functions in cwrap.c and emit.cxx, none of which
        are passed information about the function/method node.   We might
        have to change the API of those functions to make this work correctly.
        For example:

           int emit_num_required(ParmList *parms)

        might become

           int emit_num_required(ParmList *parms, int compactargs)
```

PR #2810, opened 2024-02-21
[C#] Document how to hide internals with attribute

ojwb:
@wsfulton Is there a better way to achieve this?  This was the best way I could come up with...


ojwb:
@wsfulton Seems an easy merge for 4.3.0 unless you see a better way to recommend.

wsfulton (approved):
I havn't tested this, but the additional example looks exactly like how I would implement what the previous paragraph is suggesting.

ojwb:
Thanks, I'll merge shortly.

FWIW the example was extracted from a working interface file, so should actually work (though changes in SWIG or C# could break it of course, but that's hard to avoid unless we had machinery to actually test code snippets we show in the manual...)

PR #2812, opened 2024-02-22
Check if `__cplusplus` is defined before comparing it

simark:
With SWIG 4.2.0 (and master as of today), I get:

    bt2/native_bt.c:138:5: error: "__cplusplus" is not defined, evaluates to 0 [-Werror=undef]
      138 | #if __cplusplus >=201103L
          |     ^~~~~~~~~~~

Change the pre-processor checks to see if `__cplusplus` is defined before comparing it.


Lib/swig.swg
```diff
@@ -678,7 +678,7 @@ template<typename T> class SwigValueWrapper {
 public:
   SwigValueWrapper() : pointer(0) { }
   SwigValueWrapper& operator=(const T& t) { SwigSmartPointer tmp(new T(t)); pointer = tmp; return *this; }
-#if __cplusplus >=201103L
+#if defined(__cplusplus) && __cplusplus >=201103L
```
ojwb, line 681: This isn't needed - it's already checked at line 665 above, and this is a class definition so clearly C++-only code.

Lib/swig.swg
```diff
@@ -701,7 +701,7 @@ template <typename T> T SwigValueInit() {
   return T();
 }
 
-#if __cplusplus >=201103L
+#if defined(__cplusplus) && __cplusplus >=201103L
```
ojwb, line 704: Also C++-only code so not needed.

PR #2822, opened 2024-03-04
Further improvements in GitHub Windows based action

erezgeva:
<del>Use older MSBuild versions with github win-ci</del>
<del>Check for using [cygwin64](https://cygwin.com/).</del>
Check for using <del>[microsoft/setup-msbuild@v2](https://github.com/marketplace/actions/setup-msbuild)</del> [setup-msvc](https://github.com/marketplace/actions/setup-msvc-developer-command-prompt).
Check for updating PATH in `Prepare Environment` step.


erezgeva:
As installing MSBuild and cygwin require building containers,
 we can simply use AppVeyor for these testings.
The other changes were merge into #2813.

PR #2827, opened 2024-03-05
Check whether the D compiler specified via --with-d2-compiler is functional

rlaboiss:
The D compiler can be specified by the user via the option --with-d2-compiler of configure. The specified program will then be used for compiling the D examples and unit tests (more precisely, in Examples/Makefile.in).

If the user specifies a D compiler that does not exist in the system or is not functional, then the D tests will fail. This commit introduces the check for the specified program in configure.ac, unsetting the D2COMPILER variable if necessary. This ensures that the D tests will not be compiled if the user specifies a non-working D compiler.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -679,11 +679,14 @@ else
 
   if test -z "$D2COMPILERBIN" ; then
     AC_CHECK_PROGS(D2COMPILER, dmd ldmd2 ldc2 gdmd)
-    if test -n "$D2COMPILER" ; then
-      old_ac_ext=$ac_ext
-      ac_ext=d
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks.

(Not really sure why there's a `--with-d2-compiler` option when the autotools-y way to specify it would be this (which should already just work):

```
./configure D2COMPILER=/path/to/dcompiler
```

But this odd pattern seems very well established in SWIG's configure.)

rlaboiss:
Indeed, with D2COMPILER=, it works correctly.
Thanks!

PR #2838, opened 2024-03-17
Do not use mkoctfile -p LIBS for linking - brings in too many deps

opoplawski:
octave 9.1.0 adds a bunch of dependent libraries to LIBS.  While this may be a mistake (see discussion at https://octave.discourse.group/t/change-in-mkoctfile-p-libs-output-with-octave-9-1-0/5381/2), I'm still not sure it's needed at all.

Another possibility would be to make use of the octave pkg-config configuration, which has been available since octave 5.1.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1683,11 +1683,7 @@ if test -n "$OCTAVE"; then
    for var in OCTLIBDIR; do
      OCTAVE_LDFLAGS="${OCTAVE_LDFLAGS} "-L`${mkoctfile} -p ${var}`
    done
-   for var in RDYNAMIC_FLAG RLD_FLAG OCTAVE_LIBS LIBS; do
-     # RLD_FLAG gives "mkoctfile: unknown variable 'RLD_FLAG'" on stderr
-     # with Octave 7.3 so just discard stderr here.  Apparently RLD_FLAG has
-     # reported an empty value since somewhere between 3.4.3 and 3.6.1 so
-     # can be removed below once we require Octave 4.
+   for var in RDYNAMIC_FLAG OCTAVE_LIBS; do
      OCTAVE_LDFLAGS="${OCTAVE_LDFLAGS} "`${mkoctfile} -p ${var} 2>/dev/null`
    done
    AC_MSG_RESULT([$OCTAVE_LDFLAGS])


configure.ac
```diff
@@ -1683,11 +1683,7 @@ if test -n "$OCTAVE"; then
    for var in OCTLIBDIR; do
      OCTAVE_LDFLAGS="${OCTAVE_LDFLAGS} "-L`${mkoctfile} -p ${var}`
    done
-   for var in RDYNAMIC_FLAG RLD_FLAG OCTAVE_LIBS LIBS; do
-     # RLD_FLAG gives "mkoctfile: unknown variable 'RLD_FLAG'" on stderr
-     # with Octave 7.3 so just discard stderr here.  Apparently RLD_FLAG has
-     # reported an empty value since somewhere between 3.4.3 and 3.6.1 so
-     # can be removed below once we require Octave 4.
+   for var in RDYNAMIC_FLAG OCTAVE_LIBS; do
```
ojwb, line 1686: This also drops `RLD_FLAG` which seems like it probably means dropping support for Octave 4.  Maybe that's fine, but even if it is it seems unrelated to the described purpose of your proposed change so would be better proposed separately (and ought to include an update to the manual to note the oldest version now supported).

ojwb:
The libraries probed here should only get used for building the examples and running the testsuite (if they're used for anything else, that's a bug as it would prevent generating Octave bindings on one machine and building them on another), and there shouldn't be a problem explicitly linking with the dependent libraries in this context as far as I can see.  We shouldn't end up linking unused libraries for any binary that gets shipped.

Maybe `LIBS` is not needed at all, but we only have CI for Octave 5.2 and 6.4 currently so confirming this change doesn't break anything would require some manual testing which is quite tedious to do (I manually tested 7.3, 8.4 and 9.0 for the recent changes to support Octave 9).  Or alternatively adding CI coverage for newer Octave versions - I think we used to use versions from a PPA but this stopped getting updates a while ago so currently we only test the packaged Octave in the Ubuntu versions GHA currently provides.  A switch to `pkg-config` would really want similar testing.

configure.ac
ojwb: > probably means dropping support for Octave 4

I think I meant "probably means requiring at least Octave 4".  I don't know much about Octave and what versions are still likely in use, but the oldest we CI test is 5.2 and the current version is 9 so that seems likely to be an entirely reasonable change, but my points about not lumping it in with this change and the docs needing updating still stand.

opoplawski:
I think upstream has fixed this.

Issue #2837, opened 2024-03-16
Support matching in `%rename` on C++11 attributes


mmomtchev:
I am currently wrapping a very large library which makes extensive use of gcc `visibility(hidden)` and `dllexport` to declare some class methods as internal.

I am planning to replace those attributes with the new C++11 form `[[gnu::visibility("hidden")]]`. Since https://github.com/swig/swig/pull/2286 C++11 attributes are ignored directly in the lexer. I plan to make the contents available for matching in `%rename` (`%feature` needs regexps first).

The problem is that the C++11 standard is deliberately very lax on this feature. It says that an attribute can appear everywhere, which will be difficult to implement.

My first suggestion would be to implement this as part of a `storage_class` - as this will completely cover the use case of ignoring symbols based on attribute. Support will be limited to variable attributes and function attributes. It would be missing the much more rarely used type, label, statement and enumerator attributes.

I plan to attach something of this form to every item which has an attribute:

`cpp11_attribute: Hash { 'gnu::visibility': 'hidden' }`

I also plan to use the same syntax as the regex matches:

```
%rename("$ignore", cpp11attr$gnu$visibility="hidden") "";
```

mmomtchev:
To continue ignoring them elsewhere - I will switch the lexer between ignoring and parsing depending on the context. It is somewhat ugly, but it is the only solution I see at the moment.

mmomtchev:
Does anyone know how this works in a C++ compiler? Is it considered a single token?

mmomtchev:
Or - another MacGyver solution - one could select between parsing these as a `storage_type` - or ignoring them completely - in order to not break any code that uses one of the other types.

mmomtchev:
In fact, it is not that difficult to do it properly - with everything except statement attributes.

ojwb:
Seems a plausible approach to me, except `cpp11attr` seems the wrong name as it's no longer C++-specific now since C23 added similar support (https://en.cppreference.com/w/c/language/attributes).  Probably just `attribute` would be better.

mmomtchev:
Ok, I will rename it.

I did most of the cases, but ended with very invasive changes to the grammar:
https://github.com/mmomtchev/swig/compare/main...mmomtchev:swig:cpp11-attributes

[...]

PR #2840, opened 2024-03-19
implement C++11/C23 [[attributes]]

mmomtchev:
This PR adds support for C++11/C23 `[[attributes]]` and allows to use the existing `match$` in `%rename`.

The C++ specifications is deliberately very vague about where this attributes can be placed. I have tried to support all cases that I have seen. All types of attributes are parsed and attached to the AST, but only declaration attributes for variables and functions are available to the end user at the moment.

In order to safeguard against eventual unexpected placement of attributes, `-ignoreattrs` allows to revert to the old behavior - ignoring all attributes directly in the tokenizer.

This PR also switches most of the larger bison codeblocks to use named references in order to make further modifications and especially experimenting with the grammar easier.

The new main grammar item, `attribute` contains `%empty` and has to always precede `storage_class` which also has `%empty`.

The only exception is function argument attributes which use an enumeration without `%empty` in order to avoid shift/reduce conflicts with template parameters which use the same items and do not support attributes.

The only other behavior change is the `%rename` directive where

`$match$attribute$deprecated`

without `=value` matches every occurrence of `deprecated` which can be `[[deprecated]]` but also `[[deprecated("reason")]]`.


Examples/test-suite/php/cpp11_attribute_specifiers_runme.php
```diff
@@ -3,11 +3,26 @@
...
+  '_noDiscard',
+  '_noDiscardDeprecated',
+  'maybeUnused1',
+  'maybeUnused2',
+  '_likely',
+  'test_string_literal',
+  'aligned_var_1_set',
+  'aligned_var_1_get',
+  'aligned_var_2_set',
+  'aligned_var_2_get',
```
ojwb, line 17: These getters and setters should get automatically ignored as a result of `check::globals(array('aligned_var_1', 'aligned_var_2'));`.

Examples/test-suite/php/cpp11_attribute_specifiers_runme.php
```diff
@@ -3,11 +3,26 @@
...
+  '_noReturnNoDiscard',
+  'visible',
+  'test_fn'
+));
 // New classes
 check::classes(array('cpp11_attribute_specifiers','S'));
 // No new vars
-check::globals(array());
```
ojwb, line 10: Comment is now wrong...

ojwb:
I really think we want to do the grammar change to use named references as a **separate commit** (as I said in the issue earlier) - firstly so it's easy to verify that the generated code is unchanged by this refactor, but also to make it actually feasible to review the functional changes to the grammar for this.

mmomtchev:
I see that there is a `callparms` grammar item - which should probably be used instead of a new item (`attribute_arg_list`) with the difference being that `callparms` returns a string, while `attribute_arg_list` is a numbered hash. If I reduce the arguments to a string this will render the syntax for most attributes simpler, but won't allow matching on a single argument when having multiple arguments - which probably is not a needed feature.

Source/Modules/main.cxx
```diff
@@ -965,6 +969,11 @@ int SWIG_main(int argc, char *argv[], const TargetLanguageModule *tlm) {
     }
   }
 
+  if (!cparse_ignore_attrs) {
+    Preprocessor_define("__has_cpp_attribut 201803L", 0);
```
ojwb, line 973: Typo: `__has_cpp_attribute`

But this seems wrongly defined - the expected usage is e.g. `__has_cpp_attribute(noreturn)` so your definition would mean:

```
#if __has_cpp_attribute(noreturn)
```

would expand to:

```
#if 201803L(noreturn)
```

Source/Modules/main.cxx
```diff
@@ -105,6 +105,7 @@ static const char *usage2 = "\
      -I-             - Don't search the current directory\n\
      -I<dir>         - Look for SWIG files in directory <dir>\n\
      -ignoremissing  - Ignore missing include files\n\
+     -ignoreattrs    - Ignore C++11/C23 [[attributes]]\n\
```
ojwb, line 108: Making this a command line option seems unhelpful - it's something the user is likely to need to specify for a particular interface file (because it's something that is likely to be needed based on particular uses of attributes in third party headers they're trying to wrap), so it would be better to be able to specify it in the interface file (if it's a global setting, perhaps as a parameter on `%module`; if it's something that's useful to control in a more fine-grained way, perhaps as a `%feature`.  I can see one might theoretically want it on for one wrapped header but not another, but that may be a bit of a contrived example).

mmomtchev: I did mostly because the specifications for `[[attribute]]` are very vague and I am afraid that now that parsing is implemented and mandatory, there might be cases where the parser wouldn't expect an `[[attribute]]` and this header won't pass at all in SWIG.

Source/Modules/main.cxx
mmomtchev: Yes, I will probably remove it, as this would be too complex to support.

Examples/test-suite/cpp11_attribute_specifiers.i
```diff
@@ -22,9 +27,13 @@
 #pragma warning(disable : 5030) // attribute is not recognized ('likely' and 'unlikely')
 #endif
 
+#ifndef __BIGGEST_ALIGNMENT__
+#define __BIGGEST_ALIGNMENT__ 16
+#endif
```
ojwb, line 32: Defining a macro with a reserved name makes the program ill-formed so it'd be better to use e.g. `BIGGEST_ALIGNMENT`.  If it actually matters to use `__BIGGEST_ALIGNMENT__` when it is defined by the compiler we could define it using

```
#ifdef __BIGGEST_ALIGNMENT__
#define BIGGEST_ALIGNMENT __BIGGEST_ALIGNMENT__
#else
#define BIGGEST_ALIGNMENT 16
#endif
```

and then use `BIGGEST_ALIGNMENT` below.

Examples/test-suite/cpp11_attribute_specifiers.i
```diff
@@ -41,15 +50,33 @@ bool maybeUnused2(bool a, [[maybe_unused]] bool b) { return a; }
...
 // Check that SWIG doesn't choke on ]] when it's not part of an attribute.
-// FIXME: SWIG's parser doesn't handle this case currently.
-int *a;
-int b = a[a[0]];
-#endif
+void test_fn() {
+  int data[1] = { 0 };
+  int *a = data;
+  int b = a[a[0]];
+}
```
ojwb, line 80: The point here is to ensure that SWIG can parse `int b = a[a[0]];` (in particular that the `]]` is broken apart properly.  SWIG doesn't parse function bodies, so mobing this code into a function body doesn't actually test what was intended here.

(The reason this part is currently deactivated is just that the parser doesn't currently actually handle a double array dereference in an expression, but once that is fixed we want to be testing this properly.)

mmomtchev: It will still have to be tokenized if attributes parsing is disabled - this is the only case when this could not work.

Source/Modules/main.cxx
mmomtchev: This is a parser option, it does not belong in the parsed text. Since an attribute cannot contain a `%feature` it can be made to work, but still, I don't think it belongs there.

Examples/test-suite/cpp11_attribute_specifiers.i
mmomtchev: I don't think this is the only anti-pattern in the unit tests, this will only add additional clutter.

Examples/test-suite/cpp11_attribute_specifiers.i
ojwb: The parser just skips tokens between the `{` at the start of a function body and its matching closing `}`, so putting it in a function body will parse OK even if the `]]` is scanned as a single token.

Examples/test-suite/cpp11_attribute_specifiers.i
ojwb: Please change this.

Examples/test-suite/cpp11_attribute_specifiers.i
mmomtchev: The next time you permit yourself to make a comment in order to pass a criminal message, I will seriously reconsider my decision from last year to skip sending an email to the legal department of every institution mentioned in the SWIG license files to inform them of a copyright problem involving plagiarism for a criminal extortion related to a series of falsified legal procedures in the EU. I will be completely honest and I will attest that you most probably do not have anything to do with the organized judicial and police corruption, the sexual elements, the international prostitution ring or the drug trade taking part in this affair. My patience has limits.

erezgeva:
> This PR adds support for C++11/C23 `[[attributes]]` and allows to use the existing `match$` in `%rename`.
> 
> The C++ specifications is deliberately very vague about where this attributes can be placed. I have tried to support all cases that I have seen. All types of attributes are parsed and attached to the AST, but only declaration attributes for variables and functions are available to the end user at the moment.
> 
> In order to safeguard against eventual unexpected placement of attributes, `-ignoreattrs` allows to revert to the old behavior - ignoring all attributes directly in the tokenizer.
> 
> This PR also switches most of the larger bison codeblocks to use named references in order to make further modifications and especially experimenting with the grammar easier.
> 
> The new main grammar item, `attribute` contains `%empty` and has to always precede `storage_class` which also has `%empty`.
> 
> The only exception is function argument attributes which use an enumeration without `%empty` in order to avoid shift/reduce conflicts with template parameters which use the same items and do not support attributes.
> 
> The only other behavior change is the `%rename` directive where
> 
> `$match$attribute$deprecated`
> 
> without `=value` matches every occurrence of `deprecated` which can be `[[deprecated]]` but also `[[deprecated("reason")]]`.
> 
> Res: #2837

Pardon for asking.
I understand in general the need to support C/C++ attributes.
I would like to ask some clarification questions:

1. Do you mean parsing C/C++ code that uses the attributes or add ones to the generated code?
2. I understand C++ 11 is the default, why C23? It seems like a big gap. Perhaps I missed it?
3. What about newer C++ standard attributes optionally? At least as for parsing?

Erez

Examples/test-suite/cpp11_attribute_specifiers.i
wsfulton: @mmomtchev, this last comment appears threatening, is not constructive and is not welcome in the collaborative SWIG development community that we strive for. @ojwb is merely trying to help review and improve your (very welcome) contribution. I don't understand the comment as it appears out of context and I hope it has been posted on the wrong issue, in which case, please withdraw it. If not and if you have some non-technical issue about SWIG that you would like to discuss, please raise a separate discussion thread. Otherwise, I ask you to kindly keep discussions on topic to C++11 attributes. Thank-you for you understanding.

PR #2851, opened 2024-03-28
[C#] Create  a new language cs_il2cpp to generate code used for il2cpp used on android 64 bit.

lijinlong:
calling swig with arg -c++ -cs_il2cpp


wsfulton:
Hi @lijinlong. Please take a good look at https://swig.org/Doc4.2/Extending.html especially the prerequisites section.

It's hard for me to tell at this stage, but I don't understand the motivation for a new variant of the C# target language instead of enhancing the current C# target language. Have you considered this? If you have, please elaborate with plenty of detail as my gut instinct is that a new target language language is not the right way here.

ojwb:
@lijinlong Please can you respond to wsfulton's questions above.

ojwb:
No response for almost a year so closing.

@lijinlong If you still want this to progress, respond to wsfulton's questions above and we can reopen and continue the discussion.

PR #2852, opened 2024-04-01
more robust out of source build detection

mmomtchev:
When `configure` is called with a full path, it will try to do an out-of-source build even if the current working directory is its own directory

https://github.com/conan-io/conan-center-index/pull/19058

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2859,7 +2859,9 @@ AC_CONFIG_FILES([CCache/ccache_swig_config.h])
 # source directory, and provides a rule for updating the Makefile from
 # its original source.
 AC_CONFIG_COMMANDS([Examples],[
-  if test "x${srcdir}" != "x." ; then
+  CONFIG_SRC_DIR=`cd ${srcdir} && pwd`
+  CONFIG_TARGET_DIR=`pwd`
+  if test "x${CONFIG_SRC_DIR}" != "x${CONFIG_TARGET_DIR}" ; then
     AC_MSG_NOTICE([generating Examples build tree])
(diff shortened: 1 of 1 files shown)


wsfulton:
What are the simple steps to reproduce this? 
Surely, doing a caseless compare isn't going to work on normal (non-windows) file systems?

ojwb:
You also don't really want the `[`...`]` with `tr` - just `tr A-Z a-z` would convert ASCII to lower case, but without worrying about whether `[` and `]` are suitably escaped from m4 and the shell.  As written it'll probably go wrong if there are files with single character letter names present, e.g. consider:

```
$ touch a b Y Z
$ echo hello|tr [a-z] [A-Z]
tr: extra operand ‘Y’
Try 'tr --help' for more information.
$ echo tr [a-z] [A-Z]
tr a b Y Z
```

(If no such files are present, it only works as you want because `tr` thinks you're telling it to translate `[` to `[` and `]` to `]`.)

mmomtchev:
Yes, this means that if you have your sources in `path/swig` you won't be able to do an out-of-source build in `path/SWIG`

mmomtchev:
This is with `bash` from `git`:

```
mmom@mmom-desktop MINGW64 /c
$ cd /C

mmom@mmom-desktop MINGW64 /C
$ cd /c

mmom@mmom-desktop MINGW64 /c
$
```

wsfulton:
Sorry to be pedantic here, but what are all the exact steps to reproduce the problem?

mmomtchev:
Launch `<path>/configure` on Windows with `path` being the source directory - `configure` will decide to do an out-of-source build and it will try copying each `Makefile` - the operation will fail because the source and the target being the same.

PR #2867, opened 2024-04-10
Fix mismatched method names in the manual

pfusik:


ojwb:
Thanks, merged.

I also fixed another instance just above in 3496472fe1de0befc8b3d882f68a4a1737085f06.

Issue #2800, opened 2024-02-12
testsuite failures with Ruby 3.1.5 and 3.2.3


ojwb:
The ruby3.2 job is failing in CI.  It's failed two runs in a row, starting c60734add6eefc343db8a59f3b8155a74305eac7 which doesn't seem a likely culprit and CI passed for its parent commit.

I can reproduce it failing locally, though the error I get seems to be in a different testcase:

```
checking ruby testcase import_fragments (with run test)
<internal:/usr/lib/ruby/vendor_ruby/rubygems/core_ext/kernel_require.rb>:151:in `require': machine stack overflow in critical region (fatal)
make[1]: *** [Makefile:78: import_fragments.cpptest] Error 1
```

The most recent change to touch ruby-specific code seems to be 8157a83b59660c8a61e4a3086f62b399bc7e0f5a but that passed all CI jobs.

wsfulton:
Probably we should raise a bug report to Ruby about this one, it's seems like a Ruby problem. The only problem is it is an import example, so requires two separate shared objects to recreate and will need quite a bit of work to provide a decent bug report for them.

wsfulton:
Same crash is now affecting our testing with Ruby 3.1.5 (was okay with 3.1.4) so I've pinned CI testing of 3.1 to use 3.1.4 until resolved in b6f7460.

ojwb:
Still happens with Ruby 3.2.5 (the latest release) and 12528200d06a43f10cbe4b77d7f3e0f345cb63e1.  There is a 3.1.6 release but the latest 3.1.x RVM has seems to be 3.1.5 (which still fails).

wsfulton:
Reported to ruby developers, see https://bugs.ruby-lang.org/issues/21276

wsfulton:
The segfault won't be fixed in the 3.1.x nor 3.2.x series, so I have disabled the import_fragment test from running the code that segfaults ruby for these versions.

PR #2873, opened 2024-04-13
Fix issue with -fvirtual and using

### LINKED ISSUE #2872: Parsing issue when fvirtual is enabled in SWIG 4.1.2
Noticed this issue on a much larger codebase and was able to create a smaller, standalone version. Basically, the issue is that when enabling `-fvirtual` in SWIG 4.2.1, there is an instance where a wrapped Python method should have two options (and thus uses `*args` when passing things to SWIG), but it only has one. Disabling the `-fvirtual` flag causes things to be wrapped correctly. This appears to be a new issue in 4.2.1, as 4.1.1 works as expected. I've included the files I used in my standalone test in the expandable section below. I'll also include them as a `.zip` attached to this issue to make re-creating it easier.

<details><summary> Click to expand files used in the standalone test </summary>

`test.h`
```c++
#ifndef TEST_H
#define TEST_H
class Base
{
public:
    virtual int works(){return _x;};
    virtual void does_not_work(int x) = 0;
protected:
    int _x = 0;
};

class D1 : public virtual Base
{
public:
    using Base::works;
    virtual void works(int x){_x = x;};
    virtual void does_not_work(int x){_x = x;};
};

class D2 : public virtual D1
{
public:
    using D1::does_not_work;
    virtual int does_not_work() const {return _x;};
};
#endif
```

`test.i`
```swig
#ifndef MY_SWIG_EXAMPLE
#define MY_SWIG_EXAMPLE

%module(directors="1") test_Py

%{
#include "test.h"
%}

%include "test.h"

#endif
```

`test.cxx`
```c++
#include "test.h"
```

`test.py`
```python
from test_Py import D2

d2 = D2()

d2.works()
d2.works(4)

d2.does_not_work()

# This call is broken if -fvirtual is included and we use the 
# "using" keyword to include functions
d2.does_not_work(5)
```

Commands run by make (this example with `-fvirtual`):
```
swig -python -c++ -fvirtual -module test_Py -o test_swigwrap_Py.cxx test.i                                                                                                   
g++ -fPIC -o test_swigwrap_Py.o -c test_swigwrap_Py.cxx -I/usr/include/python3.12                                                                                            
g++ -fPIC -o test.o -c test.cxx -I/usr/include/python3.12                                                                                                                    
g++ -shared -o libtest.so test.o                                                                                                                                             
g++ -shared -o _test_Py.so test_swigwrap_Py.o -L. -ltest -I/usr/include/python3.12
```
</details>

Running `LD_LIBRARY_PATH=. python -i test.py` yields an error 
```
  File "/home/leake/Downloads/test2/test.py", line 12, in <module>                                                                                                           
    d2.does_not_work(5)                                                                                                                                                      
TypeError: D2.does_not_work() takes 1 positional argument but 2 were given 
```

Essentially, SWIG does not create wrappers for both versions of `does_not_work`. It only creates a wrapper for the version defined in `D2`, even though `Base` defines another version. 

## Other notes
Here are some other notes that may be useful:
* If I remove `-fvirtual` from the `swig` command, things work as expected.
* If I use `using Base:does_not_work` rather than `D1:does_not_work`, then things work as expected.
* If I use SWIG 4.1.1 rather than 4.2.1 things work as expected.

[test.zip](https://github.com/swig/swig/files/14960769/test.zip)

leakec:
#2872 is for a parsing problem for a corner case where:
* A member is included on a grandchild class (call this `D2` for derived 2) via a `using` directive that points to the parent class (let's call this `D1` for derived 1). The parent does not do any specialization, so it's version of this method is the same as the `base` class (this is the grandparent class of `D2` and the parent class of `D1`).
* The `-fvirtual` flag is used. As a result, when parsing `D1`, the method will pick up the `feature:ignore` flag.
* The `D2` class overloads the method that is including via the `using` directive.

What should happen here is a `_wrap_SWIG_0` for the version of the method included via `using`. `wrap_SWIG_n` wrapped versions of the method for any overloads. However, the `feature:ignore` flag set when parsing `D1` was keeping this from happening, so only the overloaded methods were being wrapped, i.e., the method included via the `using` directive was getting missed. This is all explained in more detail with an example in #2872.

This PR fixes this issue by temporarily unsetting the `feature:ignore` flag if and only if that flag was set as a result of `fvirtual`. The way this is tracked is via a new flag called `ignored_by_fvirtual`, which is set whenever `feature:ignore` is set as a result of `fvirtual`. 

There is also a test case included with #2872 that would be nice to include here so that this corner case can be unit tested going forward. I'm not sure of the best way to add this, e.g., the unit test is written for C++ and Python, is there a good way to adapt it so other languages can run the test case as well? Any feedback/pointers on this would be helpful.

### DIFF
--- a/Examples/test-suite/java/using_member_multiple_inherit_runme.java
+++ b/Examples/test-suite/java/using_member_multiple_inherit_runme.java
@@ -42,5 +42,10 @@ public class using_member_multiple_inherit_runme {
     mb.multmethod("hi");
     mb.multmethod(123, 234);
     mb.multmethod(123, 345, 567);
+
+    // Multiple inheritance, with no override on parent class. 
+    Vusing2 vu = new Vusing2();
+    vu.usingmethod(3);
+    vu.usingmethod("hi");
   }
 }
--- a/Examples/test-suite/python/using_member_multiple_inherit_runme.py
+++ b/Examples/test-suite/python/using_member_multiple_inherit_runme.py
@@ -30,3 +30,8 @@ mb.multmethod(123)
 mb.multmethod("hi")
 mb.multmethod(123, 234)
 mb.multmethod(123, 345, 567)
+
+# Multiple inheritance, with no override on parent class. 
+vu = Vusing2()
+vu.usingmethod(3)
+vu.usingmethod("hi")
--- a/Examples/test-suite/using_member_multiple_inherit.i
+++ b/Examples/test-suite/using_member_multiple_inherit.i
@@ -113,13 +113,16 @@ void cplusplus_testB() {
 class Vusing {
 public:
   virtual void usingmethod(int i) {}
+  virtual ~Vusing() {};
 };
 class Vusing1 : public Vusing {
 public:
   virtual void usingmethod(int i) {}
+  virtual ~Vusing1() {};
 };
 class Vusing2 : public Vusing1 {
 public:
+  virtual ~Vusing2() {};
   using Vusing1::usingmethod;
   virtual void usingmethod(const char* c) {};
 };
--- a/Examples/test-suite/using_member_multiple_inherit.i
+++ b/Examples/test-suite/using_member_multiple_inherit.i
@@ -107,6 +107,23 @@ void cplusplus_testB() {
   m.multmethod(123, 345, 567);
 }
 
+// Multiple inheritance, with no override on parent class. 
+// This checks a corner case with fvirtual and using
+// See #2872
+class Vusing {
+public:
+  virtual void usingmethod(int i) {}
+};
+class Vusing1 : public Vusing {
+public:
+  virtual void usingmethod(int i) {}
+};
+class Vusing2 : public Vusing1 {
+public:
+  using Vusing1::usingmethod;
+  virtual void usingmethod(const char* c) {};
+};
+
 /* TODO: fix when using declaration is declared before method, for example change MultMiddleA to:
 struct MultMiddleB : Mult1, Mult2 {
 protected: // Note!
--- a/Source/Modules/allocate.cxx
+++ b/Source/Modules/allocate.cxx
@@ -234,6 +234,7 @@ class Allocate:public Dispatcher {
 			// Don't eliminate if an overloaded method as this hides the method
 			// in the scripting languages: the dispatch function will hide the base method if ignored.
 			SetFlag(n, "feature:ignore");
+            SetFlag(n, "feature:ignored_by_fvirtual");
 		      }
 	      } else {
 		// Some languages need to know about covariant return types
@@ -633,6 +634,14 @@ class Allocate:public Dispatcher {
  * ----------------------------------------------------------------------------- */
 
   void add_member_for_using_declaration(Node *c, Node *n, int &ccount, Node *&unodes, Node *&last_unodes) {
+    if (GetFlag(c, "feature:ignored_by_fvirtual"))
+    {
+        // This node was ignored by fvirtual. Thus, it has feature:ignore set. 
+        // However, we may have new sybling overrides that will make us want to keep it.
+        // Hence, temporarily unset the feature:ignore flag.
+        UnsetFlag(c, "feature:ignore");
+    }
+
     if (!(Swig_storage_isstatic(c)
 	  || checkAttribute(c, "storage", "typedef")
 	  || Strstr(Getattr(c, "storage"), "friend")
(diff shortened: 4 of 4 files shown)


wsfulton:
Test-suite information: https://swig.org/Doc4.2/Extending.html#Extending_test_suite

I think the reason we haven't found this problem before is because most of the using/method overloading tests have not used virtual methods, so could be a massive category of extra testing that could be added. Just trying to see which testcase to use/enhance...
1. I don't see any director usage in your example, any change with/without directors?
2. Any difference using plain inheritance instead of virtual inheritance?

Some potential testcases to enhance/use:
* using_member_multiple_inherit.i
* using_inherit.i (not more than 2 deep)
* cpp11_using_constructor.i Section: Deeper inheritance chain
* cpp11_director_using_constructor.i

leakec:
@wsfulton 
1. In the example there is a `%module(directors="1") test_Py` line near the top of the `test.i` file, so I believe the test is using directors. I also tried the test without directors. Having directors on vs. off does not seem to make a difference. In other words, the test fails with and without directors before this PR and passes with and without directors afterwards.
2. I do not see any change with virtual vs. non-virtual inheritance. In other words, even with non-virtual inheritance the test from #2872 still fails before this PR and still passes afterwards.

Thanks for the link to the test suit info, will read through it. Took a glance and I think enhancing `using_member_multiple_inherit.i` could cover the test from #2872. What I think needs to be added is a `struct` that will not wrap the method under `fvirtual`, i.e., a struct that inherits the method from the `Using1`, but does not do any overloading (call it `Vusing` for now). Then,  another `struct` needs to inherit from `Vusing` (let's just call it `Vusing2`). It needs to have a `using` statement to get the method from `Vusing` and it should overload the method. Then, the Python/Ruby files need to check that both versions of the method are available in `Vusing2`. 

`using_inherit.i` won't work, since as you mentioned it's only 2 deep. 

It's possible this could be an issue with constructors as well. I can check into this when I next have some time (may be a few days). If it is, then enhancing `cpp11_using_constructor.i` and `cpp11_director_using_constructor.i` may also be needed.

leakec:
@wsfulton Okay, just added the test case. I added to `using_member_multiple_inherit.i`. It looks like Python and Java are the only ones that use this, so I added the appropriate lines to their `runme` files. I also rolled back `allocate.cxx` to ensure that the text breaks before the changes in this PR. Here were the results:
* Running Python with the old version (before the PR) and without `fvirtual` passes. This is as expected since this was only a problem with `fvritual`.
* Running Python with the old version (before the PR) and with `fvirtual` fails. This is as expected since this is the bug #2872 found. 
* Running Python with the new version (after the PR) and without `fvirtual` passes. This is as expected. Good to check I didn't mess anything up with the new test case.
* Running Python with the new version (after the PR) and with `fvirtual` passes. This is as expected and confirms the PR fixes #2872. 
* Running Java passes in all version (before/after the PR and with/without `fvirtual`). I haven't run much with Java before, but apparently this language doesn't have the same issue that Python did.

leakec:
This test change is causing an issue with a specific version of Ruby. Other versions of Ruby are passing. I'm not familiar with Ruby. Is this something anyone can offer advice on?

leakec:
> This test change is causing an issue with a specific version of Ruby. Other versions of Ruby are passing. I'm not familiar with Ruby. Is this something anyone can offer advice on?
> 
> Also, the test that is failing is `import_fragments`, which doesn't appear to use any of the code in the test case I changed. However, maybe there is something I'm missing here.

Pinging @wsfulton as I'm pretty unfamiliar with Ruby or why this test might be failing but only on a specific version of Ruby. Do you have anything to recommend other than installing this specific version of Ruby and trying to get a stack trace?

leakec:
Just pinging this thread so it doesn't buried. @wsfulton

leakec:
Actually, I just realized there is another issue related to ruby testing. According to [this comment](https://github.com/swig/swig/issues/2800#issuecomment-2083391047) Ruby 3.1.5 was having problems, so it was switched to 3.1.4. The checks in this PR ran on April 27th before the change was made to use 3.1.4 on April 29th. Re-running the checks may be enough to fix this.

leakec:
Re-running the checks worked! @wsfulton as far as I know, I think everyone has been done in order to merge. Please let me know if there is anything else that is needed.

wsfulton:
Hi @leakec. Very nice analysis and the fix looks good to me. Thanks very much, merging shortly.

leakec:
@wsfulton Thank you!

PR #2890, opened 2024-04-24
fix wchar.i utf32 string converting

sunnyqeen:


wsfulton:
Can you please describe in a good amount of detail what this fixes along with some examples of what it fixes and which operating systems you've tested on, ideally with an addition to a testcase. Thanks.

sunnyqeen:
@wsfulton This is a bugfix, the buffer is created but the content is not copied.
Please check the same file, line 46
```
    public static string CreateWStringFromUTF32([global::System.Runtime.InteropServices.MarshalAs(global::System.Runtime.InteropServices.UnmanagedType.LPWStr)]global::System.IntPtr cString, int length) {
      if (length == 0)
        return string.Empty;

      byte[] buffer = new byte[length * 4];
      global::System.Runtime.InteropServices.Marshal.Copy(cString, buffer, 0, buffer.Length);
      return global::System.Text.Encoding.UTF32.GetString(buffer);
    }
```

sunnyqeen:
It seems these functions are not used anymore.
same functions from SWIGWStringHelper are used.

ojwb:
@wsfulton This looks good to me - these two indeed seem unused (there are methods of the same name in `SWIGWStringHelper` which are used) - and CI passes with this patch.  I don't think a testcase is possible, since the patch is just removing unused code (in its current version - the initial submission was fixing bugs in that unused code).  OK to merge?

PR #2907, opened 2024-05-14
Fix returning null from functions with output parameters in Python

### LINKED ISSUE #2905: Don't discard null return value in function with output parameters
Consider the following example:

```cpp
%module outparam

%include <std_string.i>

%{
#include <string>
%}

%apply std::string* OUTPUT { std::string* out };

%inline {
std::string hi = "Hello ";

std::string* TestOutput(bool ok, std::string* out) {
  *out = "world";
  return ok ? &hi : nullptr;
}
}
```

and Python code to test it:

```python
import outparam

print(type(outparam.TestOutput(True)), type(outparam.TestOutput(False)))
```

I'd expect this to output

```
<class 'list'> <class 'list'>
```

but it actually outputs

```
<class 'list'> <class 'str'>
```

because `SWIG_Python_AppendOutput()` [replaces null values](https://github.com/swig/swig/blob/b6f7460fda0c754aacefc822759c1d356da9d159/Lib/python/pyrun.swg#L123-L125) instead of appending to them and so when the function returns `nullptr` its output parameter is returned directly, as a string, instead of returning a list.

This is pretty annoying as it means you can't use natural

```python
first, second = outparam.TestOutput(whatever)
```

in Python as this gives a run-time error if the function happens to return null for some value of `whatever`.

This behaviour dates from a very long time ago, see 6e77a6ba2 (fixes for swigrun example, 2005-12-07), but I don't think this was done intentionally or makes sense even if it was, and would like to fix this, but please let me know if you think it is and the current behaviour is actually expected.

vadz:
Null pointers returned from functions also using output parameters or C strings typemaps were lost because the generated code didn't distinguish between Py_None corresponding to them and Py_None indicating the return value of a "void" function.

This doesn't seem to be easy to solve properly because the function return type and output parameters typemaps are independent and there is no way to customize the behaviour of the latter depending on the former.

So we do it using a hack involving a global variable which is always 0 and is passed to SWIG_Python_AppendOutput() from non-void functions, but is shadowed by a local variable with the same name set to 1 in the void ones. This allows us to only discard Py_None if it corresponds to a void function return value.

This is obviously ugly, but requires only minimal changes, doesn't impose much extra run-time overhead and is MT-safe (unlike any solution involving global variables).

Also add a unit test checking for this for Python and for JS, where the same function already worked correctly (unlike in some other languages, including at least Ruby and PHP).

Closes #2905.

### DIFF
--- a/Examples/test-suite/inout.i
+++ b/Examples/test-suite/inout.i
@@ -55,4 +55,9 @@ void AddOne1r(double& INOUT);
   inline void CharNot(char* INOUT) {
     if (*INOUT) { *INOUT = '\0'; } else { *INOUT = '\xff'; }
   }
+
+  inline void* NonVoidOut(int* INOUT) {
+    *INOUT += 42;
+    return NULL;
+  }
 %}
--- a/Examples/test-suite/javascript/inout_runme.js
+++ b/Examples/test-suite/javascript/inout_runme.js
@@ -1,0 +1,13 @@
+var inout = require("inout");
+
+if (inout.AddOne1(1) != 2) {
+    throw new Error;
+}
+
+if (JSON.stringify(inout.AddOne3(1, 1, 1)) != JSON.stringify([2, 2, 2])) {
+    throw new Error;
+}
+
+if (JSON.stringify(inout.NonVoidOut(-42)) != JSON.stringify([null, 0])) {
+    throw new Error;
+}
--- a/Examples/test-suite/python/inout_runme.py
+++ b/Examples/test-suite/python/inout_runme.py
@@ -19,3 +19,9 @@ if a != [(2, 2), 2]:
 a = inout.AddOne3p(1, (1, 1), 1)
 if a != [2, (2, 2), 2]:
     raise RuntimeError
+
+ret, out = inout.NonVoidOut(-42)
+if ret is not None:
+    raise RuntimeError
+if out != 0:
+    raise RuntimeError
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -114,13 +114,16 @@ SWIG_Python_SetConstant(PyObject *d, const char *name, PyObject *obj) {
 
 #endif
 
+/* See "out" typemap for void in pytypemaps.swg */
+SWIGINTERN int swig_is_void = 0;
+
 /* Append a value to the result obj */
 
 SWIGINTERN PyObject*
-SWIG_Python_AppendOutput(PyObject* result, PyObject* obj) {
+SWIG_Python_AppendOutput(PyObject* result, PyObject* obj, int is_void) {
   if (!result) {
     result = obj;
-  } else if (result == Py_None) {
+  } else if (result == Py_None && is_void) {
     SWIG_Py_DECREF(result);
     result = obj;
   } else {
--- a/Lib/python/pytypemaps.swg
+++ b/Lib/python/pytypemaps.swg
@@ -25,7 +25,7 @@
 
 /* Backward compatibility output helper */
 %fragment("t_output_helper","header") %{
-#define t_output_helper SWIG_Python_AppendOutput
+#define t_output_helper(r, o) SWIG_Python_AppendOutput(r, o, swig_is_void)
 %}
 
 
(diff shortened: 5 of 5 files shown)


ojwb:
Won't this result in `-Wshadow` warnings?

vadz:
> Won't this result in `-Wshadow` warnings?

It will (as well as `-Wunused-variable` for the void functions without output params), if the warning is enabled for SWIG-generated code but personally I always disable all warnings in it as there are just too many of them otherwise (especially if you enable `-Wall -Wextra` and a few more globally for your projects, as I do).

And this was the only way to make this work that I found. I don't know if I should feel proud or horrified, but I think this solution is nice, in spite of being so ugly, because it's pretty minimal, both at SWIG and at the generated code level.

But I'll obviously need to wait to see what @wsfulton thinks of it. I hope it can be accepted because it solves a real problem, but if not, it won't be the first patch I'll just maintain in my own fork.

ojwb:
Warnings in generated code do seem of somewhat limited use, though SWIG generated files can contain significant amounts of non-generated code from the user interface file and they can be useful there.  Also it's often much simpler to enable warnings across a whole project.

For whatever reasons, enough people expect SWIG-generated code not to trigger warnings that knowingly introducing such warnings is just setting up for future bug reports we then have to deal with.  Better to address this now while it's fresh in our minds, and we already have `SWIGUNUSED` to add suitable attributes for at least GCC and compilers with compatible attribute syntax (and this could presumably be adjusted to use C++11 attribute syntax with modern compilers).

vadz:
Sorry, I don't know how to avoid `-Wshadow` here, as shadowing is what allows this hack to work, other than by disabling it in the entire file which is hardly going to be very helpful -- but if you really think we should do this, it's trivial to add. I'd like to repeat that IME SWIG-generated code always gives warnings with a high enough warning level/sufficiently many warnings enabled of MSVC/gcc/clang so I always end up using `-w` or equivalent with it anyhow.

ojwb:
Oh true `SWIGUNUSED` doesn't help with `-Wshadow`, but it should avoid the `-Wunused-variable` warnings you mention.

Pretty much any code will generate warnings if you turn on obscure enough warning options, but it seems reasonable to aim for being clean with `-Wall -Wextra`.

vadz:
I couldn't find a way to use `SWIGUNUSED` with a typemap-declared variable, but I can push this:

```diff
diff --git a/Lib/python/pytypemaps.swg b/Lib/python/pytypemaps.swg
index e1f649885..3b14fb232 100644
--- a/Lib/python/pytypemaps.swg
+++ b/Lib/python/pytypemaps.swg
@@ -81,7 +81,8 @@
    than the return value of a void function. By intentionally shadowing the
    global swig_is_void which is always 0 by this local variable set to 1 here,
    we can tell it to do this only for void functions. */
-%typemap(out,noblock=1) void (int swig_is_void = 1) {$result = VOID_Object;}
+%typemap(out,noblock=1) void (int swig_is_void = 1) {(void)swig_is_void;
+  $result = VOID_Object;}
 
 /* Consttab, needed for callbacks, it should be removed later */
 
```

which fixes the warnings given by `-Wall -Wextra` (but not `-Wshadow` if you enable it explicitly, of course).

To avoid rerunning the CI uselessly, I'll wait for William's review before doing it.

vadz:
@wsfulton This is yet another (relatively) simple bug fix correcting a real problem in SWIG and yet there has been no reaction since almost a month. I'd really appreciate if it could be applied or if I could get at least some feedback about it. TIA!

wsfulton:
Hi @vadz, sure, I can look at this one today for you.

wsfulton:
Clever hack, but I can put in a proper fix. I've added 790f7d61c61e2476f4f0ca0ce3e32f23a45910f1 to your [py-nullptr-with-outparam](https://github.com/swig/swig/tree/py-nullptr-with-outparam) branch rebased on master with a quick implementation of it. It's quite similar to the `$null` special variable used for handling void return type differently in the statically typed languages. 


Java.html#Java_special_variables

vadz:
Thanks, this is indeed a much better solution (but I didn't think about it as I wanted to do it entirely at typemaps level, for some reason)!

If you'd like, I can try to clean it up and generalize to the other languages now — but if you'd prefer to do it yourself, I wouldn't mind at all, of course, just please let me know.

Thanks again!

wsfulton:
Thanks, but I can clean it up. I spotted some deprecated stuff and a few quirks in this area while looking at it that I'd like to deal with.

wsfulton:
See commits f3d46986d34d3abb14ed1268b1d0996dca3e2bb2...d9401905f2316aa95376a6128f1e31f8cd5acc1f

Issue #2906, opened 2024-05-13
[ruby] select wrapper function does not work with std::vector and std::map


binaire10:
I got an issue with swig generating a ruby wrapper for `std::map` and `std::vector`. 
The select operation of an `Array` should return a copy of the array with selected value but it segfaults.
https://ruby-doc.org/3.2.2/Array.html#method-i-select

The bug are in those file :
- https://github.com/swig/swig/blame/b6f7460fda0c754aacefc822759c1d356da9d159/Lib/ruby/rubycontainer.swg#L681
- https://github.com/swig/swig/blame/b6f7460fda0c754aacefc822759c1d356da9d159/Lib/ruby/std_map.i#L217

The fixe should be in both files like this :
```diff
- $self->insert( r->end(), *i);
+ r->insert(r->end(), *it);
```

ojwb:
Please open a PR with your proposed change.

ojwb:
We should also have a regression test for this.  If you can add one as part of the PR, that'd be great, but if not a small, self-contained reproducer which can be added as a testcase would be useful.

PR #2912, opened 2024-05-25
[ruby] Fix implementation of container

binaire10:
This fix the misuse of `std::vector::insert( const_iterator pos, const T& value )` describe in issue #2906.

The error come from `std::vector` when the address of insertion is outside the memory of the result object.
That lead to segmentation fault.


ojwb:
Thanks, looks good - merged.

PR #2925, opened 2024-06-13
Python 3.13 support

jschueller:


kloczek:
Is it possible to release new version because this MR? 🤔

jschueller:
it still generates valid code for 3.13 although with possible compilation warnings, so a release would be nice of course but not mandatory (python 3.13 is scheduled for 1st october 2024)

PR #2929, opened 2024-06-20
Allow suppressing progress messages when running test suite

vadz:
Use $(ECHO_PROGRESS) instead of the hard-coded "echo" to allow setting ECHO_PROGRESS=: on make command line to suppress the (many hundreds of) messages given when running every unit test.

This makes it much easier to see any warnings given during the test suite execution.

----

This is just a cosmetic change, but I find it quite useful and it should be completely harmless as it doesn't change anything by default.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -71,6 +71,9 @@ INTERFACEDIR = ../
 SRCDIR     = $(srcdir)/
 SCRIPTDIR  = $(srcdir)
 
+# This can be set to ":" on make command line to suppress progress messages.
+ECHO_PROGRESS := echo
+
 # Regenerate Makefile if Makefile.in or config.status have changed.
 Makefile: $(srcdir)/Makefile.in ../../../config.status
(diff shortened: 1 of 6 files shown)


vadz:
FYI, scilab CI failure is bogus and is due to Ubuntu servers being unavailable.

wsfulton:
Looks okay to me.

wsfulton:
Squash merged in dbdbbe3fdf2c542369181d7315036aa80909c445.

ojwb:
I spotted a neat hack leveraging this feature:

```
make -s ECHO_PROGRESS="printf '\eM% 80s\\r%s\\n' ''" check-test-suite
```

This shows progress but unless there's other output, each progress line just replaces the previous one, so you kind of get the best of both worlds.

(It's not perfect though: if a progress line is > 80 characters, it'll leave a "dropping", which can happen with `partialcheck` at least; it'll also overwrite the last line of any other output from the previous testcase.  Maybe it's possible to do better!)

PR #2931, opened 2024-06-20
Avoid duplicating std::vector<> declaration for JavaScript

vadz:
Use a helper macro abstracting the difference between the generic version and the bool specialization of std::vector<> instead of duplicating the entire class declaration.

No real changes, but this will help with future modifications by allowing to make them only once instead of twice.

### DIFF
--- a/Lib/javascript/v8/std_vector.i
+++ b/Lib/javascript/v8/std_vector.i
@@ -10,14 +10,14 @@
 %}
 
 
-%define SWIG_STD_VECTOR_INTERNAL(CTYPE, REFERENCE, CONST_REFERENCE)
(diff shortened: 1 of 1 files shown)


wsfulton:
You should be able to use the same two argument macro approach used in Lib/java/std_vector.i, which doesn't have REFERENCE, it instead uses `CTYPE &` instead.

vadz:
Sure, we can do this too.

vadz:
Scilab failure is clearly unrelated as this doesn't change any non-JS files, so this can be (squash?) merged IMO.

wsfulton:
Squash merged in 628e356f8b96f7b2a4f6d145834755364b38980e.

PR #2932, opened 2024-06-20
Minor fixes to Boost detection in autoconf

vadz:
I got tired of seeing `-I/include/boost-0` in the generated makefiles and in the compilation command lines and so fixed it in several different ways, so that you could choose what to apply:

1. If you'd like the minimal fix, please apply the second commit, which adds a minimal workaround to SWIG's own configure.
2. If you'd like to fix it at the source, please apply the last commit, which I've also submitted to the upstream.
3. In any case, please apply the first commit, as it's just a minor cleanup removing an unnecessary line.
4. And please also apply third commit just to keep our macro in sync with the upstream version (as of right now).

### DIFF
--- a/Tools/config/ax_boost_base.m4
+++ b/Tools/config/ax_boost_base.m4
@@ -33,7 +33,7 @@
 #   and this notice are preserved. This file is offered as-is, without any
 #   warranty.
 
-#serial 54
+#serial 55
 
 # example boost program (need to pass version)
 m4_define([_AX_BOOST_BASE_PROGRAM],
@@ -289,6 +289,8 @@ AC_DEFUN([_AX_BOOST_BASE_RUNDETECT],[
         else
             AC_MSG_NOTICE([Your boost libraries seems to old (version $_version).])
         fi
+        BOOST_LDFLAGS=""
+        BOOST_CPPFLAGS=""
         # execute ACTION-IF-NOT-FOUND (if present):
         ifelse([$3], , :, [$3])
     else
--- a/Tools/config/ax_boost_base.m4
+++ b/Tools/config/ax_boost_base.m4
@@ -11,9 +11,9 @@
 #   Test for the Boost C++ libraries of a particular version (or newer)
 #
 #   If no path to the installed boost library is given the macro searches
-#   under /usr, /usr/local, /opt and /opt/local and evaluates the
-#   $BOOST_ROOT environment variable. Further documentation is available at
-#   <http://randspringer.de/boost/index.html>.
+#   under /usr, /usr/local, /opt, /opt/local and /opt/homebrew and evaluates
+#   the $BOOST_ROOT environment variable. Further documentation is available
+#   at <http://randspringer.de/boost/index.html>.
 #
 #   This macro calls:
 #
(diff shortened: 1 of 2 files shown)


wsfulton:
So it sounds like `-I/include/boost-0` occurs when boost has not been detected at all. In this case the test-suite fails. However, we really need boost to be found in order for the test-suite to work, so while this is a nice change, it's not making a difference in a properly configured environment, or have I overlooked something?

Do you have the upstream autoconf pull request I can track? Hopefully they merge this soon as ideally we use the upstream version.

vadz:
> So it sounds like `-I/include/boost-0` occurs when boost has not been detected at all.

Yes, this is correct, sorry for not making this clear.

> In this case the test-suite fails. However, we really need boost to be found in order for the test-suite to work, so while this is a nice change, it's not making a difference in a properly configured environment, or have I overlooked something?

I didn't want to fail configure if Boost is not found because it shouldn't prevent building SWIG and 99% of the tests still pass without it, even if the entire test suite does not.

The difference it makes is that I don't see nonsensical `-I/include/boost-0` in the compiler command lines when running the (other) test cases. This is admittedly not much, but it was pretty confusing to see it, and I thought there was some other problem resulting in it appearing. If you really, really don't want to update the autoconf macro, please at least merge the first 2 commits which do absolutely no harm but get rid of this already.

> Do you have the upstream autoconf pull request I can track?

It's in the commit message of the last commit, but here it is once again: https://github.com/autoconf-archive/autoconf-archive/pull/302

wsfulton:
> It's in the commit message of the last commit, but here it is once again: https://github.com/autoconf-archive/autoconf-archive/pull/302

Sorry missed that, am tracking that now and will get back to the SWIG patch in a couple of weeks, when it is hopefully merged upstream.

ojwb:
It's been more than a couple of weeks now...

FWIW, my recent experience is that patches to autoconf-archive aren't getting reviewed promptly (I have an open one ~1 month older than vadz's which also has no feedback yet) so it probably doesn't make sense to block on the change getting accepted there.  The change proposed to the macro is small and seems an obvious fix.

wsfulton:
Maybe chase the autoconf pull request or otherwise I would just go ahead and merge this now. I'd still like to see the final upstream version copied into SWIG when done.

vadz:
I can merge if this is fine with you, I don't see any activity in autoconf-archive/autoconf-archive since 2 months, so I don't think it makes much sense to wait for anything happening there.

ojwb:
This doesn't seem a situation where chasing upstream seems helpful.  My past experience is they do merge clean patches, but sometimes it takes a while for them to get to them.  I know when I'm busy I don't really appreciate impatient pings on patches I just haven't had time to get to yet.

So I've cherry-picked 3 of the 4 commits from here (as bab75232218cfc99690fec67ce23455ddfccdd4e and the two before it), omitting "the second commit, which adds a minimal workaround to SWIG's own configure".

vadz:
Just FYI, my PR in autoconf archive got merged.

ojwb:
> I'd still like to see the final upstream version copied into SWIG when done.

The current upstream version is now identical to what we are already using.

PR #2935, opened 2024-06-23
Fix specialization without primary template

### LINKED ISSUE #2934: Specializing without primary template definition doesn't work any longer
Commit 817c700ce (Template partial specialization improvements, 2023-01-30) broke parsing of the injected class name in the specializations of templates without the primary declaration. To see it, apply the following:

```diff
diff --git a/Examples/test-suite/template_specialization.i b/Examples/test-suite/template_specialization.i
index 00a0cf8ce..bff841684 100644
--- a/Examples/test-suite/template_specialization.i
+++ b/Examples/test-suite/template_specialization.i
@@ -9,9 +9,7 @@
     
     template <class ArgType>
     struct UnaryFunction 
-    {
-      UnaryFunction operator-() const { return *this; }
-    };
+    ;
 
     template <>
     struct UnaryFunction<bool>
```

The test then fails (for any language, but here for C#) with

```
checking csharp testcase template_specialization
template_specialization_wrap.cxx: In function ‘void* CSharp_template_specializationNamespace_UnaryFunction_bool_not1(void*)’:
template_specialization_wrap.cxx:383:24: error: class template argument deduction failed:
  383 |   vfncs::UnaryFunction result;
      |                        ^~~~~~
compilation terminated due to -fmax-errors=1.
make[1]: *** [../../../../Examples/Makefile:235: csharp_cpp] Error 1
make: *** [Makefile:71: template_specialization.cpptest] Error 2
```

because SWIG now generates just `UnaryFunction` instead of the correct `UnaryFunction<bool>`.

vadz:
Due to a regression in 817c700 (Template partial specialization improvements, 2023-01-30), specializations of templates without a primary template definition were not handled correctly any more and, notably, resulted in misparsing the injected name inside such specializations.

Fix this by handling the nodes with template type "classforward" created for the template declaration in the same way as those created for the primary template definition.

Add a trivial test case which works (again) now but doesn't work without the fix in this commit.

Closes #2934.

### DIFF
--- a/Examples/test-suite/template_specialization.i
+++ b/Examples/test-suite/template_specialization.i
@@ -27,6 +27,13 @@
       
     };
     
+    // Also test specialization without the primary template.
+    template <typename T> struct OnlySpecialized;
+
+    template <> struct OnlySpecialized<int>
+    {
+        void bar(const OnlySpecialized& other) { }
+    };
   }
 %}
 
--- a/Source/CParse/templ.c
+++ b/Source/CParse/templ.c
@@ -1362,7 +1362,8 @@ static void expand_defaults(ParmList *expanded_templateparms) {
 ParmList *Swig_cparse_template_parms_expand(ParmList *instantiated_parms, Node *primary, Node *templ) {
   ParmList *expanded_templateparms = CopyParmList(instantiated_parms);
 
-  if (Equal(Getattr(primary, "templatetype"), "class")) {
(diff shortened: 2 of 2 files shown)


wsfulton:
Fix looks sensible to me, thanks.

wsfulton:
Committed in 6b0d355e4dec18385aa50727b86d593d27d98a30, thanks.

PR #2937, opened 2024-06-27
Add reverse length string typemap

erezgeva:
As we already have the reverse LENTH-STRING typemap.

Add testing and add typemap to missing languages.


ojwb (commented):
There also seems to be no test coverage for the new director typemaps (like the existing testing for `STRING, LENGTH` are tested in the `director_binary_string` testcase).

Examples/test-suite/tcl/rev_len_str_runme.tcl
```diff
@@ -0,0 +1,18 @@
...
+
+Test t
+set str "hile"
+if { [ t strlen $str ] != 4 } {
+   puts stderr "bad multi-arg typemap"
+   exit 1
+}
+
+# The universal character Escape Sequence is 2 bytes
+set str2 "hi\u0000"
```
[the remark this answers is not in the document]
ojwb: I didn't trivially find a canonical source to cite for this, but e.g. see point 2 here: https://wiki.tcl-lang.org/page/Tcl+vs+Perl+vs+Python

erezgeva:
> There also seems to be no test coverage for the new director typemaps (like the existing testing for `STRING, LENGTH` are tested in the `director_binary_string` testcase).

[Add tests for reverse length string typemap.](https://github.com/swig/swig/pull/2937/commits/6ad3bef4eb928827b0fd141e9d3d39977dde6eda)

Examples/test-suite/tcl/rev_len_str_runme.tcl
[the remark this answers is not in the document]
erezgeva: It is actually UTF-16 (replacing the older UCS-2).
UTF-16 stand for 16-bit Unicode Transformation Format.
So it is Unicode. And not UTF-8 but UTF-16.
The comment is not misleading.
But if you want a more precise explenation, we can change.

Lib/csharp/csharp.swg
```diff
@@ -1177,6 +1179,49 @@ SWIGEXPORT char *SWIGSTDCALL SWIG_csharp_string_str(SWIG_csharp_string *p) {
...
+/* Length & string reverse order typemap */
+
+%typemap(cstype) (size_t LENGTH, const char *STRING) "string"
+%typemap(ctype)  (size_t LENGTH, const char *STRING) "void *"
+%typemap(imtype) (size_t LENGTH, const char *STRING) "global::System.IntPtr"
+%typemap(csin)   (size_t LENGTH, const char *STRING) "$modulePINVOKE.SWIGStringWithLengthHelper.SWIG_csharp_string_to_c($csinput)"
+%typemap(in, canthrow=1, fragment="SWIG_csharp_string") (size_t LENGTH, const char *STRING) {
+  SWIG_csharp_string *p;
+  p = (SWIG_csharp_string *)$input;
+  if(p != SWIG_NULLPTR) {
```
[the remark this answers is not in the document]
erezgeva: just a habit.
I found only twice.
I fix it.

Lib/typemaps/strings.swg
```diff
@@ -451,20 +451,20 @@
...
-  (size_t LENGTH, Char *STRING) (int res, Char *buf = 0, size_t size = 0, int alloc = 0)
+  (size_t LENGTH, const Char *STRING) (int res, Char *buf = 0, size_t size = 0, int alloc = 0)
 {
   res = SWIG_AsCharPtrAndSize($input, &buf, &size, &alloc);
   if (!SWIG_IsOK(res)) {
     %argument_fail(res,"$type",$symname, $argnum);
   }
   $2 = %reinterpret_cast(buf, $2_ltype);
-  $1 = %numeric_cast(size - 1, $1_ltype);
+  $1 = %numeric_cast((size - 1), $1_ltype);
```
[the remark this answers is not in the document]
erezgeva: Probably I miss it during merge.
Fixed.

Examples/test-suite/tcl/rev_len_str_runme.tcl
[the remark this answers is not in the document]
erezgeva: See [Unicode and UTF-8](https://wiki.tcl-lang.org/page/Unicode+and+UTF-8)

ojwb:
> > There also seems to be no test coverage for the new director typemaps (like the existing testing for `STRING, LENGTH` are tested in the `director_binary_string` testcase).
> 
> [Add tests for reverse length string typemap.](https://github.com/swig/swig/pull/2937/commits/6ad3bef4eb928827b0fd141e9d3d39977dde6eda)

That commit doesn't appear to add test coverage for the new director typemaps though.

erezgeva:
> > > There also seems to be no test coverage for the new director typemaps (like the existing testing for `STRING, LENGTH` are tested in the `director_binary_string` testcase).
> > 
> > 
> > [Add tests for reverse length string typemap.](https://github.com/swig/swig/pull/2937/commits/6ad3bef4eb928827b0fd141e9d3d39977dde6eda)
> 
> That commit doesn't appear to add test coverage for the new director typemaps though.

What do you think is missing?

I did not add unsigned string pointer, it is a C pointer, the actual C type is not really matter.
The `pchar` is not testing the typemap.
And director is a director test, not typemap.

ojwb:
You add new `directorin` typemaps but there's no test coverage for them.

There's not even anything which uses them let alone any runtime tests, so we can't even be confident the code in them compiles - consider this patch which changes the PHP `directorin` to `#error` but `make check-php-test-suite` still passes:

```
diff --git a/Lib/php/php.swg b/Lib/php/php.swg
index 02719f01..ca02067b 100644
--- a/Lib/php/php.swg
+++ b/Lib/php/php.swg
@@ -109,11 +109,7 @@
    $1 = ($1_ltype) Z_STRLEN($input);
 %}
 %typemap(directorin, noblock=1) (size_t LENGTH, const char *STRING) {
-   if (!$2) {
-     ZVAL_EMPTY_STRING($input);
-   } else {
-     ZVAL_STRINGL($input, (const char*)$2, (size_t)$1);
-   }
+%#error BROKEN
 }
 %apply (size_t LENGTH, const char *STRING) { (int LENGTH, const char *STRING) }
 %apply (size_t LENGTH, const char *STRING) { (size_t LENGTH, char *STRING) }
```

We want test coverage that exercises new typemaps so CI can catch a change which breaks a typemap.  As I said above, there's testing of `director_binary_string testcase` for the existing non-reverse equivalents of what you're adding so a reverse version of that would give us coverage as good as we have for the existing string+length typemaps.

Examples/test-suite/rev_len_str.i
```diff
@@ -0,0 +1,17 @@
+/*
+ * A test case for reverse length string typemap
+ */
+
+%module rev_len_str
+
+#if !defined(SWIGC) && !defined(SWIGOCAML) && !defined(SWIGGUILE)
```
wsfulton: Any reason you couldn't include guile? There is a char_binary.i test including run test which works and should be easily ported.

Examples/test-suite/director_rev_len_str.i
```diff
@@ -0,0 +1,37 @@
+%module(directors="1") director_rev_len_str;
+
+%feature("director") Callback;
+
+#if !defined(SWIGC) && !defined(SWIGOCAML) && !defined(SWIGGUILE)
```
wsfulton:
```suggestion
/* There is no concept of a higher level binary string in C */
#if !defined(SWIGC)
```
ocaml is ignored in its Makefile and guile should work, that is, compile without a runtest, like it does for the director_binary_string.i testcase.

Now this matches the recently updated director_binary_string.i in commit 07ce0648149bf8ecff4eada4a45e4c1b805e712e

Examples/test-suite/rev_len_str.i
wsfulton:
```suggestion
/* There is no concept of a higher level binary string in C */
#if !defined(SWIGC)
```
ocaml is ignored in its Makefile and guile should work, that is, compile without a runtest, like it does for the char_strings.i testcase.

Suggested change matches the recently updated char_strings.i in commit 07ce0648149bf8ecff4eada4a45e4c1b805e712e

wsfulton (changes requested):
These are binary string typemaps so same as char *STRING, size_t LENGTH, but order reversed. I think ojwb's comments have all been addressed. I got a couple more...

I am a bit surprised there are problems with Go and D, but let's deal with that afterwards, I might be able to help. Do you have a quick summary of what's not working perhaps?

Could you enhance the Library.html docs with these new multi-arg typemaps please.

Finally can you rename the testcases to match the non-reversed tests:

* non-reversed: char_binary.i and director_char_binary.i
* suggest: char_binary_rev_len.i and director_char_binary_rev_len.i

Finally, this is a comprehensive and very useful contribution, sorry it's been on the review queue for so long, I'll try get it merged as soon as these minor points are addressed. Thanks.

erezgeva:
> These are binary string typemaps so same as char *STRING, size_t LENGTH, but order reversed. I think ojwb's comments have all been addressed. I got a couple more...
> 
> I am a bit surprised there are problems with Go and D, but let's deal with that afterwards, I might be able to help. Do you have a quick summary of what's not working perhaps?

Did initial work:
https://github.com/erezgeva/swig/commit/99ab5d5c9a77e5f6938f386c5c62145fc0c4ad7b
I run the failing tests here:
https://github.com/erezgeva/swig/actions/runs/16036934534

> 
> Could you enhance the Library.html docs with these new multi-arg typemaps please.

Sure, I will update Library.htm.
I do wish to notice, this is **not** a new typemap.
It is already defined in `typemaps/strings.swg`.
I only add languages which do not use `typemaps/strings.swg`.

> 
> Finally can you rename the testcases to match the non-reversed tests:
> 
> * non-reversed: char_binary.i and director_char_binary.i
> * suggest: char_binary_rev_len.i and director_char_binary_rev_len.i

Sure, I will rename.

> 
> Finally, this is a comprehensive and very useful contribution, sorry it's been on the review queue for so long, I'll try get it merged as soon as these minor points are addressed. Thanks.

I am happy on my bid.
Thanks for your work.

Erez

Examples/test-suite/rev_len_str.i
erezgeva: > Any reason you couldn't include guile? There is a char_binary.i test including run test which works and should be easily ported.

Sadly guile fails:
https://github.com/erezgeva/swig/actions/runs/16036934534/job/45250789811

Examples/test-suite/director_rev_len_str.i
erezgeva: update

Examples/test-suite/rev_len_str.i
erezgeva: update as you require

wsfulton:
So sorry, I got the director name wrong, 

director_char_binary_rev_len.i should be director_binary_string_rev_len.i as these are really the reverse of the tests in director_binary_string.i.

One other minor point, can you please keep the lists in alphabetical order in common.mk and Makefile.in.

erezgeva:
> So sorry, I got the director name wrong,
> 
> director_char_binary_rev_len.i should be director_binary_string_rev_len.i as these are really the reverse of the tests in director_binary_string.i.
> 
> One other minor point, can you please keep the lists in alphabetical order in common.mk and Makefile.in.

Update

wsfulton:
Thanks very much. I made some additional minor changes and fixed the guile and ocaml problems. See 07ce0648149bf8ecff4eada4a45e4c1b805e712e...dcf69d3e0ff427417a6fa428a3e792afe7e3d30a.

wsfulton:
I've used your Go changes and fixed them in 5f36c9ab9a2739da22d0a464e86f51e2c7242436. Just D is missing which seems to have a multi-arg typemap lookup problem I need to investigate.

erezgeva:
> Thanks very much. I made some additional minor changes and fixed the guile and ocaml problems. See [07ce064...dcf69d3](https://github.com/swig/swig/compare/07ce0648149bf8ecff4eada4a45e4c1b805e712e...dcf69d3e0ff427417a6fa428a3e792afe7e3d30a).

Thank
I found it too. removing the free of the string.
I try to test why the D is breaking.

erezgeva:
@wsfulton

What I see in D,
is that `%typemap(dtype)` is not used. 

That explain why `ustrlen` in  `char_binary.i` fails.
As the `%typemap(dtype)` state `string` but the  D class  `str` parameter uses `ubyte *` type.

erezgeva:
> I've used your Go changes and fixed them in [5f36c9a](https://github.com/swig/swig/commit/5f36c9ab9a2739da22d0a464e86f51e2c7242436). Just D is missing which seems to have a multi-arg typemap lookup problem I need to investigate.

@wsfulton 
Found the issue with D.
The `dtype` lookup missed the attached flag.
See #3234

I also update the Go test, see #3231

PR #2941, opened 2024-06-30
Avoid warnings for %ignored using declarations

vadz:
It doesn't make sense to give WARN_PARSE_USING_UNDEF if the identifier being brought in scope by a using declaration is explicitly %ignored, but we still did it even in this case.

Just ignore the declaration completely instead in this case.

----

This is mostly cosmetic and I didn't add a test for this as I don't see any simple way to check for an absence of a warning, but the problem could be reproduced with e.g.

```
%module t

%feature("flatnested");

%ignore B::f;

%{
struct B {
  void f() { }
};
%}

%inline {
class C {
protected:
  struct D : B {
    using B::f;
  };
};
}
```

and JS backend: without this change, you'd get warning 315 for the using declaration which doesn't make much sense.

### DIFF
--- a/Source/Modules/allocate.cxx
+++ b/Source/Modules/allocate.cxx
@@ -1000,6 +1000,9 @@ Allocate():
 
   virtual int usingDeclaration(Node *n) {
 
+    if (GetFlag(n, "feature:ignore"))
+      return SWIG_OK;
+
     if (!Getattr(n, "namespace")) {
       Node *ns;
       /* using id */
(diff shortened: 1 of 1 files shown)


ojwb:
> I don't see any simple way to check for an absence of a warning

If it's reproducible with Python (as this one should be) see `Examples/test-suite/errors/`- from its `Makefile.in`:

```
# This test-suite is for checking SWIG errors and warnings and uses
# Python as the target language.
```

See `cpp_namewarn.stderr` and `cpp_namewarn.i` there for a testcase checking for expected warnings.

wsfulton:
Thanks @vadz, I've committed this patch in b2cf9d8c2730eee6ec8ca55e51709a98fb52c5ce and added a testcase in 9a06c17e19e5b025d0f4b9a9c3fabb4ec0dec3c0.

vadz:
Thanks and sorry for not adding the test myself, I got very busy with other stuff in the meanwhile (but I hope to get back to SWIG soon).

PR #2943, opened 2024-06-30
Allow using keywords for members in JS

### LINKED ISSUE #2940: Keyword detection in JS is overeager
When using JS backend (I'm using it with v8 emitter but I think it's the same for all the others), any variables using one of the names from `Lib/javascriptkw.swg` get renamed, but this seems wrong because keywords can be used as property names in JS and so there is no problem with using them for at least the member variables (but actually it looks like keywords can also be used for the function names too, as they're properties of the module object).

In particular, as a side effect of the merge of #2545, `yield` was added to the list of keywords and this broke previously working wrappers as expressions like `something.yield` don't work any more because it got renamed to `_yield`.

I'd like to fix this but I'm not sure how to do it, as `need_name_warning()` can't be customized per language and is called at parse time (via `Swig_name_make()`, `make_name()` and `add_symbols()` finally called by the parser itself when handling `c_declaration`). Any idea?

Trivial example reproducing the problem:

```
%module member

%inline {
struct MyStruct {
  double yield = 0.2;
};
}
```

vadz:
Names of object properties can be keywords in JS, so don't force renaming them to have a leading underscore.

Closes #2940.

----

I'd also move this entire file to `Lib/javascript` but I didn't want to do it in this PR to avoid conflicts later.

Note that this fixes a regression (with members called `yield` and a few more recently added JS keywords) so IMO it is important to apply this.


ojwb:
I agree it ought to be moved.  Also it should be included by `Lib/allkw.swg`.  I can probably take care of those two after this gets merged.

A regression test would be nice - the reproducer in the issue plus checking the attribute is exists under the unrenamed name in a `_runme.js` should do it.

vadz:
I've added the unit test and moved the file.

IMO this can/should be merged.

ojwb:
Thanks, merged.

> Also it should be included by `Lib/allkw.swg`.

Done in 2dc94292a0a041ff12d22a9ab7d8060e85a3021f.

vadz:
> > Also it should be included by `Lib/allkw.swg`.
> 
> Done in [2dc9429](https://github.com/swig/swig/commit/2dc94292a0a041ff12d22a9ab7d8060e85a3021f).

Oops, sorry, forgot about this one, thanks for taking care of it.

mmomtchev:
@vadz Alas, this test is invalid and this PR does not work.

In JavaScript you can always assign a new property to an object. Try this:

```js
var keyword_rename = require("keyword_rename")
keyword_rename._instanceof(1)
keyword_rename._finally(1)
keyword_rename._yield(0)
if (typeof keyword_rename.garga !== 'number') console.log('as expected');
keyword_rename.garga = 42
if (typeof keyword_rename.garga === 'number') console.log('as expected');
if (typeof keyword_rename._yield === 'number') console.log('as expected');
if (typeof keyword_rename.yield !== 'number') throw new Error('yield not included');
keyword_rename.yield = 0;
```

mmomtchev:
In fact, the renaming is for the `S` structure, but it is not included in the test - the renaming in `S` is valid.

vadz:
Sorry, I don't understand what do you mean. There is no renaming for `S`, it's `yield` which used to get automatically renamed, which was wrong and I don't see how the fact that you can add new properties to JS objects changes anything here.

In any case, the current behaviour is correct, so I don't see much point in returning to this PR any more.

mmomtchev:
The test is not valid. It always works.

vadz:
Ah, indeed, thanks, I see. I've created #3015 to fix this.

Issue #2369, opened 2022-09-15
C# std_wstring.i and wchar.i may have problem with UTF-16 to UTF-32


sunnyqeen:
The functions inside to convert UTF-16/UTF-32 use directly casting.
That will not work for Unicode character code which is big that 0x10000
See https://en.wikipedia.org/wiki/UTF-16
We need correct conversion methods, not simple casting.
Example characters:
𐐷	U+10437
𤭢	U+24B62

wsfulton:
You havn't mentioned which version you are using and which platform. The Unicode string handling has been reworked for the pending swig-4.1.0 release, so please try that if you havn't yet.

sunnyqeen:
Hi 
I mean the current code in master.
See the code blow in wchar.i

`static wchar_t * Swig_csharp_UTF16ToWCharPtr(const unsigned short *str) {
  if (sizeof(wchar_t) == 2) {
    return (wchar_t *)str;
  } else {
    wchar_t *result = 0;

    if (str) {
      const unsigned short *pBegin(str);
      const unsigned short *pEnd(pBegin);
      wchar_t *ptr = 0;

      while (*pEnd != 0)
        ++pEnd;

#ifdef __cplusplus
      result = ptr = new wchar_t[pEnd - pBegin + 1];
#else
      result = ptr = (wchar_t *)malloc(sizeof(wchar_t) * (pEnd - pBegin + 1));
#endif
      while(pBegin != pEnd)
        *ptr++ = *pBegin++;
      *ptr++ = 0;
    }

    return result;
  }
}`

emmenlau:
Dear @sunnyqeen that sounds interesting! Can you help with suggestions for code, or provide a PR?

sunnyqeen:
I have written some code before that could be as reference.( it ignores bad encoding character)
There is a lot of code in the internet that do the same thing.

```
int32_t utf16to32(const uint16_t* u16str, int32_t u16len, uint32_t* u32str, int32_t u32len)
{
    int32_t u32index = 0;

    if (u16len < 0)
    {
        u16len = 0x7FFFFFFF;
    }

    for (int32_t i = 0; i < u16len; i++)
    {
        uint16_t uc = u16str[i];
        if (uc <= 0xD7FFu || uc >= 0xE000u && uc <= 0xFFFFu)
        {
            if (u32str && u32index < u32len)
            {
                u32str[u32index] = uc;
            }
            else if (u32str)
            {
                break;
            }
            u32index++;

[...]

PR #2944, opened 2024-07-02
Fix UTF16 to UTF32 string coverting for C#

sunnyqeen:
Code points from U+010000 to U+10FFFF should not cast directly
See https://github.com/swig/swig/issues/2369

### DIFF
--- a/Lib/csharp/std_wstring.i
+++ b/Lib/csharp/std_wstring.i
@@ -31,8 +31,22 @@ static std::wstring Swig_csharp_UTF16ToWString(const unsigned short *str) {
 
     std::wstring result;
     result.reserve(ptr - pBegin);
-    while(pBegin != ptr)
-      result.push_back(*pBegin++);
+    while (pBegin != ptr) {
+      unsigned short uc = *pBegin++;
+      if (uc <= 0xD7FFu || uc >= 0xE000u) {
+        result.push_back(uc);
+      }
+      else if (pBegin != ptr && (uc & 0xFC00u) == 0xD800u) {
+        unsigned short uc1 = *pBegin++;
+        if ((uc1 & 0xFC00u) == 0xDC00u) {
+          unsigned int u32c = uc & 0x03FFu;
+          u32c <<= 10;
+          u32c += uc1 & 0x03FFu;
+          u32c += 0x10000u;
+          result.push_back(u32c);
+        }
+      }
+    }
 
     return result;
   }
--- a/Lib/csharp/wchar.i
+++ b/Lib/csharp/wchar.i
@@ -171,8 +171,21 @@ static wchar_t * Swig_csharp_UTF16ToWCharPtr(const unsigned short *str) {
 #else
       result = ptr = (wchar_t *)malloc(sizeof(wchar_t) * (pEnd - pBegin + 1));
 #endif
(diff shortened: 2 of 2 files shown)


Lib/csharp/std_wstring.i
```diff
@@ -31,8 +31,21 @@ static std::wstring Swig_csharp_UTF16ToWString(const unsigned short *str) {
...
     result.reserve(ptr - pBegin);
-    while(pBegin != ptr)
-      result.push_back(*pBegin++);
+    while (pBegin != ptr) {
+      unsigned short uc = *pBegin++;
+      if (uc <= 0xD7FFu || uc >= 0xE000u) {
+        result.push_back(uc);
+      } else if (pBegin != ptr && (uc & 0xFC00u) == 0xD800u) {
+        unsigned short uc1 = *pBegin++;
+        if ((uc1 & 0xFC00u) == 0xDC00u) {
```
ojwb, line 40: In the case of a lone high surrogate pair codepoint followed by a valid character, this will swallow that valid character.

However, Unicode rules are to treat ill-formed data (which unpaired surrogates are) as an error - see https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf C10 page 12 (or 81 by the page numbers in the document).

Simply stripping them like you're attempting to do here can lead to security holes - e.g. a bad string can be smuggled past a vetting check by inserting an unpaired surrogate in the middle of it.

sunnyqeen: I updated the code to insert replacement character 0xFFFD, when wrong surrogate pair were found.

wsfulton:
@sunnyqeen, do you have a simple C# example that is fixed by this. I can add it as a testcase if you don't, but I would test case coverage for this change.

Lib/csharp/std_wstring.i
wsfulton: @ojwb, please resolve this if you are happy with this commit.

ojwb: Using a replacement character is reasonable for a web browser or similar, but as a general strategy I think following the Unicode rules and flagging this as an error is a better approach (which is what I already suggested above).  

Maybe some users (e.g. who aren't working on data that could be from an attacker or are using the data in a way where there are no security concerns) will want something different, but they can always provide their own typemaps.

(Or maybe we could just delegate the handling of this to the C# runtime, but that only works if it provides an API to do this conversion - I don't know if it does or not.)

sunnyqeen:
@wsfulton I do not have a sample yet, but it's easy to add a test case inside test-suite.
```
//li_std_wstring.h
#define U32_WSTRING L"U32: 𐐷𤭢"

//li_std_wstring.i
bool test_equal_u32(const std::wstring &s) {
  return test_equal(U32_WSTRING, s);
}

//li_std_wstring_runme.cs
        ss = "U32: 𐐷𤭢";
        if (!li_std_wstring.test_equal_u32(ss))
            throw new Exception("Not equal " + ss);
```

Lib/csharp/std_wstring.i
sunnyqeen: Most users expect some useful output, replacement characters at least give a level of the secure concerns.
Error handling may be a bit overkill.

PR #2945, opened 2024-07-02
Restore C# empty() containers method but warn when using it

vadz:
In d8c21410e (Replace empty() method with IsEmpty C# property in STL containers, 2023-11-10) the previously existing empty() method was made private, making it impossible to use it by default and breaking existing code that called it and worked fine with the previous SWIG versions.

Instead of breaking it completely, still define this method but deprecate it using the standard Obsolete() attribute and a message pointing to IsEmpty property.

----

I found another way of making this work in a preferable (IMNSHO) way not requiring introducing an extra `is_empty()` private helper. AFAICS the changes of this PR don't have any drawbacks at all and provide much better migration path since SWIG < 4.2.

### DIFF
--- a/Lib/csharp/std_array.i
+++ b/Lib/csharp/std_array.i
@@ -47,7 +47,7 @@
 
   public bool IsEmpty {
     get {
-      return empty();
+      return $modulePINVOKE.$csclazznameempty(swigCPtr);
     }
   }
 
@@ -237,7 +237,8 @@
 #include <stdexcept>
 %}
 
-%csmethodmodifiers std::array::empty "private"
+%csattributes std::array::empty "[global::System.Obsolete(\"Use IsEmpty property instead\")]"
+
 %csmethodmodifiers std::array::getitemcopy "private"
 %csmethodmodifiers std::array::getitem "private"
 %csmethodmodifiers std::array::setitem "private"
--- a/Lib/csharp/std_map.i
+++ b/Lib/csharp/std_map.i
@@ -50,7 +50,7 @@
 
   public bool IsEmpty {
     get {
-      return empty();
+      return $modulePINVOKE.$csclazznameempty(swigCPtr);
     }
   }
 
@@ -294,7 +294,8 @@
 
 %enddef
 
-%csmethodmodifiers std::map::empty "private"
+%csattributes std::map::empty "[global::System.Obsolete(\"Use IsEmpty property instead\")]"
+
 %csmethodmodifiers std::map::size "private"
 %csmethodmodifiers std::map::getitem "private"
 %csmethodmodifiers std::map::setitem "private"
--- a/Lib/csharp/std_set.i
+++ b/Lib/csharp/std_set.i
@@ -16,7 +16,8 @@
 #include <stdexcept>
 %}
 
-%csmethodmodifiers std::set::empty "private"
+%csattributes std::set::empty "[global::System.Obsolete(\"Use IsEmpty property instead\")]"
+
 %csmethodmodifiers std::set::size "private"
 %csmethodmodifiers std::set::getitem "private"
 %csmethodmodifiers std::set::create_iterator_begin "private"
@@ -47,7 +48,7 @@ class set {
 
   public bool IsEmpty {
     get {
-      return empty();
+      return $modulePINVOKE.$csclazznameempty(swigCPtr);
     }
   }
 
--- a/Lib/csharp/std_unordered_map.i
+++ b/Lib/csharp/std_unordered_map.i
@@ -50,7 +50,7 @@
 
   public bool IsEmpty {
     get {
-      return empty();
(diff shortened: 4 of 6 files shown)


wsfulton:
This pull request has one aspect that is a nice idea and another that misses the mark. But I build on this with a couple of solutions. Sorry for the long response, but I've given it a lot of consideration and it requires some explanation in order to fully understand...

There are two types of changes in this pull request:

1. The `return $modulePINVOKE.$csclazznameempty(swigCPtr);` changes are a good idea to skirt around the problem of IsEmpty calling the empty() function marked as obsolete. Alas, it does not generate the correct code when using %nspace and -namespace. I can't think of a way to correct this, it might need some sort of special new feature, I don't have any good ideas after giving it a bit of thought, but given enough time and effort might be solvable.

2. Make `empty()` public again but obsoleted with `[global::System.Obsolete(\"Use IsEmpty property instead\")]`. This reverts what I put in to a less ideal default interface as `empty()` is public again. The original commit removes it from being public by default, so I won't accept this aspect of the patch. Apart from this change being in the wild for many months, it is also is not a workable strategy for the SWIG tool to **unnecessarily** get into the time frames for obsoleting APIs. Whatever time frame the SWIG tools decides to remove the `empty()` function has no chance of suiting everyone. What will work and is what we always aim for is giving SWIG user's the control over the target language API's they make available to their clients. In this case we provide an opinionated default 'best API' (SWIG does this everywhere) and then provides users with the flexibility to remove/obsolete the function for whatever release time frames is suitable for their project's releases.

So in this spirit, I do think it is absolutely important that SWIG users can choose to restore `empty()` to be public and optionally obsoleted. My suggestion in d8c21410e2b4cb2f04da4c7e2be93a29ae2d756b did not meet this goal as it does not allow for an obsolete `empty()` function without causing unnecessary obsolete warnings in calling `IsEmpty`.

There is one solution that can be used right now:

* Copy std_vector.i into your own project and customise it to make `empty()` Obsolete. Make a few releases of your project with `empty()` being obsoleted, then go back to using SWIG's default std_vector.i in a few month's time/whatever time scale is chosen. This is a design feature of SWIG, where you can put any library file replacement into SWIG's search path to modify the default. Run with the `-v` option to see the search paths.

The second solution is as @vadz suggested (I think):

> I found another way of making this work in a preferable (IMNSHO) way not requiring introducing an extra is_empty() private helper. AFAICS the changes of this PR don't have any drawbacks at all and provide much better migration path since SWIG < 4.2.

Can you show this patch? Maybe the following is what you had in mind, this was the best I could come up:
```patch
diff --git a/Lib/csharp/std_vector.i b/Lib/csharp/std_vector.i
index 982306e0d..15a82f9b2 100644
--- a/Lib/csharp/std_vector.i
+++ b/Lib/csharp/std_vector.i
@@ -71,7 +71,7 @@
 
   public bool IsEmpty {
     get {
-      return empty();
+      return is_empty();
     }
   }
 
@@ -217,6 +217,7 @@
     %rename(Add) push_back;
     void push_back(CTYPE const& x);
     size_type size() const;
+    %rename(is_empty) empty;
     bool empty() const;
     size_type capacity() const;
     void reserve(size_type n);
```

If this was committed, the CHANGES.current entry would need to be something like:
```text

            [C#] The private C# empty() proxy class method for implementing the C# IsEmpty property in the std::vector, std::list, std::map wrappers has been renamed to is_empty(). Prior to swig-4.2.0 this method was public until its replacement by the public IsEmpty property.

            For backwards compatibility, the method can still be made public, like it was prior to swig-4.2.0, by using %extend for vectors. Below shows how to additionally Obsolete it via C#'s Obsolete attribute:

              %extend std::vector {
                %proxycode %{
                [global::System.Obsolete("Use IsEmpty property instead. The empty method will be removed in the next release of myproject.")]
                public bool empty() {
                  return IsEmpty;
                }
                %}
              }
              %include "std_vector.i"

              Anyone making empty() public since swig-4.2.0 using the %csmethodmodifiers suggestion in the 4.2.0 release notes will need to switch to the above solution instead going forwards.
```

The downside of this second solution is that it will silently break any usage of the suggestion in d8c21410e2b4cb2f04da4c7e2be93a29ae2d756b and would require users to use different code for the newer version of SWIG as suggested. So potential for frustration for something not absolutely necessary given the first solution.

My preference is to just use the first solution as it works now and doesn't have the downsides of the second solution. However, I'm open to implementing the second solution if there are any reasonable objections to using the first solution that would outweigh the downsides of the second solution, or anything else that I have overlooked.

vadz:
I don't understand the logic according to which it's OK for SWIG to completely remove a previously existing function but somehow it isn't to mark it as deprecated. I promise you that I really did try to to look at things from your point of view, but I just can't make any sense of this argument...

Anyhow, previous experience shows that arguing is pointless, so I'm just closing this PR and will keep my commit in my fork.

PR #2959, opened 2024-07-16
Java director_thread forget to call stop.

erezgeva:
This cause the test to sometimes to pass and sometimes to fail.
Depends if the thread exit before the main or the other way around.


wsfulton:
Looks okay to me, thanks.

erezgeva:
Hi @wsfulton,

You can add Java version 17 in `windows.yml`.
I comment it only because it failed with `director_thread`.
I guess  Java version 17 with MingW-W64 should work as well.

PR #2962, opened 2024-07-17
Add Java 17 to windows testing

erezgeva:
Java director_thread should work.
We can test  Java 17 with both compilers on Windows.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -49,9 +49,8 @@ jobs:
           VER: 8
         - SWIGLANG: java
           VER: 11
-       #  FIXME: JDK 17 sometimes fail on director_thread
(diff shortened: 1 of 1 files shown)


wsfulton:
Thanks, I've used eb5a599e1c7f25346e2697873c9999640466bd06 instead. Basically, we need to keep a handle on the matrix of versions we test and this is largely covered in Linux, so testing two jdk versions with msvc is ample.

erezgeva:
> Thanks, I've used [eb5a599](https://github.com/swig/swig/commit/eb5a599e1c7f25346e2697873c9999640466bd06) instead. Basically, we need to keep a handle on the matrix of versions we test and this is largely covered in Linux, so testing two jdk versions with msvc is ample.

That does cover too.

PR #2966, opened 2024-07-18
Fix Python docstring for overloads with some Doxygen comments

vadz:
The existing code didn't work correctly if the last element of the overload set didn't have a Doxygen comment: in this case, no docstring was generated at all.

Fix this by trying to find any overload with a Doxygen comment, as Python docstring is common for all of them: add a helper function to do it and use it for both all kinds of ordinary functions (global, member and static) and __init__ functions generated for C++ ctors, as docstring was generated in the same wrong way for all of them in different places.

Note that currently the overloads without Doxygen comments are not documented at all at Python level, as saying "there exist other overloads but there is no documentation for them" doesn't seem very useful and there doesn't seem anything else that we could do.

Add a new unit test for testing all the different combinations of overloaded functions with and without Doxygen comments.

----

This is a simple change which should hopefully be uncontroversial: we need to pass a node with some Doxygen comment to get any docstring out of it, but for overloaded functions (including ctors), we always passed it the node corresponding to the last element of the overload set, which might not have had any Doxygen comment, unlike the other elements.

Before this change, the unit test would fail because having a Doxygen comment for the first function but not the second one didn't generate any docstring at all.

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -730,6 +730,7 @@ DOXYGEN_TEST_CASES += \
 	doxygen_ignore \
 	doxygen_misc_constructs \
 	doxygen_nested_class \
+	doxygen_overloads \
 	doxygen_parsing \
 	doxygen_parsing_enums \
 	doxygen_translate \
--- a/Examples/test-suite/doxygen_overloads.i
+++ b/Examples/test-suite/doxygen_overloads.i
@@ -1,0 +1,95 @@
+%module doxygen_overloads
+
+%inline %{
+
+void overloadWithNoDoc(int) { }
+void overloadWithNoDoc(double) { }
+
+/// Doc for first overload.
+void overloadWithFirstDoc(int) { }
+void overloadWithFirstDoc(double) { }
+
+void overloadWithSecondDoc(int) { }
+/// Doc for second overload.
+void overloadWithSecondDoc(double) { }
+
+/// Doc for both overloads, first.
+void overloadWithBothDocs(int) { }
+/// Doc for both overloads, second.
+void overloadWithBothDocs(double) { }
+
+/// Doc for some overloads, first.
+void overloadWithSomeDocs(int) { }
+void overloadWithSomeDocs(double) { }
+/// Doc for some overloads, third.
+void overloadWithSomeDocs(char) { }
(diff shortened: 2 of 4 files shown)


vadz:
OK, I spoke too soon, there is also the "builtin" case which is completely different and broken in its own way.

I've fixed it too now but I'm much less sure about whether this is the right thing to do because the existing code is completely incomprehensible and I couldn't really find out when is `add_method()` supposed to be called with (non-)null node. And, of course, there are no comments whatsoever... If anybody understand this code better, it would be great if you could please document the requirements of this function and maybe — as long as I'm dreaming anyhow — explain where do different things happen in normal and builtin cases, as they're very different.

vadz:
P.S. I wonder if we want to have this:

```diff
diff --git a/Source/Modules/python.cxx b/Source/Modules/python.cxx
index f96d9220d..43d813a40 100644
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -1502,6 +1502,10 @@ public:
    * node itself which needs to be passes to docstring() later.
    * ------------------------------------------------------------ */
   Node* find_overload_with_docstring(Node* n) {
+    if (Getattr(n, "sym:nextSibling")) {
+      Swig_error(Getfile(n), Getline(n), "Internal error: find_overload_with_docstring() called with a node that is not the last element of an overload set\n");
+    }
+
     for (Node* node_with_doc = n; node_with_doc; node_with_doc = Getattr(node_with_doc, "sym:previousSibling")) {
       if (have_docstring(node_with_doc))
 	return node_with_doc;
```

to check that the function is only called for the expected nodes. I've checked that this is the case currently, but nothing really catastrophic happens even if this is not the case, so I'm not sure if this check is worth it. But I could push it if you think it is.

wsfulton:
Thanks @vadz. Looks okay to me, noting that the documentation states that the comments should be combined, see

https://github.com/swig/swig/blob/8d1ac40dc676e1306b25036e56131d01675ca705/Doc/Manual/Doxygen.html#L1158-L1161

Using your new doxygen_overloads test case showed that previously SWIG would combine the comments but not always and now you've now filled in the cases where it didn't.

The check doesn't seem necessary to me. If anything it should be an assert.

PR #2969, opened 2024-07-22
Use -fPIC on Linux instead of -fpic

matoro:
On most platforms, these are identical.  However, on a small list of platforms, the small -fpic will enforce a limit on GOT size.  However, if -fPIC is used instead, then the GOT on these platforms will have no limits.  The affected platforms and the respective GOT limits with -fpic are:  8k on sparc, 28k on aarch64, and 32k on m68k and powerpc.

See: https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html#index-fpic


ojwb:
`-fPIC` is certainly better as general advice to end users.

I'm not entirely sure about changing `configure.ac` though - that only gets used for the tests which are all small so should work within any `-fpic` limitations.  The code with `-fPIC` is apparently slower than with `-fpic` on architectures where there's a functional difference, though I don't recall seeing anything quantifying the additional overhead so it's hard to know if it's actually a concern here.

matoro:
> `-fPIC` is certainly better as general advice to end users.
> 
> I'm not entirely sure about changing `configure.ac` though - that only gets used for the tests which are all small so should work within any `-fpic` limitations. The code with `-fPIC` is apparently slower than with `-fpic` on architectures where there's a functional difference, though I don't recall seeing anything quantifying the additional overhead so it's hard to know if it's actually a concern here.

Oops sorry, I should have included it here, but this is a direct result of a downstream bug where I encountered exactly this problem:  https://bugs.gentoo.org/900769

So it definitely can and does occur in the test suite on at least one of these platforms, and I confirmed that this change fixes it and the test suite passes afterwards.

ojwb:
> Oops sorry, I should have included it here, but this is a direct result of a downstream bug where I encountered exactly this problem: https://bugs.gentoo.org/900769

OK, well that's a strong argument for changing it.

I guess C++ templates can result in a lot object code for small source code.

ojwb:
I've also updated uses of `-fpic` on other platforms in `configure.ac` and updated example compilation commands in the manual.

Issue #1922, opened 2021-01-11
AIX 7.1 : tcl : missing -ltcl at test link time


trex58:
I've ported Swig 4.0.2 to AIX 7.1 .

When testing with TCL, it appeared that the check link stage was missing the -ltcl option.
I fixed this like I did for Python, by adding  $(TCL_LINK) to the $(TCLCXXSHARED) line of the "tcl:" and "tcl_cpp:" targets in the Examples/Makefile.in .

See the attached patch.
[tcl.patch.txt](https://github.com/swig/swig/files/5796364/tcl.patch.txt)

trex58:
It seems that this additional patch is required too.

On AIX, OpenSource packages are delivered in the /opt/freeware/lib directory. Thus, it must also be looked atwhen searching for libraries.

And "TCLDYNAMICLINKING" from configure.ac is required by the previous patch.

[tcl2.patch.txt](https://github.com/swig/swig/files/5796621/tcl2.patch.txt)

PR #2972, opened 2024-07-26
Follow autoconf recommendation

erezgeva:
Follow recommendations from:
https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Limitations-of-Builtins.html

Use 'test !' instead of '! test'

Use '&&' and '||' instead of of 'test -a' and 'test -o'

Replace '()' with '{ ; }' when using a compound condition.

When calling an M4 macro:
Empty parameters needed only before a non empty parameter.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -857,7 +857,7 @@ else
         AC_MSG_RESULT([$guile_bindir])
         GUILE="$guile_bindir/guile"
       fi
-      if ! test -f "$GUILE" ; then
+      if test ! -f "$GUILE" ; then
         GUILE=
         AC_PATH_PROG(GUILE, guile)
       fi
@@ -1252,12 +1252,12 @@ else
 
   JSCOREVERSION=
 
-  if test -z "$JSCOREINCDIR" -a -n "$JSCORELIB"; then
+  if test -z "$JSCOREINCDIR" && test -n "$JSCORELIB"; then
     AC_MSG_ERROR([Either both or none of --with-jcoreinc --with-jscorelib should be specified])
-  elif test -n "$JSCOREINCDIR" -a -z "$JSCORELIB"; then
+  elif test -n "$JSCOREINCDIR" && test -z "$JSCORELIB"; then
     AC_MSG_ERROR([Either both or none of --with-jcoreinc --with-jscorelib should be specified])
-  elif test -z "$JSCOREINCDIR" -a -z "$JSCORELIB"; then
-    if test -z "$JSCORELIB" -a -n "$PKG_CONFIG "; then
+  elif test -z "$JSCOREINCDIR" && test -z "$JSCORELIB"; then
+    if test -z "$JSCORELIB" && test -n "$PKG_CONFIG"; then
       AC_MSG_CHECKING(for JavaScriptCore/Webkit)
       if $PKG_CONFIG  javascriptcoregtk-4.1 2>/dev/null ; then
         JSCORELIB=`$PKG_CONFIG  --libs javascriptcoregtk-4.1`
@@ -1654,7 +1654,7 @@ if test -n "$OCTAVE"; then
    save_CXXFLAGS="${CXXFLAGS}"
    CXXFLAGS="-Werror -O0"
    AC_COMPILE_IFELSE([
-      AC_LANG_PROGRAM([],[])
+      AC_LANG_PROGRAM
    ],[
       OCTAVE_CXXFLAGS="${OCTAVE_CXXFLAGS} -O0"
    ])
@@ -1904,7 +1904,7 @@ else
     PYEPREFIX=`($PYTHON -c "import sys; sys.stdout.write(sys.exec_prefix)") 2>/dev/null`
     AC_MSG_RESULT($PYEPREFIX)
 
-    if test x"$PYOSNAME" = x"nt" -a x"$PYSEPARATOR" = x"\\"; then
+    if test x"$PYOSNAME" = x"nt" && test x"$PYSEPARATOR" = x"\\"; then
       # Windows installations are quite different to posix installations (MinGW path separator is a forward slash)
       PYPREFIX=`echo "$PYPREFIX" | sed -e 's,\\\\,/,g'` # Forward slashes are easier to use and even work on Windows most of the time
       PYTHON_SO=.pyd
@@ -2025,7 +2025,7 @@ else
     PYVER=0
   fi
   if test "x$PY3BIN" = xyes; then
-    if test x"$PYOSNAME" = x"nt" -a x"$PYSEPARATOR" = x"\\" -a $PYVER -ge 3; then
+    if test x"$PYOSNAME" = x"nt" && test x"$PYSEPARATOR" = x"\\" && test $PYVER -ge 3; then
       PYTHON3="$PYTHON"
     else
       for py_ver in 3 3.15 3.14 3.13 3.12 3.11 3.10 3.9 3.8 3.7 3.6 3.5 3.4 3.3 ""; do
@@ -2060,7 +2060,7 @@ else
     PYSEPARATOR=`($PYTHON3 -c "import sys, os; sys.stdout.write(os.sep)") 2>/dev/null`
     AC_MSG_RESULT($PYSEPARATOR)
 
-    if test x"$PY3OSNAME" = x"nt" -a x"$PYSEPARATOR" = x"\\"; then
+    if test x"$PY3OSNAME" = x"nt" && test x"$PYSEPARATOR" = x"\\"; then
       # Windows installations are quite different to posix installations
       # There is no python-config to use
       AC_MSG_CHECKING(for Python 3.x prefix)
@@ -2311,7 +2311,7 @@ if test -n "$RUBY"; then
 
 		if test "$rb_libruby" != ""; then
 			for i in $dirs; do
-				if (test -r $i/$rb_libruby;) then
+				if test -r "$i/$rb_libruby"; then
 					RUBYLIB="$i"
 					break
 				fi
@@ -2642,7 +2642,7 @@ AC_SUBST(SKIP_JAVA)
 
 
 SKIP_JAVASCRIPT=
-if test -z "$JAVASCRIPT" || ( test -z "$NODEJS" && test -z "$JSCENABLED" && test -z "$JSV8ENABLED" && test -z "$JSNAPIENABLED" ) ; then
+if test -z "$JAVASCRIPT" || { test -z "$NODEJS" && test -z "$JSCENABLED" && test -z "$JSV8ENABLED" && test -z "$JSNAPIENABLED" ; } ; then
     SKIP_JAVASCRIPT="1"
 fi
 AC_SUBST(SKIP_JAVASCRIPT)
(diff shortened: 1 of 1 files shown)


vadz:
Are there any known problems with these constructs on any more or less contemporary systems? The linked page starts its description of `!` limitations with the mention of Unix v7...

erezgeva:
> Are there any known problems with these constructs on any more or less contemporary systems? The linked page starts its description of `!` limitations with the mention of Unix v7...

It is hard to know all the different systems at hand.
I guess that some of the recommendation are based on old system.

But on the other side we keep backward comparability with MSVS 12,
 which Microsoft declare on its end of life and we support 32 bits build on Windows.

Among other, we support HP Unix, AIX and solaris. Not very popular these days.
Did we try all of them? Including the old system.

As we use autoconf, than I think we should follow there guidance.
And try to make our configure coherent.

ojwb:
> Among other, we support HP Unix, AIX and solaris. Not very popular these days. Did we try all of them?

SWIG is a community project and portability improvements to a less used system happens because somebody cares about portability to that system.  The core maintainers don't actively seek out access to platforms to test on that they don't use themselves.  We struggle even to keep on top of issues and PRs without finding ourselves more jobs.

At least for AIX, we've had patches from users fairly recently (e.g. #1922 from 2021: "I've ported Swig 4.0.2 to AIX 7.1").  The conundrum here is that any FOSS project generally only hears from people on particular platforms when it doesn't work, so we may have regular use of SWIG on any or all these platforms.

> Including the old system.

There are different considerations for removing support for an older system which SWIG is known to have worked on at some point and people may be using versus fixing potential portability issues which are only relevant to systems sufficiently old that it seems unlikely anyone is still using them.  If people were trying to build SWIG on, say, UNICOS 10.0.0.6 and running into the limitations of `test -z` in its shell why don't we have a bug report about it?

I don't think we have a formal policy on what platforms we'll support, but I'd suggest a sensible base requirement is that the platform is still supported by its developer.  Based on wikipedia, from your list above HP-UX had a release in May 2024, AIX had a release in November 2023, Solaris has a release this month.  Checking some platforms mentions in the autoconf manual, UNICOS is "Discontinued", AIX 3.2 is from 1992 and no longer maintained, Digital Unix support ended December 2012; Solaris 10 is still supported, though I think there configure will re-exec itself with a more capable shell than `/bin/sh` anyway.

I'm supportive of actual portability improvements (e.g. `-a` -> `&&` makes some sense because POSIX 2008 marked that as obsolete in 2008 so it's arguably less portable going forwards, though all modern bourne-compatible shells support `-a` and it seems inconceivable they'll drop support and unnecessarily break a lot of existing scripts).

The `(`...`)` -> `{`...`; }` change makes sense too, though AIUI it's an optimisation rather than portability as it avoids a subshell.

However I'm unconvinced by the changes here to avoid `-z` and `-n`  in particular - they harm readability of the code, and anyway the problematic cases in some older shells are when the argument looks like an operator, and that's not a concern in most of our uses - e.g. `$JSCOREINCDIR` isn't realistically going to be `!`, `=`, etc or start with `-`.

vadz:
I just don't have a good feeling about making changes not actually fixing any kind of problem and not obviously improving the code quality (in fact, I'd argue that some of those make it worse). There is always a change of breaking something when changing code and if it isn't offset by fixing anything, I don't really see why should these changes be made. But this is just MHO.

erezgeva:
> I just don't have a good feeling about making changes not actually fixing any kind of problem and not obviously improving the code quality (in fact, I'd argue that some of those make it worse). There is always a change of breaking something when changing code and if it isn't offset by fixing anything, I don't really see why should these changes be made. But this is just MHO.

I respect your opinion. Any change can lead to an error.
That is part of development. A good reason to take it slower and more strictly.
But not necessarily to avoid changes.

Following the instructions of GNU Autoconf.
A GNU project that works since 1991.
Does not sound so drastic.
As long as we follow the rules strictly.

Now, you may ask, why bother. And my answer robustness.
We want to increase the portability of configure to works on as many systems there are out there.
Including the many systems we do not test our-self.

In MHO, we are too much divert from Autoconf guidance.
And yet I think, it is safer to take it in smaller steps, so it is easier to ensure we did not fuck thing up.

In my estimation, following Autoconf guidance, we would need a 90% rewrite of the configure, without changing it. Only syntax changes.

Another point for other PRs. Our configure tend to make assumptions in places there are not really needed.
I do know, we must use some assumptions, but only in places where there is no easy alternative.
For example. we assume Python on Windows does not have python-configure and we do not try to probe it. MinGW-w64 python does have python-configure, which does works.

erezgeva:
> > Among other, we support HP Unix, AIX and solaris. Not very popular these days. Did we try all of them?
> 
> SWIG is a community project and portability improvements to a less used system happens because somebody cares about portability to that system. The core maintainers don't actively seek out access to platforms to test on that they don't use themselves. We struggle even to keep on top of issues and PRs without finding ourselves more jobs.
> 
> At least for AIX, we've had patches from users fairly recently (e.g. #1922 from 2021: "I've ported Swig 4.0.2 to AIX 7.1"). The conundrum here is that any FOSS project generally only hears from people on particular platforms when it doesn't work, so we may have regular use of SWIG on any or all these platforms.
> 
> > Including the old system.
> 
> There are different considerations for removing support for an older system which SWIG is known to have worked on at some point and people may be using versus fixing potential portability issues which are only relevant to systems sufficiently old that it seems unlikely anyone is still using them. If people were trying to build SWIG on, say, UNICOS 10.0.0.6 and running into the limitations of `test -z` in its shell why don't we have a bug report about it?
> 
> I don't think we have a formal policy on what platforms we'll support, but I'd suggest a sensible base requirement is that the platform is still supported by its developer. Based on wikipedia, from your list above HP-UX had a release in May 2024, AIX had a release in November 2023, Solaris has a release this month. Checking some platforms mentions in the autoconf manual, UNICOS is "Discontinued", AIX 3.2 is from 1992 and no longer maintained, Digital Unix support ended December 2012; Solaris 10 is still supported, though I think there configure will re-exec itself with a more capable shell than `/bin/sh` anyway.
> 
> I'm supportive of actual portability improvements (e.g. `-a` -> `&&` makes some sense because POSIX 2008 marked that as obsolete in 2008 so it's arguably less portable going forwards, though all modern bourne-compatible shells support `-a` and it seems inconceivable they'll drop support and unnecessarily break a lot of existing scripts).
> 
> The `(`...`)` -> `{`...`; }` change makes sense too, though AIUI it's an optimisation rather than portability as it avoids a subshell.
> 
> However I'm unconvinced by the changes here to avoid `-z` and `-n` in particular - they harm readability of the code, and anyway the problematic cases in some older shells are when the argument looks like an operator, and that's not a concern in most of our uses - e.g. `$JSCOREINCDIR` isn't realistically going to be `!`, `=`, etc or start with `-`.


I will revert to `-z' and `-n`.

erezgeva:
> > I just don't have a good feeling about making changes not actually fixing any kind of problem and not obviously improving the code quality (in fact, I'd argue that some of those make it worse). There is always a change of breaking something when changing code and if it isn't offset by fixing anything, I don't really see why should these changes be made. But this is just MHO.
> 
> I respect your opinion. Any change can lead to an error. That is part of development. A good reason to take it slower and more strictly. But not necessarily to avoid changes.
> 
> Following the instructions of GNU Autoconf. A GNU project that works since 1991. Does not sound so drastic. As long as we follow the rules strictly.
> 
> Now, you may ask, why bother. And my answer robustness. We want to increase the portability of configure to works on as many systems there are out there. Including the many systems we do not test our-self.
> 
> In MHO, we are too much divert from Autoconf guidance. And yet I think, it is safer to take it in smaller steps, so it is easier to ensure we did not fuck thing up.
> 
> In my estimation, following Autoconf guidance, we would need a 90% rewrite of the configure, without changing it. Only syntax changes.
> 
> Another point for other PRs. Our configure tend to make assumptions in places there are not really needed. I do know, we must use some assumptions, but only in places where there is no easy alternative. For example. we assume Python on Windows does not have python-configure and we do not try to probe it. MinGW-w64 python does have python-configure, which does works.

I did find one error though.
`test -n "$PKG_CONFIG "` had a space at its end, which means it would always be true.
You may argue no one needs Webkit without `pkg-config`.
But I simply think it is hard to run all possible scenarios of configure.

configure.ac
```diff
@@ -2686,8 +2686,8 @@ AC_SUBST(SKIP_PERL5)
 
 
 SKIP_PYTHON=
-if (test -z "$PYINCLUDE" || test -z "$PYLINK") &&
-   (test -z "$PY3INCLUDE" || test -z "$PY3LINK") ; then
```
wsfulton, line 2690: What's the point of adding in a redundant semicolon and making a whitespace change (we tend not to put in this kind of whitespace)?

wsfulton: oh sorry, ignore, it's not the same bracket

ojwb:
> Following the instructions of GNU Autoconf.
> A GNU project that works since 1991.
> Does not sound so drastic.
> As long as we follow the rules strictly.

That "since 1991" is essentially the problem with rigidly following all suggestions in the autoconf manual's portability sections - over time increasing amounts of it have become (and continue to become) only relevant to platforms or versions of platforms which are now obsolete.

(I do appreciate it's not easy work for them to update as they really need to carefully check if there's any still current platform versions affected before removing something, and most of the problematic platforms are proprietary; also they have done some updating over the years to remove or indicate parts that you no longer need to worry about.)

erezgeva:
> > Following the instructions of GNU Autoconf.
> > A GNU project that works since 1991.
> > Does not sound so drastic.
> > As long as we follow the rules strictly.
> 
> That "since 1991" is essentially the problem with rigidly following all suggestions in the autoconf manual's portability sections - over time increasing amounts of it have become (and continue to become) only relevant to platforms or versions of platforms which are now obsolete.
> 
> (I do appreciate it's not easy work for them to update as they really need to carefully check if there's any still current platform versions affected before removing something, and most of the problematic platforms are proprietary; also they have done some updating over the years to remove or indicate parts that you no longer need to worry about.)

Autoconf relays on shell, that is where most of portability issues are originated.
Personally, I would not mind if we stay or move to `cmake` or `Mason`
If we stick with Autoconf, we know we can not test all platform, I think rely on someone else work may improve our Autoconf robustness.

Yet I understand we do not have to follow all the recommendation from Autoconf. 
I try to take it easy :-)

ojwb:
Thanks, now merged.

@erezgeva I notice there are still a few uses of `test ... -a ...` in `configure.ac` though - was there a reason to leave those alone?  For example:

```
    if test -z "$JAVA_HOME" -a -x /usr/libexec/java_home; then
```

(That one is specific to `*-*-darwin*` so we can probably assume things about the shell, but the others look like they can get used on any platform.)

erezgeva:
> s specific to `*-*-darwin*` so we can probably assume things about the shell, but the others look like they can get used on any platform.)

I prefer to change less, and be sure my changes works and do not disturb our delicate configuration.

But I believe, we can continue and replace all `test ... -a` and `test ... -o`, step by step.

PR #2976, opened 2024-07-29
Add MinGW-W64 python

erezgeva:
Follow Autoconf way. Probe if possible!
- Fix Windows python probing. MinGW-W64 python does use 'python3-config'.
- Although 'cygpath' comes from CygWin, it can be found on other Windows "bash" based. Use 'Tools/convertpath' only if we do not have 'cygpath'.
- Add 'EXAMPLE_LIBS', add libraries for link on example.


erezgeva:
Hi @wsfulton,

Your changes in `friends_nested` breaks in 
[Environment: SWIGLANG=csharp, VSVER=12](https://ci.appveyor.com/project/swig/swig/builds/50537743/job/b348xn09x3iqtwni)
[Environment: SWIGLANG=csharp, VSVER=14](https://ci.appveyor.com/project/swig/swig/builds/50537743/job/qkf26yp97n5687ce)

configure.ac
```diff
@@ -251,6 +251,22 @@ then
...
+dnl Fall back to convertpath
+AS_IF([test -z "$CYGPATH"],
+      [AC_MSG_CHECKING([alternative for cygpath])
+       AS_VAR_SET([CYGPATH], [`(cd ${srcdir} && pwd)`/Tools/convertpath])
+       AC_SUBST(CYGPATH)
+       AC_MSG_RESULT($CYGPATH)])
+      dnl '--mixed' and '-u' mimic cygpath '-m'
+      dnl '-w' mimic cygpath '-w'
+      dnl '-m' is bizzare. FIXME: why do we have this mode?
+
```
wsfulton, line 269: Can we try remove this and go back to only using Tools/convertpath? That is more consistent as cygpath is not always available and in my experience over many years, keeps changing. As far as I know Tools/convertpath does all we need, it is a very simple implementation of cygpath.

erezgeva: 1. MSYS2 uses the tool `cygpath`, the current assumption that the tool exist with Cygwin only is incorrect. 
   I think it is always better to assume less and probe more.
   I try to follow the autoconf approach.
2. MSYS2 as Cygwin uses different mounting points
    for example `/mingw64/bin` Windows path is `C:\msys64\mingw64`
    The `convertpath`  convert solids to back slash, but do not handle mounting points.
3. Yes, `cygpath` may not be available, therefor we  `Fall back to convertpath`.



I did thought we can  probe in a windows environment.
By doing it like:

````
case $host in
  *-*-cygwin* | *-*-mingw* | *-*-msys*)
    AC_CHECK_TOOL([CYGPATH], [cygpath])
    ;;
esac
dnl Fall back to convertpath
AS_IF([test -z "$CYGPATH"],
...
````

**Comment by wsfulton** [comment]
> Can we try remove this and go back to only using Tools/convertpath? 

Can we try this please. If convertpath has a problem in our setup, I'd like to know what it is and using the same tools where possible helps reduce the crazy dependency matrix we have.

**Comment by erezgeva** [comment]
> > Can we try remove this and go back to only using Tools/convertpath?
> 
> Can we try this please. If convertpath has a problem in our setup, I'd like to know what it is and using the same tools where possible helps reduce the crazy dependency matrix we have.

Looking in `configure.ac`
The places we use `CYGPATH`.

- `JAVA_HOME` we set in `windows.yml`
- `ABS_SRCDIR` is used with `SWIG_LIB`, which we skip when using Win32 API build.
- `ROOT_DIR` is used with JavaScript, which we do not test.
- `SWIG_LIB_WIN_UNIX` we do not really use.

The point is, we can remove the `CYGPATH` and the tests will pass, as we do not use the results in this tests.
But once users will actually need `CYGPATH`, they will get wrong values with `convertpath`.
As `convertpath` do not convert mounting points.

As for C# using `convertpath`, I would not be surprise if users will have issues there.
But as the `convertpath` is used in testing and **NOT** on production, I am less concern.

As for `Fall back to convertpath`.
If we run in environment without mounting points, it will work, else we will need to handle the bridge.

**Comment by erezgeva** [comment]
> convertpath

One point though, I will test the Fallback to `convertpath` works.
I think I did test it, but I will do another test.

**Comment by erezgeva** [comment]
Another mess is the `Tools/convertpath` itself.
I have no idea what is `-m` for, `-m` is suppose to be identical to `--mixed`.
What are the `zap unexpected trailing`.
And why using `-u` when we know `convertpath` do not  handle mounting points.

File: configure.ac (hunk 1 above)
Comment (erezgeva) [reply configure.ac:269]: Hi @wsfulton,

I tested the fallback of using the `Tools/convertpath`.
I stand with my opinion that using `cygpath` when it is available is better than using the `Tools/convertpath`.
If you wish to simplify, there are many other places we can.
I have some suggestion, and I hope you have more.
However with the  `cygpath` tool I only see benefits over using the `Tools/convertpath`.

File: Tools/convertpath [hunk 2]
```diff
@@ -25,7 +25,7 @@ case "$option" in
         *-msys) echo `cmd //c echo "$@" | sed -e "s/[ ]*$//"`;;
         *) echo "The -m option is only supported on MinGW MSYS" 1>&2; exit 1 ;;
       esac ;;
-  -u) shift; echo $@ | sed -e 's,\\,/,g' ;;
+  -u|--mixed) shift; echo $@ | sed -e 's,\\,/,g' ;;
```
wsfulton: This is confusing and just wrong as this is not what cygpath does. The long form should be called `--unix` not `--mixed`.

configure.ac
```diff
@@ -2774,6 +2792,28 @@ AC_SUBST(ROOT_DIR)
...
+    AC_EGREP_CPP([yes],
+    [#ifdef _WIN32
+     yes
+    #endif
+    ], [SWIG_LIB_SET="env SWIG_LIB="
+        AS_VAR_SET([HAVE_WIN32], [true])])
+fi
+AC_SUBST(SWIG_LIB_SET)
+AC_SUBST(HAVE_WIN32)
+
```
wsfulton: Please don't move code unnecessarily. I'm not sure why it now above the comment on the next line when it should be below it.

Why does HAVE_WIN32 need to be set? I can't see any change that uses it.
I don't really understand AS_VAR_SET, does it work with the stated minimum configure.ac?

configure.ac
```diff
@@ -279,6 +279,22 @@ then
 fi
 AC_MSG_RESULT($RPATH)
 
+dnl cygpath: Converts POSIX and Windows format paths
+dnl -m, --mixed    like --windows, but with regular slashes (C:/WINNT)
+dnl -u             (default) print Unix form of NAMEs (/cygdrive/c/winnt)
+dnl -w             print Windows form of NAMEs (C:\WINNT)
+dnl See: https://cygwin.com/cygwin-ug-net/cygpath.html
```
wsfulton: Leave the documentation in the cygpath tool.

configure.ac
```diff
@@ -279,6 +279,22 @@ then
...
+AC_CHECK_TOOL([CYGPATH], [cygpath])
+dnl Fall back to convertpath
+AS_IF([test -z "$CYGPATH"],
+      [AC_MSG_CHECKING([alternative for cygpath])
+       AS_VAR_SET([CYGPATH], [`(cd ${srcdir} && pwd)`/Tools/convertpath])
+       AC_SUBST(CYGPATH)
+       AC_MSG_RESULT($CYGPATH)])
+      dnl '--mixed' and '-u' mimic cygpath '-m'
+      dnl '-w' mimic cygpath '-w'
+      dnl '-m' is bizzare. FIXME: why do we have this mode?
```
wsfulton: What's the point of these comments? -m is --mixed and clearly documented by cygpath, nothing bizzare about it. It's the most compatible option for mixed windows/unix tools use in my experience.

configure.ac
```diff
@@ -2784,27 +2824,18 @@ AC_DEFINE_DIR(SWIG_LIB, swig_lib, [Directory for SWIG system-independent librari
 
 case $build in
   # Windows does not understand unix directories. Convert into a windows directory with drive letter.
-  *-*-mingw*) SWIG_LIB_WIN_UNIX=`${srcdir}/Tools/convertpath -m $SWIG_LIB`;;
-  *-*-cygwin*) SWIG_LIB_WIN_UNIX=`cygpath --mixed "$SWIG_LIB"`;;
+  # FIXME: Why do we need this for?
+  *-*-cygwin* | *-*-mingw* )
```
wsfulton:
```suggestion
  *-*-cygwin* | *-*-mingw* | *-*-msys*)
```

.github/workflows/windows.yml
```diff
@@ -143,6 +141,10 @@ jobs:
             case "$SWIGLANG" in
             python)
               MORE_MSYS_PKGS+=" $mingw_variant-python"
+              # MinGW-w64 with iostream need static link
```
wsfulton: Do you have any reference for this? Seems weird or maybe some workaround??

Tools/convertpath
erezgeva: You are correct.
It is a mess.
But it seems like you start it.
What  is this `The -m option is only supported on MinGW MSYS` all about.

I was just trying to minimise the mess by using `cygpath` where possible.
And leave the `Tools/convertpath` as fall back.

configure.ac
erezgeva: I add  `HAVE_WIN32` as we can use it in the future.
We make today too much decisions based on host type, we should use more probing.
It is true that I did not use it my patch, Rome was not build in a day.
We need to start somewhere.

I move the block before `SWIG_LIB_WIN_UNIX`.
As the value of `SWIG_LIB_WIN_UNIX` should be based on `HAVE_WIN32` value,
in future patches.
After all `SWIG_LIB_WIN_UNIX` is used in code under the same condition `#if defined(_WIN32)`.

I think we should get rid of both `SWIG_LIB_SET` and `SWIG_LIB_WIN_UNIX`.
But I take it step by step. As we are supporting many different environments, we need to be sure they all works before we change.
So I am cautious on.

`AS_VAR_SET` is the recommend way to set variables.
I think it goes a long way back.
I do not plan in changing existing assignments in configure.ac.
But now code in configure.ac, can surly do so.

configure.ac
erezgeva: These are the flags we use in configure.ac.
As `Tools/convertpath` confuse things up.
It is a good reference to how it should behave.
Seems that we have a lot of mess around it, including the `SWIG_LIB_SET` and `SWIG_LIB_WIN_UNIX` flags.

This is a very important reference to sort the mess.
So we are clear on the three path formats:

1. Windows native paths
2. Unix mounting of Windows drives
3. Windows paths using slash instead of backslash.

configure.ac
erezgeva: The bizzare is the fact that `Tools/convertpath` does not work as cygpath.
But produce different result, as it does not handle Unix (MSYS) mounting points.

I understand that there are windows environment lacking the `cygpath` tool.
So we do need `Tools/convertpath`.
But we also need to clear this mess.
So I start by comment it in the configure.ac, so we can mitigate it over time.

configure.ac
erezgeva: That is the point.
When we use MSYS2, we do not need `SWIG_LIB_WIN_UNIX`.
The real question is do we need `SWIG_LIB_WIN_UNIX` at all?

So the minimum is to leave it out of MSYS2.
And try to remove it completely.
But for that we need to test in all Windows environment, which I am not sure how.
Are the tests in AppVeyor covers that?
Do we have more environments?

.github/workflows/windows.yml
erezgeva: Yes, it is a workaround.

It seems an old issue solved when MingW-w64 moves to ucrt.
It was also like that with Ruby before moving to ucrt version of MingW-w64.
Perhaps python will move to ucrt in future versions.

There seems an issue in locating the proper libraries in runtime.
So as a workaround we link these libraries statically.
We could try and find the proper environment in run time, I just do not think it worth our time.

I did not find any official report,
I simply found the workaround in some forums, and once it works, I simple left it as such.

Tools/convertpath
erezgeva: I will move the cygpath changes to a separate PR.

configure.ac
erezgeva: I will move the cygpath changes to a separate PR.

configure.ac
erezgeva: I will move the cygpath changes to a separate PR.

configure.ac
erezgeva: I will move the cygpath changes to a separate PR.

configure.ac
erezgeva: I will move the cygpath changes  to a separate PR.
MSYS2 does not require `SWIG_LIB_WIN_UNIX`.
I doubt if we need `SWIG_LIB_WIN_UNIX` at all.
We can continue the conversation on a PR dedicated to cygpath and other Windows related changes.

.github/workflows/windows.yml
erezgeva: I will add a note that this is a workaround.
For an issue of locating the proper standard libraries,
An issue that will probably be fixed by using the newer MinGW-w64 using UCRT.

erezgeva:
I will move the cygpath changes to a separate PR.
I was wrong to bundle the changes, they do not depend on each other.
I suggest we add python using MinGW-W64 compiler first.

We can continue the conversation on a PR dedicated to `cygpath` and other Windows related changes.

PR #2997, opened 2024-08-17
String literals with zero bytes

### LINKED ISSUE #2996: Sort out handling of string literals with embedded zero bytes
SWIG's "DOH String" can hold a string containing zero bytes, but SWIG is a bit careless in handling literal string values currently and they can get truncated at the zero byte.

This is a fairly uncommon case, but I think it's also pretty close to just working.  Also #2990 broke some cases which previously worked (in exchange for fixing some more common cases which were previously mishandled) and it would be good not to regress these.  I'll try to do this for 4.3.0.

ojwb:
Improve handling of zero bytes in string literals
    
Many of the target languages don't handle this currently.  Those
that don't support this in their strings never will, but others
can probably be fixed (e.g. I've fixed this for SWIG/PHP so far).
    
Fixes #2996

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,13 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.3.0 (in progress)
 ===========================
 
+2024-08-29: olly
+	    #2996 Fix handling of string literals with embedded zero bytes.  This
+	    fix is to the generic machinery - for some target languages SWIG
+	    still truncates such string literals at a zero byte (and for target
+	    languages where the native string can't contain zero bytes it likely
+	    always will).
+
 2024-08-23: wsfulton
             [Java] #2991 Document solutions for mismatch in C++ access specifiers
             and Java access modifiers in an inheritance hierarchy.
--- a/Examples/test-suite/csharp/preproc_constants_runme.cs
+++ b/Examples/test-suite/csharp/preproc_constants_runme.cs
@@ -39,6 +39,9 @@ public class runme {
     assert( typeof(char) == preproc_constants.CONST_CHAR.GetType() );
     assert( typeof(string) == preproc_constants.CONST_STRING1.GetType() );
     assert( typeof(string) == preproc_constants.CONST_STRING2.GetType() );
+    assert( typeof(string) == preproc_constants.CONST_STRING3.GetType() );
+    assert( typeof(string) == preproc_constants.CONST_STRING4.GetType() );
+    assert( preproc_constants.CONST_STRING4 == "zer\0zer\0" );
 
     assert( typeof(int) == preproc_constants.INT_AND_BOOL.GetType() );
     assert( typeof(int) == preproc_constants.INT_AND_CHAR.GetType() );
--- a/Examples/test-suite/java/preproc_constants_runme.java
+++ b/Examples/test-suite/java/preproc_constants_runme.java
@@ -1,0 +1,19 @@
+
+import preproc_constants.*;
+
+public class preproc_constants_runme {
+  static {
+    try {
+        System.loadLibrary("preproc_constants");
+    } catch (UnsatisfiedLinkError e) {
+      System.err.println("Native code library failed to load. See the chapter on Dynamic Linking Problems in the SWIG Java documentation for help.\n" + e);
+      System.exit(1);
+    }
+  }
+
+  public static void main(String argv[]) 
+  {
+    if (preproc_constants.CONST_STRING4 != "zer\0zer\0")
+      throw new RuntimeException("Failed");
+  }
+}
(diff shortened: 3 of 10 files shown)


ojwb:
This fixes:

* the scanner now returns string literals as `String*` rather than `const char*`
* string literal concatenation no longer truncates at zero bytes
* PHP string literals are now created to avoid truncating at zero bytes

Zero bytes now work end to end (at least for `#define CONST_STRING4 "zer\0zer\0"`) for C#, D, Go, Java and PHP.  For C#, D and I think also Java they don't work unless emitted as target language literals (`%csconst`/`%dmanifestconst`/`%javaconst`) because the mechanism to pass them from C/C++ to the target language treats them as `const char*`.

D testing isn't hooked up in the branch as currently pushed.

Go isn't automatically tested but the generated Go wrapper looks good: `const CONST_STRING4 string = "zer\x00zer\x00"`

They don't currently work for Guile, Lua, mzscheme, Ocaml, Octave, Perl, Python, Ruby, Scilab, Tcl.

R doesn't seem to wrap these constants at all so it's hard to tell if they would be mishandled.

I don't think I'm going to attempt to fix all the target languages - if I fix the generic machinery and establish which target languages work (and ensure there's test coverage for those which do) then the target language maintainers (or users who care) can fix target languages which can handle strings with zero bytes.

I think we ought to have tests for such strings in other situations though.

ojwb:
The fix for PHP breaks other cases (which I missed locally despite testing - I guess the C++11 tests didn't get run):

```
cpp11_decltype_wrap.cxx: In function ‘zend_result zm_startup_cpp11_decltype(int, int)’:
cpp11_decltype_wrap.cxx:3426:40: error: initializer fails to determine size of ‘swig_string’
 3426 |   static const char swig_string[] = B::should_be_string;
      |                                     ~~~^~~~~~~~~~~~~~~~
cpp11_decltype_wrap.cxx:3426:40: error: array must be initialized with a brace-enclosed initializer
```

For PHP we need a way to emit different code for the case where the constant value is a string literal.  One current quirk of SWIG doesn't help here - string literals should really have type `const char[]` but SWIG seems to treat them as `const char*` - if that was fixed we could have a different `consttab` typemap for the `char*` and `char[]` cases.  Maybe fixing that is the way to go, but I think I might just revert the PHP fix and leave it as a language where this isn't currently supported.

ojwb:
Rebased onto current master and added a draft CHANGES.current entry. Also:

> D testing isn't hooked up in the branch as currently pushed.

Done.

> I think I might just revert the PHP fix and leave it as a language where this isn't currently supported.

Done.

ojwb:
Python test failures with MSVC don't seem connected to my changes, but the last CI build of git master didn't trigger these:

```
li_std_vector_enum_wrap.cxx(6666): error C2275: 'std::vector<EnumVector::numbers,std::allocator<EnumVector::numbers>>::value_type': expected an expression instead of a type
li_std_vector_enum_wrap.cxx(6666): note: prefix the qualified-id with 'typename' to indicate a type
li_std_vector_enum_wrap.cxx(6666): error C2923: 'swig::traits': 'std::vector<EnumVector::numbers,std::allocator<EnumVector::numbers>>::value_type' is not a valid template type argument for parameter 'Type'
li_std_vector_enum_wrap.cxx(6666): note: see declaration of 'std::vector<EnumVector::numbers,std::allocator<EnumVector::numbers>>::value_type'
li_std_vector_enum_wrap.cxx(6666): error C2955: 'swig::traits': use of class template requires template argument list
li_std_vector_enum_wrap.cxx(3915): note: see declaration of 'swig::traits'
li_std_vector_enum_wrap.cxx(6666): warning C4346: 'swig::traits::category': dependent name is not a type
li_std_vector_enum_wrap.cxx(6666): note: prefix the qualified-id with 'typename' to indicate a type
li_std_vector_enum_wrap.cxx(6666): error C2923: 'swig::container_owner': 'swig::traits::category' is not a valid template type argument for parameter 'T'
li_std_vector_enum_wrap.cxx(6666): note: see declaration of 'swig::traits::category'
li_std_vector_enum_wrap.cxx(6666): error C2955: 'swig::container_owner': use of class template requires template argument list
li_std_vector_enum_wrap.cxx(5011): note: see declaration of 'swig::container_owner'
```

ojwb:
> I think we ought to have tests for such strings in other situations though.

The other relevant situation is default arguments, but actually there a string literal with a zero byte gets truncated in C++ anyway so there isn't anything we can usefully test, consider:

```
std::string f(std::string s = "zer\0zer\0") { return s; }
```

`f()` returns `std::string("zer")`.

ojwb:
The CI failure seems to be due to an MSVC change so I'm going to ignore that and merge this.

PR #3005, opened 2024-08-26
[Go] Use sync.RWMutex for director feature

bardiz12:
Hi!. I've been using swig for golang for sometimes and there's something about the Swig director features that I think we can improve.

```go
// file: /Lib/go/goruntime.swg

var swigDirectorTrack struct {
	sync.Mutex
	m map[int]interface{}
	c int
}

func swigDirectorLookup(c int) interface{} {
	swigDirectorTrack.Lock()
	defer swigDirectorTrack.Unlock()
	ret := swigDirectorTrack.m[c]
	if ret == nil {
		panic("C++ director pointer not found (possible	use-after-free)")
	}
	return ret
}
```

I notice that the swigDirectorTrack variable is using `sync.Mutex` rather than `sync.RWMutex`. I think it would be beneficial in term of performance if we can implement the swigDirectorTrack using sync.RWMutex since it will allow concurrent read. So, this PR aims to change sync.Mutex to sync.RWMutex in the swig's director feature.

### DIFF
--- a/Lib/go/goruntime.swg
+++ b/Lib/go/goruntime.swg
@@ -174,7 +174,7 @@ type _ sync.Mutex
 %insert(go_director) %{
 
 var swigDirectorTrack struct {
-	sync.Mutex
+	sync.RWMutex
 	m map[int]interface{}
 	c int
 }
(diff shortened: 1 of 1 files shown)


ojwb:
@ianlancetaylor Does this proposed change make sense to you, or is there some reason we're not using `RWMutex` already?

ianlancetaylor:
A `sync.RWMutex` is more complex than a `sync.Mutex`.  It's worth using a `sync.RWMutex` when the read lock can be held for more than a brief period of time.  But in this case the read lock is held only very briefly.  For a case like this the more efficient `sync.Mutex` is likely to outweigh the benefit of permitting multiple concurrent readers.  So I would not be inclined to make this change, unless of course there are some benchmarks showing that it would make the code run faster.  Thanks.

ojwb:
Thanks Ian, I'm going to close this then.  If there's benchmarking evidence then please discuss in #3001.

Issue #3012, opened 2024-09-04
Incorrect enum qualification of typedef to anonymous enum


ojwb:
This has been reported multiple times over the years, often conflated with related but separate issues with a superficially similar qualifying problem with C++11 `enum class`.  Opening one issue to bring all these together in one place (and because I think I've fixed the other issue so those other issues will be closed soon).

Seems to be in generic code as it affects most (probably all) target languages.

# https://sourceforge.net/p/swig/bugs/682/

Reproducer:

```
%module x
%inline %{
typedef enum { BAZ1 } baz;
void enumbugexample( baz const & ) { }
%}
```

Then:

```
$ swig -c++ -python simplified.i
$ grep 'enum baz' simplified_wrap.cxx
  enumbugexample((enum baz const &)*arg1);
static swig_type_info _swigt__p_baz = {"_p_baz", "baz *|enum baz *", 0, 0, (void*)0, 0};
```

The `const &` seems to be needed.

I notice the type info entry includes `enum baz *` which seems wrong too.

# https://sourceforge.net/p/swig/bugs/923/

Reproducer:

```
%module test
%inline %{
class VR
{
public:
typedef enum {
Foo
} VRType;
static bool IsValid(const char *vr1, const VRType &vr2);
};
%}
```

```
$ swig -c++ -python test.i
$ g++ -Wall -W `python3-config --cflags` test_wrap.cxx -shared -o demo.so
test_wrap.cxx: In function ‘PyObject* _wrap_VR_IsValid(PyObject*, PyObject*)’:
test_wrap.cxx:3413:59: error: using typedef-name ‘VR::VRType’ after ‘enum’
 3413 |   result = (bool)VR::IsValid((char const *)arg1,(enum VR::VRType const &)*arg2);
      |                                                           ^~~~~~
test_wrap.cxx:3090:3: note: ‘VR::VRType’ has a previous declaration here
 3090 | } VRType;
      |   ^~~~~~
test_wrap.cxx:3413:50: error: expected primary-expression before ‘enum’
 3413 |   result = (bool)VR::IsValid((char const *)arg1,(enum VR::VRType const &)*arg2);
      |                                                  ^~~~
test_wrap.cxx:3413:50: error: expected ‘)’ before ‘enum’
 3413 |   result = (bool)VR::IsValid((char const *)arg1,(enum VR::VRType const &)*arg2);
      |                                                 ~^~~~
      |                                                  )
```

# #197

[...]

PR #3013, opened 2024-09-04
Fix qualify of "enum class" enumerator names

### LINKED ISSUE #197: Incorrect C++ Emitted For Template of Template with Enum Parameter
Take the following declarations:

```
%include <std_vector.i>

enum class Foo
{
    a,
    b,
    c
};

template<Foo F>
class Bar
{   
};

%template(FooVector) std::vector<Bar<Foo::a>>;
```

This will generate code like the following:

``` c++
SWIGINTERN std::vector< Bar< enum Foo::a > >::const_reference std_vector_Sl_Bar_Sl_Foo_a_Sg__Sg__get(std::vector< Bar< Foo::a > > *self,int i){
                int size = int(self->size());
                if (i>=0 && i<size)
                    return (*self)[i];
                else
                    throw std::out_of_range("vector index out of range");
```

The return type is specified as `std::vector<Bar<enum Foo::a>>`. The `enum` keyword before the qualified enum member is invalid. The correct syntax is `std::vector<Bar<Foo::a>>`.

The issue seems to be in `SwigType_typedef_resolve` (typesys.c:650). In the process of fully qualifying types for the template, the qualified enum member is split into base and prefix. The prefix is then fully qualified on its own, yielding `enum Foo` and then joined together with the base, resulting in `enum Foo::a`.

Simple workaround would be to just strip `enum` from the fully qualified typename if if is an enum member that is being resolved. I've applied the following workaround locally and it seems to do the trick:

``` c++
rnameprefix = SwigType_typedef_resolve(nameprefix);
Replaceall(rnameprefix, "enum ", "");
```

### LINKED ISSUE #675: Extra "enum" keyword when inheriting templated class which contains enum class parameter
It can be reproduced with the following interface in both version 3.08 and the in-developing 3.09:

```
%module ReactHandler
enum class DispatchMode
{
    Base,
    Process,
    Remote,
    Proxy
};

template <DispatchMode m>
class DispatcherClientServer
{
};

%template(SoaProcessDispatcherClientServer) DispatcherClientServer<DispatchMode::Process>;

%inline %{
    class SoaProcessDispatcherServer : public DispatcherClientServer<DispatchMode::Process>
    {
    };
%}
```

Swig will generate an extra "enum" keyword, which causes syntax error:

`‘Process’ in ‘enum class DispatchMode’ does not name a type`

The problem code is:

```
static void *_p_SoaProcessDispatcherServerTo_p_DispatcherClientServerT_DispatchMode__Process_t(void *x, int *SWIGUNUSEDPARM(newmemory)) {
    return (void *)((DispatcherClientServer< enum DispatchMode::Process > *)  ((SoaProcessDispatcherServer *) x));
}
```

It compiles if manually change `< enum DispatchMode::Process >` to `< DispatchMode::Process >`

### LINKED ISSUE #1677: Non-compilable wrap code when using static_cast<>
While using SWIG 4.0.1, I may have found a way for which SWIG fails to make wrap code compilable by `g++`. This is related to another issue that I have also posted, but independent from it, so I post this here.

In short, I have this header file, which I called `my_enum.hpp`:
```
#pragma once
enum class my_enum {
     type_0, type_1, type_2, type_3,
     type_4, type_5, type_6, type_7, type_8
};
typedef std::bitset<static_cast<uint32_t>(my_enum::type_8)> type;
// I added the following two functions so as to provide a usage example.
type classify(int a) {
	type k(0);
	k.set(a, true);
	return k;
}
bool has_type(const type& T, const my_enum& t) {
	return T[static_cast<uint32_t>(t)];
}
```
This code is simple enough: a simple `enum class`, a `std::bitset`, and two very small functions. The code above can be used with the following minimal example:
```
#include <iostream>
#include "my_enum.hpp"
int main() {
    std::cout << type().size() << std::endl;
    type t = classify(5);
    std::cout << has_type(t, my_enum::type_8) << std::endl;
}
```
which compiles perfectly with `g++` (9.2.0).

Now, although SWIG (4.0.1) does not crash when making wrap code for `my_enum.i`:
```
%module my_enum
%{
#include "my_enum.hpp"
%}
%include "my_enum.hpp"
```
using the command `swig -Wall -c++ -python -py3 my_enum.i`, the wrap code is not compilable by `g++` (9.2.0). The errors issued by the compiler are:
```
my_enum_wrap.cxx: In function ‘PyObject* _wrap_has_type(PyObject*, PyObject*)’:
my_enum_wrap.cxx:2988:80: error: template argument 1 is invalid
 2988 |   result = (bool)has_type((std::bitset< static_cast< uint32_t > (enum my_enum::type_8) > const &)*arg1,(enum my_enum const &)*arg2);
      |                                                                                ^~~~~~
my_enum_wrap.cxx:2988:88: error: expected primary-expression before ‘>’ token
 2988 |   result = (bool)has_type((std::bitset< static_cast< uint32_t > (enum my_enum::type_8) > const &)*arg1,(enum my_enum const &)*arg2);
      |                                                                                        ^
my_enum_wrap.cxx:2988:90: error: expected primary-expression before ‘const’
 2988 |   result = (bool)has_type((std::bitset< static_cast< uint32_t > (enum my_enum::type_8) > const &)*arg1,(enum my_enum const &)*arg2);
      |                                                                                          ^~~~~
```
The wrap file generated by SWIG (4.0.1) contains the line:
```
result = (bool)has_type((std::bitset< static_cast< uint32_t > (enum __type::type_8) ....;
```
which is what is causing `g++` not to compile the file. If I replace that piece of code (line `2988` in my generated file) with:
```
result = (bool)has_type((std::bitset< static_cast< uint32_t > (__type::type_8) ...
```
(the `enum` keyword inside the `static_cast<uint32_t>()` has been removed) the code compiles perfectly.

I hope to have described the problem clearly and to have provided enough instructions for its reproducibility.

### LINKED ISSUE #2047: const enum class reference fails to compile
# What's the problem ?

If, in a C++ code, there is something like
```c++
class MyClass
{
    public:
        enum class PRINT_SETUP
        {
            TO_CONSOLE,
            TO_CSV
        };
        void my_func(const PRINT_SETUP& e = PRINT_SETUP::TO_CONSOLE);
};
```
SWIG will generate invalid C++ code as follows:
```c++
test_wrap.cxx: In function ‘PyObject* _wrap_MyClass_my_func(PyObject*, PyObject*, PyObject*)’:
test_wrap.cxx:2965:40: error: expected primary-expression before ‘enum’
 2965 |   MyClass::PRINT_SETUP arg2_defvalue = enum MyClass::PRINT_SETUP::TO_CONSOLE ;
      |                                        ^~~~
```

# Workaround

If we just forget about the fact that using `const &` for a `enum class` is debatable, SWIG is not able to parse the statement correctly.

The workaround is simply to remove the reference...

Here is the code (`test.hpp`,`test.cpp`,`test.i`,`test.sh` (to compile)):
[test.zip](https://github.com/swig/swig/files/6782627/test.zip)

ojwb:
Fix incorrect inclusion of "enum " when qualifying C++11 "enum class" enumerator names.

Fixes #197
Fixes #675
Fixes #1677
Fixes #2047

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,10 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.3.0 (in progress)
 ===========================
 
+2024-09-09: olly
+	    #197 #675 #1677 #2047 Fix incorrect inclusion of "enum " when
+	    qualifying C++11 "enum class" enumerator names.
+
 2024-09-07: wsfulton
             #2875 Fix swig-4.1.0 regression using the %interface family of macros
             for multiple inheritance and common bases.
--- a/Examples/test-suite/cpp11_strongly_typed_enumerations.i
+++ b/Examples/test-suite/cpp11_strongly_typed_enumerations.i
@@ -263,3 +263,37 @@ private:
     Val2 = 1172,
   };
 %}
+
+// Regression tests for cases where SWIG incorrectly included "enum " when
+// qualifying the emunerator name:
+
+// #197
+%include <std_vector.i>
+%inline %{
+template<Enum17 E> class Bar {};
+%}
+%template(Enum17Vector) std::vector<Bar<Enum17::Val1>>;
+
+// #675
+%template(Enum17Bar) Bar<Enum17::Val2>;
+%inline %{
+class SubEnum17Bar : public Bar<Enum17::Val2> {};
+%}
+
+// #1677
+%inline %{
+#include <bitset>
+typedef std::bitset<static_cast<unsigned>(Enum17::Val1)> type1677;
+type1677 classify(int a) { return type1677{}; }
+bool has_type(const type1677&, const Enum17&) { return false; }
+%}
+
+// #2047
+%feature("compactdefaultargs") MyClass::my_func;
+%inline %{
+class MyClass {
+  public:
+    enum class PRINT_SETUP { TO_CONSOLE, TO_CSV };
+    void my_func(const PRINT_SETUP& e = PRINT_SETUP::TO_CONSOLE) { (void)e; }
+};
+%}
--- a/Source/Swig/typesys.c
+++ b/Source/Swig/typesys.c
@@ -502,6 +502,9 @@ static Typetab *SwigType_find_scope(Typetab *s, const SwigType *nameprefix) {
     String *qname = Getattr(ss, "qname");
     if (qname) {
       full = NewStringf("%s::%s", qname, nameprefix);
+      if (Strncmp(full, "enum ", 5) == 0) {
+	Delslice(full, 0, 5);
+      }
     } else {
       full = NewString(nameprefix);
     }
@@ -740,9 +743,11 @@ SwigType *SwigType_typedef_resolve(const SwigType *t) {
 	    type = Copy(namebase);
 	    Insert(type, 0, "::");
 	    Insert(type, 0, rnameprefix);
-	    if (strncmp(Char(type), "::", 2) == 0) {
-	      Delitem(type, 0);
-	      Delitem(type, 0);
+	    if (Strncmp(type, "enum ", 5) == 0) {
+	      Delslice(type, 0, 5);
+	    }
+	    if (Strncmp(type, "::", 2) == 0) {
+	      Delslice(type, 0, 2);
 	    }
 	    newtype = 1;
 	  } else {


ojwb:
@wsfulton I'd appreciate you having a look at this.  It's a simple change but this is an area of the code I'm less familiar with.

The key point is we should not include `enum ` when qualifying a C++11 `enum class` enumerator.

ojwb:
One issue I've spotted is that `4` here should be `5` (and similarly in another case):

```
            if (Strncmp(type, "enum ", 4) == 0) {  
              Delslice(type, 0, 4);  
            }   
```

I'll fix these, but if you have any other comments let me know and I'll address those at the same time

ojwb:
I noticed my comment formatting was inconsistent and the CHANGES entry had the wrong month (because I'd copied the previous entry with the wrong month, which I've since fixed on master), and CI failed for unrelated reasons.  Given this hasn't been reviewed yet I've pushed a rebased and squashed version with all these fixed, which will hopefully also demonstrate a clean CI run.

ojwb:
I'm going to merge this now.  My concern motivating wanting a review is really that there might be a better approach, but the change here does fix all the reported cases, and is very limited in the cases it can affect so is really unlikely to break anything which it isn't attempting to fix.  Therefore even if there is a better approach, this seems a suitable fix for 4.3.0 and it makes little sense to leave these cases broken just to get 4.3.0 out.  I would still appreciate a review though.

wsfulton:
My only concern was removing "enum " from the type via DelSlice in `SwigType_typedef_resolve` as removing the "enum " prefix would break C enums. Because the code happens to be in the `if (Swig_scopename_check(base))` block it will only strip the prefix for C++ code as only C++ code uses `::` scoping. The code base relies on enum as prefix for C code, so I'm not sure if it'll help when the type should resolve to the current scope as the prefix would then not be stripped off, so it may be a partial fix. Given the fix is targetting enum classes, not enums, I would want to understand why enum classes have enum in the type in the first place.

Given the test-suite doesn't break and it fixes many cases, on the face of it, it looks okay and is worth using given I just don't have the time to investigate deeply. Thanks for the work in this area.

ojwb:
> I'm not sure if it'll help when the type should resolve to the current scope as the prefix would then not be stripped off, so it may be a partial fix.

I'm not sure I understand what you're thinking of here - if you can give an example of such a case I can try poking around and see if there are still problems.

There are still bugs with handling a `typedef` to an anon `enum` (e.g. `typedef enum { Foo } VRType;`) but while superficially similar the problem there seems to be that we should **never** emit `enum VRType` whereas here we just want to avoid `enum Enum17::Val1` .  Therefore I opened a separate issue: #3012

PR #3014, opened 2024-09-07
Python use MSVC from older GitHub Windows 2019

erezgeva:
Once VC++ issue is resolved we can revert.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -55,12 +55,14 @@ jobs:
         - SWIGLANG: java
           COMPILER: gcc
           VER: 11
-       # Next two are ignored due to new visual c++ not handling containers of enums
+       # Next two are using old VC++ as the new visual c++ not handling containers of enums
        # See https://github.com/swig/swig/issues/3008
-       #- SWIGLANG: python
-       #  VER: '3.7'
-       #- SWIGLANG: python
-       #  VER: '3.12'
+        - SWIGLANG: python
+          VER: '3.7'
+          os: 'windows-2019'
+        - SWIGLANG: python
+          VER: '3.12'
+          os: 'windows-2019'
        #  TODO require fixing of probing in configure.ac
        #- SWIGLANG: python
        #  INSTALL: 'true'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,15 +21,17 @@ environment:
     VSVER: 16
     VER: 39
     APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
-  - SWIGLANG: python
-    VSVER: 17
-    VER: 310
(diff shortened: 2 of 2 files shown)


erezgeva:
The same errors appears in AppVeyor

[SWIGLANG=python, VSVER=17, VER=310, APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2022](https://ci.appveyor.com/project/swig/swig/builds/50555937/job/kb644va9x4s3sb6y)
Failed in 21 min 59 sec
[SWIGLANG=python, VSVER=17, VER=310, Platform=x86, APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2022](https://ci.appveyor.com/project/swig/swig/builds/50555937/job/r5bc2e57n1ylcadq)
Failed in 24 min 37 sec

wsfulton:
committed in dda6f691e58472ce2fad7278c1e8436030977024.

Surely there must be a way to use the older docker images that https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md provides. Docker images don't usually just disappear! @erezgeva, perhaps try and search or ask on some support forums.

erezgeva:
> committed in [dda6f69](https://github.com/swig/swig/commit/dda6f691e58472ce2fad7278c1e8436030977024).
> 
> Surely there must be a way to use the older docker images that https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md provides. Docker images don't usually just disappear! @erezgeva, perhaps try and search or ask on some support forums.

I have no idea if it possible.

Seems AppVeyor upgrade their Windows in a similar way.

erezgeva:
> committed in [dda6f69](https://github.com/swig/swig/commit/dda6f691e58472ce2fad7278c1e8436030977024).
> 
> Surely there must be a way to use the older docker images that https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md provides. Docker images don't usually just disappear! @erezgeva, perhaps try and search or ask on some support forums.

The image runs as virtual machine, not as a Docker container image.
I did search if I can create a Docker container image out of the runner image, and fail to find any reference on it.

wsfulton:
Could you ask Github support how one can use an older version if (and when in this case) there is a bug in the toolsets provided (visual c++) in a new VM? Uninstalling and re-installing a different version of visual c++ on the fly doesn't seem like a viable workaround option as you pointed out.

erezgeva:
> Could you ask Github support how one can use an older version if (and when in this case) there is a bug in the toolsets provided (visual c++) in a new VM? Uninstalling and re-installing a different version of visual c++ on the fly doesn't seem like a viable workaround option as you pointed out.

I open a bug report:
https://github.com/actions/runner-images/issues/10575

From the questions, it does not seems they provide an option to use older images.
I think the way to avoid such is to create our own Docker images.

PR #3018, opened 2024-09-12
Straighten out handling of integer constants

ojwb:
This provides a generic framework to aid converting C/C++ integer and boolean literals to target language literals, replacing custom code in several target language backends.

This fixes some bugs in that code.

TODO: Add regression tests for these bugs.

### DIFF
--- a/Examples/test-suite/d/enum_thorough_runme.2.d
+++ b/Examples/test-suite/d/enum_thorough_runme.2.d
@@ -5,6 +5,7 @@ import enum_thorough.enum_thorough;
 import enum_thorough.AnonStruct;
 import enum_thorough.colour;
 import enum_thorough.DifferentTypes;
+import enum_thorough.Enum2995;
 import enum_thorough.FirStruct;
 import enum_thorough.HairStruct;
 import enum_thorough.IgnoreTest;
@@ -444,5 +445,12 @@ void main() {
     enforce(cast(int)globalDifferentTypesTest(global_enum) == 'C', "global differentTypes 5 failed");
     global_enum = global_typedefaultint;
     enforce(cast(int)globalDifferentTypesTest(global_enum) == 'D', "global differentTypes 6 failed");
+  }
+  // Regression tests for bad handling of bool expressions (#2995)
+  {
+    enforce(cast(int)Enum2995.T1 == 1, "Enum2995 T1 failed");
+    enforce(cast(int)Enum2995.T2 == 1, "Enum2995 T2 failed");
+    enforce(cast(int)Enum2995.F1 == 0, "Enum2995 F1 failed");
+    enforce(cast(int)Enum2995.F2 == 0, "Enum2995 F2 failed");
   }
 }
--- a/Examples/test-suite/enum_thorough.i
+++ b/Examples/test-suite/enum_thorough.i
@@ -660,6 +660,16 @@ struct EnumCharStruct {
     enumcharAE3 = '\xC6' // AE (latin1 encoded)
   };
 };
+
+// We can't test this for C# and Java it seems.
+#if !defined(SWIGCSHARP) && !defined(SWIGJAVA)
+enum Enum2995 {
+   T1 = (1 == 1),
+   T2 = !false,
+   F1 = (1 != 1),
+   F2 = !true,
+};
+#endif
 %}
 
 #if defined(SWIGJAVA)
--- a/Source/CParse/cscanner.c
+++ b/Source/CParse/cscanner.c
@@ -19,6 +19,7 @@
 #include "parser.h"
 #include <string.h>
 #include <ctype.h>
+#include <errno.h>
 
 /* Scanner object */
 static Scanner *scan = 0;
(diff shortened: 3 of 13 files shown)


ojwb:
This is getting close - mostly I need to sort out regression tests for the bugs it fixes (or at least those I'm aware of).

I've compared partialcheck output.  The main change is SWIG converts integer literals to decimal in the version offered in `numval`, which means things like C++11 binary literals (e.g. `0b101010`) are now automatically converted when generated in target language code so work even if the target language doesn't support this syntax.

Even in Python where we converted integers before the code is simpler:

```
diff -ru partialcheck-base/python/default_args.py Examples/test-suite/python/default_args.py
--- partialcheck-base/python/default_args.py    2024-09-13 13:18:45.705715867 +1200
+++ Examples/test-suite/python/default_args.py  2024-09-13 13:54:19.804246176 +1200
@@ -65,28 +65,28 @@
     def value_m1(self, first, pos=-1):
         return _default_args.TrickyInPython_value_m1(self, first, pos)
 
-    def value_0xabcdef(self, first, rgb=0xabcdef):
+    def value_0xabcdef(self, first, rgb=11259375):
         return _default_args.TrickyInPython_value_0xabcdef(self, first, rgb)
 
-    def value_0644(self, first, mode=int("0644", 8)):
+    def value_0644(self, first, mode=420):
         return _default_args.TrickyInPython_value_0644(self, first, mode)
 
     def value_perm(self, *args):
         return _default_args.TrickyInPython_value_perm(self, *args)
 
-    def value_m01(self, first, val=int("-01", 8)):
+    def value_m01(self, first, val=-1):
         return _default_args.TrickyInPython_value_m01(self, first, val)
 
     def booltest2(self, *args):
         return _default_args.TrickyInPython_booltest2(self, *args)
 
-    def max_32bit_int1(self, a=0x7FFFFFFF):
+    def max_32bit_int1(self, a=2147483647):
         return _default_args.TrickyInPython_max_32bit_int1(self, a)
 
     def max_32bit_int2(self, a=2147483647):
         return _default_args.TrickyInPython_max_32bit_int2(self, a)
 
-    def min_32bit_int1(self, a=-0x80000000):
+    def min_32bit_int1(self, a=-2147483648):
         return _default_args.TrickyInPython_min_32bit_int1(self, a)
 
     def too_big_32bit_int1(self, *args):
```

The downside of this is when the original base of the value is more helpful to understand the value, e.g. `mode=int("0644", 8)` probably means more to a human reader very familiar with Unix permissions than `mode=420`.  That becomes an issue in cases we are generating doc comments, or where the language allows the user to inspect the default parameter values.  However it seems Python itself does a similar conversion, e.g. consider:

```
>>> def f(x = int("0644", 8)): return x
... 
>>> help(f)
```
outputs:
```
Help on function f in module __main__:

f(x=420)
```

We also now handle string and character literals as Python default arguments (maybe this needs more thought in the light of bytes vs unicode strings?):

```
@@ -186,26 +186,26 @@
 def casts1(*args):
     return _default_args.casts1(*args)
 
-def casts2(*args):
-    return _default_args.casts2(*args)
+def casts2(m='Hello'):
+    return _default_args.casts2(m)
 
-def chartest1(*args):
-    return _default_args.chartest1(*args)
+def chartest1(c='x'):
+    return _default_args.chartest1(c)
 
-def chartest2(*args):
-    return _default_args.chartest2(*args)
+def chartest2(c='\0'):
+    return _default_args.chartest2(c)
 
-def chartest3(*args):
-    return _default_args.chartest3(*args)
+def chartest3(c='\1'):
+    return _default_args.chartest3(c)
 
-def chartest4(*args):
-    return _default_args.chartest4(*args)
+def chartest4(c='\n'):
+    return _default_args.chartest4(c)
 
-def chartest5(*args):
-    return _default_args.chartest5(*args)
+def chartest5(c='B'):
+    return _default_args.chartest5(c)
 
-def chartest6(*args):
-    return _default_args.chartest6(*args)
+def chartest6(c='C'):
+    return _default_args.chartest6(c)
 NoType = _default_args.NoType
 
 def dummy(*args):
```

There's an odd case in `preproc_constants_c.go`:

```
@@ -248,8 +250,20 @@
 }
 
 var EXPR_MOD int = _swig_getEXPR_MOD()
-const EXPR_PLUS int = 0xFF+2
-const EXPR_MINUS int = 0xFF+2
+func _swig_getEXPR_PLUS() (_swig_ret int) {
+       var swig_r int
+       swig_r = (int)(C._wrap_EXPR_PLUS_preproc_constants_c_219c889b37716cca())
+       return swig_r
+}
+
+var EXPR_PLUS int = _swig_getEXPR_PLUS()
+func _swig_getEXPR_MINUS() (_swig_ret int) {
+       var swig_r int
+       swig_r = (int)(C._wrap_EXPR_MINUS_preproc_constants_c_219c889b37716cca())
+       return swig_r
+}
+
+var EXPR_MINUS int = _swig_getEXPR_MINUS()
 func _swig_getEXPR_LSHIFT() (_swig_ret int) {
        var swig_r int
        swig_r = (int)(C._wrap_EXPR_LSHIFT_preproc_constants_c_219c889b37716cca())
```

This is actually because the previous Go-specific code was a bit liberal and decided `0xFF+2` was a simple integer literal.  It isn't, but we happen to get away with using the expression in Go.  I'm thinking this change is reasonable.

Octave has this change to a doc comment, which I think is a bug fix (`None` seems correct, while `nil` seems to be from cutting and pasting from `ruby.cxx`):

```
diff -ru partialcheck-base/octave/cpp11_null_pointer_constant_wrap.cxx Examples/test-suite/octave/cpp11_null_pointer_constant_wrap.cxx
--- partialcheck-base/octave/cpp11_null_pointer_constant_wrap.cxx       2024-09-13 13:14:19.917423749 +1200
+++ Examples/test-suite/octave/cpp11_null_pointer_constant_wrap.cxx     2024-09-13 13:49:55.127706499 +1200
@@ -2584,7 +2584,7 @@
 };
 
 static const char* _wrap_A_NullPtrMethod_texinfo = "-*- texinfo -*-\n\
-@deftypefn {Loadable Function} NullPtrMethod (@var{self}, @var{ptr} = nil)\n\
+@deftypefn {Loadable Function} NullPtrMethod (@var{self}, @var{ptr} = None)\n\
 @var{self} is of type A. @var{ptr} is of type double *. \n\
 @end deftypefn";
 
```

Ruby uses double quotes for strings which are character literals in doc comments now (this is actually a bug fix, since Ruby single quoted strings don't support escape sequences, while double quoted strings do):

```
@@ -6356,7 +6356,7 @@
   Document-method: Default_args.chartest4
 
   call-seq:
-    chartest4(c='\n') -> char
+    chartest4(c="\n") -> char
     chartest4 -> char
 
 A module function.
```

Also e.g. `bool x = 0` as a parameter now shows as `False` in Ruby doc comments:

```
@@ -2628,7 +2628,7 @@
   Document-method: Default_arg_values::Display.mybool0
 
   call-seq:
-    mybool0(x=0) -> bool
+    mybool0(x=False) -> bool
     mybool0 -> bool
 
 An instance method.

```

ojwb:
For 4.3.0 I'm leaning towards leaving `TRUE`, `FALSE` and Python char/string literal handling in default values as-is, but I'm definitely open to discussion on these points.

I'm definitely not intending to go further and handle floating point values or `NULL`/`nullptr` via `numval` for 4.3.0 (but doing so is the longer term plan).

PR #3024, opened 2024-09-16
Fix cstrings.swg cstring_output_allocate example freeing

rlee287:
The memory freeing needs to be done with `free(*$1)` instead of `free($1)`: because $1 is a `char**` (a pointer to a heap-allocated string), one dereference is required to peel back a layer of pointer indirection and pass the correct heap-allocated pointer to `free`.

Also noting that the examples at https://swig.org/Doc4.2/Library.html#Library_nn12 have the correct invocation of `free`.

### DIFF
--- a/Lib/typemaps/cstrings.swg
+++ b/Lib/typemaps/cstrings.swg
@@ -236,7 +236,7 @@
  * This macro is used to return Character data that was
  * allocated with new or malloc.
  *
- *     %cstring_output_allocate(Char **outx, free($1));
+ *     %cstring_output_allocate(Char **outx, free(*$1));
  *     void foo(Char **outx) {
  *         *outx = (Char *) malloc(512);
  *         strcpy(outx,"blah blah\n");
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks - merged (with the very long single line commit description wrapped - it's really best to not use github's built in editor to make PRs as it does stupid things like that).

The testsuite coverage for this also has the dereference, so that's OK:

```
Examples/test-suite/li_cstring.i:%cstring_output_allocate(char **out7, delete [] *$1);
Examples/test-suite/li_cstring.i:%cstring_output_allocate_size(char **out8, int *size, delete [] *$1);
Examples/test-suite/li_cstring.i:%cstring_output_allocate(char **out7, free(*$1));
Examples/test-suite/li_cstring.i:%cstring_output_allocate_size(char **out8, int *size, free(*$1));
```

I pushed a fix for the same problem in the ocaml-specific version (not sure why it has one exactly when no other target languages seem to need to but not going to visit that rabbit hole today).

Issue #2897, opened 2024-04-30
#999 Not used by default in MSVC


bartoli:
It seems like #999  could work at least up to MSVC2015. But, the condition in SwigValueWrapper depends on __cplusplus value being >=2011XX, which by default is stuck to 199711 for MSVC (https://learn.microsoft.com/fr-fr/cpp/build/reference/zc-cplusplus?view=msvc-170) unless we are at least in MSVC2017 and have the option /Zc:__cplusplus

Could the condition be updated to also check _MSVC_LANG value?

wsfulton:
SWIG expects a standards compliant compiler as it generates standards compliant code, `_MSVC_LANG` is not in the C or C++ standard. Please configure your compiler to be standards compliant or use a compiler that is standards compliant. You've already identified that the `/Zc:__cplusplus` option is required for MSVC to be compliant, so you do need to use this option. Related to #2860.

bartoli:
I actually am with VS2015 compilers, where the /Zc poption was not availabe yet, hence why my request, since the compiler can build the code, but not unless i patch the condition.
I'll just patch my swig output

wsfulton:
`__cplusplus` is just a macro so define it to an appropriate value using the `\D` cl.exe option or add it to the top of your generated wrapper code using `%begin`, for example:
```swig
%begin %{
#define __cplusplus 201103L
%}
```

ojwb:
Redefining `__cplusplus` for the compiler like that would be undefined behaviour though (citation: https://en.cppreference.com/w/cpp/preprocessor/replace says "Attempts to redefine or undefine these macros result in undefined behavior.")

Issue #2860, opened 2024-04-04
Should SWIG workaround MSVC stupidity?


ojwb:
By default in MSVC it seems `__cplusplus` has the value for C++98 **even if you explicitly request a newer language version with e.g. `/std:c++17`** - see the last line in the table here:

https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

This seems a nasty trap - I've just wasted a fair bit of time trying to debug a problem which turned out to be SWIG's move semantics support not getting enabled due to this (because the SWIG-generated code checks the value of `__cplusplus` which reports C++98).

You can make MSVC do the sensible thing with command line option `/Zc:__cplusplus`, but you somehow need to know that you need to do this, and the docs for `/std:` don't seem to tell you: https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170

MSVC defines `_MSVC_LANG` with the value `__cplusplus` would have with a sensible compiler.  SWIG could check this, though the obvious approach would make any check for the `__cplusplus` value a bit unwieldly - a current check such as:

```
#if defined(__cplusplus) && __cplusplus >=201103L
```

becomes something like:

```
#if define(__cplusplus) && (defined(_MSVC_LANG) ? _MSVC_LANG : __cplusplus) > 201103L
```

We could define a SWIG-specific macro at the top of the generated C/C++ wrapper, which could actually simplify these checks by being defined (to say `-1L`) for C:

```
#ifdef __cplusplus
# ifdef _MSVC_LANG
#  define SWIG_CPPSTD _MSVC_LANG
# else
#  define SWIG_CPPSTD __cplusplus
# endif
#else
# define SWIG_CPPSTD -1L
#endif
```

Then the check above can just be:

```
#if SWIG_CPPSTD >= 201103L
```

(Such a macro could also solve #2818 if we take into account whether SWIG was run with `-c++` when generating the code to set it.)

Or we could instead alert users about the need to use `/Zc:__cplusplus` by generating something like this in the C++ wrapper:

```
#if defined(_MSVC_LANG) && _MSVC_LANG != __cplusplus
# warning Use MSVC option /Zc:__cplusplus to get __cplusplus correctly defined
#endif
```

wsfulton:
> By default in MSVC it seems `__cplusplus` has the value for C++98 **even if you explicitly request a newer language version with e.g. `/std:c++17`** -

[...]

PR #3027, opened 2024-09-17
Use noexcept instead of throw() in a number of places

r-barnes:
[Dynamic throw specifications](https://en.cppreference.com/w/cpp/language/except_spec) (denoted by `throw()`) are being removed from C++. [`noexcept`](https://en.cppreference.com/w/cpp/language/noexcept_spec) is the appropriate replacement.

This PR makes the `throw()` -> `noexcept` switch in a number of places.

### DIFF
--- a/Examples/test-suite/cpp11_lambda_functions.i
+++ b/Examples/test-suite/cpp11_lambda_functions.i
@@ -27,7 +27,7 @@
 %warnfilter(SWIGWARN_CPP11_LAMBDA) Space1::Space2::lambda20;
 
 // throw is invalid in C++17 and later, only SWIG to use it
-#define TESTCASE_THROW0() throw()
+#define TESTCASE_THROW0() noexcept
 #define TESTCASE_THROW1(T1) throw(T1)
 %{
 #define TESTCASE_THROW0()
@@ -120,4 +120,3 @@ int(*lambda101notauto)(int, int) = [] (int a, int b) { return a + b; };
 int lambda102 = [] (int a, int b) mutable { return a + b; }(1, 2);
 void lambda_init(int = ([]{ return 0; })());
 %}
-
--- a/Lib/lua/std_except.i
+++ b/Lib/lua/std_except.i
@@ -19,9 +19,9 @@ namespace std
   class exception 
   {
   public:
-    exception() throw() { }
-    virtual ~exception() throw();
-    virtual const char* what() const throw();
+    exception() noexcept { }
+    virtual ~exception() noexcept;
+    virtual const char* what() const noexcept;
   }; 
 }
 
--- a/Source/CParse/parser.y
+++ b/Source/CParse/parser.y
@@ -3603,7 +3603,7 @@ cpp_alternate_rettype : primitive_type
    auto myFunc = [] { return something; };
    auto myFunc = [](int x, int y) { return x+y; };
    auto myFunc = [](int x, int y) -> int { return x+y; };
-   auto myFunc = [](int x, int y) throw() -> int { return x+y; };
+   auto myFunc = [](int x, int y) noexcept -> int { return x+y; };
    auto six = [](int x, int y) { return x+y; }(4, 2);
    ------------------------------------------------------------ */
 cpp_lambda_decl : storage_class AUTO idcolon EQUAL lambda_introducer lambda_template LPAREN parms RPAREN cpp_const lambda_body lambda_tail {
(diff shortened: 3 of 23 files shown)


Examples/test-suite/cpp11_lambda_functions.i
```diff
@@ -27,7 +27,7 @@
 %warnfilter(SWIGWARN_CPP11_LAMBDA) Space1::Space2::lambda20;
 
 // throw is invalid in C++17 and later, only SWIG to use it
-#define TESTCASE_THROW0() throw()
+#define TESTCASE_THROW0() noexcept
```
ojwb, line 30: As the comment above notes, only SWIG gets fed this - it is testing SWIG's support for `throw()`, which we want to continue to test.

ojwb:
I didn't check through all of this, but there's a fundamental problem with the proposed change as-is which is that `noexcept` was added in C++11, so using it unconditionally in generated code (like this change does in many places) will stop the output being valid C++98 which is still the aim when not wrapping features from a newer C++ standard (maybe you think that's not a very useful aim at this point, but it's not me you would need to convince on that point).

So this probably needs to conditionally use `throws()` or `noexcept`, but note we can't just check `__cplusplus` because Microsoft - MSVC always defines `__cplusplus` to `199711L` regardless of the C++ standard it's actually using (newer versions support an option to tell it behave sensibly, but we can't rely on people compiling SWIG generated code enabling that).

Lib/lua/std_except.i
```diff
@@ -19,9 +19,9 @@ namespace std
...
   {
   public:
-    exception() throw() { }
-    virtual ~exception() throw();
-    virtual const char* what() const throw();
+    exception() noexcept { }
+    virtual ~exception() noexcept;
+    virtual const char* what() const noexcept;
   }; 
 }
```
ojwb, line 26: This is only shown to SWIG and the `%ignore` should mean no code gets generated.  However we can rely on SWIG supporting `noexcept` so this change should be safe and perhaps worthwhile even if it doesn't actually do anything for the stated goal here.

Source/CParse/parser.y
```diff
@@ -3603,7 +3603,7 @@ cpp_alternate_rettype : primitive_type
    auto myFunc = [] { return something; };
    auto myFunc = [](int x, int y) { return x+y; };
    auto myFunc = [](int x, int y) -> int { return x+y; };
-   auto myFunc = [](int x, int y) throw() -> int { return x+y; };
+   auto myFunc = [](int x, int y) noexcept -> int { return x+y; };
```
ojwb, line 3606: You're changing a comment here.

r-barnes:
From your description, it seems as though SWIG's commitment to MSVC using C++98 means that you will never be able to support the C++20 standard on any compiler. That doesn't seem like a good outcome.

ojwb:
> From your description, it seems as though SWIG's commitment to MSVC using C++98 means that you will never be able to support the C++20 standard on any compiler.

We can support C++20 for decent compilers just by checking the `__cplusplus`.  Microsoft decided to hobble their compiler though and my point was that means this is not enough to support MSVC.

> you haven't suggested a course of action

Check the value of `__cplusplus` for decent compilers; check the stupid Microsoft-specific macro for MSVC: see #2860 for details.

If there was a feature test macro defined for having `noexcept` then we could use that, but I don't see one at https://en.cppreference.com/w/cpp/feature_test - there's `__cpp_noexcept_function_type` but that's for "Make exception specifications be part of the type system" from C++17 which is a narrower requirement.

ojwb:
I just spotted that William essentially WONTFIXED #2897 with _the /Zc:__cplusplus option is required for MSVC to be compliant, so you do need to use this option._  - if that's our policy on Microsoft's `__cplusplus` stupidity then we can just test the value of `__cplusplus` here and rely on MSVC users figuring out they need to use this option.

There's the question of MSVC 2015 which doesn't have the magic option to make it conform, but in this particular case that just means MSVC 2015 would compile the version still using `throw()` which is probably not a problem since MSVC 2015 doesn't know about newer C++ versions where `throw()` is getting removed.

wsfulton:
As @ojwb says, we need to test for `throw()` for c++98 users, even if they are in the minority. gcc doesn't give a deprecation warning for dynamic exception specifications, however, I found clang has a `-Wdeprecated-dynamic-exception-spec` warning which I used to flush out the problems in the generated code. I've also put some hacks in to make the test-suite warning free for `-Wdeprecated-dynamic-exception-spec`.

PR #3035, opened 2024-09-26
Implementation for std::optional for java code wrappers

AnDevi:
I provided support for wrapping to java std::optional used with types listed below:
- Custom cpp types
 - Primitive types:
  |- bool
  |- int 
  |- flaot 
  |- long
  |- double
 - Types from std library:
  |- std::string
  
  I'm using it for quite a while in our project without any problems yet. 
  I'll also try to provide some tests in here for that implementation.

### DIFF
--- a/Lib/java/std_optional.i
+++ b/Lib/java/std_optional.i
@@ -1,0 +1,365 @@
+// This implementation adds support for std::optional for:
+// - Custom types
+// - Primitive types:
+//  |- bool
+//  |- int 
+//  |- flaot 
+//  |- long
+//  |- double
+// - Types from std library:
+//  |- std::string
+
+
+// 1) Implementation for custom types
+
+// Main user macro for defining optional typemaps for both const and non-const pointer types
+%define %optional(TYPE...)
+SWIG_STD_OPTIONAL_TYPEMAPS(, TYPE)
+SWIG_STD_OPTIONAL_TYPEMAPS(const, TYPE)
+%enddef
+
+#ifndef SWIG_STD_OPTIONAL_TYPEMAPS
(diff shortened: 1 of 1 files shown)


vadz:
Sorry, but how is it actually used on Java side? You need to add some unit tests checking that this works as expected, and they would be useful to understand this.

FWIW I'm surprised that this doesn't seem to use `java.util.Optional` at all. My own SWIG typemaps for `optional` map it to this Java type, allowing to use it naturally (or at least I think so) in Java, e.g. calling `isPresent()` on the returned values etc.

AnDevi:
In my case the returned value to java doesn't have any api as `isPresent()`,
`has_value()` etc.
Ex. function `optional<Point> getPoint()` is generated as `public @Nullable
Point getPoint()`, under the hood it checks in cpp generated code if it has
value.
In the end for me java code generated by swig is only a proxy between cpp
and kotlin.
So in final usage in Kotlin I need to use elvis syntax ex.
`rect?.getPoint()` so when `getPoint()` is described as ***@***.***` without
this "elvis check" it won't compile.
In my opinion it is a pretty convenient solution which is natural in
Kotlin.

In the first iteration I added an additional layer with `java.util.Optional`
to generate code, but in the end I decided to skip it.







czw., 26 wrz 2024 o 16:54 VZ ***@***.***> napisał(a):

vadz:
Is this Android and/or Kotlin-specific then?

christophe-calmejane:
std::optional was introduced in c++17, shouldn't your tests in Examples/test-suite be named cpp17_xxx instead of cpp11_xxx?

AnDevi:
Thanks @christophe-calmejane , you're right. 
I changed those names, and I also added support for enum types hold by optional. I recently found out it's missing.

PR #3066, opened 2024-11-01
fix: support Python 3.14

### LINKED ISSUE #2967: __package__ will cease being set/used  in Python 3.15
The files generated by SWIG depend on `__package__` being present, but Python 3.14 will removing this in favor of `__spec__.parent`. The 3.13 release notes https://docs.python.org/3.13/whatsnew/3.13.html#pending-removal-in-python-3-14 say:

> [importlib](https://docs.python.org/3.13/library/importlib.html#module-importlib): `__package__` and `__cached__` will cease to be set or taken into consideration by the import system ([gh-97879](https://github.com/python/cpython/issues/97879)).

Both could be checked, or just `__spec__.parent` - I can verify that works back to 3.6, which is the oldest Python I had to check.

henryiii:
Fix #2967.

I don't remember the exact details of the patch I was using locally, so let's see if this works, can iterate on it if not.

Ah, I didn't realize Python 2 was still supported, I might need to fix a branch for Python 2.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -138,6 +138,9 @@ jobs:
           VER: '3.13'
           CSTD: gnu99
         - SWIGLANG: python
+          VER: '3.14'
+          CSTD: gnu99
+        - SWIGLANG: python
           PY2: 2
           SWIG_FEATURES: -builtin
         - SWIGLANG: python
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -6548,7 +6548,7 @@ An input interface file, foo.i, results in the two modules foo.py and _foo.so fo
 
 <p>
 In this configuration, the pure Python module, foo.py, tries to load the C/C++ module, _foo, from the same package foo.py is
-located in.  The package name is determined from the <tt>__spec__.parent</tt>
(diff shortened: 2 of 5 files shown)


ojwb:
@henryiii What's still to do here?

Was it CI coverage for 3.14?  If that's hard to install in a GHA CI image I think it would be reasonable to merge this knowing that it doesn't break 3.13 and earlier (which we can see from existing CI results) and that we'll get CI testing for 3.14 once it's actually released and packaged by Ubuntu.

ojwb:
I use https://github.com/actions/setup-python in a different project, and that appears to support `3.14-dev` - not sure how easy it would be to fit into SWIG's existing CI configuration though (it would probably need to be used conditionally).

henryiii:
It wasn’t true when I made the PR, but now there are several ways to get alpha builds.

ojwb:
@henryiii You didn't actually answer the key question though - what's still to do here?

It's still marked as a draft so presumably something (or maybe you forgot to mark it as ready for review?)

henryiii:
It's draft because I didn't have a way to test it. Adding a 3.14 job would finish it off, I think.

wsfulton:
Closing in favour of #3159 which has evolved this patch with all tests passing.

PR #3067, opened 2024-11-01
[python] Define SWIG_HEAPTYPES for limited api mode

jschueller:
Fixes 500d3e1, Seems SWIG_HEAPTYPES is needed when compiling with Py_LIMITED_API (py3.9) else I get this error:
swigpyrun.h:1415:23: error: variable has incomplete type 'PyTypeObject' (aka '_typeobject')
  static PyTypeObject varlink_type;

/cc @ojwb

### DIFF
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -335,6 +335,10 @@ swig_varlink_setattr(PyObject *o, char *n, PyObject *p) {
   return res;
 }
 
+#if defined(Py_LIMITED_API)
+# define SWIG_HEAPTYPES
+#endif
+
 SWIGINTERN PyTypeObject*
 swig_varlink_type(void) {
   static char varlink__doc__[] = "Swig var link object";
--- a/Examples/python/check.list
+++ b/Examples/python/check.list
@@ -9,6 +9,7 @@ enum
 exception
 exceptproxy
 extend
+external_runtime
 funcptr
 funcptr2
 functor
--- a/Examples/python/external_runtime/Makefile
+++ b/Examples/python/external_runtime/Makefile
@@ -1,0 +1,26 @@
+TOP        = ../..
+SWIGEXE    = $(TOP)/../swig
+SWIG_LIB_DIR = $(TOP)/../$(TOP_BUILDDIR_TO_TOP_SRCDIR)Lib
+CXXSRCS    = example.cxx
+TARGET     = example
+INTERFACE  = example.i
+LIBS       = -lm
+SWIGOPT    =
+
+check: build
+	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' python_run
+
+build:
+	SWIG_LIB='$(SWIG_LIB_DIR)' $(SWIGEXE) -python $(SWIGOPT) -external-runtime swig_runtime.hxx
+	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' $(SWIGLIB) CXXSRCS='$(CXXSRCS)' \
+	SWIG_LIB_DIR='$(SWIG_LIB_DIR)' SWIGEXE='$(SWIGEXE)' \
+	SWIGOPT='$(SWIGOPT)' TARGET='$(TARGET)' INTERFACE='$(INTERFACE)' python_cpp
+
+static:
+	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' $(SWIGLIB) CXXSRCS='$(CXXSRCS)' \
+	SWIG_LIB_DIR='$(SWIG_LIB_DIR)' SWIGEXE='$(SWIGEXE)' \
+	SWIGOPT='$(SWIGOPT)' TARGET='mypython' INTERFACE='$(INTERFACE)' python_cpp_static
+
+clean:
+	rm -f swig_runtime.hxx
+	$(MAKE) -f $(TOP)/Makefile SRCDIR='$(SRCDIR)' TARGET='$(TARGET)' python_clean
--- a/Examples/python/external_runtime/example.cxx
+++ b/Examples/python/external_runtime/example.cxx
@@ -5,24 +5,5 @@
 Function::Function(PyObject * pyCallable)
 : pyObj_(pyCallable), meshValue_(0)
 {
-  if (pyCallable)
-  {
-    Py_XINCREF(pyCallable);
-    PyObject * pyMesh = PyObject_CallMethod(pyObj_, const_cast<char *>("mesh"), const_cast<char *>("()"));
-    if (!pyMesh)
-      throw std::runtime_error("null pyMesh");
-    void * ptr = 0;
-    if (SWIG_IsOK(SWIG_ConvertPtr(pyMesh, &ptr, SWIG_TypeQuery("Mesh *"), 0)))
-    {
-      Mesh *mesh = reinterpret_cast< Mesh * >(ptr);
-      if (!mesh)
-        throw std::runtime_error("null mesh");
-      meshValue_ = mesh->value();
-    }
-    else
-    {
-      throw;
-    }
-    Py_XDECREF(pyMesh);
-  }
+  // stuff happens in the typemap of PyObject * pyCallable
 }
--- a/Examples/python/external_runtime/example.cxx
+++ b/Examples/python/external_runtime/example.cxx
@@ -1,0 +1,28 @@
+#include <Python.h>
+#include "swig_runtime.hxx"
+#include "example.h"
+
+Function::Function(PyObject * pyCallable)
+: pyObj_(pyCallable), meshValue_(0)
+{
+  if (pyCallable)
+  {
+    Py_XINCREF(pyCallable);
+    PyObject * pyMesh = PyObject_CallMethod(pyObj_, const_cast<char *>("mesh"), const_cast<char *>("()"));
+    if (!pyMesh)
+      throw std::runtime_error("null pyMesh");
+    void * ptr = 0;
+    if (SWIG_IsOK(SWIG_ConvertPtr(pyMesh, &ptr, SWIG_TypeQuery("Mesh *"), 0)))
+    {
+      Mesh *mesh = reinterpret_cast< Mesh * >(ptr);
+      if (!mesh)
+        throw std::runtime_error("null mesh");
+      meshValue_ = mesh->value();
+    }
+    else
+    {
+      throw;
+    }
+    Py_XDECREF(pyMesh);
+  }
+}
--- a/Examples/python/external_runtime/example.h
+++ b/Examples/python/external_runtime/example.h
@@ -17,8 +17,8 @@ class Function
 {
 public:
 	explicit Function(PyObject * pyCallable = 0);
-	int meshValue() { return meshValue_;}
+	int meshValue() { return meshValue_;}	int meshValue_;
+
 private:
 	PyObject * pyObj_;
-	int meshValue_;
 };
--- a/Examples/python/external_runtime/example.h
+++ b/Examples/python/external_runtime/example.h
@@ -1,0 +1,24 @@
+/* File : example.h */
+
+#include <iostream>
+#include "Python.h"
+
+class Mesh
+{
+public:
+	Mesh(const int value = 0)
+	: value_(value) {}
+	int value() { return value_;}
+private:
+	int value_;
+};
+
+class Function
+{
+public:
+	explicit Function(PyObject * pyCallable = 0);
+	int meshValue() { return meshValue_;}
+private:
+	PyObject * pyObj_;
+	int meshValue_;
+};
--- a/Examples/python/external_runtime/example.i
+++ b/Examples/python/external_runtime/example.i
@@ -4,6 +4,28 @@
 #include "example.h"
 %}
 
+%typemap(in) PyObject * pyCallable
+{
+  if ($input)
+  {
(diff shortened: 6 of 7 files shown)


Lib/python/pyrun.swg
```diff
@@ -335,6 +335,10 @@ swig_varlink_setattr(PyObject *o, char *n, PyObject *p) {
   return res;
 }
 
+#if defined(Py_LIMITED_API)
+# define SWIG_HEAPTYPES
+#endif
+
```
ojwb, line 341: This seems the wrong fix - with this change we'd have multiple different places where this gets defined.

Travelling currently so not easy for me to look but why isn't it getting defined by the code I moved in the commit you referenced?

jschueller: because the code pyruntime.swg is not used (only pyrun.swg has effect here):  in the client code I only include the swig runtime header: here is the same sample code that I used to report a previous error:
https://github.com/jschueller/swig-cmake-example/blob/master/PythonFieldFunction.cxx#L2

ojwb:
Ah, so the external runtime case.  Sigh, I really carefully checked this change wouldn't break anything, but missed that case.

I think we really need to move that code back then.  I can take a proper look once I'm back.

jschueller:
hello @ojwb, did you have a look at this ? maybe I can help to add a testcase with external runtime ?

ojwb:
>  did you have a look at this ?

Sorry, not yet.  I caught covid and have been rather backlogged since.

> maybe I can help to add a testcase with external runtime ?

If you're able to add a testcase which fails for current git master but passed before 500d3e1470a56eeb79cafb917d85e2819a125a0d that would help a lot.  Currently we don't seem to have a single testcase which uses `-external-runtime`.

jschueller:
> If you're able to add a testcase which fails for current git master but passed before [500d3e1](https://github.com/swig/swig/commit/500d3e1470a56eeb79cafb917d85e2819a125a0d) that would help a lot. Currently we don't seem to have a single testcase which uses `-external-runtime`.

I added a test-case if you want to look at it

ojwb:
> I added a test-case if you want to look at it

Thanks, though it would be better to have something which doesn't make use of the Python C API directly and to add it to the testsuite (i.e. under `Examples/test-suite`) rather than as an example (i.e. under `Examples/<targetlanguage>`) as then we can easily test this for more target languages.  (I don't know why our testsuite is hidden under "Examples" - it seems confusing to me too...)

Also the examples runtime testing isn't really set up for failing tests based on runtime checks, though that's less relevant for this particular bug as the problem manifests when building the generated bindings.

ojwb:
If I revert your proposed change to the library files the new example still seems to pass:

```
$ git checkout master Lib/python/pyrun.swg
Updated 1 path from b57550e9
$ make -s
$ make -s -C Examples/python/external_runtime 
42
OK

python exit
$ echo $?
0
```

Am I missing something?

jschueller:
> Am I missing something?
yes, this example wont compile in limited api mode

jschueller:
> Thanks, though it would be better to have something which doesn't make use of the Python C API directly and to add it to the testsuite (i.e. under `Examples/test-suite`) rather than as an example (i.e. under `Examples/<targetlanguage>`) as then we can easily test this for more target languages. (I don't know why our testsuite is hidden under "Examples" - it seems confusing to me too...)

I dont know how I could test these SWIG_ConvertPtr/SWIG_TypeQuery functions without using Python directly, any idea ?

ojwb:
> yes, this example wont compile in limited api mode with current master

Aha!  Thanks.

> I dont know how I could test these SWIG_ConvertPtr/SWIG_TypeQuery functions without using Python directly, any idea ?

Use them in a typemap?  You can avoid referencing `PyObject*` by using `SWIG_Object` instead, which gets `#define`-d to a suitable type by most target languages, so I think it ought to be possible to write one typemap which works for most languages.

Or if that's not suitable (I don't really understand what the external runtime feature is for TBH) maybe wrap these two functions with SWIG (e.g. see `template_typedef.i`) and then you can call them from the testcase's `_runme.py`, etc?

jschueller:
it doesnt like I can use SWIG_Object in a typemap or even from swig_runtime.hxx
```
%typemap(in) PyObject *pyCallable %{
  std::cout << "TYpemap"<<std::endl;
  if ($1)
  {
    Py_XINCREF($1);
    SWIG_Object * pyMesh = PyObject_CallMethod(pyObj_, const_cast<char *>("mesh"), const_cast<char *>("()"));
    if (!pyMesh)
      throw std::runtime_error("null pyMesh");
    void * ptr = 0;
    if (SWIG_IsOK(SWIG_ConvertPtr(pyMesh, &ptr, SWIG_TypeQuery("Mesh *"), 0)))
    {
      Mesh *mesh = reinterpret_cast< Mesh * >(ptr);
      if (!mesh)
        throw std::runtime_error("null mesh");
      meshValue_ = mesh->value();
    }
    else
    {
      throw;
    }
    Py_XDECREF(pyMesh);
  }
%}
example_wrap.cxx:3686:5: error: ‘SWIG_Object’ was not declared in this scope; did you mean ‘SwigPyObject’?
```

ojwb:
> it doesnt like I can use SWIG_Object in a typemap

You need to use `{`...`}` delimiters - SWIG's preprocessor doesn't process any text enclosed in `%{`...`%}` (as documented in the manual).  With that change I get:

```
example_wrap.cxx: In function ‘PyObject* _wrap_new_Function__SWIG_0(PyObject*, Py_ssize_t, PyObject**)’:
example_wrap.cxx:3682:49: error: ‘pyObj_’ was not declared in this scope
 3682 |       PyObject * * pyMesh = PyObject_CallMethod(pyObj_, const_cast<char *>("mesh"), const_cast<char *>("()"));
      |                                                 ^~~~~~
example_wrap.cxx:3686:37: error: cannot convert ‘PyObject**’ {aka ‘_object**’} to ‘PyObject*’ {aka ‘_object*’}
 3686 |       if (SWIG_IsOK(SWIG_ConvertPtr(pyMesh, &ptr, SWIG_TypeQuery("Mesh *"), 0)))
      |                                     ^~~~~~
      |                                     |
      |                                     PyObject** {aka _object**}
example_wrap.cxx:353:37: note: in definition of macro ‘SWIG_IsOK’
  353 | #define SWIG_IsOK(r)               (r >= 0)
      |                                     ^
example_wrap.cxx:1149:57: note: in expansion of macro ‘SWIG_Python_ConvertPtr’
 1149 | #define SWIG_ConvertPtr(obj, pptr, type, flags)         SWIG_Python_ConvertPtr(obj, pptr, type, flags)
      |                                                         ^~~~~~~~~~~~~~~~~~~~~~
example_wrap.cxx:3686:21: note: in expansion of macro ‘SWIG_ConvertPtr’
 3686 |       if (SWIG_IsOK(SWIG_ConvertPtr(pyMesh, &ptr, SWIG_TypeQuery("Mesh *"), 0)))
      |                     ^~~~~~~~~~~~~~~
example_wrap.cxx:2569:40: note:   initializing argument 1 of ‘int SWIG_Python_ConvertPtrAndOwn(PyObject*, void**, swig_type_info*, int, int*)’
 2569 | SWIG_Python_ConvertPtrAndOwn(PyObject *obj, void **ptr, swig_type_info *ty, int flags, int *own) {
      |                              ~~~~~~~~~~^~~
example_wrap.cxx:3691:9: error: ‘meshValue_’ was not declared in this scope
 3691 |         meshValue_ = mesh->value();
      |         ^~~~~~~~~~
```

which look to just be problems with the typemap (e.g. `pyObj_` and `meshValue_` aren't declared anywhere).

jschueller:
that would work, but the point of this testcase is to show that swig the runtime is usable from pure c++ side (example.cxx)
if I move everything to example.i it loses its goal

ojwb:
> the point of this testcase is to show that swig the runtime is usable from pure c++ side

For me the point really is to catch when we change something such that things break for bindings generated with `-external-runtime`, and all cases of such breakage I can remember were compile time errors similar to the one here.  From `git log --grep external-runtime` I found:

* 84a2c45d66f0334f8fe67077311383af11a2d5c8
* db50dc8154c5c63593e80c5cc5f96d6303dd0214
* 49b6c39d98967c5bad997bf1490deab0161b18a5 (not familiar with this one but the diff looks like it was this type of issue)

The runtime functions themselves are heavily tested by other testcases, so the only thing a testcase would really check is that they're actually exposed by `-external-runtime`, which it seems we don't manage to break.

So a testcase which worked for all (or many) target languages but which didn't do any runtime testing or for test calling runtime functions from outside the wrapper would actually catch more problems.

jschueller:
you were right, using a typemap also make the bug appear with the stable api

hovewer now I cannot seem to be able to use SWIG_Object to make it cross-language:
```
 example.cxx:9:3: error: 'SWIG_Object' was not declared in this scope; did you mean 'SwigPyObject'?
    9 |   SWIG_Object b = 0;
      |   ^~~~~~~~~~~
      |   SwigPyObject
```
any idea ?

ojwb:
`SWIG_Object` is a SWIG-level macro so the code using it needs to pass through SWIG's preprocessor.

I'll see if I can adjust the example.  My first attempt worked, but also still worked with your fix reverted so was no longer a useful regression test...

wsfulton:
Please see 124d8a8c37a8047c1cb190db24893d3bf346bb2c...61f0bbe97debe5e28d49f6ca79a384bb65a0724e committed to master. 56afa6e6de2ed3c5dd92a8e3668a3e9b5a0e9280 fixes the problem and the other two commits take the example in this patch and fixes it to use the runtime to expose the problem.

I'd also prefer to have the problem replicated in the test-suite, but the test-suite is not so flexible for invoking SWIG multiple times for one test, so keeping it as an example is a reasonable compromise. For anyone with the interest, it would be useful it is was copied and adapted to the other target languages that also support the external runtime.

jschueller:
Thanks a lot, everything is working fine now!

ojwb:
Hmm, I already had a fix for this ready to go.  I was just holding off pending us sorting out a proper regression testcase for it (which I see we still don't have).

My fix was different though:

```
diff --git a/Lib/python/pyrun.swg b/Lib/python/pyrun.swg
index 4a327547..3e474f67 100644
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -335,6 +335,18 @@ swig_varlink_setattr(PyObject *o, char *n, PyObject *p) {
   return res;
 }
 
+#if !defined(SWIGPYTHON_BUILTIN) && PY_VERSION_HEX >= 0x03030000
+# define SWIG_HEAPTYPES
+
+/* Note: Currently this won't activate - it is in place ready for when the
+ * SWIGPYTHON_BUILTIN condition above gets removed. */
+# if PY_VERSION_HEX < 0x030c0000 && defined(SWIGPYTHON_BUILTIN)
+#  include <structmember.h>
+#  define Py_READONLY READONLY
+#  define Py_T_PYSSIZET T_PYSSIZET
+# endif
+#endif
+
 SWIGINTERN PyTypeObject*
 swig_varlink_type(void) {
   static char varlink__doc__[] = "Swig var link object";
diff --git a/Lib/python/pyruntime.swg b/Lib/python/pyruntime.swg
index 6d18afa1..cdd9dd87 100644
--- a/Lib/python/pyruntime.swg
+++ b/Lib/python/pyruntime.swg
@@ -31,18 +31,6 @@
 # include <Python.h>
 #endif
 
-#if !defined(SWIGPYTHON_BUILTIN) && PY_VERSION_HEX >= 0x03030000
-# define SWIG_HEAPTYPES
-
-/* Note: Currently this won't activate - it is in place ready for when the
- * SWIGPYTHON_BUILTIN condition above gets removed. */
-# if PY_VERSION_HEX < 0x030c0000 && defined(SWIGPYTHON_BUILTIN)
-#  include <structmember.h>
-#  define Py_READONLY READONLY
-#  define Py_T_PYSSIZET T_PYSSIZET
-# endif
-#endif
-
 #if __GNUC__ >= 7
 #pragma GCC diagnostic pop
 #endif
```

The above handles it all in one place, just moving that place to `pyrun.swg` whereas 56afa6e6de2ed3c5dd92a8e3668a3e9b5a0e9280 moves the `SWIG_HEAPTYPES` defining to `pyrun.swg` but leaves the rest behind.  I think moving everything is cleaner, and show give more consistency between the normal and external runtime case.  @wsfulton Thoughts?

wsfulton:
> proper regression testcase for it (which I see we still don't have).

I've modified @jschueller's example as mentioned to create a regression test, so I'm not sure what else you are expecting? It took me a while to figure out how to recreate the actual problem, but once I had, I'd also done the work to fix it, so I committed it all together. Sorry if I stepped on your toes.

The approach I took is what I'd consider a minimal low risk change mainly for the 4.3.1 release. I've not moved the `#include` of Python headers for a patch release just in case some user code is relying on them in their current position near the start of the generated code. Ultimately I also feel they should all be together where they were in the 4.3.0 release (Python.h and structmember.h right after each other). Also with the `-external-runtime` option, the user has to `#include` the python headers themselves and my change doesn't change the way this work.

I also set up my changes so that a user can try out `SWIG_HEAPTYPES` and -builtin, if desired, by defining `SWIG_HEAPTYPES`. It makes it easy for interested users/developers to run the test-suite for working on fixes. While this combination is not officially ready as the test-suite is not 100% and we don't run CI with this, it does mostly work. I hope we can get this fully working for 4.4 and maybe we need to rethink all this once the actual solutions are available and understood where a more breaking change might be necessary for a bigger release. But if I've missed something let me know, I'd like to get it right for the 4.3.1 release which I posted about on the swig-devel list yesterday.

Issue #3059, opened 2024-10-25
Generated code using pybuffer.i, fails to compile with Py_LIMITED_API and Python-3.13


julian-smith-artifex-com:
If `Py_LIMITED_API` is defined and `pybuffer.i` is used, SWIG generated code uses `PyObject_AsReadBuffer()`, which is no longer available with Python-3.13.

This is with SWIG 4.2.1.

To reproduce:

Create file `foo.i` containing:
```
%include pybuffer.i

%pybuffer_binary(
        const unsigned char* PYTHON_BUFFER_DATA,
        size_t PYTHON_BUFFER_SIZE
        );

%{
const unsigned char* python_buffer_data(
        const unsigned char* PYTHON_BUFFER_DATA,
        size_t PYTHON_BUFFER_SIZE
        )
{
    return PYTHON_BUFFER_DATA;
}

%}

const unsigned char* python_buffer_data(
        const unsigned char* PYTHON_BUFFER_DATA,
        size_t PYTHON_BUFFER_SIZE
        );
```

Then run swig:
```
swig -Wall -c++ -python -module foo -o foo.i.cpp foo.i
```

The generated foo.`i.cpp` file has this sequence:
```
SWIGINTERN PyObject *_wrap_python_buffer_data(PyObject *self, PyObject *args) {
  PyObject *resultobj = 0;
  unsigned char *arg1 = (unsigned char *) 0 ;
  size_t arg2 ;
  PyObject *swig_obj[1] ;
  unsigned char *result = 0 ;
  
  (void)self;
  if (!args) SWIG_fail;
  swig_obj[0] = args;
  {
    int res; Py_ssize_t size = 0; const void *buf = 0;
#ifndef Py_LIMITED_API
    Py_buffer view;
    res = PyObject_GetBuffer(swig_obj[0], &view, PyBUF_CONTIG_RO);
#else
#if defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 6))
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated"
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#elif defined(_MSC_VER)
#pragma warning(push)
#pragma warning(disable: 4996)
#endif
    res = PyObject_AsReadBuffer(swig_obj[0], &buf, &size);
[...]

Issue #2826, opened 2024-03-05
Potential avenue for stable-ABI non-compliance 


HendersonSC:
I just wanted to raise this [discussion](https://github.com/trailofbits/abi3audit/issues/82) here for consideration.

In full disclosure, I am short on knowledge of python's ABI and on swig's implementation. However, since there is a bit of crossroads here and it could help prevent a real problem in the future, I figured it better to raise the alert for consideration.

Thanks for the work you do!

ojwb:
It seems Python could just solve this globally at least for new releases by doing `#define Py_DECREF(X) Py_DecRef(X)` when stable ABI is on...

As for SWIG, we could just change to use `Py_DecRef(X)` everywhere, but that imposes an overhead on people who don't care about the stable ABI so seems suboptimal.    We probably need a `SWIG_Py_DECREF(X)` which we can point at one or the other.

wsfulton:
I thought the performance impact of the functions compared to the macros would be negligible. Until I measured it. While not negligible, it is noticable. The time spent calling the Py_IncRef function is about 10% of the most simple wrapper function. Measuring the performance reliably for the macro forms, Py_INCREF and Py_XINCREF, is rather hard. For example when calling it in a loop n times and using g++ -03, the code was pretty much optimised away with one call to increment the ob_refcnt in PyObject by n instead of n increments. Using g++ -O2 was a bit more realistic and showed that the macro forms are
more than twice as fast. More than twice as fast is at least a 5% performance improvement for the simplest/fastest wrapper function, so not negligible. Hence I decided to not use the Py_IncRef function form throughout and instead provide macros as @ojwb suggests that switch to the function forms for the stable abi and the macros otherwise resulting in 84d4cc716f81695f3cc253550a175d8bb53d5b2c.

For those interested, here are the performance tests:

```
$ cat example.i
%module example

%inline %{
void Test_Py_INCREF(PyObject *o) {
  Py_INCREF(o);
}
void Test_Py_XINCREF(PyObject *o) {
  Py_XINCREF(o);
}
void Test_Py_IncRef(PyObject *o) {
  Py_IncRef(o);
}

void Test_Py_INCREF_n(PyObject *o, int n) {
  for (int i=0; i<n; ++i) {
    Py_INCREF(o);
  }
}
void Test_Py_XINCREF_n(PyObject *o, int n) {
  for (int i=0; i<n; ++i) {
    Py_XINCREF(o);
  }
}
[...]

PR #3073, opened 2024-11-07
Avoid warnings about using Py_TYPE from abi3audit

### LINKED ISSUE #3069: Python abi version mismatch ?
I tried running the Python test suite on github actions in limited api mode with versions 3.9 to 3.13 (currently only python 3.8 is enabled)
And I noticed the 3.11-3.13 jobs failed (see https://github.com/swig/swig/pull/3068)
with abi3audit errors
I tried to reproduce locally (from python 3.12 with limited api 3.9):
```
$ ./build/install/bin/swig -python Examples/python/contract/example.i
$ gcc Examples/python/contract/example_wrap.c -I/usr/include/python3.12 -DPy_LIMITED_API=0x03090000 -fPIC -shared -o example.abi3.so
$ abi3audit example.abi3.so  --verbose --summary --assume-minimum-abi3 3.9
[14:25:38] 👎 : example.abi3.so uses the Python 3.14 ABI, but is tagged for the Python 3.2 ABI                                                                                                                                              
           ┏━━━━━━━━━━━━━━━━┳━━━━━━━━━┓                                                                                                                                                                                                     
           ┃ Symbol         ┃ Version ┃                                                                                                                                                                                                     
           ┡━━━━━━━━━━━━━━━━╇━━━━━━━━━┩                                                                                                                                                                                                     
           │ Py_TYPE        │ 3.14    │                                                                                                                                                                                                     
           └────────────────┴─────────┘                                                                                                                                                                                                     
           💁 example.abi3.so: 1 extensions scanned; 1 ABI version mismatches and 0 ABI violations found
```
how should I interpret this error ?

vadz:
These warnings are spurious because Py_TYPE is an inline function and so using it doesn't prevent the extension from using stable Python ABI from earlier versions, but it is still better to avoid it, especially because this warning may become an actual error soon.

Closes #3069.

### DIFF
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -25,6 +25,37 @@
 #  define SWIG_Python_str_FromFormat PyString_FromFormat
 #endif
 
+/*
+  Py_TYPE and related functions are officially part of the limited API only
+  since 3.11, and while we don't actually break ABI compatibility by using
+  them even with the earlier versions because they are implemented as inline
+  functions in Python headers, doing this does provoke an audit failure from
+  abi3audit tool, so avoid using them if we target anything earlier than 3.11.
+ */
+#if defined(Py_LIMITED_API) && Py_LIMITED_API+0 < 0x030B0000
+#   undef Py_TYPE
+#   define Py_TYPE SWIG_Py_TYPE
+
+#   undef Py_IS_TYPE
+#   define Py_IS_TYPE SWIG_Py_IS_TYPE
+
+#   undef PyObject_TypeCheck
+#   define PyObject_TypeCheck SWIG_PyObject_TypeCheck
+
+/*
+  These functions are verbatim copies of the corresponding functions in
+  Python's own object.h.
+ */
+inline PyTypeObject *SWIG_Py_TYPE(const PyObject *ob) {
+    return ob->ob_type;
+}
+inline int SWIG_Py_IS_TYPE(PyObject *ob, PyTypeObject *type) {
+    return Py_TYPE(ob) == type;
+}
+inline int SWIG_PyObject_TypeCheck(PyObject *ob, PyTypeObject *type) {
+    return Py_IS_TYPE(ob, type) || PyType_IsSubtype(Py_TYPE(ob), type);
+}
+#endif /* !Py_LIMITED_API || Py_LIMITED_API >= 3.11 */
 
 /* Wrapper around PyUnicode_AsUTF8AndSize - call Py_XDECREF on the returned pbytes when finished with the returned string */
 SWIGINTERN const char *


jschueller:
Should we add a new CI job with Python >= 3.11 ? We only have python 3.8 / api 3.4:
```
        - SWIGLANG: python 
          VER: '3.8'
          PY_ABI_VER: '3.4'
```
Maybe python 3.13 / abi 3.9 (I dont think its relevant to test all combinations)

Lib/python/pyhead.swg
```diff
@@ -25,6 +25,37 @@
...
 #endif
 
+/*
+  Py_TYPE and related functions are officially part of the limited API only
+  since 3.11, and while we don't actually break ABI compatibility by using
+  them even with the earlier versions because they are implemented as inline
+  functions in Python headers, doing this does provoke an audit failure from
+  abi3audit tool, so avoid using them if we target anything earlier than 3.11.
+ */
+#if defined(Py_LIMITED_API) && Py_LIMITED_API+0 < 0x030B0000
```
vadz, line 35: Actually this could, and probably should, be simplified to just
```suggestion
/*
  Py_TYPE and related functions are officially part of the limited API only
  since 3.14, so avoid using them if we target anything earlier than this.
 */
#if defined(Py_LIMITED_API) && Py_LIMITED_API+0 < 0x030E0000
```
because while we may use them with 3.11, 3.12 and 3.13 without breaking the ABI, this still results in abi3audit error.

I don't know if I should push this change or if you can do it when merging? This won't affect our CI anyhow, as it only uses `PY_ABI_VER==3.4`.

vadz:
> Should we add a new CI job with Python >= 3.11 ? We only have python 3.8 / api 3.4:

Yes, I think it would be useful to have. But then there are so many combinations that we can't test all of them anyhow.

FWIW currently we have tests comparing `Py_LIMITED_API` with 3.7, 3.10 and, with this PR, 3.11 (or 3.14 if the correction above is applied).

jschueller (approved):

Lib/python/pyhead.swg
jschueller: maybe its better to avoid any warning, as I definetely want to add a CI job for python>=3.11 when I try to fix https://github.com/swig/swig/issues/3059

jschueller:
hello @vadz, did you have time to go back to this one ?

vadz:
Sorry, not sure what do you mean, i.e. what am I expected to do here?

jschueller:
your own comment is not resolved so I thought you wanted to get back to it
imho its good to go as is, so I guess the only thing left to do is ask @wsfulton to review or merge

vadz:
I didn't want to change the PR in case William disagrees with it, applying the diff above is simple if he does agree with it.

wsfulton:
I've added an abi3audit run in 9eb75b46f96c8d5902b7e9dc11e804ac45ab5861 for python-3.14 which passes as the supposed stable ABI problems flagged by abi3audit are fixed in python-3.14. There are various ABI issues and non-issues, alluded to here and also please see [my comment in #2826](https://github.com/swig/swig/issues/2826#issuecomment-2019045039). These are being addressed by the Python maintainers as the ABI evolves. 

I originally added abi3audit to the SWIG test-suite as a way to check for stable ABI conformance. However, the tool is overly restrictive and I agree they are spurious. It shows problems with what the Python developers have purposefully put in the stable ABI. It is not usable for Python versions 3.11 to 3.13, so let's just use it for versions where it is actually useful. SWIG's goal is to generate ABI conformant code (and it does) by defining `Py_LIMITED_API`. It does not generate code that is necessarily abi3audit clean, but this is not the goal nor should it be. abi3audit is still useful for versions for 3.14 and versions <= 3.10. Shame the tool doesn't provide a warning override/suppression mechanism in particular for "future_abi3_objects" it even understands.

I suggest anyone who has a goal of abi3audit clean builds incorporates the ideas in this patch using something in a `%begin` section if comfortable with using the code that the circumventions therein, or otherwise works with abi3audit to add in a warning suppression mechanism.

vadz:
> I suggest anyone who has a goal of abi3audit clean builds incorporates the ideas in this patch using something in a `%begin` section if comfortable with using the code that the circumventions therein, or otherwise works with abi3audit to add in a warning suppression mechanism.

Just for the record, it's not really trivial to put this code in any section: adding it into `%begin` doesn't compile as is, you need to include `Python.h` and to do this properly you have to reproduce the debug checks in `Lib/python/pyruntime.swg`. And adding it to `%runtime` doesn't work because it's too late: code included in this section earlier already uses `Py_TYPE()`.

So I'm finally just carrying this patch in my fork because it's much easier to have this in `pyhead.swg` than do anything else and it doesn't seem to have any drawbacks. It's a pity that it was rejected for the inclusion in SWIG itself, as this would be clearly helpful to SWIG users.

Issue #1307, opened 2018-08-15
typemapping c++ boost::optional<int> to C# int?


nanan14:
Hi ,

is there a way to typemap boost::optional<int> to c# int? and vice versa
and also for other type like bool  , string and other classes?

so far I only found this implementation.
https://stackoverflow.com/questions/15912937/can-you-swig-a-boostoptional
```
%template(optionalInt) optional<int>;
%typemap(ctype) optional<int> "void *"; 
%typemap(imtype) optional<int> "global::System.IntPtr"; 
%typemap(cstype) optional<int> "int?"; 
%typemap(csin) optional<int> "$csinput";
%typemap(out) optional<int> %{
    std::vector<int> result_vec;
    if (!!$1)
    {
        result_vec = std::vector<int>(0, $1.get());
    }
    else
    {
        result_vec = std::vector<int>();
    }
    $result = new std::vector< int >((const std::vector< int > &)result_vec); 
%}
%typemap(csout, excode=SWIGEXCODE) boost::optional<int> {
     global::System.IntPtr cPtr =$imcall;$excode
	 if (cPtr != global::System.IntPtr.Zero)
	 {
	     return (int)($imcall);
	 }
	 return null;
}
%typemap(csvarout, excode=SWIGEXCODE) optional<int>
%{ 
    get 
    { 
         global::System.IntPtr cPtr =$imcall;$excode
	    if (cPtr != global::System.IntPtr.Zero)
	    {
	        return (int)($imcall);
	    }
	    return null;
    } 
%}
```

vadz:
There are no standard typemaps for this in SWIG but I do have my own typemaps for `std::optional`-like class for C#, Java and Python (OK, so the latter one is trivial, but still). I could tidy them up for the inclusion in SWIG itself if there is sufficient interest.

aksswami:
It would be great if you can share the snippet as I am also looking for the same. I  am new to the SWIG and type mapping complex type is something I don’t understand as of now.

wsfulton:
@vadz, yes! They ought to port to `std::optional` which we ought to also have in SWIG.

diegoferigo:
@vadz Is the `std::optional` support a feature you are still working on? Can you please provide the snippet you mentioned in the meantime?

vadz:
No, sorry, other higher priority tasks took over and I was never able to work on this.

I've created some gists showing the typemaps we're using:

[...]

PR #3078, opened 2024-11-25
Added support for std::optional for C# language

christophe-calmejane:
std::optional support for C#

### DIFF
--- a/Examples/test-suite/cpp17_optional.i
+++ b/Examples/test-suite/cpp17_optional.i
@@ -13,7 +13,8 @@
 
 // Enum types with different underlying types
 %optional_enum_uint8(test::SmallEnum)
-%optional_enum_int64(test::BigEnum)
+%optional_enum_uint64(test::BigEnum)
+%optional_primitive(AliasEnum)
 
 %feature("director") test::TestObjectDirected;
 
@@ -28,6 +29,7 @@ struct NoNamespaceStruct
     int a;
 };
 using NoNamespaceStructOptional = std::optional<NoNamespaceStruct>;
+using AliasEnum = std::int64_t;
 
 namespace test
 {
@@ -39,11 +41,11 @@ namespace test
         Value3 = 255
     };
 
-    enum class BigEnum : std::int64_t
+    enum class BigEnum : std::uint64_t
     {
-        Small = -1000000000000LL,
+        Small = 10ULL,
         Zero = 0,
-        Large = 1000000000000LL
+        Large = 1000000000000ULL
     };
 
     // Example using simple type (eg. std::uint32_t)
@@ -65,6 +67,8 @@ namespace test
         std::uint32_t a;
         std::uint32_t b;
         SubStructOptional sub;
+        std::optional<BigEnum> bigEnumOpt;
+        std::optional<AliasEnum> aliasEnumOpt;
     };
     using StructOptional = std::optional<Struct>;
 
--- a/Examples/test-suite/cpp17_optional.i
+++ b/Examples/test-suite/cpp17_optional.i
@@ -1,6 +1,7 @@
 %module(directors="1") cpp17_optional
 #if defined SWIGCSHARP
 #define SWIG_STD_OPTIONAL_DEFAULT_TYPES
+%include <stdint.i>
 %include <std_optional.i>
 
 %optional(test::Struct)
@@ -121,7 +122,7 @@ namespace test
         void               setInt(int i) { _numberInt = i; }
         std::optional<int> getIntOpt() const { return _numberIntOpt; }
         void               setIntOpt(std::optional<int> i) { _numberIntOpt = i; }
-        // Flaot optional testing
+        // Float optional testing
         float                getFloat() const { return _numberFloat; }
         void                 setFloat(float f) { _numberFloat = f; }
         std::optional<float> getFloatOpt() const { return _numberFloatOpt; }
--- a/Examples/test-suite/cpp17_optional.i
+++ b/Examples/test-suite/cpp17_optional.i
@@ -1,0 +1,317 @@
+%module(directors="1") cpp17_optional
+#if defined SWIGCSHARP
+%include <std_optional.i>
+
+%optional(test::Struct)
+%optional(test::Point)
+%optional(test::Circle)
+%optional(test::Rect)
+
+%feature("director") test::TestObjectDirected;
+
+%inline %{
+
+#include <cstdint>
+#include <optional>
+#include <string>
+
+struct NoNamespaceStruct
+{
+    int a;
+};
+using NoNamespaceStructOptional = std::optional<NoNamespaceStruct>;
+
+namespace test
+{
+    // Example using simple type (eg. std::uint32_t)
+    using SimpleOptional = std::optional<std::uint32_t>;
+
+    // Example using a string
+    using StringOptional = std::optional<std::string>;
+
+    // Example using a struct
+    struct Struct
+    {
+        std::uint32_t a;
+        std::uint32_t b;
+    };
+    using StructOptional = std::optional<Struct>;
+
+
+    // Define a class that uses all the above types (input and output parameters)
+    class TestOptionals
+    {
+    public:
+        TestOptionals() = default;
+
+        SimpleOptional getSimpleOptional() const
+        {
+            return _simpleOptional;
+        }
+        void setSimpleOptional(SimpleOptional const value)
+        {
+            _simpleOptional = value;
+        }
+
+        StringOptional getStringOptional() const
+        {
+            return _stringOptional;
+        }
+        void setStringOptional(StringOptional const& value)
+        {
+            _stringOptional = value;
+        }
+
+        StructOptional getStructOptional() const
+        {
+            return _structOptional;
+        }
+        void setStructOptional(StructOptional const& value)
+        {
+            _structOptional = value;
+        }
+
+    private:
+        SimpleOptional _simpleOptional {};
+        StringOptional _stringOptional {};
+        StructOptional _structOptional {};
+    };
+
+    // Examples/test-suite/cpp17_optional_primitive_types.i
+    class TestObjectPrimitives
+    {
+    
+    public:
+        TestObjectPrimitives(bool done, int i, float f, double d)
+            : _isDone(done)
+            , _numberInt(i)
+            , _numberFloat(f)
+            , _numberDouble(d)
+            , _isDoneOpt(done)
+            , _numberIntOpt(i)
+            , _numberFloatOpt(f)
+            , _numberDoubleOpt(d)
+            , _isNullopt(std::nullopt)
+        { }
+    
+        // Bool optional testing
+        bool                getBool() const { return _isDone; }
+        void                setBool(bool b) { _isDone = b; }
+        std::optional<bool> getBoolOpt() const { return _isDoneOpt; }
+        void                setBoolOpt(std::optional<bool> b) { _isDoneOpt = b; }
+        // Int optional testing
+        int                getInt() const { return _numberInt; }
+        void               setInt(int i) { _numberInt = i; }
+        std::optional<int> getIntOpt() const { return _numberIntOpt; }
+        void               setIntOpt(std::optional<int> i) { _numberIntOpt = i; }
+        // Flaot optional testing
+        float                getFloat() const { return _numberFloat; }
+        void                 setFloat(float f) { _numberFloat = f; }
+        std::optional<float> getFloatOpt() const { return _numberFloatOpt; }
+        void                 setFloatOpt(std::optional<float> f) { _numberFloatOpt = f; }
+        // Double optional testing
+        double                getDouble() const { return _numberDouble; }
+        void                  setDouble(double d) { _numberDouble = d; }
+        std::optional<double> getDoubleOpt() const { return _numberDoubleOpt; }
+        void                  setDoubleOpt(std::optional<double> d) { _numberDoubleOpt = d; }
+    
+        std::optional<bool> getNullopt() const { return _isNullopt; }
+    
+    private:
+        bool   _isDone;
+        int    _numberInt;
+        float  _numberFloat;
+        double _numberDouble;
+    
+        std::optional<bool>   _isDoneOpt;
+        std::optional<int>    _numberIntOpt;
+        std::optional<float>  _numberFloatOpt;
+        std::optional<double> _numberDoubleOpt;
+    
+        std::optional<bool> _isNullopt;
+    };
+    
+    // Examples/test-suite/cpp17_optional_std_types.i
+    class TestObjectString
+    {
+    
+    public:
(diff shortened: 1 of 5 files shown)


vadz (changes requested):
I was curious to see how this worked because, as I wrote, I had my own typemaps for an optional-like class, but after a brief look it looks like this is pretty much based on my own typemaps (it could have been nice to mention this...) except that this doesn't use C# nullable types AFAICS, `cstype` for `optional<T>` is just `T` and not `T?` as in my version.

Unsurprisingly, I think this is wrong, and that the natural representation of `optional<T>` is `T?` and not (possibly null) `T` because the former allows working with them more conveniently at C# level.

IOW there is clearly a lot of useful stuff here (my typemaps didn't include director support), but I think this shouldn't be merged as is and the changes to my original typemaps replacing `T` with `T` should be reverted.

christophe-calmejane:
> I was curious to see how this worked because, as I wrote, I had my own typemaps for an optional-like class, but after a brief look it looks like this is pretty much based on my own typemaps (it could have been nice to mention this...)

I've seen quite a lot of tentative implementation of optional that all look pretty much the same, not sure if *yours* is shared somewhere but after testing quite a lot of solutions to be as clean as possible, this is what came up entirely rewritten based on an implementation I made quite a few years ago.

> except that this doesn't use C# nullable types AFAICS, `cstype` for `optional<T>` is just `T` and not `T?` as in my version.
> Unsurprisingly, I think this is wrong, and that the natural representation of `optional<T>` is `T?` and not (possibly null) `T` because the former allows working with them more conveniently at C# level.

Indeed, this is a last minute change to acommodate with Mono restriction that doesn't understand this (needs C#>=8), but then I forgot it is not a C# nullable reference anymore. I guess I need to find a way to properly switch from one or other implementation, certainly with a define. But then how to have proper unit testing since we only have one version of the C# compiler?

Lib/csharp/std_optional.i
```diff
@@ -0,0 +1,256 @@
...
+%typemap(csin) std::optional<std::string> * "$csinput"
+%typemap(csout, excode=SWIGEXCODE) std::optional<std::string> * {
+    string ret = $imcall;$excode
+    return ret;
+  }
+
+%enddef
+
+
+#if !defined(SWIG_NO_STD_OPTIONAL_DEFAULT_TYPES)
```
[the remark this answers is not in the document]
christophe-calmejane: If I include std_optional.i I would expect it to just work by default for basic types, but I'm open to discussion. This is much like including cstdint, you have all types defined directly. Either way it's a decision to be taken (or is there some precedence we can rely on?)

vadz:
> I've seen quite a lot of tentative implementation of optional that all look pretty much the same, not sure if _yours_ is shared somewhere but after testing quite a lot of solutions to be as clean as possible, this is what came up entirely rewritten based on an implementation I made quite a few years ago.

Well, this contains several comments that I wrote verbatim (which is how I recognized it, the writing style looked familiar), so I have little doubts that this is based on [my typemaps](https://gist.github.com/vadz/c6a9cf0c62d8db7ad687992735a9ae08) originally written in 2013 which I posted [at least here](https://github.com/swig/swig/issues/1307#issuecomment-469258404) and I believe in some other places, so you can easily compare the two versions.

> Indeed, this is a last minute change to acommodate with Mono restriction that doesn't understand this (needs C#>=8)

I don't understand it, as I wrote on the mailing list, we definitely didn't use C# 8 in 2013 and yet my typemaps worked just fine with whatever version we did use back then. Why do you think C# 8 is needed?

Lib/csharp/std_optional.i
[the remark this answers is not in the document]
vadz: To me, this is similar to `std_vector.i` predefining `vector<int>` etc — which is actually also done for C# currently but was IMO a mistake and just creates a gratuitous difference between it and the other languages.

christophe-calmejane:
> > I've seen quite a lot of tentative implementation of optional that all look pretty much the same, not sure if _yours_ is shared somewhere but after testing quite a lot of solutions to be as clean as possible, this is what came up entirely rewritten based on an implementation I made quite a few years ago.
> 
> Well, this contains several comments that I wrote verbatim (which is how I recognized it, the writing style looked familiar), so I have little doubts that this is based on [my typemaps](https://gist.github.com/vadz/c6a9cf0c62d8db7ad687992735a9ae08) originally written in 2013 which I posted [at least here](https://github.com/swig/swig/issues/1307#issuecomment-469258404) and I believe in some other places, so you can easily compare the two versions.

Indeed you are right. This is a 5y long work that was started by an old colleague, I didn't know he started (or was inspired) from your work. Sorry about that I didn't know when I took over as it wasn't mentioned. I'm curious though if it's been so long since you did that work, why isn't it in the official repository?

christophe-calmejane:
> > Indeed, this is a last minute change to acommodate with Mono restriction that doesn't understand this (needs C#>=8)
> 
> I don't understand it, as I wrote on the mailing list, we definitely didn't use C# 8 in 2013 and yet my typemaps worked just fine with whatever version we did use back then. Why do you think C# 8 is needed?

Well, I'm not at all a C# expert so I'm not entirely sure what is wrong. When I try to use nullable reference like you suggest, I get CI errors like this:
```
cpp17_optional/TestOptionals.cs(85,10): error CS0453: The type `cpp17_optionalNamespace.Struct' must be a non-nullable value type in order to use it as type parameter `T' in the generic type or method `System.Nullable<T>'
cpp17_optional/TestOptionals.cs(91,33): error CS0453: The type `cpp17_optionalNamespace.Struct' must be a non-nullable value type in order to use it as type parameter `T' in the generic type or method `System.Nullable<T>'
```

I don't get any of these errors when compiling a C# project (outside the CI) that uses a recent C# compiler (ie. dotnet), and also if I try to force a C#<8 in my project I correctly get this error:
```
Program.cs(71,42): error CS8107: Feature 'nullable reference types' is not available in C# 7.0. Please use language version 8.0 or greater.
```

That's where I assumed Mono is not implementing C#>=8, or at least not nullable reference. Again, not an expert, so it might just be something totally different, but still everything is working fine when I define the typemap as nullable reference for all my projects, just the CI that don't accept it.
I'd be glad to revert to nullable reference as it was the original plan, so if you have any clue what is wrong, feel free to advice.

On a side note, I only have access to macOS for testing, it's a quite a pain to rely on the CI for testing small changes (although for compilation only and not running it might be sufficient. I'll try to setup things again locally for that purpose).

christophe-calmejane:
Also, there is one more thing that I cannot seem to fix when using nullable references, I get this warning with my modern C# compiler:
```
warning CS8669: The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source.
```

And I don't know how to add the requested `#nullable enable` directive using typemap (adding a csimports typemap has no effect, I don't know if there is a way to add things to the generated .cs files). Ofc this directive makes no sense for a mono compiler (might even be an error), this the reason to be able to target C#<8 or C#>=8

christophe-calmejane:
@wsfulton @ojwb This PR has one remaining issue in this code https://github.com/swig/swig/blob/8fd4df01bd4111e5f7322fe4f9d4817ffd6f0c6d/Source/Modules/csharp.cxx#L4252-L4257

When using a `nullable reference type` this type of line is generated:
```csharp
private static global::System.Type[] swigMethodTypes1 = new global::System.Type[] { typeof(Rect?) };
```
which fails to compile with this error
```
error CS8639: The typeof operator cannot be used on a nullable reference type
```

I tested by manually changing the generated line to `typeof(Rect)` and it is working just fine and `SwigDerivedClassHasMethod` is happy.

Can you hint me to what needs to be changed in `csharp.cxx` module to properly detect the type is a nullable reference and remove the `?` (but only for this, we still want to have `?` for basic types and structs).
Thanks

Examples/test-suite/cpp17_optional.i
```diff
@@ -0,0 +1,334 @@
...
+        bool                getBool() const { return _isDone; }
+        void                setBool(bool b) { _isDone = b; }
+        std::optional<bool> getBoolOpt() const { return _isDoneOpt; }
+        void                setBoolOpt(std::optional<bool> b) { _isDoneOpt = b; }
+        // Int optional testing
+        int                getInt() const { return _numberInt; }
+        void               setInt(int i) { _numberInt = i; }
+        std::optional<int> getIntOpt() const { return _numberIntOpt; }
+        void               setIntOpt(std::optional<int> i) { _numberIntOpt = i; }
+        // Flaot optional testing
```
ojwb, line 124: Typo: "Flaot" -> "Float"

Lib/csharp/std_optional.i
```diff
@@ -0,0 +1,329 @@
...
+#define SWIG_STD_OPTIONAL_STRINGIFY_(x) #x
+#define SWIG_STD_OPTIONAL_STRINGIFY(x) SWIG_STD_OPTIONAL_STRINGIFY_(x)
+#endif
+
+
+%{
+    #include <optional>
+%}
+
+%include <stdint.i>
```
ojwb, line 58: Implicitly including `stdint.i` opens something of a can of worms.  By default SWIG aims to generate code independent of the platform it is run on, which can then be built anywhere.  A few places violate this, and `stdint.i` is one of them as the code generated is tailored to the system it will be built on (the problem is that for most target languages SWIG needs to work in terms of `int`, `long`, `long long`, etc which of those e.g. `int64_t` corresponds to depends on the platform).

So I think it would be better to avoid doing this by default.

ojwb:
> Can you hint me to what needs to be changed in `csharp.cxx` module to properly detect the type is a nullable reference and remove the `?` (but only for this, we still want to have `?` for basic types and structs).

I think this is addressed by #3114?

Lib/csharp/std_optional.i
AdamLithgow: @ojwb as someone less familiar with SWIG that's looking to help close this out, what would be preferred to use instead of `stdint.i` in cases like this where numeric types would need to be translated?

christophe-calmejane: Oh dear I didn't get the notification on April 2nd when Olly reviewed the PR. I can try without the inclusion and see if that works. I've been using this code for more than a year in multiple projects now.

christophe-calmejane:
> > Can you hint me to what needs to be changed in `csharp.cxx` module to properly detect the type is a nullable reference and remove the `?` (but only for this, we still want to have `?` for basic types and structs).
> 
> I think this is addressed by #3114?

Yes indeed, we worked together on these PR and decided to create a separate one to fix the issue as it can be fixed independently to this one.

Lib/csharp/std_optional.i
christophe-calmejane: Ok, I didn't notice any change in the generated code of many of my project, so I'll update the file without including stdint.

christophe-calmejane:
Hi, is there still any blocker for this one?
I'd like to add optional for lua, but it depends on this one, and on directors for lua. It's a bit complicated to handle all theses branches. Thanks

wsfulton:
I think I've seen a bunch of other `std::optional` merge requests which ought to be looked at together with this one to ensure we have a common set of features for across all the target languages. I'll get up to speed on this merge request next.

@vadz, any objections to using this merge request over any of your older patches (that this work was derived from)?

vadz:
No objections for reusing the typemaps I originally wrote in any way, of course, but I'd like to repeat that I think it would be much better, from usability point of view, to represent `optional<T>` as `T?` for the simple types.

FWIW I have `optional` typemaps for C, C#, Java, JS and Python that work well for me, but they don't provide directors support. I could still share the latest version with you somehow (send them to you directly? open a draft PR with them?) if you think this could be useful.

christophe-calmejane:
As I explained in https://github.com/swig/swig/pull/3078#issuecomment-2500182289 , changing to `T?` like the original code is causing CI to fail. I don't know what is wrong nor how to solve it for it to work both on CI and real projects.

If anybody has any clue on how to solve it, feel free to edit or suggest changes. I've been using it like this in multiple C# projects for 4y now and it works fine (except for the less-optimal simple types) if used along side #3114 .

vadz:
But less optimal simply types are a big problem, our C# API works with them since 15 years and it would be a huge step back to require all its users to use much less convenient code (without even speaking of requiring them to change something working perfectly well at all).

christophe-calmejane:
I've spent the last 2 days trying to find a solution for satisfy this but it's very tricky as there is always something breaking (either enums, directors, nullable ref, ...). The most difficult part is solving directors handling for enums (I absolutely need it) that #3114 is not solving, but I may have a better solution that #3114 (I worked with the author of that PR and it was made for this one) by using a new attribute (so we only rely on what the typemap declares instead of checking types in swig, which doesn't handle enums).

I hope I can find a solution, otherwise I still think it's better to have support for optionals, even if not optimal for primitive types, than no support at all (we'll have the option to improve it in a next release). The good thing is you said you already have it in use for 15y, so you won't be forced to use this one if it breaks your codebase, as you probably have your own file anyway.

vadz:
> The good thing is you said you already have it in use for 15y, so you won't be forced to use this one if it breaks your codebase, as you probably have your own file anyway.

Yes, I do, and I have no problems merging SWIG master into my branch as long as SWIG doesn't have its own typemaps for optional, but when it adds them, I'm going to have conflicts galore and it will be really painful to resolve them.

I still didn't have time to look into this, but I don't understand what is the fundamental problem of defining typemaps differently for the primitive types (using `T?`) and not?

christophe-calmejane:
There are quite a few issues, I don't remember exactly all of them, that's why I created a quite important test-suite for optionals (more incoming). But mainly the problem is related to directors and enums (marshaling issues), and correctly supporting both C#<8 and C#>8 (passing the CI is something, real world applications is something else).

Let's see if my latest improvements are passing the CI.

christophe-calmejane:
So, I've:
* Rebased to master (otherwise the CI doesn't run anymore due to github deprecated ubuntu-20.04)
* Fully rewritten the primitive and enums types macros, so it work on the CI (C# < 8) and recent C# compilers with `nullable reference types` support
* Added more tests to check all kind of directors usage
* Incorporated a different approach to #3114 that work for primitive types **and** enums, meaning #3114 is no longer needed
* Tried to beautify my changes in the cxx file by following the guidelines, but it didn't run (`../IndentBaks directory does not exist`), so I hope I got it right this time
* Updated my own projects using the first version of this PR to match the new macros and confirm it's working fine

I hope everybody will be pleased with this new version.

wsfulton:
> Tried to beautify my changes in the cxx file by following the guidelines, but it didn't run (../IndentBaks directory does not exist), so I hope I got it right this time

The error is telling you to make the directory (it's a check).

I really need latest dotnet (>8) working with our test-suite to progress this. I started on this a while back, got a fairly good idea of what to change in the build system and then got distracted. Anyway, I'll finish it off and then get back to this pull request.

PR #3089, opened 2024-12-12
Correct docs for Python abi3 mode

larsoner:
# Doc update

This PR updates the docs based on my experience working with the new `abi3` mode from https://github.com/swig/swig/pull/2727. It might also qualify as a bug report, which I can separate out (see below :point_down: ) but want to mention here to explain the necessity of the second point:

1. The `-fast*` option that is specifically problematic in `-O` is `-fastproxy` (I found this by trial-and-error so I might be wrong)
2. Instead of `-O` you should use `-fastdispatch` if you use typemaps (and it should be abi3-compatible, or at least is according to `abi3audit` for me!)

cc @wsfulton since I think you wrote those docs and worked on the abi3 stuff.

# Possible bug

Regarding the need for `-fastdispatch`, I use swig over in OpenMEEG with a some simple typemaps like:

https://github.com/openmeeg/openmeeg/blob/4cf760b23f88ddb0523b66235e01debe2795bd42/wrapping/python/openmeeg/_openmeeg.i#L357-L362

If I `swig` with `-O`, everything is fine. According to the docs, I can't do that now that I'm trying to use `abi3` (and I do see missing symbols during compilation as suggested!), so I removed that option. But when I did that, I started getting duck-typing errors in my tests, where I could no longer instantiate objects that I could before, almost as if the typemap didn't exist anymore:
```
FAILED ../venv/lib/python3.10/site-packages/openmeeg/tests/test_vector.py::test_vector - TypeError: Wrong number or type of arguments for overloaded function 'new_Vector'.
  Possible C/C++ prototypes are:
    OpenMEEG::Vector::Vector()
    OpenMEEG::Vector::Vector(OpenMEEG::Dimension const)
    OpenMEEG::Vector::Vector(OpenMEEG::Vector const &,OpenMEEG::DeepCopy const)
    OpenMEEG::Vector::Vector(OpenMEEG::Matrix const &)
    OpenMEEG::Vector::Vector(OpenMEEG::SymMatrix const &)
    OpenMEEG::Vector::Vector(PyObject *)
```
So then by trial and error I started putting back some `-O` options to see if anything fixed things for me, and it looks like the one that works is:
```
     -fastdispatch   - Enable fast dispatch mode to produce faster overload dispatcher code
```
You can see it working in the Linux run of the [CI job](https://github.com/openmeeg/openmeeg/actions/runs/12303396171/job/34338423522?pr=708) where [I set `-fastdispatch`](https://github.com/openmeeg/openmeeg/pull/708/commits/36630de943339185f0567066c288cef143af2817) but failing in the [CI job](https://github.com/openmeeg/openmeeg/actions/runs/12303318513/job/34338169903?pr=708#step:7:1222) where I [remove the switch](https://github.com/openmeeg/openmeeg/pull/708/commits/16e8b768bfa1399c74edaafbe3ef66da58947275).

It seems like a potential bug that this needs to be enabled for my type mapping to work. From this description it sounds like it will only affect the *speed* of produced code rather than typemap functionality.

If this is a bug that needs to be fixed in `swig`, I'm happy to open a new issue but I'd be way out of my depth to try to fix it.

If it's not a bug, then maybe an additional note to the `-fastdispatch` command line option either in the `swig -python --help` or `swig --help` would be warranted to make it clearer it's not just a speed switch but also affects functionality?

FYI this part is not `abi3`-specific -- this possibly errant typemap behavior based on the presence or absence of `-fastdispatch` happens either way.

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -4136,7 +4136,7 @@ or simply define the macro using your C/C++ compiler's <tt>-D</tt> command line 
 The default SWIG command line options generate code that enable the limited API/stable ABI.
 Some options, such as <tt>-builtin</tt>, <tt>-fastproxy</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
 Compatibility of user's custom typemaps is of course dependent on the Python APIs used in the typemaps.
-Use of typemaps requires passing the <tt>-fastdispatch</tt> option to SWIG (which is typically otherwise enabled by the abi3-incompatible <tt>-O</tt>).
+If you disable <tt>-O</tt>, you could consider enabling <tt>-fastdispatch</tt> as this option is abi3 compatible.
 </p>
 
 <p>
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -4135,8 +4135,8 @@ or simply define the macro using your C/C++ compiler's <tt>-D</tt> command line 
 <p>
 The default SWIG command line options generate code that enable the limited API/stable ABI.
 Some options, such as <tt>-builtin</tt>, <tt>-fastproxy</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
+Limited API is only guaranteed to work with the default code generation options.
 Compatibility of user's custom typemaps is of course dependent on the Python APIs used in the typemaps.
-If you disable <tt>-O</tt>, you could consider enabling <tt>-fastdispatch</tt> as this option is abi3 compatible.
 </p>
 
 <p>
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -4134,8 +4134,9 @@ or simply define the macro using your C/C++ compiler's <tt>-D</tt> command line 
 
 <p>
 The default SWIG command line options generate code that enable the limited API/stable ABI.
-Some options, such as <tt>-builtin</tt>, <tt>-fast</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
+Some options, such as <tt>-builtin</tt>, <tt>-fastproxy</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
 Compatibility of user's custom typemaps is of course dependent on the Python APIs used in the typemaps.
+Use of typemaps requires passing the <tt>-fastdispatch</tt> option to SWIG (which is typically otherwise enabled by the abi3-incompatible <tt>-O</tt>).
 </p>
 
 <p>


Doc/Manual/Python.html
```diff
@@ -4134,8 +4134,9 @@ <H3><a name="Python_stable_abi">33.6.6 Stable ABI</a></H3>
 
 <p>
 The default SWIG command line options generate code that enable the limited API/stable ABI.
-Some options, such as <tt>-builtin</tt>, <tt>-fast</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
+Some options, such as <tt>-builtin</tt>, <tt>-fastproxy</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
```
wsfulton, line 4137: This is a correct typo correction.

Doc/Manual/Python.html
```diff
@@ -4134,8 +4134,9 @@ <H3><a name="Python_stable_abi">33.6.6 Stable ABI</a></H3>
...
 Compatibility of user's custom typemaps is of course dependent on the Python APIs used in the typemaps.
+Use of typemaps requires passing the <tt>-fastdispatch</tt> option to SWIG (which is typically otherwise enabled by the abi3-incompatible <tt>-O</tt>).
```
wsfulton, line 4139: If you provide your own typemaps then it is up to you to ensure it works with or without `-fastdispatch`, so this statement is wrong.

I'm not sure what your problem is given the information in this issue. Please provide a small cutdown testcase if you need help writing a typemap.

larsoner: > If you provide your own typemaps then it is up to you to ensure it works with or without -fastdispatch, so this statement is wrong.

The problematic typemap is [very simple](https://github.com/openmeeg/openmeeg/blob/4cf760b23f88ddb0523b66235e01debe2795bd42/wrapping/python/openmeeg/_openmeeg.i#L357-L362):
```
    %typemap(in) Vector& {
        $1 = new_OpenMEEG_Vector($input);
    }
```
Which just calls [this little helper](https://github.com/openmeeg/openmeeg/blob/4cf760b23f88ddb0523b66235e01debe2795bd42/wrapping/python/openmeeg/_openmeeg.i#L201-L218). Would you expect this to work?

I could try to trim the example but I am far from a `swig` expert. Do you happen to have some tiny sample project I could start from and build out? Going the opposite direction starting from OpenMEEG and cutting down would be a nightmare :sweat_smile:

wsfulton: The SWIG developers do not know OpenMEEG, you really ought to ask them for help.

larsoner: Sorry I thought that would be simple enough to have a sense of whether this could be a SWIG problem rather than an OpenMEEG one. Any thoughts/ideas on my second question regarding:

> Do you happen to have some tiny sample project I could start from and build out?

wsfulton: I'm not sure what `Vector` is but if it is anything like `std::vector`, then simply start with:
```
%include <std_vector.i>
```

Doc/Manual/Python.html
```diff
@@ -4134,8 +4134,9 @@ <H3><a name="Python_stable_abi">33.6.6 Stable ABI</a></H3>
 
 <p>
 The default SWIG command line options generate code that enable the limited API/stable ABI.
-Some options, such as <tt>-builtin</tt>, <tt>-fast</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
+Some options, such as <tt>-builtin</tt>, <tt>-fastproxy</tt> (used by <tt>-O</tt>) do not use the limited API and hence when <tt>Py_LIMITED_API</tt> is defined there will be missing symbols during compilation.
 Compatibility of user's custom typemaps is of course dependent on the Python APIs used in the typemaps.
+If you disable <tt>-O</tt>, you could consider enabling <tt>-fastdispatch</tt> as this option is abi3 compatible.
```
larsoner, line 4139: @wsfulton removed the part about needing `-fastdispatch`, but left in some mention of it. Do you think this is worthwhile?

ojwb: I think it would be helpful to document a complete list of options which don't work with the limited API (the current "Some options, such as [...]" is useful in telling the user some options to avoid, but it does suggest the list is incomplete).

It's clear you can't use `-O` because of `-fastproxy`.  It seems `-fastdispatch` is OK, but for example what about `-fvirtual` (the other option currently enabled by `-O`)?  And are any other options broken with the limited API?

If we had a list of options we expect to work we could have a CI build which tries to use them all with the limited API (or at least uses most if some are mutually incompatible).

wsfulton: As far as I know the limited api only works with the default optimisations. We don't test it for any other options but I know it does not work with some of the important flags like `-builtin`. It is a *limited api* after all and is quite restricted to what python api calls can be made!

It is probably possible to get the limited api to work with the various other flags and combinations with restrictions longer time. Consider it a project for someone to take on board. Additionally, the limited api is historically too limited and is constantly being expanded by the Python developers. In particular python-3.14 has a much improved limited api. I think the many rough edges of the limited api with the default  SWIG flags should be ironed out for 3.14 as a priority and is where I will try and spend my time in this area. Right now CI testing won't work for different versions of Python, despite it being supposedly version agnostic! These need sorting out first but work for `-O`, `-builtin` etc needs to be done too.

Regarding this documentation patch, I would add to the limited api section that it is only known to work with the default code generation.

larsoner: > Regarding this documentation patch, I would add to the limited api section that it is only known to work with the default code generation.

Okay pushed that change!

wsfulton:
Merged see 52ceedd2a8bf894fa6b183ab29a58f3ded5914e7 and ea00382379e27f602dbc4ee8d1a5caa6450711b8

PR #3105, opened 2025-01-14
Fix redeclaration error in std_string_view.i for python

gzzi:
When using multiple std::string_view as argument of a function, gcc complain that variable are redeclared. see error below:

```
libpctool_wrap.cxx: In function 'PyObject* _wrap_new_DeviceView(PyObject*, PyObject*)':
libpctool_wrap.cxx:4867:14: error: redeclaration of 'Py_ssize_t len'
 4867 |   Py_ssize_t len;
      |              ^~~
libpctool_wrap.cxx:4848:14: note: 'Py_ssize_t len' previously declared here
 4848 |   Py_ssize_t len;
      |              ^~~
libpctool_wrap.cxx:4873:15: error: redeclaration of 'const char* p'
 4873 |   const char *p;
      |               ^
libpctool_wrap.cxx:4854:15: note: 'const char* p' previously declared here
 4854 |   const char *p;
      |               ^
libpctool_wrap.cxx:4886:14: error: redeclaration of 'Py_ssize_t len'
 4886 |   Py_ssize_t len;
      |              ^~~
libpctool_wrap.cxx:4848:14: note: 'Py_ssize_t len' previously declared here
 4848 |   Py_ssize_t len;
      |              ^~~
libpctool_wrap.cxx:4892:15: error: redeclaration of 'const char* p'
 4892 |   const char *p;
      |               ^
libpctool_wrap.cxx:4854:15: note: 'const char* p' previously declared here
 4854 |   const char *p;
      |               ^
ninja: build stopped: subcommand failed.
```

This change use various names for each argument.

### DIFF
--- a/Lib/python/std_string_view.i
+++ b/Lib/python/std_string_view.i
@@ -29,23 +29,23 @@ namespace std {
     %}
 
     %typemap(in) string_view (PyObject *bytes = NULL) {
-        Py_ssize_t len;
+        Py_ssize_t $1_len;
 %#ifdef SWIG_PYTHON_STRICT_BYTE_CHAR
-        const char *p = PyBytes_AsString($input);
-        if (!p) SWIG_fail;
-        len = PyBytes_Size($input);
+        const char *$1_p = PyBytes_AsString($input);
+        if (!$1_p) SWIG_fail;
+        $1_len = PyBytes_Size($input);
 %#else
-        const char *p;
+        const char *$1_p;
         if (PyUnicode_Check($input)) {
-          p = SWIG_PyUnicode_AsUTF8AndSize($input, &len, &bytes);
-          if (!p) SWIG_fail;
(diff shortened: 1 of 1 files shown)


ojwb:
There's already a testcase for this in `Examples/test-suite/cpp17_string_view.i` which was added as a regression test when this problem was fixed in 6226d58065005a31f3737530ae93eaa628ae4995 (which was in 4.3.0).

@gzzi Are you sure this problem is reproducible with git master (or SWIG >= 4.3.0)?

I'm wondering if you hit this with e.g. SWIG 4.2.0, patched it locally, and then rebased your fix onto git master to submit without spotting that the change of delimiters had already addressed it in a different way...

gzzi:
My bad, I confirm it's fixed in 4.3.0 and the test added in https://github.com/swig/swig/commit/6226d58065005a31f3737530ae93eaa628ae4995 cover my use case. I close the PR.

PR #3112, opened 2025-01-27
Fix crash when using directors in Python without limited API

### LINKED ISSUE #2889: Director methods using 'PyThread_allocate_lock' will crash at runtime (Regression in SWIG 4.2.1)
SWIG 4.2.1 or more specifically this commit (21f1c923ca5a2065ffcfea5fd4f01af8811e54d1) will result in crashes or deadlocks at runtime, when the generated code actually gets executed. Reason for this is, that this wrapper class actually confuses the functions `PyThread_release_lock` and `PyThread_free_lock`:

```
class Mutex
   {
   public:
       Mutex() {
           mutex_ = PyThread_allocate_lock();
       }

       ~Mutex() {
           PyThread_release_lock(mutex_);
       }

   private:
       void Lock() {
           PyThread_acquire_lock(mutex_, WAIT_LOCK);
       }

       void Unlock() {
           PyThread_free_lock(mutex_);
       }

       PyThread_type_lock mutex_;

       friend class Guard;
   };
```

`PyThread_free_lock` should be called in the destructor and `PyThread_release_lock` in the `Unlock` function. The way it is right now will actually free/delete the lock object upon the first call to `Unlock`! Everything after that is undefined behaviour.

vadz:
Call correct functions when unlocking/destroying a mutex implemented using Python API: they were accidentally exchanged in 21f1c923c (Make directors implementation for Python work with limited API, 2019-12-23) resulting in crashes when using directors and threads together.

Closes #2889.

### DIFF
--- a/Lib/python/director_py_mutex.swg
+++ b/Lib/python/director_py_mutex.swg
@@ -16,9 +16,9 @@ namespace Swig {
        PyThread_type_lock mutex_;
    public:
        Mutex() : mutex_(PyThread_allocate_lock()) {}
-       ~Mutex() { PyThread_release_lock(mutex_); }
+       ~Mutex() { PyThread_free_lock(mutex_); }
        void lock() { PyThread_acquire_lock(mutex_, WAIT_LOCK); }
-       void unlock() { PyThread_free_lock(mutex_); }
+       void unlock() { PyThread_release_lock(mutex_); }
    };
 }
 #endif


vadz:
@wsfulton OK to push this?

ojwb:
LGTM but deserves a CHANGES.current entry.

vadz:
I hesitated about it, what about

```diff
diff --git a/CHANGES.current b/CHANGES.current
index f804f39e4..f366a4f74 100644
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,10 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.4.0 (in progress)
 ===========================
 
+2025-01-27: StefanBattmer
+	    [Python] #2889 Fix regression when using directors and threads
+	    but not using limited API.
+
 2024-12-07: arbrauns
 	    [Lua] #3083 Fix "unsigned long long" being interpreted as "signed
 	    long long".
```

? I can push it along with this commit.

ojwb:
Sounds OK

StefanBattmer:
Thanks for dealing with this guys!

vadz:
Pushed now, see the commit above.

PR #3114, opened 2025-01-28
C#: Check nullable reference types and omit ? during typeof 

### LINKED ISSUE #3108: C# generator creates invalid code when nullable reference type is used in director method
While trying to add support for `std::optional` for C#, specifically for director support (see https://github.com/swig/swig/pull/3078) I found an issue.
The C# generator creates invalid C# code here:

https://github.com/swig/swig/blob/8fd4df01bd4111e5f7322fe4f9d4817ffd6f0c6d/Source/Modules/csharp.cxx#L4252-L4257

When using a `nullable reference type` this type of line is generated:
```csharp
private static global::System.Type[] swigMethodTypes1 = new global::System.Type[] { typeof(Rect?) };
```
which fails to compile with this error
```
error CS8639: The typeof operator cannot be used on a nullable reference type
```

I tested by manually changing the generated line to `typeof(Rect)` and it is working just fine and `SwigDerivedClassHasMethod` is happy.

I'm willing to help fix that but I need guidance on how to do it properly. I think of 2 possible solutions:
- In the `csharp.cxx` file if we have all the information to detect the type is a nullable reference and remove the `?` (but only for this, we still want to have `?` for basic types and structs). Basically another `else if`.
- By adding a new special `typemap` that will need to be used by my `std::optional` typemaps.

Thanks

synvima:
Omit the `?` for generated `swigMethodTypes` when using a `nullable reference type`.

For `nullable value types` the generated code stays the same. E.g:
- `sbyte?`, `byte?`,
- `short?`, `ushort?`,
- `int?`, `uint?`,
- `long?`, `ulong?`,
- `char?`,
- `float?`, `double?`,
- `bool?`,
- `decimal?`

Fixes #3108.

### DIFF
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -3786,6 +3786,35 @@ public:
   }
 
   /* -----------------------------------------------------------------------------
+   * containsCSharpNullableValueType()
+   * Determine whether the cstype in the typemap is a C# nullable value type.
+   * ----------------------------------------------------------------------------- */
+
+  bool containsCSharpNullableValueType(String* typemap)
+  {
+    const char* csharp_basic_types[] = {
+        "sbyte?", "byte?",
+        "short?", "ushort?",
+        "int?", "uint?",
+        "long?", "ulong?",
+        "char?",
+        "float?", "double?",
+        "bool?",
+        "decimal?"
+    };
+
+    for (size_t i = 0; i < sizeof(csharp_basic_types) / sizeof(csharp_basic_types[0]); ++i)
+    {
+      if (Strcmp(typemap, csharp_basic_types[i]) == 0)
+      {
+        return true;
+      }
+    }
+
+    return false;
+  }
+
+  /* -----------------------------------------------------------------------------
    * excodeSubstitute()
    * If a method can throw a C# exception, additional exception code is added to
    * check for the pending exception so that it can then throw the exception. The
@@ -4253,7 +4282,10 @@ public:
 		  Printf(proxy_method_types, "typeof(%s).MakeByRefType()", tm);
 		} else if (Replace(tm, "out ", "", flags) || Replace(tm, "out\t", "", flags)) {
 		  Printf(proxy_method_types, "typeof(%s).MakeByRefType()", tm);
+		} else if (containsCSharpNullableValueType(tm)) {
+		  Printf(proxy_method_types, "typeof(%s)", tm);
 		} else {
+		  Replace(tm, "?", "", DOH_REPLACE_ID_END | DOH_REPLACE_NOCOMMENT);
 		  Printf(proxy_method_types, "typeof(%s)", tm);
 		}
 	      } else {


wsfulton:
Can you please provide a motivating example or even better add a test to the patch. Without an example and without any new documentation in the Manual/Doc directory, it's hard to know what feature you are trying to add as currently SWIG does not have a need to parse any nullable reference types. While this patch seems to be needed for `std::optional`, what is a simple stand-alone use case?

christophe-calmejane:
Hi William, motivation for this comes from https://github.com/swig/swig/pull/3078

wsfulton:
A testcase for #3078 is of course needed. Maybe it will help, but for the generic feature put in here I need a standalone simple testcase.

Source/Modules/csharp.cxx
```diff
@@ -3785,6 +3785,35 @@ class CSHARP:public Language {
...
+        "sbyte?", "byte?",
+        "short?", "ushort?",
+        "int?", "uint?",
+        "long?", "ulong?",
+        "char?",
+        "float?", "double?",
+        "bool?",
+        "decimal?"
+    };
+
```
ojwb, line 3805: Unless most calls to this are likely to succeed (my guess is actually most will fail) it'd be more efficient to check if the final character is `?` first, something like:

```
if (Len(typemap) == 0 || Char(typemap)[Len(typemap) - 1] != '?') return false;
```

That way we avoid the loop of checks when there's no hope of any of them matching.

synvima: You are right, i also would expect that for most cases this will be false. I updated the source.

synvima:
I'm preparing a minimal example. This should also apply to null pointer types.

synvima:
I've created a minimal example using pointers that maps to nullable types in C#.
However, I'm currently unable to run the CI tests successfully.

So the current test definitions as they are.
_`director_nullable_args.i:`_
```c++
%module(directors="1") director_nullable_args
#if defined SWIGCSHARP

#include <typemaps.i>
#include <std_string.i>

%typemap(cstype)       std::string const * "string?"
%typemap(csin)         std::string const * "new System.Runtime.InteropServices.HandleRef()"
%typemap(csdirectorin) std::string const * "($iminput != global::System.IntPtr.Zero) ? System.Runtime.InteropServices.Marshal.PtrToStringAnsi($iminput) : null"
%typemap(directorin)   std::string const * "$input = ($1 != nullptr) ? (void*)$1->c_str() : nullptr;"

%typemap(cstype)       int const * "int?"
%typemap(csin)         int const * "new System.Runtime.InteropServices.HandleRef()"
%typemap(csdirectorin) int const * "($iminput != global::System.IntPtr.Zero) ? System.Runtime.InteropServices.Marshal.ReadInt32($iminput) : null"

%feature("director") TestObjectDirected;

%typemap(csimports) TestObjectDirected %{
#nullable enable
%}

%inline %{
class TestObjectDirected
{
public:
    TestObjectDirected()          = default;
    virtual ~TestObjectDirected() = default;

    virtual void onIndex(int const* index)
    {
    }

    virtual void onMessage(std::string const* message)
    {
    }

    void nextMessage()
    {
        if (counter++ % 2 == 0)
        {
            onIndex(nullptr);
            onMessage(nullptr);
        }
        else
        {
            std::string msg("Hello " + std::to_string(counter));
            onIndex(&counter);
            onMessage(&msg);
        }
    }

private:
    int counter { 0 };
};
%}

#endif
```

_`director_nullable_args.cs:`_
```csharp
#nullable enable

using System;
using director_nullable_argsNamespace;

public class runme
{
  public class DeriveredTestObject : TestObjectDirected
  {
      public override void onIndex(int? index)
      {
          base.onIndex(index);
          Indexes.Add(index);
      }

      public override void onMessage(string? message)
      {
          base.onMessage(message);
          Messages.Add(message);
      }

      public readonly List<string?> Messages = [];
      public readonly List<int?> Indexes = [];
  }
  
  private static void check(string got, string expected)
  {
    if (got != expected)
      throw new ApplicationException("Failed, got: " + got + " expected: " + expected);
  }
  
  static void Main()
  {
    var testDerivedDirected = new DeriveredTestObject();
    
    testDerivedDirected.nextMessage(); // counter = 0
    testDerivedDirected.nextMessage(); // counter = 1
    testDerivedDirected.nextMessage(); // counter = 2
    testDerivedDirected.nextMessage(); // counter = 3

    // Check Indexes
    check(testDerivedDirected.Indexes.Count.ToString(), "4");
    check(testDerivedDirected.Indexes[0]?.ToString() ?? "null", "null");
    check(testDerivedDirected.Indexes[1]?.ToString() ?? "null", "2");
    check(testDerivedDirected.Indexes[2]?.ToString() ?? "null", "null");
    check(testDerivedDirected.Indexes[3]?.ToString() ?? "null", "4");

    // Check Messages
    check(testDerivedDirected.Messages.Count.ToString(), "4");
    check(testDerivedDirected.Messages[0] ?? "null", "null");
    check(testDerivedDirected.Messages[1] ?? "null", "Hello 2");
    check(testDerivedDirected.Messages[2] ?? "null", "null");
    check(testDerivedDirected.Messages[3] ?? "null", "Hello 4");
  }
}

```

------------
In the windows pipeline, i'm getting this errors:
```
error CS0518: Predefined type 'System.Runtime.CompilerServices.NullableAttribute' is not defined or imported
error CS0518: Predefined type 'System.Runtime.CompilerServices.NullableContextAttribute' is not defined or imported
```

And in the linux pipeline, i'm getting this error:
```
error CS1024: Wrong preprocessor directive
```

It appears that the Windows pipeline targets an older .NET Framework 4 runtime, while the Linux pipeline runs entirely within the Mono runtime. Neither (?) of them supports the newer nullable reference types feature.

wsfulton:
From the docs at https://swig.org/Doc4.3/CSharp.html#CSharp_introduction:

> SWIG 3 and later requires .NET 2.0 at a minimum. There are some minor exceptions, where the minimum required is .NET 4.0. This is when using the std::complex and std::list STL containers. 

Hence we use .NET 4 to test. Ideally we test the majority of the code on .NET 2 except where .NET 4 is the stated minimum, but we havn't figured out any creative solutions for that. I've actually got some work in progress to get the dotnet tool to work for .net core 6/8 as it is radically different to the .NET framework, but I need to get back to it so we can at least additionally test later .NET versions. This work is related to #3078.

I think we'll still need creating solutions to only test .NET 8 features on .NET 8. So maybe take that as a challenge to get this merge request to pass on older .NET versions. We'll probably need to check the version of csc.exe / dotnet and only enable .NET 8 tests if .NET is later enough. So something similar to what we do for testing newer c++ standards. If fancy tackling all that, please do, but you may want to just provide a testcase that works only on .NET 8 and later and update your pull request until we have that machinery in place.

AdamLithgow:
> I think we'll still need creating solutions to only test .NET 8 features on .NET 8. So maybe take that as a challenge to get this merge request to pass on older .NET versions. We'll probably need to check the version of csc.exe / dotnet and only enable .NET 8 tests if .NET is later enough. So something similar to what we do for testing newer c++ standards. If fancy tackling all that, please do, but you may want to just provide a testcase that works only on .NET 8 and later and update your pull request until we have that machinery in place.

@synvima I think this should be able to be accomplished by generating some C# preprocessor directives that switches based on the framework version. Specifically, excluding the generation of the `#nullable enable` preprocessing directive, and changing the generation of anything that uses `string?` to just use `string` if the lang version is not at least .NET Core 3.0, as that seems to be the only place you're using the NRTs.

I think this, or something pretty close would be what you'd want to generate:
```
#if NETCOREAPP3_0_OR_GREATER
	#nullable enable
#endif

using System;
using director_nullable_argsNamespace;

public class runme
{
  public class DeriveredTestObject : TestObjectDirected
  {
      public override void onIndex(int? index)
      {
          base.onIndex(index);
          Indexes.Add(index);
      }

#if NETCOREAPP3_0_OR_GREATER
      public override void onMessage(string? message)
      {
          base.onMessage(message);
          Messages.Add(message);
      }

      public readonly List<string?> Messages = [];
#else
	public override void onMessage(string message)
      {
          base.onMessage(message);
          Messages.Add(message);
      }

      public readonly List<string> Messages = [];
#endif
      public readonly List<int?> Indexes = [];
  }
  
  private static void check(string got, string expected)
  {
    if (got != expected)
      throw new ApplicationException("Failed, got: " + got + " expected: " + expected);
  }
  
  static void Main()
  {
    var testDerivedDirected = new DeriveredTestObject();
    
    testDerivedDirected.nextMessage(); // counter = 0
    testDerivedDirected.nextMessage(); // counter = 1
    testDerivedDirected.nextMessage(); // counter = 2
    testDerivedDirected.nextMessage(); // counter = 3

    // Check Indexes
    check(testDerivedDirected.Indexes.Count.ToString(), "4");
    check(testDerivedDirected.Indexes[0]?.ToString() ?? "null", "null");
    check(testDerivedDirected.Indexes[1]?.ToString() ?? "null", "2");
    check(testDerivedDirected.Indexes[2]?.ToString() ?? "null", "null");
    check(testDerivedDirected.Indexes[3]?.ToString() ?? "null", "4");

    // Check Messages
    check(testDerivedDirected.Messages.Count.ToString(), "4");
    check(testDerivedDirected.Messages[0] ?? "null", "null");
    check(testDerivedDirected.Messages[1] ?? "null", "Hello 2");
    check(testDerivedDirected.Messages[2] ?? "null", "null");
    check(testDerivedDirected.Messages[3] ?? "null", "Hello 4");
  }
}
```

AdamLithgow:
Correction to my above comment, `#nullable` is not a valid preprocessor directive prior to C# 8.0. The above should normally work, but I believe since SWIG is running on mono, the conditional directives needed to allow it to work don't exist (it's building with an older compiler, instead of a newer compiler compiling an old version of .NET). As far as I can tell, the minimal example code @synvima provided works in any .net version that supports C# 8.0, but since the test pipelines run .NET 4, there's no way to get them to run in that context, at least not without much larger-scale changes.

christophe-calmejane:
As explained in #3078 I provided in my PR a more robust alternative to this PR.

synvima:
As described by Christophe, this fix is included in the referenced PR, and as far as I know, this implementation is mainly needed there. I'm closing it here.

Issue #3132, opened 2025-02-28
EOL looming for ubuntu-20.04 in GHA


ojwb:
Actual EOL is 2025-04-01 (so a month away as I write) but they're going to be doing "brownouts" on and off during March: https://github.com/actions/runner-images/issues/11101

Our current default for Linux jobs is `ubuntu-22.04` so this will be very disruptive if CI tries to run during a brownout so switching before the end of the month seems more sensible than eking out 4 more weeks on ubuntu-20.04 in exchange for some pain.  I'll push a branch shortly to see how things look with the obvious references updated.

Issue #2830, opened 2024-03-07
Support for Racket 8


erezgeva:
Debian racket package in bookworm and up, do not provide any mzdyn object.
Not sure if that is a bug in Debian or racket provides a better solution?

[bookworm racket version 8.7+dfsg1-1](https://packages.debian.org/bookworm/racket), [Files]( https://packages.debian.org/bookworm/amd64/racket/filelist)
[trixie racket version 8.10+dfsg1-2](https://packages.debian.org/trixie/racket), [Files]( https://packages.debian.org/trixie/amd64/racket/filelist)

Old versions have:
/usr/lib/racket/mzdyn3m.o
[bookworm racket version 7.9+dfsg1-2](https://packages.debian.org/bullseye/racket), [Files]( https://packages.debian.org/bullseye/amd64/racket/filelist)

Ubuntu follows. Newer version remove the mzdyn object.
[8.10+dfsg1-2  universe](https://packages.ubuntu.com/noble/racket), [Files](https://packages.ubuntu.com/noble/amd64/racket/filelist)

ojwb:
Racket 8 changed the underlying scheme implementation so SWIG doesn't currently work with this version or newer - see discussion in #920.  We've not had a SWIG/Racket maintainer for a long time but maybe somebody cares and will step forward and fix this.  The plan is to drop Racket support if that doesn't happen and we reach the point where we can no longer test Racket in CI.

ojwb:
This is already documented in the manual BTW:

> SWIG works with Racket versions &lt; 8 (Racket 8 switched to be based on a different scheme interpreter and SWIG hasn't been updated for this change).

```
$ rmadison -uubuntu racket
 racket | 5.3.6+dfsg1-1      | trusty/universe | source, amd64, armhf, i386, powerpc
 racket | 6.3-1              | xenial/universe | source, amd64, arm64, armhf, i386, powerpc, s390x
 racket | 6.11+dfsg1-1       | bionic/universe | source, amd64, arm64, armhf, i386, ppc64el
 racket | 7.2+dfsg1-2ubuntu3 | focal/universe  | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 racket | 8.2+dfsg1-2ubuntu1 | jammy/universe  | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
 racket | 8.7+dfsg1-1        | mantic/universe | source, amd64, arm64, armhf, ppc64el, riscv64, s390x
[...]

PR #3133, opened 2025-02-28
[ci] Update for ubuntu-20.04 EOL

ojwb:
See #3132

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -35,13 +35,9 @@ jobs:
         - SWIGLANG: ""
           CPPFLAGS: "-DDOH_POISON"
         - SWIGLANG: ""
-          GCC: 7
+          GCC: 9
           CPPSTD: c++98
           CSTD: c99
-        - SWIGLANG: ""
-          GCC: 7
-        - SWIGLANG: ""
-          GCC: 8
         - SWIGLANG: ""
           GCC: 9
         - SWIGLANG: ""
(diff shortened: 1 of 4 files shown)


ojwb:
Oddly despite this showing as failing CI here, in the actions tab I see green runs for the latest build (at time of writing).

I don't think this is quite ready to merge yet anyway though as it just turns off some jobs which ought to be runnable on ubuntu-22.04.

To summarise:

* We lose Linux builds with GCC 7 and 8
* We lose Guile 2.0 testing
* We lose Python 3.5 and 3.6 testing
* We need to adjust things to be able to still test python2 (because the package to install is now `python2-dev` instead of `python-dev` and `/usr/bin/python` is now always Python3)
* We lose mzscheme/racket testing (we have already flagged that support would be dropped at this point unless somebody updated SWIG to work with Racket 8: see #2830)
* I've had to comment out Ruby 2.4, 2.5, 2.6, 2.7 and 3.0 for now as installing them with `rvm` fails with an unhelpful error message

The Ruby one needs addressing.  We could probably use Ubuntu's ruby3.0 package for 3.0, but 2.x isn't available as a package for ubuntu-22.04.  It seems Ruby 3.0 is only just over 4 years old so losing testing of all 2.x wouldn't be ideal (losing the oldest/older 2.x would be reasonable at this point).  Do we have an active SWIG/Ruby maintainer currently?

wsfulton:
Thanks @ojwb. I might tweak and tidy it up a bit later, but for now this is great.

PR #3136, opened 2025-03-05
Python: Hide SwigPyIterator class

jschueller:
since there is no way to ignore that class, lets add a "_" prefix so that it doesnt show up in the interpreter autocompletion


ojwb:
A comment in `Lib/python/pyiterators.swg` reads (before patching):

_Users can derive form the SwigPyIterator to implement their own iterators._

So the proposed renaming would break compatibility with any existing user interface file which does that.

Also using identifiers which _begin with an underscore followed by an uppercase letter_ is not a good idea in C++ since _If the programmer uses such identifiers, the program is ill-formed, no diagnostic required_:

https://en.cppreference.com/w/cpp/language/identifiers

This problem could probably be avoided by keeping the C++ name the same and using `%rename` to add an underscore to the Python wrapper name.

jschueller:
indeed ignore is enough, thanks

Issue #103, opened 2013-11-04
Thread Safety Bugfix


tjachmann:
There is a severe bug regarding thread savity in the functions 
SWIG_TypeCheckStruct and SWIG_TypeCheck, that did cost me quite some time to figure out. I propose to generally apply the following bugfix, so others do not run into this too. 

The commented code below in these functions is not thread safe, as the swig_type_info structures are static. So using for example multiple Lua States, running on different threads results in hard to find and reproduce, mind blowing, random errors and all kinds of strange misbehaviour...

The function is at least used commonly in the generated wrapper functions for Lua.
And I guess it is also used for all the other languages as well, but I have not checked this. So I guess without this bugfix, probalbly all language wrappers are NOT thread safe without the following fix:

```
/*
  Check the typename
*/
SWIGRUNTIME swig_cast_info *
SWIG_TypeCheck(const char *c, swig_type_info *ty) {
  if (ty) {
    swig_cast_info *iter = ty->cast;
    while (iter) {
      if (strcmp(iter->type->name, c) == 0) {
        if (iter == ty->cast)
          return iter;

        /*
        THREAD SAVITY BUGFIX. See also SWIG_TypeCheckStruct

        Move iter to the top of the linked list
        iter->prev->next = iter->next;
        if (iter->next)
          iter->next->prev = iter->prev;
        iter->next = ty->cast;
        iter->prev = 0;
        if (ty->cast) ty->cast->prev = iter;
        ty->cast = iter;
        */
        return iter;
      }
      iter = iter->next;
    }
  }
  return 0;
}

/*
  Identical to SWIG_TypeCheck, except strcmp is replaced with a pointer comparison
*/
SWIGRUNTIME swig_cast_info *
SWIG_TypeCheckStruct(swig_type_info *from, swig_type_info *ty) {

  if (ty) {
    swig_cast_info *iter = ty->cast;
    while (iter) {
      if (iter->type == from) {
        if (iter == ty->cast)
          return iter;

        /*                                                         
        THREAD SAVITY BUGFIX. 

        Move iter to the top of the linked list
        iter->prev->next = iter->next;
        if (iter->next)
          iter->next->prev = iter->prev;
        iter->next = ty->cast;
        iter->prev = 0;
        if (ty->cast) ty->cast->prev = iter;
        ty->cast = iter;
        */

        return iter;
      }
      iter = iter->next;
    }
  }
  return 0;
}
```

Alternate solutions:
[...]

Issue #3121, opened 2025-02-06
Support free-threaded Python


ngoldbaum:
It should be possible to use SWIG to build a Python C extension for use with the free-threaded build of CPython.

I'm working at Quansight Labs along with a small team to improve ecosystem support for free-threaded Python. We've put together some documentation that is a good place to start: https://py-free-threading.github.io.

One big wrinkle that might come up is that free-threaded 3.13 (and maybe 3.14) won't support the limited API, so free-threaded extensions must use the version-specific ABI.

When we added free-threading support in other wrapper tools like Cython, PyO3, and F2Py the job of adding free-threaded support was first to make sure generated wrapping code is thread-safe. Once that's done, there needs to be a way for the extension author to signal that they think the extension is thread-safe.

wsfulton:
Your link led me to [Quansight Labs blog post about work on free-threading](https://labs.quansight.org/blog/free-threaded-python-rollout) and it looks like a great initiative. Indeed it should be possible for the code that SWIG generates to support free-threaded Python.

Given SWIG generates bindings for hundreds of Python extensions, it makes sense for SWIG to be involved. Good to have a heads up of this initiative and we can do what we can to help. Just so that you know, the SWIG developers are struggling to keep up with patches and necessary improvements and if Quansight Labs can assist, as you do other Python open source projects, it would certainly help the wider Python ecosystem and free-threaded Python stability. The most beneficial thing would be to find an advanced C Python API expert who can join the SWIG developer community to manage patches/pull requests and make other improvements into the SWIG code base. I say this as the SWIG developers are volunteers, are very few and without a sponsored SWIG developer, we're never going to keep up. Feel free to contact me offline or discuss further here if you feel Quansight can help.

wsfulton:
Please try out the free threaded Python support from #3137 which has now been merged to master for inclusion in swig-4.4.0.

wsfulton:
@ngoldbaum, perhaps you could update the compatibility site for SWIG at https://py-free-threading.github.io/tracking/.

ngoldbaum:
> [@ngoldbaum](https://github.com/ngoldbaum), perhaps you could update the compatibility site for SWIG at https://py-free-threading.github.io/tracking/.

see https://github.com/Quansight-Labs/free-threaded-compatibility/pull/173

crusaderky:
Follow-up to this:
- https://github.com/swig/swig/pull/3215

PR #3137, opened 2025-03-09
added support for Python free threading (aka no-gil)

klausspanderen:
Python 3.13 now supports free threading (aka disabling the GIL). Some marcos are deemed unsafe for this use case and should be disabled. In addition a variant of [Thread Safety Bugfix #103](https://github.com/swig/swig/issues/103) needs to be put in place.

### DIFF
--- a/Examples/test-suite/python/langobj_runme.py
+++ b/Examples/test-suite/python/langobj_runme.py
@@ -3,6 +3,10 @@ from langobj import *
 
 
 x = 256*256+1
+
+# avoid immortal object in no-gil scenario
+x += 1
+
 rx = sys.getrefcount(x)
 v = identity(x)
 rv = sys.getrefcount(v)
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -103,7 +103,7 @@ SWIG_Python_str_FromChar(const char *c)
 #endif
 
 /* Increment and Decrement wrappers - for portability when using the stable abi and for performance otherwise */
-#if defined(Py_LIMITED_API) || defined(Py_GIL_DISABLED)
+#ifdef Py_LIMITED_API
 # define SWIG_Py_INCREF Py_IncRef
 # define SWIG_Py_XINCREF Py_IncRef
 # define SWIG_Py_DECREF Py_DecRef
--- a/Lib/python/pyhead.swg
+++ b/Lib/python/pyhead.swg
@@ -75,7 +75,20 @@ SWIG_Python_str_FromChar(const char *c)
 #define Py_hash_t long
 #endif
 
-#ifdef Py_LIMITED_API
+#ifdef Py_GIL_DISABLED
+# undef PyTuple_GET_ITEM
+# undef PyTuple_SET_ITEM
+# undef PyTuple_GET_SIZE
+# undef PyCFunction_GET_FLAGS
+# undef PyCFunction_GET_FUNCTION
+# undef PyCFunction_GET_SELF
+# undef PyList_GET_ITEM
+# undef PyList_SET_ITEM
+# undef PySliceObject
+#endif
+
+
+#if defined(Py_LIMITED_API) || defined(Py_GIL_DISABLED)
 # define PyTuple_GET_ITEM PyTuple_GetItem
 /* Note that PyTuple_SetItem() has different semantics from PyTuple_SET_ITEM as it decref's the original tuple item, so in general they cannot be used
   interchangeably. However in SWIG-generated code PyTuple_SET_ITEM is only used with newly initialized tuples without any items and for them this does work. */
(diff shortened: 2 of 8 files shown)


Lib/python/pyhead.swg
```diff
@@ -75,7 +75,20 @@ SWIG_Python_str_FromChar(const char *c)
...
+# undef PyTuple_SET_ITEM
+# undef PyTuple_GET_SIZE
+# undef PyCFunction_GET_FLAGS
+# undef PyCFunction_GET_FUNCTION
+# undef PyCFunction_GET_SELF
+# undef PyList_GET_ITEM
+# undef PyList_SET_ITEM
+# undef PySliceObject
+#endif
+
```
wsfulton, line 89: Do you have a reference or reason for undefiningthese?

Lib/swigrun.swg
```diff
@@ -278,6 +278,7 @@ SWIG_TypeCheck(const char *c, swig_type_info *ty) {
       if (strcmp(iter->type->name, c) == 0) {
         if (iter == ty->cast)
           return iter;
+#ifndef Py_GIL_DISABLED
```
wsfulton, line 281: I presume this is a performance optimisation you've disabled so that `SWIG_TypeCheck` can be called from multiple threads without having to add a lock ???

wsfulton: Seems like it is having digested https://github.com/swig/swig/issues/103 more thoroughly.

Lib/python/pyhead.swg
klausspanderen: https://py-free-threading.github.io/porting/ states that some low level macros like PyList_SET_ITEM aren't safe with free threading. Hence I've undef the macros and redefined them using the limited api version in the next define section.

klausspanderen:
I had to make a few minor changes to get CI working with python3.13t.

configure.ac
```diff
@@ -2075,6 +2075,13 @@ else
     if test -z "$PYVER"; then
       PYVER=0
     fi
+
+    AC_MSG_CHECKING([for $PYTHON3 GIL is disabled])
+    PYGIL=`($PYTHON3 -c "import sysconfig; print(not bool(sysconfig.get_config_var('Py_GIL_DISABLED')))") 2>/dev/null`
+    AC_MSG_RESULT($PYGIL)
+    if test x"$PYGIL" = x"False"; then
+      PYTHON3="$PYTHON3 -Xgil=0"
+    fi
```
wsfulton, line 2084: This results in output:
```
checking for python3.13t GIL is disabled... False
```
Isn't the logic wrong here? It should surely output True. Then I don't think you need `-Xgil=0` to turn off the GIL if PYGIL is True. But if it is really needed the docs imply it should be with a space `-X gil=0` ... https://docs.python.org/3/using/cmdline.html#cmdoption-X.

We'll need to think of some way to show clearly that the GIL is turned off in the build or configure output, but probably the additional `AC_MSG_RESULT` is good enough once corrected.

Examples/test-suite/python/langobj_runme.py
```diff
@@ -3,6 +3,10 @@
 
 
 x = 256*256+1
+
+# avoid immortal object in no-gil scenario
+x += 1
+
```
wsfulton, line 9: Could you elaborate on this please? I don't understand why immortal objects are not allowed in no-gil scenarios.

configure.ac
klausspanderen: arrg...you are right, the comment needs to be other way around, fixed it. W.r.t. the command line option -Xgil=0, I think we need it because it ensures that the GIL stays disabled even if test cases import libraries, which are not explicitly build for the free threading case (the space between -X and gil is not needed), otherwise an import might enable the GIL back again.

wsfulton:
Thanks @klausspanderen for this. I've squashed your changes and added some refinements and docs, see d96024bb0e3ee35f13349df0d004c5b4a5a9cfed...298c489dc623046466485c60d61849db5b40ae6.

configure.ac
wsfulton: I made a few tweaks in 1b09f4f11112f0276f7b64c54e2a1ea73062fe52 to make it give more standard configure output.

wsfulton:
Related issue #3121.

ngoldbaum:
This probably deserves at least a mention in the SWIG documentation. In particular, that SWIG isn't doing anything to guard against any thread-unsafe wrapped native code (at least from what I can tell, reading the PR diff) and it's up to users to prevent unsafe shared mutation using either Python-level locking in Python wrapper code or using native locking (e.g. making the wrapped native code thread-safe).

wsfulton:
Hopefully 4167b0275adc574f86914e42fb5ef9639b2b25a5 makes this clear enough now.

PR #3143, opened 2025-03-25
Remove redundant casts to const char* after using GetStringUTFChars method

DLuchyk:
This PR introduces an improvement aimed at reducing SonarQube issues in SWIG. Our team has been maintaining local modifications to address static analysis findings, and we would like to contribute these changes upstream.

The signature of the function:
`const char * GetStringUTFChars(...); `
There is no reason of casting the return value to `const char *`


ojwb:
Thanks.  Looks like these probably came from copying typemaps which used `GetStringChars()` - that returns `const jchar *` so a cast may be needed.

PR #3145, opened 2025-03-25
Refactor SWIG Code to Address SonarQube Rule cpp:S5827

DLuchyk:
This PR refactors SWIG's C++ code to comply with SonarQube rule cpp:S5827, which recommends using auto to eliminate redundant type declarations. By applying this change, we improve code readability, maintainability, and reduce unnecessary verbosity while preserving functionality.


ojwb:
Sorry, we can't take this patch.

The generated code needs to be valid C++98 (unless only emitted when wrapping features from newer C++ standards - e.g. if actually wrapping `auto` it's OK to emit code using `auto` or indeed other C++11 features).

You'll have to just turn off this code style check for SWIG-generated code.

PR #3146, opened 2025-03-26
Address s5416 SonarQube issue for carrays.i&javahead.swg files

DLuchyk:
This PR introduces an improvement aimed at reducing SonarQube issues in SWIG. Our team has been maintaining local modifications to address static analysis findings, and we would like to contribute these changes upstream.
This pull request addresses SonarQube rule cpp:S5416: "using" should be preferred to "typedef" for type aliasing.


ojwb:
This form of `using` was new in C++17 so we can't merge this.  Essentially the same reason as #3145.

It's really not useful to keep submitting such patches.

Issue #2942, opened 2024-06-30
Scilab failures in SWIG test-suite


wsfulton:
Scilab testing in Github Actions is proving to be somewhat flaky. I've gone back over the last couple of months and dug out some examples of the failures which I've summarised in the table below.

|Test failure URL|Github Actions job name|test name|Scilab version|
|---|---|---|---|
|https://github.com/swig/swig/actions/runs/9695230329/job/26754542543|scilab 2023.0.0 gcc|inout test|2023.0.0.1678464761|
|https://github.com/swig/swig/actions/runs/9582066278/job/26420260029|scilab 2023.0.0 gcc|overload_null test|2023.0.0.1678464761|
|https://github.com/swig/swig/actions/runs/9392556110/job/25867005911|scilab gcc|constants example|6.1.1.988271013|
|https://github.com/swig/swig/actions/runs/9652896968/job/26624050087|scilab gcc c++14|contract example|6.1.0.0|
|https://github.com/swig/swig/actions/runs/9669957277/job/26677659483|scilab 2023.0.0 gcc|contract example|2023.0.0.1678464761|
|https://github.com/swig/swig/actions/runs/9584238501/job/26427413888|scilab 2023.0.0 gcc|contract example|2023.0.0.1678464761|

The stack traces point to a file system error and looks to me to be a problem internal to Scilab.

@vincentcouvert @sylvestre could you please investigate and come up with a solution to make these reliable? I doesn't look as simple as disabling just a few tests as the same sort of error happens in a number of different tests.

wsfulton:
The only versions tested on Github are those in the table. I suggest @sylvestre or @smarchetto makes a change to add a new test of the later version and the monitor. However, I think it requires some real work either using tools to analyse (valgrind?) or someone who understands the stack trace to look as it happens infrequently, so is not easily reproduced.

vincentcouvert:
Hi William,

I updated the tested Scilab versions in https://github.com/swig/swig/pull/2952
The issue is related to a random concurrent access to a file when multiple Scilab instances are run at the same time and this issue was fixed in Scilab 2024.1.0.

Vincent

ojwb:
Thanks, I've merged #2952.

It looks like we're now testing with scilab 5.5.2, 6.0.2, 6.1.0, 6.1.1, 2023.1.0 (instead of 2023.0.0) and 2024.1.0.

[...]

PR #3154, opened 2025-04-02
Ci scilab debug

ojwb:
This is to check the current status of #2942 - it is not intended to be merged as-is.

(There is a new job in here for scilab 2025.0.0 which I'll merge once I've determined that it works, and whether it can run in parallel or not.)


ojwb:
I've retried the builds a few times and finally CI failed, but because the downloads from scilab.org failed with HTTP error 503 - not due to the apparent parallelism issue(s).

I'm not sure it's useful to keep retrying - the issues didn't trigger very often before, and while it's possible they've been fixed by changes elsewhere, it seems most likely they are due to bugs in Scilab itself.  If a particular old version which has the bug it will continue to do so, and the interesting question is really if a new release does.

The only newer release is 2025.0.0 and that currently doesn't work with SWIG, but we have an open issue for that (#3155).

The workaround of turning off parallel make for SWIG's scilab testsuite isn't terrible either - it slows down those CI jobs but they're still testing all the same features.

If we think it's really useful to recheck, probably turning off the workaround on master and watching for CI failures in jobs we're running anyway would be best, but I don't currently plan to do that.

Therefore closing.

PR #3158, opened 2025-04-10
fix the placement of open namespace for c# interface class.

pxyn:


wsfulton:
Could you please give some details and an example that requires this change? Our regression test-suite doesn't currently show up any problems, so am interested in exactly what it is wrong.

ojwb:
@pxyn Please can you provide an example showing the problem your proposed change is intended to solve?

I checked all the C# testcases which use `swiginterface.i` and the C# code SWIG generates for each seems to be **identical** with or without your patch.

wsfulton:
I can't work out why this is needed, it also seems wrong to do move just this one addOpenNamespace and not the other one just above it to after the call to `Language::classHandler`. Closing, but please reopen if a motivating example is provided.

Issue #2881, opened 2024-04-18
DeprecationWarning when building with Py_LIMITED_API


bartoli:
When running a module generated by swig 4.2, built with Py_LIMITED_API,

The console output gets 3 warnings
`<frozen importlib._bootstrap>:241: DeprecationWarning: builtin type swigvarlink has no __module__ attribute`

For swigPyPacked, SwigPyObject, swigvarlink types.

It seems we could remove them by modifying, in pyrun.swg where the 3 PyType_Spec are filled for thise types:
```
  PyType_Spec spec = {
  #ifndef Py_LIMITED_API
    "SwigPyPacked",
  #else
    SWIG_name ".SwigPyPacked"
  #endif
    sizeof(SwigPyPacked),
```

But in the generated cpp file, SWIG_name is actually defined after those code fragments

I will patch output files on my side to avoid the warning. Do you have an idea on how to make a clean change that could be pushed? Can SWIG_name be defined earlier in the output?

ojwb:
> Can SWIG_name be defined earlier in the output?

```
diff --git a/Source/Modules/python.cxx b/Source/Modules/python.cxx
index 9d195d14..bb832547 100644
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -782,7 +782,8 @@ public:
     Printf(f_header, "#else\n");
     Printf(f_header, "#  define SWIG_init    init%s\n\n", module);
     Printf(f_header, "#endif\n");
-    Printf(f_header, "#define SWIG_name    \"%s\"\n", module);
+
+    Printf(f_runtime, "#define SWIG_name    \"%s\"\n\n", module);
 
     Printf(f_wrappers, "#ifdef __cplusplus\n");
     Printf(f_wrappers, "extern \"C\" {\n");
```

(BTW, you're missing a comma in your code snippet above.)

wsfulton:
I've done a lot of testing with the limited API and python versions but I haven't seen this. In order to reproduce, could you please provide the value of Py_LIMITED_API that you are using, the version of python you are using to compile against and the version of python you are running with.

bartoli:
This is with python 3.11.7, swig 4.2.0,  Py_LIMITED_API=0x3040000.
It MIGHT also be only when this is a debug build of python and the swig module

The stackoverflow thread where i saw the first solution suggestion: https://stackoverflow.com/questions/66614672/python-limited-api-howto-set-module-attribute-of-extension-type

[...]

PR #3160, opened 2025-04-11
Python: Avoid DeprecationWarning

jschueller:
For internal classes SwigPyPacked, SwigPyObject, swigvarlink

See #2881


wsfulton:
Thanks, I've used this as a starting point in commit 49a7d2b0c6aed00bd07a9a28569020d560c2658c as the basis for the solution described in #2881.

jschueller:
Thanks!

PR #3163, opened 2025-04-16
Avoid warnings about using Py_TYPE from abi3audit

### LINKED ISSUE #3069: Python abi version mismatch ?

jschueller:
These warnings are spurious because Py_TYPE is an inline function and so using it doesn't prevent the extension from using stable Python ABI from earlier versions, but it is still better to avoid it, especially because this warning may become an actual error soon.

Closes #3069.


vadz:
Also would close #3073, if merged.

jschueller:
oh indeed I didnt know you already opened a PR for that, sorry

PR #3165, opened 2025-04-18
Use SWIG_NULLPTR instead of NULL for director methods to address sona…

DLuchyk:
…rqube issue cpp:s4962

### DIFF
--- a/Source/Modules/java.cxx
+++ b/Source/Modules/java.cxx
@@ -2722,21 +2722,6 @@ public:
       String *mangled_overname = Swig_name_construct(getNSpace(), overloaded_name);
       String *imcall = NewString("");
 
-#if 0
-      // TODO: add full support for Java annotations - implementation should be same as C# attributes
-      const String *annotations = Getattr(n, "feature:java:annotations");
-#else
-      const String *annotations = NULL;
-#endif
-      // Default annotation for director constructors is a warning suppression
-      static const String *suppress_warning_this_escape = NewString("@SuppressWarnings(\"this-escape\")");
-      if (!annotations && feature_director)
-	annotations = suppress_warning_this_escape;
-      if (annotations) {
-	Printf(function_code, "  %s\n", annotations);
-	Printf(helper_code, "  %s\n", annotations);
-      }
-
       const String *methodmods = Getattr(n, "feature:java:methodmodifiers");
       methodmods = methodmods ? methodmods : (is_public(n) ? public_string : protected_string);
 
@@ -3780,7 +3765,7 @@ public:
 
       Printf(f_runtime, "namespace Swig {\n");
       Printf(f_runtime, "  namespace {\n");
-      Printf(f_runtime, "    jclass jclass_%s = NULL;\n", imclass_name);
+      Printf(f_runtime, "    jclass jclass_%s = SWIG_NULLPTR;\n", imclass_name);
       Printf(f_runtime, "    jmethodID director_method_ids[%d];\n", n_methods);
       Printf(f_runtime, "  }\n");
       Printf(f_runtime, "}\n");
@@ -4182,7 +4167,7 @@ public:
 
       Wrapper_add_localv(w, "swigjnienv", "JNIEnvWrapper", "swigjnienv(this)", NIL, NIL);
       Wrapper_add_localv(w, jenvstr, "JNIEnv *", jenvstr, "= swigjnienv.getJNIEnv()", NIL);
-      Wrapper_add_localv(w, jobjstr, "jobject", jobjstr, "= (jobject) NULL", NIL);
+      Wrapper_add_localv(w, jobjstr, "jobject", jobjstr, "= (jobject) SWIG_NULLPTR", NIL);
       Delete(jenvstr);
       Delete(jobjstr);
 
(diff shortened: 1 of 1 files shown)


Source/Modules/java.cxx
```diff
@@ -2731,10 +2731,10 @@ class JAVA:public Language {
...
       static const String *suppress_warning_this_escape = NewString("@SuppressWarnings(\"this-escape\")");
       if (!annotations && feature_director)
-	annotations = suppress_warning_this_escape;
+	  annotations = suppress_warning_this_escape;
       if (annotations) {
-	Printf(function_code, "  %s\n", annotations);
-	Printf(helper_code, "  %s\n", annotations);
+	    Printf(function_code, "  %s\n", annotations);
+	    Printf(helper_code, "  %s\n", annotations);
       }
```
ojwb, line 2738: Please don't make changes to re-indent code entangled with functional changes.  Especially not code that's unrelated to the functional changes.

In this case the indentation looks correct as-is, so this change is also unwanted anyway (maybe you have your tab width wrongly set - it should move to the next multiple of 8 columns for SWIG code).

If the indentation actually is wrong please open a separate PR for fixing it.

ojwb (changes requested):

wsfulton:
We have pointer initialisation using `0` all over the place, it's what is generally used and I can't see that changing too easily. Pointer initialisation using `0`, `NULL` or `nullptr` is ultimately the same thing. Does sonarqube not also warn about `0` pointer initialisation? I don't know that changing it just in this one place is going to add much value, I'd rather see `0`, or NULL in the code to be honest, as I know what I'm getting then very easily.

ojwb:
Presumably the motivator for this check is that for C++, `NULL` must effectively be defined as `0` (it can't be `(void*)NULL` because that C++ won't implicitly convert `void*` to a pointer to another type, only the reverse).

That means e.g. `int i = NULL;` expands to `int i = 0;` and doesn't get warned about, and similarly for passing `NULL` for a non-pointer function argument.  AIUI this is the main reason why `nullptr` was created.

However GCC and clang both have magic definitions of `NULL` so you do get a warning:

```
$ printf '%s\n' '#include <stdlib.h>' 'int i = NULL;'|g++ -xc++ -c -
<stdin>:2:9: warning: converting to non-pointer type ‘int’ from NULL [-Wconversion-null]
$ printf '%s\n' '#include <stdlib.h>' 'int i = NULL;'|clang++ -xc++ -c -
<stdin>:2:9: warning: implicit conversion of NULL constant to 'int' [-Wnull-conversion]
    2 | int i = NULL;
      |     ~   ^~~~
      |         0
1 warning generated.
```

I do wonder if this situation is better addressed by users of this code analysis tools telling it to suppress this check for SWIG-generated code.  We've already had to reject changes aimed at placating this tool (e.g. #3145 where it wants us to use `auto` which we can't without raising the baseline C++ requirement to C++11) so they're going to need to suppress some checks anyway.

ojwb:
> We have pointer initialisation using `0` all over the place, it's what is generally used and I can't see that changing too easily. Pointer initialisation using `0`, `NULL` or `nullptr` is ultimately the same thing. Does sonarqube not also warn about `0` pointer initialisation? I don't know that changing it just in this one place is going to add much value, I'd rather see `0`, or NULL in the code to be honest, as I know what I'm getting then very easily.

It looks like `SWIG_NULLPTR` was introduced in the first place due to warnings from this same "sonarqube" tool: #2499

This does seem like it's going to be a nearly bottomless pit of additional work we really don't need which it's hard to argue makes the code better (if anything it makes it slightly worse as having `SWIG_NULLPTR` everywhere instead of the standard `NULL` or `0` is ugly and a minor distraction for those of use maintaining the SWIG code).

If we could switch to using `nullptr` everywhere I'd agree that'd be a bit better, but that would require C++11.  We'd also still need the ugly `SWIG_NULLPTR` for code that can be used from C or C++.

As I noted above, we've rejected changes to use `auto` to placate sonarqube warnings.  I wonder if we should just instate a wider policy that users of this tool need to suppress certain warnings from it for SWIG-generated code.  We could perhaps even include a suppression file if someone wants to contribute one - it's not something we really want to get into maintaining, but having a central place where it can be housed would be helpful to people using it with SWIG.

wsfulton:
@ojwb, agreed with your last comments. Let's remove the SWIG_NULLPTR macro altogether and go back to use 0 or NULL. Sonarqube is a great tool for static analysis, but like all static analysis tools, they often require tons of mindless warning suppressions and given SWIG's target standards and goals, the history of SWIG and the fact there is more than enough work to do for the contributors available, it is too much work to fix this warning. Recommendation is to suppress the Sonarqube warning in generated code.

ojwb:
> Let's remove the SWIG_NULLPTR macro altogether and go back to use 0 or NULL.

OK, done.

I'd be supportive of patches to migrate to using `NULL` more consistently instead of `0` - I think there was a school of thought that `0` was a better null pointer constant in C++ because `NULL` was just defined to `0`, but as I noted above GCC and clang now both have magic definitions of `NULL` and with those using `NULL` both signals to the reader that this is a null pointer literal not an integer literal, and also allows the compiler to warn if `NULL` is misused in a non-pointer context.

wsfulton:
Could be challenging to implement as a lot of code initialises to '0' as it works with pointers and integral types alike without having to inspect the type. I'll be focusing on real problems!

ojwb:
Uses in SWIG itself (rather than generated code) don't have that problem, though I guess users care more about warnings from compiling or statically analysing the generated code.

I'm not about to spend time changing this - I really just meant if someone submitted a clean patch I'd definitely try to review and merge it.

PR #3166, opened 2025-04-18
Use SWIG_NULLPTR instead of NULL for java director methods

DLuchyk:
We use SWIG as part of our system and rely on SonarQube for static code analysis. Replacing NULL with SWIG_NULLPTR eliminates the S4962 warnings, improving code quality and maintaining compliance with modern C++ best practices.


wsfulton:
#3165 is a duplicate

ojwb:
> Copilot AI left a comment
> [...]

Wow, that really is staggeringly useless.

> #3165 is a duplicate

Not an exact duplicate, though there's overlap in scope and some of the same changes are in both.

PR #3167, opened 2025-04-19
Makefile.am: avoid double `bison` execution on parallel build

trofi:
Before the change `make -j16` ran `bison` twice on the same input:

    $ make -j16
    ...
    make[1]: Entering directory '/build/source/Source'
    bison -d -Wall -Werror  --output=CParse/parser.c ./CParse/parser.y
    gcc -g -O2 -Wall -W -I.   -c -o mdfour.o mdfour.c
    bison -d -Wall -Werror  --output=CParse/parser.c ./CParse/parser.y
    ...

Note how two parallel `bison` calls both generate the same files. It works most of the time, but is prone to generate unexpected output.

This happens because Makefile rules like:

    a b: c; COMMAND
    all: a b

will run COMMAND for each target (twice here: once for `a` and once for `b`),

The fix uses GNU make's Grouped Targets to use single command to produce
multiple targets.

### DIFF
--- a/Source/Makefile.am
+++ b/Source/Makefile.am
@@ -99,7 +99,9 @@ eswig_SOURCES =	CParse/cscanner.c		\
 
 bin_PROGRAMS = eswig
 
-CParse/parser.c CParse/parser.h: CParse/parser.y
+# Use GNU make's grouped targets to avoid repeated bison execution for
+# each target.
+CParse/parser.c CParse/parser.h &: CParse/parser.y
 	$(BISON) $(AM_YFLAGS) $(YFLAGS) --output=CParse/parser.c $(srcdir)/CParse/parser.y
 
 # The executable is copied to the root directory for installation and running the test-suite.


ojwb:
Your solution misbehaves unhelpfully if `parser.h` is deleted:

```
make
rm Source/CParse/parser.h
make
```

The second `make` doesn't rerun bison because `parser.c` is up to date, then `touch` creates an **empty** `parser.h`, which results in stuff failing to compile with many errors and warnings, none of which usefully point to the problem.

`touch -c` would help, and is [specified by POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/touch.html) - I'm don't know if it's sufficiently portable in practice or not.  However, then just running `make` unhelpfully doesn't recover from removing `parser.h` - the user will need to also delete `parser.c` or `make clean` (at least the error points more clearly to the problem if not the solution):

```
CParse/cscanner.c:19:10: fatal error: parser.h: No such file or directory
   19 | #include "parser.h"
      |          ^~~~~~~~~~
```

The automake manual discusses this problem and has several solutions (all of them sadly rather clunky feeling): https://www.gnu.org/software/automake/manual/html_node/Multiple-Outputs.html

We do already require GNU make though and its "grouped targets" seem to nearly solve this problem:

```
CParse/parser.c CParse/parser.h &: CParse/parser.y
       $(BISON) $(AM_YFLAGS) $(YFLAGS) --output=CParse/parser.c $(srcdir)/CParse/parser.y
```

Does that work for you?

trofi:
> Your solution misbehaves unhelpfully if `parser.h` is deleted:

Oh, great point!

> We do already require GNU make though and its "grouped targets" seem to neatly solve this problem:
> 
> ```
> CParse/parser.c CParse/parser.h &: CParse/parser.y
>        $(BISON) $(AM_YFLAGS) $(YFLAGS) --output=CParse/parser.c $(srcdir)/CParse/parser.y
> ```
> 
> Does that work for you?

That's perfect! Updated the patch to use your solution as is.

ojwb:
I was curious how non-GNU make would handle such a rule.  Testing with the Debian `bmake` package which is apparently "a port of the NetBSD make tool" it appears the `&` is simply parsed as an additional target, which is pretty much harmless as nobody/nothing will realistically try to `make '&'` and even if they did it would at worst run such a rule.

I didn't test other make implementations, but it seems likely they'd behave the same way.  So probably you get a parallel-make safe rule for GNU make and a rule which works in non-parallel builds for other makes.

(This is only tangentially relevant to SWIG since we require GNU make already.)

PR #3181, opened 2025-05-21
Patch for go int type. The original integer type translation is incorrect.

rocket049:
The original integer type translation is incorrect.
C  int = Go  int32
C long = Go int32


ojwb:
C `long` **is** 64-bit on many 64-bit platforms

ojwb:
This breaks CI and seems to be based on incorrect assumptions about how C and Go types are related.  No response from submitter, so closing.

If something doesn't work correctly with the code as it currently is, please provide a small, self-contained example which shows the problem.

PR #3184, opened 2025-05-25
Python: Use multi-phase initialization

### LINKED ISSUE #3168: Support Python multi-phase init
Currently SWIG generates python extensions with [single-phase initialization](https://docs.python.org/3/c-api/module.html#single-phase-initialization): 

https://github.com/swig/swig/blob/4167b0275adc574f86914e42fb5ef9639b2b25a5/Lib/python/pyinit.swg#L275

The more modern [multi-phase initialization](https://docs.python.org/3/c-api/module.html#multi-phase-initialization) was introduced in 2013 and offers advantages over the older single-phase init. For a more detailed discussion on the advantages of multi-phase init and possible deprecation of single-phase init see:

https://discuss.python.org/t/proposal-officially-deprecate-support-for-legacy-single-phase-init-extension-modules/89262

jschueller:
Implements https://peps.python.org/pep-0489/
Bumps minimal python3 version from 3.4 to 3.5.
The idea is to move the initialization of the module in a new SWIG_mod_exec function.

Closes #3168

TODO: Figure out why it makes the swigvarlink warning reappear: DeprecationWarning: builtin type swigvarlink has no __module__ attribute make python_pybuffer.cpptest

### DIFF
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -11,8 +11,8 @@
 # error "This version of SWIG only supports Python >= 2.7"
 #endif
 
-#if PY_VERSION_HEX >= 0x03000000 && PY_VERSION_HEX < 0x03030000
-# error "This version of SWIG only supports Python 3 >= 3.3"
+#if PY_VERSION_HEX >= 0x03000000 && PY_VERSION_HEX < 0x03050000
+# error "This version of SWIG only supports Python 3 >= 3.5"
 #endif
 
 /* Common SWIG API */
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -162,7 +162,7 @@ jobs:
           SWIG_FEATURES: -builtin -O
         - SWIGLANG: python
           VER: '3.8'
-          PY_ABI_VER: '3.4'
+          PY_ABI_VER: '3.5'
         - SWIGLANG: r
           os: ubuntu-24.04
         - SWIGLANG: ruby
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -4174,19 +4174,19 @@ This is enabled by defining <tt>Py_LIMITED_API</tt> during the compilation of th
 </p>
 
 <p>
-SWIG supports the stable ABI, but only version 3.4 of Python and later is supported.
-There are two recommended approaches for using SWIG and the stable ABI and both require setting the <tt>Py_LIMITED_API</tt> macro to be set to <tt>0x03040000</tt> as a minimum value (Python 3.4).
+SWIG supports the stable ABI, but only version 3.5 of Python and later is supported.
+There are two recommended approaches for using SWIG and the stable ABI and both require setting the <tt>Py_LIMITED_API</tt> macro to be set to <tt>0x03050000</tt> as a minimum value (Python 3.5).
 Either set this using <tt>%begin</tt> by adding the following into your interface file so that this macro appears at the beginning of the generated C/C++ code:
 </p>
 
 <div class="code"><pre>
 %begin %{
-#define Py_LIMITED_API 0x03040000
+#define Py_LIMITED_API 0x03050000
 %}
 </pre></div>
 
 <p>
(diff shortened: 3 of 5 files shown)


wsfulton:
Which Github Actions build (python version) shows the warning? swigvarlink no longer exist in master since 53453d51ec3ca111c5c97530d8ee4c6b5184c85e, so maybe you are using some other version.

jschueller:
indeed you're right; no warnings in a clean rebuild

Lib/python/pyrun.swg
```diff
@@ -11,8 +11,8 @@
 # error "This version of SWIG only supports Python >= 2.7"
 #endif
 
-#if PY_VERSION_HEX >= 0x03000000 && PY_VERSION_HEX < 0x03030000
-# error "This version of SWIG only supports Python 3 >= 3.3"
+#if PY_VERSION_HEX >= 0x03000000 && PY_VERSION_HEX < 0x03050000
+# error "This version of SWIG only supports Python 3 >= 3.5"
 #endif
```
ojwb, line 16: This seems to update the hard requirement to Python 3 >= 3.5 but you only updated the docs to say that for the stable ABI case.

Is updating the hard requirement what you intended?  If so, the docs should be updated as currently they say:

> SWIG is compatible with all recent Python versions (Python 2.7 and Python &gt;= 3.5).  Python 3.3 and 3.4 may still work but we are no longer regularly testing with them. SWIG-4.0 supported Python 3.2.  SWIG-3.0 supported older Python 2.x and 3.x.

So delete the second sentence and record the last version to support 3.4.

jschueller: yes, it updates the requirement for both stable and non stable, I removed the sentence

wsfulton:
Please don't merge this just yet, I need some other pending changes put in before accepting this patch.

wsfulton:
Hi @jschueller, could you please rebase on top of current master and adjust if necessary for the recent changes for heap types for the builtin wrappers.

jschueller:
hello, I rebased on master and saw no conflicts

jschueller:
there was a problem which I fixed but now there is a problem with only one job, may be unrelated

chris-eibl:
`LINK : fatal error LNK1104: cannot open file 'python313t.lib'` pretty much looks like https://github.com/python/cpython/issues/135151 which resulted (amongst others) in an unscheduled 3.13.5 release https://discuss.python.org/t/python-3-13-5-is-now-available-yes-really/95211

ojwb:
> `LINK : fatal error LNK1104: cannot open file 'python313t.lib'` pretty much looks like [python/cpython#135151](https://github.com/python/cpython/issues/135151) which resulted (amongst others) in an unscheduled 3.13.5 release https://discuss.python.org/t/python-3-13-5-is-now-available-yes-really/95211

Thanks, but we already know (see #3198).

jschueller:
it's all green now

wsfulton:
Thanks for the patch.

This patch bumps the minimum python 3 version from python-3.4 to python-3.5 with the switch to `PyModuleDef_Init`. I think we've decided that is okay. But I'll note here that python-3.4 could still have been supported if necessary (with single phase initialization by using the original approach that is still currently in the code for python-2.7).

I note that the patch implements something very similar to the legacy init described at https://peps.python.org/pep-0489/#legacy-init, where 'legacy' is between python-3.0 and python-3.4. In our case, it provides single phase initialization for python-2.7 and multi-phase initialization > python-3.5. 

However, there is just a Py_mod_exec slot and no Py_mod_create slot. I understand this is still multi-phase initialization as a PyModuleDef type is returned by calling PyModuleDef_Init. However no Py_mod_create slot is provided, so a default is used. I don't think we need to customise this phase with a custom Py_mod_create slot.

I suppose the next steps for anyone interested is to support module reloading.

wsfulton:
Merged in 6c3bc2d18a864a2515d4abe59ba4f9c2ff36f84f. Also see 2e363dc8b2a39abc4d8645689c58557731167845.

PR #3192, opened 2025-06-09
[Python] Eliminate swig::stop_iteration exception

### LINKED ISSUE #3189: Multiple modules leads to unhandled swig::stop_iteration
This bug was [previously submitted by @benmwebb on Sourceforge in 2008](https://sourceforge.net/p/swig/bugs/881/), and then closed by @wsfulton  with the note "*Closing as can't reproduce. Please re-open if reproducible on a modern Linux distribution.*"

Recently, @weinhuber reported [a similar issue](https://gitlab.lre.epita.fr/spot/spot/-/issues/603) against a package of mine that would work fine in Ubuntu, but not on a Mac M2.  I don't have access to such a machine, but I was able to reproduce the issues (the one in my package, and the small example from 2008) on a modern OpenBSD distribution.  The common factor is that both the Mac and OpenBSD are using LLVM's libc++, while Linux-based distributions use libstdc++.

Here is the bug as seen on OpenBSD 7.7: 

```console
verycloudy4$ cat module1.i                                                                                                                                                                    
%module module1
%include "std_vector.i"
%template(vectori) std::vector<int>;
verycloudy4$ cat module2.i  
%module module2
%include "std_vector.i"
%template(vectorf) std::vector<float>;
verycloudy4$ swig -python -c++ module1.i                                                  
verycloudy4$ swig -python -c++ module2.i 
verycloudy4$ c++ -g -shared -fPIC -I /usr/local/include/python3.12 -o _module1.so module1_wrap.cxx 
verycloudy4$ c++ -g -shared -fPIC -I /usr/local/include/python3.12 -o _module2.so module2_wrap.cxx 
verycloudy4$ cat test.py                                                                                                                                                                      
import module1
import module2
a = module1.vectori()
for b in a: print(b)
a = module2.vectorf()
for b in a: print(b)
verycloudy4$ python3 test.py                                                                                                                                                                  
libc++abi: terminating due to uncaught exception of type swig::stop_iteration
Abort trap (core dumped) 
verycloudy4$ egdb python3 python3.12.core                                                          
GNU gdb (GDB) 15.2
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-unknown-openbsd7.7".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from python3...
(No debugging symbols found in python3)
[New process 530494]
Core was generated by `python3.12'.
Program terminated with signal SIGABRT, Aborted.
#0  thrkill () at /tmp/-:2

warning: 2      /tmp/-: No such file or directory
(gdb) bt
#0  thrkill () at /tmp/-:2
#1  0x765bc84b1729365a in ?? ()
#2  0x00000c41c47448cb in _libc_abort () at /usr/src/lib/libc/stdlib/abort.c:51
#3  0x00000c4176ad6f2e in abort_message (format=0xc4176aaa5b0 "terminating due to %s exception of type %s") at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/abort_message.cpp:78
#4  0x00000c4176ad58c6 in demangling_terminate_handler () at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_default_handlers.cpp:77
#5  0x00000c4176ab5a0a in std::__terminate (func=0x0) at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_handlers.cpp:59
#6  0x00000c4176ab538c in __cxxabiv1::failed_throw (exception_header=0xc411c33b400) at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_exception.cpp:152
#7  0x00000c4176ab52e0 in __cxa_throw (thrown_object=0xc411c33b480, tinfo=0xc41d5b6f218 <typeinfo for swig::stop_iteration>, dest=0x0)
    at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_exception.cpp:283
#8  0x00000c41d5b5d10f in swig::SwigPyForwardIteratorClosed_T<std::__1::__wrap_iter<int*>, int, swig::from_oper<int> >::value (this=0xc41b949ea80) at module1_wrap.cxx:4294
#9  0x00000c4180528e8b in swig::SwigPyIterator::next (this=0xc41b949ea80) at module2_wrap.cxx:3391
#10 0x00000c4180528ee3 in swig::SwigPyIterator::__next__ (this=0xc41b949ea80) at module2_wrap.cxx:3400
#11 0x00000c4180519e5f in _wrap_SwigPyIterator___next__ (self=<module at remote 0xc40f593e7f0>, args=Python Exception <class 'gdb.error'>: No type named PyDictValues.
) at module2_wrap.cxx:5504
#12 0x00000c41b8513450 in cfunction_vectorcall_O.llvm.15958724407970784865 () from /usr/local/lib/libpython3.12.so.0.0
#13 0x00000c41b84a5d32 in PyObject_Vectorcall () from /usr/local/lib/libpython3.12.so.0.0
#14 0x00000c41b85ed323 in _PyEval_EvalFrameDefault () from /usr/local/lib/libpython3.12.so.0.0
#15 0x00000c41b853db8c in vectorcall_method () from /usr/local/lib/libpython3.12.so.0.0
#16 0x00000c41b8541a44 in slot_tp_iternext () from /usr/local/lib/libpython3.12.so.0.0
#17 0x00000c41b85ebe3b in _PyEval_EvalFrameDefault () from /usr/local/lib/libpython3.12.so.0.0
#18 0x00000c41b85dfbfa in PyEval_EvalCode () from /usr/local/lib/libpython3.12.so.0.0
#19 0x00000c41b866a6a7 in run_mod.llvm () from /usr/local/lib/libpython3.12.so.0.0
#20 0x00000c41b8667c5b in _PyRun_SimpleFileObject () from /usr/local/lib/libpython3.12.so.0.0
#21 0x00000c41b86670f8 in _PyRun_AnyFileObject () from /usr/local/lib/libpython3.12.so.0.0
#22 0x00000c41b869cceb in Py_RunMain () from /usr/local/lib/libpython3.12.so.0.0
#23 0x00000c41b869d130 in pymain_main () from /usr/local/lib/libpython3.12.so.0.0
#24 0x00000c41b869d1ac in Py_BytesMain () from /usr/local/lib/libpython3.12.so.0.0
#25 0x00000c3edd31e94b in _start ()
(gdb) quit
verycloudy4$ c++ --version
OpenBSD clang version 16.0.6
Target: amd64-unknown-openbsd7.7
Thread model: posix
InstalledDir: /usr/bin
verycloudy4$ swig -version  

SWIG Version 4.2.1

Compiled with c++ [x86_64-unknown-openbsd7.7]

Configured options: +pcre

Please see https://www.swig.org for reporting bugs and further information
```

Here is my understanding of how the problem unfolds:
1. both modules define a `swig::stop_iteration` type that can be thrown.
2. both modules define the same `SwigPyIterator` base class that provides common implementation methods for iterators.  In particular `SwigPyIterator::__next__()` calls a virtual `value()` method and will convert any `swig::stop_iteration()` thrown by that method into Python's StopIteration. 
3. Each module will define a different subclass of `SwigPyIterator` to implement iteration over `vectori` or over `vectorf`.  In particular, they implement the aforementioned `value()` method that may throw `swig::stop_iteration`.
5. When Python loads a module with dlopen(), this is done with `RTLD_LOCAL` so the dynamic linker will *not* merge the type_info for the two instances of `swig::stop_iteration` as normally done when multiple libraries define the same symbols.  So these two instances of type info for `swig::stop_iteration` have different addresses.
6. When both modules are initialized, they share their Swig types via some mechanism in the swig runtime, so the two definitions from SwigPyIterator are unified: this Python type now refers to one of the two implementations of that class depending on load order.
7. When the first iteration of `test.py` calls `SwigPyIterator::__next__()` this is resolved as _wrap_SwigPyIterator___next__ from the second module.
8. the above function eventually dispatches a call to `swig::SwigPyForwardIteratorClosed_T<...>::value()` from module1, that raises `swig::stop_iteration` with the type_info from module1.
9. the catch of this exception is supposed to happen in `_wrap_SwigPyIterator___next__`, but that function is the one from module 2, so it will only recognize the `swig::stop_iteration` with the type_info from module 2.    Since the raised exception does not have the correct type_info it is not caught.

We don't see this on Linux, because [since 2009](https://gcc.gnu.org/legacy-ml/gcc-patches/2009-07/msg01239.html) `libstdc++` is aware that in this kind of situation (dlopen with RTLD_LOCAL) type_info symbols might not be merged by the dynamic linker.  Therefore it defaults to comparing type_info by their name string if the pointer comparison fails.  On system using libc++, [the default is to assume that type info symbols are unique](https://github.com/llvm/llvm-project/blob/c3656afdf111901977792c1d9589cb32ee1b6015/libcxx/include/typeinfo#L120-L187) unless you are on a system where those macros decide otherwise.

Is there something that can be done in Swig, or in my use of Swig, that will make this work with both setups?   From my point of view I'd be happy if `swig::stop_iteration` was some internal detail in each module, and if the code throwing and catching that exception was always in the same module.  This way it does not matter if the C++ runtime consider the two `swig::stop_iteration` as separate types or not.   Is there an easy way to rename the SwigPyIterator class in each module (so that one will not shadow the other one), while allowing type conversions between both types?

ojwb:
Instead return NULL and check for that, which should be lower overhead and avoids problems with having swig::stop_iteration classes thrown as exceptions in multiple SWIG-wrapped modules.

Fixes #3189

### DIFF
--- a/Lib/python/pyiterators.swg
+++ b/Lib/python/pyiterators.swg
@@ -13,9 +13,6 @@
 
 %fragment("SwigPyIterator","header",fragment="<stddef.h>") {
 namespace swig {
-  struct stop_iteration {
-  };
-
   struct SwigPyIterator {
   private:
     SwigPtr_PyObject _seq;
@@ -37,7 +34,7 @@ namespace swig {
     // Backward iterator method, very common in C++, but not required in Python
     virtual SwigPyIterator *decr(size_t /*n*/ = 1)
     {
-      throw stop_iteration();
+      return NULL;
     }
 
     // Random access iterator methods, but not required in Python
@@ -271,7 +268,7 @@ namespace swig {
     
     PyObject *value() const {
       if (base::current == end) {
-	throw stop_iteration();
+	return NULL;
       } else {
 	return from(static_cast<const value_type&>(*(base::current)));
       }
@@ -286,7 +283,7 @@ namespace swig {
     {
       while (n--) {
 	if (base::current == end) {
-	  throw stop_iteration();
+	  return NULL;
 	} else {
 	  ++base::current;
 	}
@@ -321,7 +318,7 @@ namespace swig {
     {
       while (n--) {
 	if (base::current == base0::begin) {
-	  throw stop_iteration();
+	  return NULL;
 	} else {
 	  --base::current;
 	}
@@ -366,17 +363,18 @@ namespace swig {
 %fragment("SwigPyIterator");
 namespace swig 
 {
-  /*
-    Throw a StopIteration exception
-  */
-  %ignore stop_iteration;
-  struct stop_iteration {};
-  
-  %typemap(throws) stop_iteration {
-    (void)$1;
-    SWIG_SetErrorObj(PyExc_StopIteration, SWIG_Py_Void());
-    SWIG_fail;
-  }
+#define SWIG_PY_RAISE_STOP_ITERATION { $action if (result == NULL) { SWIG_SetErrorObj(PyExc_StopIteration, SWIG_Py_Void()); SWIG_fail; } }
+  %exception SwigPyIterator::value() const SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::incr(size_t n = 1) SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::decr(size_t n = 1) SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::__next__() SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::next() SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::previous() SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::advance(ptrdiff_t n) SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::operator += (ptrdiff_t n) SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::operator -= (ptrdiff_t n) SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::operator + (ptrdiff_t n) const SWIG_PY_RAISE_STOP_ITERATION
+  %exception SwigPyIterator::operator - (ptrdiff_t n) const SWIG_PY_RAISE_STOP_ITERATION
 
   /* 
      Mark methods that return new objects
@@ -397,19 +395,8 @@ namespace swig
   }
 #endif
 
-  %catches(swig::stop_iteration) SwigPyIterator::value() const;
-  %catches(swig::stop_iteration) SwigPyIterator::incr(size_t n = 1);
-  %catches(swig::stop_iteration) SwigPyIterator::decr(size_t n = 1);
   %catches(std::invalid_argument) SwigPyIterator::distance(const SwigPyIterator &x) const;
   %catches(std::invalid_argument) SwigPyIterator::equal (const SwigPyIterator &x) const;
-  %catches(swig::stop_iteration) SwigPyIterator::__next__();
-  %catches(swig::stop_iteration) SwigPyIterator::next();
-  %catches(swig::stop_iteration) SwigPyIterator::previous();
-  %catches(swig::stop_iteration) SwigPyIterator::advance(ptrdiff_t n);
-  %catches(swig::stop_iteration) SwigPyIterator::operator += (ptrdiff_t n);
-  %catches(swig::stop_iteration) SwigPyIterator::operator -= (ptrdiff_t n);
-  %catches(swig::stop_iteration) SwigPyIterator::operator + (ptrdiff_t n) const;
-  %catches(swig::stop_iteration) SwigPyIterator::operator - (ptrdiff_t n) const;
 
   struct SwigPyIterator
   {


adl:
This looks nice!

The only downside is that it will break user-defined iterators.  In my wrapper I have a few iterator-like objects that do not follow the STL begin/end interface.  I still use the SwigPyIterator mechanism by implementing a subclass manually.   To try this patch, I had to adjust several places in my own `*.i` files as follows:

```patch
--- a/python/spot/impl.i
+++ b/python/spot/impl.i
@@ -223,7 +223,11 @@ namespace swig
 
     PyObject *value() const {
       if (base::current == end) {
+#ifdef SWIG_PY_RAISE_STOP_ITERATION
+        return nullptr;
+#else
         throw stop_iteration();
+#endif
       } else {
        return from(static_cast<value_type>(&*(base::current)));
       }
```

Aside from that, the patch works for me.  Thank you!

adl:
I forgot this side note on `NULL` and `nullptr`.   Maybe that's a separate issue.

The wrapper code generated by Swig for Python bindings define the following:
```c++
#if defined(__cplusplus) && __cplusplus >=201103L
# define SWIG_NULLPTR nullptr
#else
# define SWIG_NULLPTR NULL
#endif 
```
Yet, currently all the generated code (this patch included) uses `NULL` directly (sometimes even `0`), including in C++ constructs.  It seems `SWIG_NULLPTR` is used in bindings for some other target languages.  Using `SWIG_NULLPTR` instead of `NULL` would help people compiling with `-Wzero-as-null-pointer-constant`.

ojwb:
I removed `swig::stop_iteration` in the patch as it's not mentioned in the documentation, so seems to be entirely an internal implementation detail.  If you're using something like that then you shouldn't be surprised if it arbitrarily disappears or changes.  AIUI you're using it rather than defining your own `struct my_stop_iteration {};` because you're relying on SWIG's code catching it?  If so leaving `swig::stop_iteration` in place would actually be worse because the code would compile but when you throw the exception it won't get caught and handled like before.  Leaving the exception handling for it in place as well seems overkill just to keep compatibility with people using SWIG internals.

`SWIG_NULLPTR` was added by people wanting to placate a particular static analysis tool.  However the same tool complains if we don't use `auto` which we can't sensibly work around with more ugly macros (we'd need to replace `int` with something like `SWIG_AUTO(int)` which expands to `auto` or `int`, but then we'd have to worry about having to test both versions because `auto` could deduce a different type to the static type, etc).  Our development resources are limited and we struggle to keep up with bugs and patches, and this just seems to be sucking up time and effort for marginal benefit.  I think we're concluding it was a mistake to introduce `SWIG_NULLPTR` and we should just direct people using such tools to suppress certain warnings for SWIG code (see https://github.com/swig/swig/pull/3165#issuecomment-2846485979 and following comments).

adl:
I was just mentioning the two points I noticed while trying the patch to provide feedback.  I don't mind them.

The main reason I had to write my own iterator code is that the default iterators provided by Swig return values by const references, while I want values returned by pointers (so that I can use the iterators to modify the object I'm iterating on).  Rather than implementing Python's iterator protocol in my own iterator object, I simply defined a class similar to `SwigPyForwardIteratorOpen_T`, but that returns pointers.  Yes, I depends on several internal details of `pyiterators.swg` and will have to adapt my code when that changes.  I'm fine with this, it's still less effort than writing my own class.

Having said that, I still have a question about such an interface change.  Even if `SwigPyIterator` can be considered as in implementation detail, it is a public type exported to Python.  In #3189, the issue was that two modules were defining SwigPyIterator differently (each expecting a different copy of `swig::stop_iteration`), and Swig's type system would only keep one copy of this type.  With this patch `swig::stop_iteration` is out of the equation (good riddance), but this actually introduces a different implementation for `SwigPyIterator` that keeps the same name.  So if you load two Python packages generated with different versions of Swig (one with the patch and one without) the iterators will break for one package, won't they?

ojwb:
Ah yes, that's likely to be a problem.

My first thought was to rename the C++ SwigPyIterator class (with `%rename` to keep `SwigPyIterator` in Python, each in the namespace of its module).  However if you are right that SWIG's type system is picking one of the `SwigPyIterator` classes to use everywhere that's not going to help.

ojwb:
I took your example and patched the `SwigPyIterator` constructor to report the name of the module it was in like so:

```
diff -u ./module1_wrap.cxx patched/module1_wrap.cxx
--- ./module1_wrap.cxx	2025-06-10 10:28:27.236737685 +1200
+++ patched/module1_wrap.cxx	2025-06-10 10:26:09.700655140 +1200
@@ -3561,6 +3561,7 @@
   protected:
     SwigPyIterator(PyObject *seq) : _seq(seq)
     {
+	printf("SwigPyIterator from %s\n", SWIG_name);
     }
       
   public:
diff -u ./module2_wrap.cxx patched/module2_wrap.cxx
--- ./module2_wrap.cxx	2025-06-10 10:28:27.704737964 +1200
+++ patched/module2_wrap.cxx	2025-06-10 10:26:13.224657272 +1200
@@ -3561,6 +3561,7 @@
   protected:
     SwigPyIterator(PyObject *seq) : _seq(seq)
     {
+	printf("SwigPyIterator from %s\n", SWIG_name);
     }
       
   public:
```

Then with this branch using both `g++` and `clang++ -stdlib=libc++` I get output:

```
SwigPyIterator from _module1
SwigPyIterator from _module2
```

So I think the problem in #3189 is just due to symbol collisions from dynamic linking at the C/C++ level, and not related to SWIG's type system (the Python classes are `module1.SwigPyIterator` and `module2.SwigPyIterator` so I wouldn't really expect them to be treated as the same - AIUI the case where SWIG's type system combines typeinfo is when one module knows about classes from another module thanks to `%import` but only has stub typeinfo for them until runtime when the other module is loaded).

Possibly we should encode the module name in the C++ name of `SwigPyIterator` and similar classes - e.g. `swig::mod_example::SwigPyIterator` where `example` is the module name - that would help protect against these sort of dynamic linking problems more generally.

ojwb:
Or put them in an anonymous namespace which seems cleaner, assuming it can be made to work here.

adl:
> I took your example and patched the `SwigPyIterator` constructor to report the name of the module it was in like so:

I'll try this on OpenBSD when I get a chance (in a few hours). 

> So I think the problem in #3189 is just due to symbol collisions from dynamic linking at the C/C++ level, and not related to SWIG's type system (the Python classes are `module1.SwigPyIterator` and `module2.SwigPyIterator` so I wouldn't really expect them to be treated as the same - AIUI the case where SWIG's type system combines typeinfo is when one module knows about classes from another module thanks to `%import` but only has stub typeinfo for them until runtime when the other module is loaded).

I don't fully understand the inner workings of Swig's type system.  All I know is that in the backtrace I posted, `swig::SwigPyIterator::next` from module 2 is calling `swig::SwigPyForwardIteratorClosed_T` from module 1, during the iteration on the vector from module 1.  If the two SwigPyIterator are not combined (that was my assumption), I don't understand how `swig::SwigPyIterator::next` get called from the wrong module.   Dynamic linking with RTLD_LOCAL will not merge symbols, that's the point.  I'll try to add more traces too.  

> Possibly we should encode the module name in the C++ name of `SwigPyIterator` and similar classes - e.g. `swig::mod_example::SwigPyIterator` where `example` is the module name - that would help protect against these sort of dynamic linking problems more generally.

That's currently my workaround: I do `sed -i 's/SwigPyIterator/SwigPyIterator_module_1/g' module1_wrap.cxx module1.py` once Swig has generated its code.

adl:
I did that:
```patch
--- module1_wrap.cxx.orig       Tue Jun 10 10:03:57 2025
+++ module1_wrap.cxx    Tue Jun 10 10:18:24 2025
@@ -3354,6 +3354,7 @@
   protected:
     SwigPyIterator(PyObject *seq) : _seq(seq)
     {
+       printf("SwigPyIterator from %s (this=%p)\n", SWIG_name, this);
     }
       
   public:
@@ -3387,6 +3388,7 @@
 
     PyObject *next()     
     {
+      printf("SwigPyIterator::next from %s (this=%p)\n", SWIG_name, this);
       SWIG_PYTHON_THREAD_BEGIN_BLOCK; // disable threads       
       PyObject *obj = value();
       incr();       
@@ -3397,6 +3399,7 @@
     /* Make an alias for Python 3.x */
     PyObject *__next__()
     {
+      printf("SwigPyIterator::__next__ from %s (this=%p)\n", SWIG_name, this);
       return next();
     }
 
```
and did the same in the wrapper for module2.  On OpenBSD, I'm using Swig 4.2.1 **unpatched**.

```console
$ python3                                        
Python 3.12.10 (main, May  5 2025, 16:26:16) [Clang 16.0.6 ] on openbsd7
>>> import module1
>>> import module2
>>> a = module1.vectori()
>>> i = a.begin()
SwigPyIterator from _module1 (this=0x83344090060)
>>> type(i)
<class 'module2.SwigPyIterator'>
>>> next(i)
SwigPyIterator::__next__ from _module2 (this=0x83344090060)
SwigPyIterator::next from _module2 (this=0x83344090060)
```
So SwigIterator `i` is indeed constructed by _module1, but it gets cast into the version from _module2, somehow.

ojwb:
Testing git master unpatched on Linux, with `g++` I get:
```
SwigPyIterator from _module1 (this=0x180ae690)
SwigPyIterator::__next__ from _module2 (this=0x180ae690)
SwigPyIterator::next from _module2 (this=0x180ae690)
SwigPyIterator from _module2 (this=0x180ae690)
SwigPyIterator::__next__ from _module2 (this=0x180ae690)
SwigPyIterator::next from _module2 (this=0x180ae690)
```
And with `clang++ -stdlib=libc++`:
```
SwigPyIterator from _module1 (this=0x1024ea70)
SwigPyIterator::__next__ from _module2 (this=0x1024ea70)
SwigPyIterator::next from _module2 (this=0x1024ea70)
libc++abi: terminating due to uncaught exception of type swig::stop_iteration
Aborted (core dumped)
```

At first glance this seems like it could be explained by there being a `SwigPyIterator::next()` symbol (strictly speaking the mangled form of that symbol) in both `_module1.so` and `_module2.so` and the dynamic linker resolving uses of it in `_module1.so` to the version in `_module2.so`.

However you said in the issue that `RTLD_LOCAL` was in use, which should prevent that happening.  Also with `RTLD_GLOBAL`, the docs suggest the symbol from the module loaded **first** gets be used (which also makes sense if symbols are resolved at `dlopen()` time as they are with `RTLD_NOW`).

I used `ltrace` to check what flags are getting passed to `dlopen()` and it's `2` in both cases for me, which looking in the headers is `RTLD_NOW`.  Linux man page as `RTLD_LOCAL` is the default.

I get essentially the same output with the PR branch, except without the uncaught exception.

And with the PR branch plus a patch to put `SwigPyIterator` in a anonymous namespace, I still get essentially the same output.  Patch:

```
diff --git a/Lib/python/pyiterators.swg b/Lib/python/pyiterators.swg
index 532095e8..3edc9934 100644
--- a/Lib/python/pyiterators.swg
+++ b/Lib/python/pyiterators.swg
@@ -12,10 +12,10 @@
 %include <std_common.i>
 
 %fragment("SwigPyIterator","header",fragment="<stddef.h>") {
-namespace swig {
+namespace {
   struct SwigPyIterator {
   private:
-    SwigPtr_PyObject _seq;
+    swig::SwigPtr_PyObject _seq;
 
   protected:
     SwigPyIterator(PyObject *seq) : _seq(seq)
@@ -129,6 +129,9 @@ namespace swig {
   }
 %#endif
 }
+namespace swig {
+  using ::SwigPyIterator;
+}
 }
 
 %fragment("SwigPyIterator_T","header",fragment="<stddef.h>",fragment="SwigPyIterator",fragment="StdTraits",fragment="StdIteratorTraits") {
```

So it seems this indeed isn't due to dynamic linking symbol resolution, or at least not of symbols in `SwigPyIterator`.

Adding more printf calls, we call `_wrap_SwigPyIterator___next__() in _module2` but that's `static`, and that's set as the C function that Python `SwigPyIterator___next__` calls, which is a function in each module.

And in each module we have:

```
class SwigPyIterator(object):
```
...
```
    def __next__(self):
        return _module2.SwigPyIterator___next__(self)
```

Adding a Python `print` there I get (with gcc, FWIW):

```
SwigPyIterator from _module1 (this=0x2ad03f90)
module2 SwigPyIterator.next()
_wrap_SwigPyIterator___next__() in _module2
SwigPyIterator::__next__ from _module2 (this=0x2ad03f90)
SwigPyIterator::next from _module2 (this=0x2ad03f90)
SwigPyIterator from _module2 (this=0x2ad03f90)
module2 SwigPyIterator.next()
_wrap_SwigPyIterator___next__() in _module2
SwigPyIterator::__next__ from _module2 (this=0x2ad03f90)
SwigPyIterator::next from _module2 (this=0x2ad03f90)
```

Looking around I find:

```
# Register SwigPyIterator in _module2:
_module2.SwigPyIterator_swigregister(SwigPyIterator)
```

and:

```
SWIGINTERN PyObject *SwigPyIterator_swigregister(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
  PyObject *obj = NULL;
  if (!SWIG_Python_UnpackTuple(args, "swigregister", 1, 1, &obj)) return NULL;
  SWIG_TypeNewClientData(SWIGTYPE_p_swig__SwigPyIterator, SWIG_NewClientData(obj));
  return SWIG_Py_Void();
}
```

Presumably that then gets sucked into the process which is intended to resolve types (the case where there's a type defined in moduleA which moduleB knows about via `%import`) and we end up with one `SwigPyIterator` entry in SWIG's type system which points to module2's `SwigPyIterator`.  Not sure why we get `SwigPyIterator from _module1` though.

adl:
> Not sure why we get SwigPyIterator from _module1 though.

I think it's because the method that creates the iterator is a method from `module1.vectori`, so the C++ code that creates the iterator is still code from `module1`, and at that point the iterator hasn't been turned into a Python object yet.

ojwb:
Ah yes, that seems plausible.

The major problem here is that any change needs to be compatible in both directions because with multiple modules the load order will (I think) determine module's `SwigPyIterator` gets used for all of them.  So we would need a "flag day" compatibility break where you can only combine modules which are either all before or all after.

**However**, we've already bumped `SWIG_RUNTIME_VERSION` to 5 for 4.4.0, so our timing is very fortunate and I think we can actually merge this for 4.4.0 without worrying about backward compatibility.  I'll set the milestone so we don't miss that window without a conscious decision to.

@wsfulton Does that seem OK?  If so I can try to extend this change to all the languages which currently throw `swig::stop_iteration`.

adl:
> we've already bumped SWIG_RUNTIME_VERSION to 5 for 4.4.0, so our timing is very fortunate

Perfect!

vadz:
Should this be still marked as "Draft"? AFAICS it should be merged and it seems like a much better approach to detecting the end of the iteration than the current one (without speaking of the fact that it fixes #3189).

ojwb:
I have left it as draft because I was hoping for some sort of consensus that it's OK to make this change as part of runtime version 5 before going to the effort of doing it for other target languages which use essentially the same approach.

We don't absolutely have to change this everywhere at once but I really think we should as consistency is helpful and it just seems a much better mechanism, but we don't bump the runtime version often these days (though interestingly we used to do it every year or two: v5 in 2025, v4 in 2008, v3 in 2006, v2 in 2005).

ojwb:
@wsfulton offered on the mailing list to take this over, so I'm going to close this PR as he wanted a different approach and it'll be easier for him to work in his own branch anyway.

I'll leave my branch in place for now at least so it's easier to pull out any parts that are useful.

Issue #3203, opened 2025-06-24
Python -builtin segfault if attempt to delete struct member


jim-easterbrook:
I've found a bug if a user tries to delete an attribute of a struct wrapped with SWIG.

Example interface file `struct_del_member_segfault.i`:
```
%module struct_del_member_segfault

%inline %{
struct Foo {
  int x;
};
%}
```

Extension built with `setup.py`
```
from distutils.core import setup, Extension

setup (name = 'struct_del_member_segfault',
       ext_modules = [Extension(
           '_struct_del_member_segfault',
           sources=['struct_del_member_segfault_wrap.c'])],
       py_modules = ["struct_del_member_segfault"],
       )
```

Test file `test.py`:
```
import struct_del_member_segfault

s = struct_del_member_segfault.Foo()
print('get member')
print(s.x)
print('set member')
s.x = 123
print('del member')
del s.x
```

Build and run:
```
$ swig -python -builtin -nofastunpack struct_del_member_segfault.i 
$ python3 setup.py build_ext --inplace
running build_ext
building '_struct_del_member_segfault' extension
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -DOPENSSL_LOAD_CONF -fwrapv -fno-semantic-interposition -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -fPIC -I/usr/include/python3.6m -c struct_del_member_segfault_wrap.c -o build/temp.linux-x86_64-3.6/struct_del_member_segfault_wrap.o
gcc -pthread -shared build/temp.linux-x86_64-3.6/struct_del_member_segfault_wrap.o -L/usr/lib64 -lpython3.6m -o /home/jim/Documents/projects/swig/bugs/struct_del_member_segfault/_struct_del_member_segfault.cpython-36m-x86_64-linux-gnu.so
$ python3 test.py 
get member
0
set member
[...]

PR #3206, opened 2025-06-26
Check for NULL in Python builtin setter closure

jim-easterbrook:
This prevents a segfault, as reported in bug
https://github.com/swig/swig/issues/3203


wsfulton:
Merged in e5a76eef91e10b79e9c21edf7720dc8f29e356ef and tweaked and a testcase added in 31dba786a21916dc9874c7bc83fe977b21a929a5. Thanks.

jim-easterbrook:
Thanks William. And I like your more consistent choice of exception class.

Issue #3209, opened 2025-07-02
Python: unneeded PyErr_SetString if PyType_Ready fails


jim-easterbrook:
The `PyType_Ready` function used during type initialisation sets an exception if it fails, but swig over writes that exception with a "Could not create type ..." message.

An example module `PyType_Ready_error.i` (for use with swig v4.4):
```
%module PyType_Ready_error

%inline %{
// trigger a PyType_Ready failure by mixing heap and static types
#undef SWIG_HEAPTYPES

struct Foo {
  int x;
};
%}
```

`setup.py`
```
from distutils.core import setup, Extension

setup (name = 'PyType_Ready_error',
       ext_modules = [Extension(
           '_PyType_Ready_error',
           sources=['PyType_Ready_error_wrap.c'])],
       py_modules = ["PyType_Ready_error"],
       )
```

Build and run:
```
$ swig -python -builtin PyType_Ready_error.i
$ python3 setup.py build_ext --inplace
running build_ext
building '_PyType_Ready_error' extension
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -DOPENSSL_LOAD_CONF -fwrapv -fno-semantic-interposition -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -fmessage-length=0 -grecord-gcc-switches -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -g -fPIC -I/usr/include/python3.6m -c PyType_Ready_error_wrap.c -o build/temp.linux-x86_64-3.6/PyType_Ready_error_wrap.o
gcc -pthread -shared build/temp.linux-x86_64-3.6/PyType_Ready_error_wrap.o -L/usr/lib64 -lpython3.6m -o /home/jim/Documents/projects/swig/bugs/PyType_Ready_error/_PyType_Ready_error.cpython-36m-x86_64-linux-gnu.so
$ python3 -c "import PyType_Ready_error"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
[...]

PR #3210, opened 2025-07-02
Remove PyErr_SetString if type init fails

jim-easterbrook:
PyType_Ready sets an exception which should not be over written. See bug #3209.


wsfulton:
Thanks. I checked the other locations we use PyType_Ready they handle the error correctly.

Issue #3212, opened 2025-07-02
Python free-threading: segfault on all Python versions (including with GIL)


crusaderky:
I've recently finished updating sentencepiece to support free-threading (https://github.com/google/sentencepiece/issues/1110).
The current status is that I can
1. regenerate bindings with `swig -c++ -python` using SWIG git tip (1fa70fb4a7adeb89b0b29e66d909db7f6310e7d0)
2. successfully run Python tests suite on Python 3.6 to 3.13
3. successfully run Python tests suite on Python 3.13t, including with pytest-run-paralllel, by setting the `PYTHON_GIL=0` environment variable.

This sane baseline state is reflected here https://github.com/crusaderky/sentencepiece/pull/7.

However, as soon as I regenerate the bindings with `swig -c++ -python -threads`, the test suite segfaults very fast - including on non-free-threaded Python, on tests that don't leverage any multi-threading.

The delta (including generated bindings) is here: https://github.com/crusaderky/sentencepiece/pull/6
Error message here: https://github.com/crusaderky/sentencepiece/actions/runs/16023450799/job/45205642614?pr=6
CI segfaults on Python 3.6; I'm reproducing the same failure locally on 3.13 and 3.13t too.

CC @ngoldbaum @wsfulton @klausspanderen from the original work on this (https://github.com/swig/swig/issues/3121, https://github.com/swig/swig/pull/3137)

ngoldbaum:
Have you reproduced it outside CI? If so, it'd probably be useful to see a C traceback for the segfault from gdb or lldb.

crusaderky:
Yes, I'm reproducing it locally.
Additionally to the above issue, the `-flags` parameter does not remove the need for the `PYTHON_GIL=0` flag.

gdb traceback:
```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00005555556d93e5 in _PyInterpreterState_GET () at /usr/local/src/conda/python-3.13.5/Include/internal/pycore_pystate.h:214
warning: 214    /usr/local/src/conda/python-3.13.5/Include/internal/pycore_pystate.h: No such file or directory
(gdb) bt
#0  0x00005555556d93e5 in _PyInterpreterState_GET () at /usr/local/src/conda/python-3.13.5/Include/internal/pycore_pystate.h:214
#1  _PyType_LookupRef (type=type@entry=0x3a80c294600, name=name@entry=0x555555a557e0 <_PyRuntime+85664>)
[...]

PR #3215, opened 2025-07-03
Python free-threading without PYTHON_GIL=0

crusaderky:
Run python 3.13t without needing to specify the PYTHON_GIL=0 environment variable. Clarify documentation.

# XREFs
- #3212
- #3121
- #3137
- https://github.com/google/sentencepiece/issues/1110


Lib/python/pyinit.swg
```diff
@@ -206,6 +206,9 @@ SWIG_init(void) {
 #if PY_VERSION_HEX >= 0x03000000
   static PyModuleDef_Slot SwigSlots[] = {
     { Py_mod_exec, (void *)SWIG_mod_exec },
+    #ifdef Py_GIL_DISABLED
+    {Py_mod_gil, Py_MOD_GIL_NOT_USED},
+    #endif
```
crusaderky: This is the change. Everything else is tests and documentation.

ojwb: I don't know Python well enough to comment on the content, but a coding style nit: please don't put whitespace before the `#`.

ojwb: Also the whitespace inside the `{`...`}` in the new code is inconsistent with the existing code just above it.

crusaderky:
FYI @ngoldbaum @wsfulton @klausspanderen

Lib/python/pyinit.swg
ngoldbaum: In other projects we made this setting configurable. Does it make sense in swig to do something similar?

You could imagine generating bindings for a thread-unsafe C library with swig. The bindings themselves might be thread-safe, but nothing in swig prevents someone from writing a Python program with the `threading` module that calls into the C library in a thread-unsafe way, unless the author of the wrapping library handles that with e.g. a lock outside of the bindings.

In other binding generators there's an option in the mechanism that creates a module to allow setting the gil used slot. 

Heres pybind11's docs for creating python modules with free-threaded support: https://pybind11.readthedocs.io/en/latest/advanced/misc.html#free-threading-support

Cython has a per-module compiler directive: https://cython.readthedocs.io/en/latest/src/userguide/freethreading.html

PyO3 also has an option for the PyModule macro: https://docs.rs/pyo3/latest/pyo3/types/trait.PyModuleMethods.html#tymethod.gil_used

wsfulton:
I suggest then that the `SwigSlots[]` are modified to contain:
```c
#ifdef Py_GIL_DISABLED
#ifdef SWIGPYTHON_NOGIL
      { Py_mod_gil, Py_MOD_GIL_NOT_USED },
#endif
#endif
```
and `SWIGPYTHON_NOGIL` is a macro generated by a new SWIG Python command line option `-nogil` which will work in the same way that `SWIGPYTHON_BUILTIN` is defined by the `-builtin` option. We can run the test-suite for this like we do for `-builtin` in .github/workflows/linux.yml. @crusaderky would you be able to make these changes to your pull request? 

It would also be useful to have our docs contain the advice to check if the GIL is disabled both before and after loading the SWIG module by using `sys._is_gil_enabled()` - or point to https://py-free-threading.github.io/running-gil-disabled/.

crusaderky:
Yes, I can make these changes.

crusaderky:
Changes to code and documentation done. Tests are WIP.

crusaderky:
@wsfulton @ngoldbaum this is ready for review again.

crusaderky:
> It would also be useful to have our docs contain the advice to check if the GIL is disabled both before and after loading the SWIG module by using `sys._is_gil_enabled()` - or point to https://py-free-threading.github.io/running-gil-disabled/.

This seems redundant? There is already a very loud warning from the python interpreter for that, and many (most?) test suites run with `-Werror`.

PR #3218, opened 2025-07-10
Python: add weakref support to builtin types

### LINKED ISSUE #1792: Python weakref with -builtin
When a python extension is compiled with -builtin, it can't work with weakref, 
```
>>> i = SomeSwigPyObject
>>> from weakref import ref
>>> ref(i)

TypeError: cannot create weak reference to 'SomeSwigPyObject' object
```

It seems to me that this is due to SwigPyObject doesn't have the `in_weakreflist` field.  Is there a switch or option to turn that on for -builtin python module?

jim-easterbrook:
These changes add a weakreflist member to SwigPyObject, and set tp_weaklistoffset to its offset. This fixes #1792.

The Py_TPFLAGS_MANAGED_WEAKREF introduced in Python 3.12 requires Py_TPFLAGS_HAVE_GC to be set as well, which it currently isn't for SwigPyObjectType. (See https://github.com/python/cpython/issues/134786).


wsfulton:
> The Py_TPFLAGS_MANAGED_WEAKREF introduced in Python 3.12 requires Py_TPFLAGS_HAVE_GC to be set as well, which it currently isn't for SwigPyObjectType. 

I don't follow how this is relevant to this pull request which doesn't explicitly use `Py_TPFLAGS_MANAGED_WEAKREF`. Maybe you had in mind that this is for the future?

If we need to set Py_TPFLAGS_HAVE_GC, let's have a separate issue for that which will need to discuss the implications.

Lib/python/builtin.swg
```diff
@@ -580,6 +580,8 @@ SWIGINTERN void
 SwigPyBuiltin_destructor_closure(SwigPyWrapperFunction wrapper, const char *wrappername, PyObject *a) {
   SwigPyObject *sobj;
   sobj = (SwigPyObject *)a;
+  if (sobj->weakreflist != NULL)
+    PyObject_ClearWeakRefs(a);
```
wsfulton, line 584: Can you think of some way to test that this is being called, that is, a way to test the weak reference is cleared?

The python_weakref_runme.py runtime test is lightweight, so if you can think of any way to beef it up, that would be good.

jim-easterbrook:
> > The Py_TPFLAGS_MANAGED_WEAKREF introduced in Python 3.12 requires Py_TPFLAGS_HAVE_GC to be set as well, which it currently isn't for SwigPyObjectType.
> 
> I don't follow how this is relevant to this pull request which doesn't explicitly use `Py_TPFLAGS_MANAGED_WEAKREF`. Maybe you had in mind that this is for the future?
> 
> If we need to set Py_TPFLAGS_HAVE_GC, let's have a separate issue for that which will need to discuss the implications.

I was attempting to explain why I'm not using `Py_TPFLAGS_MANAGED_WEAKREF` at all.

Lib/python/builtin.swg
jim-easterbrook: I'll give it some thought. I'm about to go on holiday, so probably won't do anything for a while.

jim-easterbrook: I found time to add some tests this morning.

wsfulton:
Merged thanks. I've made some tweaks in 3438289cadfe137bfcfda985600a8eaab1ed01ed as part of the longer term goal of also supporting the limited API for builtin wrappers.

PR #3223, opened 2025-07-19
Don't call PyWeakref_GET_OBJECT if limited API

jim-easterbrook:
PyWeakref_GET_OBJECT is not part of the limited API/stable ABI so call PyWeakref_GetObject instead.


wsfulton:
Thanks. I've cherry-picked this in 44684b9b55b6bd188f340fd3819f9396cd6492ec which then led me to make additional changes in 738db8f288bf5c7111598c8321db6a1252a67a33.

PR #3225, opened 2025-07-23
Free-threading Python: revert special treatment for PyList_GET_ITEM etc.

crusaderky:
Partially revert #3137.
This change did nothing to improve thread safety.
e.g. to get an item from a list you can use:
- `PyList_GET_ITEM` - thread unsafe vs. list element swap due to borrowed references; thread unsafe against list shrinking; no bounds checking regardless of threading
- `PyList_GetItem` - thread unsafe vs. list element swap; unsure about thread safety vs. list shrinking; bounds checked. Slower than `PyList_GET_ITEM`.
- `PyList_GetItemRef`. Thread safe and bounds checked; slower than both of the above.

To clarify there is absolutely nothing wrong with using `PyList_GET_ITEM` or `PyList_GetItem` in free-threading Python, as long as you can guaranteed that either
- the list is private to the thread; or
- your code is protected by a [critical section](https://docs.python.org/3/c-api/init.html#python-critical-section-api) and is not going to be suspended within it; or
- your code is protected by a lock

CC @klausspanderen @ngoldbaum

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -209,10 +209,9 @@ Examples/python/**/foo.py
 Examples/python/**/robin.py
 Examples/python/**/runme3.py
 Examples/python/**/spam.py
-Examples/python/doxygen/example.html
-Examples/python/external_runtime/swig_runtime.hxx
 Examples/python/import_packages/module_is_init/pkg1/__init__.py
 Examples/python/import_packages/namespace_pkg/path4.zip
+Examples/python/doxygen/example.html
 
 # R
 Examples/test-suite/r/*.R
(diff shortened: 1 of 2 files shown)


crusaderky:
Appveyor failure is unrelated to PR
> Build execution time has reached the maximum allowed time for your plan (60 minutes).

wsfulton:
I will merge this shortly. It's not really up to SWIG to remove/modify the available Python APIs for users. These Python APIs can quite legitimately be used safely in a multi-threaded environment with appropriate locking methods as suggested here. Additionally, not all code has to be thread-safe - it's up to users as to what code is called in a multi-threaded context and vast swathes of wrapped C/C++ APIs are neither thread-safe nor re-entrant either.

PR #3227, opened 2025-07-23
Remove SWIG_AsArgcArgv

crusaderky:
This is 20 years old code, untested, undocumented, and broken - just lint it and you'll realise that parenthesis opening and closing make no sense.

Rather than fix it, I would advise deleting it since nobody felt the need for it in 20 years.


crusaderky:
there was a test but it escaped my grepping. Apologies. Will work on a different PR to fix thread-safety.

PR #3230, opened 2025-07-29
Python: SWIG_AsArgcArgv thread safety

crusaderky:
- Make `SWIG_AsArgcArgv` thread-safe in free-threading Python.
- Fix memory leak in case of invalid inputs, with and without GIL. e.g. `mainv(["foo", 1])` raises TypeError; previously it would leak both a `char *[3]` and a `char *` for "foo".
- Improve test coverage.


DNM_build.sh
crusaderky, line 1: DNM

Lib/python/pyhead.swg
```diff
@@ -23,6 +23,14 @@
...
 #endif
 
+/* Compatibility macros for Python <3.13 */
+#if PY_VERSION_HEX < 0x030D00B3
+#  define Py_BEGIN_CRITICAL_SECTION(op) {
+#  define Py_END_CRITICAL_SECTION() }
+#  define Py_BEGIN_CRITICAL_SECTION2(a, b) {
+#  define Py_END_CRITICAL_SECTION2() }
+#endif
+
```
crusaderky, line 33: Copied from https://github.com/python/pythoncapi-compat/blob/b541b98df1e3e5aabb5def27422a75c876f5a88a/pythoncapi_compat.h#L1354

Lib/python/argcargv.i
```diff
@@ -2,53 +2,51 @@
...
+    return res;
+
   } else {
     /* seems dangerous, but the user asked for it... */
     size_t i = 0;
-    if (argv) { while (*argv[i] != 0) ++i;}
+    if (argv && *argv) { while ((*argv)[i] != 0) ++i; }
     if (argc) *argc = i;
     if (owner) *owner = 0;
     return SWIG_OK;
```
crusaderky, line 52: AFAICT these last two code paths are untested

crusaderky:
I've hit a block.
The threaded torture test demonstrates a segmentation fault.
I can successfully reproduce it on 3.13t but also on 3.13 and 3.12. In the NoGIL build it is just more frequent.

All generated files here: https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804

To my understanding:
1. swig_argcargv_runme.py [calls mainv in the generate python wrapper](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_runme-py-L91), passing as a parameter a list[str] that is being modified by other threads (enlarge, shrink, and dereference elements)
2.  the generated python wrapper [calls mainv in the C++ module](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest-py-L66)
3. on GIL-enabled interpreters, this acquires the GIL
4. run [_wrap_mainv](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3966)
5. _wrap_mainv [calls SWIG_AsArgcArgv](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3982) on the list object
6. on NoGIL interpreters, SWIG_asArgcArgV [acquires a critical section](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3665) on the list object. This to my understanding prevents the list from changing size or contents, and also prevents the referenced string objects from getting dereferenced. Note that this is true specifically because nothing blocks within the critical section.
7. either SWIG_AsCharPtrAndSize or SWIG_asArgcArgV [deep-copy the buffer](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3674-L3677) of each PyString to a new, thread-private `char*`. This to my understanding is thread-safe because PyStrings are immutable and the object as a whole can't be dereferenced, thanks to the critical section on the list that holds a reference to it.
8. On NoGIL interpreters, SWIG_AsArgcArgv [releases the critical section](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3692), thus allowing the list to mutate and the str arguments to get dereferenced
9. SWIG_AsArgcArgv returns
10. _wrap_mainv [calls mainv](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3995), passing the thread-local, deep-copied `char **argv` as a parameter
11. [mainv returns a reference](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3548) to either one of the elements of `char **argv` or to a constant
12. _wrap_mainv [calls SWIG_FromCharPtr](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3996) on the char* returned by mainv
13. SWIG_FromCharPtr [calls SWIG_FromCharPtrAndSize(cptr, strlen(cptr))](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3900-L3903)
14. SWIG_FromCharPtrAndSize calls [PyUnicode_FromStringAndSize](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3885), which deep-copies the returned `char*` to a new PyString object
15. _wrap_mainv [dereferences the temporary `char**`](https://gist.github.com/crusaderky/f7b90bb01bb2293b9c5654e62f811804#file-argcargvtest_wrap-cxx-L3999-L4002)
16. _wrap_mainv returns.
17. On GIL-enabled interpreters, this releases the GIL.

However, I'm getting a segfault on step 13, where the `char*` returned by `mainv` has already been released - but I can't figure out by what, since it's supposed to be thread-private.
The stack trace is identical in 3.13 and 3.13t:
```
#0  __strlen_avx2 () at ../sysdeps/x86_64/multiarch/strlen-avx2.S:76
#1  0x00007b82f6933a6f in SWIG_FromCharPtr(char const*) () from /home/crusaderky/github/swig/Examples/test-suite/python/_argcargvtest.so
#2  0x00007b82f6933ded in _wrap_mainv () from /home/crusaderky/github/swig/Examples/test-suite/python/_argcargvtest.so
#3  0x000063c5d9f15539 in cfunction_call (func=0x556a4e89330, args=<optimised out>, kwargs=<optimised out>)
    at /usr/local/src/conda/python-3.13.5/Objects/methodobject.c:551
```

Lib/python/argcargv.i
```diff
@@ -2,57 +2,55 @@
...
-    }
-  } else {
+  if (SWIG_IsOK(res)) {
     /* seems dangerous, but the user asked for it... */
     size_t i = 0;
-    if (argv) { while (*argv[i] != 0) ++i;}
+    if (argv && *argv) { while ((*argv)[i] != 0) ++i; }
     if (argc) *argc = i;
     if (owner) *owner = 0;
     return SWIG_OK;
```
crusaderky, line 16: AFAICT this code path is untested. Adding a test is out of scope for this PR.

Lib/python/argcargv.i
```diff
@@ -2,57 +2,55 @@
...
+  int is_list = PyList_Check(input);
+  if (!is_list && !PyTuple_Check(input)) return SWIG_TypeError;
+
+  size_t i;
+  res = SWIG_OK;
+  size_t size = is_list ? PyList_Size(input) : PyTuple_Size(input);
+  if (argv) *argv = %new_array(size + 1, char*);
+
+  for (i = 0; i < size; ++i) {
+    PyObject *obj = is_list ? PyList_GetItemRef(input, i) : PyTuple_GetItem(input, i);
```
crusaderky, line 33: An earlier version of this PR was using `Py_BEGIN_CRITICAL_SECTION` and `PyList_GET_ITEM`. The problem with critical sections is that many Python C API functions can release it; namely `PyUnicode_AsUTF8AndSize` which is called internally by `SWIG_AsCharPtrAndSize`.

Lib/python/argcargv.i
```diff
@@ -2,57 +2,55 @@
...
-	  char *cptr = 0; size_t sz = 0; int alloc = 0;
-	  res = SWIG_AsCharPtrAndSize(obj, &cptr, &sz, &alloc);
-	  if (SWIG_IsOK(res)) {
-	    if (cptr && sz) {
-	      (*argv)[i] = (alloc == SWIG_NEWOBJ) ? cptr : %new_copy_array(cptr, sz, char);
-	    } else {
-	      (*argv)[i] = 0;
-	    }
-	  } else {
-	    return SWIG_TypeError;
```
crusaderky, line 31: this was leaking `*argv` and all `(*argv)[j]` for j < i

Lib/python/argcargv.i
```diff
@@ -2,57 +2,55 @@
...
+
+  size_t i;
+  res = SWIG_OK;
+  size_t size = is_list ? PyList_Size(input) : PyTuple_Size(input);
+  if (argv) *argv = %new_array(size + 1, char*);
+
+  for (i = 0; i < size; ++i) {
+    PyObject *obj = is_list ? PyList_GetItemRef(input, i) : PyTuple_GetItem(input, i);
+    if (!obj) {
+      // List was shrunk by another thread. Return early with SWIG_OK.
```
crusaderky, line 35: This behaviour is coherent with shriking a list while an Iterator is running on it.
Raising `RuntimeError: list changed size during iteration` coherently with the behaviour of iterators running on python dicts was another legitimate option.

crusaderky:
@wsfulton @ngoldbaum this is ready for review

Lib/python/pyhead.swg
```diff
@@ -93,6 +93,17 @@ SWIG_Python_str_FromChar(const char *c)
 #define Py_hash_t long
 #endif
 
+// gh-114329 added PyList_GetItemRef() to Python 3.13.0a4
```
crusaderky, line 96: Copy-paste from https://github.com/python/pythoncapi-compat/blob/main/pythoncapi_compat.h

Lib/python/argcargv.i
ngoldbaum: I don't have a strong sense for whether it's more coherent to proceed as normal or alert the user they're doing something wrong. An opinion from a SWIG maintainer reviewer would be really helpful here.

Lib/python/argcargv.i
```diff
@@ -2,57 +2,55 @@
...
     if (owner) *owner = 0;
     return SWIG_OK;
   }
+  PyErr_Clear();
+  int is_list = PyList_Check(input);
+  if (!is_list && !PyTuple_Check(input)) return SWIG_TypeError;
+
+  size_t i;
+  res = SWIG_OK;
+  size_t size = is_list ? PyList_Size(input) : PyTuple_Size(input);
```
ngoldbaum, line 24: `PyList_Size` and `PyTyple_Size` return [`Py_ssize_t`](https://docs.python.org/3/c-api/intro.html#c.Py_ssize_t), which is almost always a typedef for `ssize_t` instead of `size_t`.

Also if you need to use the stable ABI I'd write this instead like:

```
Py_ssize_t size = PyList_Size(input);
if (size < 0) {
    size = PyTuple_Size(input);
    if (size < 0) {
        # not a list or tuple
        return SWIG_TypeError;
    }
}
```

Since both `PyList_Size` and `PyTuple_Size` do `PyList_Check` and `PyTuple_Check` internally this lets you skip unnecessary repeated `isinstance` checks.

Lib/python/argcargv.i
ngoldbaum: I'd add a `# thread safety:` comment block above this explaining why strong references are necessary for lists but not for tuples.

ngoldbaum (commented):
I saw a few minor places to refactor a little but other than that this looks great!

Lib/python/argcargv.i
crusaderky: I think it's missing a line?

```C
Py_ssize_t size = PyList_Size(input);
if (size < 0) {
    PyErr_Clear();  // added this
    size = PyTuple_Size(input);
    if (size < 0) {
        // not a list or tuple
        return SWIG_TypeError;
    }
}
```

I've tweaked it to incorporate your suggestion, thanks!

Lib/python/pyhead.swg
crusaderky: See https://github.com/swig/swig/pull/3235 for perplexities re. SWIG's limited API tests.

Lib/python/argcargv.i
wsfulton: Oh boy, that is really hard to say. Some C++ programmers would say it's quite acceptable for your computer to burst into flames should someone try and write to a container from multiple threads as this well and truly falls under 'undefined behaviour'. However, generally, SWIG library code tries not to get in the way of the user and I would suggest meeting some sort of Python expectations in this area and defined behaviour. Given the logic here can detect when the Python container has been shrunk by another thread, I would document this as unexpected and throw an appropriate exception. Is there some free-threading Python standard approach to this? Some sort of equivalent to Java's ConcurrentModificationException. If there isn't one in Python, perhaps there should be.

Lib/python/argcargv.i
```diff
@@ -2,57 +2,65 @@
...
     if (argc) *argc = i;
     if (owner) *owner = 0;
     return SWIG_OK;
   }
+  PyErr_Clear();
+  res = SWIG_OK;
+  
+  int is_list = 0;
+  size_t i;
+  Py_ssize_t size = PyTuple_Size(input);
```
wsfulton: Any reason why PyTuple_Check and PyList_Check are not used to determine which is the appropriate function to call for getting the size?

ngoldbaum: I suggested this to avoid unnecessary extra calls to `PyList_Check`, since `PyList_Size` has an inline `PyList_Check`: https://github.com/python/cpython/blob/c5cebe1b5afa58fc0ee95153a4b1905229dce7dc/Objects/listobject.c#L303-L312

but maybe that's too clever

wsfulton: It's just okay I suppose to utilise internal knowledge of the implementation if it is open source, won't change in the future and is consistent across all Python versions. That's a bit risky, but acceptable. However, passing in either a tuple or list is expected and in the case of the list, the Python code raises a Python exception. Certainly in C++, it is bad practice to throw an exception as normal, non-error code flow and it severely affects performance. I'm less certain of this in the Python space though so would like to be enlightened. The PyList_Check and PyTuple_Check functions can be used to avoid raising a Python error, so on the face of it they should be used.

ngoldbaum: I think I agree, just deal with the duplicated checks and do the idiomatic thing.

Lib/python/argcargv.i
crusaderky: > Is there some free-threading Python standard approach to this?

The "standard approach" in Python is that the interpreter must not burst into flames under any circumstances. Both raising and returning early are sensible outcomes here.

If you remove elements from a list while iterating on it, the iterator just returns early.
If you do the same to a dict, you get a RuntimeError.

Lib/python/argcargv.i
crusaderky: Reverted

crusaderky:
> We avoid assert in the test-suite as assert may be completely ignored by the Python interpreter. As you've used unittest, shouldn't these be assertTrue() or assertEquals() calls instead?

AFAICT in modern python (e.g. the last 10 years) unit tests always use plain `assert` for the sake of readability, as unit tests are never run with the `-O` flag anyway.  In pytest, which does not offer `assertTrue` or `assertEqual`, numpy has a function [numpy.testing.assert_](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_.html) which however AFAICT has fallen into disuse in favour of just using `assert`. I've reviewed the SWIG CI and I have not seen any use case of the `-O` flag in any of the test benches.

> We've never used unittest before in the SWIG test-suite, I see it prints a pile of unwanted output [...]
> The test-suite output is expected to be nothing except on error or if a warning is output by one of the tools invoked during the test. Can you configure unittest to suppress this output? If so, we should probably make better use of it going forwards.

No, I can at most remove the first line of dots by passing `--quiet` but I can't silence it completely.
Is this a dealbreaker? To me it feels that the benefit of using a test suite vastly outweighs having 5 less lines in the CI output.

> I'm not a fan at all of tests that are not reproducible such as using random inputs. Reproducing tests when c++ undefined behaviour is often at fault is challenging enough, so adding more undefined/random test behaviour should be avoided.

Welcome to the world of thread safety testing. I could change `test_threading_race_conditions` to pick from a predefined deterministic sequence of strings but it would change very little about its non-determinism.

> I'm not sure I see any value in using random strings in TestArgcArgv.test_bad_args_memory_leak to be honest, so can you instead use a predefined selection of strings of varying length. Include one at least greater than 15 in size in order to avoid the small string optimization.

Done.

> I'm also unsure of the purpose of running torture() in parallel, other than to try make it segfault

The _only_ purpose of running it in parallel is to make it segfault. In fact the new test vs. the old code was segfaulting a solid 80% of the times you ran it.

> which from the comment in that function indicates that it will segfault as I didn't spot a lock to prevent it

It can no longer segfault thanks to the replacement of PyList_GetItem with PyList_GetItemRef, plus testing if the function actually returns a value, as per free-threaded python best practices. No locks are needed; the new code just deals with all possible race conditions.

> ... and a test that has the possibility of segfaulting is not going to be accepted into master by me! Sorry if I've missed something obvious.

The test never segfaults because the code has been now made thread-safe.

> Given this is the first true multi-threaded test using the free-threaded interpreter and a new area for both Python and SWIG, we ought to discuss and document what sort of testing SWIG should do in this area. This could be a massive project making all the SWIG generated code behave in some sort of defined manner, so we need some expectations set. 

Speaking for Python exclusively here. 
From the [current documentation](https://github.com/swig/swig/blob/be943d11a11b4b901bd673f9c2b4dbc500d993f3/Doc/Manual/Python.html#L7659-L7661):

> SWIG does not add any thread-unsafe code generation, barring some corner cases where a warning is issued (warning 470 in the Warnings chapter). However, this does not automatically make the code thread-safe for using in a free threading Python interpreter as it requires the code being wrapped by SWIG to also be thread-safe.

The new test just verifies the statement that "SWIG does not add any thread-unsafe code generation" (which, for `SWIG_AsArgcArgv`, was in fact false before this PR).

> Probably all the tests in test-suite should have a way to run all the tests in a multi-threaded manner.

[pytest-run-parallel](https://github.com/Quansight-Labs/pytest-run-parallel) does that. However implementing it in SWIG would require rewriting all unit tests to use pytest to begin with, which would be a monumental task I don't have the capacity for. pytest-run-parallel would also _not_ detect the use case I fixed in this PR - sharing a local object between threads - as pytest-run-parallel paralellizes test functions, and most test functions create and destroy their own objects. Broadly speaking, pytest-run-parallel is good at detecting race conditions in global state access.

> Inspiration from other Python free-threading project is probably needed. 

Both myself and @ngoldbaum are deployed to delivering free-threading support across the whole Python ecosystem.
The general attack plan is described here: https://py-free-threading.github.io/
In short, test-wise, each project should
- Run the whole test suite with pytest-run-parallel
- Run the whole test suite with TSAN
- Manually inspect code for known offenders: [functions that return borrowed references](https://docs.python.org/3.14/howto/free-threading-extensions.html#borrowed-references) and global variables
- Write ad-hoc multi-threaded tests for known fragile patterns, e.g. mutating a shared state.

> have you got some bigger picture goal in mind here with the patches you are providing?

The goal is that SWIG must not introduce thread unsafety, as long as the wrapped C/C++ code is itself thread safe.

> Multi-threaded testing can get complicated and we need some guidelines as to what should and shouldn't be tested. 

There are levels:
1. thread safety is guaranteed as long as you never share objects across threads, e.g. the global state if any (e.g. global caches) is thread-safe
2. thread safety is guaranteed as long as shared objects are read-only across threads. e.g. local caches must become thread-safe.
3. thread safety is guaranteed even when mutating shared objects 

As a rule of thumb, the project I'm working on strives for level 2 across the whole Python ecosystem, with nuance.
This PR hits level 3, for the specific function `SWIG_AsArgcArgv`. I would however advise pragmatism. In general, if by just glancing at the code you can say it is going to segfault (like in this case) it may be a good idea to spend some time to make it robust. If code is probably OK as-is, it may be overkill to spend extensive amounts of time and energy to write torture tests for it.

> It has the potential to be constantly causing problems as well as being slow resulting in these multi-threaded just being disabled

There is a general aim with free-threading PRs to not introduce any material slowdowns or in general downsides for GIL-enabled interpreters.

>  if they are not reliable and stable given the lack of resources for general SWIG maintenance. There just isn't anyone sitting around the bandwidth to chase hard-to-fix threading issues in the tests. It's a big commitment providing thread-safety guarantees

I would treat thread safety like any other code robustness issue, to be prioritised depending on how many people are getting burned by bugs. There is a hope and expectation that the general public will substantially ramp up usage of free-threading towards the end of this year (with Python 3.14 coming out plus many fundamental packages gaining robust free-threading support as I write).

>  and someone needs to commit to this in the long term if done properly(which would be great).

As with all open sources projects, people will commit to it if they have a stake in it.

Examples/test-suite/python/argcargvtest_runme.py
```diff
@@ -1,54 +1,119 @@
...
+                if nargs == 0 or random.random() > 0.4:
+                    args.append(random_string())
+                else:
+                    idx = random.randint(0, nargs - 1)
+                    try:
+                        del args[idx]
+                    except IndexError:  # Race condition
+                        pass
+                # Race conditions will cause this to segfault unless SWIG_AsArgcArgv is
+                # protected by a lock.
```
wsfulton:
```suggestion
                # Race condition testing...
                # This used to cause SWIG_AsArgcArgv to segfault unless protected by a lock.
```
Comment clarification to no longer imply that a lock is still required.

wsfulton:
@crusaderky - appreciate for the enlightening comments and I'd like to thank your and @ngoldbaum's contributions to SWIG as well as making the wider Python ecosystem better in this somewhat challenging area.

Regarding the test-suite, it's very simple and designed for multiple target languages. It probably could be better, but this looks non-trivial using unittest given the verbiage can't be stopped. I looked at pytest too quickly and it seems deficient in this area as well. It' a bit surprising; SWIG uses tools from so many different languages and I can't remember coming across this lack of output control over the last 25 years or so except with Mono and abi3audit but they happily accepted patches to remedy it. Anyway, we can discuss the improvements to the test-suite in a separate ticket as I don't want this ticket to turn into a 'fix the python test-suite issue' before we accept these thread-safety fixes. So with that in mind, could you please adjust the tests to fit in with the way the rest of the test-suite works, which means silent by default as I see this as invaluable for spotting warnings and other non-python tool problems when running the entire test-suite. 

We tend to write a simple Python test specific `check` function equivalent to your assert statements. I would prefer to avoid assert; we get so many obscure bugs and the hundreds of combinations of running the test-suite are certainly not captured by CI and are sometimes just done by hand - we have 130 separate test-suite invocations and that is just a tiny fraction of what we could be doing. I'd like to still be able to usefully run:
```
make check-python-test-suite PYFLAGS=-O
```
Normally test `utils` are added to each test - that's an exaggeration as it's normally just an assert like helper function. Your test_utils.py file could expand into something useful, but can you rename `test_utils.py -> swig_test_utils.py` and suggest adding a `swig_assert` function and possibly prefixing all the functions in there with `swig_`

Your approach of adding in the odd torture test where it is known to be needed is nicely pragmatic.

crusaderky:
@wsfulton I've done the changes you requested

Lib/python/argcargv.i
wsfulton: I was just viewing @ngoldbaum's video on free-threaded python - https://www.youtube.com/watch?v=EuU3ksI1l04 and it discusses a `ConcurrentUseException`, but that looks to be project specific. Seems a shame there is no such thing planned for inclusion in Python.

wsfulton:
Thanks, the swig_test_utils.py additions should be useful going forwards.

PR #3233, opened 2025-07-31
Python: Thread-safe `SWIG_runtime_data_module`

crusaderky:
[EDIT] `SWIG_runtime_data_module` is thread-safe as the first call to the function holds the GIL even in free-threaded python. This PR uses the new API when available (this is just cosmetic) and adds a comment explaining it.

Original OP below:

~On free-threaded Python, prevent a situation where a thread calls `import module1` and another calls `import module2`, where `module1` and `module2` are two separate SWIG-generated Python modules, and the two modules end up with two completely separate copies of the `swig_runtime_data5` module, only one of which is referenced by `sys.modules`.~

~`SWIG_runtime_data_module` creates the empty module `swig_runtime_data5` when the first SWIG-compiled module is imported. To my understanding, `PyImport_AddModuleRef` only prevents a segfault due to a pointer to a dereferenced PyObject. It does not prevent from ending up with two separate modules, only one of which is referenced by `sys.modules`.~

~There is no protection against race conditions caused by `reimport`. Unless a user deliberately runs `del sys.modules["swig_runtime_data5"]`, reimporting the actual user's SWIG modules (`module1` / `module2` in the above example) will just return a reference to the same `swig_runtime_data5` that had been created previously.~

~No unit tests for this as it would be quite hard to concoct one.~

### DIFF
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -120,16 +120,10 @@ static PyObject *Swig_runtime_data_module_global = NULL;
 SWIGINTERN PyObject *
 SWIG_runtime_data_module() {
   if (!Swig_runtime_data_module_global) {
-#if PY_VERSION_HEX >= 0x030D00A0
-    /* On free-threaded Python, handle race condition with other SWIG-compiled
-       Python modules creating the same module at the same time.
-       Note that global variables are local to the current Python module,
-       so a mutex would not help!
-     */
-    PyImport_AddModule(SWIG_RUNTIME_MODULE);
-    /* Now we are guaranteed that a module is referenced by sys.modules,
-       but not necessarily the one returned by the previous function call.
-     */
+#if PY_VERSION_HEX >= 0x030D0000
+    /* free-threading note: the GIL is always enabled when this function is first called
+       by SWIG_init, so there's no risk of race conditions
+     */  
     Swig_runtime_data_module_global = PyImport_AddModuleRef(SWIG_RUNTIME_MODULE);
 #elif PY_VERSION_HEX >= 0x03000000
     Swig_runtime_data_module_global = PyImport_AddModule(SWIG_RUNTIME_MODULE);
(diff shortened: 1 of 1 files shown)


ngoldbaum:
> To my understanding, PyImport_AddModuleRef only prevents a segfault due to a pointer to a dereferenced PyObject. It does not prevent from ending up with two separate modules, only one of which is referenced by sys.modules.

Just so I understand what this does PR does: after this PR swig still might leak an empty module object, but no longer crash, which might happen without this PR. Is that right?

crusaderky:
> Just so I understand what this does PR does: after this PR swig still might leak an empty module object, but no longer crash, which might happen without this PR. Is that right?

There shouldn't be any leaks. However, I've been rubber-ducking this and I think there's a problem in my code. Given two threads A and B:

1. thread A imports moduleA, which calls [PyImport_AddModule](https://github.com/python/cpython/blob/08ce7c695899d2c62b2b32dbb733052e1ce9301c/Python/import.c#L390) for `swig_runtime_data5`
2. at the same time, thread B imports moduleB, which does the same. Note that it is not the same code (it's copy-pasted from the template), so no shared lock is possible.
3. in thread A `PyImport_AddModule` creates a new module and inserts it in sys.modules. This sets its refcount to 1.
4. in thread B, `PyImport_AddModule` does the same. The refcount of the module created by thread A goes down to 0 and gets released.
5. now one of the two threads starts [PyImport_AddModuleRef](https://github.com/python/cpython/blob/08ce7c695899d2c62b2b32dbb733052e1ce9301c/Python/import.c#L339). There are 2 possible use cases:
  a. thread A called setitem first and thread B called setitem second, or vice versa. `PyImport_AddModuleRef` returns the surviving module in both threads. :heavy_check_mark: 
  b. thread A called setitem, but thread B hasn't done it yet. thread A runs `PyImport_AddModuleRef`, which returns a valid reference. Shortly afterwards, thread B calls its own SetItem, thus expunging the reference from module A. Now in thread A we have a PyObject with a valid reference stored in SWIG's global variable for moduleA which is not the same as the one stored in sys.modules. :red_circle:

crusaderky:
After discussing this offline, this function is actually thread-safe.
This function is always first called by `SWIG_init`, which is an alias of `PyInit_<modulename>`, which acquires the GIL even in free-threaded builds.

I've updated the PR to just clean up the API and add a comment.

PR #3234, opened 2025-07-31
Add D reverse length string

erezgeva:
Fix use of 'dtype' attribute.
Add Length & string reverse order typemap.
Add testes for Length & string typemaps.


wsfulton:
Nice fix as this is fixing  a generic multi-argument typemap problem, thanks. Merged in f52c25b767d9e3075c968e6b353b9c19d275c0cb and changes entry added in b0e3faefd10dabbcfda2e789fc602c9e53956074.

Also see #3236 for a similar D typemap issue that needs addressing for the reverse length support.

Issue #3236, opened 2025-08-01
LENGTH STRING and STRING LENGTH typemap issues


wsfulton:
Following on from #3234, I've spotted another D issue with the reverse string length typemaps. When wrapping constructors, the 'din' typemap is not applied. Test and fix is:

```diff
diff --git a/Examples/test-suite/char_binary_rev_len.i b/Examples/test-suite/char_binary_rev_len.i
index 84c94b452..9831915ea 100644
--- a/Examples/test-suite/char_binary_rev_len.i
+++ b/Examples/test-suite/char_binary_rev_len.i
@@ -8,6 +8,7 @@
 #if !defined(SWIGC)
 /* There is no concept of a higher level binary string in C */
 %apply (size_t LENGTH, const char *STRING) { (size_t len, const char *str, ) }
+%apply (size_t LENGTH, const char *STRING) { (size_t leng, const char *stri, ) }
 #endif
 
 %inline %{
@@ -15,5 +16,7 @@ struct Test {
   size_t strlen(size_t len, const char *str) {
     return len;
   }
+  Test() {}
+  Test(size_t leng, const char *stri) {}
 };
 %}
diff --git a/Source/Modules/d.cxx b/Source/Modules/d.cxx
index 3a59775e4..21aa08b69 100644
--- a/Source/Modules/d.cxx
+++ b/Source/Modules/d.cxx
@@ -1191,7 +1191,7 @@ public:
 
       // Get the D code to convert the parameter value to the type used in the
       // intermediary D module.
-      if ((tm = lookupDTypemap(p, "din"))) {
+      if ((tm = lookupDTypemap(p, "din", true))) {
 	Replaceall(tm, "$dinput", arg);
 	String *pre = Getattr(p, "tmap:din:pre");
 	if (pre) {
```
Runtime tests are needed, and probably for all languages.

There is an additional problem here and that is the typecheck typemaps are missing for the LENGTH STRING and STRING LENGTH typemaps for all languages. The typecheck typemaps are needed to support wrapping overloaded methods.

@erezgeva would you be able to fill in these missing gaps?

erezgeva:
> untime tests are needed, and probably for all languages.
> 
> There is an additional problem here and that is the typecheck typemaps are missing for the LENGTH STRING and STRING LENGTH typemaps for all languages. The typecheck typemaps are needed to support wrapping overloaded methods.

Hi @wsfulton
Fixing `dytpe` make `din:nativepointer` obsolete.
[...]

PR #3238, opened 2025-08-01
Add typecheck to string length typemaps

erezgeva:
Following #3236
Add typecheck and test with 2 constructors in `char_binary_rev_len`

### DIFF
--- a/Examples/test-suite/char_binary_rev_len.i
+++ b/Examples/test-suite/char_binary_rev_len.i
@@ -8,6 +8,7 @@
 #if !defined(SWIGC)
 /* There is no concept of a higher level binary string in C */
 %apply (size_t LENGTH, const char *STRING) { (size_t len, const char *str, ) }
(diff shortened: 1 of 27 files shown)


wsfulton:
The only changes required are around the typecheck typemaps.

1. Use `SWIG_TYPECHECK_STRING` instead of the two typecheck types that were added as the inputs are ultimately marshalled as a target language string.
2. The typemap typemaps are needed for all languages and are missing for some.

erezgeva:
> The only changes required are around the typecheck typemaps.
> 
> 1. Use `SWIG_TYPECHECK_STRING` instead of the two typecheck types that were added as the inputs are ultimately marshalled as a target language string.
> 2. The typemap typemaps are needed for all languages and are missing for some.

Updated

wsfulton:
Thanks, merged, see de1ad94c78f5ab07039f89bef0b7649c2db814ba...29acbcebccda0dc26809cb481a8da8069ce09eb7

PR #3240, opened 2025-08-02
Python default builtin fix

jim-easterbrook:
Merged current master branch


wsfulton:
Hi @jim-easterbrook, you've raised this in error I believe. If so, please close.

jim-easterbrook:
> Hi @jim-easterbrook, you've raised this in error I believe. If so, please close.

Yes, I was attempting to check if this old fix still works with the current master. I should probably have tried merging the old fix into a branch of the current master, not _vice versa_.

PR #3241, opened 2025-08-04
Fix 2786 (Python single default parameters with builtin and fastunpack)

jim-easterbrook:
This is a simpler solution than #2790 (by @ojwb), but still passes all the tests I've tried on it.


jim-easterbrook:
I've redone this (with proper commit history) as #3243.

PR #3243, opened 2025-08-06
Fix 2786

jim-easterbrook:


jim-easterbrook:
This is the same as #3241, but preserves the commit history from @ojwb.

PR #3246, opened 2025-08-09
tcl: Fix memory leak in constant registration

QuantamHD:
When creating a new string object you need to increment and decrement the ref counter to make sure the object is correctly collected.


wsfulton:
Thanks

wsfulton:
@QuantamHD, I do apologize, but I lost the authorship of your commit as I was trying out a new approach to merging `gh pr merge` and it somewhat surprisingly dropped you as author. I have credited you in the CHANGES file though, see de1ad94c78f5ab07039f89bef0b7649c2db814ba.

PR #3250, opened 2025-08-20
python -builtin: remove __dict__ docstring

jim-easterbrook:
The docstring from the previously documented atribute or method was being applied to the builtin `__dict__` member. This now has no docstring, like `this` and `thisown`.

(This trivially small change doesn't merit an entry in the change log.)

### DIFF
--- a/Source/Modules/python.cxx
+++ b/Source/Modules/python.cxx
@@ -3377,8 +3377,7 @@ public:
       }
       Setattr(h, "getter", "SwigPyObject_get___dict__");
       if (!Getattr(h, "doc")) {
-	Setattr(n, "doc:high:name", Getattr(n, "name"));
-	Setattr(h, "doc", cdocstring(n, AUTODOC_VAR));
+	Setattr(h, "doc", "");
       }
     }
 


Source/Modules/python.cxx
```diff
@@ -3377,8 +3377,7 @@ class PYTHON:public Language {
       }
       Setattr(h, "getter", "SwigPyObject_get___dict__");
       if (!Getattr(h, "doc")) {
-	Setattr(n, "doc:high:name", Getattr(n, "name"));
-	Setattr(h, "doc", cdocstring(n, AUTODOC_VAR));
+	Setattr(h, "doc", "");
       }
```
ojwb, line 3381: It would probably make more sense to remove this `if` entirely instead if we can't generate a sensible default docstring (at least I'd generally expect a SWIG attribute to be handled the same when unset as when set to an empty string, but explicitly setting it adds some memory and CPU overhead).

jim-easterbrook: If unset it gets set (somewhere) to `class_name.__dict__`, which is not a helpful docstring. I prefer to make it blank.

wsfulton:
Old output using autodoc testcase: `help(autodoc.A)`:
```
 |  Data descriptors defined here:
 |  
 |  __dict__
 |      A : p.A
 |  
 |  this
 |  
 |  thisown
 |  
```
New output:
```
 |  Data descriptors defined here:
 |  
 |  __dict__
 |  
 |  this
 |  
 |  thisown
 |  
```
Non-builtin output is:
```
 |  Data descriptors defined here:
 |  
 |  __dict__
 |      dictionary for instance variables (if defined)
...
 |  
 |  thisown
 |      The membership flag  
```

While this patch is arguably better, the real fix is for the builtin doc output to match the non-builtin output. The description for from the non-builtin `__dict__` is correct given a pure Python instance variable can be created:
```python
>>> import autodoc
>>> a = autodoc.A(1, 2, 0)
>>> a.__dict__
{}
>>> a.madeup = 123
>>> a.__dict__
{'madeup': 123}
```
@jim-easterbrook would you be able to adjust your patch please?

jim-easterbrook:
OK, I'll sort it out tomorrow (UK time).

jim-easterbrook:
Hmm. The "dictionary for instance variables" message is generated by Python, and is version dependent. (See https://github.com/python/cpython/issues/112266)

Python3.11:
```
>>> class Foo(object):
...     pass
... 
>>> help(Foo)
Help on class Foo in module __main__:

class Foo(builtins.object)
 |  Data descriptors defined here:
 |  
 |  __dict__
 |      dictionary for instance variables
 |  
 |  __weakref__
 |      list of weak references to the object
```

Python3.6:
```
>>> class Foo(object):
...     pass
... 
>>> help(Foo)
Help on class Foo in module __main__:

class Foo(builtins.object)
 |  Data descriptors defined here:
 |  
 |  __dict__
 |      dictionary for instance variables (if defined)
 |  
 |  __weakref__
 |      list of weak references to the object (if defined)
```

I can simply put one of those texts into swig's `python.cxx`, but it would be better to find out what triggers Python to generate that message.

jim-easterbrook:
As far as I can tell, Python classes derived from `builtins.object` have a `__dict__` (with suitable docstring), even though `builtins.object` doesn't. Swig Python `-builtin` classes are also derived from `builtins.object` but only have a `__dict__` because it's provided by `SwigPyObject.swigdict`. Therefore we need to provide our own docstring.

PR #3254, opened 2025-09-01
Do not mark modules no-gil compatible by default

AnnaAraslanova:
Do not mark modules no-gil compatible by default, but only if SWIGPYTHON_NOGIL is initialized

This must be linked to https://github.com/swig/swig/pull/3215


wsfulton:
Thanks, committed in 70e284938eb4ab5d0d2bc57a0c340af7746847a3.

PR #3259, opened 2025-09-05
Drop Python 2.x support

### LINKED ISSUE #2424: don't automatically rename print. print is not a Python keyword
SWIG automatically renames C++ names that are Python keywords. However, `print` is not a keyword and thus should not _automatically_ be renamed. Such renaming may in many cases be a good idea, but that decision should be left up to developers.

Here are all the keywords in Python 3.10:
```Python
>>> import keyword
>>> keyword.kwlist
['False', 'None', 'True', 'and', 'as', 'assert', 'async', 'await', 'break', 'class', 'continue', 'def', 'del', 'elif', 'else', 'except', 'finally', 'for', 'from', 'global', 'if', 'import', 'in', 'is', 'lambda', 'nonlocal', 'not', 'or', 'pass', 'raise', 'return', 'try', 'while', 'with', 'yield']
>>> print in keyword.kwlist
False
```
Note, `print` is _not_ a keyword and therefor it _can_ be used as a name in Python code. For instance the following is valid Python:
```Python
print = 1
```
The following is also (of course) valid:
```Python
import sys

class foo:
    def print(self):
        sys.stderr.write('foo\n')

f = foo()
f.print()
```

For an example of the problem with the current SWIG implementation, when wrapping the following C++ class:
```C++
class foo
{
public:
    void print();
};
```
SWIG generates the following warning:
```
Warning 314: 'print' is a python keyword, renaming to '_print'
```
One calls wrapped method like `foo._print()`. The rename is unnecessary as `print` is no a keyword.  `print` should not be automatically renamed because it is not a keyword in Python.

### LINKED ISSUE #3201: Drop support for python 2.7
Is it time to drop support for python-2.7?

python-2.7 went end of life 6 years ago, the same year python-3.5 went end of life, see https://devguide.python.org/versions/. #3184 will shortly require 3.5 as SWIG's minimum supported python 3 version.

Dropping support will help with a lot of code cleanup in the Python module. With the switch to heap heap types, see #3196, nearly complete, we could remove all the code for static types as python-2.7 does not support heap types. As well as a bunch of other python-2.7 specific code.

jschueller:
Closes #3201
Closes #2424


Doc/Manual/Python.html
```diff
@@ -173,7 +172,7 @@ <H2><a name="Python_nn2">33.1 Overview</a></H2>
 
 <p>
 SWIG is compatible
-with all recent Python versions (Python 2.7 and Python &gt;= 3.5).
+with all recent Python versions (Python &gt;= 3.5).
 SWIG-4.0 supported Python 3.2.  SWIG-3.0 supported older Python 2.x and 3.x.
```
ojwb, line 176: Add "SWIG-4.3 supported Python 2.7."

Doc/Manual/Varargs.html
```diff
@@ -538,26 +537,10 @@ <H2><a name="Varargs_nn6">17.5 Varargs and typemaps</a></H2>
...
 }
-
-%typemap(freearg) (...) {
-%#if PY_VERSION_HEX&gt;=0x03000000
-  int i;
-  for (i = 0; i &lt; 10; i++) {
-    free(vargs$argnum[i]);
-  }
-%#endif
-}
```
ojwb, line 560: This should be kept and only the conditional removed.

Lib/python/pyabc.i
```diff
@@ -1,14 +1,12 @@
 %define %pythonabc(Type, Abc)
   %feature("python:abc", Abc) Type;
 %enddef
-%pythoncode %{if _swig_python_version_info[0:2] >= (3, 3):
+%pythoncode %{
     import collections.abc
-else:
```
ojwb, line 6: Should the `import` be right after the `%{` on the same line, like the existing code is?  I don't remember the details of how indentation inside `%pythoncode` is handled, but keeping it like the existing code should definitely avoid misindentation in the generation Python code.

Lib/python/pyhead.swg
```diff
@@ -1,12 +1,9 @@
-#if PY_VERSION_HEX >= 0x03030000 && !defined(SWIG_NO_HEAPTYPES)
+#if !defined(SWIG_NO_HEAPTYPES)
 #if !defined(SWIG_HEAPTYPES)
 #define SWIG_HEAPTYPES
 #endif
 #endif
 
-/* Compatibility macros for Python 3 */
-#if PY_VERSION_HEX >= 0x03000000
-
```
ojwb, line 9: If we're keeping these macros (which probably makes sense as user typemaps may be relying on them) then we should keep the comment saying what they're for.

ojwb: And also note that these are no longer used by SWIG itself but kept for compatibility with user typemaps.

Lib/python/pyrun.swg
```diff
@@ -952,9 +911,6 @@ SwigPyObject_TypeOnce(void) {
     (binaryfunc)0, /*nb_subtract*/
     (binaryfunc)0, /*nb_multiply*/
     /* nb_divide removed in Python 3 */
-#if PY_VERSION_HEX < 0x03000000
```
ojwb, line 955: It doesn't seem useful to leave this comment in place.

Lib/python/std_map.i
```diff
@@ -85,10 +85,8 @@
 	SWIG_PYTHON_THREAD_BEGIN_BLOCK;
 	if (PyDict_Check(obj)) {
 	  SwigVar_PyObject items = PyObject_CallMethod(obj,(char *)"items",NULL);
-%#if PY_VERSION_HEX >= 0x03000000
           /* In Python 3.x the ".items()" method returns a dict_items object */
           items = PySequence_Fast(items, ".items() didn't return a sequence!");
-%#endif
```
ojwb, line 91: The code which was inside the removed conditional has an extra indent, which no longer makes sense.

Lib/python/std_multimap.i
```diff
@@ -13,10 +13,8 @@
 	int res = SWIG_ERROR;
 	if (PyDict_Check(obj)) {
 	  SwigVar_PyObject items = PyObject_CallMethod(obj,(char *)"items",NULL);
-%#if PY_VERSION_HEX >= 0x03000000
           /* In Python 3.x the ".items()" method returns a dict_items object */
           items = PySequence_Fast(items, ".items() didn't return a sequence!");
-%#endif
```
ojwb, line 19: The code which was inside the removed conditional has an extra indent, which no longer makes sense.

Lib/python/std_unordered_map.i
```diff
@@ -71,10 +71,8 @@
 	SWIG_PYTHON_THREAD_BEGIN_BLOCK;
 	if (PyDict_Check(obj)) {
 	  SwigVar_PyObject items = PyObject_CallMethod(obj,(char *)"items",NULL);
-%#if PY_VERSION_HEX >= 0x03000000
           /* In Python 3.x the ".items()" method returns a dict_items object */
           items = PySequence_Fast(items, ".items() didn't return a sequence!");
-%#endif
```
ojwb, line 77: The code which was inside the removed conditional has an extra indent, which no longer makes sense.

Lib/python/std_unordered_multimap.i
```diff
@@ -20,10 +20,8 @@
 	int res = SWIG_ERROR;
 	if (PyDict_Check(obj)) {
 	  SwigVar_PyObject items = PyObject_CallMethod(obj,(char *)"items",NULL);
-%#if PY_VERSION_HEX >= 0x03000000
           /* In Python 3.x the ".items()" method returns a dict_items object */
           items = PySequence_Fast(items, ".items() didn't return a sequence!");
-%#endif
```
ojwb, line 26: The code which was inside the removed conditional has an extra indent, which no longer makes sense.

appveyor.yml
```diff
@@ -117,7 +117,6 @@ install:
 - if not "%WITHLANG%"=="" set SWIGWITHLANG==%WITHLANG%
 - if not "%WITHLANG%"=="" where %WITHLANG%
 - if "%SWIGLANG%"=="python" set PY3=3
-- if "%PY2%"=="2" set PY3=
```
ojwb: It seems silly to keep `PY3` which is now always set to `3` for Python.  It looks like the root of this is that `configure` takes `--with-python3` but the make rules are e.g. `check-python-version`.  The GHA config also has to deal with this.

Maybe `--with-python` should be an alias for `--with-python3` now.  This could reasonably be done as a follow-on clean-up though, as this change is already rather substantial.

ojwb: Looking into this, I noticed that `Tools/CI-linux-install.sh` checks `PY2` but isn't updated by this patch.

ojwb: Nor is `configure.ac`.

jschueller: I tried to update it for with-python to be an alias to with-python3, but its not fully cleaned up

jschueller:
hi, do you have any more remarks about the changes ?

Tools/CI-linux-install.sh
```diff
@@ -155,11 +155,7 @@ case "$SWIGLANG" in
...
 			pip install --user 'abi3audit>=0.0.12'
 		fi
-		if [[ "$PY2" ]]; then
-			WITHLANG=$SWIGLANG
-		else
-			WITHLANG=${SWIGLANG}3
-		fi
+		WITHLANG=${SWIGLANG}3
 		if [[ "$VER" ]]; then
 			if [[ -z "$PY2" ]] && [[ $VER =~ ^[0-9.]+$ ]]; then
```
erezgeva: Remove here `[[ -z "$PY2" ]] &&`

erezgeva (changes requested):
gcc 13 with Ubuntu 22 is broken.

I will PR to update to Ubuntu 24.

Erez

jschueller:
hello, are there still plans to integrate this for the next release ?

erezgeva:
> hello, are there still plans to integrate this for the next release ?

As #3201 was open by @wsfulton 
There is a good chance he would merge this one.
I simply think William is a bit busy.

jschueller:
yes, makes sense

erezgeva:
### I found `SWIG_PYTHON_2_UNICODE`.
```
$ git grep SWIG_PYTHON_2_UNICODE
Examples/test-suite/python/unicode_strings_runme.py:# Testing SWIG_PYTHON_2_UNICODE flag which allows unicode strings to be passed to C
Examples/test-suite/unicode_strings.i:#define SWIG_PYTHON_2_UNICODE
```
<br>

### More stuff with python2

```
$ git grep -i  python2
Doc/Manual/Python.html:        -DHAVE_CONFIG_H -I/usr/include/python2.7 \
Doc/Manual/Python.html:        -I/usr/lib/python2.7/config-x86_64-linux-gnu \
Doc/Manual/Python.html:        -I/usr/lib/python2.7/config \
Doc/Manual/Python.html:        -L/usr/lib/python2.7/config -lpython2.7 -lm -ldl \
Doc/Manual/Python.html:$ g++ -O2 -fPIC -c example_wrap.cxx -I/usr/local/include/python2.5
Doc/Manual/Python.html:$ CC -c example_wrap.cxx -I/usr/local/include/python2.5
Doc/Manual/Python.html:was built with (the python2.4 and python2.5 distributed by python.org
Doc/Manual/Python.html:As of python2.5, the distutils support building extensions with MingGW out
Doc/Manual/Python.html:For example "python27.lib".  Also, set the name of the output file to
Examples/python/import_packages/from_init1/README:For example with python2.x the py2/pkg2/__init__.py imports Pkg2_Bar class
configure.ac:    AC_CHECK_PROGS(PYTHON, [python python2.7])
```
<br>

###  `configure.ac`
The section `Look for Python` is for Python 2.  
It need to be removed completely til the `Look for Python 3.x`.  
 **Yes the entire block from line 1873 to line 2019!**
and any reference using these 4 variables
```
AC_SUBST(PYINCLUDE)
AC_SUBST(PYLIB)
AC_SUBST(PYLINK)
AC_SUBST(PYTHONDYNAMICLINKING)
```
The `AC_SUBST` means thay can be used in Makefiles.
`git grep` them to verify a proper clean.
For example to grep `PYINCLUDE` use `$git grep '\<PYINCLUDE\>'`

erezgeva:
[Remove-python2-from-configure.patch](https://github.com/user-attachments/files/26913551/Remove-python2-from-configure.patch)

jschueller:
oh, thanks! applied

erezgeva (approved):
The changes I request are done.

wsfulton:
Merged in 3e62339f7...79f7a2b7c. Thanks @jschueller, that was quite a lot of mopping up work you did and it's good to finally see the end of python 2. I added a tidy up commit and also removed the legacy python 2 macros.

jschueller:
awesome!

PR #3268, opened 2025-09-19
update Python example with setuptools + add -M option note

phetdam:
This PR addresses the discussion in #2586 on how `Python.html` is still using distutils for the Python extension module example instead of using [setuptools](https://setuptools.pypa.io/en/latest/), its replacement and *de facto* standard build backend for any Python package that needs to build extension modules. An additional section in `SWIG.html` is also added to document the use of the `-M` family of options, which instructs SWIG to generate a Make-style rule for the C/C++ source.

The `-M` options are not well documented but are invaluable for build system integration to ensure correct dependency tracking by the build system and ensuring optimal incremental rebuild, avoiding both the case where one has to re-run SWIG every time a build is triggered, and the case where only the `.i` file is tracked (e.g. in a naive Make rule), so the wrapper is out-of-date if a processed header of `.i` file required by the extension has been modified.


wsfulton:
Thanks, merged to master in de640927515eba2c9944b8e516455c916a82e261...3acc4ec53cd40d1fafca13fd66548896d0d4572e

phetdam:
Great, thank you!

PR #3272, opened 2025-10-09
Using cygpath

erezgeva:
Although 'cygpath' comes from CygWin,
it can be found on other Windows "bash" based. Use 'Tools/convertpath' only if we do not have 'cygpath'.

Remove `SWIG_LIB_WIN_UNIX`.
Windows does not need a separate environment.
Windows users can set `SWIG_LIB` like any other platform.

Remove `SWIG_LIB_SET`.
We have a separate test for native Windows library setting. No need to complicate the test suit.


configure.ac
```diff
@@ -2790,30 +2783,9 @@ SWIG_LIB_INSTALL=${swig_lib}
...
-esac
-AC_DEFINE_UNQUOTED(SWIG_LIB_WIN_UNIX, ["$SWIG_LIB_WIN_UNIX"], [Directory for SWIG system-independent libraries (Unix install on native Windows)])
-
 SWIG_LIB_PREINST=$ABS_SRCDIR/Lib
 AC_SUBST(SWIG_LIB_PREINST)
 
-dnl For testing purposes, clear SWIG_LIB when building SWIG in the source
-dnl directory under Windows because it is supposed to work without SWIG_LIB
-dnl being set. Otherwise it always needs to be set.
-SWIG_LIB_SET="env SWIG_LIB=\$(SWIG_LIB_DIR)"
```
wsfulton, line 2817: This is an important check and I don't see how it has been replaced with an equivalent.

wsfulton (changes requested):
I understand the motivation for detecting cygpath on mingw and msys on modern versions; cygpath used to only be available on cygwin. But I don't understand why main.cxx has changed nor why SWIG_LIB has been tampered with.

configure.ac
erezgeva: The `Test swiglib` is checking that.
The test verify that the Windows SWIG library path depends on the location of the SWIG executable in case the `SWIG_LIB` is not defined.

Why changing all of the testing that focus on actual testing, on a small testing that only windows need.
That seems a big overhead.

erezgeva:
> I understand the motivation for detecting cygpath on mingw and msys on modern versions; cygpath used to only be available on cygwin. But I don't understand why main.cxx has changed nor why SWIG_LIB has been tampered with.

`SWIG_LIB` was not tamper, but was equal to other platforms.
So all platform uses `SWIG_LIB` in the same manner.
The only different is that with Linux, MacOS, Unix and other POSIX platforms, the instillation directory is set before the build, while Windows is using its own `swig.exe` application location as base for the libraries path.
There is no real reason for any other different.
In configure the only change should be handle with the `cygpath` application.


So yes, we need to verify with a test that when  `SWIG_LIB` is not set, the windows application library path is based on swig application, so we make a specific test `Test swiglib` for that.
We can improve the test by copying the swig application to another folder and check the `./swig.exe -swiglib` change with it.
My point is that every test need to be oriented, that is why we have 127 tests on Linux and 680 tests in the test-suite, so every test tell us where the problem is.
Testing the windows default swig application library path with all the examples and test-suite testing is a huge overkill.

As for the use of `SWIG_LIB_WIN_UNIX`, I do not understand why do we need 2 libraries location for the swig windows application.
Can you prove the need? Do you have a test that prove its needed?

configure.ac
erezgeva: I moved it to another future PR.

I focus on the use of `cygpath`.

Too much for one PR :smile:

erezgeva:
I do think the removing  `SWIG_LIB_WIN_UNIX` and use of `SWIG_LIB_SET` in the same manner all across platforms is a good change, but seems require a separate PR.

The handle of `cygpath` is big enough for one PR.

erezgeva:
Restoring `SWIG_LIB_WIN_UNIX` cause `./swig.exe -swiglib` to print 2 paths.
The first is based on application location, the second is the value of `SWIG_LIB_WIN_UNIX`.

I update the `Test swiglib` test.

wsfulton:
The changes to convertpath for `-m` are regressive. Previously, the invocation of `cmd` would really convert to mixed path, for example:
```bash
# bash shell
$ cmd //c echo "/c/Users"

C:/Users
```
now it doesn't convert to mixed paths.

When convertpath was added, MinGW did not have any decent equivalent to cygpath and it was a necessary minimal workaround. That was a long time ago! I think it reasonable to require cygpath on one of msys/cygwin/mingw now. So, let's just delete convertpath and rely on cygpath being available. Set CYGPATH to false or something like that if it is missing, so that an error occurs at testing time. An error at configure time should not occur as cygpath should not be needed to just build the SWIG executable and we only error at configure time if there is a build dependency missing. I think this should be the final change for this pull request.

Agreed it might be overkill to set SWIG_LIB at every invocation of SWIG for all tests if it was just a sanity test for Windows. However, setting SWIG_LIB should remain as it provides a very helpful testing environment in the event that SWIG_LIB is already set on a computer. It makes sure testing is not affected by another SWIG installation.

I'm a little hazy on why we added two default paths being set to the Lib directory on Unix like environments on Windows. It's got to do with invoking the SWIG executable in a native Windows shell and in a Unix (such as MinGW bash) shell. It needs to work on both.

erezgeva:
> now it doesn't convert to mixed paths.

The mapping of drivers depend on simulation environment in which the bash runs in.
`MSYS2` does map `C:\` to `/c/`, but there are also other mapping.
We can not be sure what is the proper conversion in the simulation environment.
Most simulation environment adopted `cygpath` for a good reason :smile: 

There is another Windows mapping,
https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#dos-device-paths
Using path like `\\.\C:\`, but they are not POSIX compatible either.

<br>  

> Set CYGPATH to false or something like that if it is missing.

OK, on it.
`false` it is :+1: 
I will check if we need the `Tools/convertpath` on C# on non-windows platforms.

<br>  

> However, setting SWIG_LIB should remain as it provides a very helpful testing environment in the event that SWIG_LIB is already set on a computer

I aim in always set the `SWIG_LIB` during the tests.
Regardless whether `SWIG_LIB` is set in the system, or whether we use `WIN32`. I think we should always set the`SWIG_LIB` during the tests.
Anyhow, we can continue this conversation on my next PR regarding `SWIG_LIB`.

This PR focus on `cygpath`.

PR #3280, opened 2025-11-03
SwigPyIterator: Add override keyword

jschueller:


ojwb:
The generated code still needs to compile as C++98 - we could define a macro such as `SWIG_OVERRIDE` which expands to nothing for C++98 and `override` for newer standards versions if there are real benefits to being able to use `override`.

jschueller:
yes a macro would work, I wrote this to avoid warnings (-Wsuggest-override enabled)

PR #3287, opened 2025-11-11
Python: Fix warnings about implicit type conversions

jaw:
When compiling with -Wconversion -Wsign-conversion on gcc, these places resulted in warnings.

Add explicit type casts to be clear to the compiler.

### DIFF
--- a/Tools/testflags.py
+++ b/Tools/testflags.py
@@ -4,6 +4,10 @@ def get_cflags(language, std, compiler):
     if std == None or len(std) == 0:
         std = "gnu89"
     c_common = "-fdiagnostics-show-option -std=" + std + " -Wno-long-long -Wreturn-type -Wmissing-field-initializers -Wunused-variable"
+
+    # signed to unsigned conversions warnings, type conversion
+    sign_conversion_flags = " -Wconversion -Wsign-conversion"
+
     if std == "gnu89" or std == "gnu90":
         # gnu89 standard allows declaration after headers
         # use c99 or gnu99 if feature is necessary for using target language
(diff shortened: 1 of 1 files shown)


wsfulton:
To prevent this warning from returning into the code base could you add the warning to Tools/testflags.py, preferably for all languages if you don't mind probably fixing a few more.

jaw:
@wsfulton I've fixed a few more type conversion warnings for Python but I don't understand where these latest warnings come from (see the most recent CI run).

I tried applying the warnings to everything, but it seems you get warnings in thirdparty libs...
So, to limit the scope I've only enabled it for the Python build.

I thought this would be possible to finish fairly quickly but it's bigger than anticipated. Sadly I don't have more time to put in.

I see 3 options for this PR:
1. Remove the warning from testflags for now and merge what's already in here (it does solve a few compilation warnings including the ones I experienced in the project I'm working on).
2. Hand this PR over to someone with more experience with the project to fix the remaining warnings.
3. Close/cancel the PR, maintain local patches.

You decide what is best :)

wsfulton:
Looks like just a couple of simple fixes for li_cstring and li_cwstring testcases are required in the latest run. Surely that's doable to get a clean run for python?

Lib/python/pystrings.swg
```diff
@@ -32,7 +32,7 @@ SWIG_AsCharPtrAndSize(PyObject *obj, char **cptr, size_t *psize, int *alloc)
     /* The returned string is only duplicated if the char * returned is not owned and memory managed by obj */
     if (bytes && cptr) {
       if (alloc) {
-        cstr = %new_copy_array(cstr, len + 1, char);
+        cstr = %new_copy_array(cstr, (size_t)len + 1, char);
```
wsfulton:
```suggestion
        cstr = %new_copy_array(cstr, (size_t)(len + 1), char);
```
For consistency with other changes.

jaw: Done I think!

wsfulton (approved):
Thanks

PR #3288, opened 2025-11-18
cmake - remove unused policy

Andrej730:
`CMP0074` was introduced in cmake 3.12 and since min cmake version is now 3.13, then setting this policy have no use anymore and can be safely removed.

https://github.com/swig/swig/blob/f38f249c9ca397a548b31440e540bf6b0d953abf/CMakeLists.txt#L1

Ref https://cmake.org/cmake/help/latest/policy/CMP0074.html


jschueller (approved):

jschueller:
good to go @wsfulton

PR #3291, opened 2025-11-18
Go: Add `no-unique-id` flag

ysmilda:
This pull request adds the `no-unique-id` flag to the swig/go cli. This disables the generation of the unique id that is added to the generated names.

This allows for much smaller diffs when committing the files to version control.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -132,6 +132,10 @@ jobs:
           VER: '3.13'
           CSTD: gnu99
         - SWIGLANG: python
+          VER: '3.13'
+          PY_ABI_VER: '3.9'
+          CSTD: gnu99
+        - SWIGLANG: python
           VER: '3.13-dbg'
           CSTD: gnu99
         - SWIGLANG: python
--- a/CHANGES
+++ b/CHANGES
@@ -5,6 +5,19 @@ See the RELEASENOTES file for a summary of changes in each release.
 Issue # numbers mentioned below can be found on Github. For more details, add
 the issue number to the end of the URL: https://github.com/swig/swig/issues/
 
+Version 4.4.1 (7 Dec 2025)
+==========================
+
+2025-12-03: jschueller
+            [Python] #3285 Fix builds with python>=3.13 and Py_LIMITED_API<3.13.
+
+2025-11-14: jschueller
+            [Python] #3280 Add override specifier to SwigPyIterator methods for c++11
+            and later.
+
+2025-11-06: degasus
+	    #3284 Fix -fsanitize=undefined problem in the SWIG runtime code.
+
 Version 4.4.0 (20 Oct 2025)
 ===========================
 
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -12,6 +12,12 @@ Version 4.5.0 (in progress)
                - Add `-unique-id` and `-no-unique-id` flags to enable/disable the addition of a unique id to the generated function names.
                - Deprecate `-cgo` flag as this doesn't do anything.
                - Cleanup docs and help output.
+               
+2025-12-12: christophe-calmejane
+            [Lua] #3295 Add director support for Lua.
+
+2025-12-05: jaw
+            [Python] #3287 Fixes for gcc warnings: -Wconversion -Wsign-conversion.
 
 2025-11-08: wsfulton
             Fix handling of typedef to void in class methods/constructor's parameter
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,10 +4,6 @@ set (CMAKE_BUILD_TYPE Release CACHE STRING "Build type")
 
 project (swig)
 
-if (POLICY CMP0074)
-  cmake_policy (SET CMP0074 NEW)
-endif()
-
 file (STRINGS configure.ac line LIMIT_COUNT 1 REGEX "AC_INIT\\(.*\\)" )
 if (line MATCHES "AC_INIT\\(\\[(.*)\\],[ \t]*\\[(.*)\\],[ \t]*\\[(.*)\\]\\)" )
   set (SWIG_VERSION ${CMAKE_MATCH_2})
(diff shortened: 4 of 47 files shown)


ojwb:
@ianlancetaylor What are your thoughts on this?

ianlancetaylor:
I have no objection to adding an option like this. My personal bias would be to also have a `-unique-id` option, which would be the default. And, of course, the option needs documentation in Doc/Manual/Go.html which explains the risks of using it, namely that you can have name collisions if a program has multiple SWIG inputs.

ysmilda:
Would the addition of the `-unique-id` be purely decorative? As it is the default behaviour and setting it doesn't change any behaviour.

As for the documentation and the mentioned risks, would adding the module name mitigate the naming collisions? Or do multiple SWIG files not necessarily mean multiple modules.

ianlancetaylor:
Adding `-unique-id` is not purely decorative, as it can be used to override a `-no-unique-id` earlier on the command line. This kind of thing is a trivial point until you are working with an ugly Makefile that uses `SWIGFLAGS` along with its own flags.

Adding the module name is a good idea. It's not perfect as nothing ensures that module names are unique. Note that the module name is already part of the unique string. It would make sense for `-no-unique-id` to just drop the hash code, not the module name.

As background, I added the unique ID based on experience at Google where a single program could contain hundreds of different SWIG wrappers written by different people.

ysmilda:
I see, hadn't thought of that use case. Added the `-unique-id` flag and changed it so that the module name is left in when the `-no-unique-id` flag is set. Also added some documentation, not sure if this is enough/up to standard.

ianlancetaylor:
Thanks, this looks fine to me. @ojwb It's fine with me to merge this if it looks OK to you.

ojwb:
@wsfulton has previously stripped out options which are on by default as "redundant" or "pointless" (e.g. a quick search found 398ac5f01c6d99c7167c7fafb67d7ea695e08e47 and 67f5ade7ad03f15a2811190df899196dbf5a12fe) so I'm hesitant to merge a patch which adds such an option without his input.

Personally I tend to agree with Ian's point that it's useful to be able to override an option earlier on the command line (and it also makes for a smoother user experience if we want to flip the default at some point because they can specify an option to get the old default behaviour with both old and new versions), but consistency is also important and I don't really care enough to argue about it.  OTOH we also have "work more like a compiler" as a general guiding principle (also a form of consistency), and GCC's boolean options standardly have a way to set the default.

Also with regard to consistency, SWIG option naming would seem to favour `-nouniqueid` over `-no-unique-id`.  Again, not my personal preference as the latter seems significantly easier to read (what does "noun i que id" mean?); also `-no-unique-id` is more like GCC's style of option naming.  SWIG is not currently very consistent on this point though.

> [...] multiple SWIG inputs [...]

What's the situation where we can get a collision exactly?

I'm guessing you have the case of loading multiple SWIG-generated bindings for different modules in mind and are worrying about collisions between the interfaces from Go to the SWIG-generated C/C++ code?  However those symbols are all prefixed by `_` and the module name, so should only collide if the module name is the same (in which case can you load both in Go anyway?), or if there's an ambiguous split (e.g. `_foo_bar_baz` could be symbol `bar_baz` in module `foo`, or symbol `baz` in module `foo_bar`).

I think we currently just don't worry about potential collisions from such ambiguous splits in other target languages and nobody has complained that I can recall so it seems not to be a problem in real-world use, but perhaps it would be better addressed by a bijective mapping from (module, symbol) to name.  For example, we could encode the length of the module name to give e.g. `SWIG3_foo_bar_baz` vs `SWIG7_foo_bar_baz` to avoid potential collision in the case above.  The key advantage that has over the current hash approach is that the symbols don't all arbitrarily change for any change to the SWIG interface file so diffing SWIG/Go generated code becomes much less painful (and it could probably be implemented for most other target languages too).

ianlancetaylor:
Although I don't remember the details it was an actual problem at Google, and I added the unique ID to fix the problem. I would guess that two different .swig files used the same module name. At Google scale that is nearly impossible to avoid. The actual breakage could occur any time both .swig files were included in the same program, which could be quite a long time after the .swig files were created.

ojwb:
OK, thanks.  I guess you can include two different Go modules with the same name in the same program if you include them from different Go source files, or something like that.

wsfulton:
Regarding `-nouniqueid` over `-no-unique-id`, the former is what is generally the SWIG convention (I also don't like it btw). But I notice that a `-no-cgo` has crept into the Go module, so I suppose we can go with `-no-unique-id` and `-unique-id`.

I wasn't aware of the complex Makefile use case for having both an on and off flag. I've certainly seen complex Makefiles, and there are ways to deal with this. One issue of having both is it requires documenting which option is a default and which isn't. That is currently missing in the patch, so please add a `(default)` to the end of the `-unique-id` help output.

The options should also be in alphabetical order in the output of -help and in the added docs. Unfortunately this means that the on and off versions of the flag are not necessarily adjacent to each other.

ysmilda:
I've sorted the list. As for the `(default)` tag, should this also be added to the `-cgo` flag? And from the documentation the flags `-no-cgo`, `-use-shlib`, and `-soname` are all obsolete, shouldn't this also be mentioned in the CLI help?

wsfulton:
> I've sorted the list. As for the `(default)` tag, should this also be added to the `-cgo` flag? And from the documentation the flags `-no-cgo`, `-use-shlib`, and `-soname` are all obsolete, shouldn't this also be mentioned in the CLI help?

Options which are deprecated should be completely removed from the documentation and help output (even if they are accepted and ignored in the code).

* -no-cgo is in the changes file as being removed so should not appear in the docs nor help output.
* -cgo should be mentioned in the changes file as being deprecated (it doesn't do anything now) and should be removed from docs and help output.
* -use-shlib seems to do something, doesn't look obsolete to me.
* -so-name is still used from what I can make out, so also does not look obsolete.

Tidying that up will be appreciated. Thanks.

ysmilda:
Tidied it up by removing the `-cgo` and `-no-cgo` flags from the docs and help files. Left the handling intact so that existing setups that use `-cgo` won't break. 

Changed the documentation for the `-use-shlib` and `-soname` flags to match the help output.

PR #3295, opened 2025-12-09
Directors support for lua

christophe-calmejane:
Adding directors support for lua.

### DIFF
--- a/Doc/Manual/Contents.html
+++ b/Doc/Manual/Contents.html
@@ -1281,6 +1281,13 @@
 <li><a href="Lua.html#Lua_nn37">Userdata and Metatables</a>
 <li><a href="Lua.html#Lua_nn38">Memory management</a>
 </ul>
+<li><a href="Lua.html#Lua_directors">Cross language polymorphism using directors</a>
+<ul>
+<li><a href="Lua.html#Lua_directors_enabling">Enabling directors</a>
+<li><a href="Lua.html#Lua_directors_usage">Using directors in Lua</a>
+<li><a href="Lua.html#Lua_directors_exceptions">Exception handling in directors</a>
+<li><a href="Lua.html#Lua_directors_caveats">Director caveats and limitations</a>
+</ul>
 </ul>
 </div>
 <!-- INDEX -->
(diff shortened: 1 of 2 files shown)


christophe-calmejane:
Hi.

Just so I fully understand you: Are you planing on merging *like this* and then I'll have to *make another PR with all your requested fixes*? Or are you updating it *yourself while merging*?

- 1/ Sorry about the file naming, I've noticed it after pushing and didn't have the time to fix it as I tried to improve std::optional for C#.


- 4/ Yes this is correct. Please feel free to add/update/suggest anything in the documentation, you'll probably write english sentences way better than me, and you have a better global understanding of it and better see what could be missing/not clear for users.

Regarding the coding conventions, I tried to do it then my global clang-format messed up with the whole file. I tried to manually fix everything it changed but it took too long. Would it be possible to provide your own clang-format file so we don't have to think too much about it when contributing (if you have one of course)?

Thanks

Note: I've quite a lot of other things prepared for lua (set, array, optional, ...) but I've too many dependencies (at this time) from other pending PR that depend on it. Hopefully I can provide them quite soon.

wsfulton:
I've merged this in now, see 3392cfdb87d1beb71dfab49df6e8eae1ee08998c...f66873434caeeb7b5080401526de333a9a1719d4.

Code formatting is documented at https://swig.org/Doc4.4/Extending.html#Extending_coding_style_guidelines. I recommend reading the whole of the Extending chapter for any non-trivial contributions to SWIG. I fixed the current style issues in 3392cfdb87d1beb71dfab49df6e8eae1ee08998c, then 'beautified' your lua.cxx additions, so if you compare 9d455889413feac5633ae3df0f575753c6dea008 and your original submission, you'll see just code style fixes. Alas the beautify tool is by no means  perfect and often needs fixing up formatting issues it creates. A reliable clang-format alternative would be nice to have.

Does swig_get_override always get a Lua function? What happens if there is no Lua override function? Clarification of this is needed in the docs.

Oh and the documentation additions are very welcome and are in fact vital. Documentation additions, even if they have a few English issues, are way more important than no docs at all.

christophe-calmejane:
Thanks, created https://github.com/swig/swig/pull/3302 
(I noticed a couple issues in my examples in the doc that I also fixed)

Issue #3297, opened 2025-12-10
Parameter packs in function arguments are not expanded correctly


mmomtchev:
The following input:

```swig
%module template_pack

%inline %{
  template <typename T, typename... U>
  class Object {};

  template <typename T, typename... U>
  int fn1(Object<T, U...>) {
    return 1;
  }
%}

%template(fn1_3) fn<int, int, int>;
```

produces a local variable of invalid type:

```cpp
  SwigValueWrapper< Object< int,int ... > > arg1 ;
```

mmomtchev:
Certainly a bug because the following is correct:

```swig
%module template_pack

%inline %{
  template <typename T, typename... U>
  class Object {};

  template <typename T, typename... U>
  int fn2(T (*fn)(U...)) {
    return 1;
  }
%}

%template(fn2_3) fn2<int, int, int>;
```

produces:

```cpp
int (*arg1)(int,int) = 0 ;
```

mmomtchev:
The problem is that `SwigType_variadic_replace` uses `SwigType_split` which does not work for `Object<(T,v.U)>` - It is considered to be an unidivisble type.

Issue #3172, opened 2025-05-06
c++ to  c# 。  std::function <void(double&, double&)>,  how to write the interface File.? and Why Swig not privide Default std_function.i


qd1308504206:
Hello SWIG Team!
I Use the Version is 4.3.1

I hope this message finds you well. I'm writing to inquire about a technical detail regarding SWIG's support for C++ standard library features.

While working on wrapping a C++ project to C#, I noticed that the std_function.i module file appears to be missing from SWIG's standard library interfaces. This is critical for my current development task, as I need to expose std::function objects (or compatible C++ callable entities) to C# code.

Could you kindly explain:

    Is there a specific reason why std_function.i isn't included in SWIG's default distributions?
    Are there alternative approaches or existing typemaps I could use to achieve similar functionality?
    Would this be a candidate for future SWIG development, or is there a philosophical limitation I should be aware of?

Any guidance you could provide would be greatly appreciated! I'm happy to share more context about my use case if needed.

Thank you for your time and for maintaining such a powerful tool.

Best regards,
`
// for example
//coord.h
class Coord
{
public:
 bool   transform(std::function<void(double&, double&y)> fun) 
{ 
      fun(this->x, this->y);
      return true;
}

public:
    double x = 0; double y = 0;
}
`

`
//coord.i
%module(directors="1") ESharp
%feature("director");
%include "std_function.i"
%template(DoubleFunc) std::function<void(double&, double&)>;
%include "coord.h"

`

ojwb:
There's not really a good reason why there's no explicit support for `std::function`, just that SWIG's a volunteer project and there are more things that could be done than there is time to do them.

I've never used `std::function` myself so I'm not familiar with exactly how it works, but maybe you can `%ignore` the method taking `std::function` then use `%extend` to provide a shim alternative which creates a suitable temporary `std::function` object from something SWIG can wrap, then calls the method with that temporary `std::function`.

Closing as a duplicate of #341.

wsfulton:
Now fixed, see issues #341 and #3298.

Issue #3021, opened 2024-09-13
C++-specific varargs syntax not supported


ojwb:
I only noticed this when looking at https://en.cppreference.com/w/cpp/language/variadic_arguments rather than in real world code, but the original C++ syntax for varargs was this, which SWIG fails to parse (but is accepted by GCC and clang):

```
int f(int n...);
```

Apparently when C added prototypes, it adopted this syntax except it added a `,`:

```
int f(int n, ...);
```

Then when C++ was standardised (C++98) they added support for the C syntax for compatibility.

It seems in practice the original C++ syntax is used much less often than the C syntax.

My quick attempt to implement it adds 9 shift/reduce conflicts:

```
diff --git a/Source/CParse/parser.y b/Source/CParse/parser.y
index 3339ff5b..72c62e01 100644
--- a/Source/CParse/parser.y
+++ b/Source/CParse/parser.y
@@ -5222,6 +5222,14 @@ rawparms	: parm { $$ = $parm; }
 		  set_comment($parm, $DOXYGENPOSTSTRING);
 		  $$ = $parm;
 		}
+		| parm ELLIPSIS {
+		  SwigType *t = NewString("v(...)");
+		  $$ = NewParmWithoutFileLineInfo(t, 0);
+		  Setfile($$,cparse_file);
+		  Setline($$,cparse_line);
+		  set_nextSibling($parm, t);
+		  $$ = $parm;
+		}
 		| %empty {
 		  $$ = 0;
 		}
```

ojwb:
The conflict seems to be with `rawtype:` [...] `| rawtype[in] ELLIPSIS {` [...] which I think happens for the case `int f(int...)`.

That rule was added in 70837bbc2638ec18bd07d333c6e474c89f4ce695 for template parameter pack arguments.

wsfulton:
Just a note that this is still a syntax error and is not fixed by #3298, but will probably benefit from #3298 once the parser error has been fixed.

Issue #341, opened 2015-02-14
Feature Request: Add support for std::function


hlprasu:
Is std::function support for adding callbacks from target language, planned?

In https://stackoverflow.com/questions/11516809/c-back-end-call-the-python-level-defined-callbacks-with-swig-wrapper#new-answer,
a cool solution via a custom class PyCallback is used in the answer by Flexo. Any chance this could make to mainstream SWIG?

bzeller:
Is there any work going on for this issue? That would be a really cool feature to have.

ojwb:
I don't know of anyone currently working on this.

wsfulton:
Please see #3298.

PR #3298, opened 2025-12-12
support expanding parameter packs and wrapping `std::function`

mmomtchev:
The main goal of this PR is to support wrapping `std::function`. It has simple default language-independent support that works as an object that can be called using a normally wrapped function that takes the `std::function` as argument, as well as the call arguments. Languages which have support for callable objects can use `%extend` to make it work as a normal function. Other languages will have to use an external helper.

This PR:
 * Fixes a number of problems related to expanding parameter packs when generating local variables
 * Fixes casting of templated types - it did not work before for any templated types
 * Merges the casting and the moving in a single method in order to allow non-copyable objects to be properly casted

There are two new tests which should work for all languages - one for parameter packs in general and one for the `std::function` support itself.

Refs: #3297 
Refs: #3021
Refs: #3172
Refs: #341


wsfulton:
Nice. Couple of initial questions:

1. Why have "extern C++" in the testcase?
2. Any objections to squash merging this. Otherwise you could rebase it into say a handful of smaller independent commits.

> Merges the casting and the moving in a single method in order to allow non-copyable objects to be properly casted

I'd like to look closely at this bit. Is there a simple example that was causing this to be a problem (I thought we had tests covering this already).

mmomtchev:
`extern "C"` is necessary for sure because many languages have `extern "C"` sections in their headers. I haven't checked if `extern "C++"` is necessary.  Anyway, this part was because of the code splitting, I replaced it with a simple `%inline`.
You can squash merge.

wsfulton:
>>   Merges the casting and the moving in a single method in order to allow non-copyable objects to be properly casted

> I'd like to look closely at this bit. Is there a simple example that was causing this to be a problem (I thought we had tests covering this already).

What is a cut down testcase that caused non-copyable objects to not work? Bit hard to tell with many features rolled into one merge request.  If there is a move problem, I'd like to see clearly what it is without all the std::function changes as I thought I'd solved all those problems.

mmomtchev:
The `cpp11_template_pack` test does not work. This is #3297. It is about expanding parameter packs.

I included non-copyable objects because I had to enable casting on all templated types - something that did not work before. This change impacts non-copyable types. 

> > > Merges the casting and the moving in a single method in order to allow non-copyable objects to be properly casted
> 
> > I'd like to look closely at this bit. Is there a simple example that was causing this to be a problem (I thought we had tests covering this already).
> 
> What is a cut down testcase that caused non-copyable objects to not work?

wsfulton:
Thanks, this triggered quite a bit of interesting improvements! I've split the 64 commits in this pull request into 2 and applied them as 45c0bd45008892fc1820d5da130063d53e4e5f01...6f66d3d94c78204115b44ba8acc19acdf78f72cc.

I've then improved this area in various ways resulting in 6f66d3d94c78204115b44ba8acc19acdf78f72cc...98dd0a3160bf19104068986c800f7c052b00aa7d and 21d51452ea16aaec46685ef5a17b11f456f0ea8a...a4e8e182dbab1d81672adac694ccf8845c7e930c - hope it all makes sense and helps.

PR #2334, opened 2022-08-03
Add QuickJS Javascript Emitter and Interfaces

sax1johno:
Created a new Javascript emitter and interface files for QuickJS.

PR #3303, opened 2025-12-17
Proposed Javascript engine: QuickJs

didier-jorand:
Hello
QuickJS is a fast and light javascript engine ([link](https://bellard.org/quickjs/)). 
This branch implements the JSEmitter for this engine, update the Tools/javascript to run the examples and test-suite and updates some tests (runme.js) to get them running in "strict" mode;
The documentation (javascript.html) is updated.
All tests succeed for the QuickJS variant and still succeed for all the other javascript variants (no regression or side-effect).
Can you review this implementation and make any comment that would improve it and facilitate its integration in the swig main branch ?
Thanks,
Didier


wsfulton:
Hi @didier-jorand. Thanks for your contribution to SWIG. A new contribution requires fulfilling the requirements in https://swig.org/Doc4.4/Extending.html#Extending_prerequisites. As you're adding a new engine to the already existing Javascript target language rather than adding a completely new language, it is a lot easier to fulfill these requirements. However, please check them, especially the code formatting style in javascript.cxx.

After a quick cursory look, the only obvious areas I found to address are:
1. The patch seems to break the current javascript testing in Github.
2. There is no Github testing added for quickjs. You'll need to add quickjs to .github/workflows/linux.yml and the Tools/CI-linux-* files.

As this is the 4th Javascript engine, it would be nice to also see in the SWIG documentation a section listing the available engines and why one would choose each over another, so pros/cons of each. In particular relating to wrapping C++ code, that is, information that a SWIG user might want to know.

I would appreciate a review/comments by any other Javascript experts / Javascript SWIG contributors who see this.

ojwb:
There was also #2334 to add QuickJS support which got closed due to lack of any response from the submitter, but might be worth checking for useful parts.

didier-jorand:
Hi @wsfulton @ojwb , thanks for your feedback.
I have corrected the format on the javascript.cxx file, identified and corrected issues with the currrent javascript testing. 
I also added some github testing, but limited to 3 use cases because, at this stage, the code generated for quickjs requires gcc-13 or above.
Regarding the 2022 pull request, not much to get from it as the API of QuickJS evolved a lot since then.
For the pro/cons of the different javascript engines, it is beyond our knowledge to make such a comparison. At this stage, we are looking at using  QuickJS to provide javascript scripting to applications (similar to what we have seen a lot with Lua).
Best regards,
Didier

aidanhs:
Hi, I've been playing with this branch a bit and have some comments/improvements I've made in a PR at https://github.com/didier-jorand/swig-quickjs/pull/1. To detail here, I have:

1. added a TODO about overloaded functions not being exception safe
2. added some additional bits of exception handling
3. avoided setting base_prototype on object_class because it causes a prototype loop that quickjs detects and throws an exception for (which previously wasn't being detected before 2). I don't know if this is the correct solution - it's possible there's actually a sane base_prototype to pick, but I couldn't quite wrap my head around it at the time and this seemed to work
4. moved DEFAULT_NAMESPACE to be initialised earlier so `SWIG_QuickJS_swig_object_class_definition` and `SWIG_QuickJS_packed_data_class_definition` can be registered on it (this was also causing an exception)
5. allowed bigints to be passed to C++ functions. I probably wouldn't put this into the plugin as it will work very poorly in 32bit, and only in restricted cases for 64 bit - I just have it so I can cleanly round trip a uint64_t into js and back out, which is not guaranteed otherwise

### Some thoughts on exceptions
Exceptions in quickjs are stored as global state on the runtime. Functions in the C API *must* be checked for their return value and the exception from quickjs retrieved (so it gets unset on the interpreter), otherwise future calls to the interpreter may incorrectly infer that they have errored.

Checking quickjs return values is made tricky because you need to read the comments (and sometimes the function itself) to understand if it's just an infallible `JSValue`, a `JSValue` which may be an exception which indicates an error, a -1/0/1 (like `JS_GetOwnPropertyInternal` which means error/doesn't exist/exists) or a -1/0 (for error/ok).

I've added a few bits of additional exception checking, but there's a lot more that I think is missing - but it works well enough for me now so I'm not motivated to do more.

didier-jorand:
Hello,
Thanks for your valuable analysis and contribution.
In details:
1. "added a TODO about overloaded functions not being exception safe": good catch, to solve the issue, I have added a call to JS_GetException when the ctor is successful, this will clear any previous exception
2. "exception handling": you're right, I missed some checks on returned value (assuming maybe that exception would never occur...). I have integrated your modifications and I will check (later) if I missed other
3. "base_prototype on object_class": tricky one, as object_class is the top of the class hierarchy, it should not be addressed like other classes ; therefore I implement your proposal to exclude it in 'registerClass'
4. "moved DEFAULT_NAMESPACE": OK, but this breaks the test-suite in dynamic_cast (or any use of %init in swig file that expect to insert code in the init function). I have added a "local_module_initialize" function, but it is not 'elegant'.
5. I applied your tweak to long and unsigned long for the time being, but I am not sure on what to do with BigInt in a more general way.

I have to test a little bit more all these modifications (the examples and test-suite are OK, so t looks good now) and I will push in the master branch soon.

Lib/javascript/quickjs/javascriptrun.swg
```diff
@@ -364,7 +365,7 @@ SWIGRUNTIME int SWIG_QuickJS_ConvertPtr(JSContext *ctx, JSValueConst val, void**
         JSAtom atom = JS_NewAtom(ctx, pClsDef->class_name);
         if(atom != JS_ATOM_NULL) {
           found = JS_HasProperty(ctx, base_map, atom);
-          if(found == -1} { return SWIG_ERROR; } /* exception */
+          if(found == -1) { return SWIG_ERROR; } /* exception */
```
ojwb: Please ensure the testsuite at least passes locally **before** pushing.

Each time you push all the SWIG devs get another github notification; also ~130 CI jobs are run, which burns a lot of CPU time - pointlessly so when there are multiple trivial syntax errors in the code (I see at least this one and two in `Lib/javascript/quickjs/javascripthelpers.swg`).

(If this branch really just isn't ready yet, please close the PR and open a new PR when it is.)

Also the code formatting here doesn't match SWIG conventions - should be space after `if`, i.e. `if (found == -1)`.

didier-jorand: Agreed, that push was a mistake.

wsfulton:
Hi @didier-jorand. Is this ready for review now? If so, please take a look at the requirements for adding a new target language and make an assessment regarding Experimental vs Supported status. Create and add a list of the requirements as a comment here and then add the evidence for meeting each requirement. I'll then review it accordingly.

Just one small thing, can you try use apt to install quickjs. It should be quicker and more reliable than downloading the source and building it in each CI job (although 40s to download and build is quite fast). Also rebase onto latest master. Thanks.

didier-jorand:
Hi @wsfulton. In my opinion, the module is ready for review.


Regarding the requirements for the supported status (according to Doc4.4), it seems to comply:
- _It is in a mature, well functioning state_ => yes
- _It has its own comprehensive chapter in the documentation. The level of documentation should be comprehensive and match the standard of the other mature modules._ => javascript chapter has been updated, ready for review (it precises in particular the preconditions (quickjs-2025-09-13)).
- _It passes all of the main SWIG test-suite. The main test-suite is defined by the tests in the C_TEST_CASES, CPP_TEST_CASES and MULTI_CPP_TEST_CASES lists in Examples/test-suite/common.mk. All the newer C++ standard tests need to work and are grouped together, such as CPP11_TEST_CASES for C++11. These more 'modern' C++ standards are only tested though if the compiler is detected as supporting the given standard.
The test-suite must also include at least twenty wide-ranging runtime tests. The most mature languages have a few hundred runtime tests. Note that porting runtime tests from another language module is a quick and easy way to achieve this._ => all current javascript test-suite (from other engines like node or jsc) pass (717 tests OK on fedora 44).
- _It supports the vast majority of SWIG features. Some more advanced features, such as, directors, full nested class support and target language namespaces (nspace) may be unimplemented. A few support libraries may be missing, for example, a small number of STL libraries._ => same level of features as other javascript engines
- _It provides strong backwards compatibility between releases. Each point release must aim to be fully backwards compatible. A point release version is the 3rd version digit, so each of the x.y.* versions should be backwards compatible. Backwards compatibility breakages can occur in a new major or minor version if absolutely necessary and if documented. A major or minor version is the first or second digit in the three digit version.
Fixing unintended regressions in the Supported languages will be given higher priority over the Experimental languages by the core SWIG developers._ => not relevant at this stage
- _Examples must be available and run successfully._ => all current javascript examples work + added a multiple inheritance example (QuickJS only)
- _The examples and test-suite must be fully functioning on the Github Actions Continuous Integration platform._ => yes


Regarding the use of apt to install quickjs, Ubuntu is providing a 2021 version of QuickJS, obsolete (the API has changed significantly).

Rebase onto latest master done.

wsfulton:
Thanks, an initial observation is this pull request is overwhelmed by runtime tests changes. Adding in 'var' or 'let' might be good Javascript practice, but is not something that should be part of this merge request, nor is it in any way important for SWIG wrappers. What is the motivation for this? What I'd like to see is zero runme.js file changes and quickjs passing them all. Any reason quickjs cannot support this?

didier-jorand:
Hi @wsfulton. 
This deserve indeed some explanations. 

Short version : the runtime tests changes were initially needed because quickjs ran in 'strict' mode ; this is not required anymore and I have reverted 99% of the runme.js files in their original version in the PR.

Long version : the initial intend was to use the out-of-the-box _qjs_ interpreter. But it supported only modern javascript construct, in particular the "`import`" module statement that triggers the javascript "strict" mode. This implied to replace dynamically all the "`require`" statements by "`import`" statements in the examples and tests (sed + awk managed to do that) but also to do other modifications, like declaring all variables (`var/let`) and properly use the `delete` statement. At the end of the day, it seemed too complex and we seek another way. It was then decided to implement the `require` function in quickjs and to have a dedicated javascript executable (Tools/javascript/quickjs) as it was already the case for the other engines. But the modified runtime tests remained : the examples and tests were successful, not only for QuickJS but also for the other javascript engines and we disregarded these "improvements". These should have been transferred in another pull request as, for sure, it has nothing to do anymore with the quickjs engine implementation... 
So, for now, the swig-quickjs master branch (the one in the PR) has only very few runme.js modifications and we keep the "strict" version in another branch (js_strict).

Not related, but a new version of QuickJS was published on June 4th. It implies no API changes and passes all the tests. The documentation (javascript.html) has been updated and CI tests have been added. To avoid multiple push in the repository, this is also included in the master.

Tools/javascript/javascript.cxx
```diff
@@ -31,6 +32,15 @@ int main(int argc, char* argv[]) try {
       shell = JSShell::Create(JSShell::V8);
     } else if(strcmp(argv[idx], "-jsc") == 0) {
       shell = JSShell::Create(JSShell::JSC);
+#ifdef QJSPATH
```
wsfulton: Should be in path, not a compile time option.

Tools/javascript/javascript.cxx
```diff
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
```
wsfulton, line 4: This is not portable, can you make it cross-platform?

wsfulton:
## Review: QuickJS JavaScript engine

Thanks for this — a well-scoped contribution that adds a working fifth backend to the JavaScript module with examples, docs, and broad runtime coverage. Findings below are grouped by severity. 

### Blocking

- **Default destructor frees objects with the wrong allocator** — `Lib/javascript/quickjs/javascriptcode.swg:123` (the default `js_dtor` fragment; the `js_dtoroverride` variant at ~`:152` is fine).
  The default `js_dtor` hardcodes `js_free_rt(rt, ($jstype)swig_object->ptr);`. The payload is allocated by the standard SWIG constructor action (`new T(...)`), so freeing it with QuickJS's runtime pool allocator is an allocator-*family* mismatch (C++ heap vs the QuickJS pool) **and** the C++ destructor never runs — undefined behaviour, not just a leak. The `js_dtoroverride` variant correctly uses `${destructor_action}`, so this only affects the default / no-custom-destructor path. The model to follow is the v8 engine, whose default `js_dtor` uses the `$jsfree` placeholder (`Lib/javascript/v8/javascriptcode.swg:133`) — `$jsfree` is built in `Source/Modules/javascript.cxx` as `delete (T*)` / `delete[] (T*)` / `free` and so respects how the object was allocated. (For reference, the jsc engine hardcodes `free()` in its default dtor — a pre-existing milder flaw that skips the C++ destructor but at least pairs with `malloc`; QuickJS's `js_free_rt` is worse because the allocator family is wrong too.) Please route the default dtor through `$jsfree swig_object->ptr;` as v8 does.

- **`JS_NewClassID` called inside `assert()`** — `Lib/javascript/quickjs/javascriptinit.swg:164` and `:286`.
  Under `-DNDEBUG` the whole expression is compiled out, so the class IDs are never allocated in release builds. Call the function unconditionally and check the result separately. (The `registerClass` path at `javascripthelpers.swg:830` already does this correctly — these two are the outliers.)

### Should-fix

- **quickjs-ng vs. vanilla QuickJS is contradictory.** The runtime uses quickjs-ng-only APIs — `JS_PrintValue` (callback form), `JS_TAG_SHORT_BIG_INT` / `JS_TAG_STRING_ROPE` (`javascriptrun.swg:642-696`), `JS_SetModuleLoaderFunc2` (`Tools/javascript/swigqjs.c:354`), `__JS_NewFloat64` (`javascripthelpers.swg:574`) — yet `configure.ac` has a quickjs-ng detection branch while the docs state "quickjs-ng fork is not supported yet." Please make the config, CI, and docs agree on exactly which fork and versions are supported.

- **`SWIG_QuickJS_registryGet` off-by-one** — `javascripthelpers.swg:208`. `if ((uint32_t)idx > len)` should be `>=`; valid indices are `0..len-1`, so `idx == len` currently passes the bounds check and reads past the end.

- **Public header missing its own include** — `Lib/javascript/quickjs/swigquickjs.h` uses `JSContext` / `JSValue` / `JSValueConst` but does not `#include <quickjs.h>`, relying on include ordering. As a header intended for external modules it should include quickjs itself or document the requirement.

- **Init error paths leak / discarded dup** — `createRegistries` in `javascripthelpers.swg` returns on several error paths without freeing `globalObject` / `registry` / the current atom, and `JS_DupValue(ctx, registry)` at `javascripthelpers.swg:34` discards its result (looks like a logic slip rather than a deliberate simplification).

- `Tools/javascript/swigqjs.c` hardcodes the `/usr/lib/qjs/*.so` require path, has an unimplemented `QUICKJS_CPATH` (already flagged with `XXX`), and bakes an absolute `QJSPATH` — Linux-only and non-relocatable. Can you make cross-platform and drop the unistd.h (unix only) header?

### Minor

- `QuickJSEmitter::QuickJSEmitter()` passes `JSEmitter::JavascriptCore` to the base constructor in `Source/Modules/javascript.cxx` — should be `JSEmitter::QuickJS`. Latent only (the base `engine` member is read just via the local in `main()`), but wrong and confusing.
- Leftover commented-out debug code in the new emitter (`// Printv(stdout, ...)`, `/* XXX trace for debug */`) and in the `.swg` files (`#if 0` blocks at `javascriptrun.swg:595` and `javascriptinit.swg:297`).
- `QuickJSEmitter::enterFunction` is a no-op wrapper around the base method — drop it.
- C++ `//` comments inside C-compilable runtime fragments — `javascriptrun.swg:67`, `javascriptcode.swg:53,56`.
- Typos: `mechanisem` (emitter comment); docs `interpretor`, `is build`, `compiled..`, double-slash URL `bellard.org/quickjs//`; runtime `retreive`, `funtion`, `swig_packet_data` (→ `packed`), `$jsdor`. Several `<p>` tags are left unclosed in the new documentation section.
- Trailing whitespace in `configure.ac`, `Examples/Makefile.in` (`endif    `), `Examples/test-suite/langobj.i` (`#endif    `), and `swigquickjs.h`.
- Delete the added multiinherit and defargs examples. We only ship a small set of examples, so you can copy from other target languages, but not add new examples. The test-suite should cover these already.

## Final steps
1. Can you fix the **Blocking** and **Should-fix** issues above. I can deal with the **Minor** prior to merge.
2. QuickJS meets the fully supported category as far as I'm concerned. I don't think there is much more to do apart from addressing the few small points above. Can you please take on the role of maintainer for the QuickJS module for the time being?
3. Can you merge/rebase onto master before you next push to pick up the latest.

## Questions
1. Why do you have `--with-jsquickjssrc` option for configure in CI? We don't use it in CI for anything else on Linux. Configure should 'just work' and find the installed version of target languages.
2. When done, are you okay if I squash merge this into master or do you have some history you would really like to preserve?

wsfulton (changes requested):

Tools/javascript/javascript.cxx
didier-jorand: Hi @wsfulton.

Blocking, should-fix and minor (hopefully) fixed.

`--with-jsquickjssrc` option for configure removed, the requirement is to have QuickJS installed (documentation updated).

Tools/javascript/javascript for QuickJS has been fully rewritten to create a QuickJS shell in C++, similar to V8, resolving portability issue.
Question: Needed to add the flag '-rdynamic' when linking the javascript executable. Works in Linux but not needed for Mingw and no idea regarding MacOS. Should it be checked somehow in the Makefile ?

No problem to maintain this module.

wsfulton:
Hi @didier-jorand, I see you rebased this branch. Are you able to respond to the review?

wsfulton:
> Hi @didier-jorand, I see you rebased this branch. Are you able to respond to the review?

Ah, I see you have responded. It was in response to a single code comment, not in the main issue, so sorry I missed it. I should be able to take a close look at the changes soon.

PR #3310, opened 2026-01-14
octave: Fix save_binary prototype

### LINKED ISSUE #3282: Octave 10.3 C++ warnings
With Octave 10.3 locally on Debian unstable (I don't see an easy way to test with this in CI) I see a lot of warnings from the testsuite compiling generated C++ code which look concerning:

```
/usr/include/octave-10.3.0/octave/../octave/ov-base.h:780:30: warning: ‘virtual bool octave_base_value::save_binary(std::ostream&, bool)’ was hidden [-Woverloaded-virtual=]
  780 |   virtual OCTINTERP_API bool save_binary (std::ostream& os, bool save_as_floats);
      |                              ^~~~~~~~~~~
```

SWIG defines this method except the last parameter is `bool&` - it looks like this method signature has changed in Octave's C++ API and used to be `bool&` there too.  I'm not sure which Octave version this changed in.  I'd expect SWIG needs to generate different code for newer Octave, though the tests still seem to pass which surprises me.  Perhaps this `save_binary()` method doesn't get exercised by anything in our test suite?

jschueller:
Closes #3282

### DIFF
--- a/Lib/octave/octrun.swg
+++ b/Lib/octave/octrun.swg
@@ -949,7 +949,7 @@ SWIGRUNTIME void swig_acquire_ownership_obj(void *vptr, int own);
       return true;
     }
 
-    virtual bool save_binary (std::ostream& os, bool& save_as_floats) {
+    virtual bool save_binary (std::ostream& os, bool save_as_floats) {
       return true;
     }
 
@@ -1237,7 +1237,7 @@ SWIGRUNTIME void swig_acquire_ownership_obj(void *vptr, int own);
     virtual bool load_ascii (std::istream& is)
       { return ptr->load_ascii(is); }
 
-    virtual bool save_binary (std::ostream& os, bool& save_as_floats)
+    virtual bool save_binary (std::ostream& os, bool save_as_floats)
       { return ptr->save_binary(os, save_as_floats); }
 
     virtual bool load_binary (std::istream& is, bool swap, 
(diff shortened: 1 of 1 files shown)


ojwb:
Thanks.

In #3282 you wrote:
> it looks like this is from octave 6:

So presumably this patch creates a similar problem for octave < 6 to the one it solves for octave >= 6.  However `Doc/Manual/Octave.html` indicates we still support 5.2:

> SWIG is regularly tested against the following versions of Octave: 5.2, 6.4. SWIG 4.2.1 has also been manually tested with 7.3, 8.4 and 9.0.

So either this needs to be conditional, or we need to drop support for 5.2 (and update the documentation to reflect that).  It looks like since GHA added Ubuntu 24.04 and dropped 20.04 our CI actually tests 6.4 and 8.4, so given we're no longer testing 5.2 and it's more than 5 years old dropping support seems reasonable.

Testing manually with Octave 10.3.0 this patch does fix the `save_binary` problem.  So I've merged and fixed up the docs.

We still don't get a clean compile.  Lots of deprecation warnings like this from `make check-octave-examples` but this seems like an upstream Octave problem we can't really do much about:

```
/usr/include/octave-10.3.0/octave/../octave/oct-stream.h:236:24: warning: ‘template<class _Codecvt, class _Elem, class _Tr> class std::wbuffer_convert’ is deprecated [-Wdeprecated-declarations]
  236 |   std::unique_ptr<std::wbuffer_convert<convfacet_u8, char>> m_converter;
      |                        ^~~~~~~~~~~~~~~
```

PR #3311, opened 2026-01-14
Deprecate %typedef directive

### LINKED ISSUE #3019: Deprecate %typedef
SWIG accepts `%typedef` but just treats it as an SWIG-specific alias for `typedef` (and poking git history this seems to be true as far back as we have useful history).  It's not documented and has no test coverage.  The only mentions outside the source code are in two CHANGES entries from 1996:

> 14. %typedef modified to insert a corresponding C typedef into
>     the output file.

> 7.  Better implementation of %typedef directive.

This feature seems to serve no useful purpose, but I spotted a use "in the wild":

https://sources.debian.org/src/soapysdr/0.8.1-5/python/SoapySDR.in.i/?hl=105#L105

That one looks to me like it's just a typo (or thinko) for `typedef` that SWIG quietly accepted so the developer didn't notice.

I think it make sense to deprecate this, and either just remove it, or emit a warning (and actually remove it in a few years).  The recommended solution would be just to use `typedef` instead, which should work with existing SWIG versions going back a long way.

@wsfulton Thoughts?

jschueller:
Closes #3019

### DIFF
--- a/Examples/test-suite/errors/pp_deprecated_typedef_directive.i
+++ b/Examples/test-suite/errors/pp_deprecated_typedef_directive.i
@@ -1,0 +1,4 @@
+%module xxx
+
+/* Checks the deprecated %typedef directive warning is issued */
+%typedef int MYINT;
--- a/Examples/test-suite/errors/pp_deprecated_typedef_directive.stderr
+++ b/Examples/test-suite/errors/pp_deprecated_typedef_directive.stderr
@@ -1,0 +1,1 @@
+pp_deprecated_typedef_directive.i:4: Warning 127: %typedef directive is deprecated, use plain typedef instead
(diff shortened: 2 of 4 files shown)


ojwb:
It'd be better to have test coverage for this to confirm the warning actually triggers, check that it gets reported on the line number where the `%typedef` is, and to catch if a later change would regress on either of these.

Testcases in `Examples/testsuite/errors` get stderr checked against the corresponding `.stderr` file (the testcase doesn't actually need to trigger an error - it's really tests of "diagnostics" rather than just "errors").

jschueller:
yes, done

ojwb:
Thanks, merged.  I added a CHANGES entry in ea62c86e82b34348a37c0816f2aea4ec523b9915 (but referenced the wrong bug number in the commit message).

PR #3312, opened 2026-01-14
Use clang-format on SWIG source

wsfulton:
I find during code reviews I end up spending time on looking at, getting distracted by and fixing code formatting. I would like to remove this problem to speed up reviews as well as helping contributors understanding SWIG's coding style with a bonus of keeping the formatting consistent in the source code base.

My first attempt back in 2006 was to introduce the 'beautify' target in Source/Makefiles.am. This used the GNU indent tool to format the code. This was problematic, as while it worked okay with C code, it managed to make a mess of C++ code and required manual fixups. So, it could never be automatically invoked. In fact it was hardly used Twenty years later, the obvious replacement choice is clang-format.

In an ideal world, we would switch to clang-format to use the identical formatting (for C code) that GNU indent uses in order to avoid a large purely whitespace commit. I've spent some time now experimenting with clang-format and propose using the .clang-format config file in this merge request. I think I've found a fairly good close equivalent but with one important change. I found that its output was a bit buggy/weird if the tab setup we use was used for indentation. But I have had good success if we switch to spaces instead of tabs. I think our tab and indentation setup was always viewed as a bit odd, so I'm proposing to use a 2 space indent and no tabs going forwards.

The other notable changes are:
1. Fixing up the currently haphazard and inconsistent formatting of calling a function that take many parameters. They can either be one long line now or one parameter per line.
2. Fixing up the currently haphazard formatting of multi-line if statements. Single line if statements only (broken at 160 characters if too long).

From what I understand, the clang-format tool parses the code losing the input format, then strictly reformats the code according to the global input config file. There is no way to 'break the rules' except by using `\\ clang-format off` comments. There is also a regex tool, but it isn't a serious option really. This means that we can't 100% match the original format nor can we have many localised special exception cases. But this is one of the strengths; once we've reformatted the code, it will remain consistently formatted going forwards if we use the tool for every commit.

I havn't enforced the use of the tool in this merge request, nor documented it. I'll work on that next. I thought I'd put in this merge request as a heads up to see if anyone with experience of clang-format would like to offer some guidance and improvements before I land it in master.

I havn't tackled parser.y yet as clang does not support yacc files. Instead, I plan on moving the C functions out of parser.y into a .c file and whitespace correct parser.y for the time being. I also plan to add checks to enforce a ban on tabs in the Source directory (except of course in Makefiles).

Description of different commits
================================
The first 5 commits are a few small changes to make the current code base more clang-format friendly and removes a couple of subtle differences different versions (I tested versions 14-20). The next 3 commits are whitespace only changes for which clang-format is not necessarily needed. They convert tabs to 8 spaces and remove trailing whitespace in the same way that git commit does. clang-format can do these whitespace changes, but I've committed these separately to remove noise in the actual (more interesting) code reformatting. The final commit, 93af82c1ae793069a87bd148436a41edc1bd28ac, contains the remaining/actual formatting changes/corrections that clang-format makes.

Some stats to consider:
Total lines of code (.c .h .cxx), using wc -l: 111491

| Commit    | Lines changed | Lines changed (%) | Commit heading |
------------|---------------|-------------------|----------------|
| 7166e2e4e | 25    |   0.0  | Minor code changes in preparation for clang-format |
| 6f8ea0001 | 90    |   0.1  | Remove some use of tab4 and tab8 in perl, lua, python, perl |
| e5bf1c891 | 89    |   0.1  | Tweaks in python.cxx for clang-format |
| e485e1b1d | 6     |   0.0  | clang-format changes for older versions of clang-format |
| d5e375a14 | 19245 |   17.3 | Remove tabs at start of source files - cosmetic whitespace only change |
| c70d946d9 | 519   |   0.5  | Remove trailing whitespace in source - cosmetic whitespace only change |
| 399581948 | 749   |   0.7  | Convert remaining tabs to spaces |
| 93af82c1a | 7259  |   6.5  | Beautify using swig specific .clang-format |

The 'Lines changed' column is the number of lines deleted in the output of `diff diff --stat`.

From the above the switch to tabs from spaces is 18.0% of the lines and the remaining formatting adjustments are 6.5% of the line count. The 18% is higher than I'd hoped, but it'll be a one off and there should be no more whitespace only commits going forwards so seems like a good trade-off to me.

### DIFF
--- a/Source/.clang-format
+++ b/Source/.clang-format
@@ -1,0 +1,26 @@
+# This file format requires clang-format version 14 or later
+# SWIG requires clang-format version 18 or later for .clang-format-ignore support
+
+Language: Cpp
+BasedOnStyle: Google
+AlignArrayOfStructures: Left
+AlignTrailingComments: true
+AccessModifierOffset: -2
+AlignConsecutiveMacros: AcrossEmptyLinesAndComments
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakBeforeMultilineStrings: false
+BinPackArguments: false
+BreakConstructorInitializers: AfterColon
+ColumnLimit: 160
+ConstructorInitializerIndentWidth: 2
+ContinuationIndentWidth: 2
+DerivePointerAlignment: false
+IndentCaseBlocks: true
+IndentCaseLabels: false
+IndentGotoLabels: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+PackConstructorInitializers: CurrentLine
+PointerAlignment: Right
+SortIncludes: Never
--- a/Source/Modules/xml.cxx
+++ b/Source/Modules/xml.cxx
@@ -1,5 +1,5 @@
 /* -----------------------------------------------------------------------------
- * This file is part of SWIG, which is licensed as a whole under version 3 
+ * This file is part of SWIG, which is licensed as a whole under version 3
  * (or any later version) of the GNU General Public License. Some additional
  * terms also apply to certain portions of SWIG. The full details of the SWIG
  * license and copyrights can be found in the LICENSE and COPYRIGHT files
@@ -30,7 +30,7 @@ public:
 
   XML() :indent_level(0) , id(0) {
   }
-  
+
   ~XML() {
   }
 
--- a/Source/Modules/xml.cxx
+++ b/Source/Modules/xml.cxx
@@ -37,11 +37,11 @@ public:
   virtual void main(int argc, char *argv[]) {
     for (int iX = 0; iX < argc; iX++) {
       if (strcmp(argv[iX], "-help") == 0) {
-	fputs(usage, stdout);
+        fputs(usage, stdout);
       }
       if (strcmp(argv[iX], "-xmllite") == 0) {
-	Swig_mark_arg(iX);
-	xmllite = 1;
+        Swig_mark_arg(iX);
+        xmllite = 1;
       }
     }
 
@@ -61,8 +61,8 @@ public:
       Append(outfile, ".xml");
       out = NewFile(outfile, "w", SWIG_output_files());
       if (!out) {
-	FileErrorDisplay(outfile);
-	Exit(EXIT_FAILURE);
+        FileErrorDisplay(outfile);
+        Exit(EXIT_FAILURE);
       }
     }
     Printf(out, "<?xml version=\"1.0\" ?> \n");
@@ -98,51 +98,51 @@ public:
     while (ki.key) {
       k = ki.key;
       if ((Cmp(k, "nodeType") == 0)
-	  || (Cmp(k, "firstChild") == 0)
-	  || (Cmp(k, "lastChild") == 0)
-	  || (Cmp(k, "parentNode") == 0)
-	  || (Cmp(k, "nextSibling") == 0)
-	  || (Cmp(k, "previousSibling") == 0)
-	  || (*(Char(k)) == '$')) {
-	/* Do nothing */
+          || (Cmp(k, "firstChild") == 0)
+          || (Cmp(k, "lastChild") == 0)
+          || (Cmp(k, "parentNode") == 0)
+          || (Cmp(k, "nextSibling") == 0)
+          || (Cmp(k, "previousSibling") == 0)
+          || (*(Char(k)) == '$')) {
+        /* Do nothing */
       } else if (Cmp(k, "module") == 0) {
-	Xml_print_module(Getattr(obj, k));
+        Xml_print_module(Getattr(obj, k));
       } else if (Cmp(k, "baselist") == 0) {
-	Xml_print_baselist(Getattr(obj, k));
+        Xml_print_baselist(Getattr(obj, k));
       } else if (!xmllite && Cmp(k, "typescope") == 0) {
-	Xml_print_typescope(Getattr(obj, k));
+        Xml_print_typescope(Getattr(obj, k));
       } else if (!xmllite && Cmp(k, "typetab") == 0) {
-	Xml_print_typetab(Getattr(obj, k));
+        Xml_print_typetab(Getattr(obj, k));
       } else if (Cmp(k, "kwargs") == 0) {
-	Xml_print_kwargs(Getattr(obj, k));
+        Xml_print_kwargs(Getattr(obj, k));
       } else if (Cmp(k, "parms") == 0 || Cmp(k, "pattern") == 0) {
-	Xml_print_parmlist(Getattr(obj, k));
+        Xml_print_parmlist(Getattr(obj, k));
       } else if (Cmp(k, "catchlist") == 0 || Cmp(k, "templateparms") == 0) {
-	Xml_print_parmlist(Getattr(obj, k), Char(k));
+        Xml_print_parmlist(Getattr(obj, k), Char(k));
       } else {
-	DOH *o;
-	print_indent(0);
-	if (DohIsString(Getattr(obj, k))) {
-	  String *ck = NewString(k);
-	  o = Str(Getattr(obj, k));
-	  Replaceall(ck, ":", "_");
-	  Replaceall(ck, "<", "&lt;");
-	  /* Do first to avoid aliasing errors. */
-	  Replaceall(o, "&", "&amp;");
-	  Replaceall(o, "<", "&lt;");
-	  Replaceall(o, "\"", "&quot;");
-	  Replaceall(o, "\\", "\\\\");
-	  Replaceall(o, "\n", "&#10;");
-	  Printf(out, "<attribute name=\"%s\" value=\"%s\" id=\"%ld\" addr=\"%p\" />\n", ck, o, ++id, o);
-	  Delete(o);
-	  Delete(ck);
-	} else {
-	  o = Getattr(obj, k);
-	  String *ck = NewString(k);
-	  Replaceall(ck, ":", "_");
-	  Printf(out, "<attribute name=\"%s\" value=\"%p\" id=\"%ld\" addr=\"%p\" />\n", ck, o, ++id, o);
-	  Delete(ck);
-	}
+        DOH *o;
+        print_indent(0);
+        if (DohIsString(Getattr(obj, k))) {
+          String *ck = NewString(k);
+          o = Str(Getattr(obj, k));
+          Replaceall(ck, ":", "_");
+          Replaceall(ck, "<", "&lt;");
+          /* Do first to avoid aliasing errors. */
+          Replaceall(o, "&", "&amp;");
+          Replaceall(o, "<", "&lt;");
+          Replaceall(o, "\"", "&quot;");
+          Replaceall(o, "\\", "\\\\");
+          Replaceall(o, "\n", "&#10;");
+          Printf(out, "<attribute name=\"%s\" value=\"%s\" id=\"%ld\" addr=\"%p\" />\n", ck, o, ++id, o);
+          Delete(o);
+          Delete(ck);
+        } else {
+          o = Getattr(obj, k);
+          String *ck = NewString(k);
+          Replaceall(ck, ":", "_");
+          Printf(out, "<attribute name=\"%s\" value=\"%p\" id=\"%ld\" addr=\"%p\" />\n", ck, o, ++id, o);
+          Delete(ck);
+        }
       }
       ki = Next(ki);
     }
--- a/Source/Modules/xml.cxx
+++ b/Source/Modules/xml.cxx
@@ -21,14 +21,12 @@ XML Options (available with -xml)\n\
 static File *out = 0;
 static int xmllite = 0;
 
-
-class XML:public Language {
+class XML : public Language {
 public:
-
   int indent_level;
   long id;
 
-  XML() :indent_level(0) , id(0) {
+  XML() : indent_level(0), id(0) {
   }
 
   ~XML() {
@@ -97,13 +95,8 @@ public:
     ki = First(obj);
     while (ki.key) {
       k = ki.key;
-      if ((Cmp(k, "nodeType") == 0)
-          || (Cmp(k, "firstChild") == 0)
-          || (Cmp(k, "lastChild") == 0)
-          || (Cmp(k, "parentNode") == 0)
-          || (Cmp(k, "nextSibling") == 0)
-          || (Cmp(k, "previousSibling") == 0)
-          || (*(Char(k)) == '$')) {
+      if ((Cmp(k, "nodeType") == 0) || (Cmp(k, "firstChild") == 0) || (Cmp(k, "lastChild") == 0) || (Cmp(k, "parentNode") == 0) ||
+          (Cmp(k, "nextSibling") == 0) || (Cmp(k, "previousSibling") == 0) || (*(Char(k)) == '$')) {
         /* Do nothing */
       } else if (Cmp(k, "module") == 0) {
         Xml_print_module(Getattr(obj, k));
@@ -172,8 +165,7 @@ public:
     Printf(out, "</%s>\n", nodeType(obj));
   }
 
-
-  void Xml_print_parmlist(ParmList *p, const char* markup = "parmlist") {
+  void Xml_print_parmlist(ParmList *p, const char *markup = "parmlist") {
 
     print_indent(0);
     Printf(out, "<%s id=\"%ld\" addr=\"%p\">\n", markup, ++id, p);
@@ -240,7 +232,6 @@ public:
 
     Xml_print_hash(p, "typetab");
   }
-
 
   void Xml_print_hash(Hash *p, const char *markup) {
 
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -66,7 +66,7 @@ jobs:
           esac
 
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
       with:
         show-progress: false
 
@@ -121,7 +121,7 @@ jobs:
       run: systeminfo | findstr /B /C:"OS Name" /B /C:"OS Version"
 
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
       with:
         show-progress: false
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -22,9 +22,11 @@ permissions:
 
 jobs:
   # Code format checks, see https://github.com/marketplace/actions/clang-format-github-action
-  clang-format-oldest:
+  # The clang-format version used is slightly different for each OS workflow file
+  # This job gates the main build job so must succeed before the main build job is run
+  clang-format:
     runs-on: ubuntu-latest
-    name: clang-format 18
+    name: clang-format 18 # 18 is the oldest supported version
     steps:
       - name: Checkout
         uses: actions/checkout@v6
@@ -32,20 +34,6 @@ jobs:
           show-progress: false
       - name: Check with clang-format
         uses: RafikFarhad/clang-format-github-action@v4.1
-        with:
-          sources: "Source/*/*.c,Source/*/*.cxx,Source/*/*.h"
-          style: file
-
-  clang-format-newest:
-    runs-on: ubuntu-latest
-    name: clang-format 21
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v6.0.1
-        with:
-          show-progress: false
-      - name: Check with clang-format
-        uses: RafikFarhad/clang-format-github-action@v6.0.1
         with:
           sources: "Source/*/*.c,Source/*/*.cxx,Source/*/*.h"
           style: file
@@ -62,7 +50,7 @@ jobs:
     # can define the additional "desc" field with any additional information to include in the name.
     name: ${{ matrix.SWIGLANG || 'none' }}${{ matrix.PY2 }} ${{ matrix.ENGINE}} ${{ matrix.VER }} ${{ matrix.PY_ABI_VER && format('abi={0}', matrix.PY_ABI_VER) }} ${{ matrix.SWIG_FEATURES }} ${{ (matrix.compiler || 'gcc') }}${{ matrix.GCC }} ${{ matrix.CPPSTD }} ${{ matrix.CSTD }} ${{ matrix.desc }} ${{ matrix.continue-on-error && '(can fail)' }}
 
-    needs: [clang-format-oldest, clang-format-newest]
+    needs: [clang-format]
 
     strategy:
       matrix:
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -21,6 +21,35 @@ permissions:
   contents: read
 
 jobs:
+  # Code format checks, see https://github.com/marketplace/actions/clang-format-github-action
+  clang-format-oldest:
+    runs-on: ubuntu-latest
+    name: clang-format 18
+    steps:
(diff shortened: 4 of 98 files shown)


Source/.clang-format
```diff
@@ -0,0 +1,23 @@
...
+ConstructorInitializerIndentWidth: 2
+ContinuationIndentWidth: 2
+DerivePointerAlignment: false
+IndentCaseBlocks: true
+IndentCaseLabels: false
+IndentGotoLabels: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+PackConstructorInitializers: CurrentLine
+PointerAlignment: Right
+SortIncludes: Never
```
vadz, line 26: I'd enable this, IMHO having includes in alphabetical order is nice because it makes it much easier to see at a glance if the file includes some header or not.

Source/Modules/xml.cxx
```diff
@@ -96,52 +95,47 @@ class XML:public Language {
...
       k = ki.key;
-      if ((Cmp(k, "nodeType") == 0)
-	  || (Cmp(k, "firstChild") == 0)
-	  || (Cmp(k, "lastChild") == 0)
-	  || (Cmp(k, "parentNode") == 0)
-	  || (Cmp(k, "nextSibling") == 0)
-	  || (Cmp(k, "previousSibling") == 0)
-	  || (*(Char(k)) == '$')) {
-	/* Do nothing */
+      if ((Cmp(k, "nodeType") == 0) || (Cmp(k, "firstChild") == 0) || (Cmp(k, "lastChild") == 0) || (Cmp(k, "parentNode") == 0) ||
```
vadz, line 98: This is just a random example of code which I find so much _less_ readable after reformatting...

Source/.clang-format
```diff
@@ -0,0 +1,23 @@
...
+AlignTrailingComments: true
+AccessModifierOffset: -2
+AlignConsecutiveMacros: AcrossEmptyLinesAndComments
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakBeforeMultilineStrings: false
+BinPackArguments: false
+BreakConstructorInitializers: AfterColon
+ColumnLimit: 160
```
vadz, line 16: Isn't this a bit excessive? 120, or maybe 130, columns should really be enough. Especially considering that clang-format *will* produce lines of 160 characters.

vadz (commented):
I won't discuss the decision to do this as I assume it has been already taken, but I'd just like to say that we absolutely need a CI job verifying that all PRs conform to the new format, i.e. running `clang-format --dry-run`. FWIW I have a script I use in some other projects doing this, but it might be a bit more flexible than needed, as it allows specifying extra per-file formatting options. Please let me know if you'd like me to adapt it for SWIG use.

Source/.clang-format
ojwb: I'd say it's much too wide.

My ideal is a width such that I can have 2 windows side-by-side with readable text size which makes tools like vimdiff much more usable as the whole line length on each side can be seen without wrapping or truncation.  I find that's a struggle for lines much over 80 characters.  I also find it handy to be able to edit a file and its corresponding header side by side.

SWIG code is already a bit of a pain in this regard in places, but if clang-format is going to aggressively unwrap lines which are currently wrapped but would fit in a single 160 character line that will make such tools much less usable.

ojwb:
> I thought I'd put in this merge request as a heads up to see if anyone with experience of clang-format would like to offer some guidance and improvements before I land it in master.

I've investigated using clang-format for Xapian in the past.  A lot of it seems good (especially that it reuses a C/C++ compiler's parser which should the sort of problems which the `beautify-file` target suffers from for C++), and there's a lot of configurability.  However I also found it frustrating to work with, and ultimately we haven't ended up using it (yet at least) because we never managed to resolve some blocking issues with using it.

One problem was that the interaction between some related options seemed to be unobvious and I found it hard to find the right combination of such options to achieve a particular result (a downside of a lot of configurability).

A more significant problem was that in some cases it resulted in horribly unreadable reformatted code.  I remember reading complaints about this from other people too - someone summed it up well with something along the lines of "it can reformat code in a way no human ever would".  I suspect some configuration options combinations suffer from this problem much worse than others so perhaps SWIG will suffer less from this.

I don't think we need to exactly match the current formatting (frankly it's pretty poor in many places, with variable adherence to the coding conventions we do have) but I also don't want to be fighting a tool which keeps insisting on reformatting code in ways that make it harder to read.  While `beautify-file` also does stupid reformatting, we don't force it to run on every commit.

wsfulton:
> I won't discuss the decision to do this as I assume it has been already taken, but I'd just like to say that we absolutely need a CI job verifying that all PRs conform to the new format, i.e. running `clang-format --dry-run`. FWIW I have a script I use in some other projects doing this, but it might be a bit more flexible than needed, as it allows specifying extra per-file formatting options. Please let me know if you'd like me to adapt it for SWIG use.

Absolutely, sorry if it wasn't clear, but adding checks that are always run during CI and locally are vital to the success of keeping the code correctly formatted and for helping developers get new code correct before creating pull requests. My next steps were to figure out how to do this best. Any experience in successful workflows will be very helpful. My half-baked thoughts were:

1. If configure finds clang-format, it will add a clang-format to the build of the c and c++ source files and fail or warn if the format is incorrect. As clang-format would be optional, we'd need to encourage developers to install it to help them.
2. Add a Make variable so that a 'make' will also autocorrect with clang-format during builds as another easy to use way to keep modified code correctly formatted. Maybe by simply adding `--Werror` to clang-format if it is always invoked during a build.
3. Document/encourage use of editor auto-correct options when saving (not sure if this will be successful given parser.y doesn't autoformat properly).
4. Modify 'make beautify' to autoformat.

The above are all optional. But a mandatory CI check would be required.

@vadz with regards to your script, I don't envisage per-file options in SWIG as being useful or necessary as the formatting and consistency that clang-format provides across the code base looks fine to me. Maybe I've misunderstood what it provides though. Could you elaborate on what it offers that a 'clang-format -i file` doesn't do. BTW, I'm not keen on per-file variations as a single `Source/.clang-format` file allows many tools that support clang-format easily out of the box.

wsfulton:
BTW, I experimented with the git integration https://clang.llvm.org/docs/ClangFormat.html#git-integration but didn't get that to work. I don't think it is ready for prime time yet, which is a shame as in theory it would be good.

Source/.clang-format
wsfulton: I was trying to also minimise the number of changes/overrides to the Google format. I was quite amazed that formatting the code modified the include order to be honest as this changes the code, it is very definitely not just a code formatting option and other projects I work on have a very definite set of rules as to the order of the #includes and alphabetical ordering would break the code. Anyway, I don't fancy getting distracted by weird #include order bugs for now, but possibly at a later date we can switch this on. Having said that, SWIG hardly has any includes.

Source/Modules/xml.cxx
wsfulton: Agreed. In fact, I'd say this is clang-format's one thing that I find more frustrating. I experimented, but couldn't find a good way to get these if statements format as we'd previously had them. There is one such example that I really objected and is the only case of using `//clang-format off`, see the end of e5bf1c89105a6e3abeded55f5d3262020c4e1465. The if statement was just too long in this case to accept clang-format's squashed up set of expressions. If you find other if statement formatting truly objectionable, we can turn off the formatting for a few more, let me know. I think it is a balancing act, there are a few downsides, but mostly upside by using clang-format.

Source/.clang-format
wsfulton: SWIG's width has formally been 160 for 20 years since the beautify code was added. Opinion will of course vary with those thinking 80 is an objectionable relic from 1970s terminals to those thinking they can't stand anything over 80 wide. There'll never be consensus, so I'm going to just leave it as it has been for 20 years.

wsfulton:
> One problem was that the interaction between some related options seemed to be unobvious and I found it hard to find the right combination of such options to achieve a particular result (a downside of a lot of configurability).
> 

Yup, I found this too. I spent ages configuring it, but feel that for the SWIG code base the options are actually quite good, with just one failure (the if statement formatting - if only it would put one expression per line once a line is too long ... frustrating as it manages to do this for function calls with many arguments). Anyway, bigger picture is it is not such a big deal and one can resort to `// clang-format off` occasionally.

> A more significant problem was that in some cases it resulted in horribly unreadable reformatted code. I remember reading complaints about this from other people too - someone summed it up well with something along the lines of "it can reformat code in a way no human ever would". I suspect some configuration options combinations suffer from this problem much worse than others so perhaps SWIG will suffer less from this.
> 

I think we're lucky with our code base. Maybe because it is simple C like C++ code.

Source/.clang-format
ojwb: That's missing my real point here though.

While SWIG has allowed > 80 columns for some years now, whether to take advantage of that has been down to individual developers, which seems OK - sometimes a long line is more readable, sometimes wrapping is more readable.  Different developers will disagree on the details, but I'd expect there to be at least approximately aligned agreement.

However, if switching to enforced auto-reformatting with clang-format will unwrap any currently wrapped line which fits within the allowed number of columns, we'll have many more very long lines.  That really is not "just leave it as it is".

vadz:
> My next steps were to figure out how to do this best. Any experience in successful workflows will be very helpful.

I do have experience with a couple of projects using clang-format and a few more using formatters for the other languages (ruff, rustfmt, ...) and I think it's by far the best to keep it as simple as possible, i.e.

1. Have a separate CI job just running clang-format.
2. Have a simple way of doing it locally, ideally without using make because some people will use CMake or whatever, so maybe just a trivial `Tools/format.sh`.
3. Have explanations about how to set up clang-format locally in `README-DEV.md` or similar, including how to set up your IDE/editor to do it automatically (e.g. I use Vim autocommand running it on save for the projects that need it).

> My half-baked thoughts were:
> 
>     1. If configure finds clang-format

I would absolutely not do this, using clang-format must not be optional and should not depend on configure.

>     2. Add a Make variable so that a 'make' will also autocorrect with clang-format during builds as another easy to use way to keep modified code correctly formatted.

I would also not do this because it again shouldn't be optional and also because you may want to format the code which doesn't even compile (yet).

>     3. Document/encourage use of editor auto-correct options when saving

Yes.

> (not sure if this will be successful given parser.y doesn't autoformat properly).

So it should be excluded from auto formatting, right?
 
>     4. Modify 'make beautify' to autoformat.

As mentioned above, I think a separate script would be preferable, you don't gain anything by using make for this.
 
> @vadz with regards to your script, I don't envisage per-file options in SWIG as being useful or necessary

If you don't need this, then the script is not needed. I had to do it to make exceptions for some files that looked significantly worse with the default options in that project and also to exclude some files from formatting (useful if you run clang-format with `**/*.cpp`, but not so much if the only file you want to exclude from formatting is `parser.y`).

wsfulton:
Hi @vadz, I've pushed a change which tries to cover your suggestions.

- I've put in two separate CI jobs and started created a pipeline so that the two clang-format X builds are run prior to the linux build matrix, so now you see the beginnings of a pipeline:

image


- I've updated the docs from the previous version for GNU indent, you can view it here for now (scroll down to _Coding style guidelines_ section): https://htmlpreview.github.io/?https://github.com/swig/swig/blob/clang-format-script/Doc/Manual/Extending.html. I've referenced the standard clang-format docs for setting up editors, but if there is a better site for setting up editors, let me know.

- I'm going for supporting clang-format-18 minimum as that is the first version that supports .clang-format-ignore and the ignore feature is really great and simplifies things, such as not having to explicitly specify each file that is allowed to be clang-formatted. One of the things that concerns me is new versions of clang introducing different formatting. The first commits sort this out for versions 17 and 19 (I was originally thinking of supporting 14 and later). With version 21 also available, I had to make one more tweak: de623f46113fec5345cd1b715c087a7cd69eb6bb. So CI testing with first supported version (18) and latest (21) seems to be sensible options.

- We can't tell what future changes clang-format will introduce and we cannot have builds breaking if a newer version is used when building SWIG, so making it mandatory during a build is going to be very annoying for users who just want to build SWIG. So while making clang-format mandatory for development in the SWIG GHA pipelines is necessary, I have put in a set of flexible default options for non-developers, see the html docs mentioned above. Hopefully all use cases are covered and documented and gets the balance right of being mandatory for development and optional for non-development.

- autoconf + make is the primary build and development system, so hence it is supported as first class and is much more flexible than invoking clang-format manually or via a script. I can add in a one liner script or just document it for someone not using make and would rather invoke clang manually:
```bash
clang-format -i  */*.c */*.cxx */*.h
```
@vadz, if you are using cmake, is there not a way to add it into the build akin to make?

vadz:
Thanks for your work on this but I still don't understand why running clang-format should be related to any build system, be it configure, CMake, meson or whatever. Running `clang-format -i  */*.c */*.cxx */*.h` is really simple and doesn't require anything, so what's wrong with just doing it? IOW what benefits does integration with configure have? I just don't see any at all.

FWIW I'd also document adding the command to run clang-format to Git commit hook, this is very helpful to avoid accidentally committing non-formatted changes even locally (and, again, completely independent of configure — but I think I've already made this point).

Speaking of commit hooks, I am not sure it's worth having a pipeline for this, there is no real dependency between formatting and building and it just adds unnecessary (IMHO) complexity to the CI.

Concerning different versions of clang-format: I'd definitely stick to a single version, whichever it is, as other versions can and do format code differently and nobody wants to deal with this. So just document that SWIG must be formatted with clang-format-18 and provide download links for its binaries for the common platforms (see e.g. https://github.com/muttleyxd/clang-tools-static-binaries/releases).

Finally, I'd still very much like to ask to reconsider using 160 as line length, this is excessive and it would be better to use something more reasonable, like 120 or 132, perhaps.

wsfulton:
Hiya @vadz, 

> Thanks for your work on this but I still don't understand why running clang-format should be related to any build system, be it configure, CMake, meson or whatever. Running `clang-format -i */*.c */*.cxx */*.h` is really simple and doesn't require anything, so what's wrong with just doing it? IOW what benefits does integration with configure have? I just don't see any at all.

To me it seems the best solution. A kind of shift-left if you like. The formatting is part of the edit, save, compile, test cycle and the earlier one sees problems the less you need to repeat later in the cycle. To me the clang-format warnings are akin to the occasional code formatting warnings that the compiler emits, such as `-Wmisleading-indentation`. So, seeing additional formatting warnings along with compiler warnings seems entirely the right time to see them and deal with them. The benefit I see is a faster cycle - manually remembering to invoke a separate tool later in the cycle is error prone and in the wrong order (for those that havn't configured editors to autoformat on save). It might not be your preferred way of working and that's okay. The documented options enable both choices.

> FWIW I'd also document adding the command to run clang-format to Git commit hook, this is very helpful to avoid accidentally committing non-formatted changes even locally (and, again, completely independent of configure — but I think I've already made this point).

I really like the idea, but I couldn't get it to work, I refer you to my earlier comment - https://github.com/swig/swig/pull/3312#issuecomment-3758761086. Maybe just me and I could try again if there is a better guide to go in the docs, but given the difficulty, it should be one of the various optional choices.

> Speaking of commit hooks, I am not sure it's worth having a pipeline for this, there is no real dependency between formatting and building and it just adds unnecessary (IMHO) complexity to the CI.

Seems quite simple to me, but I'm just a bit confused as to what exactly you're looking for as earlier you mentioned _using clang-format must not be optional_ and _Have a separate CI job just running clang-format._. What I've put in place goes a long way to 'must not be optional' and it does run in a separate CI job. It is mandatory by failing the build and **preventing** the running of the majority of the tests (linux builds). This is deliberate as it prevents unnecessary CI repeated builds for additional formatting commits - I'm getting emails from Github about using too much compute on GHA.

> Concerning different versions of clang-format: I'd definitely stick to a single version, whichever it is, as other versions can and do format code differently and nobody wants to deal with this. So just document that SWIG must be formatted with clang-format-18 and provide download links for its binaries for the common platforms (see e.g. https://github.com/muttleyxd/clang-tools-static-binaries/releases).

Can we not just try supporting the proposed small range of versions by testing for them in CI - more flexible and developer friendly right? My experience showed that it really is not a big deal supporting them all. Trying to install a single mandated version is going to be a source of great frustration, so being flexible with a choice is good as we want to make it easy for developers to install clang-format. It's very similar to a compiler - I might chose to develop with a particular version of gcc and CI tests with a huge range. Usually this results in no problems with just the odd exception. Also, I have already and would rather continue to install from a trusted, secure source over any other source of binaries and those trusted sources don't always provide a full range. I may also need to meet the needs of different projects, so flexibility is key.

The feedback is interesting for an insight into a to using clang-format, and taking a step back, is there any blocker in the current pull request preventing you from using clang-format how you're used to working with it? I'm kinda struggling to get it absolutely perfect for differing views and it may need to evolve.

> Finally, I'd still very much like to ask to reconsider using 160 as line length, this is excessive and it would be better to use something more reasonable, like 120 or 132, perhaps.

Okay, I'll play with that next.

vadz:
> Hiya @vadz,
> 
> > Thanks for your work on this but I still don't understand why running clang-format should be related to any build system, be it configure, CMake, meson or whatever. Running `clang-format -i */*.c */*.cxx */*.h` is really simple and doesn't require anything, so what's wrong with just doing it? IOW what benefits does integration with configure have? I just don't see any at all.
> 
> To me it seems the best solution.

I usually work on projects that build under more than one platform and use more than one build system and maybe this explains why I had never used it like this. And while SWIG is _mostly_ just using configure/make to build, I still think it's not crazy to think of people who use CMake, MSBuild or whatever else to build it. Of course, all of them could redo what you did for configure but IMNSHO it's clearly not worth it. Moreover, even though I do use make myself, I'd definitely want to specifically avoid running clang-format automatically as part of the build for many reasons, one of which is that I *never* want the build system to modify any source files.

> A kind of shift-left if you like. The formatting is part of the edit, save, compile, test cycle and the earlier one sees problems the less you need to repeat later in the cycle.

Yes, it's basically part of "Edit" and IMO it should remain there.

[speaking of Git hook]
> I really like the idea, but I couldn't get it to work

Sorry, I'm not sure what the problem is. I just have something like this in a section of my `.git/hooks/pre-commit`:

```sh
files_to_format=$(git diff-index --cached --diff-filter=d --name-only HEAD ':**/*.cpp' ':**/*.h')
if [ -n "$files_to_format" ]; then
    echo "Checking formatting of ${files_to_format}..."
    if ! clang-format --Werror --dry-run ${files_to_format}; then
        printf "Error: Fix formatting before committing.\n"
        exit 1
    fi
fi
```

Why doesn't this work for SWIG?

> > Speaking of commit hooks, I am not sure it's worth having a pipeline for this, there is no real dependency between formatting and building and it just adds unnecessary (IMHO) complexity to the CI.
> 
> Seems quite simple to me, but I'm just a bit confused as to what exactly you're looking for as earlier you mentioned _using clang-format must not be optional_ and _Have a separate CI job just running clang-format._.

Sorry for being unclear, for me just having a separate CI job running it was enough as failure of any CI jobs should prevent merging (i.e. make this "not optional").

> What I've put in place goes a long way to 'must not be optional' and it does run in a separate CI job. It is mandatory by failing the build and **preventing** the running of the majority of the tests (linux builds). This is deliberate as it prevents unnecessary CI repeated builds for additional formatting commits - I'm getting emails from Github about using too much compute on GHA.

I see, this would definitely help with this.
 
> Can we not just try supporting the proposed small range of versions by testing for them in CI - more flexible and developer friendly right? My experience showed that it really is not a big deal supporting them all. 

Maybe they improved their compatibility story then. Back when I was putting it in place (~clang 14 time I think) I tried 3 versions which all gave different outputs.

So, of course, if there are no problems with supporting multiple versions, all the better, but I wouldn't waste any time on it if you can save it by just requiring a specific one.

> Trying to install a single mandated version is going to be a source of great frustration, so being flexible with a choice is good as we want to make it easy for developers to install clang-format. It's very similar to a compiler

It's much, much simpler — it's a single binary without any dependencies.

> The feedback is interesting for an insight into a to using clang-format, and taking a step back, is there any blocker in the current pull request preventing you from using clang-format how you're used to working with it? I'm kinda struggling to get it absolutely perfect for differing views and it may need to evolve.

I haven't tried using it yet, sorry. I hope to upgrade my SWIG branch to latest master soon (maybe next week) and will probably try it too then.

wsfulton:
> ... I'd definitely want to specifically avoid running clang-format automatically as part of the build for many reasons, one of which is that I _never_ want the build system to modify any source files.

It doesn't. 

> I haven't tried using it yet, sorry. I hope to upgrade my SWIG branch to latest master soon (maybe next week) and will probably try it too then.

I think you'll find you'll get what you want by default or with a simple configure, please read the docs! The branch is not up to date to master yet and I will probably be rebasing it before merging to master, you may need to experiment in a separate git worktree.

> Maybe they improved their compatibility story then. Back when I was putting it in place (~clang 14 time I think) I tried 3 versions which all gave different outputs.
> 
> So, of course, if there are no problems with supporting multiple versions, all the better, but I wouldn't waste any time on it if you can save it by just requiring a specific one.

I had to put in 3 simple workarounds to keep the formatting the same for version 14 to 21. That was 3 simple changes for 7 years of clang-format development. However, I also noticed like you that prior to 14, different versions generated various differences. So it seems that clang-format has reached a good point of stability since 14. I'm proposing version 18+ for now.

wsfulton:
Your pre-commit hook works very well and so I've tweaked it for SWIG and added it and referenced it in the docs. Thanks!

Alas, the git integration docs https://clang.llvm.org/docs/ClangFormat.html#git-integration don't mention this approach.

wsfulton:
I tried 120, but to be perfectly frank, I don't like it, there are lots of less readable changes and I got quite annoyed looking at how 120 has screwed up the code format.

It's not even possible to declare some 3 argument methods without splitting in places such as having this enforced change:
```diff
-void PyDocConverter::handleDoxyHtmlTagNoParam(DoxygenEntity &tag, std::string &translatedComment, const std::string &arg) {
+void PyDocConverter::handleDoxyHtmlTagNoParam(DoxygenEntity &tag, std::string &translatedComment,
+                                              const std::string &arg) {
```
Over the last 20 years I've written code and comments aiming to fit in with 160 columns and a lot of that work is now thrown away with daft line breaks.

With < 160 columns, I can see less on the screen, modern monitors are much wider than they are high, so one can see more on the screen with wider lines. Coding is easier seeing more on the screen at once, so I just really struggle to understand why 160 is not better when monitors can easily see over 200 characters per line. It's like someone has modified my editor's settings to uglify the code with a force wrap. I appreciate that this is just my perspective, but there it is and why I'd prefer to stay with the status quo.

With the given column widths, there'd be additional lines:
```
80: 88 files changed, 17988 insertions(+), 7136 deletions(-)
120: 60 files changed, 3862 insertions(+), 1474 deletions(-)
132: 54 files changed, 2329 insertions(+), 977 deletions(-)
```
I can adapt, but the request to change this is hardly a ringing endorsement with two votes for changing a 20 year format that has not been challenged before. If anyone else feels really strongly about any width, speak up here. I suspect most people working on the SWIG code base don't care too much though.

christophe-calmejane:
My 2 cents: 

Just like indentation "size", it's something that should be left to each person to choose from.
That's why in my company we force:
- use of tabulation char (not space) for indentation, so every dev is free to define its own width count as it pleases without impacting git commits
- unlimited chars per line, use IDE line wrapping (automatically adapts to windows size, monitor size, ...)

These are minors things (visual) but IDE should be the ones handling this kind of things, not your commits.

On a side note, I'm fully in favor of having a clang-format for this projet. Due to a lot of formatting issues with my different IDEs (that completely messed up the files) when working on pull requests, I had to create my own .clang-format file a few months ago, so I don't have to think about how to write the code, and let the IDE reformat as the project likes it.
Thanks a lot for bringing a single source of truth for the file formatting.

Issue #3313, opened 2026-01-16
python gcc CI tests are failing


wsfulton:
I've disabled this test from running in 097fb16e89e458708889a5285a373cf92ee8c709. @erezgeva do you have some time to fix this seems to be due to a switch to depending on the ucrt? Noting your comments:

https://github.com/swig/swig/blame/097fb16e89e458708889a5285a373cf92ee8c709/.github/workflows/windows.yml#L148-L157

erezgeva:
Yes, using UCRT does solve the issue.

erezgeva:
Looking on 
https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md

I see:
Python
- 3.10.11
- 3.11.9
- 3.12.10
- 3.13.11
- 3.14.2

Python 3.9 was indeed removed.
Python 3.14 should be available 😄

PR #3314, opened 2026-01-16
Fix Python GCC windows CI test

erezgeva:
As Python GCC windows CI fail.
Use UCRT, to fix see  #3313


erezgeva:
Hi @wsfulton

This fix #3313

PR #3318, opened 2026-01-22
python: AsLong: Use PyLong_AsLong unconditionally on Python3.10+.

degasus:
Since Python 3.10, PyLong_AsLong does not use `__int__()` any more. So it shall fail when the input is not supposed to be an integer.

This fixes "casting" to integer for types declared as integers, e.g. numpy.int32.

Please see https://peps.python.org/pep-0357/ for `__index__()`.

Also switch to the ..AndOverflow method as this allows us to easily decide if we have an overflow or an type error.

Same for `long long`.

Not the same for `unsigned long`, as the PyLong API only supports the *Mask calls. They always truncate and never report for overflow. Kind of makes sense IMO, but this would be a large behavior change.

Note:
This does not affect the cast_mode on python. float->int still requires the cast mode. This only supports implicit "casting" of integer-alike variables to C-integers.

Fixes implicit "casting" of numpy.int32 to int.
Does not fix the "casting" of numpy.uint32 to unsigned int.

### DIFF
--- a/Lib/python/pyprimtypes.swg
+++ b/Lib/python/pyprimtypes.swg
@@ -81,18 +81,24 @@ SWIG_AsVal_dec(long)(PyObject *obj, long* val)
     return SWIG_OK;
   } else
 %#endif
-  if (PyLong_Check(obj)) {
-    long v = PyLong_AsLong(obj);
-    if (!PyErr_Occurred()) {
+%#if PY_VERSION_HEX < 0x030a0000
+  if (PyLong_Check(obj))
+%#endif
+  {
+    int overflow;
+    long v = PyLong_AsLongAndOverflow(obj, &overflow);
+    if (overflow != 0) {
+      return SWIG_OverflowError;
+    } else if (!PyErr_Occurred()) {
       if (val) *val = v;
       return SWIG_OK;
     } else {
       PyErr_Clear();
-      return SWIG_OverflowError;
     }
   }
 %#ifdef SWIG_PYTHON_CAST_MODE
   {
+%#if PY_VERSION_HEX < 0x030a0000
     int dispatch = 0;
     long v = PyInt_AsLong(obj);
     if (!PyErr_Occurred()) {
@@ -101,7 +107,9 @@ SWIG_AsVal_dec(long)(PyObject *obj, long* val)
     } else {
       PyErr_Clear();
     }
-    if (!dispatch) {
+    if (!dispatch)
+%#endif
+    {
       double d;
       int res = SWIG_AddCast(SWIG_AsVal(double)(obj,&d));
       // Largest double not larger than LONG_MAX (not portably calculated easily)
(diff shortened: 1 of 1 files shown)


jschueller:
the testsuite needs to cover this

wsfulton:
Hi @degasus, we need a motivating example to consider this pull request, ideally a test added as already suggested.

degasus:
My motivation can be summerized by this example:

AsLong.h
```C++
#include <numeric>
#include <vector>

template <typename T> T my_sum(const std::vector<T> &values) {
  return std::accumulate(values.begin(), values.end(), T(0));
}
```

AsLong.i
```
%module AsLong
%include "std_vector.i"

%{
  #include "AsLong.h"
%}

%include "AsLong.h"

%define DECLARE_TEMPLATE(TYPE, NAME)
%template(NAME ## Vector) std::vector<TYPE>;
%template(NAME ## Sum) my_sum<TYPE>;
%enddef

DECLARE_TEMPLATE(char, Char)
DECLARE_TEMPLATE(signed char, SChar)
DECLARE_TEMPLATE(unsigned char, UChar)
DECLARE_TEMPLATE(short, Short)
DECLARE_TEMPLATE(unsigned short, UShort)
DECLARE_TEMPLATE(int, Int)
DECLARE_TEMPLATE(unsigned int, UInt)
DECLARE_TEMPLATE(long, Long)
DECLARE_TEMPLATE(unsigned long, ULong)
DECLARE_TEMPLATE(long long, LongLong)
DECLARE_TEMPLATE(unsigned long long, ULongLong)
```

test_AsLong.py
```Python
import numpy as np
from AsLong import *

for my_range_helper in (range, np.arange):
    print("testing with the helper", my_range_helper)
    assert SCharSum(my_range_helper(5)) == 10
    assert UCharSum(my_range_helper(5)) == 10
    assert ShortSum(my_range_helper(5)) == 10
    assert UShortSum(my_range_helper(5)) == 10
    assert IntSum(my_range_helper(5)) == 10
    assert UIntSum(my_range_helper(5)) == 10
    assert LongSum(my_range_helper(5)) == 10
    assert ULongSum(my_range_helper(5)) == 10
    assert LongLongSum(my_range_helper(5)) == 10
    assert ULongLongSum(my_range_helper(5)) == 10
```

This patch fixes the handling of the signed versions of the numpy checks.

For the swig test suite, do you want me to include numpy, or shall I create a custom type with the property `__index__`?

I think we can/should also discuss if a custum type with the property `__index__` shall be implicitly "casted" to integer (without castmode).

For the real fix, we should also decide how to handle unsigned types. We could use the same code and check for the signed boundaries first, loosing half of the range. Or we could switch to the Python's way and just always truncate the upper bits for unsigned types.

wsfulton:
Hi @degasus, depending on numpy is overly complicated and your suggestion of providing a type with `__index__` defined would be much better: just add some comments in the definition that it is a simplified numpy type. Adding your motivating example, with a simplified numpy type, into the test-suite would be perfect. Use swig_assert instead of assert please.

> This patch fixes the handling of the signed versions of the numpy checks.

Your test shows both signed and unsigned C types being checked. Are you saying that if a signed number, say -10 is used instead of 10 in your example, there is more work to do? Sorry, am just a bit confused.

> I think we can/should also discuss if a custum type with the property `__index__` shall be implicitly "casted" to integer (without castmode).

Oh blimey! We have %implicitconv to think about too. The approach generally is to add a feature to add in various implicit castings. Having said that, your change unconditionally uses `PyLong_AsLongAndOverflow` and so if the passed in type has an `__index__`, it will be implicitly cast to a number as per the docs of https://docs.python.org/3/c-api/long.html#c.PyLong_AsLongAndOverflow.

wsfulton:
Hi @degasus, will you get a chance to address my comments?

PR #3322, opened 2026-01-29
Support using range-based for loops for iterating over wrapped vectors

vadz:
I'd like to submit some enhancements that accumulated in my branch, starting with these ones, making using `vector` from C++ more ergonomic.

I'm not happy about having to use `SWIG_STD_VECTOR_ITERATOR()` but I don't see how can I avoid it, i.e. how to make `%template(foos) std::vector<foo>` to also have the effect of `%template(foos_const_iter) swig::vector_citr<foo>`. Any ideas would be welcome!

### DIFF
--- a/Examples/test-suite/c/li_std_vector_runme.cxx
+++ b/Examples/test-suite/c/li_std_vector_runme.cxx
@@ -33,9 +33,25 @@ void test_int_vector_cxx() {
     assert( sum_modulos == 1 );
 }
 
+void test_bool_vector_cxx() {
+    li_std_vector::BoolVector bv;
+    for ( int n = 0; n < 3; ++n )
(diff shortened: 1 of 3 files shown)


vadz:
This is completely unrelated to this PR but something weird has happened to Python in the python/gcc Windows job, as it now outputs:

```
checking for Python 3.x version... string
...
checking for Python 3.x library... -lstring
```

and then later fails to link due to `-lstring` not being able to find the "string" library.

Previous successful build had

```
checking for Python 3.x library... -lpython3.13
```

so something must have changed in the build image but I have no idea what. I also have no idea why does SWIG use such a baroque way of getting Python version in the first place...

P.S. I reran the failing job and it still failed in the same way.

PR #3324, opened 2026-02-01
More fixes for GitHub windows CI

erezgeva:
- Add dot after the `VER` variable, in search paths, to avoid fetching the wrong version. i.e. prevent `VER=3.1` to use version `3.10`.
- Ruby installations on GitHub Windows image, never use the MSVC compiler.
- There is a default Java. The currently default Java version is 17. With proper `JAVA_HOME` variable in the environment.
- Fix `PY3VERSION`: old `string.__file__`ends with `string.py`, while new ends with `string/__init__.py`. Using the file path requires more filtering, and might prone to more changes in the future. Use `sys.version_info` to fetch proper python version. This method might bent too, but is less likely to change.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -157,7 +157,7 @@ jobs:
               mingw_variant=mingw-w64-ucrt-x86_64
               MORE_MSYS_PKGS+=" $mingw_variant-gcc"
               if [[ -n "$VER" ]]; then
-                ruby_dir=$(ls -d /c/hostedtoolcache/windows/Ruby/$VER*)/x64/bin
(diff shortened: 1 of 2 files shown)


erezgeva:
@wsfulton 
The failure of python using GCC on windows CI is due to wrong value of `PY3VERSION`.
The error exist also on new versions of python on Linux as well.
The different is that Linux CI do not use the value of  `PY3VERSION` for linking of testing, while  python using GCC on windows does rely.

wsfulton:
Thanks, merged. I was a bit confused as to what it was fixing as master didn't have any failures, but there are failures in more recent merge requests due to recent Github Actions image updates.

PR #3325, opened 2026-02-03
runtime: Insert include guard

jschueller:
Allows to merge source files with several inclusions of the runtime header into a single TU.

### DIFF
--- a/Source/Modules/main.cxx
+++ b/Source/Modules/main.cxx
@@ -379,6 +379,11 @@ static void SWIG_dump_runtime() {
   Swig_banner(runtime);
   Printf(runtime, "\n");
 
+  /* include guard begin */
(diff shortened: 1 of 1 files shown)


Source/Modules/main.cxx
```diff
@@ -379,6 +379,11 @@ static void SWIG_dump_runtime() {
   Swig_banner(runtime);
   Printf(runtime, "\n");
 
+  /* include guard begin */
+  Printf(runtime, "#ifndef __SWIG_RUNTIME_HEADER__\n");
```
wsfulton:
```suggestion
  Printf(runtime, "#ifndef SWIG_RUNTIME_HEADER_\n");
```
Or similar to use standard compliant symbols.

jschueller: done

PR #3326, opened 2026-02-04
Ruby: Avoid usage of deprecated untyped API

### LINKED ISSUE #3170: [Ruby] The old (non-Typed) Data_XXX macro family has been deprecated.
It seems that using `Data_Wrap_Struct(...)` as described [here](https://www.swig.org/Doc4.3/SWIGDocumentation.html#Ruby_nn51) and as used in `rubyrun.swg` is deprecated.

Described here: https://github.com/ruby/ruby/blob/master/doc/extension.rdoc#label-Encapsulate+C+data+into+a+Ruby+object

I am getting a bunch of deprecation warnings on fedora 42 for example by building the ruby [class](https://github.com/swig/swig/tree/master/Examples/ruby/class) example using **clang++** with: `ruby-3.4.1-22.fc42.x86_64`

I am using the current upstream swig repo.

I believe it has been deprecated for quite a while now so I am not sure why it showed up only now and only with clang.
Also when I downgrade ruby to `ruby-3.3.8-19.fc41.x86_64` the warning goes away.

larskanis:
This is only a starting point for now. Some examples already work, some don't.

My use case is https://github.com/larskanis/fxruby which doesn't work at the current state.

I know the ruby internals quite well, but have trouble to understand the swig internals. It rewrites the mark and free pointers of object instances in certain cases, which is no longer possible with the new API. So I guess we need static mark and free functions, which jump to the dynamic (per instance) mark/free function.

Fixes #3170

### DIFF
--- a/Lib/ruby/rubyhead.swg
+++ b/Lib/ruby/rubyhead.swg
@@ -112,7 +112,15 @@
 #ifndef RTYPEDDATA_P
 # define RTYPEDDATA_P(x) (TYPE(x) != T_DATA)
 #endif
+#ifndef TYPED_DATA_EMBEDDED
+# define RTYPEDDATA_GET_DATA(x) RTYPEDDATA_DATA(x)
+#endif
 
+struct ruby_wrapped_object {
+  void *data;
+  void (*dmark)(void *);
+  void (*dfree)(void *);
+};
 
 
 /*
--- a/Lib/ruby/rubyrun.swg
+++ b/Lib/ruby/rubyrun.swg
@@ -69,6 +69,7 @@ typedef struct {
   VALUE mImpl;
   void  (*mark)(void *);
   void  (*destroy)(void *);
+  rb_data_type_t cext_type;
   int trackObjects;
 } swig_class;
 
@@ -162,6 +163,27 @@ SWIG_Ruby_define_class(swig_type_info *type)
   free((void *) klass_name);
 }
 
+static void SWIG_Ruby_mark_swig_type(void *_wrobj)
+{
+  struct ruby_wrapped_object *wrobj = (struct ruby_wrapped_object *)_wrobj;
+  if (wrobj->dmark) {
+    wrobj->dmark(wrobj->data);
+  }
+}
+
+static void SWIG_Ruby_free_swig_type(void *_wrobj)
+{
+  struct ruby_wrapped_object *wrobj = (struct ruby_wrapped_object *)_wrobj;
+  if (wrobj->dfree) {
+    wrobj->dfree(wrobj->data);
+  }
+  free(wrobj);
+}
+
+static const rb_data_type_t swig_type_type = {
+    .wrap_struct_name = "swig_type"
+};
+
 /* Create a new pointer object */
 SWIGRUNTIME VALUE
 SWIG_Ruby_NewPointerObj(void *ptr, swig_type_info *type, int flags)
@@ -172,6 +194,7 @@ SWIG_Ruby_NewPointerObj(void *ptr, swig_type_info *type, int flags)
   swig_class *sklass;
   VALUE klass;
   VALUE obj;
+  struct ruby_wrapped_object *wrobj;
 
   if (!ptr)
     return Qnil;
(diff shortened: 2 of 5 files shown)


wsfulton:
All the Ruby tests fail. Are you going to have another go?

larskanis:
The current implementation simulates more or less the old untyped API. This is done by storing the dmark and dfree function pointers in a 24 byte heap memory block. That way static mark and free functions of the new typed API can be used and they jump to the per-object function pointers of the old API.

The downside is that this implementation doesn't follow the assumption that the data pointer of a ruby T_DATA object directly points to the related C++ object. This assumption is made in the one test case that fails (ruby_manual_proxy). According to the documentation it is extracted from the SVN-Ruby binding. So the current state isn't fully backward compatible. Some usages of swig have to be updated to no longer assume that DATA_PTR(Ruby_obj) equals to the C++ object. Also my use case FXRuby needs to be updated in this regard.

I'll check if I find a better way to use the typed API and still follow the above assumption.

larskanis:
This is now ready for review. The patch avoids the old untyped object API of Ruby and replaces it with the typed API. As written earlier there is no way to make the object wrapping 100% compatible with the old API. It's therefore necessary to change direct usages of `DATA_PTR` and `Data_Get_Struct` with `SWIG_ConvertPtr` calls as done in the second commit which changes one related test case. If `DATA_PTR` and `Data_Get_Struct` are avoided or replaced by `SWIG_ConvertPtr`, then the code is compatible with new swig (using typed ruby API) and old swig (using untyped API).

I had to do the same changes to FXRuby here in this PR: https://github.com/larskanis/fxruby/pull/86
With these changes all the deprecation warnings are gone.

There are some further possible changes like #3377, which I would propose, but this PR is a precondition to them.

Lib/ruby/rubyrun.swg
```diff
@@ -162,6 +163,27 @@ SWIG_Ruby_define_class(swig_type_info *type)
...
+  if (wrobj->dfree) {
+    wrobj->dfree(wrobj->data);
+  }
+  free(wrobj);
+}
+
+static const rb_data_type_t swig_type_type = {
+    .wrap_struct_name = "swig_type"
+};
+
```
wsfulton, line 186: Generated code should be C90 syntax.

Lib/ruby/rubyrun.swg
```diff
@@ -162,6 +163,27 @@ SWIG_Ruby_define_class(swig_type_info *type)
   free((void *) klass_name);
 }
 
+static void SWIG_Ruby_mark_swig_type(void *_wrobj)
```
wsfulton, line 166: `_` prefix is stylistically unusual. Suggest `wrapped_obj` instead.

Lib/ruby/rubyhead.swg
```diff
@@ -112,7 +112,15 @@
 #ifndef RTYPEDDATA_P
 # define RTYPEDDATA_P(x) (TYPE(x) != T_DATA)
 #endif
+#ifndef TYPED_DATA_EMBEDDED
+# define RTYPEDDATA_GET_DATA(x) RTYPEDDATA_DATA(x)
+#endif
 
+struct ruby_wrapped_object {
```
wsfulton, line 119: ruby prefixes in the global namespace ought to only be used by the real Ruby project. Suggest `SWIG_ruby_wrapped_object` instead.

Lib/ruby/rubyrun.swg
```diff
@@ -214,7 +238,11 @@ SWIG_Ruby_NewPointerObj(void *ptr, swig_type_info *type, int flags)
     SWIG_snprintf(klass_name, klass_len, "TYPE%s", type->name);
     klass = rb_const_get(_mSWIG, rb_intern(klass_name));
     free((void *) klass_name);
-    obj = Data_Wrap_Struct(klass, 0, 0, ptr);
+
+    obj = TypedData_Make_Struct(klass, struct ruby_wrapped_object, &swig_type_type, wrobj);
```
wsfulton, line 242: `Data_Wrap_Struct` that is being replaced seems to come from an internal Ruby API in ruby/internal/core/rtypeddata.h. `TypedData_Make_Struct` is also in the internal API. Just wondering on the off-chance if there is a documented API that could instead be used?

Lib/ruby/rubyrun.swg
```diff
@@ -398,6 +434,11 @@ SWIG_Ruby_ConvertPacked(VALUE obj, void *ptr, int sz, swig_type_info *ty) {
   return SWIG_ERROR;
 }
 
+
+static const rb_data_type_t swig_runtime_data_type = {
+    .wrap_struct_name = "swig_runtime_data" SWIG_RUNTIME_VERSION SWIG_TYPE_TABLE_NAME
```
wsfulton, line 439: Generated code should be C90 syntax.

wsfulton (changes requested):
Hi @larskanis. We don't have any active Ruby maintainers, so these deprecation fixes are much appreciated. I don't know the Ruby API very well, but it all seems sensible and the tests pass. There are just a few minor cosmetics to address, then I can merge this in.

wsfulton:
The following series of commits completes the migration off the deprecated API:

ef8c5af40ae15ad00d0e5916446c9252bb1c3d97...dbe3f5ec793e328f741745606f84045cbc886693

The first commit extracts large parts from this merge request.

PR #3191, opened 2025-06-09
update PCRE2 find module to be compatible with upstream CMake config script

phetdam:
## Background

SWIG's CMake build config makes it much simpler to build SWIG on Windows assuming an existing Bison and PCRE2 installation. PCRE2 [10.38 and above](https://github.com/PCRE2Project/pcre2/commit/2410fbe3869cab403f02b94caa9ab37ee9f5854b) provides a CMake config script, although only after the 10.45 release did this config script become idiomatic, i.e. using the [CMake config helpers](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html). These allow `find_package` to be used in the "standard" way instead of using find-module logic as provided in `Tools/cmake/FindPCRE2.cmake`, e.g. with:

```cmake
if(WITH_PCRE)
  find_package(PCRE2 10.39 REQUIRED COMPONENTS 8BIT)
  set(HAVE_PCRE 1)
endif()
```

We can also link against the PCRE2 8-bit library target as follows when necessary:

```cmake
if(PCRE2_FOUND)
  target_link_libraries(swig PRIVATE PCRE2::8BIT)
endif()
```

The upstream PCRE2 config script will correctly handle defining `PCRE2_STATIC` during compilation.

## Changes

The changes are mostly to the `Tools/cmake/FindPCRE2.cmake` find module, which have the following logic:

1. Attempt to locate PCRE2 via CMake package config script if possible
2. If failed, use the existing find module logic while considering on Windows differences in static/shared PCRE2
   * For example, Windows PCRE2 static libraries have the name `pcre2-[bitness]-static`
   * [`PCRE2_USE_STATIC_LIBS`](https://github.com/PCRE2Project/pcre2/blob/bf50eeef64fc4f5ddfc93a041e2f4d7357f3c431/cmake/pcre2-config.cmake.in#L17) is respected and if not defined on Windows, implicitly set to `ON`
3. Additional enhancements to more closely mirror the upstream CMake config script's behavior
   * The `PCRE2::8BIT` [`IMPORTED`](https://cmake.org/cmake/help/latest/command/add_library.html#imported-libraries) library target is defined by `FindPCRE2.cmake` if necessary
   * The version (read from `pcre2.h`) and components passed to `find_package` are both checked

## Testing

[...]

PR #3332, opened 2026-02-11
Add MacOS CMake CI and Linux CMake CI

erezgeva:
This PR consist of two commits

1. Derek Huang,  @phetdam from #3191, I clean the PR, rebase it and stash it into a single commit.
2. Add a new MacOS CMake CI, some addition to the new Linux CMake CI and improves of Windows CMake CI.

The PR consist of:
- Rewrite of `FindPCRE2.cmake` by Derek Huang, @phetdam.
- Integration of new FindPCRE2 withing `CMakeLists.txt`
- Update of Windows document of using the CMake.
- Update of existing `windows-cmake.yml`
- Adding `linux-cmake.yml` and `macos-cmake.yml`

### Note:
Just to avoid any misunderstanding. This PR does **not** intend to replace any `autoconf` or add any further testing with `CMake`. The `CMake` is provided to users as an alternative way of building the swig application itself!


.github/workflows/linux-cmake.yml
```diff
@@ -0,0 +1,76 @@
...
+      - 'appveyor.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  cmake:
+    runs-on: ${{ matrix.os }}
+
+    name: CMake ${{ matrix.os }}
```
wsfulton:
```suggestion
    name: cmake ${{ matrix.os }}
```
Let's keep to lower case like everywhere else. You'll need to make this change for all the cmake names.

wsfulton:
With the addition of linux-cmake and macos-cmake, we'll have 6 different workflows per commit. I think this is just too many, especially as each of the cmake workflows only runs two jobs. Let's combine windows-cmake, linux-cmake and macos-cmake into one workflow called cmake, which will go into .github/workflows/cmake.yml. So, let's aim for 4 workflows: linux, windows, macos, cmake.

The simple way to do this is to copy the 3 jobs into the new cmake.yml file as 3 different jobs. However, having 3 separate cmake files would be preferred... it was nice for me to diff the 3 *-cmake.yml files to look at the diffs. So I think we need to investigate a way to include the 3 *-cmake.yml files into the one cmake.yml workflow file. I believe this is called composite actions or reusable workflows which is implemented via the 'uses' keyword. I haven't any experience with this, but would you mind looking at that @erezgeva ?

.github/workflows/linux-cmake.yml
erezgeva: Sure

erezgeva:
> With the addition of linux-cmake and macos-cmake, we'll have 6 different workflows per commit. I think this is just too many, especially as each of the cmake workflows only runs two jobs. Let's combine windows-cmake, linux-cmake and macos-cmake into one workflow called cmake, which will go into .github/workflows/cmake.yml. So, let's aim for 4 workflows: linux, windows, macos, cmake.
> 
> The simple way to do this is to copy the 3 jobs into the new cmake.yml file as 3 different jobs. However, having 3 separate cmake files would be preferred... it was nice for me to diff the 3 *-cmake.yml files to look at the diffs. So I think we need to investigate a way to include the 3 *-cmake.yml files into the one cmake.yml workflow file. I believe this is called composite actions or reusable workflows which is implemented via the 'uses' keyword. I haven't any experience with this, but would you mind looking at that @erezgeva ?

To my understanding, `reuse workflow` is a case in which a workflow uses another workflow. Meaning one workflow from one YML uses another workflow  in another workflow.
https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#creating-a-reusable-workflow

https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action

I tried it once, using `reuse workflow`. And it only seems more work than benefit.
I did not tried composite action yet, but I do not see how it works with mixing bash and powershell.
I find it more easier to multiplex inside bash.

I think on a more simple approach.
I think we can merge `macos-cmake.yml` and `linux-cmake.yml` into `cmake.yml` as both uses bash. 
And add windows-cmake.yml as  a second job named `windows-cmake`, as it uses powershell.

As the new YML will have 2 jobs, it should be easy to compare the 6 jobs very easily.

On it :smile:

erezgeva:
I forgot to change the `[itHub Actions YAML file` URL in `Doc/Manual/Windows.html`
To the new `cmake.yml`.

wsfulton:
The combined cmake.yml file looks fine.

Please also see my comments from today in #3191.

erezgeva:
With the consent of @phetdam, I removed https://github.com/swig/swig/commit/c2353b9ab642b680028af2e265f08c03a31ffbef 

And rebased.

erezgeva:
@wsfulton 
Seems issues with gcc13 
I do find it here
https://launchpad.net/ubuntu/+source/gcc-13/13.3.0-6ubuntu2~24.04.1
Perhaps they update the package?
Looking here:
https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/ppa
They support all the way from gcc4, so no reason to remove gcc13.


It is also pre installed on MacOS CI
https://github.com/swig/swig/actions/runs/22435671027/job/64964741257?pr=3332

wsfulton:
Thanks, merged see 7e5bea54016e4cbf6681426b6fad85ad1389121c and abe055f7ec1489cd53ed53d5a534cede7d750f5f

PR #3334, opened 2026-02-14
Python: Add PEP 484 annotations for simple types

Nerixyz:
Towards #735.

This adds [PEP 484](https://peps.python.org/pep-0484/) annotations where it's easy to infer (as suggested in https://github.com/swig/swig/issues/735#issuecomment-3804827060). Specifically, where `SwigType_type` returns a simple type like `T_INT`, `T_STRING`, or `T_VOID`.

Future PRs can extend this to user types like classes or instantiated templates.


wsfulton:
Thanks, this is a useful step towards the PEP484 type hints.

#735, which seems to have died, was talking about adding stub files as well as in-source type hints. Starting with the in-source type hints is a reasonable incremental step.

The majority of the pull request is okay, however, I'm afraid that `transformToPythonAnnotation` is a SWIG anti-pattern and will not be accepted. The correct approach is to use typemaps. In particular this could easily be based off the "doctype" typemaps used for Doxygen comments, by adding in similar "pytyping" typemaps.

The following typemaps should provide the equivalent to this function and the basis for future enhancements:
```swig
%typemap(pytyping) SWIGTYPE "typing.Any"
%typemap(pytyping) bool "bool"
%typemap(pytyping) signed char, unsigned char, short, unsigned short, int, unsigned int, long, unsigned long, long long, unsigned long long "int"
%typemap(pytyping) float, double, long double "float"
%typemap(pytyping) char, wchar_t, char *, wchar_t * "str"
%typemap(pytyping) void "None"
```
Coding it up should be easy too. So will extending it cover more types. The only other thing is some docs are required in Python.html.

laundmo:
While i understand this is a PR for minimal annotations, if more complex ones are ever going to be added, its incredibly likely a simple 1:1 mapping will not suffice - so a method like `transformToPythonAnnotation` might become necessary for more complex logic, and i can imagine that might have been the thought process behind making it a method.

Tho due to the force-push, I can't see how it was written originally, so this is not meant as any kind of criticism. Its just the 2 cents of a passer-by who'd benefit from this feature.

Nerixyz:
> its incredibly likely a simple 1:1 mapping will not suffice

Do you have an example for this? For reference, typemaps are quite powerful in Swig (see [docs](https://www.swig.org/Doc1.3/Typemaps.html#Typemaps_pattern_matching)). An advantage of typemaps is that they can be customized by users who can already customize how types are transferred between C/C++ and Python.

There are two things I'm still uncertain about:

- We might need separate typemaps for parameters and return values, since they could be implemented differently (by customizing `typemap(in)`/`typemap(out)`).
- I don't know how names of types in modules can be prefixed with this approach. For the current module this isn't necessary, but if you're building more than one module, you'll need to qualify the names with `modname.Thing`.

> Tho due to the force-push, I can't see how it was written originally, so this is not meant as any kind of criticism.

You can click on the commit hashes to see the old versions. The `transformToPythonAnnotation` function is here: https://github.com/swig/swig/commit/94ee3494f8a0bf904c78d98af4d0bea747d2aa3c#diff-c686001fe7d4424a7542f21deceb3f4c8e5dc806ec3edbf9cb36444747456a8bR2533.

laundmo:
> Do you have an example for this?

It might already be possible with typemaps, but some kinds of typing in python require entirely separate definitions which are then used in the actual typehint.

Contrived example:
```py
from typing import TypeVar

T = TypeVar("T") # this also has arguments for covariance/contravariance and can be passed a list of types to limit the typevar to

def push(a: MutableSequence[T], b: T) -> MutableSequence[T]:
    a.append(b)
```
tho this is a concern mostly with codegen, not with translating types. I also dont know if it would ever come up. Besides `TypeVar`, i can think of `class MyProto(Protocol)` which is used to represent pythons duck-typing: its a types/signatures-only class which says "anything with the types/signatures i specify matches" - very useful for breaking import loops and not requiring users to inherit from your base class to pass type-check. Again, unsure if this is ever relevant.

> don't know how names of types in modules can be prefixed with this approach. For the current module this isn't necessary, but if you're building more than one module, you'll need to qualify the names with modname.Thing.

if theres some way imports can be generated, you can write either `import modname; var: modname.Thing = ...` or `from modname import Thing; var: Thing = ...`. The latter might be easier to generate, as all it needs is appending one line to the top of the file (and post-processing with a import-sorting formatter like Ruff, if deduplicating is difficult on your end)

> You can click on the commit hashes to see the old versions.

ugh, how did i never notice that?

---

Question: With this basic support, would it already be possible for users of swig to add their own typemaps to extend the simple defaults here?

Source/Modules/python.cxx
```diff
@@ -2411,32 +2443,57 @@ class PYTHON:public Language {
...
-	p = Getattr(p, "tmap:argout:next");
+        switch (anno) {
+        case TYPE_ANNOTATION_C:
+          tm = SwigType_str(tm, 0);
+          break;
+        case TYPE_ANNOTATION_TYPING:
+          tm = Swig_typemap_lookup("pytyping", n, Swig_cresult_name(), 0);
+          if (!tm)
+            // There can't be empty elements in a list, so fall back to Any.
+            tm = NewString("typing.Any");
```
wsfulton: You should provide a typemap warning here instead, something like the C# "imtype" typemap warning:
```c++
         Swig_warning(WARN_CSHARP_TYPEMAP_CSTYPE_UNDEF, input_file, line_number, "No imtype typemap defined for %s\n", SwigType_str(pt, 0));
```
Once you've added in:
```swig
%typemap(pytyping) SWIGTYPE "typing.Any"
```
it'll be the fallback type. I can't quite remember, but you may also need some array or function pointer SWIGTYPE typemaps to cover all possible fallback typemap lookups, but if you familiarise yourself with the `-debug-tmsearch` option, you can easily find out what the lookup search types are for any given type. In any case, there should not be any hard coded fallback, they should only come from typemaps.

wsfulton (changes requested):

wsfulton:
The C# and Java implementations are heavily dependent on obtaining the correct types in the generated code and is very flexible/powerful. A good look at this module is needed as we can leverage these implementations quite heavily.

>    We might need separate typemaps for parameters and return values, since they could be implemented differently (by customizing typemap(in)/typemap(out)).

The C# "ctype", "imtype" and "cstype" typemaps support an "out" typemap attribute which, if specified, will override the main typemap body for output types. Have a look in the C# chapter for "asymmetric type marshalling". You'll find the implementation in csharp.cxx using code like
```c++
Getattr(n, "tmap:cstype:out")`.
```
>    I don't know how names of types in modules can be prefixed with this approach. For the current module this isn't necessary, but if you're building more than one module, you'll need to qualify the names with modname.Thing.

In C# we use a family of special variables `$csclassname`, `$*csclassname`, `$&csclassname` (in Java they're named `$javaclassname`) which are converted into the proxy class name (C# symbol name after %rename). The magic is in the getProxyName methods. It also deals with C# namespaces or Java packages and should be easily adapted for Python.

Source/Modules/python.cxx
Nerixyz: Added the fallback and checks for arrays and function pointers. They're already handled by the `SWIGTYPE` fallback.

Nerixyz:
> The C# "ctype", "imtype" and "cstype" typemaps support an "out" typemap attribute which, if specified, will override the main typemap body for output types

I added support for specifying this attribute for `pytyping` as well. Seems simple enough to be able to have it in this PR.

> In C# we use a family of special variables `$csclassname`, `$*csclassname`, `$&csclassname` (in Java they're named `$javaclassname`) which are converted into the proxy class name (C# symbol name after %rename). The magic is in the getProxyName methods. It also deals with C# namespaces or Java packages and should be easily adapted for Python.

Ah, thank you! I'll add a similar `$pytypename` in a followup PR, since this is only for basic/simple types.

> Question: With this basic support, would it already be possible for users of swig to add their own typemaps to extend the simple defaults here?

Yes, but without something like a `$pytypename`, it's not as useful yet.

Nerixyz:
> I'll add a similar `$pytypename` in a followup PR, since this is only for basic/simple types.

I implemented the `$pytypename` and friends here: https://github.com/Nerixyz/swig/tree/feat/pytypename. Still feels better to do it in another PR, but I'm also open to add it here.

Nerixyz:
Is there anything I can do to help this move forward?

wsfulton:
I'll have a good look at it again today.

Source/Modules/python.cxx
```diff
@@ -224,6 +232,15 @@ static String *getClosure(String *functype, String *wrapper, int funpack = 0) {
   return NULL;
 }
 
+static type_annotation_t getTypeAnnotationMode(Node *n) {
+  String *annotationType = Getattr(n, "feature:python:annotations");
```
wsfulton: We don't use camelCase variables as the norm. Suggest `annotations_type`.

Source/Modules/python.cxx
```diff
@@ -2421,14 +2452,34 @@ class PYTHON : public Language {
...
+          tm = lookupPytyping(p, /*out=*/true);
+          if (!tm)
+            Swig_warning(WARN_PYTHON_TYPEMAP_PYTYPING_UNDEF,
+                         input_file,
+                         line_number,
+                         "Missing required entry in pytyping typemap for %s\n",
+                         SwigType_str(Getattr(p, "tmap:argout:match_type"), 0));
+          break;
+        case TYPE_ANNOTATION_NONE:
+          break;  // unreachable
```
wsfulton, line 2480: Looks reachable to me. Remove confusing comment.

Source/Modules/python.cxx
```diff
@@ -2421,14 +2452,34 @@ class PYTHON : public Language {
...
      * however the result may not accurate. */
     while (p) {
       if ((tm = Getattr(p, "tmap:argout:match_type"))) {
-        tm = SwigType_str(tm, 0);
+        switch (anno) {
+        case TYPE_ANNOTATION_C:
+          tm = SwigType_str(tm, 0);
+          break;
+        case TYPE_ANNOTATION_TYPING:
+          tm = lookupPytyping(p, /*out=*/true);
```
wsfulton: A bit pedantic maybe, but this comment is unnecssary and may get out of date, just remove it, or use a variable with the name 'out'.

Source/Modules/python.cxx
```diff
@@ -2441,12 +2492,20 @@ class PYTHON : public Language {
...
+      case TYPE_ANNOTATION_C:
+        ret = Getattr(n, "type");
+        if (ret)
+          ret = SwigType_str(ret, 0);
+        break;
+      case TYPE_ANNOTATION_TYPING:
+        ret = lookupPytyping(n, /*out=*/true);
+        break;
+      case TYPE_ANNOTATION_NONE:
+        break;  // unreachable
```
wsfulton, line 2505: Looks reachable to me. Remove comment.

Source/Modules/python.cxx
```diff
@@ -2441,12 +2492,20 @@ class PYTHON : public Language {
...
-      if (ret)
-        ret = SwigType_str(ret, 0);
+      switch (anno) {
+      case TYPE_ANNOTATION_C:
+        ret = Getattr(n, "type");
+        if (ret)
+          ret = SwigType_str(ret, 0);
+        break;
+      case TYPE_ANNOTATION_TYPING:
+        ret = lookupPytyping(n, /*out=*/true);
```
wsfulton: A bit pedantic maybe, but this comment is unnecssary and may get out of date, just remove it, or use a variable with the name 'out'.

Source/Modules/python.cxx
```diff
@@ -2456,16 +2515,46 @@ class PYTHON : public Language {
...
     Delete(type);
     return annotation;
   }
 
+  /* ------------------------------------------------------------
+   * lookupPytyping()
+   *
+   * Lookup the annotation string for the PEP 484 pytyping
+   * typemap. If `out` is true, look for overrides with `out`
+   * too, as `n` is a return value.
```
wsfulton: We don't use backticks in comments. Use single quotes if needed.

Examples/test-suite/python/python_annotations_typing_runme.py
```diff
@@ -0,0 +1,90 @@
+import sys
+import inspect
+
+if sys.version_info[0:2] >= (3, 2):
+    from python_annotations_typing import *
+
+    # No __annotations__ support with -builtin or -fastproxy
```
wsfulton, line 7: Add something in the docs about this (even though it might be obvious when stated!)

wsfulton (changes requested):
Barring some minor changes, this will be okay to merge.

I'd like to see some more realistic examples in the docs that use the typemaps as well as some that use the typing module. These should be replicated in the testcase too.  The callable example in the docs is not going to be a common use case (but fine for a testcase).

For a better test and documentation, change 'int' test types to something in testcase (to say C short) as the pytyping type should be something that is not the same as the 'c' annotations nor the same as the C type. Then the test would have the C type as short and the pytyping type something that is not identical (int in this case). Or C++ double results in Python pytyping float.

Likewise, in the docs, a variable called 'str' is confusing for a pytyping type 'str'.

Source/Modules/python.cxx
Nerixyz: This is unreachable, because we early-out at the top of the function: https://github.com/Nerixyz/swig/blob/f6db0d770948890449c208ec2065ee2d549799e5/Source/Modules/python.cxx#L2456-L2457.

Examples/test-suite/python/python_annotations_typing_runme.py
Nerixyz: This is already mentioned in the parent paragraph about type annotations in general:
https://github.com/swig/swig/blob/9a60f9573ff6d379756dbd193ffa8aa1122a0ee6/Doc/Manual/Python.html#L7007-L7017

wsfulton: Oh cool, I missed that, thanks.

Nerixyz:
> I'd like to see some more realistic examples in the docs that use the typemaps as well as some that use the typing module. These should be replicated in the testcase too. The callable example in the docs is not going to be a common use case (but fine for a testcase).

I used an example that shows an optional type. It uses `typing.Optional[...]`.

> For a better test and documentation, change 'int' test types to something in testcase (to say C short) as the pytyping type should be something that is not the same as the 'c' annotations nor the same as the C type.

The `int` type is only used in the out attribute (for `MyType`). There it specifies a PEP 484 type.

In the first example, `float parse_name(const char *name);`, there's already a `const char*` which would be added as-is when the 'c' annotations are used. But here, it's added as "str".

wsfulton:
Okay, looks good, thanks.

PR #3337, opened 2026-02-18
Fix cross-module type cast lookup in SWIG_InitializeModule

benzwick:
When multiple SWIG modules share a type table (via `SWIG_TYPE_TABLE`), `SWIG_InitializeModule` merges each module's cast chains into the shared type entries. However, when adding a new cast entry to a type that already exists in the shared table, the cast entry's `type` pointer was not updated from the module-local `swig_type_info` to the shared `swig_type_info`.

`SWIG_TypeCheckStruct` uses pointer comparison (`head->type == from`) to match cast entries, so the lookup always failed when the local and shared pointers differed.

The first branch (line 264 of `swiginit.swg`, for types that are still local) correctly does `cast->type = target_type`, but the second branch (line 272, for types already in the shared table) was missing this update.

This affects any multi-module SWIG setup where:
1. Multiple modules share a type table (`SWIG_TYPE_TABLE`)
2. A derived type (in module B) needs to be cast to a base type first registered by a different module (module A)
3. The derived type was also first registered by a third module that loaded before module B (so the type already exists in the shared table when module B processes its cast entries)

Discovered while building Guile bindings for MFEM with 11 modules sharing `SWIG_TYPE_TABLE=GuileMFEM`. Passing a `DenseMatrix` smob (from the densemat module) to `Operator-Height` (from the operators module) would fail with "Wrong type argument", despite the DenseMatrix→Operator cast being correctly registered in the shared type table (verified with `SWIGRUNTIME_DEBUG`).

### DIFF
--- a/Lib/swiginit.swg
+++ b/Lib/swiginit.swg
@@ -269,7 +269,10 @@ SWIG_InitializeModule(SWIG_INIT_CLIENT_DATA_TYPE clientdata) {
 #ifdef SWIGRUNTIME_DEBUG
             if (ocast) printf("SWIG_InitializeModule: skip old cast %s\n", target_type->name);
 #endif
-            if (!ocast) target_type = 0;
+            if (!ocast) {
+              cast->type = target_type;
+              target_type = 0;
+            }
           }
         }
       }


wsfulton:
Thanks for the very detailed analysis and fix. I suspect the regression was introduced in c840bb728a.

I would like to replicate the issue as a test to prevent regressions in the future. I have created a branch [pull-3337](https://github.com/swig/swig/tree/pull-3337) and commit 956eaf529a18f3c0b30e11c96e7dff1166311b91 attempts to write the testcase that you describe in the commit message for 5dab6a8eda70d9e05b11a37bd80be200fbd975cd. I apologise for writing the runtime test in Python (I really rather struggle with Guile). I couldn't replicate the issue, so could you please tweak the test to replicate the issue. Feel free to switch the runtime test to Guile.

You can run the test as follows:
```bash
cd Examples/test-suite/python
make import_share.multicpptest
```

benzwick:
@wsfulton I've pushed a branch [`pull-3337-bz`](https://github.com/benzwick/swig/tree/pull-3337-bz) based on your `pull-3337` with the test tweaked to reproduce the issue.

**Why the original test didn't reproduce it:**

Module C's `%import "import_share_b.i"` gave SWIG full knowledge of the `DenseMatrix : public Operator` inheritance. When C initialized, it added the `DenseMatrix→Operator` cast to the shared type table. When B loaded later, the cast already existed — so the buggy `else` branch in `swiginit.swg` was never reached.

**How the fix works:**

1. **Forward-declare `DenseMatrix` in `import_share_c.i`** instead of `%import`ing `import_share_b.i`. This makes module C register `DenseMatrix` in the shared type table as an opaque type (no inheritance cast to `Operator`).
2. **Import order A, C, B** — C must load before B so that `DenseMatrix` is already in the shared table when B tries to add the inheritance cast. This hits the `else` branch in `SWIG_InitializeModule` where `cast->type` isn't updated from the module-local pointer to the shared pointer.

**Test results without the `swiginit.swg` fix:**

- **Guile: FAILS** ✓ — reproduces the bug:
  ```
  ERROR: In procedure Operator-Height: Wrong type argument in position 1:
  #<collectable-swig-pointer DenseMatrix * ...>
  ```
- **Python: PASSES** — Python uses `SWIG_TypeCheck` (string-based name comparison) rather than `SWIG_TypeCheckStruct` (pointer comparison), so the stale pointer doesn't cause a failure.

The bug is specific to languages that use `SWIG_TypeCheckStruct` for type checking (Guile, and likely others that use pointer comparison). I've added both a Guile runtime test (`guile/` + `schemerunme/`) and updated the Python runtime test with proper assertions and the correct import order.

PR #3341, opened 2026-02-20
Add support for Java 25

erezgeva:
- Suppress restricted methods warning
- Add Java 25 to Windows
- Add versioned Java to Linux CI
- Skip MacOS flow in in Linux CI
- Improve Linux CI CSTD code

### DIFF
--- a/Tools/CI-linux-install.sh
+++ b/Tools/CI-linux-install.sh
@@ -38,6 +38,12 @@ case "$SWIGLANG" in
 		  $HOME/bin/gimme --list
 		fi
 		;;
+	"java")
+		if [[ -n "$VER" ]]; then
+			declare -n java_path="JAVA_HOME_${VER}_X64"
+			echo "JAVA_HOME=$java_path" >> $GITHUB_ENV
+		fi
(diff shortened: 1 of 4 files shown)


Tools/CI-linux-install.sh
```diff
@@ -38,6 +38,12 @@ case "$SWIGLANG" in
 		  $HOME/bin/gimme --list
 		fi
 		;;
+	"java")
+		if [[ -n "$VER" ]]; then
+			declare -n java_path="JAVA_HOME_${VER}_X64"
```
wsfulton, line 43: declare -n is pointless here, suggest you remove it in a later commit as I'll merge this now

erezgeva: `-n        make NAME a reference to the variable named by its value`

You can use older approach that works since bash 3 (the version MacOS uses).

                `java_path="JAVA_HOME_${VER}_X64"`
                `echo "JAVA_HOME=${!java_path}" >> $GITHUB_ENV`

i.e. using the exclamation point to do indirect expansion.

PR #3342, opened 2026-02-21
Lua improvements

erezgeva:
- Add missing section in document table of contents
- lua.cxx: variable 'outputs' set but not used
- Add director tests

### DIFF
--- a/Source/Modules/lua.cxx
+++ b/Source/Modules/lua.cxx
@@ -2574,12 +2574,6 @@ public:
       Parm *p;
       char source[256];
 
-      #if 0 // warning: variable 'outputs' set but not used [-Wunused-but-set-variable]
-      int outputs = 0;
-      if (!is_void)
-	outputs++;
-      # endif
-
       /* Get lua state */
       Printf(w->code, "lua_State *L = swig_get_lua_state();\n");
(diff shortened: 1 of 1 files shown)


wsfulton:
Can you please just delete the code in lua.cxx instead of disabling it with `#if 0`.

Did one of our CI jobs not pick up this unused code? Or maybe you found it on one of your own specific compiler setups?

More runtime tests are nice :smile:, thanks.

erezgeva:
> lease just delete the code in lua.cxx instead of disabling it with `#if 0`.

I saw it on MacOS build.
It uses a newer version of clang.
See for example `build` in 
https://github.com/swig/swig/actions/runs/22257437873/job/64390363381

Looking at `git blame Source/Modules/lua.cxx`.

I see @christophe-calmejane add it on adding support for directors.
We should ask him if we can remove the code, or was it a seed for more expansions.

I did found some issues and crash with Lua director which I investigate at the moment.
So I though perhaps the code can be used in the future.

erezgeva:
@wsfulton, I remove the code.

christophe-calmejane:
Yes sorry, it was a leftover of debugging directors. I thought the code was configured as "warning-as-error" as I usually do, didn't see it. Thanks for fixing it.

PR #3344, opened 2026-02-23
Use cached tools on Linux CI

erezgeva:
- Use cached tools for python, ruby, go and javascript/node. The cached tools are used in case the version match.
- Check if required PHP version is already installed.
- Move compiler settings to `CI-linux-install.sh` and merge them with actual installation of versioned GCC.
- Add `update_env()` and `update_path()` functions to use in `CI-linux-install.sh` as the script should not use GitHub variables directly.
- Add support for new versions of ruby and go in `Tools/testflags.py`.
- Fix tests with integer in configure.ac, integer conditions should not use string notation.
- Print used C/C++ standard.

### DIFF
--- a/Tools/CI-linux-environment.sh
+++ b/Tools/CI-linux-environment.sh
@@ -3,6 +3,7 @@
 # e.g. RETRY=travis-retry SWIGLANG=python
 # Sets up environment for using various target languages
 # For Github Actions where the environment is not preserved between steps
+set -e # exit on failure (same as -o errexit)
 
 case "$SWIGLANG" in
 	"go")
--- a/Tools/CI-linux-install.sh
+++ b/Tools/CI-linux-install.sh
@@ -1,32 +1,28 @@
 #!/bin/bash
 # Expected to be called from elsewhere with certain variables set
 # e.g. RETRY=travis-retry SWIGLANG=python GCC=7
+# Also provide update_env() and update_path()
+set -e # exit on failure (same as -o errexit)
 
-run_apt()
-{
-	$RETRY sudo apt-get -qq $@
-}
-add_apt_repository()
-{
-	$RETRY sudo add-apt-repository -y ppa:$1
-}
+# See list of cached tools in:
+# https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cached-tools
 probe_cached_tool()
 {
 	tool_path=$(ls -d /opt/hostedtoolcache/$1/$VER.*/x64/bin 2> /dev/null | head -1 || true)
 }
 
 if [[ "$compiler" = 'clang' ]]; then
-	run_apt update
+	$RETRY sudo apt-get -qq update
 	CC="clang"
 	CXX="clang++"
 elif [[ -n "$GCC" ]]; then
-	add_apt_repository ubuntu-toolchain-r/test
-	run_apt update
-	run_apt install g++-$GCC
+	$RETRY sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+	$RETRY sudo apt-get -qq update
+	$RETRY sudo apt-get -qq install g++-$GCC
 	CC="gcc-$GCC"
 	CXX="g++-$GCC"
 else
-	run_apt update
+	$RETRY sudo apt-get -qq update
 	CC="gcc"
 	CXX="g++"
 fi
@@ -36,7 +32,7 @@ ls -la $(which $CC) $(which $CXX)
 $CC --version
 $CXX --version
 
-run_apt install libboost-dev libpcre3-dev
+$RETRY sudo apt-get -qq install libboost-dev libpcre3-dev
 # Note: testflags.py needs python, but python is pre-installed
 
 WITHLANG=$SWIGLANG
@@ -44,14 +40,14 @@ WITHLANG=$SWIGLANG
 case "$SWIGLANG" in
 	"")     ;;
 	"csharp")
-		run_apt install mono-devel
+		$RETRY sudo apt-get -qq install mono-devel
 		;;
 	"d")
 		if [[ $VER =~ ^2\. ]]; then
 			$RETRY wget http://downloads.dlang.org/releases/2.x/${VER}/dmd_${VER}-0_amd64.deb
 			$RETRY sudo dpkg -i dmd_${VER}-0_amd64.deb
 		else
-			run_apt install "$VER"
+			$RETRY sudo apt-get -qq install "$VER"
 		fi
 		;;
 	"go")
@@ -94,7 +90,7 @@ case "$SWIGLANG" in
 					fi
 					case "$VER" in
 						0.10|0.12|4|6)
-							# run_apt install nodejs node-gyp
+							# $RETRY sudo apt-get -qq install nodejs node-gyp
 							$RETRY npm install -g node-gyp@$VER
 							;;
 						8)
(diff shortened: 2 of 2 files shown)


Tools/CI-linux-environment.sh
```diff
@@ -3,34 +3,31 @@
 # e.g. RETRY=travis-retry SWIGLANG=python
 # Sets up environment for using various target languages
 # For Github Actions where the environment is not preserved between steps
-set -e # exit on failure (same as -o errexit)
```
wsfulton, line 6: We need exit on failure to work. Previously this line was necessary and I'd rather have it definitely working so best left as is.

Tools/CI-linux-install.sh
```diff
@@ -1,105 +1,156 @@
 #!/bin/bash
 # Expected to be called from elsewhere with certain variables set
 # e.g. RETRY=travis-retry SWIGLANG=python GCC=7
-set -e # exit on failure (same as -o errexit)
```
wsfulton, line 4: We need exit on failure to work. Previously this line was necessary and I'd rather have it definitely working so best left as is.

Tools/CI-linux-environment.sh
```diff
@@ -3,34 +3,31 @@
...
 
 case "$SWIGLANG" in
 	"go")
-		if [[ "$VER" ]]; then
+		if [[ -x "$HOME/bin/gimme" ]]; then
 		  eval "$($HOME/bin/gimme ${VER}.x)"
 		  $HOME/bin/gimme --list
 		fi
```
wsfulton, line 13: This should error out if VER is set and gimme is not installed.

Tools/CI-linux-install.sh
```diff
@@ -1,105 +1,156 @@
...
-	$RETRY sudo apt-get -qq update
-	$RETRY sudo apt-get install -qq g++-$GCC
+run_apt()
+{
+	$RETRY sudo apt-get -qq $@
+}
+add_apt_repository()
+{
+	$RETRY sudo add-apt-repository -y ppa:$1
+}
```
wsfulton: I don't really like these particular one liner functions. An important requirement for me is to replicate the installs on my own machine. I want to easily copy/paste these commands from this script in order to install the packages on my own machine. It's more readily digestible if the actual commands are also in the script.

wsfulton (commented):
Is /opt/hostedtoolcache documented? Maybe put a link to it in probe_cached_tool().

I havn't got the energy to check that the versions requested are actually installed. It's quite easy to make a mistake and so generally I would go for minimal refactoring. So, if you havn't, could you please validate the changes are good (checking versions/compilers etc in the logs).

When we had just a few versions of each language, it wasn't so hard to mentally know what version was being tested. Now with so many versions, it is hard to know from the results which version is being tested if it is the default. I hard started to explicitly specify the version to that installed by default, say running `apt-get install ruby`. I'd then specify VER=3.2 and `apt-get install ruby=3.2` would be the same. Perhaps you can look at that as a separate commit?

erezgeva:
> Is /opt/hostedtoolcache documented? Maybe put a link to it in probe_cached_tool().

GitHub is not that great with documentation.
They are mention in
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cached-tools
I add this reference to script


> 
> I havn't got the energy to check that the versions requested are actually installed. It's quite easy to make a mistake and so generally I would go for minimal refactoring. So, if you havn't, could you please validate the changes are good (checking versions/compilers etc in the logs).

I did passed the logs and check that all cached tools are used.
Including JavaScript with node.
There is a new JavaScript node version 24 which does not works, probably need some updates on our SWIG JavaScript node generation.
Unlike MacOS CI which only use cached tools, Linux will try to  install any tool which is not in the cached tools.
So while MacOS CI and Window CI (which also based on cached tools) need updating every time GitHub update the cached tools, Linux will be more smooth.
I also checked the PHP is used without reinstalling.


> 
> When we had just a few versions of each language, it wasn't so hard to mentally know what version was being tested. Now with so many versions, it is hard to know from the results which version is being tested if it is the default. I hard started to explicitly specify the version to that installed by default, say running `apt-get install ruby`. I'd then specify VER=3.2 and `apt-get install ruby=3.2` would be the same. Perhaps you can look at that as a separate commit?

It is possible to add a message when a VER is equal to the default version. Would that helps?

Tools/CI-linux-environment.sh
erezgeva: No if the used VER exist in the cached tools, we will not use gimme.
So testing VER will generate error as gimme is not in HOME/bin in that case.

So only when we actually use gimme, gimme is in HOME/bin, which is what we should test!

Go 1.20 does not have cached tool and gimme is used
https://github.com/swig/swig/actions/runs/22288383397/job/64471074114?pr=3344

While Go 1.22 does have cached tool in
/opt/hostedtoolcache/go/1.22.12/x64/bin 
And gimme is **NOT** used
https://github.com/swig/swig/actions/runs/22288383397/job/64471073989?pr=3344

They both have `VER` defined.

Tools/CI-linux-install.sh
erezgeva: You do not have $RETRY defined.

My concern was inconsistent with running `apt`.
Some place did not use  `$RETRY` and some used 
`sudo apt-get  install -qq`.

That can be audit and rearrange manually, yet using this one liner is a more robust way to ensure consistent.

The one liner was in purpose only the `sudo apt-get -qq` part.
Since I do not use `-qq` when testing.
And I tend to run apt in root shell, so I do not use `sudo`.

I will revert and let you think about the points I raise and how to sort it in a way you better like.

Tools/CI-linux-environment.sh
erezgeva: This is a duplicate, as this is the GitHub default.

But I will revert.

wsfulton:
> I did passed the logs and check that all cached tools are used. Including JavaScript with node. There is a new JavaScript node version 24 which does not works, probably need some updates on our SWIG JavaScript node generation. Unlike MacOS CI which only use cached tools, Linux will try to install any tool which is not in the cached tools. So while MacOS CI and Window CI (which also based on cached tools) need updating every time GitHub update the cached tools, Linux will be more smooth. I also checked the PHP is used without reinstalling.
> 
:+1: 

> > When we had just a few versions of each language, it wasn't so hard to mentally know what version was being tested. Now with so many versions, it is hard to know from the results which version is being tested if it is the default. I hard started to explicitly specify the version to that installed by default, say running `apt-get install ruby`. I'd then specify VER=3.2 and `apt-get install ruby=3.2` would be the same. Perhaps you can look at that as a separate commit?
> 
> It is possible to add a message when a VER is equal to the default version. Would that helps?

That would be useful in the logs. My motivation really was for it to be easier to digest what target language version is being used what with a hundred odd jobs being run if the version number is displayed in the name. That's all. Not terribly important.

erezgeva:
> > I did passed the logs and check that all cached tools are used. Including JavaScript with node. There is a new JavaScript node version 24 which does not works, probably need some updates on our SWIG JavaScript node generation. Unlike MacOS CI which only use cached tools, Linux will try to install any tool which is not in the cached tools. So while MacOS CI and Window CI (which also based on cached tools) need updating every time GitHub update the cached tools, Linux will be more smooth. I also checked the PHP is used without reinstalling.
> 
> 👍
> 
> > > When we had just a few versions of each language, it wasn't so hard to mentally know what version was being tested. Now with so many versions, it is hard to know from the results which version is being tested if it is the default. I hard started to explicitly specify the version to that installed by default, say running `apt-get install ruby`. I'd then specify VER=3.2 and `apt-get install ruby=3.2` would be the same. Perhaps you can look at that as a separate commit?
> > 
> > 
> > It is possible to add a message when a VER is equal to the default version. Would that helps?
> 
> That would be useful in the logs. My motivation really was for it to be easier to digest what target language version is being used what with a hundred odd jobs being run if the version number is displayed in the name. That's all. Not terribly important.

<del>I will try to use this</del>
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#run-name


This value can include expressions and can reference the [github](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) and [inputs](https://docs.github.com/en/actions/learn-github-actions/contexts#inputs-context) contexts.

Which means we can not rename the job after it start.
Same problem as doing dynamic filtering of the matrix based on changed files. The matrix and the names are hard coded.

PR #3345, opened 2026-02-23
Allow typedef enum values to use the typedef type name.

waj334:
Previously, the type name for enum values derived from a C-style typedef enum would resolve to int even though a proper Go type for the enum was properly emitted. Now the enum value emitter checks if the enum value type has a "tdname" attribute first before falling back to int.

### DIFF
--- a/Source/Modules/go.cxx
+++ b/Source/Modules/go.cxx
@@ -1930,7 +1930,13 @@ private:
     Node *parent = parentNode(n);
 
     if (Getattr(parent, "unnamed")) {
-      Setattr(n, "type", NewString("int"));
+      char* tdname = GetChar(parent, "tdname");
+      if (tdname && strlen(tdname)) {
+        // Use the typedef as the typename.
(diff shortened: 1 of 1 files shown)


wsfulton:
Hi @waj334, we'll need a motivating C example to consider this. Include a go runtime test if you can please.

waj334:
@wsfulton Thanks for replying. This is one of 2 strange behaviors I'm finding when generating bindings for LLVM's C API which uses C-Style typedef declarations for most types. In the case of enums, they are declared like the following:

```c
typedef enum {
    LLVMCodeGenLevelNone,
    LLVMCodeGenLevelLess,
    LLVMCodeGenLevelDefault,
    LLVMCodeGenLevelAggressive
} LLVMCodeGenOptLevel;
```

SWIG sees `LLVMCodeGenOptLevel` and creates a Go type for it, but it doesn't associate any of the enum values with the typedef and defaults to `int` for all of them. I'll create tests for this and update the PR.

PR #3347, opened 2026-02-23
Ruby: Avoid usage of deprected untyped API (but use internal API)

larskanis:
Use two `rb_data_type_t`'s, one for own and one for foreign objects. This is in contrast to #3326, which uses one `rb_data_type_t` with constant function pointers.

That way the C++ object pointer can be stored directly into the Ruby object and the assumption of `DATA_PTR(Ruby_obj)` equals C++ object pointer is still true.
The downside is, that the rb_data_type_t of an Ruby object is const and can therefore only changed by internal ruby API.

So this PR essential changes from deprecated to internal Ruby API. :-(

### DIFF
--- a/Lib/ruby/director.swg
+++ b/Lib/ruby/director.swg
@@ -255,7 +255,7 @@ namespace Swig {
     }
 
     void swig_acquire_ownership_obj(void *vptr, swig_ruby_owntype own) const {
-      if (vptr && own.datafree) {
+      if (vptr && own.cext_type->function.dfree) {
 	SWIG_GUARD(swig_mutex_own);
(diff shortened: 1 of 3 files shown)


larskanis:
Only [Ruby-4.0 succeeds](https://github.com/swig/swig/actions/runs/22319195081/job/64572508475?pr=3347), the older versions need some adjustments. But I don't think that this PR is the way to go and it makes sense to continue here. It still breaks compatibility, when ruby extensions use functions related to the old untyped API. `DATA_PTR()` is usable with objects of the new _and_ the old API, but most others like `Data_Get_Struct()` aren't. So compatibility issues can't be avoided when switching to the typed API.

PR #3348, opened 2026-02-24
Prevent Segmentation fault in director

erezgeva:
Prevent Segmentation fault (core dumped).
Make sure an error is rise properly upward.

- Add protection after calling `Swig_director_emit_dynamic_cast()`.
- Add try-catch around the action code of the director enabled method.

The dynamic cast yeild null and not a proper director object. As the object was allocated inside the C++ code using a `create` method and was not wrapped when passed back to language.

This patch aims preventing the Segmentation fault and forward the error to the script.

**We can add protection after calling `Swig_director_emit_dynamic_cast()` to other director languages.**

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,15 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.5.0 (in progress)
 ===========================
 
+2026-03-19: wsfulton
+            [C#, D] Always initialise the return type in the C/C++ code. Users with custom
+            typemaps may need to provide a valid initial value via the "null" attribute
+            in the "out" typemap if 0 is not a valid initial value.
+
+2026-03-19: erezgeva
+            Avoid a segfault by handling a failed dynamic_cast in generated director code
+            by instead throwing a language specific exception "'self' is not a director".
+
 2026-03-05: FilipAlg
             [C#] #3354 Fully qualify generated C# interfaces when using the interface
             and nspace features.
--- a/Lib/d/cdata.i
+++ b/Lib/d/cdata.i
@@ -22,7 +22,7 @@
 %typemap(dtype) SWIGCDATA "ubyte[]"
 %typemap(imtype) SWIGCDATA "ubyte[]"
 %typemap(ctype) SWIGCDATA "SWIG_c_array"
-%typemap(out) SWIGCDATA %{
+%typemap(out, null="{0, 0}") SWIGCDATA %{
   $result.array = $1.data;
   $result.len = $1.len;
 %}
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -855,7 +855,7 @@ public:
 
     is_void_return = Cmp(c_return_type, "void") == 0;
     if (!is_void_return)
-      Wrapper_add_localv(f, "jresult", c_return_type, "jresult", NIL);
+      Wrapper_add_localv(f, "jresult", c_return_type, "jresult = $null", NIL);
 
     Printv(f->def, " SWIGEXPORT ", c_return_type, " SWIGSTDCALL ", wname, "(", NIL);
 
@@ -1007,11 +1007,10 @@ public:
         Setattr(n, "csharp:canthrow", "1");
         Append(f->code, "if (!darg) {\n");
         Append(f->code, "  SWIG_CSharpSetPendingExceptionArgument(SWIG_CSharpArgumentNullException, \"'self' is not a director\", 0);\n");
-        if (!is_void_return) {
-          Append(f->code, "  jresult = 0;\n");
-          Append(f->code, "  return jresult;\n");
-        } else
+        if (is_void_return)
           Append(f->code, "  return;\n");
+        else
+          Append(f->code, "  return jresult;\n");
         Append(f->code, "}\n");
       }
       String *actioncode = emit_action(n);
@@ -1082,10 +1081,13 @@ public:
       Setattr(n, "csharp:canthrow", "1");
     }
 
-    if (!null_attribute)
+    if (!null_attribute) {
       Replaceall(f->code, "$null", "0");
-    else
+      Replaceall(f->locals, "$null", "0");
+    } else {
       Replaceall(f->code, "$null", null_attribute);
+      Replaceall(f->locals, "$null", null_attribute);
+    }
 
     /* Dump the function out */
     if (!native_function_flag) {
--- a/Source/Modules/d.cxx
+++ b/Source/Modules/d.cxx
@@ -1578,7 +1578,7 @@ public:
 
     is_void_return = (Cmp(c_return_type, "void") == 0);
     if (!is_void_return)
-      Wrapper_add_localv(f, "jresult", c_return_type, "jresult", NIL);
+      Wrapper_add_localv(f, "jresult", c_return_type, "jresult = $null", NIL);
 
     Printv(f->def, " SWIGEXPORT ", c_return_type, " ", wname, "(", NIL);
 
@@ -1723,11 +1723,10 @@ public:
         Setattr(n, "d:canthrow", "1");
         Append(f->code, "if (!darg) {\n");
         Append(f->code, "  SWIG_DSetPendingException(SWIG_DNullReferenceException, \"'self' is not a director\");\n");
-        if (!is_void_return) {
-          Append(f->code, "  jresult = 0;\n");
-          Append(f->code, "  return jresult;\n");
-        } else
+        if (is_void_return)
           Append(f->code, "  return;\n");
+        else
+          Append(f->code, "  return jresult;\n");
         Append(f->code, "}\n");
       }
       String *actioncode = emit_action(n);
@@ -1800,10 +1799,13 @@ public:
       Setattr(n, "d:canthrow", "1");
     }
 
-    if (!null_attribute)
+    if (!null_attribute) {
       Replaceall(f->code, "$null", "0");
-    else
+      Replaceall(f->locals, "$null", "0");
+    } else {
       Replaceall(f->code, "$null", null_attribute);
+      Replaceall(f->locals, "$null", null_attribute);
+    }
 
     /* Dump the function out */
     if (!native_function_flag) {


Source/Modules/lua.cxx
```diff
@@ -763,15 +763,28 @@ class LUA:public Language {
...
+    }
+
     // Remember C name of the wrapping function
     Setattr(n, "wrap:name", wname);
 
     /* Emit director dynamic cast for protected/non-public member access */
-    Swig_director_emit_dynamic_cast(n, f);
+    if(Swig_director_emit_dynamic_cast(n, f)) {
+        /* Add protection */
+        Printf(f->code, "if(!darg){Swig::DirectorMethodException::raise(\"'self' is not a director\");}\n");
```
wsfulton:
```suggestion
        Printf(f->code, "if(!darg) {\n  Swig::DirectorMethodException::raise(\"'self' is not a director\");\n}\n");
```
Generated code should follow normal formatting conventions.

wsfulton:
Just a note / comment, that the additional try/catch is already in use in Ruby, Perl and Python. Perhaps there is a testcase that already tests this.

erezgeva:
> Just a note / comment, that the additional try/catch is already in use in Ruby, Perl and Python. Perhaps there is a testcase that already tests this.

Yap, I copy it from Python, but I am unaware of a test.
I used a draft of  `director_protected`  in Lua.

I see this in python:

```
protected = 1
try:
    f.ping()
    protected = 0
except:
    pass
if not protected:
    raise RuntimeError("Foo::ping is protected")
```

Looking in python, there is no protection on `darg`.
Perhaps we should.

Source/Modules/lua.cxx
erezgeva: Hope that I format code properly

wsfulton:
> Looking in python, there is no protection on darg.
There is a try/catch handler around the dynamic cast in python, maybe you mean some other protection?

I would separate the Lua specific enhancement into one commit and then have another commit that reduces the number of places that the `Swig_director_emit_dynamic_cast` is emitted for all languages. Some runtime tests showing the fix is needed too.

erezgeva:
> Swig_director_emit_dynamic_cast

@wsfulton
I understand separation, but what do you mean by reducing?
Do you want to reduce the use of `Swig_director_emit_dynamic_cast` or do you mean we need to add protection same as I add to Lua?

wsfulton:
> I understand separation, but what do you mean by reducing?

This part of the patch conditionally adds in the dynamic cast, so it reduces when the cast is emitted (compared to the other languages):
```patch
-    Swig_director_emit_dynamic_cast(n, f);
+    if (Swig_director_emit_dynamic_cast(n, f)) {
```
So am suggesting the first commit unconditionally adds the dynamic cast as an enhancement for Lua. Then the second commit can conditionally add the cast, but does it for all languages.

erezgeva:
> his part of the patch conditionally adds in the dynamic cast, so it reduces when the cast is emitted (compared to the other languages):

I think we have a misunderstanding.
The director dynamic cast is already part of Lua. It **does** works, at least partially.

What I did found, is a bug that when the object is passed from C++ to language using a dedicated method, it does not wrapped properly.
When a new object is created in the Lua script it is wrapped properly.
This is another bug that I do not handle in this PR.

The issue, when an object is not wrapped, require a separate PR. This PR do not try to fix this bug.  
Not really sure if and how to solve this issue, I will need a separate investigation on the matter.

My aim is to add protection to the already exist `dynamic cast` and guarantee that we do not get core dump, but forward the error up.

I agree that we should add a similar protection to all languages that uses the `Swig_director_emit_dynamic_cast`.

I am working on it :smile:

erezgeva:
@wsfulton,

Update on adding protection to emit dynamic cast:

|Language | Change  | Excepation  | Message | try-catch |
|---------|---------|-------------|---------|-----------|
|lua      | Yes     | Yes         | Yes     | Yes       |
|perl5    | Yes     | Yes         | Yes     | Yes       |
|php      | Yes     | Yes         | Yes     | Yes       |
|python   | Yes     | Yes         | Yes     | Yes       |
|go       | Yes     | Yes         | not     | none      |
|ruby     | Yes     | none        | ?       | Yes       |
|octave   | Yes     | none        | ?       | not       |
|java     | Yes     | none        | ?       | none      |
|ocaml    | Yes     | none        | ?       | none      |
|scilab   | yes     | no director |         | none      |
|csharp   | non     |             |         | none      |
|d        | non     |             |         | none      |

- go - the error/exception happens, but i can not see the error/exception message, instead I see a general message. 
- ruby, octave, java, ocaml - I add protection with exception,
                            but i can not see any exception when running the test.
                            Even when removing the condition so the exception should always happen.

- scilab      - Do not support the `director` feature.
                Remove `Swig_director_emit_dynamic_cast()`.

- csharp + d  - Need a callback for excepation. The C API of d is similar to C#.
                I do **not know** how to add protection.

Source/Modules/go.cxx
```diff
@@ -1637,7 +1637,12 @@ class GO : public Language {
   void emitGoAction(Node *n, List *base, ParmList *parms, SwigType *result, Wrapper *f) {
     String *actioncode;
     if (!base || isStatic(n)) {
-      Swig_director_emit_dynamic_cast(n, f);
+      if (Swig_director_emit_dynamic_cast(n, f)) {
+        /* Add protection */
+        Append(f->code, "if(!darg) {\n");
```
wsfulton:
```suggestion
        Append(f->code, "if (!darg) {\n");
```
clang-format doesn't check the generated code, but this is how it should be. Similarly for other changes.

erezgeva: Sure

wsfulton:
For C#, generate:
```c++
if (!darg) {
  SWIG_CSharpSetPendingExceptionArgument(SWIG_CSharpArgumentNullException, "'self' is not a director", 0);
  return <0 or nothing>;
}
```
Like the Java implementation, use is_void_return to determine whether to return nothing or jresult. jresult will need initialising to zero though by default as an additional change.

For D, generate:
```c++
  if (!jarg1) {
    SWIG_DSetPendingException(SWIG_DNullReferenceException, "'self' is not a director");
    return <0 or nothing>;
  }
```

erezgeva:
> For C#, generate:
> 
> ```c++
> if (!darg) {
>   SWIG_CSharpSetPendingExceptionArgument(SWIG_CSharpArgumentNullException, "'self' is not a director", 0);
>   return <0 or nothing>;
> }
> ```
> 
> Like the Java implementation, use is_void_return to determine whether to return nothing or jresult. jresult will need initialising to zero though by default as an additional change.
> 
> For D, generate:
> 
> ```c++
>   if (!jarg1) {
>     SWIG_DSetPendingException(SWIG_DNullReferenceException, "'self' is not a director");
>     return <0 or nothing>;
>   }
> ```

I thought I try something similar.
I'll try once more.

erezgeva:
Yap, did try that

```
diff --git a/Source/Modules/csharp.cxx b/Source/Modules/csharp.cxx
index 50bc35754..353889796 100644
--- a/Source/Modules/csharp.cxx
+++ b/Source/Modules/csharp.cxx
@@ -855,7 +855,7 @@ public:
 
     is_void_return = Cmp(c_return_type, "void") == 0;
     if (!is_void_return)
-      Wrapper_add_localv(f, "jresult", c_return_type, "jresult", NIL);
+      Wrapper_add_localv(f, "jresult", c_return_type, "jresult = 0", NIL);
 
     Printv(f->def, " SWIGEXPORT ", c_return_type, " SWIGSTDCALL ", wname, "(", NIL);
 
@@ -1001,8 +1001,16 @@ public:
     String *null_attribute = 0;
     // Now write code to make the function call
     if (!native_function_flag) {
-
-      Swig_director_emit_dynamic_cast(n, f);
+      if (Swig_director_emit_dynamic_cast(n, f)) {
+        /* Add protection */
+        Append(f->code, "if (!darg) {\n");
+        Append(f->code, "  SWIG_CSharpSetPendingExceptionArgument(SWIG_CSharpArgumentNullException, \"'self' is not a director\", 0);\n");
+        if (is_void_return)
+          Append(f->code, "  return;\n");
+        else
+          Append(f->code, "  return jresult;\n");
+        Append(f->code, "}\n");
+      }
       String *actioncode = emit_action(n);
 
       /* Return value if necessary  */

```

I get this error
````
.././../director_protected.i:37: Warning 845: Unmanaged code contains a call to a SWIG_CSharpSetPendingException method and C# code does not handle pending exceptions via the canthrow attribute.
.././../director_protected.i:42: Warning 845: Unmanaged code contains a call to a SWIG_CSharpSetPendingException method and C# code does not handle pending exceptions via the canthrow attribute.
...
````

Same story in D

**Comment by wsfulton** [comment]
Adding
```c++
      Setattr(n, "csharp:canthrow", "1");
```
should fix that and add in exception handling in the C# code.

Reminder to self: to take a close look at the generated code for handling exception here if no runtime test.

erezgeva:
Add C# and D

erezgeva:
Looking on
https://en.cppreference.com/w/c/language/scalar_initialization.html
https://en.cppreference.com/w/c/language/struct_initialization.html
https://en.cppreference.com/w/cpp/language/aggregate_initialization.html

`Wrapper_add_localv(f, "jresult", c_return_type, "jresult = {0}", NIL);`
Should work for most  scalars, structres and objects.


Scalar fails with old C++ :cry:

wsfulton:
Thanks, committed in ff003116ef96cbaf3a1aa3266e28de275e1ee451 5a142e7f6dad44079a5e393081a292e876c61bb1 and a follow up be56529251dbef7d92d5777c0691331f412adac0

erezgeva:
> Thanks, committed in [ff00311](https://github.com/swig/swig/commit/ff003116ef96cbaf3a1aa3266e28de275e1ee451) [5a142e7](https://github.com/swig/swig/commit/5a142e7f6dad44079a5e393081a292e876c61bb1) and a follow up [be56529](https://github.com/swig/swig/commit/be56529251dbef7d92d5777c0691331f412adac0)

Thanks.
Always new thing to learn :smile:

PR #3349, opened 2026-02-25
Use of `Tools/testflags.py` in Windows CI using MingW64 GCC

erezgeva:
- Update python in `Tools/testflags.py`
- As `cccl` do not support GCC warnings, skip it.


Tools/testflags.py
```diff
@@ -39,6 +39,9 @@ def get_cflags(language, std, compiler):
     if compiler == "clang":
         cflags["guile"] += " -Wno-attributes" # -Wno-attributes is for clang LLVM 3.5 and bdw-gc < 7.5 used by guile
 
+    if platform.system() == 'Windows':
+        cflags["python"] += " -Wno-float-conversion"
+
     if language not in cflags:
```
wsfulton: Can we not fix whatever is triggering this instead of suppressing this warning?

.github/workflows/windows.yml
```diff
@@ -265,7 +271,7 @@ jobs:
           set -x
           if [[ -z "$COMPILER" ]]; then
             which cl.exe
-            cl.exe /? 2>&1 | head -n1
+            cl.exe -? 2>&1 | head -n1
```
wsfulton, line 362: This works as is, why change it to arguably something non-conventional?

erezgeva: Looking at `https://github.com/swig/cccl/blob/master/cccl`.
I saw 
```
operating_system=$(uname -o)
case $operating_system in
    Msys)
        slash="-"
        ;;
    *)
        slash="/"
        ;;
esac
```
And when I use  `cl.exe -E -Zc:__cplusplus -Zc:preprocessor -PD dummy.cpp`.
I saw the `/` was not working properly.
We are using MSYS2 bash here.

Tools/testflags.py
erezgeva: I'll have a look.

erezgeva:
```
example_wrap.c:4059:35: warning: conversion from 'double' to 'float' may change value [-Wfloat-conversion]
 4059 |     if (SWIG_Float_Overflow_Check(v)) {
      |                                   ^
```

Perhaps we need to have `SWIG_Float_Overflow_Check()`  using `double`.

I continue to invastigate.

erezgeva: In
`Lib/typemaps/fragments.swg`
```
%fragment("SWIG_Float_Overflow_Check","header",fragment="<float.h>,SWIG_isfinite") %{
/* Accept infinite as a valid float value unless we are unable to check if a value is finite */
#ifdef SWIG_isfinite
# define SWIG_Float_Overflow_Check(X) ((X < -FLT_MAX || X > FLT_MAX) && SWIG_isfinite(X))
#else
# define SWIG_Float_Overflow_Check(X) ((X < -FLT_MAX || X > FLT_MAX))
#endif

```

We refer to float.

But in example `/Examples/python/variables/example_wrap.c`
```
  int res = SWIG_AsVal_double (obj, &v);
  if (SWIG_IsOK(res)) {
    if (SWIG_Float_Overflow_Check(v)) {
```

We use double.
That is the source of the warning.

I think we need to change `SWIG_Float_Overflow_Check` to use double.

What do you suggest?

erezgeva: I  find where this comes from
```
SWIGINTERN int
SWIG_AsVal_float (PyObject * obj, float *val)
{
  double v;
  int res = SWIG_AsVal_double (obj, &v);
  if (SWIG_IsOK(res)) {
    if (SWIG_Float_Overflow_Check(v)) {
      return SWIG_OverflowError;
    } else {
      if (val) *val = (float)(v);
    }
  }  
  return res;
}
```
We need casting in `SWIG_Float_Overflow_Check()`

erezgeva: The warning comes from `SWIG_isfinite(X))`

erezgeva: Using preproccesor the code looks like
```

SWIG_AsVal_float (PyObject * obj, float *val)
{
  double v;
  int res = SWIG_AsVal_double (obj, &v);
  if ((res >= 0)) {
    if (((v < -3.40282346638528859811704183484516925e+38F 
# 4059 "example_wrap.c"
       || v > 3.40282346638528859811704183484516925e+38F
# 4059 "example_wrap.c"
       ) && (
# 4059 "example_wrap.c" 3
       ((__builtin_choose_expr ( __builtin_types_compatible_p (__typeof__ (
# 4059 "example_wrap.c"
       v
# 4059 "example_wrap.c" 3
       ), double), __fpclassify(
# 4059 "example_wrap.c"
       v
# 4059 "example_wrap.c" 3
       ), __builtin_choose_expr ( __builtin_types_compatible_p (__typeof__ (
# 4059 "example_wrap.c"
       v
# 4059 "example_wrap.c" 3
       ), float), __fpclassifyf(
# 4059 "example_wrap.c"
       v
# 4059 "example_wrap.c" 3
       ), __builtin_choose_expr ( __builtin_types_compatible_p (__typeof__ (
# 4059 "example_wrap.c"
       v
# 4059 "example_wrap.c" 3
       ), long double), __fpclassifyl(
# 4059 "example_wrap.c"
       v
# 4059 "example_wrap.c" 3
       ), (__builtin_trap(),0)))) & 0x0100) == 0)
# 4059 "example_wrap.c"
       ))) {
      return -7;
    } else {
      if (val) *val = (float)(v);
```


The Linux equivalent is simply using `__builtin_isfinite(v)`

I do not know how to fix it.

erezgeva: I do not know how to suppress the worming of using  `SWIG_isfinite` with MingW64 GCC to .
I added GCC `#pragma` to suppress the worming with Ming compilers.

erezgeva:
I add `-Werror`  to MSVC `cflags`.
However it fails with `cxxflags`, the code need some C++ cleaning.

.github/workflows/windows.yml
wsfulton: Understanding msys handling of forward slashes is tricky. It tries to only convert paths and something beginning with `/` could be a path. So sometimes it will convert options beginning with `/` and sometimes not. I've never seen it try to convert `/?`. So while `cl -?` is okay, so is `cl /?`. You can't say the same for an option such as `/c` as that could be `c:\`. cccl doesn't try to be clever, it will unconditionally convert `/` to `-` (depending) as per the docs.

erezgeva:
Hi, @wsfulton 

Doing more deep in into MingW64 code:

```
       ((__builtin_choose_expr(
             __builtin_types_compatible_p(__typeof__(v),double), __fpclassify(v),
             __builtin_choose_expr(
             __builtin_types_compatible_p(__typeof__(v),float), __fpclassifyf(v),
             __builtin_choose_expr(
             __builtin_types_compatible_p(__typeof__(v),long double), __fpclassifyl(v),
             (__builtin_trap(),0)))) & 0x0100) == 0)
```
This complex expression is using a 3 variants of [`fpclassify`](https://cppreference.com/w/c/numeric/math/fpclassify.html).
The expression is equivalent to `fpclassify(X) != FP_NAN && fpclassify(X) != FP_INFINITE`.
As `FP_NAN = 0x0100` and `FP_INFINITE = (0x0100 | 0x0400)`.
The 3 variants of `fpclassify` are:
- `__fpclassify` for `double`
- `__fpclassifyf` for `float`
- `__fpclassifyl` for `long double`

The proper `fpclassify` is selected using the `__builtin_choose_expr` build in function.
See: https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html
So the expression always uses the proper `fpclassify` function for the actual type of `v`.
The warning comes from `__fpclassifyf` which uses `float` while  `v` is double.

The proper way to fix the warning is to add cast to `v` inside the 3  `fpclassify`  variants:
- `__fpclassify((double)v)`
- `__fpclassifyf(float)v)`
- `__fpclassifyl((long double)v)`

We should open an issue in mingw64.

We have 2 options:
- Add warning suppression
- Use a different code for `SWIG_isfinite` when using mingw64.

We should add a proper note in `SWIG_isfinite` regarding mingw64.

erezgeva:
Hi @wsfulton,

After further investigation, found the source of the warning.

The MingW-W64 code is here:
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/crt/math.h#l495

There is a clone is GitHub
https://github.com/mingw-w64/mingw-w64/blob/master/mingw-w64-headers/crt/math.h#L495

`#define isfinite(x) ((fpclassify(x) & FP_NAN) == 0)`

`#define __mingw_types_compatible_p(type1, type2) __builtin_types_compatible_p (type1, type2)`

```
#ifdef __cplusplus
#define __mingw_choose_expr(C, E1, E2) ((C) ? E1 : E2)
#else
#define __mingw_choose_expr __builtin_choose_expr
#endif
```

See explenatopn on `fpclassify`:
https://en.cppreference.com/w/c/numeric/math/fpclassify.html
And on GCC `__builtin_choose_expr` and `__builtin_types_compatible_p` here:
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html


The macro `fpclassify(x)` is using the 3 variants:
```
  extern int __cdecl __fpclassifyl (long double);
  extern int __cdecl __fpclassifyf (float);
  extern int __cdecl __fpclassify (double);
```

The expression uses the proper variant based on the type.
The **warning** is that within the unused `__fpclassifyf(x)`, `x` is implicit casting from `double` to `float`.
The fix is to add explicit casting, as done in [`isnan(x)`](https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/crt/math.h#l589
)
i.e. `__fpclassifyf((float)(x))`

We should open an issue in `mingw-w64-headers` `crt/math.h`.

On the meanwhile we can:
- Add warning suppression when using `MingW-W64`.
- Replace  `isfinite()` with `!isnan(x)` which does cast properly,
  See https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/crt/math.h#l589

erezgeva:
I tried to send a fix to the mingw-w64 project.
Not sure if the receive it.

erezgeva:
Hi @wsfulton,


Send patch to `mingw-w64-public`

https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/20260227191657.2500517-2-erezgeva%40nwtime.org/#msg59301955

erezgeva:
Use `cflags+=" -D'SWIG_isfinite(v)=(!isnan((v)))'"` until  `isfinite()` is fixed in Windows GitHub images.

erezgeva:
I am aware that the proper workaround is 
`SWIG_isfinite(v)=(!isnan((v))&&!isinf((v)))`
The problem is that `isinf(v)` is using the same `fpclassify(x)` which will give the warning. So unless someone add a real infinity number, the test will past :smile:

erezgeva:
@wsfulton 

The fix was merged:
https://github.com/mingw-w64/mingw-w64/commit/968fed5fee0475b9cbeca823e1751258593e3474

https://sourceforge.net/p/mingw-w64/mingw-w64/ci/968fed5fee0475b9cbeca823e1751258593e3474/

Now we wait for next update of MSYS2  and GitHub image.

erezgeva:
@wsfulton 

My work paid out :+1: 

7588b7bba3f300223bdbc0f455585adbffeede4b
Pass Linux, but fail on MacOS for warning.
https://github.com/swig/swig/actions/runs/22755288785

The compiler is correct, see number (8):
https://en.cppreference.com/w/cpp/language/namespace.html
This test should be limited to C++17, you may also add a C++17 to one of the Java on MacOS :smile: 

This only show this PR is importance as well.
The warnings filters improve the quality of SWIG.
In addition to `clang-format`.
The next step is to add languages warning filters.
At least for the compiling languages: Java, C#, Go, D.

.github/workflows/windows.yml
```diff
@@ -411,8 +454,25 @@ jobs:
...
+            if [[ "$SWIGLANG" != 'go' ]]; then
+              cflags+=" -D'SWIG_isfinite(v)=(!isnan((v)))'"
+            fi
+          else
+            # cccl do not translate '-Wno' to MSVC 'cl.exe'
+            cflags="-Werror -std=$CSTD"
+            cxxflags="-Werror -std=$CPPSTD"
+          fi
+          # Include for the boost headers
+          cxxflags+=" -I$BOOST_PATH/include"
```
wsfulton: This should instead be done at configure time with `--with-boost`.

erezgeva: Look on configure step.
It is there `--with-boost=/c/msys64/ucrt64` for ucrt.
And similarly with the rest.

Somehow when we define new `CXXFLAGS` the include is dropped.
I did not check why, But as I saw the same issue in macos, I figure it is not related to windows.

erezgeva: Looking on `Examples/Makefile.in`
`CXXFLAGS   = @BOOST_CPPFLAGS@`
If `CXXFLAGS` is defined when calling the make CXXFLAGS=<value>
This line is ignored.
So `@BOOST_CPPFLAGS@ ` is not used.
The reason we did not need it with Linux is that the boost headers are already placed in search path.

erezgeva: Fix.

See new commit.

erezgeva:
Seems new Ruby uses extra semicolons.
Make exception in `Tools/testflags.py` accordingly.

erezgeva:
@wsfulton 
`cpp11_template_pack_specialization` was added recently.
Updating this PR revel a new MSVC warning of implicit casting, which I fix.
This time is was only testing interface file, but next one might be an actual product code.
I think the PR is a good practice to catch more warnings issued by the MSVC.
More compilers improve the project warning coverage.

PR #3350, opened 2026-02-25
Doc: Use -undefined dynamic_lookup on macos

### LINKED ISSUE #3278: -undefined suppress is deprecated on macOS
I encountered the same issue as #2518 and #2060 . I followed the replies in those issues and solved it by adding `-flat_namespace -undefined suppress` flags to the linker. However, I got the following warnings:
```
ld: warning: -undefined suppress is deprecated
ld: warning: -undefined suppress is deprecated
```

I'm using macOS 26 and `ld-1221.4`. 

Is there any recommended workaround or fix that doesn't rely on the deprecated `-undefined suppress` flag? I'm concerned that this flag may be removed in future versions of `ld`, which could break our release builds.

jschueller:
Closes #3278

/cc @erezgeva

### DIFF
--- a/Doc/Manual/Preface.html
+++ b/Doc/Manual/Preface.html
@@ -378,7 +378,7 @@ on OS X.
 Users of OS X should be aware that Darwin handles shared libraries and linking in
 a radically different way than most Unix systems.  In order to test SWIG and run
 the examples, SWIG configures itself to use flat namespaces and to allow undefined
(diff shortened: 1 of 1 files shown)


wsfulton:
Thanks, committed, see 9a7b43d86edbfee16cee8ed0ac04fac57d40acf4.

Issue #3321, opened 2026-01-26
UD in swig/Lib/csharp/std_string_view.i ?


ArtemKozak666:
swig/Lib/csharp/std_string_view.i uses `std::string($1).c_str()` which causes undefined behavior

For example this line:

`%typemap(directorin) const string_view & %{ $input = std::string($1).c_str(); %}`

erezgeva:
The explanation is too short.
The code can be improved, but your claim is not sufficient.
Why is using the `c_str()` function a problem, and where?
The function is used in different places.
You need to be more explicit on the issue.

Do you have an example exhibiting this undefined behaviour?
Can you share the example or some output that show the issue.

If you wish us to fix, we need first to be able to create an example the exhibit the issue, so we can use it to ensure we fix the problem.

ArtemKozak666:
> Why is using the `c_str()` function a problem, and where?

Problem is `std::string($1)` which is temporary value and will be destroyed right after assignment operation, so $input will contain pointer to deleted string.

Using `%typemap(directorin) const string_view & (std::string str) %{ 
str = std::string($1);
$input = str.c_str(); 
%}` must fix that issue, but it's not the only problem.

I will provide code snippets and result later.

erezgeva:
Sound good idea.  

What do you think on submit a PR yourself?

ArtemKozak666:
I created small [repo](https://github.com/ArtemKozak666/swig_test) for this project to not spam all my code in one comment.

When i run swig, i see following code for `Logger::Message` director:
```C++
void SwigDirector_Logger::Message(std::string_view str) {
  const char * jstr  ;
  
  if (!swig_callbackMessage) {
    Swig::DirectorPureVirtualException::raise("Logger::Message");
    return;
  } else {
    jstr = std::string(str).c_str(); 
    swig_callbackMessage(jstr);
  }
}
``` 

Here we can see that invalid code, and visual studio also shows warning here:

<img width="586" height="137" alt="Image" src="https://github.com/user-attachments/assets/eb532cf8-f50b-4c3c-a52c-67898ca9c5a9" />

When running Example program written in C#, i see the following output:
```
From C#: TestLogger created
From C#: ExampleCaller created
SetString called with: Hello, World!
From C++: Hello, World!
From C#:
Retrieved string: Hello, World!
```
So, other typemaps work correctly, but this part of code shows empty string instead of "Hello World!"

[...]

PR #3351, opened 2026-02-25
Fix std::string_view directorin for csharp (#3321)

ArtemKozak666:
- Fixed assignment of deleted string in `Lib/csharp/std_string_view.i`
- Added test to existing `cpp17_director_string_view` (It calls overridden function from c++, and compares it with passed parameter. If string is created in c# from deleted string it may contain random characters from memory instead of passed string)


wsfulton:
Thanks. The temporary std::string is needed as std::string_view may not be null terminated. The previous implementation had the `$input` pointing to deleted memory as the std::string temp had been destroyed.

Problem is C# only as other languages create a language specific string by copying the contents of the string_view instance.

Merged in 9a7b43d86edbfee16cee8ed0ac04fac57d40acf4...d81d060b0aa64b8bb6ecad8d4b8fdc9ac578a24e

PR #3352, opened 2026-02-26
fix: replace 54 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.


wsfulton:
It would be better to replace Exception with the actual Exception, such as AttributeError, TypeError etc. Could you do that instead?

haosenwang1018:
Closing per backlog policy; will revisit with narrower, test-backed patches.

PR #3353, opened 2026-02-26
gcc 13 on Ubuntu 22 is broken, use Ubuntu 24

erezgeva:
GCC comes from Ubuntu toolchain repository
https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/ppa

It seems GCC 13 on Ubuntu 22 is broken, times to move to Ubuntu 24.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -53,7 +53,10 @@ jobs:
           os: ubuntu-22.04
         - SWIGLANG: ""
           GCC: 13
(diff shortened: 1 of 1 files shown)


wsfulton:
let's rather make 24.04 the default if not specified and 22.04 the exception.

wsfulton:
Ignore that, I thought you'd changed most of the builds, but you've updated just the gcc13 builds.

PR #3354, opened 2026-02-26
Fully qualify C# interfaces when using the interface and nspace features

FilipAlg:
When using both the %nspace and %interface features the emitted C# classes that implement interfaces does not fully qualify their implemented interfaces. This causes compiler issues when the interface and the implementer are located in different namespaces. For an example, see the following interface code:

```
%module InterfaceTest

%nspace;

%include <swiginterface.i>

%interface_impl(A::B::Interface);

%inline %{
namespace A::B {
  class Interface {};
}

namespace A {
  class Concrete : public A::B::Interface {};
}
%}
```

This PR updates the C# module code to fully qualify the interface types and updates the C# swiglib code to do the same.

### DIFF
--- a/Examples/test-suite/csharp_namespace_interface.i
+++ b/Examples/test-suite/csharp_namespace_interface.i
@@ -5,7 +5,7 @@
 %include <std_string.i>
 %include <swiginterface.i>
 
-%interface_impl(A::B::Implementer);
+%interface_impl(A::B::Interface);
 
 %inline %{
 #include <string>


wsfulton:
The testcase passes using SWIG master. Can you please double check the example/testcase and make sure it contains problem that is being fixed.

I can see the inheriting from the base class has some C# namespaces added in some cases, so there is a difference, it's just the testcase doesn't seem to test the improvement.

FilipAlg:
Ah yes, I had applied the interface feature to the implementor class instead of the interface class, Ive updated this now and verified that it fails on master now

wsfulton:
Thanks @FilipAlg. I renamed csharp_namespace_interface to csharp_nspace_interface to follow our conventions on testing the nspace feature.

wsfulton:
@FilipAlg, I'm terribly sorry, but I inadvertently lost your authorship in the when I squash merged your changes into master. Your name is in the CHANGES.current file as the contributor as a consolation.

I've replicated your changes to make the equivalent change for Java in 7588b7bba3f300223bdbc0f455585adbffeede4b.

FilipAlg:
@wsfulton No worries whatsoever! Just happy to get the fix in 😄

PR #3361, opened 2026-03-03
More improvements on using cached tools in Linux GHA

erezgeva:
Add matrix variable `SKIP_CACHED_TOOLS` which prevent using a cached tools and forced installing from network.

Can be used in case a cached tool is malfunction, or we want to use the installing from network in addition to the cached tool.


wsfulton:
SKIP_CACHED_TOOLS isn't used so I'm not sure what the point of this is? Maybe for debugging in private tests?

erezgeva:
> SKIP_CACHED_TOOLS isn't used so I'm not sure what the point of this is? Maybe for debugging in private tests?

Probably.
I saw some failure with Ruby 3.3 and thought it could be more easier to compare Ruby 3.3 from cached tools versa installing using the RVM site.

But as you say, it is for debugging.

PR #3365, opened 2026-03-05
Fix descriptor() type lookup when SwigPyIterator is renamed via #define

xSetech:
The hardcoded string literal "swig::SwigPyIterator *" in descriptor() is not expanded for per-module renames of SwigPyIterator. The change allows the new name(s) to be passed into SWIG_TypeQuery().

Example of a real project that defines an iterator per-module:
https://github.com/xSetech/med-file/blob/main/python/pyiterators_881.i
https://github.com/xSetech/med-file/blob/main/python/med_881.i

The change applied in this PR, as seen in pyiterators_881.i above, appears to originate from this bug report:
https://sourceforge.net/p/swig/bugs/881/#5c41

A similar issue is described in #3189.

### DIFF
--- a/Examples/test-suite/python_pyiterator_rename_a.i
+++ b/Examples/test-suite/python_pyiterator_rename_a.i
@@ -10,7 +10,11 @@
 class VecHolder_a {
 public:
   std::vector<int> data;
-  VecHolder_a() : data({10, 20, 30}) {}
+  VecHolder_a() {
+    data.push_back(10);
+    data.push_back(20);
+    data.push_back(30);
+  }
   std::vector<int>& get() { return data; }
 };
 %}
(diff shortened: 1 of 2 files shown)


wsfulton:
The correct approach is to define the template instantiation:
```swig
```
in one just module  and then import this from the modules that need it. Any reason why this is not being done?

xSetech:
> The correct approach is to define the template instantiation in one just module and then import this from the modules that need it. Any reason why this is not being done?

I ran into a project that chose a different approach. This patch provides just a small convenience for projects that don't use the recommended approach, for whatever reason.

wsfulton:
Also the two subsequent commits e82450577da383e3573bed01cbf2c8a2daa6d464 ee5ec1c3c205af83508ea4cae4b95026bb7aefa7 are related.

xSetech:
Thanks!

PR #3367, opened 2026-03-06
[MacOS + Tcl]  GHA: Add C/C++ default standard before configure 

erezgeva:
- Add C/C++ default standard before configure in GHA.
- Fix `cpp11_std_unique_ptr` on MacOS Tcl.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -203,19 +203,14 @@ jobs:
             # GCC front-end is available with versioned only!
             CC="gcc-$GCC"
             CXX="g++-$GCC"
-            CC_NAME="$CC"
-            echo "CC=$CC" >> $GITHUB_ENV
-            echo "CXX=$CXX" >> $GITHUB_ENV
           else # The defualt on MacOS is Clang
             CFLAGS+=" -Wno-deprecated-declarations"
             CXXFLAGS+=" -Wno-deprecated-declarations"
-            # These are the defualt in configure, no need to export
-            CC=cc
-            CXX=c++
-            CC_NAME='clang'
+            CC='clang'
+            CXX='clang++'
           fi
-          # Compiler name for Tools/testflags.py
-          echo "CC_NAME=$CC_NAME" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


erezgeva:
<del>The test `nspace_interface.cpptest` fails on MacOS Java.
Because the test need c++17, while these MacOS Java tests uses c++14.
The  Java test that uses c++20 pass.</del>

- <del>See #3368</del>

Examples/test-suite/tcl/cpp11_std_unique_ptr_runme.tcl
```diff
@@ -230,9 +230,18 @@ if {[moveOverloadTest] != 0} {
 if {[moveOverloadTest "NULL"] != 1} {
   error "moveOverloadTest failed"
 }
+if { $tcl_platform(os) != "Darwin" } {
+# MacOS need a real object
```
wsfulton, line 234: What was the problem here to avoid testing on MacOS? Surely this should work the same on all operating systems, or is there a bug in the MacOS TCL implementation.

erezgeva: I do not know.
Not expert on either, nor MacOS neither Tcl. 
We do not test on Windows, so I can not tell there.

wsfulton: Did you see an error? I'm just trying to understand why this pull request removes testing on MacOS.

erezgeva: Of course.
Why bother else ...

erezgeva: It say that `k` was deleted.
I do not have MacOS, or Windows, I only test on GitHub.

wsfulton: I just thought you'd have the error message. I'll take a look at some point and recreate.

PR #3368, opened 2026-03-06
Fix:  `nspace_interface` to use c++17

erezgeva:
@wsfulton,

Fix 7588b7bba3f300223bdbc0f455585adbffeede4b
Tests on MacOS fail, as test require C++17.

<br>

Rename `nspace_interface` to `cpp17_nspace_interface`.

The test uses nested namespace which are a C++17 extension. See `(8)` at: https://en.cppreference.com/w/cpp/language/namespace.html


erezgeva:
Namespace was fixed.
We can use C++ 98.
No need to this fix :smile:

PR #3370, opened 2026-03-07
improve CCache types

erezgeva:
During testing of CCache on windows (not functional at the moment).
I found some types improvements.

<br>

- Use `size_t` to represent sizes.
  As most of C functions return a size uses this type.
- Use explicit casting on types change.
- Avoid float type for expression result an integer result value.
- remove `int` from `long` , `short` and `unsigned`.

### DIFF
--- a/CCache/ccache.c
+++ b/CCache/ccache.c
@@ -201,7 +201,7 @@ static const char *tmp_string(void)
 }
 
 /* update cached file sizes and count helper function for to_cache() */
-static void to_cache_stats_helper(struct stat *pstat, char *cached_filename, char *tmp_outfiles, int *files_size, int *cached_files_count)
+static void to_cache_stats_helper(struct stat *pstat, char *cached_filename, char *tmp_outfiles, size_t *files_size, int *cached_files_count)
 {
 #if ENABLE_ZLIB
 	/* do an extra stat on the cache file for the size statistics */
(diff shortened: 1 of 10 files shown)


wsfulton:
This code base is rather delicate and can go wrong with subtle bugs, so I am reluctant to change it at all unless there is a bug reported, or a very simple compiler warning fixes. Windows and Visual C++ is not supported and won't be. The real CCache has only just decided to support Windows properly and that is after many many years of development where this original code base is has morphed into something quite unrecognisable to the original 2.4 version that we have here.

Porting the fixes into the real CCache is the only real future goal I see for swig-ccache.

So is there some real fix in here? If not, I'm inclined to reject as unnecessary change.

erezgeva:
> This code base is rather delicate and can go wrong with subtle bugs, so I am reluctant to change it at all unless there is a bug reported, or a very simple compiler warning fixes. Windows and Visual C++ is not supported and won't be. The real CCache has only just decided to support Windows properly and that is after many many years of development where this original code base is has morphed into something quite unrecognisable to the original 2.4 version that we have here.
> 
> Porting the fixes into the real CCache is the only real future goal I see for swig-ccache.
> 
> So is there some real fix in here? If not, I'm inclined to reject as unnecessary change.

The change relate to MSVC warnings.

If you observe carefully.
The changes are quite straight forward and the tests pass for a good reason.

Using `size_t` instead on `int` for sizes can not fail.
It can only improve.

Leave `int` might fall with big numbers as `int` on old 32 bits system will use 31 bits, while `size_t` is unsigned and will use the full 32 bits and more (some 32 bits system use 64 unsigned bits).
Just because you did not see problem until today is not a prof.

The other changes are more compiler be happy than actual change.
Use explicit conversion instead of implicit conversion and omit the `int`.

I do not see any risk here.

But if you do not feel safe, we can skip this.
We should probably upgrade CCache by using the new version from https://github.com/ccache/ccache

I think #3373 is more important, as it allow using CCache on Windows and actually testing on windows in addition to Linux and MacOS.

erezgeva:
We should focus on upgrading to ccache 4
I remove this PR.

PR #3372, opened 2026-03-09
Python: Fix -Wextra-semi

jschueller:
PYTHON_wrap.cxx:3890:43: warning: extra ';' inside a class [-Wextra-semi]
SWIG_GUARD_DECLARATION(swig_mutex_own); <=====

### DIFF
--- a/Tools/testflags.py
+++ b/Tools/testflags.py
@@ -53,6 +53,10 @@ def get_cxxflags(language, std, compiler):
     else:
         cxx_common += " -Wno-deprecated-declarations"
 
+    # not applicable for these languages as headers emit the warning
+    if language not in ["octave", "javascript"]:
+        cxx_common += " -Wextra-semi"
+
     cxxflags = {
              "c":"-Werror " + cxx_common,
         "csharp":"-Werror " + cxx_common,


erezgeva:
The best way to avoid warnings is to add them to `Tools/testflags.py`.

Can you add `-Wextra-sem` there?

To C++ :   `"python":"-Werror " + cxx_common,`

<del>
And if the fix is relevant to C,
 ` "python":"-Werror " + c_common + sign_conversion_flags,`.
</del>

Tools/testflags.py
```diff
@@ -4,7 +4,7 @@
 def get_cflags(language, std, compiler):
     if std == None or len(std) == 0:
         std = "gnu89"
-    c_common = "-fdiagnostics-show-option -std=" + std + " -Wno-long-long -Wreturn-type -Wmissing-field-initializers"
+    c_common = "-fdiagnostics-show-option -std=" + std + " -Wno-long-long -Wreturn-type -Wmissing-field-initializers -Wextra-semi"
```
erezgeva: `error: command-line option ‘-Wextra-semi’ is valid for C++/ObjC++ but not for C `

So two issuse:
- This is the `C` not the `C++`.
  C++ uses the `get_cxxflags` function with `cxx_common` base for all languages.
- The `cxx_common` is used by all languages.
python is just one of many.

wsfulton:
Would you be able to add the warning to all languages in Tools/testflags.py? Probably not much to do as most of the problems were in generic code or common tests.

erezgeva:
A small fix got bigger :smile: 

Excellent work @jschueller :100:

jschueller:
> Would you be able to add the warning to all languages in Tools/testflags.py? Probably not much to do as most of the problems were in generic code or common tests.

yes, I'm trying to enable it for all languages except for octave and javascript which have problems in their headers

Tools/testflags.py
jschueller: ok, added to cxx_common

erezgeva: Great, you can reolve

wsfulton:
All green, so is this ready now?

jschueller:
> All green, so is this ready now?

yes!

wsfulton:
Okay nice. The only thing I would like to see changed is the SWIG_GUARD macros. The code should use them like they used to such as:
```c
SWIG_GUARD_DEFINITION(Director, swig_mutex_own);
```
so a semi-colon at the end looks better.

jschueller:
> Okay nice. The only thing I would like to see changed is the SWIG_GUARD macros. The code should use them like they used to such as:
> 
> ```c
> SWIG_GUARD_DEFINITION(Director, swig_mutex_own);
> ```
> 
> so a semi-colon at the end looks better.

the problem is that it would lead to a dangling semicolon when these expand to nothing:
```
#else /* SWIG_THREADS */

#define SWIG_GUARD(_mutex)
#define SWIG_GUARD_DEFINITION(_cls, _mutex)
#define SWIG_GUARD_DECLARATION(_mutex)

#endif /* SWIG_THREADS */
```
so I propose to keep the code as is

wsfulton:
Thanks, merging shortly.

If you have the motivation, a couple of potential improvements:
1. Add the flag to C testing in the c_common flags. 
2. Suppress push and pop warning suppression around the inclusion of the language headers that have the problem. I'd guess it won't be all Javascript engines.

erezgeva:
> Thanks, merging shortly.
> 
> If you have the motivation, a couple of potential improvements:
> 
> 1. Add the flag to C testing in the c_common flags.

This is a C++ warning.
Does not exist on C.

> 2. Suppress push and pop warning suppression around the inclusion of the language headers that have the problem. I'd guess it won't be all Javascript engines.

wsfulton:
> > 1. Add the flag to C testing in the c_common flags.
> 
> This is a C++ warning. Does not exist on C.

Ah, even better, nothing to do then!

PR #3373, opened 2026-03-09
Build and test CCache with MinGW-w64

erezgeva:
- Update `ccache-swig` to build with MinGW-w64 under windows.
- Update `CCache/test.sh` to properly support the `NOSOFTLINKSTEST`, i.e. work on system lacking symbolic links.
- Update `Examples/test-suite/errors/Makefile.in`.
  - remove the "dos" filter. The `checkout` on GHA add windows line feed to text files. The filter is useless.
  - Adding `--ignore-all-space` to `diff` solve the problem, no further filters are needed.
  - Improve `XXX_TEST_CASES` macros.  Follow `common.mk`, `XXX_TEST_CASES` should not store the extension. Update the comments and use one definition to each `XXX_TEST_CASES`. `SWIGOPT += -doxygen` is defined in `common.mk` for `DOXYGEN_TEST_CASES`, we do the same, no trick!

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -69,7 +69,7 @@ jobs:
 
     # The name of the test follow the tested language
     name: >
-      ${{ matrix.SWIGLANG }} ${{ matrix.VER }}
+      ${{ matrix.SWIGLANG || 'none' }} ${{ matrix.VER }}
       ${{ matrix.SWIG_FEATURES }}
       ${{ matrix.COMPILER || 'msvc' }}
       ${{ matrix.CPPSTD }}
@@ -83,6 +83,13 @@ jobs:
     strategy:
       matrix:
         include:
+        - SWIGLANG: ''
+          INSTALL: 'true'
+          COMPILER: gcc
+        - SWIGLANG: ''
+          INSTALL: 'true'
+          COMPILER: gcc
+          MSYS2_ENV: ucrt64
         - SWIGLANG: csharp
           INSTALL: 'true'
         - SWIGLANG: csharp
@@ -242,9 +249,12 @@ jobs:
             #  The MSYSTEM=MSYS, which supports 'Cygwin', uses the '/usr' path prefix.
             #  We do not use it at the moment.
 
+            # MinGW-w64 package list
+            ming_pkgs+=(binutils make autotools pcre2)
+
             case "$SWIGLANG" in
             python)
-              MORE_MSYS_PKGS+=" ${mingw_variant}python"
+              ming_pkgs+=(python)
               ;;
             ruby)
               if [[ -n "$VER" ]]; then
@@ -252,7 +262,7 @@ jobs:
                 RUBYDIR="$(cygpath -w $ruby_dir)"
                 echo "$RUBYDIR" >> $GITHUB_PATH
               else
-                MORE_MSYS_PKGS+=" ${mingw_variant}ruby"
+                ming_pkgs+=(ruby)
               fi
               ;;
             go)
@@ -267,8 +277,15 @@ jobs:
               ;;
             esac
 
-            # MinGW-w64 packages to install with MSYS2
-            for n in binutils make autotools pcre2 boost; do
+            # More MSYS2 packages to install
+            MORE_MSYS_PKGS+=" base-devel"
+
+            # Boost is needed with languages, we can skip for CCache tests
+            if [[ -n "$SWIGLANG" ]]; then
+              ming_pkgs+=(boost)
+            fi
+            # Add MinGW-w64 packages to MSYS2 packages list
+            for n in "${ming_pkgs[@]}"; do
               MORE_MSYS_PKGS+=" ${mingw_variant}$n"
             done
 
@@ -276,7 +293,7 @@ jobs:
             echo "PCRE2_CFLAGS=-I$mingw_dir/include -DPCRE2_STATIC" >> $GITHUB_ENV
             echo "PCRE2_LIBS=-L$mingw_dir/lib -lpcre2-8" >> $GITHUB_ENV
 
-            echo "MORE_MSYS_PKGS=base-devel $MORE_MSYS_PKGS" >> $GITHUB_ENV
+            echo "MORE_MSYS_PKGS=$MORE_MSYS_PKGS" >> $GITHUB_ENV
             echo "BOOST_PATH=/c/msys64$mingw_dir" >> $GITHUB_ENV
 
             # Set MSYS2 GCC compiler location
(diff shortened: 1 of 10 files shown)


wsfulton:
Somehow the version being output is 4.3.1 instead of 4.5.0: https://github.com/swig/swig/actions/runs/22895200100/job/66433088699?pr=3373#step:14:45.

erezgeva:
> Somehow the version being output is 4.3.1 instead of 4.5.0: https://github.com/swig/swig/actions/runs/22895200100/job/66433088699?pr=3373#step:14:45.

There is another SWIG already install on /c/ProgramData/Chocolatey/bin using the Chocolatey tool.
<del>I'll have a look on the path.</del>

I fix the install.
It was not working on windows.

wsfulton:
1. config_win32.h* files should be deleted from git.
2. errors/Makefile.i: use replace `--ignore-all-space` with `-w` for maximum portability.
3. test.sh: keep to sh for maximum portability
4. gcc used to compile ccache-swig just fine on MinGW, so I'm not sure how many of the changes in ccache.h, util.c are actually necessary. This code base is so stable and widely tested that I would rather make absolute minimal changes. One important goal is to easily see the SWIG additions to the original ccache-2.4 so that they can be easily understood and ported into ccache-4 when someone has the motivation for it. Hence, the desire for minimal changes required for adding in SWIG support to the code base. So, could you please remove any unnecessary changes to minimise the diffs and for any changes, provide some detail as to why they are needed. As I mentioned, I don't think any of these are needed and the builds I release seem fine as is. Sorry if that is bit of a nuisance, but this code base is too delicate and not intimately understood by us. Finally hardly anyone uses ccache-swig, so there is not much point spending much effort on it unless reported as a problem. I have already spent a couple of hours on this which I would have preferred to spend on the real SWIG code base.

erezgeva:
> 1. config_win32.h* files should be deleted from git.

Sorry miss it.

> 2. errors/Makefile.i: use replace `--ignore-all-space` with `-w` for maximum portability.

Not sure if it is more portable, but i will replace.
It is shorter by no means :smile: 

> 3. test.sh: keep to sh for maximum portability

I use `${parameter :offset :length }` to convert link to copy.
It is a bash feature and does not exist in this way in standard shells.
We can remove the `ln` and `call_ln()` and leave `ln` only in the two `testsuite link`.
If you want, I can replace these `ln` with `cp`, remove `call_ln()` and change back to `sh`.

@wsfulton please acknowledge if you want this change.

> 4. gcc used to compile ccache-swig just fine on MinGW, so I'm not sure how many of the changes in ccache.h, util.c are actually necessary. This code base is so stable and widely tested that I would rather make absolute minimal changes. One important goal is to easily see the SWIG additions to the original ccache-2.4 so that they can be easily understood and ported into ccache-4 when someone has the motivation for it. Hence, the desire for minimal changes required for adding in SWIG support to the code base. So, could you please remove any unnecessary changes to minimise the diffs and for any changes, provide some detail as to why they are needed. As I mentioned, I don't think any of these are needed and the builds I release seem fine as is. Sorry if that is bit of a nuisance, but this code base is too delicate and not intimately understood by us. Finally hardly anyone uses ccache-swig, so there is not much point spending much effort on it unless reported as a problem. I have already spent a couple of hours on this which I would have preferred to spend on the real SWIG code base.

Sorry for my response, but you are wrong.
`ccache` was never tested on windows.

Reading both tests which I fix, it is obvious that the `ccache` was never tested on windows.

I do recognise that `ccache` 4 is working on windows, simply we have a very old version that does not works properly on windows.

Someone did put effort to compile it, but there is a different between compiling, testing and working.

So, no I do not think it worked with MinGW in the past.
I did try to focus only on Windows changes, and leave non-windows code out of the `_WIN32` definition, and unchanged.

There is only one change that related to swig flag and windows
```
	/* we might be being invoked like "ccache gcc -c foo.c" */
	if (strcmp(base, MYNAME) == 0
#ifdef _WIN32 /* can be invoked without the `.exe` extension */
	    || strcmp(base, PROGRAM_NAME) == 0
#endif
```
But this is because we define:
`#define MYNAME PROGRAM_NAME ".exe"`

The second change is not related to windows in anyway,
Simply that `CCACHE_SWIG` can be defined empty.
The different of `CCACHE_SWIG=` and `unset CCACHE_SWIG`
We should copy it to the new `ccache`.
```
static void detect_swig(void)
{
	char *basename = str_basename(orig_args->argv[0]);
	char *ccache_swig = getenv("CCACHE_SWIG");
	if (strstr(basename, "swig") || (ccache_swig != NULL && ccache_swig[0] != 0)) {
		swig = 1;
	}
	free(basename);
}
```


The reset are purly windows changes, which are probably already implemented in `ccache` 4 in that way or another.
So no harm done there either.
It does helps up with migratation.
We can tests with MingW in addition to Linux and MacOS :smile: 

Note there are many `ifdef _WIN32` that I did not touch, as I saw no reason to change what is reasonably.

erezgeva:
Replace link with copy outside the `testsuite link` in `test.sh`

erezgeva:
I compare `/CCache` of master to
https://github.com/ccache/ccache/releases/tag/v2.4
I get 3203 lines.

And based on `CCache/debian/changelog`
I also compare to
https://salsa.debian.org/debian/ccache/-/tags/debian%2F2.4-15
I get 3206 lines. The same patch as with original ccache.

Not really sure what clean base you talks about.
You need a cleaning any way,
So any change not related to SWIG do not cost much anyway.
The fixing of the testing will give us a base to perform the upgrading.

Anyway I thing we should start a new separate `swig_ccache` repository based on the original https://github.com/ccache/ccache/ and patch the `swig_ccache` there, so we can follow the main  ccache more easier in the future. Also skip the licence issue, as the swig tool follow swig licenc and swig_ccache follow the ccache licence!

We can leave our `swig_ccache` testing on the swig project as well as the documentation.

erezgeva:
Try to look on all the commits in `master`, there is an initial commit
swig/swig@482d91951a09969d371ec7b5aa3ec65dd9fbccdb
Which is identical to  
https://github.com/ccache/ccache/releases/tag/v2.4

On top of it found Debian patches (which can easily be clean).
But than comes few like swig/swig@8c74fa0f46faf347b936eec202543fc3dd591946

Which make my patch a walk in the park.
The best approach is to clean the diff to https://github.com/ccache/ccache/releases/tag/v2.4 from any windows change and use only SWIG feature.

Perhaps we should open an issue with the diff and discuss there what is relevant and what is not?

wsfulton:
ccache-swig is in minimal maintenance mode as far as I'm concerned, so I'm not going take any undue risk modifying it due to minimal familiarity with the code base.

The strategic approach should be to grab the latest ccache and update it with the SWIG only patches made to the original ccache code base snapshot, then work with the ccache developers to get it added into the official ccache. That is quite a project though and not really on my priority list, but no problem if someone else wants to drive that.

erezgeva:
> ccache-swig is in minimal maintenance mode as far as I'm concerned, so I'm not going take any undue risk modifying it due to minimal familiarity with the code base.
> 
> The strategic approach should be to grab the latest ccache and update it with the SWIG only patches made to the original ccache code base snapshot, then work with the ccache developers to get it added into the official ccache. That is quite a project though and not really on my priority list, but no problem if someone else wants to drive that.

I do understand you concern,
But I do not understand why you merge #3453
That does not consistent.

Any how my recommendation is to split the swig-ccache to a separate repository, based it on https://github.com/ccache/ccache/releases/tag/v2.4
And patch current version there, so we can actually work on it
As I describe in #3376 
Add the current code as patch there.
And proceed from there.

With the current state we will not get anywhere.
Erez

PR #3374, opened 2026-03-10
Remove Lua thread test

erezgeva:
Lua do not support multithreading.
Only Coroutines, see:
https://www.lua.org/pil/9.html


wsfulton:
The threading is implemented in C++ code so might be okay if called from Lua. Was this test flaky / unreliable in some way?

erezgeva:
> The threading is implemented in C++ code so might be okay if called from Lua. Was this test flaky / unreliable in some way?

It is unreliable, because the new created thread and the main running thread uses the same Lua environment `L` without any proper protection.

As I mention, when I add the test, I was not thinking on multi-threads, I was just trying to implement director tests in Lua.

Python have special means to handle this situation, the `%nothread`.
Which Lua do not implement.

Since Lua unlike Python, do not support threads, I see no reason to confuse the Lua users, and just skip the whole notation of threads with Lua.

At least if no user ask for it.

PR #3377, opened 2026-03-14
Ruby: Remove SwigGCReferences to simplify GC_VALUE and allow RUBY_TYPED_FREE_IMMEDIATELY

larskanis:
Actually the hash is not needed at all.
`rb_gc_register_address()` can register one and the same address multiple times. `rb_gc_unregister_address()` removes only the last added address when called, so that remaining registrations keep valid. This implicit works like a reference counting.

`SwigGCReferences` was implemented with `rb_hash_*` functions, but these functions are not allowed to be called in mark and free functions, when flag `RUBY_TYPED_FREE_IMMEDIATELY` is set. Since this patch get rid of `rb_hash_aref` and `rb_hash_delete` from the free function, it allows to enable `RUBY_TYPED_FREE_IMMEDIATELY` and `RUBY_TYPED_EMBEDDABLE` flags of the typed API when #7440 is merged.

### DIFF
--- a/Lib/ruby/rubyclasses.swg
+++ b/Lib/ruby/rubyclasses.swg
@@ -31,69 +31,11 @@
       %template(RubyVector) std::vector<swig::GC_VALUE>;
 
    and all the proper typemaps will be used.
-   
+
 */
 
 %fragment("GC_VALUE_definition","header") {
 namespace swig {
-  class SwigGCReferences {
-    VALUE _hash;
-
-    SwigGCReferences() : _hash(Qnil) {
-    }
-    ~SwigGCReferences() {
-      if (_hash != Qnil)
-        rb_gc_unregister_address(&_hash);
-    }
-    static void EndProcHandler(VALUE) {
-      // Ruby interpreter ending - _hash can no longer be accessed.
-      SwigGCReferences &s_references = instance();
-      s_references._hash = Qnil;
-    }
-  public:
-    static SwigGCReferences& instance() {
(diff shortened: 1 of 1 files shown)


larskanis:
Rice uses the same mechanism instead of reference counting: https://github.com/ruby-rice/rice/blob/1bd4e730ed35834344d1d10b1b65148398659d55/rice/detail/Anchor.ipp#L10
And here is the description: https://github.com/ruby-rice/rice/blob/1bd4e730ed35834344d1d10b1b65148398659d55/rice/detail/Anchor.hpp

wsfulton:
Hi @larskanis 554b3d6a1fb5d1c8c6381c98e99dc5aad45068da fix for #3385 removes rb_hash_* and replaces with an st_table. Can you please revisit the need for this merge request now in the light of that and your contributions to the removal of the deprecated non-typed API, details all in in #3326.

larskanis:
This merge request is still valid, since ruby has functions for this use case `rb_gc_register_address()` / `rb_gc_unregister_address()`. I highlighted the use case to register one address multiple times and added related tests here: https://github.com/ruby/ruby/pull/16388

However the above use case wasn't acknowledged by the ruby core team and the ruby internal implementation is backed by a linked list, whereas swig's is backed by `st_table`. A linked list can get a performance issue if there are thousands of tagged objects. In that case I think it would be best to change the ruby internal list to `st_table`, but since this didn't happen so far, I think it's best to keep the swig's st_table as it is now.

Issue #3380, opened 2026-03-20
[Python] `cdata` fails with new python


erezgeva:
Looking on `SWIG_FromCharPtrAndSize` in `Lib/python/pystrings.swg`

It may use UTF8 decoding which breaks `cdata`.
`cdata` must not use UTF8 as the string represent a binary not a real string!

This is why Java, C#, D and go switch to byte array.
Perhaps Python need to do the same?
But for sure it must not use UTF8 decoding!

```
%#if PY_VERSION_HEX >= 0x03000000
%#if defined(SWIG_PYTHON_STRICT_BYTE_CHAR)
      return PyBytes_FromStringAndSize(carray, %numeric_cast(size, Py_ssize_t));
%#else
      return PyUnicode_DecodeUTF8(carray, %numeric_cast(size, Py_ssize_t), "surrogateescape");
%#endif
%#else
      return PyString_FromStringAndSize(carray, %numeric_cast(size, Py_ssize_t));
%#endif
```

erezgeva:
Hi @wsfulton 

Looking on `Lib/cdata,i`
For `javascript`, `octave`, `perl5`, `ruby`, `python`, `scilab`, `tcl`

All uses `SWIG_AsCharPtrAndSize` `SWIG_FromCharPtrAndSize`.
`SWIG_AsCharPtrAndSize` already uses the `SWIG_NEWOBJ`  in `int alloc`. We can replace it with `flags` bitmask. `SWIG_NEWOBJ` can be used in bitmask.

We can add a similar `flags` parameters to `SWIG_FromCharPtrAndSize`.

And we can add a new flag `SWIG_RAWSTR` which will prevent using UTF8 conversions.

What do you think?

It seems more feasible to the alternative of providing new functions, or add extra parameter.

wsfulton:
Yes, #3383 implements this and I've been reviewing it.

wsfulton:
Looks like this can be closed now with #3383 being merged/implemented. @erezgeva please confirm and close.

erezgeva:
> e this can

Indeed, it does.

PR #3383, opened 2026-03-22
Add new test `li_cdata_bytes`

erezgeva:
This new `cdata` test focus on two aspects
- Ensure we can receive proper data from C/C++ and pass proper data back to C/C++.
- Use all byte values , i.e. the full range of 0 to 255.

<br>

Fix guile `cdata`

<br>

There are two potential issues:
- Some languages treat value 0 as string null terminator.
- Treat strings as Unicode, i.e. values 128-255 do not pass as is.

<br><br>

#### Other languages issues:
See  #3380
The `SWIG_AsCharPtrAndSize` and `SWIG_FromCharPtrAndSize` of some languages need support of raw data versa using Unicode encoding (UTF-8).
I think we need to add new flag to select raw data coping.
### Tcl
receive proper data from C/C++ 
But building  data to pass to C/C++ fails, Tcl uses Unicode by default however converted string to ISO 8859-1/Latin-1 fail too, further investigation required.
### scilab
Language uses Unicode by default.
The `SWIG_AsCharPtrAndSize` and `SWIG_FromCharPtrAndSize` may need updating.
### javascript
Language uses Unicode by default.
The `SWIG_AsCharPtrAndSize` and `SWIG_FromCharPtrAndSize` may need updating.
Alternativly we can switch and use [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array)  
@mmomtchev What do you think of using `cdata.i` with `Uint8Array`?
**Update** After switching to `Uint8Array` all test pass :smile: 
### python
The `SWIG_AsCharPtrAndSize` and `SWIG_FromCharPtrAndSize` need updating.
For python 3, we can use the new [bytes](https://docs.python.org/3.0/reference/lexical_analysis.html#strings) type.
@wsfulton what do you think on using `cdata.i` with python3 `bytes`
Is python3 `bytes` storage different that strings storage?
### <del>ocaml</del>
<del>There is a basic test `li_cdata`. I do not know how to translate the `li_cdata_bytes` to ocaml. This language syntax requires more learning.
@wsfulton do we have any active developer capable writing test scripts in ocaml?
</del>
### R
There are no `cdata` tests in this language.
Perhaps we will get there?
### c
The generated code of `cdata.i` does not pass compilation.
The tests were added to `FAILING_C_TESTS` and `FAILING_CPP_TESTS`

### DIFF
--- a/Lib/javascript/napi/javascriptuint8array.swg
+++ b/Lib/javascript/napi/javascriptuint8array.swg
@@ -1,0 +1,58 @@
+
+/* ------------------------------------------------------------
+ *  utility methods for Uint8Array
+ * ------------------------------------------------------------ */
+
+%fragment("SWIG_FromUint8Array", "header") {
+SWIGINTERN int
+SWIG_FromUint8Array(Napi::Value valRef, void** cptr, size_t* psize, int *alloc)
+{
+  if (alloc == NULL || psize == NULL || cptr == NULL)
+    return SWIG_TypeError;
+  *alloc = 0;
+  *cptr = NULL;
+  *psize = 0;
+  if (valRef.IsTypedArray()) {
+    Napi::Number num;
+    size_t len = 0;
+    Napi::TypedArray arr = valRef.As<Napi::TypedArray>();
+    if (arr.TypedArrayType() != napi_uint8_array) {
+      return SWIG_TypeError;
+    }
+    len = arr.ElementLength();
+    if (len > 0) {
+      size_t i;
+      unsigned char *cbuf = %new_array(len, unsigned char);
+      if (cbuf == NULL) {
+        return SWIG_TypeError;
+      }
+      *alloc = SWIG_NEWOBJ;
+      *cptr = (void*)cbuf;
+      *psize = len;
+      for (i = 0; i < len; i++) {
+        NAPI_CHECK_RESULT(arr.Get(i).ToNumber(), num);
+        cbuf[i] = (unsigned char)num.Uint32Value();
+      }
+    }
+    return SWIG_OK;
+  }
+  return valRef.IsNull() || valRef.IsUndefined() ? SWIG_OK : SWIG_TypeError;
+}
+}
+
+%fragment("SWIG_ToUint8Array", "header") {
+#define SWIG_ToUint8Array(buf, len) SWIG_EnvToUint8Array(env, buf, len)
+SWIGINTERN Napi::Value SWIG_EnvToUint8Array(Napi::Env env, void* buf, size_t len)
+{
+  if (buf != NULL && len > 0) {
+    size_t i;
+    unsigned char* cbuf = (unsigned char*)buf;
+    Napi::Uint8Array arr = Napi::Uint8Array::New(env, len);
+    for (i = 0; i < len; i++) {
+      arr.Set(i, cbuf[i]);
+    }
+    return arr;
+  }
+  return env.Undefined();
+}
+}
--- a/configure.ac
+++ b/configure.ac
@@ -1302,6 +1302,15 @@ else
           AC_SUBST(NODENAPI_DIR)
         fi
       fi
+      # Try Debian/Ubuntu 'node-addon-api' package location
+      if test -z "$JSNAPIENABLED" && which dpkg > /dev/null 2>&1; then
+        napi_dir="$(dpkg -L node-addon-api 2> /dev/null | grep '\/napi.h$' | head -1 | sed 's%/napi.h%%')"
+        if test -n "$napi_dir" && test -f "$napi_dir/napi.h"; then
+          NODENAPI_DIR="$napi_dir"
+          JSNAPIENABLED=1
+          AC_SUBST(NODENAPI_DIR)
+        fi
+      fi
       if test -n "$JSNAPIENABLED"; then
         AC_MSG_RESULT([$NODENAPI_DIR])
       else
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,32 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.5.0 (in progress)
 ===========================
 
+2026-05-24: wsfulton
+            Fixed a one-byte buffer overflow in the cdata.i memmove typemap that
+            occurred when the data length exactly matched the destination buffer size.
+
+2026-05-20: erezgeva
+            [Guile, JavaScript, Python, Scilab, Tcl] #3383 The cdata.i library
+            functions (cdata, memmove and similar) now use a binary data type
+            instead of a string, so that all byte values 0-255 round-trip
+            correctly.
+
+            The target language type used by the cdata.i functions has changed:
+
+              - Python:     bytes object (was str)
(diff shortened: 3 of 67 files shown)


mmomtchev:
In JavaScript strings are always UTF-16 and accessing a string as an array gives 16 bit values.

`UInt8Array` is a view of a contiguous memory buffer.

Only `Uint8Array` would permit the behaviour of `memmove` as described in the manual.

It is not possible for a function to receive a reference to a string and then to modify this string. Strings are a primitive data types and are always immutable. They behave as if they were passed by value (in fact, they are passed without copying by reference but since they are immutable one can only change the reference). An `Uint8Array` is an object and if passed to a function, it can modify its contents.

erezgeva:
> immutable

immutability is less problematic, as we can create a new storage before we pass.
But using UTF-16 means each value above 127 is translated, which make it improper for `cdata`.
Thanks for the clarification, it does make things more clear.
We need to move to  `Uint8Array` and join Java, C#, Go and D.
Perhaps Javascript is related to Java after all :laughing:

erezgeva:
Hi @mmomtchev,

Can you review JavaScript changes?

Thanks
Erez

erezgeva:
Hi @ZackerySpytz,

Manaage to write the test script.
This language syntax is challenging :sweat:

Lib/javascript/napi/javascriptuint8array.swg
```diff
@@ -0,0 +1,60 @@
...
+  *cptr = NULL;
+  *psize = 0;
+  if (valRef.IsTypedArray()) {
+    Napi::Number num;
+    size_t len = 0;
+    Napi::TypedArray arr = valRef.As<Napi::TypedArray>();
+    if (arr.TypedArrayType() != napi_uint8_array) {
+      return SWIG_TypeError;
+    }
+    while (arr.Has(len)) {
```
mmomtchev: `TypedArray` has `TypedArray::ElementLength`

configure.ac
```diff
@@ -1302,6 +1302,12 @@ else
           AC_SUBST(NODENAPI_DIR)
         fi
       fi
+      # Debian 'node-addon-api' package location
```
mmomtchev: This is a very strange decision by Debian, how is one supposed to find this location? It must always be hard coded?

Normally, real world native Node.js modules won't use an external `node-addon-api` installation, a real module will have it in its dependencies and the build system (whether it is Node.js' original `node-gyp` or my own `hadron`) will know how to link it. SWIG bypasses this system in order to avoid dealing with separate packages for each unit test. There is a custom installation in the CI setup.

Not really a request for changes - since it allows a Debian/Ubuntu user to run the unit tests while skipping this custom CI installation, just a rant, I will check with Debian how they intend this to be used. Especially since the installation is in `/usr/share` and not `/usr/include` where it logically belongs.

Lib/javascript/napi/javascriptuint8array.swg
erezgeva: Updated.

configure.ac
erezgeva: Is that better?
````
if test -z "$JSNAPIENABLED" && which dpkg > /dev/null 2>&1; then
  path="$(dpkg -L node-addon-api 2> /dev/null | grep '\/napi.h$' | sed 's%/napi.h%%')"
  if test -n "$path" && test -f "$path/napi.h"; then
    NODENAPI_DIR="$path"
    JSNAPIENABLED=1
    AC_SUBST(NODENAPI_DIR)
  fi
fi
````

erezgeva:
Update on languages status

Langause  |cdata object    |mutability
----------|----------------|-----------
ruby      |`""`            |yes
php       |`""`            |yes
guile     |`""`            |yes
lua       |`""`            |not
perl5     |`""`            |not
ocaml     |`""`            |not
octave    |`""`            |not
python    |**`b""`**           |not
c#        |`byte[]`        |yes
java      |`byte[]`        |yes
d         |`ubyte[]`       |yes
go        |`[]byte`        |yes
javascript|**`Uint8Array`**    |yes
scilab    |**list of `uint8`** |yes
tcl       |**list of integers**|yes

### Changes in this PR
- `guile` was fixed.
- `python` uses binary string.
- `javascript` uses `Uint8Array`.
- `scilab` uses list of `uint8` (but allow passing list of numbers to C).
- `tcl` uses list of integers.

### Broken

- `R` breaks on runtime with "_p_void" is not a defined class.
- `cdata.i` is irrelevant for `c`.
   However there is `Lib/c/cdata.i` and `li_cdata_bytes_cpp`
   fails on build.

mmomtchev:
Mr. Geva, let me be perfectly clear. I got two other simultaneous notifications with yours related to my criminal affair. I have been out of job for 6 years, I have spent 5 years living on social welfare with a few bucks per day, I have spent 15 months living on the street. Obviously you do not have a personal problem taking part in this extortion, I understand it. But I guarantee you that there is absolutely nothing this world that can make me change my mind. I won't live in a society where criminal extortion with the legal system for personal problems is allowed and I will never make any compromise with this affair.

erezgeva:
> Mr. Geva, let me be perfectly clear. I got two other simultaneous notifications with yours related to my criminal affair. I have been out of job for 6 years, I have spent 5 years living on social welfare with a few bucks per day, I have spent 15 months living on the street. Obviously you do not have a personal problem taking part in this extortion, I understand it. But I guarantee you that there is absolutely nothing this world that can make me change my mind. I won't live in a society where criminal extortion with the legal system for personal problems is allowed and I will never make any compromise with this affair.

Sorry, no idea what you talk about.
I am only participate in SWIG project on professional base.

wsfulton:
I've added some changes and modified your original commit.

1. Are you okay if I squash merge "Rename bin abbreviation to binary in cdata identifiers" commit into yours for the final merge?
2. Note that I've removed all backticks in your commit messages as commit messages should be plain text, not Markdown.
3. Please review the last 2 commits too and comment.

erezgeva:
> I've added some changes and modified your original commit.
> 
> 1. Are you okay if I squash merge "Rename bin abbreviation to binary in cdata identifiers" commit into yours for the final merge?

Sure, I agree.

> 2. Note that I've removed all backticks in your commit messages as commit messages should be plain text, not Markdown.

No problem, I do not mind using plain text over markdown.
It simply looks nicer on GitHub. I guess on git it is better using plain text.
But I do not have any preference on the matter. 

> 3. Please review the last 2 commits too and comment.

I review now, probably finish by tomorrow.

Thanks for your review and addition commits.
Erez

erezgeva:
I push changes after review.
You may squash the review as well.

In`li_cdata_bytes_cpp_runme.pl`,   `li_cdata_bytes_runme.pl`
After removing the condition, there is no need for using `eval`.

In `Lib/perl5/perlstrings.swg`, `Lib/ruby/rubystrings.swg`
As `alloc` uses the `SWIG_BINSTRMASK` flag.
We should use `SWIG_IsNewObj(*alloc)` instead of using `*alloc == SWIG_NEWOBJ`.

erezgeva:
The PR is merged

Thanks @wsfulton

PR #3389, opened 2026-03-25
Python: Annotate types of constants

Nerixyz:
Constants in modules did not have type annotations. With this PR, they're added. It uses the same functionality as for annotating variables and thus respects `python:annotations:novar` too.


wsfulton:
Thanks, added in 834bfa580afec49a3fdbeb3a85d2dd6c7b65d9af

PR #3390, opened 2026-03-25
Python: Show class typenames in PEP 484 annotations

Nerixyz:
A followup to #3334. This adds `$pytypename` which is similar to `$javaclassname` and `$csclassname`. It expands to the wrapped typename in Python.

If the typename does not have a corresponding Python class, an opaque name is used similar to Java or C#: `void*` ⇾ `SWIGTYPE_p_void`. To ensure that type checkers understand this, an empty class is generated:
```python
class SWIGTYPE_p_void(object):
    """Opaque ``void *``.
    This type is only used for type annotations and does not exist at runtime."""
    ...
```

Both C-arrays and pointers are annotated with `typing.Optional[$pytypename]`, because they use `None` for `nullptr`.

Towards https://github.com/swig/swig/issues/735.


Nerixyz:
Hmm, I noticed that my typemap for const references is wrong. They're handled like values. So `const int &` should be annotated as `"int"`. I thought that this would need the following line:

```
%typemap(pytyping) SWIGTYPE const & "$typemap(pytyping, $*1_ltype)"
```
However, that's not enough, because the replacements are done at the end and use the type from the start. If we pass a struct `Foo const &`, then we do the following lookup:

```
  Looking for: Foo const &
  Looking for: Foo &foo
  Looking for: Foo &
  Looking for: SWIGTYPE const &foo
  Looking for: SWIGTYPE const &
  Using: %typemap(pytyping) SWIGTYPE const &
  Containing: $typemap(pytyping, Foo)
../../../../Lib/python/pytyping.swg:11: Searching for a suitable 'pytyping' typemap for: Foo
  Looking for: Foo
  Looking for: SWIGTYPE
  Using: %typemap(pytyping) SWIGTYPE
```

So in the end, we'll get `"$&pytypename"` (the result of the last typemap). But we will handle this variable with the initial type (`Foo const&`) instead of the type we got from the last lookup (`Foo`). So we'll end up searching for `Foo &*` instead of `Foo *`.

Is there a way of either performing the replacement of `$pytypename` and friends during the search or a way of keeping the type name in the variable (e.g. `$pytypename{type}`)?

Nerixyz:
> Is there a way of either performing the replacement of `$pytypename` and friends during the search or a way of keeping the type name in the variable (e.g. `$pytypename{type}`)?

Looks like I forgot to overwrite `Language::replaceSpecialVariables`.

Nerixyz:
Ping

---

I think I finally understood what I did wrong with `T const &`. I didn't know that Swig only defines `T` and `T const &` for primitive types in https://github.com/swig/swig/blob/8b307eb27900a4375d4e7b6c5aaa852c62c54c5b/Lib/typemaps/primtypes.swg. So doing `%typemap(pytyping) SWIGTYPE const & "$typemap(pytyping, $*1_ltype)"` is wrong for non-primitive types.

I changed the primitive types to map both `T` and `T const &`. I'll make a separate PR for this, since it's not directly related.

wsfulton:
I've added an additional commit to this branch with explanation in the commit message. Please take a look and let me have any comments before I push it as it is if okay with you.

BTW, `-builtin` doesn't support annotations and I suppose the way forward is to generate a .pyi file for -builtin and probably also for non-builtin. I was wondering what your focus will be next as the current refinements are making the annotations quite polished. Probably best to open a separate discussion for tracking it though.

Nerixyz:
> I've added an additional commit to this branch with explanation in the commit message. Please take a look and let me have any comments before I push it as it is if okay with you.

The changes look good to me, thank you.

> BTW, `-builtin` doesn't support annotations and I suppose the way forward is to generate a .pyi file for -builtin and probably also for non-builtin.

Yes, generating annotations for `-builtin` is important. I'm not sure if the doc comments need to be included there as well - probably depends on the editor/LSP server. Should the annotations only show in the .pyi file? 

> I was wondering what your focus will be next as the current refinements are making the annotations quite polished. Probably best to open a separate discussion for tracking it though.

I have some minor additions and one fix in mind. I'll make some issues about the minor things.

wsfulton:
does this need any more work now that #3408 has been merged?

Nerixyz:
> does this need any more work now that #3408 has been merged?

No, this is ready. I don't know if you want to merge it or if I should do it. The conflicts are simple to resolve. Tested it locally, and the only change I needed was:
```diff
diff --git a/Examples/test-suite/python/python_annotations_typing_runme.py b/Examples/test-suite/python/python_annotations_typing_runme.py
index f43b87019..ed94b97e6 100644
--- a/Examples/test-suite/python/python_annotations_typing_runme.py
+++ b/Examples/test-suite/python/python_annotations_typing_runme.py
@@ -110,7 +110,9 @@ if sys.version_info[0:2] >= (3, 2):
 
         # long double has no in/out typemaps of its own, so it is wrapped as a pointer
         anno = get_annotations(argcheck_long_double)
-        if anno != make_argcheck("typing.Any", ["a_ldouble", "a_ldouble_cref"]):
+        if anno != make_argcheck(
+            "SWIGTYPE_p_long_double", ["a_ldouble", "a_ldouble_cref"]
+        ):
             raise RuntimeError("annotations mismatch: {}".format(anno))
 
         anno = get_annotations(argcheck_complex)
```
since the type annotation for `long double` is now `SWIGTYPE_p_long_double`.

Then, `check-python-test-suite` passes.

wsfulton (changes requested):
# Review: PR #3390 — Python: Show class typenames in PEP 484 annotations
I kinda apologise for what follows, but this is the rather terrifyingly incisive and detailed AI (mostly - I did edit it) review...

## Overview

Adds a `$pytypename` special variable to the `pytyping` typemap, modelled on `$javaclassname` / `$csclassname`, and rewires the default `pytyping` typemaps so that `SWIGTYPE`, `SWIGTYPE *`, `SWIGTYPE &`, `SWIGTYPE &&`, `SWIGTYPE []` and `enum SWIGTYPE` produce a real name instead of `typing.Any`. Types with no proxy class get an opaque `SWIGTYPE_p_xxx` name, backed by empty placeholder classes emitted at the end of the `.py` under `if ... typing.TYPE_CHECKING:`.

The direction is right and it is the natural completion of previous PEP 484 work (5c35726a8 → 13fb1c108 → ddb72429f → 42942f4ca → 3ce7a62a8 → 50df6ced1). The `substitutePytypingVars()` call is placed in `lookupPytyping()` — a single choke point that covers parameters, return types, member variables and `%constant` in one go. That is cleaner than java.cxx, which scatters ~20 `substituteClassname()` calls. Verified: member variables get `f: "Foo"` / `fp: "typing.Optional[Foo]"`, and `%constant Foo *` gets `CFOO: "typing.Optional[Foo]"`.

**How this was verified.** The PR head does not build against today's master, so its `python.cxx` changes and the six new `SWIGTYPE`/`enum` typemap lines were applied onto current master (leaving master's primitive typemaps intact — the realistic post-rebase state), built, and probed. All 949 `Examples/test-suite/*.i` were also swept through both binaries: the only behavioural difference is `python_annotations_typing.py`. So the change is completely inert when `python:annotations` is off, which is the default. Blast radius is opt-in users only.

---

## Blocking

### 1. The PR is out of date; `pytyping.swg` is largely superseded

`gh pr view` reports `mergeable: CONFLICTING`. Master gained 42942f4ca (2026-04-27), 3ce7a62a8 (2026-07-06) and **50df6ced1 (2026-07-23)**, which contains an evolved version of this PR's primitive-typemap work.

Nearly everything in the PR's `pytyping.swg` must be dropped on rebase, keeping master's:

- `%_pytyping_primtype` macro → master uses explicit grouped `%typemap` lists.
- `std::string`, `std::wstring` → already in `std_string.i` / `std_wstring.i`.
- `std::complex<...>` → already in `std_complex.i`.
- `wchar_t`, `wchar_t *` → moved to `pywstrings.swg`.
- `float _Complex` etc. → moved to `ccomplex.i`.
- `long double "float"` → **deliberately deleted** by 50df6ced1; re-adding it reintroduces a bug just fixed.
- `char *` → master has `typing.Optional[str]`, the PR has `str`.
- `std::size_t` / `std::ptrdiff_t` → now picked up via `%apply` since pytyping.swg moved ahead of the Unified Typemap Library.

Only these six lines are genuinely new and should survive:

```
%typemap(pytyping) SWIGTYPE      "$&pytypename"
%typemap(pytyping) SWIGTYPE []   "typing.Optional[$pytypename]"
%typemap(pytyping) SWIGTYPE *    "typing.Optional[$pytypename]"
%typemap(pytyping) SWIGTYPE &    "$pytypename"
%typemap(pytyping) SWIGTYPE &&   "$pytypename"
%typemap(pytyping) enum SWIGTYPE "int"
```

Also: per 50df6ced1's own principle, `enum SWIGTYPE`'s `in`/`out` typemaps come from `typemaps/enumint.swg`, so its `pytyping` map arguably belongs next to where Python includes that, not in `pytyping.swg`.

### 2. The `SWIGTYPE` default change makes annotations confidently wrong, not merely uninformative

50df6ced1's stated invariant is: *a `pytyping` typemap must be defined alongside the `in`/`out` typemaps for the type, otherwise a type wrapped as a pointer to an opaque type is annotated as though it were a native Python type.* This PR does not violate that rule — it inverts the failure mode. Any type that **has** a converting `in` typemap but **no** `pytyping` typemap now gets a proxy/opaque name instead of `typing.Any`.

All of these are verified output from the patched build:

| Declaration | master | this PR | actually accepted |
|---|---|---|---|
| `void take_pyobj(PyObject *o)` | `typing.Any` | `typing.Optional[SWIGTYPE_p_PyObject]` | **anything** |
| `void take_vec(std::vector<int>)` | `typing.Any` | `IntVector` | `IntVector` **or any sequence** |
| `void take_map(std::map<std::string,int>)` | `typing.Any` | `SMap` | `SMap` **or any dict** |
| `void take_sv(std::string_view)` | `typing.Any` | `SWIGTYPE_p_std__string_view` | `str` |
| `%shared_ptr(T)` + `std::shared_ptr<T>` | `typing.Any` | `SWIGTYPE_p_std__shared_ptrT_T_t` | `T` |

`PyObject *` is the worst case — it now claims an opaque pointer type that no user can ever construct.

The `Type *INPUT / *OUTPUT / *INOUT` typemaps are deliberately left out of this table: they are already identified follow-up work, not a defect of this PR. @Nerixyz raised them in the #3408 thread (*"I didn't include pytyping maps for `Type *OUTPUT, Type *INPUT, Type *INOUT` yet"*) and the agreed resolution there was to handle them in a separate PR (*"Add them to a new pull request please. Once you start looking, you'll find tons of library files that require the pytyping typemaps."*). The point of this section is narrower and about **ordering**, not about those specific typemaps: this PR flips the `SWIGTYPE` default *before* that agreed library sweep has happened, which converts every not-yet-mapped type from a silent `typing.Any` into a confidently wrong name.

`typing.Any` is *silent*; a wrong nominal type is *loud and wrong* — a checker will reject correct code. The safe ordering is to land the agreed library sweep first, then flip the default. The Python library files that define `in` typemaps but no `pytyping` typemap — the "tons of library files" anticipated in that thread:

```
Lib/typemaps/inoutlist.swg     (INPUT / OUTPUT / INOUT — already tracked for a follow-up PR, #3408)
Lib/typemaps/swigobject.swg    (PyObject *)
Lib/typemaps/strings.swg
Lib/typemaps/swigtype.swg
Lib/python/pycontainer.swg     (all STL containers)
Lib/python/boost_shared_ptr.i  (%shared_ptr)
Lib/python/std_string_view.i
```

Containers want a union, using the `out=` attribute the feature already supports:

```
%typemap(pytyping, out="IntVector") std::vector<int> "typing.Union[IntVector, typing.Sequence[int]]"
```

### 3. `typing.get_type_hints()` now raises `NameError`

`typing.get_type_hints()` is the standard-library function (from Python's built-in [`typing`](https://docs.python.org/3/library/typing.html#typing.get_type_hints) module) that resolves a function's annotations into real type objects. SWIG emits annotations as string literals (e.g. `"typing.Optional[SWIGTYPE_p_void]"`), so they are only evaluated when a consumer resolves them — and this PR's opaque `SWIGTYPE_*` names exist only inside the `if typing.TYPE_CHECKING:` block, i.e. never at runtime. Resolution therefore fails. This is inherent to the placeholder-class design and affects any opaque type — it is independent of the deferred INPUT/OUTPUT library work in section 2; the example below uses a plain `void *` so nothing hinges on a missing library typemap.

Complete, self-contained reproduction (paste and run):

```console
$ cat > e.i <<'EOF'
%module e
%feature("python:annotations", "typing");
%inline %{
struct Foo { int a; };
void take_foo(Foo *fp) {}
void *make_handle() { return 0; }
%}
EOF
$ swig -python -c++ e.i
$ g++ -fPIC -shared \
    -I"$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["include"])')" \
    e_wrap.cxx -o _e.so
$ python3 -c 'import typing, e; print(typing.get_type_hints(e.take_foo))'
{'fp': e.Foo | None, 'return': <class 'NoneType'>}
$ python3 -c 'import typing, e; print(typing.get_type_hints(e.make_handle))'
NameError: name 'SWIGTYPE_p_void' is not defined
```

`take_foo` resolves because `Foo` is a real proxy class; `make_handle` returns `void *`, whose annotation refers to `SWIGTYPE_p_void`, which only exists under `TYPE_CHECKING`.

Annotations are string literals, so `import e` still succeeds — but every runtime consumer of annotations breaks on any function touching an opaque type: `typing.get_type_hints()` itself, `inspect.signature(func, eval_str=True)` (the stdlib call that evaluates annotation strings), and the libraries built on them (dataclasses, pydantic, FastAPI, Sphinx `autodoc_typehints`). Under master these all resolved to `typing.Any`, which is harmless.

Note that Java/C#/D — which this PR models itself on — emit **real, runtime** type wrapper classes (`SWIGTYPE_p_int.java`). Emitting these as plain runtime classes would keep `get_type_hints()` working, cost a handful of empty classes per module, and be more consistent with the languages being imitated. The counter-argument (a real class is a lie for `isinstance`) applies equally to the `TYPE_CHECKING` stub. At minimum this needs a documented limitation.

### 4. Cross-module class names are wrong with `%module(package=...)`

`getProxyClassLocalName()` hand-rolls the qualified name as `modname.symname`, dropping the package. This reproduces on the existing multi-module test `packageoption` with two small additions. `Examples/test-suite/packageoption_c.i` is `%module(package="PackageC")` and already `%import`s `packageoption_a.i`, which is `%module(package="CommonPackage")` and defines class `A` (plus `Base` from `packageoption.h`). Enable the feature in both and add one function that takes the imported type:

```diff
  // packageoption_a.i, after the %module line
+ %feature("python:annotations", "typing");

  // packageoption_c.i, after the %module line
+ %feature("python:annotations", "typing");
  // ... and inside the %inline block:
+ void use_imported(A *a, Base *b) {}
```

The generated `packageoption_c.py` imports the module correctly but annotates with the bare, unbound module name:

```python
import CommonPackage.packageoption_a                                    # binds 'CommonPackage', not 'packageoption_a'
def use_imported(a: "typing.Optional[packageoption_a.A]", ...) -> ...:   # 'packageoption_a' is undefined -> broken reference
```

python.cxx already has the correct, `package`- and `-relativeimport`-aware helper (`import_name_string()`), used by `classDeclaration()`. Reading the cached `python:proxy` attribute does **not** work (it is populated lazily during traversal — verified to regress `TemplateShort` to `SWIGTYPE_p_Space__TemplateT_short_t` when the function precedes the `%template`). Calling the helper directly does work:

```c
String *symname = Getattr(n, "sym:name");
if (!symname)
  return NULL;
Node *mod = Getattr(n, "module");
if (!mod)
  return Copy(symname);
Node *options = Getattr(mod, "options");
String *pkg = options ? Getattr(options, "package") : 0;
return import_name_string(package, mainmodule, pkg, Getattr(mod, "name"), symname);
```

With the fix the annotation becomes `typing.Optional[CommonPackage.packageoption_a.A]` — correct. This modified `packageoption` case would also make a good regression test: as it stands the imported type is only ever used as a base class, never as a function parameter, so there is no annotations coverage for cross-module `%import` at all (see Test coverage below).

---

## Smaller issues

- **`$*pytypename` on a non-pointer silently leaks the literal.** Verified: `%typemap(pytyping) Bar "B:$*pytypename"` emits `def t3(b: "B:$*pytypename")`. Java has the same silent skip, so it's consistent — but a `Swig_error` here would be cheap and the variable is now documented, so users will hit it.
- **HTML in `Doc/Manual/Python.html` is invalid.** A `<dl>` inside a `<p>` implicitly closes the paragraph. Running `tidy` (as `Doc/Manual/Makefile`'s `check` target does) with the manual's real doctype gives two new warnings: `inserting implicit <p>` / `trimming empty <p>`. Also, `<dl>`/`<dt>`/`<dd>` appears nowhere in Python.html; Java.html §27.8.7 and the rest of the manual use `<p><b><tt>$var</tt></b><br>…</p>`. Matching that both fixes tidy and reads consistently. The doc also doesn't mention that `$pytypename` yields `int` for an enum.
- **Consider a proper "Python special variables" section.** Java has `Java_special_variables`, C# has its own; Python has none. `$pytypename` is the first entry that warrants one, and it would be linkable from Typemaps.html.
- **`if _swig_python_version_info >= (3, 5) and typing.TYPE_CHECKING:`** — the version half is dead weight now that Python 2 and <3.5 are gone (f505f2566, 79f7a2b7c), and a compound condition is less reliably special-cased by checkers than the bare `if typing.TYPE_CHECKING:` every checker recognises. This was not verified with mypy/pyright (not installed here), so treat it as a simplification worth making rather than a confirmed break.
- **Two leaks in `substituteTypenameSpecialVariable` / `substitutePytypingVars`** — `SwigType_manglestr()` returns a new String that `NewStringf` doesn't consume, and `SwigType_typedef_resolve_all()` returns a new object that `Copy()` discards. Both are copied verbatim from java.cxx:3446/3551, so they're pre-existing project-wide — but given 42149f507 was specifically a `lookupPytyping` ownership fix, it's worth not propagating them into new code.
- **`lookupPytyping` can return NULL** (there's a `WARN_PYTHON_TYPEMAP_PYTYPING_UNDEF` warning for exactly that), and `substitutePytypingVars(pt, tm)` is called unconditionally. It survives in practice, but an explicit `if (tm && pt)` is clearer than relying on `Strstr(NULL, …)`.
- **Cheap early-out.** `substitutePytypingVars` runs `typedef_resolve_all` + `strip_qualifiers` + 3 `Strstr` for every parameter of every function, even though `$` appears in almost none. `if (!Strchr(tm, '$')) return false;` first.
- **`Printf(f_shadow_py, "    class %s(object):\n", …)`** — three `Printf` calls where one `Printv` would do, matching the style of `emitTypeWrapperClasses` right above it.
- **The generated docstring is thinner than the PR description advertises.** The PR description shows the placeholder class with a two-line docstring: a first line naming the opaque type, then a second line "This type is only used for type annotations and does not exist at runtime." But `emitTypeWrapperClass()` emits only the first line — `"""Opaque 'void *'."""` — dropping the explanatory sentence. Either update the description to match the code, or (better) make the code emit that second sentence too: it usefully tells a reader the class is annotation-only and absent at runtime.

## CHANGES.current

Format is right (`[Python] #3390`, 12-space body). Fixes needed: the date `2026-06-08` needs bumping to the merge date; the entry doesn't mention that `enum` parameters change from `typing.Any` to `int`, that `SWIGTYPE_*` placeholder classes now appear in the generated `.py`, or the `get_type_hints()` consequence — all user-visible.

## Test coverage

The added cases are good and well chosen — `MyEnum "bool"` proves a user typemap beats `enum SWIGTYPE`; `%rename(MyNamespaced2)` proves `sym:name` is honoured; member/member-function pointers, `void *`, `short&` vs `const short&`, and the "not present at runtime" `hasattr` loop are all worth having. `swig_assert` is already imported at the top of the runme, so that's fine.

Gaps:

- **`$*pytypename` is documented but never tested**, including the silent no-pop case.
- **`$&pytypename` and bare `$pytypename` in a user typemap are only tested via an enum** — no test on an actual class type, so the class path through the special variables is only covered indirectly by the defaults.
- **No cross-module test.** Master now has `python_crossmod`; the `%import` + `%module(package=…)` path is exactly where bug 4 lives.
- **No test for variables or constants** of class type, even though `lookupPytyping` now substitutes there too.
- **No test for the `%ignore`d / forward-declared-only class fallback.**
- Removing the duplicate pre-instantiation `MakeShort` check is fine, but note it did assert annotations are stable across instantiation.
- `-builtin` / `-fastproxy` remain uncovered because the runme skips them (annotations aren't emitted there at all — confirmed).

---

## Adjacent pre-existing bug found while reviewing

`Doc/Manual/Python.html` documents `%feature("python:annotations", "pytyping")` throughout the PEP 484 section, but `getTypeAnnotationMode()` only accepts `"typing"`. Verified: with `"pytyping"` the generated code has **no annotations at all**, silently. Anyone following the manual gets nothing. Not this PR's doing, but it's editing that exact section — worth a one-line fix here or a separate commit.

---

## What's left after this lands

1. **`.pyi` stub generation** (issue #735's real endgame). It removes the `TYPE_CHECKING` hack, fixes `get_type_hints()`, and is the only way to get annotations under `-builtin` and `-fastproxy`, which are structurally unreachable today.
2. **Finish the library `pytyping` sweep** — the seven files listed above. This is the prerequisite that makes the new `SWIGTYPE` default safe rather than risky.
3. **`@typing.overload` for overloaded functions.** Today `def global_overloaded(*args)` carries no parameter annotations at all — the single largest remaining hole in the feature.
4. **Union annotations for containers**, using the `out=` attribute so `in` accepts a sequence/dict while `out` stays the proxy type.
5. **Drop the `_swig_python_version_info >= (3, 5)` guards** and import `typing` unconditionally now that Python 2 and <3.5 are gone.
6. **Fix the `"pytyping"` vs `"typing"` documentation bug.**
7. **Global variables via `cvar`** are still unannotated (only members and `%constant` are covered).
8. **Director methods** — the Python-side overrides are unannotated.
9. **Add a Python special variables section** to the manual and cross-link it from Typemaps.html.

---

**Verdict:** worth landing, but not in this shape. Items 1 and 4 are mechanical. Item 2 is the real gate — either add the missing library `pytyping` typemaps in this PR (or a precursor) before flipping the `SWIGTYPE` default, or split the PR so `$pytypename` plus explicit opt-in typemaps land first and the default change follows once the library is complete. Item 3 needs a decision on whether runtime introspection is a supported use of these annotations.



---

Probably best to discuss and agree the high level comments. If you don't have an AI agent, I expect I can get mine to fix everything quickly... let me know once we've agree the main points.

Nerixyz:
> 1. The PR is out of date; `pytyping.swg` is largely superseded

I think the AI got confused by the conflicts in the file. I pushed a merge commit now.

> 2. The `SWIGTYPE` default change makes annotations confidently wrong, not merely uninformative

Yea that's a valid issue, which is mainly due to missing `pytyping` typemaps for the mentioned types. I see two options:

1. Annotate everything in this PR. I believe that's what the AI wants to do - and probably something the AI is good at.
2. Restore the previous `%typemap(pytyping) SWIGTYPE "typing.Any"` and only add the `$pytyping` maps in the test for now. Then, in a followup, add the typemaps for the remaining types.

> 3. `typing.get_type_hints()` now raises `NameError`

You changed this in [`da00715` (this PR)](https://github.com/swig/swig/pull/3390/commits/da00715eb74f22c0580a9ff930186305e9311ebd). I'm not sure what the best solution here would be. Because the stub-classes are not used by SWIG, so ideally, they shouldn't be available at runtime. However, to allow type checking at runtime with `typing.get_type_hints`, we need them to be available. The AI mentions third party libraries consuming the extensions, so I'm leaning towards always adding the classes (i.e. removing the `if TYPE_CHECKING`).

> 4. Cross-module class names are wrong with `%module(package=...)`

Yes, this has to be fixed.

---

The "Small issues" are valid, and should be fixed. One particular stands out:

> * **`if _swig_python_version_info >= (3, 5) and typing.TYPE_CHECKING:`** — the version half is dead weight now that Python 2 and <3.5 are gone ([f505f25](https://github.com/swig/swig/commit/f505f2566fb89f633b8251f1644fc5375be404bd), [79f7a2b](https://github.com/swig/swig/commit/79f7a2b7cb7ddc256eba09c8770133148025171d)

This also means we can unconditionally `import typing`. This would be great for some LSPs that can't see through `_swig_python_version_info`.

---

> If you don't have an AI agent, I expect I can get mine to fix everything quickly... let me know once we've agree the main points.

I don't have an AI agent, but I do agree on the main points. So if you can, feel free to fix these issues.

wsfulton:
I've been thinking about point 3. The `if typing.TYPE_CHECKING` and quoted forward reference annotations is the standard convention and `get_type_hints()` not resolving the opaque `SWIGTYPE_*` names is expected. The annotations are aimed at static checkers/IDEs, which see the guarded classes fine. 

I also checked https://docs.python.org/3/library/typing.html#constant and the `annotationlib.get_annotations()` route suggested in these docs (requires Python 3.14 or later): it's the standard non-crashing reader and does avoid the NameError. Worth noting it returns the annotations as strings rather than resolved objects here, because we quote the whole annotation, but that's fine for the static-checking use case. So in conclusion, let's leave the type wrapper classes behind the `TYPE_CHECKING` if statements.

Maybe a one-line doc note that the annotations are static-only (won't survive get_type_hints()/pydantic/typeguard) - but that's optional, not a blocker.

Regarding point 2., yes, agreed. I'm basically waiting for this PEP484 work to reach a conclusion before cutting the 4.5.0 release. With the functionality in place to use the `$pytyping` special variables it'll be handy for those to add manually and test using the release, but not a risky change as it'll be turned off by default. In the 4.6 development phase, hopefully get this turned on.

I'll try an agentic session soon to address the remaining issues.

wsfulton:
Landed in 722ccac7e095db2d1566c5eee1ed41a601197df1...a0da96aff8dd118dec4a2b9cfaf101e7aa2acdf4

PR #3393, opened 2026-03-31
Task/std function lua

christophe-calmejane:
Adds lua typemaps for std::function.

Note: I have the same one for C# but needs some refactoring. I hope I can provide it "soon".

### DIFF
--- a/Doc/Manual/Contents.html
+++ b/Doc/Manual/Contents.html
@@ -500,6 +500,9 @@
 </ul>
 <li><a href="Library.html#Library_std_auto_ptr">auto_ptr smart pointer</a>
 <li><a href="Library.html#Library_std_function">std::function</a>
+<ul>
+<li><a href="Library.html#Library_std_function_callback">Passing target language functions to C++</a>
+</ul>
 </ul>
 <li><a href="Library.html#Library_nn16">Utility Libraries</a>
 <ul>
--- a/Doc/Manual/Library.html
+++ b/Doc/Manual/Library.html
@@ -2529,8 +2529,10 @@ Limitations and notes:
 through to the empty primary template and produces no useful interface.</li>
 <li>The wrapper does not let target-language code construct a
 <tt>std::function</tt> from a target-language callable. <tt>std::function</tt>
-moves a C++ callable <i>out</i> to the target language; to go the other way,
-see
+moves a C++ callable <i>out</i> to the target language; to go the other way
+(pass a target-language function to C++ as a <tt>std::function</tt>), see the
+<a href="#Library_std_function_callback">%std_function callback macro</a> below,
+or alternatively
 <a href="SWIG.html#SWIG_nn30">Pointers to functions and callbacks</a>
 or
 <a href="SWIGPlus.html#SWIGPlus_target_language_callbacks">Callbacks to the target language</a>.</li>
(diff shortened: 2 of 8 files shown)


erezgeva:
Hi,

Two questions

1. Why C++ 14?
std::function<> was added in C++ 11
https://en.cppreference.com/cpp/utility/functional/function
Lambda expressions (since C++11)
https://en.cppreference.com/cpp/language/lambda
1. Why is the test called `lua_std_function‎`? It looks perfectly appropriate to any other language. I think `cpp11_std_function` is more proper.
If we want to implement other languages, we can use the same test.
If Lua really need C++14, it may call the test in C++14, but after all `std::function<>` was added in C++11.

<br>
I think this looks like a great addition to SWIG.
Thanks for your contribution.

christophe-calmejane:
Hi, thanks for the questions:

1. Because it's much easier to work with lambdas and capture parameter in C++14 than C++11. I'm pretty sure we can switch back to C++11 compatible code, but at the expanse of readability and complexity. Not sure it's worth it.

2. Absolutely right. I probably had conflicts when working on lua and C# std::function implementations and had to rename this file. I'll fix it.

wsfulton:
There has been quite a bit of std::function work recently merged into master. See #3298.

Could you please update this pull request given the latest support for std::function. I'm trying to understand what it really offers given we have %callback, directors and the recently added std::function support. Once it's been adapted, could you add some user level documentation with clear enhancements compared to what is already available. I'm afraid my Lua is not very good, so you should add in the C# equivalents as that might help with understand. Also any new concept you are adding here needs to be applicable across different languages - that's a primary focus for SWIG. I can't really see how this is applicable across different languages as there are so quite a lot of Lua specific code.

christophe-calmejane:
Oh I didn't see that new code for expanding parameters, that is great news. The lua implementation of std::function is mainly based on the C#/Python code we developed almost 2y ago and couldn't offer as PR until the std::optional stuff was merged (for various unrelated reasons).
I hope I'll have the time to update it, if not too much, as I won't have as much spare time I had the past 2 months.

christophe-calmejane:
After studying the new `Lib/std/std_function.i`, I've updated the code to integrate with it.
Here's what the rebased+updated code adds:

**Filling a gap**:
Master's `std_function.i` covers **C++ → target language** direction. My code adds **target language → C++** direction: passing a native Lua function directly where C++ expects a `std::function` parameter.

The new documentation states this explicitly as a limitation: _"The wrapper does not let target-language code construct a std::function from a target-language callable."_
My `%std_function` macro removes that limitation.

**Changes I made**:
- `Lib/lua/std_function.i` now includes `<std/std_function.i>` first, so `%template` instantiation works alongside my `%std_function` macro
- Test renamed to `cpp14_std_function_callback` (like I explained before, moving back to cpp11 will add a lot of boilerplate code, complications, less readability) and placed in `CPP14_TEST_CASES`
- Added user documentation (AI generated, seems good enough) in `Doc/Manual/Library.html` (section 12.5.8.1)

Happy to discuss further or make additional changes. If everything is fine, I'll have to update my C# `std::function` implementation to match this (not enough time as this moment).

wsfulton:
I can't review the html changes as there are tons of pointless whitespace changes. Please revert these changes and make sure the diffs contain just your text additions.

While I can't fully understand the changes here given the html doc diffs are incomprehensible, it seems like this is a Lua only change when really we need a language neutral feature, or a feature that can be expanded to all languages. Is a language-neutral feature something you've thought about? I presume you are aware of the cross-language directors feature; have you thought how this can be used instead as this is the primary mechanism for inter-language callbacks where a Lua method can be called from C++? The directors feature is a massive SWIG feature and adding something similar to it, replicating most of much of what it does is not desirable when it can be used as is.

The recent `std::function` changes in master enable target languages to call C++ callbacks (such as a lambda or a functor) from a target language. I'm confused as you say the recent changes  _covers C++ -> target language_ when it is the other way round. Sorry, for the fundamental questions, but I need the basics to be absolutely clear.

erezgeva:
> The recent `std::function` changes in master enable target languages to call C++ callbacks (such as a lambda or a functor) from a target language. I'm confused as you say the recent changes _covers C++ -> target language_ when it is the other way round. Sorry, for the fundamental questions, but I need the basics to be absolutely clear.

That is a fundamental question @christophe-calmejane should address.
If I may, from testing Lua director implementation, I found it is not a complete solution as other target languages.
`std::function` is in some way indeed a duplicate of the directors feature but seems more natural to use in Lua environment.
I do agree with you we must take it in a more broad thinking.
But perhaps mark it as an experimental feature, as we do with other features and languages we are not fully support or test.
Although the project started with Python, we do not have to start there and may add features to specific target languages on a need base.
I do find Lua a very useful language, so I personally think we should consider starting with Lua.

Last remark though. We do need a proper marking of the two separate features, so we can understand what feature we actually use. Sharing C++ std::function as a value to target language, or calling a callback in target language, i.e. calling a director method without a class.

christophe-calmejane:
**1. Regarding the white spaces:**
That's the problem with modern editors that always want to properly reformat everything :(
I've setup a clang-format long before you added one, for the source code, but it doesn't apply to html. The original file already has a lot of indentation issues, I guess that's what my editor tried to solve. I'll rewrite it and force unformatted saving and force-push.

Now regarding your questions/concerns, correct me if I'm wrong, but this is how I understand things:

**2. Direction clarification:**

I think the confusion is terminology (I'm often confused with in/out typemaps, language direction). Let me try to be precise:

- PR #3298 (`%template(Name) std::function<R(A...)>`): wraps a `std::function` that already exists on the C++ side so the target language can invoke it (call `operator()`). The target language **consumes** a C++ callable.
- This PR (`%std_function(Name, R, A...)`): allows passing a target-language function where C++ expects a `std::function` parameter. C++ **consumes** a target-language callable.

Master's documentation (from #3298) explicitly states this as a current limitation:
> "The wrapper does not let target-language code construct a std::function from a target-language callable."

My PR removes that limitation for lua.

**3. This IS directors, not a replacement (at least it's how I see it, maybe I'm wrong):**

The `%std_function` macro generates a director-enabled class internally:
```
%feature("director") Name;
struct Name { virtual Ret Invoke(Args...) = 0; };
```
The typemap(in) then creates a director instance, wraps it in a `shared_ptr` (lifetime management), and builds a `std::function` that calls `Invoke()`. It's what a user would do manually with directors, nothing new/different.

**4. Why it can't be fully generic (AFAIK):**

The `typemap(in)` must convert a language-specific value (a `lua function` on the stack, a C# delegate, ...) into a director instance. That conversion really is language-specific, I don't think there's a way around it, just like `std_vector.i` needs per-language typemaps.

The architecture is :
- Language-independent: director class generation, shared_ptr lifetime management, the `%std_function` macro structure
- Language-specific : only the `%typemap(in)` code in `Lib/<lang>/std_function.i`

I will also provide the C# implementation (when my old branch is rebased/cleaned). For Python/Java/etc, the pattern is identical but someone with those languages knowledge would need to write the per-language `typemap(in)`.

**5. Unit test:**

The added `cpp14_std_function_callback.i` file is completely language-neutral. Each language just adds its own `_runme` file, as usual.

---
@erezgeva

Thanks a lot for for testing/improving Lua directors.

You're right that this can be seen as two distinct features:
- sharing a C++ `std::function` as a callable value to the target language
- calling a target-language callback from C++ via `std::function`

My `%std_function` macro is essentially "directors without a class" as you described it. The user doesn't need to define a class or subclass anything, they just pass a plain function and the macro handles the director stuff. I did the same thing in my C# version.

Regarding starting with Lua (or, not starting with python): agreed. I think it's better to have *some* languages supported, than having to wait for all of them. Especially if this project doesn't have maintainers for all target-language.
The approach is identical for every language, only the typemap(in) body differs.

christophe-calmejane:
Github got some nasty issues today, looks like I'm not able to re-run the failed jobs. Are you?

Issue #3395, opened 2026-04-03
[Lua] eLua status


erezgeva:
Hi @wsfulton ,  @christophe-calmejane 

I understand `eLua` is in experimental state.
But beside flags in `Source/Modules/lua.cxx` there are no tests.
No reference in `configure.ac`, non in `Examples/Makefile.in` and no CI.

Should we drop `eLua` or perhaps add a `eLua` test?

erezgeva:
The project home page `http://www.eluaproject.net/` seems offline.
Looking `https://github.com/elua/elua/tags`
The last release v0.9 on  Apr 2013.
Looks like a dead project.

ojwb:
FWIW, I also raised this (though as a side-note) about 4 years ago: https://github.com/swig/swig/issues/2213#issuecomment-1046329069

> The project home page http://www.eluaproject.net/ seems offline.

It seems to have been up recently enough to still appear in search results, but the lack of updates in git seems a strong indication that development has ceased.  Also last mailing list post seems to be 2021 (though this could be an out of date archive): http://elua-development.15.s1.nabble.com/

The iTcl support is also not exercised by the testsuite (and so not by CI) and has had an open bug for more than 20 years: https://sourceforge.net/p/swig/bugs/537/ - last upstream activity seems to be 2022 for that: https://sourceforge.net/projects/incrtcl/  We seem to lack any useful SWIG docs for iTcl (at least elua is discussed in the manual).

erezgeva:
> so raised this (t

Unlike eLua, iTcl does seems active and part of Tcl.
https://core.tcl-lang.org/itcl/
And I agree, features need to be tested.

ojwb:
> Unlike eLua, iTcl does seems active and part of Tcl. https://core.tcl-lang.org/itcl/ And I agree, features need to be tested.

OK, so that SF repo is inactive because iTcl is now maintained as part of Tcl.

It still seems unlikely anyone is actually using SWIG's iTcl support though - it is essentially undocumented, and doesn't currently seem to handle overloaded methods (which is likely to be a problem for wrapping most C++ APIs).

erezgeva:
There is a new project called Pallene.
[What about Pallene?](https://www.lua.org/wshop22/Ierusalimschy.pdf) [slides](https://www.lua.org/wshop22/Ierusalimschy.pdf) [video](https://youtu.be/pGF2UFG7n6Y)
Roberto Ierusalimschy

With active development in
https://github.com/pallene-lang

[...]

Issue #3394, opened 2026-04-01
[Lua] New Lua directors breaks on Lua 5.1


erezgeva:
Hi @christophe-calmejane ,

The new Lua directors swig/swig@9d455889413feac5633ae3df0f575753c6dea008 uses `lua_setuservalue` which was added in Lua 5.2 https://www.lua.org/manual/5.2/manual.html but is missing from Lua 5.1.

According to `Doc/Manual/Lua.html#Lua_nn2`
"29.1 Preliminaries"
"The current SWIG implementation is designed to work with Lua 5.0.x, 5.1.x and 5.2.x. "

So we need to

1.  Do we want to support Lua 5.1? If so we need to do update `Lib/lua/director.swg` so we replace `lua_setuservalue` when using Lua 5.1.
1. If we decide to drop Lua 5.1, than we need to update `configure.ac` to skip Lua 5.1 in probing and update "29.1 Preliminaries" and "The wrappers produced by SWIG can be compiled and linked with Lua 5.1.x and later. The loading is extremely simple." in "29.2.3 Compiling a dynamic module". 
1. We should add tests in `linux.yml` for all Lua versions 5.1, 5.2, 5.3 and  5.4. Currently we test only 5.2 and 5.3.

Erez

christophe-calmejane:
Hi,
I will need to look at my dev notes on directors, but I remember doing things to handle lua < 5.2 (initially code was failing the CI because of that fonction that only exists in 5.2). I'll take a look "shortly".

christophe-calmejane:
Ok you are right and I partially remember what I did.

I started the `lua directors` feature a long time ago, and just recently dug it up (and refreshed a bit). Originally I used `debug.setuservalue` because the CI actually used a lua 5.1 compiler (I think it was ubuntu 20.04).

When I refreshed it, I created some helper methods in `luadirector.swg` to "hide" using the debug library from the user, and thought it would be even better to use `lua_setuservalue` directly as it was now passing the CI. What I didn't notice it that the `lua` job (not the `lua 5.3` one) was no longer running on v5.1 but on 5.2, probably because of the increase in ubuntu runner version (meaning the CI is currently not testing old lua versions).

lua 5.1 is quite old and obsolete, so I guess it's up to @wsfulton to decide if we really want to keep compatibility at the expense of more complicated handling.

Note: I'm actually using `lua_setuservalue` directly now in my `std::function` PR (also passing CI).

erezgeva:
[...]

PR #3396, opened 2026-04-03
Lua: drop 5.0/5.1 support, remove -squash-bases and -no-old-metatable-bindings

erezgeva:
Follows #3394 <del> and #3395 </del>

- Remove `-no-old-metatable-bindings` as it is deprecated since 2014
- Remove the `-squash-bases` option as it is broken


erezgeva:
Move `Remove eLua` to last.
So we can leave `eLua` if Raman will add tests.

wsfulton:
can you make this pull request just a drop Lua5.1 for now and drop eLua in a separate pull request (if necessary given comments in #3395).

Please summarise the changes in this pull request and motivation for them. There are lot of test case changes which need an explanation.

erezgeva:
> can you make this pull request just a drop Lua5.1 for now and drop eLua in a separate pull request (if necessary given comments in #3395).

Easy, i reordered and left it last. 
I Just removed it.
As the Ubuntu servers are down, we will have many errors on the linux CI.
The Lua tests should pass.

> 
> Please summarise the changes in this pull request and motivation for them. There are lot of test case changes which need an explanation.

In terms of patchs:

1. Patch 1 remove all code related to Lua 5.1 or older
2. Patch 2 remove the `-squash-bases` flag
3. patch 3 remove the `-no-old-metatable-bindings` flag

All three have document updates, `Doc/Manual/Lua.html`.

As for "remove all code related to Lua 5.1 or older contains" which is the larger, the changes are in the commit message:


- `configure.ac`
  - Use minimum version 5.2 for Lua
  - Add lua 5.5, it was released in dec 2025
  - Syntax improvements
- GHA CI
  - Remove 'default' Lua version, it is confusing.
    Just select the version explicitly.
  - Add Lua 5.4 test, we need to test all supported versions 5.2, 5.3 and 5.4
    Lua 5.5 is not available on Ubuntu yet.
- Lua documentation `Doc/Manual/Lua.html`
  - Update version to 5.2
  - Remove mention to older SWIG version,
    whom wishes to go to the very old history can look on `CHANGES`.
  - Add references on missing operators in Lua.
- `_ENV` exist from Lua 5.2.
- Remove old compatibility code from `Lib/lua/luarun.swg`.
- Remove `LUA_VERSION_NUM` from Lua C wrapping code.
  It was used for old Lua 5.0 and Lua 5.1.
- Remove `_VERSION` from Lua tests.
  Note: _VERSION remains a keyword and developers may use it in interface files!
- Remove `Example/lua/import.lua` as it is not used.
- `Examples/test-suite/lua/import.lua`
  - Remove old Lua import.
  - Add catch "undefined" global variables enable function.
  - Add `move_to_globs` and `import_to_globs` functions.
    To add module functions, variables and classes to global scope.

<br>

Lua 5.5 is already available on Debian testing, it will probably be available on next Ubuntu LTS.

erezgeva:
Rebase after master.

Clean new test `Examples/test-suite/lua/cpp11_std_function_runme.lua`

The `import("module")` function is removed, we call `require("module")` directly.

Most cases of `if <condition> then error("<message>")` can be expressed with `assert(not <condition>, "<message>")`

wsfulton:
I've restarted the failing linux tests as the Ubuntu ppas seem to have come back to life. And will do the same for other pull requests.

wsfulton (commented):
Thanks for the cleanup! I've pushed a small follow-up commit on this branch (9fe463e84) covering the obvious cosmetic fixes (typo, eLua grammar, comment collapse). The points below need your input / further work before merge.

### Title and scope

- The title says *"Drop Lua5.1 and eLua"*, but the diff actually:
  - Drops **Lua 5.0 *and* 5.1** (`configure.ac` enforces a 5.2 minimum, all `LUA_VERSION_NUM < 502` shims are gone, examples no longer use `lua_open` / `loadlib`-based loading).
  - Does **not** drop eLua: `-elua`, `-eluac`, `-elua-emulate`, and the entire `SWIG_LUA_TARGET == SWIG_LUA_FLAVOR_ELUA[C]` runtime path stay in place. The doc just reaffirms eLua is "experimental" (which addresses issue #3395 by *deciding to keep* eLua, rather than removing it).
- Suggested rename: *"Lua: drop 5.0/5.1 support, address eLua status (#3395), remove -squash-bases and -no-old-metatable-bindings"*. Worth splitting the eLua decision out of the title if it's purely a clarification.

### `CHANGES.current` is missing entries

Three user-visible behaviour changes need recording. Suggested entries (please check the exact CLI error text users will see if they pass the removed flags, and quote it verbatim per project convention):

- `[Lua]` Minimum supported Lua version is now 5.2. SWIG no longer compiles wrappers for Lua 5.0 or 5.1 (this includes LuaJIT consumers still using the 5.1 ABI).
- `[Lua]` Removed `-squash-bases` command-line option (and the `SWIG_LUA_SQUASH_BASES` runtime path). The implementation was broken; classes now always use the standard base-lookup path.
- `[Lua]` Removed deprecated `-no-old-metatable-bindings` command-line option. Old-style metatable bindings are always generated, matching the long-standing default.

### CI matrix coverage

- `configure.ac` now probes for `lua5.5`, but the GitHub Actions matrix exercises only 5.2 (×3), 5.3 (×1), and 5.4 (×1). 5.5 isn't tested at all and 5.3 / 5.4 each get only a single configuration. Worth at least one 5.5 entry, and probably another 5.3/5.4 entry, given that this PR is the one declaring "5.2 and above" the official supported range.

### Source / behaviour question

- `Source/Modules/lua.cxx` keeps `static int old_compatible_names = 1;` as a file-static with no command-line setter (the only consumer is `make_v2_compatible = getCurrentClass() && old_compatible_names`). With `old_metatable_bindings` gone there's now no way for it to ever be 0. Was leaving it as a hook for future use intentional, or can it be removed too?

### Test-suite `import.lua` rewrite

- The three new helpers (`catch_undef_globs`, `move_to_globs`, `require_to_globs`) are deliberately *global* so that runmes can call them after `require("import")`. Worth a one-line comment at the top of `Examples/test-suite/lua/import.lua` documenting that — otherwise the lack of `local` looks accidental.
- Several runmes dropped the `require("import")` line entirely (e.g. `char_strings_runme.lua`, `chartest_runme.lua`, all the `cpp11_shared_ptr_*_runme.lua`, and most `director_*_runme.lua`). That's consistent — they don't use the helpers — but it means those tests no longer enable `catch_undef_globs()` either. Was the undefined-global protection in the *original* versions of those runmes there on purpose, or just boilerplate? If on purpose, please re-add `require("import"); catch_undef_globs()` to keep the coverage; if boilerplate, fine as-is.
- `Examples/test-suite/lua/cpp17_string_view_runme.lua` calls `require("cpp17_string_view"); move_to_globs(cpp17_string_view)` rather than the `require_to_globs("cpp17_string_view")` pattern used by the rest of the diff. Same effect, but the inconsistency is minor noise — suggest harmonising.

### Process nit

- The PR mixes a meaningful refactor with a lot of cosmetic edits (trailing whitespace stripping in `Examples/lua/embed*`, semicolon removal in many `_runme.lua` files, `if not (X) then error(Y) end` → `assert(X, Y)` rewrites). All defensible individually, but they bulk the diff up to 161 files and obscure the actual behavioural change. Easier on reviewers if cosmetic-only chunks land in a separate prep commit/PR next time.

---

Once the title is updated, `CHANGES.current` has entries, and you've decided on the `old_compatible_names` and `require("import")`-loss questions, I'm happy to re-review.

erezgeva:
> Thanks for the cleanup! I've pushed a small follow-up commit on this branch ([9fe463e](https://github.com/swig/swig/commit/9fe463e845e4b56fdc21717d528e8f1536ad1b8d)) covering the obvious cosmetic fixes (typo, eLua grammar, comment collapse). The points below need your input / further work before merge.
> 
> ### Title and scope
> * The title says _"Drop Lua5.1 and eLua"_, but the diff actually:
>   
>   * Drops **Lua 5.0 _and_ 5.1** (`configure.ac` enforces a 5.2 minimum, all `LUA_VERSION_NUM < 502` shims are gone, examples no longer use `lua_open` / `loadlib`-based loading).

The commit has the full description of both 5.0 and 5.1
The title was a short cut.

>   * Does **not** drop eLua: `-elua`, `-eluac`, `-elua-emulate`, and the entire `SWIG_LUA_TARGET == SWIG_LUA_FLAVOR_ELUA[C]` runtime path stay in place. The doc just reaffirms eLua is "experimental" (which addresses issue #3395 by _deciding to keep_ eLua, rather than removing it).

I removed the commit that remove eLua after you request.
I just did not change the PR title

> * Suggested rename: _"Lua: drop 5.0/5.1 support, address eLua status (#3395), remove -squash-bases and -no-old-metatable-bindings"_. Worth splitting the eLua decision out of the title if it's purely a clarification.

Rename title.
The PR do not address eLua after all.

> 
> ### `CHANGES.current` is missing entries
> Three user-visible behaviour changes need recording. Suggested entries (please check the exact CLI error text users will see if they pass the removed flags, and quote it verbatim per project convention):
> 
> * `[Lua]` Minimum supported Lua version is now 5.2. SWIG no longer compiles wrappers for Lua 5.0 or 5.1 (this includes LuaJIT consumers still using the 5.1 ABI).
> * `[Lua]` Removed `-squash-bases` command-line option (and the `SWIG_LUA_SQUASH_BASES` runtime path). The implementation was broken; classes now always use the standard base-lookup path.
> * `[Lua]` Removed deprecated `-no-old-metatable-bindings` command-line option. Old-style metatable bindings are always generated, matching the long-standing default.
> 

On it

> ### CI matrix coverage
> * `configure.ac` now probes for `lua5.5`, but the GitHub Actions matrix exercises only 5.2 (×3), 5.3 (×1), and 5.4 (×1). 5.5 isn't tested at all and 5.3 / 5.4 each get only a single configuration. Worth at least one 5.5 entry, and probably another 5.3/5.4 entry, given that this PR is the one declaring "5.2 and above" the official supported range.

Ubuntu 24.04 LTS do not have the Lua 5.5 package, as it was only publish in February 2016.
The package will be available in the next Ubuntu release.
As the new Lua 5.5 is backward compatible to Lua 5.4, it should pass.
Already did testing with Debian testing on my project that use Lua 5.5 with SWIG. Though It did not try to run the full SWIG test with Lua 5.5.

> 
> ### Source / behaviour question
> * `Source/Modules/lua.cxx` keeps `static int old_compatible_names = 1;` as a file-static with no command-line setter (the only consumer is `make_v2_compatible = getCurrentClass() && old_compatible_names`). With `old_metatable_bindings` gone there's now no way for it to ever be 0. Was leaving it as a hook for future use intentional, or can it be removed too?

Line 1229:
````
    const int old_compatible_names_saved = old_compatible_names;
    if (getNSpace() || (Getattr(n, "sym:nspace") != 0 && Len(Getattr(n, "sym:nspace")) > 0)) {
      old_compatible_names = 0;
    }
````

I did not check if the condition is actually true.
But on the surface it does not seems to be directly related to the removed `old_metatable_bindings` flag.

> 
> ### Test-suite `import.lua` rewrite
> * The three new helpers (`catch_undef_globs`, `move_to_globs`, `require_to_globs`) are deliberately _global_ so that runmes can call them after `require("import")`. Worth a one-line comment at the top of `Examples/test-suite/lua/import.lua` documenting that — otherwise the lack of `local` looks accidental.

Sure, I'll add a comment.

> * Several runmes dropped the `require("import")` line entirely (e.g. `char_strings_runme.lua`, `chartest_runme.lua`, all the `cpp11_shared_ptr_*_runme.lua`, and most `director_*_runme.lua`). That's consistent — they don't use the helpers — but it means those tests no longer enable `catch_undef_globs()` either. Was the undefined-global protection in the _original_ versions of those runmes there on purpose, or just boilerplate? If on purpose, please re-add `require("import"); catch_undef_globs()` to keep the coverage; if boilerplate, fine as-is.

Some tests were specifically using the `catch_undef_globs()`
And some did not.
I just replace the already existing code with the newly added functions at `import.lua`.
But I did not add or remove behaviour from any existing test.
We may add more `catch_undef_globs()` or the other functions from  `import.lua`  to any other Lua test in the future.
My purpose was to reduce repeated code, not to change the tests behaviour or cover!



> * `Examples/test-suite/lua/cpp17_string_view_runme.lua` calls `require("cpp17_string_view"); move_to_globs(cpp17_string_view)` rather than the `require_to_globs("cpp17_string_view")` pattern used by the rest of the diff. Same effect, but the inconsistency is minor noise — suggest harmonising.

As I mention above, I replaced exiting code.
The global `cpp17_string_view` is used in the test.
Using `require_to_globs()` will skip the global.
The test seems keen on this variation.
I do not consider that very important, but I did not want to change the test itself, just the start where the repeated code is.

But if you insist, I can change.


> 
> ### Process nit
> * The PR mixes a meaningful refactor with a lot of cosmetic edits (trailing whitespace stripping in `Examples/lua/embed*`, semicolon removal in many `_runme.lua` files, `if not (X) then error(Y) end` → `assert(X, Y)` rewrites). All defensible individually, but they bulk the diff up to 161 files and obscure the actual behavioural change. Easier on reviewers if cosmetic-only chunks land in a separate prep commit/PR next time.

I generally agree.
It seems as a good opportunity to do some cleaning of the tests.
Though all files were change on merit.
whitespace, semicolon comes after.
Using if-error seems copying from other languages.
Most tests already uses asserts.
Assert is a more proper syntax for tests.
And as Lua support it, and most test already use them, I think it is proper to update the few that did not.
The `error` function is yet used in other context which do not comply with assert.
I think it is better to use `error` if it is not under condition.
It is better over `assert(false, 'message')`.

> 
> Once the title is updated, `CHANGES.current` has entries, and you've decided on the `old_compatible_names` and `require("import")`-loss questions, I'm happy to re-review.

Sure.

**Comment by erezgeva** [comment]
I did a full `make check-lua-examples check-lua-test-suite`
Under Debian testing with
```
$ make check-lua-version
showing lua version
Lua 5.5.0  Copyright (C) 1994-2025 Lua.org, PUC-Rio
```

Test pass fully without any warnings

wsfulton:
Sorry, I forced pushed a couple of times. I'll wait for the CI to pass, then I have another mess up to fix in my commit message, then will merge to master.

erezgeva:
> Sorry, I forced pushed a couple of times. I'll wait for the CI to pass, then I have another mess up to fix in my commit message, then will merge to master.

You may cancel the older `erezgeva:lua5_1_drop` actions, which can save us 1 hour waiting for all actions to finish.

wsfulton:
Merged now, see a27c13e8b...c94c19d29

PR #3399, opened 2026-04-08
[GO] Migrated from GOPATH mode to module-aware mode

erezgeva:
After seen the failure
https://github.com/swig/swig/actions/runs/23852332690/job/69535727350
I think it is the time.

The module-aware mode is a new and more safer mode. The GO code is divided into models,
 each has its own `go.mod` file.
The go.mod gives a better control on where to fetch
 the imported modules.
The cost is creating multiple `go.mod` files,
 the reward is a more stable and consistent building.

See: https://go.dev/ref/mod


wsfulton:
This PR overlaps with a `make -j` race I hit in the Go test-suite, which I fixed minimally in a27c13e8b27c2c8a29876d7d8268aa4556a4ccc9 — full analysis is in that commit message. Sorry, I didn't realise this PR was here.

In brief: the plain `%.cpptest` / `%.ctest` rules never passed `GOMOD`, so it expanded empty and every ordinary test shared one `gopath/src/go.mod`, written non-atomically (`printf > "$@"`). Under `-j`, one test's `go_cpp` rewrites/clears that shared tree while another's `go build` reads a half-written `go.mod`, giving:

```
go: no module declaration in go.mod
```

The `multicpptest` path was immune because it already passes `GOMOD="$*"`. The asymmetry is latent since 8a1c09e28 (2019) and became `-j`-triggerable in #3360 (the shared non-atomic `gopath/src/go.mod`).

My fix just extends the existing `GOMOD="$*"` pattern to the plain rules (one file, ~14 lines) so each test gets its own `gopath/<test>/src/go.mod`.

Flagging the analysis and a low-risk stop-gap in case a quick fix is wanted first.  Is this module aware PR still needed or useful?

erezgeva:
> This PR overlaps with a `make -j` race I hit in the Go test-suite, which I fixed minimally in [a27c13e](https://github.com/swig/swig/commit/a27c13e8b27c2c8a29876d7d8268aa4556a4ccc9) — full analysis is in that commit message. Sorry, I didn't realise this PR was here.
> 
> In brief: the plain `%.cpptest` / `%.ctest` rules never passed `GOMOD`, so it expanded empty and every ordinary test shared one `gopath/src/go.mod`, written non-atomically (`printf > "$@"`). Under `-j`, one test's `go_cpp` rewrites/clears that shared tree while another's `go build` reads a half-written `go.mod`, giving:
> 
> ```
> go: no module declaration in go.mod
> ```
> 
> The `multicpptest` path was immune because it already passes `GOMOD="$*"`. The asymmetry is latent since [8a1c09e](https://github.com/swig/swig/commit/8a1c09e280e323ba48407a9a9ae86f19d1c7abbe) (2019) and became `-j`-triggerable in #3360 (the shared non-atomic `gopath/src/go.mod`).
> 
> My fix just extends the existing `GOMOD="$*"` pattern to the plain rules (one file, ~14 lines) so each test gets its own `gopath/<test>/src/go.mod`.
> 
> Flagging the analysis and a low-risk stop-gap in case a quick fix is wanted first. Is this module aware PR still needed or useful?

This PR include the idea that each test generate with a separate folder.
This might be sufficient.
But I got this idea after I already tried to enhanced the GO build and end up with this issue.
And as I use the method in my project, I thought it is time for a full migration, in addition to using folder per test, from GOPATH to module aware mode.
And follow Go recommendation as GOPATH was left as backward compatible.

The new mode is more stable.
It uses multiple go.mod files that point to the exact location where the code is.
In addition it leaves the decision on build to the go builder. No need to build libraries static or dynamic, just place the go.mod files and tell go to build all in one command.

PR #3404, opened 2026-04-17
Fix -doxygen: @file block bleeds into first class docstring

### LINKED ISSUE #3403: -doxygen: @file block bleeds into first class docstring
### Description

When a header file contains a `@file` Doxygen block followed by a class
with its own `//!` doc comment, SWIG's `-doxygen` parser concatenates the
file-level content into the class docstring instead of discarding it. The
`@file` command should mark the block as file-scope documentation, not as
a pending comment to attach to the next declaration.

Same for @brief, @authors, and other file header blocks.

### Minimal reproducer

**test.h**
```cpp
//! @file      test.h
//! @brief     File-level brief description.
//!
//! @homepage  http://example.com
//! @authors   Someone

//! This is the real class description.
class Foo {
public:
    Foo() {}
};
```

**test.i**
```swig
%module test
%feature("autodoc");
%{ #include "test.h" %}
%include "test.h"
```

```
swig -c++ -python -doxygen test.i
```

### Actual output (`test.py`)

```python
class Foo(object):
    r""" File-level brief description. http://example.com Authors: Someone This is the real class description."""
```

### Expected output

```python
class Foo(object):
    r"""This is the real class description."""
```

### Environment

- SWIG 4.2.1
- Python target (`-python`), likely affects other targets too

### Notes

The `@file` command is Doxygen's explicit marker for file-scope
documentation. Doxygen itself never attaches a `@file` block to subsequent
declarations. SWIG should either discard the accumulated comment when it
encounters `@file`, or scope it to the file entity and not carry it
forward to the next class or function.

As a workaround, wrapping the file header in `#ifndef SWIG … #endif`
suppresses the bleed, but this is ugly and should not be necessary.

jwuttke:
When a header uses consecutive \`//!\` (or \`///\`) single-line comments for a
file-level block starting with \`@file\`, SWIG's comment accumulation loop was
concatenating the file-header content into the following class or function
docstring.

Two bugs conspired:
- Only the \`@file\` line itself was recognised as structural and skipped;
  subsequent lines (\`@brief\`, \`@authors\`, ...) are not in \`structuralTags[]\`
  and so were accumulated into \`yylval.str\`.
- Blank lines between comment groups do not break the accumulation loop
  (all \`SWIG_TOKEN_ENDLINE\` tokens are consumed silently), so the next
  declaration's own doc comment was appended to the same string.

Fix: introduce an \`in_structural_block\` flag. When the first comment in a
group contains a structural command (\`@file\`, \`@page\`, ...), set the flag but
continue accumulating content normally. Count newlines in the inner do-while;
when \`in_structural_block\` is set and two or more consecutive newlines are seen
(a blank line), discard all accumulated content and break, so the following
declaration's doc comment is processed fresh.

Not breaking the loop on a natural exit (no blank line) is deliberate: it
correctly handles the \`@name\`/\`@{\` member-grouping pattern, where \`@{\`
immediately follows \`@name\` without a blank line and must still be attached to
the next member.

The block-comment style (\`/*! @file ... */\`) was already handled correctly
because the entire block is one scanner token and \`isStructuralDoxygen()\`
would see \`@file\` in it.

Fixes #3403

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,12 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.5.0 (in progress)
 ===========================
 
+2026-04-17: jwuttke
+            #3403 Fix -doxygen: @file block bleeding into the first class
+            docstring. Single-line //! (or ///) comments forming a file-level
+            header (starting with @file) are now correctly discarded instead of
+            being concatenated into the next declaration's docstring.
+
 2026-03-26: Nerixyz
             [Python] #3389 Add missing C annotations and PEP 484 annotations to constants.
 
--- a/Examples/test-suite/doxygen_misc_constructs.i
+++ b/Examples/test-suite/doxygen_misc_constructs.i
@@ -190,5 +190,13 @@
 
     #include "doxygen_misc_constructs.h"
 
+    // @file block (single-line //! style) must not bleed into the next class doc (GitHub issue #3403).
+    //! @file      doxygen_misc_constructs.i
+    //! @brief     File-level brief description.
+    //! @authors   Someone
+
+    //! This is the real class description.
+    struct FileHeaderTestClass {};
+
 %}
     %include "doxygen_misc_constructs.h"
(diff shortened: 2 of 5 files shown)


wsfulton:
Thanks, fixes are squashed merged into d50e2355510965ef550ca4f2dfbc2fa783be8934 and an additional test for your change added into 5fcb0ace4f8e52df8e6b726ab2a95d8a7dbe4500.

PR #3405, opened 2026-04-20
More Lua tests

erezgeva:
- Add Lua examples
- Add and fix unused Lua tests
- Add `lua_module_global` Complementary test of `lua_no_module_global`.
- Add Lua tests


Examples/test-suite/template_parameters_global_scope.i
```diff
@@ -111,7 +111,7 @@ Alloc::Rebind< TDGlobalTable >::Integer Table5() { return 5; }
 Alloc::Rebind< ::TDGlobalTable >::Integer Table6() { return 6; }
 %}
 
-#if 0
+#if defined(SWIGLUA)
```
wsfulton: As far as possible, testing should be consistent across different languages, is Lua special here?

wsfulton (commented):
That's a lot of tests, good stuff! 

Out of interest, are the test case runme files all or mostly conversions from other languages?

We've got too many examples to be honest, I'd like to trim them back. Actually, I'd like to rewrite them so that there are just a few helpful get going kind of classic use cases, but that's for another day. The test-suite covers all corner cases and some of the examples are like testcases. Anyway, for now, it's okay to copy example.i files exactly as they are from another language. Can you remove those that aren't copies. Thanks.

erezgeva:
> That's a lot of tests, good stuff!
> 
> Out of interest, are the test case runme files all or mostly conversions from other languages?
> 
> We've got too many examples to be honest, I'd like to trim them back. Actually, I'd like to rewrite them so that there are just a few helpful get going kind of classic use cases, but that's for another day. The test-suite covers all corner cases and some of the examples are like testcases. Anyway, for now, it's okay to copy example.i files exactly as they are from another language. Can you remove those that aren't copies. Thanks.

Most are conversions
There are few new tests

`lua_module_global`  is the negative of `lua_no_module_global`
The  `lua_no_module_global` verify the `--nomoduleglobal` flag works.
The `lua_module_global`  is prove that omitting  the flag is works different.

I tried 13 more tests,
 as I was not convinced on their added value, i stopped. 
You may look on them, if you find them irrelevant than we may drop them:
```
c_delete_function
c_delete
const_const_2
constructor_explicit
constructor_value
contract_c
cpp11_delegating_constructors
cpp11_director_enums
cpp11_directors
cpp11_explicit_conversion_operators
cpp11_shared_ptr_crtp_upcast
director_nspace_director_name_collision
exception_partial_info
```

Examples/test-suite/template_parameters_global_scope.i
erezgeva: Not sure.
Seems more like the first to use.
I guess some other languages fail.
The custom of having conditional languages in interfaces files is quite spread.
I do agree we need less, but I do not think it is in this PR scope.

wsfulton:
The list of tests do seem to be unique runme file tests. No  harm having them, I'll leave them in.

The two main questions are:

1. Out of interest, are the test case runme files all or mostly conversions from other languages?
2. Noting that the example.i examples are very different to testcase under Examples/test-suite, are there any examples which were not copies from another language? The example.i files should be identical to another language.

Examples/test-suite/template_parameters_global_scope.i
wsfulton: Remove this enabling just for Lua, there is no reason to make this change just for Lua. It is testing global scope prefixes in the core and this work is not complete and it should be consistent for all languages and if not, that needs an explanation.

erezgeva: I'll give it a try and find which languages breaks.

erezgeva:
> The list of tests do seem to be unique runme file tests. No harm having them, I'll leave them in.
> 
> The two main questions are:
> 
> 1. Out of interest, are the test case runme files all or mostly conversions from other languages?

Yes, apart from the list of 13 tests above.

> 2. Noting that the example.i examples are very different to testcase under Examples/test-suite, are there any examples which were not copies from another language? The example.i files should be identical to another language.

The examples are copying of other examples in python and ruby.
The `example.i` are just pass a cleaning.
You can compare and see
```
diff -u python/contract/example.i    lua/contract/example.i
diff -u python/enum/example.i        lua/enum/example.i
diff -u python/funcptr/example.i     lua/funcptr/example.i
diff -u python/operator/example.i    lua/operator/example.i
diff -u python/reference/example.i   lua/reference/example.i
diff -u python/smartptr/example.i    lua/smartptr/example.i
diff -u python/std_vector/example.i  lua/std_vector/example.i
diff -u ruby/template/example.i      lua/template/example.i
diff -u ruby/overloading/example.i   lua/overloading/example.i
diff -u ruby/value/example.i         lua/value/example.i
```
The Makefiles are also copied with proper lua script name.
Only the `runme.lua` are fully converted.

As the `example.i` are modify for each language, we can modify for lua as well.

Examples/test-suite/template_parameters_global_scope.i
erezgeva: All tests pass.
Removed the condition completly.

erezgeva:
PR was merged into master.

wsfulton:
Indeed, I split the commit into two: a686cb1993910a4e9503b0da23e6669afb578c8a  (non-Lua) changes and a1b3ca81ca4195a5d51548bf5355e13380187e3b (Lua only changes).

PR #3406, opened 2026-04-21
added pointer initializations in a format supported in C

akarantanapa:
These C# typemaps produced .c code that did not compile due to the now corrected pointer initializations using C++ syntax.

### DIFF
--- a/Lib/csharp/std_wstring.i
+++ b/Lib/csharp/std_wstring.i
@@ -23,8 +23,8 @@ static std::wstring Swig_csharp_UTF16ToWString(const unsigned short *str) {
   if (sizeof(wchar_t) == 2) {
     return std::wstring((wchar_t *)str);
   } else {
-    const unsigned short *pBegin(str);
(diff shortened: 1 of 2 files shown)


wsfulton:
The change in wchar.i can be used in C or C++ code. Just use C syntax which will also work in C++. In other words, remove the macro detecting c++.

While the change for std::wstring is not needed, you can make it the same as the change in wchar.i for consistency.

akarantanapa:
> The change in wchar.i can be used in C or C++ code. Just use C syntax which will also work in C++. In other words, remove the macro detecting c++.
> 
> While the change for std::wstring is not needed, you can make it the same as the change in wchar.i for consistency.

You're absolutely right, better use the C style and be rid of the #ifdefs 

And now that you say it out loud, ofc std_wstring.i is only used with C++. :D I just ran into this problem with wchar.i and then grepped the swig codebase to see if the same problem occurs elsewhere. Without thinking deeper that it might not be a problem in all contexts.

PR #3408, opened 2026-04-27
Python: Map more primitives for PEP 484 annotations

Nerixyz:
This is split from #3390 where I noticed that `T const &` wasn't mapped correctly for primitive types.

This PR maps all types that are included in https://github.com/swig/swig/blob/8b307eb27900a4375d4e7b6c5aaa852c62c54c5b/Lib/typemaps/primtypes.swg. Notably, this includes const-references.

### DIFF
--- a/Lib/python/pytyping.swg
+++ b/Lib/python/pytyping.swg
@@ -2,11 +2,47 @@
 
 %typemap(pytyping) bool "bool"
 
-%typemap(pytyping) signed char, unsigned char, short, unsigned short, int, unsigned int, long, unsigned long, long long, unsigned long long "int"
-%typemap(pytyping) size_t "int"
+/*
+    Map both 'T' and 'T const &' to 'PyType'.
+    Keep in sync with primtypes.swg.
+*/
+%define %_pytyping_primtype(Type, PyType)
+    %typemap(pytyping) Type, Type const & "PyType"
+%enddef
 
-%typemap(pytyping) float, double, long double "float"
+%_pytyping_primtype(bool, bool);
 
-%typemap(pytyping) char, wchar_t, char *, wchar_t * "str"
+%_pytyping_primtype(signed char       , int);
+%_pytyping_primtype(unsigned char     , int);
+%_pytyping_primtype(short             , int);
+%_pytyping_primtype(unsigned short    , int);
+%_pytyping_primtype(int               , int);
+%_pytyping_primtype(unsigned int      , int);
+%_pytyping_primtype(long              , int);
+%_pytyping_primtype(unsigned long     , int);
+%_pytyping_primtype(long long         , int);
+%_pytyping_primtype(unsigned long long, int);
+%_pytyping_primtype(size_t            , int);
+%_pytyping_primtype(std::size_t       , int);
+%_pytyping_primtype(ptrdiff_t         , int);
+%_pytyping_primtype(std::ptrdiff_t    , int);
+
+%_pytyping_primtype(float , float);
+%_pytyping_primtype(double, float);
+
+%_pytyping_primtype(std::complex< float > , complex);
+%_pytyping_primtype(std::complex< double >, complex);
+
+/* No 'T const &' maps for these */
+%typemap(pytyping) float _Complex, double _Complex, _Complex "complex"
+
+%_pytyping_primtype(char        , str);
+%_pytyping_primtype(wchar_t     , str);
+%_pytyping_primtype(std::string , str);
+%_pytyping_primtype(std::wstring, str);
+
+%typemap(pytyping) long double "float"
+
+%typemap(pytyping) char *, wchar_t * "str"
 
 %typemap(pytyping) void "None"
--- a/Lib/python/pytyping.swg
+++ b/Lib/python/pytyping.swg
@@ -1,12 +1,36 @@
 %typemap(pytyping) SWIGTYPE "typing.Any"
 
-%typemap(pytyping) bool "bool"
+%typemap(pytyping) bool, bool const & "bool"
 
-%typemap(pytyping) signed char, unsigned char, short, unsigned short, int, unsigned int, long, unsigned long, long long, unsigned long long "int"
-%typemap(pytyping) size_t "int"
+%typemap(pytyping)
+  signed char,        signed char const &,
+  unsigned char,      unsigned char const &,
+  short,              short const &,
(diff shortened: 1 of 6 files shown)


Nerixyz:
I added a second commit that turns the `str` annotation for `char*` into `typing.Optional[str]`, because `nullptr` is returned as `None` and passing `None` yields `nullptr`. There might be users where this isn't the case, but they can manually overwrite the annotation.

Can open a new PR for this if desired.

wsfulton:
I think these latest developments require better integration of the pytyping typemaps into SWIG's library files. The pytyping typemaps need to be provided in the appropriate library interface files. In particular in this commit, there should not be any pytyping typemaps for std:: string by default because by default there are no other specific typemaps for std::string. The specific pytyping typemaps should only be provided as part of the family of typemaps for a given type alongside the in, out, typecheck, etc typemaps in order to be correctly defined. So for std::string, that would be when the std_string.i file is included. The C primitive types which are available without including any C (nor SWIG) library files should of course be defined by default.

Nerixyz:
> The specific pytyping typemaps should only be provided as part of the family of typemaps for a given type alongside the in, out, typecheck, etc typemaps in order to be correctly defined.

I moved the typemaps for `std::*` types into their SWIG files. 

You mention the in and out typemaps. I didn't include pytyping maps for `Type *OUTPUT, Type *INPUT, Type *INOUT` yet, but if you'd like, I can add them in this PR.

wsfulton:
> You mention the in and out typemaps. I didn't include pytyping maps for `Type *OUTPUT, Type *INPUT, Type *INOUT` yet, but if you'd like, I can add them in this PR.

Add them to a new pull request please. Once you start looking, you'll find tons of library files that require the pytyping typemaps.

Could you drop the `_pytyping_primtype` macro? It's not really necessary as it isn't hard to get it right without the macro and there are always exceptions and it just easier to read (we have a nightmare number of macros already making the typemaps quite impenetrable for the average user when they are actually quite simple after preprocessing and grouping together). I would then just copy the style of the appropriate library file (as you've done for std::complex, std::string).

Lib/python/std_string.i
```diff
@@ -1 +1,5 @@
 %include <typemaps/std_string.swg>
+
+#ifdef SWIG_STD_STRING
+%typemap(pytyping) std::string, std::string const & "str"
+#endif
```
wsfulton, line 7: When SWIG_STD_STRING is not defined it falls back to SWIGTYPE default typing.Any. That's defensible (the proxy isn't a str), but the container std::string still accepts a Python str on input via the SWIG_AsCharPtrAndSize fallback, so typing.Any is looser than ideal. Probably it should be "str|typing.Any" ?

Lib/python/pytyping.swg
```diff
@@ -1,12 +1,36 @@
...
+    unsigned short,     unsigned short const &,     \
+    int,                int const &,                \
+    unsigned int,       unsigned int const &,       \
+    long,               long const &,               \
+    unsigned long,      unsigned long const &,      \
+    long long,          long long const &,          \
+    unsigned long long, unsigned long long const &, \
+    size_t,             size_t const &,             \
+    ptrdiff_t,          ptrdiff_t const &           \
+    "int"
```
wsfulton:
```suggestion
%typemap(pytyping)
  signed char,        signed char const &,
  unsigned char,      unsigned char const &,
  short,              short const &,
  unsigned short,     unsigned short const &,
  int,                int const &,
  unsigned int,       unsigned int const &,
  long,               long const &,
  unsigned long,      unsigned long const &,
  long long,          long long const &,
  unsigned long long, unsigned long long const &,
  size_t,             size_t const &,
  ptrdiff_t,          ptrdiff_t const &
"int"
```

Lib/python/pytyping.swg
```diff
@@ -1,12 +1,36 @@
...
 
-%typemap(pytyping) float, double, long double "float"
+%typemap(pytyping)                      \
+    float,  float const &,              \
+    double, double const &,             \
+    long double, long double const &   \
+    "float"
```
wsfulton:
```suggestion
%typemap(pytyping)
  float,       float const &
  double,      double const &
  long double, long double const &
"float"
```

Lib/python/pytyping.swg
```diff
@@ -1,12 +1,36 @@
...
 
-%typemap(pytyping) char, wchar_t, char *, wchar_t * "str"
+/* No 'T const &' maps for these */
+%typemap(pytyping) float _Complex, double _Complex, _Complex "complex"
+
+%typemap(pytyping)              \
+    char,    char const &,      \
+    wchar_t, wchar_t const &    \
+    "str"
```
wsfulton:
```suggestion
%typemap(pytyping)
    char,    char const &,
    wchar_t, wchar_t const &
    "str"
```

wsfulton (changes requested):
I've added some better formatting

Lib/python/std_string.i
Nerixyz: So in `std_string.i`, we should do:
```swig
#ifdef SWIG_STD_STRING
%typemap(pytyping) std::string, std::string const & "str"
#else
%typemap(pytyping) std::string, std::string const & "typing.Union[str, typing.Any]"
#endif
```
instead? I think that could work, yeah.

wsfulton: Yes indeed with your corrected syntax.

wsfulton: Please make this final change and test locally and push only once as CI resources are scarce.

Nerixyz: Tested this in CI on a separate branch in my fork.

Lib/python/std_wstring.i
```diff
@@ -1,3 +1,6 @@
 %include <pywstrings.swg>
 %include <typemaps/std_wstring.swg>
 
+#ifdef SWIG_STD_WSTRING
+%typemap(pytyping) std::wstring, std::wstring const & "str"
+#endif
```
wsfulton, line 8: Same comment here of course

wsfulton:
Thanks. Committed in 42942f4cae6b99f0e241163dd203f51e9da6309e and 3ce7a62a8b58047d41e97e388b9984a0fc5ba58f. 

I also made a follow up tidy up commit in 50df6ced1f22b24106ac49041b07dc443766a4f0.

PR #3409, opened 2026-04-27
Add protection to allocations in `argcargv` typemap

erezgeva:
Add protection against failing memory allocations in `argcargv` typemap

Lau:
- Fix error handling
- Shrink array once we get nil value

Python:
- Replace `%new_copy_array` with `%new_array` and explicit memory copy
  As `%new_copy_array` do not check for memory allocation failure.


wsfulton (commented):
Thanks @erezgeva — good hardening of the argcargv typemaps and the string-duplication paths.

I've pushed two follow-up commits on top of yours. Any comments prior to me merging to master?

**`argcargv / OOM-protection follow-ups`**

- **Lua** `Lib/lua/argcargv.i`: `$1 = i + 1` off-by-one corrected to `$1 = i` (the count only covers strings actually copied; the loop hasn't filled `$2[i]` yet). Also pop the nil from the Lua stack before breaking.
- **Python** `Lib/python/pystrings.swg`: the two `SWIG_MemoryError` early returns added by this PR skipped `SWIG_Py_XDECREF(bytes)` / `SWIG_Py_XDECREF(obj)`, leaking the temporary PyObject. Now decref'd before returning.
- **Octave / Python** `argcargv.i`: free already-allocated string entries (and the `**argv` array itself for Octave) before failing on per-element OOM. Python uses the existing `break` path so its `freearg` typemap performs the cleanup.
- **Ruby / Octave** `argcargv.i`: initialise `$1 = 0; $2 = NULL` up front so failure paths only need to clean up what they own. Also added an Octave runme case exercising the per-element non-string cleanup path.
- **Perl / PHP / Tcl** `argcargv.i`: dropped the unreachable `goto fail` after `SWIG_croak` / `SWIG_PHP_Error` / `SWIG_exception_fail` — those macros all already invoke `SWIG_fail`.
- Standardised the OOM error wording (`"Failed to allocate memory for 'int ARGC, char **ARGV' ..."`) and fixed a couple of indent inconsistencies (3- and 5-space blocks) in the new code.
- `Examples/perl5/xmlstring/xmlstring.i` updated to the new `%typemaps_string` signature (also picks up the missing `WarningLeakMsg` argument it never had) and a `SWIG_NewCopyXMLChArray` fragment added.
- `CHANGES.current` entry added documenting the API breakages introduced by this PR — `%new_copy_array` removal, `%typemaps_string_alloc` removal, and the extended `%typemaps_string` signature — including the literal SWIG / compiler error text users will hit (`Macro '%typemaps_string' expects 13 arguments`, `Unknown directive '%typemaps_string_alloc'`, `error: expected expression before '%' token`).

**`[Scilab] argcargv typemaps - free already-allocated strings on failure`**

Separate leak in the same area — not introduced by this PR. The per-string `malloc(aLen[i] + 1)` and third `getMatrixOfString` failure paths in `Lib/scilab/argcargv.i` were leaking `$2[0..i-1]` and `$2` itself before `SWIG_fail`. `freearg` doesn't run for a failed `in` typemap, so the memory was lost. Now freed up front before failing, and `$2` nulled so any future `freearg` is a no-op via its existing `$2 != NULL` guard.

`argcargvtest` passes for python, lua, perl5, ruby, php, tcl, octave, scilab; `li_std_string` / `li_std_wstring` runmes exercising the modified string fragments also pass. The OOM paths themselves aren't runme-testable (need an allocator hook), but Octave's new mixed-type case at least exercises the per-element cleanup block.

erezgeva:
> Thanks @erezgeva — good hardening of the argcargv typemaps and the string-duplication paths.
> 
> I've pushed two follow-up commits on top of yours. Any comments prior to me merging to master?
> 
> **`argcargv / OOM-protection follow-ups`**
> 
> * **Lua** `Lib/lua/argcargv.i`: `$1 = i + 1` off-by-one corrected to `$1 = i` (the count only covers strings actually copied; the loop hasn't filled `$2[i]` yet). Also pop the nil from the Lua stack before breaking.
> * **Python** `Lib/python/pystrings.swg`: the two `SWIG_MemoryError` early returns added by this PR skipped `SWIG_Py_XDECREF(bytes)` / `SWIG_Py_XDECREF(obj)`, leaking the temporary PyObject. Now decref'd before returning.
> * **Octave / Python** `argcargv.i`: free already-allocated string entries (and the `**argv` array itself for Octave) before failing on per-element OOM. Python uses the existing `break` path so its `freearg` typemap performs the cleanup.
> * **Ruby / Octave** `argcargv.i`: initialise `$1 = 0; $2 = NULL` up front so failure paths only need to clean up what they own. Also added an Octave runme case exercising the per-element non-string cleanup path.
> * **Perl / PHP / Tcl** `argcargv.i`: dropped the unreachable `goto fail` after `SWIG_croak` / `SWIG_PHP_Error` / `SWIG_exception_fail` — those macros all already invoke `SWIG_fail`.
> * Standardised the OOM error wording (`"Failed to allocate memory for 'int ARGC, char **ARGV' ..."`) and fixed a couple of indent inconsistencies (3- and 5-space blocks) in the new code.
> * `Examples/perl5/xmlstring/xmlstring.i` updated to the new `%typemaps_string` signature (also picks up the missing `WarningLeakMsg` argument it never had) and a `SWIG_NewCopyXMLChArray` fragment added.
> * `CHANGES.current` entry added documenting the API breakages introduced by this PR — `%new_copy_array` removal, `%typemaps_string_alloc` removal, and the extended `%typemaps_string` signature — including the literal SWIG / compiler error text users will hit (`Macro '%typemaps_string' expects 13 arguments`, `Unknown directive '%typemaps_string_alloc'`, `error: expected expression before '%' token`).
> 
> **`[Scilab] argcargv typemaps - free already-allocated strings on failure`**
> 
> Separate leak in the same area — not introduced by this PR. The per-string `malloc(aLen[i] + 1)` and third `getMatrixOfString` failure paths in `Lib/scilab/argcargv.i` were leaking `$2[0..i-1]` and `$2` itself before `SWIG_fail`. `freearg` doesn't run for a failed `in` typemap, so the memory was lost. Now freed up front before failing, and `$2` nulled so any future `freearg` is a no-op via its existing `$2 != NULL` guard.
> 
> `argcargvtest` passes for python, lua, perl5, ruby, php, tcl, octave, scilab; `li_std_string` / `li_std_wstring` runmes exercising the modified string fragments also pass. The OOM paths themselves aren't runme-testable (need an allocator hook), but Octave's new mixed-type case at least exercises the per-element cleanup block.

Thanks @wsfulton 
Review all changes and approve of them.
The `argvarg` is a bit challenging typemap.
I hope we nailed it this time.

Lib/lua/argcargv.i
```diff
@@ -34,13 +34,15 @@ SWIGINTERN int SWIG_argv_size(lua_State* L, int index) {
...
+      lua_pushstring(L,"Failed to allocate memory for 'int ARGC, char **ARGV' in '$symname'");
       SWIG_fail;
     }
     for (i = 0; i < size; i++) {
       lua_rawgeti(L,$input,i+1);
       if (lua_isnil(L,-1)) {
-        $1 = i + 1; /* shrink size to match actual size */
+        /* nil terminates the table early; shrink ARGC to the actual count */
+        $1 = i;
+        lua_pop(L,1);
```
erezgeva, line 45: <del> This  `lua_pop(L,1);` is done last, so we may drop it.
Though leaving it does not harm.</del>

<del>
@christophe-calmejane, what do you think?
</del>

<br> <br>


Looking again, we need the `pop` as the function is not done and we may have more arguments in the stack to process. after the typemap is dome.

erezgeva:
Hi @wsfulton,

Just remind that I review your changes and I approve them, including the Lua push on nill value.

You may proceed and merge.

Thanks for the review and the updates :+1:

wsfulton:
Thanks @erezgeva, great to tidy this up. Merged in c02d4fe134e072e4f0c28413f883a6f5c4e8d32f...a1e619b7cbd602500765212bf17c732e9c67abc1

PR #3411, opened 2026-04-30
[R] Fix compilation with R 4.6.0 which removed non-API macros

### LINKED ISSUE #3407: Substitute use of non-API R function SET_S4_OBJECT
R core has been cleaning the exported C API, and R v4.6.0 (released last Friday) removes a lot of non-API functions. Swig still uses `SET_S4_OBJECT`, which produces code that doesn't compile under this release. These calls should be replaced with `Rf_asS4`. See [S4 objects](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#S4-objects-1). E.g. in 

https://github.com/swig/swig/blob/8b307eb27900a4375d4e7b6c5aaa852c62c54c5b/Lib/r/rrun.swg#L270

(and two more calls), you probably want 

```C
r_obj = Rf_asS4(r_obj, TRUE, 0);
```

instead.

blowekamp:
R 4.6.0 (released April 2025) removed several non-API C macros from its
public headers, breaking compilation of SWIG-generated R wrapper code.
This PR replaces them with the proper stable R API equivalents which have
been available since R 2.x, so no version guards are needed.

Fixes #3407.

### Changes

**`Lib/r/rrun.swg`** — 3 sites where `SET_S4_OBJECT(x)` was used to mark
an R object as an S4 instance. Replaced with `x = Rf_asS4(x, TRUE, 0)`,
which is the documented API equivalent (available since R 2.4.0).

**`Lib/r/std_vector.i`** — 2 sites in `traits_from_ptr` specialisations
for `std::vector<std::string>` and `std::vector<std::vector<std::string>>`
that wrote into a character vector via `CHARACTER_POINTER(v)[i] = ...`.
`CHARACTER_POINTER` now returns a `const SEXP *` in R 4.6.0, making write
access illegal. Replaced with `SET_STRING_ELT(v, i, ...)`.

**`Lib/r/argcargv.i`** — The `(int ARGC, char **ARGV)` typemap used
`CHARACTER_POINTER` to obtain a pointer to the string array and then
indexed through it. Replaced the pointer variable with direct calls to
`STRING_ELT($input, i)`.

### Testing

The full R test suite was run both before and after the change using the
existing SWIG binary with the modified `Lib/` files
(`SWIG_LIB=... make -k check-r-test-suite`). Both runs processed 691 test
cases with identical results: the only failures were 6 boost-related tests
(`li_boost_shared_ptr*`, `director_shared_ptr`,
`multiple_inheritance_shared_ptr`) that fail due to missing boost headers
in the test environment and are unrelated to this change.

Key tests that exercise the affected code paths all passed:
- `argcargvtest` (with run test) — exercises `argcargv.i`
- `li_std_vector` (with run test) — exercises `std_vector.i` string path
- `li_std_vector_vector` (with run test) — exercises nested string vector path
- All `smart_pointer_*` and `pointer_*` tests — exercise `rrun.swg` S4 marking

<details>
<summary>AI assistance</summary>

- Tool: GitHub Copilot (Claude Sonnet 4.6)
- Role: root-cause analysis, identification of all affected code sites, and implementation of fixes
- Contribution: researched R 4.6.0 API removals, located all 6 affected sites across 3 files, identified backward-compatible replacement APIs, implemented changes, ran test suite to verify no regressions
- All code was reviewed and tested locally before committing

</details>

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -275,6 +275,8 @@ jobs:
         - SWIGLANG: python
           CPPSTD: c++11
         - SWIGLANG: r
+          VER: 4.0.0
+        - SWIGLANG: r
           CPPSTD: c++11
         - SWIGLANG: ruby
           CPPSTD: c++11
@@ -319,6 +321,7 @@ jobs:
           CPPSTD: c++14
         - SWIGLANG: r
           CPPSTD: c++14
+          VER: 4.5.0
         - SWIGLANG: ruby
           CPPSTD: c++14
         - SWIGLANG: scilab
@@ -386,6 +389,10 @@ jobs:
         - SWIGLANG: python
           CPPSTD: c++17
           GCC: 13
+          os: ubuntu-24.04
+        - SWIGLANG: r
+          CPPSTD: c++17
+          VER: 4.6.0
           os: ubuntu-24.04
         - SWIGLANG: r
           CPPSTD: c++17
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,12 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.5.0 (in progress)
 ===========================
 
+2026-04-30: blowekamp
+            [R] #3407 Fix compilation with R 4.6.0 which removed the non-API macro
+            SET_S4_OBJECT and made CHARACTER_POINTER return a const pointer.
+            Use Rf_asS4, SET_STRING_ELT and STRING_ELT instead. These replacement
+            APIs have been available since R 2.x so no version guards are required.
+
 2026-04-25: wsfulton
             Advanced %rename fixes for templates. It is now possible to use %rename
 	    on instantiated templates for various corner cases, see "Template renaming"
--- a/Lib/r/argcargv.i
+++ b/Lib/r/argcargv.i
@@ -14,7 +14,6 @@
 %}
 %typemap(in) (int ARGC, char **ARGV) {
   $1_ltype i;
-  SEXP *pstr;
   if ($input == R_NilValue) {
     /* Empty array */
     $1 = 0;
@@ -22,14 +21,13 @@
     SWIG_exception_fail(SWIG_RuntimeError, "Wrong array type.");
   } else {
     $1 = Rf_length($input);
-    pstr = CHARACTER_POINTER($input);
   }
   $2 = ($2_ltype) malloc(($1+1)*sizeof($*2_ltype));
   if ($2 == NULL) {
     SWIG_exception_fail(SWIG_MemoryError, "Memory allocation failed.");
   }
   for (i = 0; i < $1; i++) {
-    $2[i] = ($*2_ltype)STRING_VALUE(pstr[i]);
+    $2[i] = ($*2_ltype)STRING_VALUE(STRING_ELT($input, i));
   }
   $2[i] = NULL;
 }
--- a/Lib/r/rrun.swg
+++ b/Lib/r/rrun.swg
@@ -267,7 +267,7 @@ SWIG_MakePtr(void *ptr, const char *typeName, int flags)
     R_RegisterCFinalizer(external, R_SWIG_ReferenceFinalizer);
 
   r_obj = SET_SLOT(r_obj, Rf_mkString((char *) "ref"), external);
-  SET_S4_OBJECT(r_obj);
+  r_obj = Rf_asS4(r_obj, TRUE, 0);
   Rf_unprotect(2);
 
   return(r_obj);
@@ -285,7 +285,7 @@ R_SWIG_create_SWIG_R_Array(const char *typeName, SEXP ref, int len)
    Rf_protect(arr = R_do_slot_assign(arr, Rf_mkString("dims"), Rf_ScalarInteger(len)));
 
    Rf_unprotect(3); 			   
-   SET_S4_OBJECT(arr);	
+   arr = Rf_asS4(arr, TRUE, 0);
    return arr;
 }
 
@@ -308,7 +308,7 @@ SWIG_R_NewPointerObj(void *ptr, swig_type_info *type, int flags) {
   }
   rptr = R_MakeExternalPtr(ptr, 
   R_MakeExternalPtr(type, R_NilValue, R_NilValue), R_NilValue); 
-  SET_S4_OBJECT(rptr);
+  rptr = Rf_asS4(rptr, TRUE, 0);
   return rptr;
 }
 
--- a/Lib/r/std_vector.i
+++ b/Lib/r/std_vector.i
@@ -203,7 +203,7 @@
          PROTECT(result = Rf_allocVector(STRSXP, val->size()));
          for (unsigned pos = 0; pos < val->size(); pos++)
            {
-             CHARACTER_POINTER(result)[pos] = Rf_mkChar(((*val)[pos]).c_str());
+             SET_STRING_ELT(result, pos, Rf_mkChar(((*val)[pos]).c_str()));
            }
         UNPROTECT(1);
         return(result);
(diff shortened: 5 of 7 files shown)


blowekamp:
@richardbeare Are you able to review this PR?

erezgeva:
Finding R release dates and names is a bit challenging.
The project is quite old and do not maintain the information in order.

However found this site
https://cran.r-project.org/src/base/R-2/
Which seems holding the original versions.
`R-2.4.0.tar.gz` seems from   2006.
That is 20 years. old enough.
Found the original code here
https://svn.r-project.org/R/tags/R-2-4-0/
Still using subversion in 2026 ... 

<br>

My suggestion:
1. Looking on in `.github/workflows/linux.yml` test only 2 R versions each for each Ubuntu version.
   Is there a way to test other versions?
   Perhaps download original release and build it?
3. I think we need to add note in `Doc/Manual/R.html`, which versions are supported, a minimum and the latest version 4.6.0 with the release dates of these revisions.
4. I suggest adding a version verification in `configure.ac`.

blowekamp:
> My suggestion:
> 

> 3. I suggest adding a version verification in `configure.ac`.

I think these are all beyond the scope of fixing this compilation error due to non-API macros being used. But here are my thoughts:

> 1. Looking on in `.github/workflows/linux.yml` test only 2 R versions each for each Ubuntu version.
>    Is there a way to test other versions?
>    Perhaps download original release and build it?

In SimpleITK we use the GHA[ r-lib/actions/setup-r](https://github.com/SimpleITK/SimpleITK/blob/d8948b580fefe42a0c94f31e328277506d8cb123/.github/workflows/Build.yml) to get R installed. We had the version set to the 'latest' when 4.6.0 was downloaded and caused the compilation error.

I didn't see any GHA used to installed languages so there is no convention to follow for adding this to the matrixed testing. I find testing the oldest and newest versions of languages and things to be the best use of resources to convert a range of versions.

> 2. I think we need to add note in `Doc/Manual/R.html`, which versions are supported, a minimum and the latest version 4.6.0 with the release dates of these revisions.

This would be useful, determining what the SWIG commenting is able to support is beyond the scope of this bug fix. Perhaps if #1 is implemented that will guide what can be tested.

> 3. I suggest adding a version verification in `configure.ac`.

This is dependent on 2.

erezgeva:
I can not build older `R` then 4.0
The r-core fail building with
`multiple definition of  'ptr_R_ProcessEvents'`

I tried the following versions:
4.0.0, 4.1.0, 4.2.0, 4.3.0, 4.4.0, 4.5.0, 4.6.0
Version 4.5.0, 4.6.0 does not pass testing.

I will share the changes to `configure.ac`, `.github/workflows/linux.yml` and `Tools/CI-linux-install.sh`, so you can test it yourself.

Erez

erezgeva:
> > My suggestion:
> 
> > 3. I suggest adding a version verification in `configure.ac`.
> 
> I think these are all beyond the scope of fixing this compilation error due to non-API macros being used. But here are my thoughts:
> 
> > 1. Looking on in `.github/workflows/linux.yml` test only 2 R versions each for each Ubuntu version.
> >    Is there a way to test other versions?
> >    Perhaps download original release and build it?
> 
> In SimpleITK we use the GHA[ r-lib/actions/setup-r](https://github.com/SimpleITK/SimpleITK/blob/d8948b580fefe42a0c94f31e328277506d8cb123/.github/workflows/Build.yml) to get R installed. We had the version set to the 'latest' when 4.6.0 was downloaded and caused the compilation error.

This PR is in the SWIG project.
We need to be explicit to our costumers which versions they can use.


> 
> I didn't see any GHA used to installed languages so there is no convention to follow for adding this to the matrixed testing. I find testing the oldest and newest versions of languages and things to be the best use of resources to convert a range of versions.

So we are the first :smile: 
Just do some more cleaning and I will share it with you.

> 
> > 2. I think we need to add note in `Doc/Manual/R.html`, which versions are supported, a minimum and the latest version 4.6.0 with the release dates of these revisions.
> 
> This would be useful, determining what the SWIG commenting is able to support is beyond the scope of this bug fix. Perhaps if #1 is implemented that will guide what can be tested.
> 
> > 3. I suggest adding a version verification in `configure.ac`.
> 
> This is dependent on 2.

erezgeva:
[0001-Add-versioned-R-testing.patch](https://github.com/user-attachments/files/27282879/0001-Add-versioned-R-testing.patch)

Add tests with versions: 4.0.0, 4.5.0, 4.6.0

You can see the actual used R version in the `Configure` step.
Search `checking for R version`
And in the `Test` step start with `+ make check-r-version`.

Currently Ubuntu 22.04 uses version 4.3.3
And Ubuntu 24.04 uses version 4.5.3

I want to emphasise:
Fixing compilation in SWIG means the CI tests with R 4.6.0 pass in SWIG CI.

blowekamp:
Please push to the PR branch for your additional changes.

erezgeva:
> Please push to the PR branch for your additional changes.

The change need to be in a single PR.
As my patch will not pass as version 4.6.0 will fail without your patch.
So my only alternative is to open a new PR with your patch.
Alternatively, you can `git am` my patch here and prove your fix works.

blowekamp:
> > Please push to the PR branch for your additional changes.
> 
> The change need to be in a single PR. As my patch will not pass as version 4.6.0 will fail without your patch. So my only alternative is to open a new PR with your patch. Alternatively, you can `git am` my patch here and prove your fix works.

Thank you for sharing the additional changes you required. I have added them to this PR.

The PR had "Allow edits and access to secrets by maintainers" enabled. I assume you could push your changes to the PR branch to update:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

erezgeva:
@wsfulton 
Ubuntu `ppa:ubuntu-toolchain-r/test` seems offline.
All test with `GCC:` 10 to 14 fails.
`ppa:ondrej/php`, `ppa:deadsnakes/ppa` and `ppa:rael-gc/rvm` also seems off.
Seems as Ubuntu PPA is off.
Seems Ubuntu is under attack
https://www.omgubuntu.co.uk/2026/05/ubuntu-websites-ddos-attack

blowekamp:
It looks like all R tests/build passes. I don't have access to trigger retesting of only failed builds

wsfulton:
Are there not any ppa repos we can use or some more efficient way of obtaining different versions of R? Building R and installing it is time consuming - 11 minutes for 4.6.0 and then the real SWIG testing time is much less at about 3 mins. Apart from that, this looks okay to merge in and a rather helpful update. Thanks.

erezgeva:
> ppa repos

The only ppa repo I find is for CRAN packages not the R-CORE itself.
The other option is to build the Debian package itself, which takes more time.

We can however remove 4.5.0, and stay with 4 tests:

- 4.0.0 build from r-project
- Ubuntu 22.04 using version 4.3.3
- Ubuntu 24.04 using version 4.5.3
- The latest 4.6.0 build from r-project

erezgeva:
@blowekamp,

In `Examples/test-suite/r`  there are only 37 tests.
For comparison [SciLab](https://www.scilab.org) have 103.
And [GNU Octave](https://octave.org) have 179.
The more tests, the better a language is supported.

erezgeva:
[Add-R-language-windows-GHA-CI.patch](https://github.com/user-attachments/files/27475203/Add-R-language-windows-GHA-CI.patch)

GitHub Windows image have the latest R installed

This patch uses:
- 4.0.0 build from r-project, takes ~14:00
- Ubuntu 22.04 using version 4.3.3, takes ~4:30
- 2 x Ubuntu 24.04 using version 4.5.3, takes ~5:00 and ~11:00 (GCC 13)
- Windows using 4.6.0, takes ~14:30

@wsfulton hope this is better.
With this patch we use the newer version of R and download only the old version 4.0.0 to ensure backward compatible.

wsfulton:
The new version appeared in the images last night causing master to fail on Linux, see https://github.com/swig/swig/actions/runs/25885787077/job/76077029021. I'm looking at this pull request now, but given R is now available, there is no need to use the Tools/CI-linux-install.sh additions to build it. It's preferable / faster to avoid building R each time.

wsfulton:
Merged in 0601b9ca9401aed5bcf1b018be922b15a8cee92a, thanks. I added the missing "Assisted-by" to the commit message as we require this for AI assisted source code changes.

erezgeva:
> ed to use the Tools/CI-linux-install.sh additions to build it. It's preferable / faster to avoid building R eac

I suggest using the download to build the older version 3.0.0

PR #3412, opened 2026-05-04
Add CI support for testing Swig on Windows ARM64

MugundanMCW:
PR Description:
 
* This PR introduces CI support for building and testing Swig on Windows ARM64 using the MSYS2/MSVC toolchain.
 
Changes Proposed:
 
* Updated the windows.yml workflow to include ARM64 build configurations.
* Replaced NuGet with VCPKG, as NuGet packages are currently available only for Windows x64, while VCPKG supports both Windows x64 and ARM64.
* Switched to using prebuilt Boost libraries in GitHub Actions.
* Added a step to manually install MSYS2 on Windows ARM64 runners, since it is not pre-installed.
* Excluded add-stdcall-alias and mthreads options for Windows ARM64, as they are not supported by the MSYS2 CLANGARM64 compiler.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -47,7 +47,7 @@ jobs:
   # - Microsoft Visual Studio compiler.
   # - MinGW-w64 GCC compiler with MSVCRT or UCRT C standard library.
   # We use 2 package managers
-  # - NuGet packages are compatible with Visual Studio compiler.
+  # - Vcpkg packages are compatible with Visual Studio compiler.
   # - MSYS2 pacman with MSYS2 packages and MING-w64 packages.
   #
   # We try to use already installed software on GitHub windows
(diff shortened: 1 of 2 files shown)


erezgeva:
> * Added a step to manually install MSYS2 on Windows ARM64 runners, since it is not pre-installed.
> * Excluded add-stdcall-alias and mthreads options for Windows ARM64, as they are not supported by the MSYS2 CLANGARM64 compiler.

First question:
Why do you think we need Windows ARM?

NuGet was select for speed, if you wish to replace it with VCPKG, you need to provide installation times.
Looking on https://github.com/actions/runner-images/blob/main/images/windows/Windows11-Arm64-Readme.md
It does have NuGet.



Adding Clang is nice, but I think should be done in a separate PR.
I think there are too many changes in one PR.

The fact windows arm  lack MSYS2 is a big drawback.
Also the long time, the tests almost double in time.

Please avoid changes in `configure.ac`.
The `configure.ac` already support clang on Linux and MacOS, so it does not seems reason to update it.
Most conditional parameters should be done in `windows.yml`.

I recommend on start with few test, 4 at the most.

It is better to add an architecture parameter in the matrix and not use 
`${{ matrix.os == 'windows-11-arm' &&`. this condition that base on image name are bad, as images name tend to increase.

What is the benefit of  "Install Boost (MSVC)" ?
We install it with Nuget, is it faster?

erezgeva:
I repeat my first question.
Why do you think we need Windows ARM64?

You need to convince the maintainer.
This project is using standard C/C++, so it is basically insensitive to architecture selection.
We already use arm64 with MacOS, so there is no real reason to use Windows with ARM64.

wsfulton:
Hi @MugundanMCW, as mentioned, we have too many jobs slowing us down and not enough resources to test every platform around. I wouldn't object to a single extra ARM64 job though as a sanity test, either in the cmake workflow or the windows workflow.

erezgeva:
I check VCPKG:
Two remarks:

1. The VCPKG is longer as it build the package, while NuGet download already build,  the build of pcre2 is ~1 minute.
2. VCPKG do not have Bison, "MarkusJx/install-boost@v2" takes ~2 minutes, while NuGet Bison takes 2 seconds and work on ARM64.

My suggestion for using MSVC is 
1. Always Install Bison with NuGet
2. Install pcre2 with NuGet on AMD64 and VCPKG on ARM64
3. Use `RUNNER_ARCH`  to determine architecture see: https://docs.github.com/en/actions/reference/workflows-and-actions/variables
4. You can use bash `${RUNNER_ARCH,,}` to convert to lowecase, i.e. `ARM64` to `arm64`.

I added `windows-11-arm` in `cmake.yml`
See: #3423

wsfulton:
Thanks @erezgeva, I've merged #3423. Closing.

PR #3414, opened 2026-05-06
Python: Support multi-argument typemaps for pytyping

Nerixyz:
To support multi-argument typemaps for `pytyping`, we need to use `Swig_typemap_attach_parms` and lookup with `"tmap:pytyping"`. Before, only the parameter would receive the type of the first argument to the typemap. Users couldn't overwrite that type properly.

I added a test that uses the typemap described in the docs (https://www.swig.org/Doc4.4/Python.html#Python_nn60).


wsfulton:
Commits are now in master in ddb72429fbeb437d1311ac3a692377e06f1e2938 with a followup 42149f507649be6791b412121e6b519cc198294c commit.

PR #3416, opened 2026-05-07
Fix #3415 - Segfault in Allocate::usingDeclaration when using brings in a template method that also has a protected concrete override

### LINKED ISSUE #3415: Segfault in Allocate::usingDeclaration when using brings in a template method that also has a protected concrete override
SWIG segfaults when processing a class that has:
1. A public using Base::method where Base::method is a template with no SWIG-visible instantiations
2. A protected concrete override of the same method name in the derived class

Minimal reproducer:

```
%module using_member_template

%inline %{

struct Impl {};

struct Base {
  template <typename T>
  T *getImpl() const { return nullptr; }
};

struct Derived : public Base {
  using Base::getImpl;

protected:
  Impl *getImpl() const { return nullptr; }
};

%}
```

https://github.com/jmarrec/swig/commit/b5864374e75fbec1f2fe8e263d6539406b6f6709

A fuller concrete example can be found on https://godbolt.org/z/9fE1asdhT

jmarrec:
- Fix #3415

In Source/Modules/allocate.cxx, usingDeclaration() splices the using node out of the sym:overloaded linked list. When the using node is the head of that list (firstoverloaded == n), it reassigns:

https://github.com/swig/swig/blob/226cebfb9892aae5e27f39911a035232f2275867/Source/Modules/allocate.cxx#L1119

```
fc = firstChild(n)
ns = sym:nextSibling
```

When the referenced template has no instantiations visible to SWIG, no child nodes are added (ccount == 0, unodes == null), and if the using node is also the sole entry in the overloaded list (ns == null), firstoverloaded becomes null. The code then unconditionally calls:

https://github.com/swig/swig/blob/226cebfb9892aae5e27f39911a035232f2275867/Source/Modules/allocate.cxx#L1161-L1165

`clean_overloaded(firstoverloaded);   // null → crash in DohGetattr`

Fix:

Guard both calls against the null case (?)


wsfulton:
Thanks @jmarrec. Did you see the test-suite breakages? Should be easy to fix, by replacing `null` with 0 as the test case needs to be c++98 compliant. Also remove the unnecessary whitespace change. Thanks.

wsfulton:
Also it would be nice to add a runtime test, say java, python, whichever you prefer, that replicates the c++ test you have in godbolt.

jmarrec:
@wsfulton I don't think I did a whitespace change? The debug stuff is scoped inside the if statement now so it's indented.

I didn't have the pretention of making a perfect PR, just contributing *something*.

I'll try to do what you asked but might fumble and ask for help!

Source/Modules/allocate.cxx
```diff
@@ -1159,10 +1159,12 @@ class Allocate : public Dispatcher {
...
                 Delattr(n, "sym:overloaded");
                 Delattr(n, "sym:overname");
-                clean_overloaded(firstoverloaded);
+                if (firstoverloaded) {
+                  clean_overloaded(firstoverloaded);
 #ifdef DEBUG_OVERLOADED
-                show_overloaded(firstoverloaded);
+                  show_overloaded(firstoverloaded);
 #endif
+                }
```
jmarrec, line 1167: The whitespace is needed, because it's scoped to the if() now.

Full code without the noise of the of the diff

https://github.com/jmarrec/swig/blob/1d35ddb7b494f13498448ab157e70a5d1e6bd641/Source/Modules/allocate.cxx#L1162-L1167


```c++
                if (firstoverloaded) {
                  clean_overloaded(firstoverloaded);
#ifdef DEBUG_OVERLOADED
                  show_overloaded(firstoverloaded);
#endif
                }
```

Examples/test-suite/python/cpp11_using_member_template_runme.py
```diff
@@ -0,0 +1,10 @@
+from swig_test_utils import swig_assert, swig_check
+import cpp11_using_member_template as m
+
+d = m.Derived()
+swig_check(d.value(), 1)
+
+md = m.MoreDerived()
+swig_check(md.value(), 2)
+swig_assert(isinstance(md, m.Derived))
+swig_assert(isinstance(md, m.Base))
```
jmarrec, line 10: runtime test

Examples/test-suite/cpp11_using_member_template.i
```diff
@@ -0,0 +1,72 @@
...
+  std::shared_ptr<ImplType> getImpl() const {
+    return Base::getImpl<ImplType>();
+  }
+};
+
+// A further-derived class exercises the inherited using declaration path.
+struct MoreDerived : public Derived {
+  MoreDerived() : Derived(std::shared_ptr<detail::Derived_Impl>(new detail::MoreDerived_Impl())) {}
+};
+%}
```
jmarrec, line 72: Full example, matches https://godbolt.org/z/9fE1asdhT

jmarrec (commented):
@wsfulton I've added a full example targetting c++11 (shared_ptr...) and removed the original one that wasn't as interesting (it's still in the branch history, so we can put it back if needed, revert 1d35ddb7b494f13498448ab157e70a5d1e6bd641 in that case)

hopefully I got it right, let me know if not.

wsfulton:
Thanks @jmarrec, tests are just what is needed and fix is good.

PR #3421, opened 2026-05-15
Skip Ubuntu Toolchain PPA if versioned GCC already installed

erezgeva:
Most versioned GCC are already installed on GitHub Ubuntu images. 

Although we only spare ~10 seconds of adding the PPA repository.
We skip potential errors in case the PPA is not reachable.

### DIFF
--- a/Tools/CI-linux-install.sh
+++ b/Tools/CI-linux-install.sh
@@ -9,11 +9,15 @@ if [[ "$compiler" = 'clang' ]]; then
 	CC="clang"
 	CXX="clang++"
 elif [[ -n "$GCC" ]]; then
-	$RETRY sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
-	$RETRY sudo apt-get -qq update
-	$RETRY sudo apt-get -qq install g++-$GCC
 	CC="gcc-$GCC"
 	CXX="g++-$GCC"
+	if ! dpkg -l 'g++*' | grep "^ii  $CXX " > /dev/null; then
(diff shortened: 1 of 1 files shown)


Tools/CI-linux-install.sh
```diff
@@ -9,11 +9,15 @@ if [[ "$compiler" = 'clang' ]]; then
 	CC="clang"
 	CXX="clang++"
 elif [[ -n "$GCC" ]]; then
-	$RETRY sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
-	$RETRY sudo apt-get -qq update
-	$RETRY sudo apt-get -qq install g++-$GCC
 	CC="gcc-$GCC"
 	CXX="g++-$GCC"
+	if [[ ! -x "$(which $CC || true)" ]] || [[ ! -x "$(which $CXX || true)" ]]; then
```
wsfulton: I don't follow the logic here. Can you make it a bit simpler using variable names or something to indicate the intention?

erezgeva: The logic is very simple.
We want to use `gcc-n` and `g++-n`.
We mark the wanted values in `CC` and `CXX` as before.
We check if the versioned compilers are already installed.
If they are NOT already installed, we install the versioned compilers from the PPA.

Basic logic, does not really need explicit comment.

erezgeva: We can use this test
`if ! dpkg -l 'g++*' | grep "^ii  $CXX" > /dev/null; then`

Same as we do with PHP.
The compilers do comes from the packages.

Would that be better?

wsfulton (approved):

PR #3422, opened 2026-05-15
Add versioned `R` testing and windows GHA CI

erezgeva:
- Add versioned `R` testing with version 4.0.0, to ensure backward compatible
- Add check for `R` version.
- Add windows testing with already installed `R`
  - GitHub tend to update Window first.
  -  Update `R` test-suite to support windows build.


wsfulton:
R tests should run in Examples/test-suite/r, like all the other scripting languages. What's the motivation to change this to run in a subdirectory?

erezgeva:
> R tests should run in Examples/test-suite/r, like all the other scripting languages. What's the motivation to change this to run in a subdirectory?

Well most are, some not: C#, d and C as target uses sub-folder.

The reason is that the build is done with R own Makefile which create temporary files with fixed names (at least on Windows).
As Makefile run the tests in parallel, the files collide and tests fails.

The basic example (which R have only 2), already build in separated folders, so they do not need any further tweak.

Although it happens on Windows, as the Makefile comes from `R`, it may fail with Linux in the future.

The Makefile is called by `R CMD SHLIB`.

wsfulton:
It's only the non-scripting languages that use a sub director, so will be out of step with the other accepting languages. This needs a comment in the Makefile.in as to why a sub directory is needed. What are the name of the files that clash. I'm a bit surprised in this day and she that concurrent builds are a problem, but there you go! Maybe there is an option to move the temp files somewhere else?

erezgeva:
> It's only the non-scripting languages that use a sub director, so will be out of step with the other accepting languages. This needs a comment in the Makefile.in as to why a sub directory is needed. What are the name of the files that clash. I'm a bit surprised in this day and she that concurrent builds are a problem, but there you go! Maybe there is an option to move the temp files somewhere else?

`R CMD SHLIB` itself do not provide temporary folder build option.
So we end up with only implement one ourselves.

As the build it done in `Examples/Makefile.in` and as the problem is in `Examples/test-suite/r`, while sub-folders of  `Examples/r` are already sub-folders and do not need any further change.

It seems more simple to take an already prove solution.
You can use the same argument of using a temporary folder to build C#, D and C.
If we think the build should be done in temporary folder, it should be a more broad usage by more than the `R` language.

Notice that you statement for non-script is incorrect as Java  which is a non-script build without sub-folder.

I think the more proper distinguish is full controlled builds versa non controlled builds.

I know `D` simply follows C#. And I am not sure why C# or C needs sub-folders. The fact that they are non-script do not explain it.

`Go` need sub-folders because it uses the external `go` tool to build.
See my PR #3399

PR #3424, opened 2026-05-17
Fix PHP `%typemap(directorin)` for pointers

erezgeva:
- `director_classes`
  - Add `ConstPtrCall()` method and add it to tests
  - Remove numbering from `D`, it is easier to use function names. And easier to modify in future.
  - Fix PHP test
  - Fix Perl test
  - Add Python and Ruby tests

### DIFF
--- a/Lib/php/php.swg
+++ b/Lib/php/php.swg
@@ -461,13 +461,21 @@
   SWIG_SetPointerZval($result, (void *)*$1, $*1_descriptor, $owner);
 %}
 
-%typemap(directorin) SWIGTYPE *,
-                     SWIGTYPE [],
-                     SWIGTYPE &,
-                     SWIGTYPE &&
+/* map C++ reference to SWIG pointer and C++ reference of pointer to SWIG pointer of pointer */
+%typemap(directorin) SWIGTYPE &,
+                     SWIGTYPE &&,
+                     SWIGTYPE *const&
 %{
   ZVAL_UNDEF($input);
-  SWIG_SetPointerZval($input, (void *)&$1, $1_descriptor, $owner);
+  SWIG_SetPointerZval($input, (void *)&($1), $1_descriptor, $owner);
+%}
+
+/* map C/C++ pointer to SWIG pointer */
+%typemap(directorin) SWIGTYPE *,
+                     SWIGTYPE []
+%{
+  ZVAL_UNDEF($input);
+  SWIG_SetPointerZval($input, (void *)($1), $1_descriptor, $owner);
 %}
 
 %typemap(out, phptype="SWIGTYPE") SWIGTYPE (CLASS::*)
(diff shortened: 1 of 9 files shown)


Lib/php/php.swg
```diff
@@ -461,13 +461,21 @@
   SWIG_SetPointerZval($result, (void *)*$1, $*1_descriptor, $owner);
 %}
 
+/* References and pointers of pointer */
+%typemap(directorin) SWIGTYPE &,
+                     SWIGTYPE &&,
+                     SWIGTYPE *&
```
ojwb: That's a reference to a pointer, so "pointers of pointer" seems wrong in the comment above.

Is this case really correctly handled this way?  I'd expect the generated code would need to look like the code for `SWIGTYPE *` not like the code for `SWIGTYPE &`.

erezgeva: <del>It is correct that neither of these are C/C++ pointer of pointer.
But the SWIG generate pointer of pointer type i.e. `SWIGTYPE_p_p_XXX`
Which means this is a pointer of pointer typemap.
Not sure why the SWIG use this typemap?
Perhaps add TODO or a question mark?</del>

To be more accuracy.
The typemap maps
C++ reference to swig pointer and C/C++ reference of pointer to swig pointer of pointer.

When looking on `ConstPtrRef` of `director_classes` in python, ruby and Perl5,  they use swig pointer to object.
Only PHP uses pointer of pointer to object.
Not really sure what is the proper.
Yet the `director_classes` test assume it is passing pointer to `DoubleHolder` and NOT pointer of pointer.

We should probably open a discussion on the matter.
In this PR, I focus on fixing PHP `Ptr` and `ConstPtr` of `director_classes`.
Open an issue to discuss the matter #3447

ojwb: I'm not sure why PHP would be doing something different here.

But anyway, the point that's relevant to the proposed change here is that this comment you've added seems confusing.

ojwb: Oh I see, you've updated that (but github was showing the old version).

I guess the new version's OK.

ojwb:
The PHP part looks plausible to me, though I didn't review in full details.  I don't have time or energy right now to review all the other languages touched though, sorry.

Lib/php/php.swg
```diff
@@ -461,13 +461,21 @@
   SWIG_SetPointerZval($result, (void *)*$1, $*1_descriptor, $owner);
 %}
 
+/* map C++ reference to SWIG pointer and C++ reference of pointer to SWIG pointer of pointer */
+%typemap(directorin) SWIGTYPE &,
+                     SWIGTYPE &&,
+                     SWIGTYPE *&
```
wsfulton: SWIG does not provide typemaps for `SWIGTYPE *&`. It does however provide typemaps for `SWIGTYPE *const&` though, so maybe you meant to use the latter.

erezgeva: The test match your description.
`DoubleHolder *const& ConstPtrRef(DoubleHolder *const& cprx)`
I can rename the typemap, and test it.

erezgeva: I change it to `SWIGTYPE *const&`
The conversion remains.
It is C++ reference of pointer to SWIG pointer of pointer.
Simply a C++ constant reference.
It convert to `SWIGTYPE_p_p_DoubleHolder`, pointer to pointer of `DoubleHolder`.

While all other languages use pointer of `DoubleHolder`.
As described in #3447

erezgeva:
@wsfulton  You may squash the last commit after review.

PR #3429, opened 2026-05-20
Add std::optional support

jschueller:


Maxhy:
+1 for such official support

PR #3434, opened 2026-05-20
Fix duplicate using declaration

### LINKED ISSUE #2933: Duplicate using declarations for `T` break `std::vector<T>` wrapping
This is a rather crazy one and it took me a long time to make a SSCCE example for it as I couldn't believe that simply duplicating a using declaration could result in this, but here is a minimal diff showing the problem:

```diff
diff --git a/Examples/test-suite/li_std_vector.i b/Examples/test-suite/li_std_vector.i
index a76604b3d..3bf4cc51f 100644
--- a/Examples/test-suite/li_std_vector.i
+++ b/Examples/test-suite/li_std_vector.i
@@ -119,6 +119,8 @@ namespace aa {
     Holder(int n = 0) : number(n) {}
     int number;
   };
+
+  struct UsedFromNS {};
 }
 %}
 
@@ -131,6 +133,10 @@ std::vector< ::aa::Holder > vec1(std::vector< ::aa::Holder > x) { return x; }
 %}
 #endif
 
+using aa::UsedFromNS;
+using aa::UsedFromNS; // Remove this to make the test pass
+%template(VectorUsedFromNS) std::vector< UsedFromNS >;
+
 // exercising vectors of strings
 %inline %{
 std::vector<std::string> RevStringVec (const std::vector<std::string> &In)
```

With this change, the test breaks for all languages (tested with C#, Java, Python) because the generated code uses unqualified `UsedFromNS` instead of `aa::UsedFromNS` that is used with just a single `using aa::UsedFromNS` line in all `std::vector<>` occurrences in the generated code.

I have no idea what's going on here yet, but this used to work before, so I'll try to bisect it and hopefully fix it, but opening this issue now just in case somebody already has some idea.

jschueller:
`Source/Swig/symbol.c:Swig_symbol_cadd()` chains duplicate `using` declarations (e.g., two `using aa::UsedFromNS;`) via `csym:nextSibling`. When `Swig_symbol_clookup()` later resolves the unqualified name `UsedFromNS`, it sees `csym:nextSibling` set and treats it as "overloaded using declarations", breaking out of the resolution loop and returning the `using` node itself instead of following `uname` to the underlying `aa::UsedFromNS` class node. The type qualification then fails, generating `std::vector< UsedFromNS >` instead of `std::vector< aa::UsedFromNS >` in wrapper code.

In **`Source/Swig/symbol.c:615`**, added a new `else if` clause in `Swig_symbol_cadd()` that detects when both the existing entry (`cn`) and the new node (`n`) are `using` declarations with identical `uname` attributes, and returns early without chaining. This prevents the spurious `csym:nextSibling` that confuses `Swig_symbol_clookup`.

Closes #2933

### DIFF
--- a/Examples/test-suite/li_std_vector.i
+++ b/Examples/test-suite/li_std_vector.i
@@ -119,8 +119,14 @@ namespace aa {
     Holder(int n = 0) : number(n) {}
     int number;
   };
+
+  struct UsedFromNS {};
 }
 %}
+
+using aa::UsedFromNS;
+using aa::UsedFromNS; // Duplicate using declarations (regression test for #2933)
+%template(VectorUsedFromNS) std::vector< UsedFromNS >;
 
 #if !defined(SWIGOCTAVE)
 // To fix: something different in Octave is preventing this from working
--- a/Source/Swig/symbol.c
+++ b/Source/Swig/symbol.c
@@ -607,6 +607,10 @@ void Swig_symbol_cadd(const_String_or_char_ptr name, Node *n) {
     Swig_error(Getfile(n), Getline(n), "Declaration of '%s' shadows template parameter,\n", name);
     Swig_error(Getfile(cn), Getline(cn), "previous template parameter declaration '%s'.\n", name);
     return;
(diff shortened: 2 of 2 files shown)


wsfulton:
Thanks, merged in as e2fca65752ad78f484e807d1938ec806c2ae5824 and a follow up commit for testing 65fbe448ac5fe20713b05b0b7a0cb0a01eb66224.

PR #3435, opened 2026-05-20
Python: Implement nspace support

### LINKED ISSUE #2289: Error Python Enum Value is multiply defined
Hello,

Swig seems to ignore c++ namespaces enum values in python (csharp and java works).

Example:
```cpp
namespace M_Common {
  namespace OperationModeStateKind {
    enum Type {
      Normal,
      Manual,
      Auto,
      Calibrating,
      Maintenance
    };
  }
}
namespace M_SignalProcessing {
  namespace ThresholdingKind {
    enum Type {
      Manual,
      Offset,
      Global
    };
  }
}
```
My swig file:
```
%rename(M_Common_OperationModeStateKind_Type) M_Common::OperationModeStateKind::Type;
%rename(M_SignalProcessing_ThresholdingKind_Type) M_SignalProcessing::ThresholdingKind::Type;
```

I got the following Error:
```bash
swig.exe -c++ -python my_swig.i
. . .
C:\SignalProcessingTypes.h(173) : Error: 'Manual' is multiply defined in the generated target language module.
C:\CommonTypes.h(437) : Error: Previous declaration of 'Manual'
```

Is this a swig bug?
Are there any workarounds to solve this? 🙈

jschueller:
Add %nspace feature support for the Python target language. Names from C++ namespaces are prefixed into Python symbol names using underscore separators (Outer::Inner1::Color -> Outer_Inner1_Color), the same approach used by the Lua module.

Key changes:
- Prefixed enum value names in constantWrapper() to avoid collisions
- Added nspace-aware class naming in classHandler() and classDeclaration()
- Replaced all NSPACE_TODO (= 0) with getNSpace() in Swig_name_* calls
- Fixed variableWrapper() to avoid double-namespacing
- Updated Language::staticmembervariableHandler() to use NSpace for unique static member wrapper names across namespaces
- Enabled SWIGPYTHON in 6 nspace test interface files
- Created python/nspace_runme.py test

Closes #2289

### DIFF
--- a/Examples/test-suite/cpp17_nspace_nested_namespaces.i
+++ b/Examples/test-suite/cpp17_nspace_nested_namespaces.i
@@ -5,7 +5,7 @@ SWIG_JAVABODY_PROXY(public, public, SWIGTYPE)
 #endif
 
 // nspace feature only supported by these languages
-#if defined(SWIGJAVA) || defined(SWIGCSHARP) || defined(SWIGD) || defined(SWIGLUA) || defined(SWIGJAVASCRIPT) || defined(SWIGPYTHON)
+#if defined(SWIGJAVA) || defined(SWIGCSHARP) || defined(SWIGD) || defined(SWIGLUA) || defined(SWIGJAVASCRIPT) || defined(SWIGPYTHON) || defined(SWIGPHP)
 %nspace;
 #endif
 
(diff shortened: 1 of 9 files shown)


jschueller:
maybe #3498 is better

PR #3440, opened 2026-05-20
Use Py_TPFLAGS_MANAGED_DICT for builtin wrappers on Python >= 3.12

### LINKED ISSUE #3202: Use Py_TPFLAGS_MANAGED_DICT instead of tp_dictoffset for builtin wrappers
For python-3.12+ one could use [Py_TPFLAGS_MANAGED_DICT](https://docs.python.org/3/c-api/typeobj.html#c.Py_TPFLAGS_MANAGED_DICT) instead of __dict__ (and similar for weakref), for memory/speed improvements in SwigPyObject used by the builtin wrappers. (from #3196). Currently we use https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_dictoffset.

As documented in the links above, this will require also require using [Py_TPFLAGS_HAVE_GC](https://docs.python.org/3/c-api/typeobj.html#c.Py_TPFLAGS_HAVE_GC). To be researched: side effects and consequences of turning on garbage collection for builtin types.

jschueller:
For the heaptypes (builtin) path, set Py_TPFLAGS_MANAGED_DICT and Py_TPFLAGS_MANAGED_WEAKREF in the type spec instead of relying on tp_dictoffset/tp_weaklistoffset. The instance dict is now accessed via _PyObject_GetDictPtr rather than direct sobject->swigdict access, and manual swigdict/weakreflist management is skipped at object creation and destruction time.  Non-builtin types continue to use the existing tp_dictoffset mechanism.

Closes #3202

### DIFF
--- a/Lib/python/builtin.swg
+++ b/Lib/python/builtin.swg
@@ -580,9 +580,11 @@ SWIGINTERN void
 SwigPyBuiltin_destructor_closure(SwigPyWrapperFunction wrapper, const char *wrappername, PyObject *a) {
   SwigPyObject *sobj;
   sobj = (SwigPyObject *)a;
+#if PY_VERSION_HEX < 0x030c0000
   if (sobj->weakreflist != NULL)
     PyObject_ClearWeakRefs(a);
   SWIG_Py_XDECREF(sobj->swigdict);
+#endif
   if (sobj->own) {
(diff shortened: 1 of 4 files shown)


jschueller:
seems not possible

PR #3441, opened 2026-05-20
Go: Fix director returning reference to local variable

### LINKED ISSUE #3398: [GO] director_frob - function returns address of local variable
Hi

Running  `make director_frob.cpptest`
Gives 
```
director_frob_wrap.cxx: In member function 'virtual SwigDirector_Ops::operator int&()':
director_frob_wrap.cxx:590:11: warning: function returns address of local variable [-Wreturn-local-addr]
  590 |   return *c_result;
      |           ^~~~~~~~
director_frob_wrap.cxx:587:7: note: declared here
  587 |   int f = (int)*result;
      |       ^

```

jschueller:
The directorout typemaps for int& and int&& in Lib/go/go.swg created a local variable and returned its address:

  int f = (int)*result;
  c_result = (int *)&f;
  return *c_result;      // warning: function returns address of local variable

Fix by heap-allocating the value with 'new' and tracking it with swig_acquire_pointer, matching the existing pattern used by const int&.

  c_result = new int((int)*result);
  swig_acquire_pointer(&swig_mem, c_result);
  return *c_result;

Closes #3398

### DIFF
--- a/Examples/test-suite/director_frob.i
+++ b/Examples/test-suite/director_frob.i
@@ -80,6 +80,10 @@
     virtual unsigned long long ull(unsigned long long i, unsigned long long j) { return i + j; }
     unsigned long long callull(int i, int j) { return ull(i, j); }
   };
+
+  int take_ops_intref(Ops& ops) {
+    return ops.operator int&();
+  }
 %}
 
 
--- a/Examples/test-suite/go/director_frob_runme.go
+++ b/Examples/test-suite/go/director_frob_runme.go
@@ -9,4 +9,12 @@ func main() {
 	if s != "Bravo::abs_method()" {
 		panic(s)
 	}
+
+	// Test take_ops_intref with a non-director Ops (no virtual dispatch)
+	ops := NewOps()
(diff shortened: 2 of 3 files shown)


erezgeva:
The fix is explained.

But no explanation on the test changes.
What is the purpose of this change.
I would except explanation in the commit message in addition to tests files.

Secondly the addition test is with a non-director test.
Perhaps we can add a non-director test which is missing in Go and actually test the feature?
I'd rather use director tests to actually test director.

wsfulton:
The `-Wreturn-local-addr` warning only seems to appears in combination with optimisation warning `-O2`. Merged in 677034fa7670870d8b899f6fbcb7765f973bbb8d. This patch missed the warning from the source typemaps that were copied and didn't test the new typemaps. When I added some tests there was another problem which, see follow on commits 677034fa7670870d8b899f6fbcb7765f973bbb8d...e1374a8e0b5a002163e7c8ac1b8ccacf142791d9. Anyway all sorted now.

PR #3442, opened 2026-05-20
Ruby: Fix crash in GC_register when obj is Qnil

### LINKED ISSUE #3385: [ruby] ruby 3.3 on Linux
For example here:
https://github.com/swig/swig/actions/runs/23414103303/job/68106498857

Looking on `Segmentation fault`

`ruby 3.3.10 (2025-10-23 revision 343ea05002) [x86_64-linux]`

Crash happens on in GC:

Inside `GC_register()` when calling `rb_hash_aref()`
`GC_register()` is called from` GC_VALUE(const GC_VALUE& item)`  constructor
called by `BinaryPredicate(VALUE obj = Qnil) `constructor

```
namespace swig {
  class SwigGCReferences {
    VALUE _hash;
    void GC_register(VALUE& obj) {
      if (FIXNUM_P(obj) || SPECIAL_CONST_P(obj) || SYMBOL_P(obj))
        return;
      if (_hash != Qnil) {
=>      VALUE val = rb_hash_aref(_hash, obj);
        unsigned n = FIXNUM_P(val) ? NUM2UINT(val) : 0;
        ++n;
        rb_hash_aset(_hash, obj, INT2NUM(n));
      }
    }
    ...
  }
  class GC_VALUE {
  protected:
    VALUE  _obj;
  public
    GC_VALUE(const GC_VALUE& item) : _obj(item._obj)
    {
      SwigGCReferences::instance().GC_register(_obj);
    }
    ...
  }
  ...
  template <class _T = GC_VALUE, class _DefaultFunc = std::less<GC_VALUE> >
  struct BinaryPredicate : GC_VALUE
  {
    BinaryPredicate(VALUE obj = Qnil) : GC_VALUE(obj) { }
    ...
  }
```

jschueller:
The BinaryPredicate default constructor (VALUE obj = Qnil) passes Qnil to GC_VALUE, which calls GC_register(Qnil).  GC_register already guards against FIXNUM/SPECIAL_CONST/SYMBOL but not against NIL_P, so rb_hash_aref(_hash, Qnil) was called, causing a segfault in Ruby 3.3's GC.

Add NIL_P(obj) to the early-return checks in both GC_register and GC_unregister.

Closes #3385

### DIFF
--- a/Lib/ruby/rubyclasses.swg
+++ b/Lib/ruby/rubyclasses.swg
@@ -66,7 +66,7 @@ namespace swig {
       }
     }
     void GC_register(VALUE& obj) {
-      if (FIXNUM_P(obj) || SPECIAL_CONST_P(obj) || SYMBOL_P(obj))
+      if (NIL_P(obj) || FIXNUM_P(obj) || SPECIAL_CONST_P(obj) || SYMBOL_P(obj))
         return;
       if (_hash != Qnil) {
         VALUE val = rb_hash_aref(_hash, obj);
@@ -76,7 +76,7 @@ namespace swig {
       }
     }
     void GC_unregister(const VALUE& obj) {
-      if (FIXNUM_P(obj) || SPECIAL_CONST_P(obj) || SYMBOL_P(obj))
+      if (NIL_P(obj) || FIXNUM_P(obj) || SPECIAL_CONST_P(obj) || SYMBOL_P(obj))
         return;
       // this test should not be needed but I've noticed some very erratic
       // behavior of none being unregistered in some very rare situations.


Lib/ruby/rubyhead.swg
```diff
@@ -1,3 +1,13 @@
+/* Ruby >= 3.4 deprecates the untyped Data_* API (Data_Wrap_Struct,
+ * Data_Get_Struct) in favour of TypedData.  SWIG-generated wrappers
+ * use these macros throughout the file body, not just at the include
+ * site, so these suppressions are file-scoped (no push/pop). */
```
erezgeva: Do you have a link/reference to that?

jschueller: no, this comment seems incorrect

jschueller: turns out its much more involved than that: its was hiding an api migration to typed data, I will drop this commit

wsfulton:
Thanks @jschueller for digging into this. I spent some time reproducing #3385 on a real Ruby 3.3 build, and I don't think this change can fix it - the `Qnil` case never reaches the crashing code, and the actual fault is on a different path. Details below so we can track the real fix.

### 1. `NIL_P(obj)` is already covered by `SPECIAL_CONST_P(obj)`

`Qnil` is a special constant (an immediate: value `0x04`, inside the `0x07` immediate mask), so `SPECIAL_CONST_P(Qnil)` is already true. The existing guard

```c
if (FIXNUM_P(obj) || SPECIAL_CONST_P(obj) || SYMBOL_P(obj))
  return;
```

therefore already returns early for `Qnil` before `rb_hash_aref` is reached. Adding `NIL_P(obj)` returns for exactly the same set of inputs - it doesn't change behaviour. A one-line check confirms `SPECIAL_CONST_P(Qnil) == 1`, and instrumenting `GC_register` over a full `li_std_functors` run (tens of thousands of calls, thousands with `obj == Qnil`) shows zero `Qnil` values ever reaching `rb_hash_aref` - all are caught by the existing guard. So the `Qnil` in the original backtrace was a red herring (it reads that way under `-O2`, but it can't actually get past the guard).

### 2. The crash reproduces under GC compaction, deterministically

On an ubuntu-22.04 container with RVM Ruby 3.3.11 (matching CI), the test passes normally, but the segfault reproduces reliably (8/8 here) under compaction stress with a long-lived container:

```ruby
require "li_std_functors"
M = Li_std_functors::Map
maps = []
40.times { maps << M.new(proc { |a, b| b < a }) }   # comparator procs stored in C++
maps.each_with_index { |m, i| m["k#{i}"] = i }
GC.stress = true
150.times do |i|
  GC.verify_compaction_references(expand_heap: true, toward: :empty)  # moves every object
  maps.each_with_index { |m, j| m["x#{i}_#{j}"] = i }                 # reuse stored proc/keys
end
```

### 3. Root cause: a raw `VALUE` dangling after compaction, in the comparator

gdb backtrace at the crash:

```
#0  rb_id_table_lookup (tbl=0x0, id=3425, ...)                         id_table.c:230
...
#11 rb_funcall (...)
#12 swig::BinaryPredicate<swig::GC_VALUE, std::less<swig::GC_VALUE> >::operator()(...)
#13 std::_Rb_tree<...>::_M_lower_bound (...)
#16 std::map<...>::operator[] (...)
#17 std_map_..._setitem__ (...)
#18 _wrap_Map___setitem__ (...)
```

`swig::GC_VALUE` (`Lib/ruby/rubyclasses.swg`) stores a raw Ruby `VALUE` in C++ - both the `std::map`/`std::set` keys and the `BinaryPredicate` comparator proc. Ruby 3.x compaction relocates those objects but cannot update SWIG's C++ copies, so they dangle. The fatal access is the comparator calling `rb_funcall` on the moved proc: `rb_id_table_lookup(tbl=0x0)` because the relocated slot no longer has a class/method table.

Note this is **not** in `GC_register`. Instrumenting `GC_register` separately did catch a stale object reaching it with `BUILTIN_TYPE == T_NONE` (a freed slot) - never `Qnil` - but even adding a `T_NONE`/`T_MOVED` guard there still segfaults 8/8 via the comparator path above. So a guard inside `GC_register` can't fix this.

### 4. Suggested direction

Because the `Qnil` path already returns early and the crash is on the comparator path, I don't think we should merge this as the fix for #3385, and I'd keep #3385 open. A real fix has to stop the `VALUE`s that SWIG stores in containers from being moved by compaction (pin them), or update them when objects move - a broader change in `GC_VALUE`/`SwigGCReferences` in `Lib/ruby/rubyclasses.swg` rather than a guard in `GC_register`. I think we should pin the objects.

PR #3443, opened 2026-05-20
Python: Fix SWIG_NewPointerObj requiring 'self' in scope with -builtin

### LINKED ISSUE #2613: SWIG_NewPointerObj requires self when using -builtin (Python)
Example to reproduce:
```swig
%module example

%inline %{
struct Foo {
    int n;
};

void foo(PyObject* cb) {
    Foo foo{123};
    auto* arg = SWIG_NewPointerObj(&foo, SWIGTYPE_p_Foo, 0);
    Py_XDECREF(PyObject_CallFunctionObjArgs(cb, arg, NULL));
    Py_DECREF(arg);
}
%}
```
This works fine without -builtin:
```
$ python -c "import example; example.foo(print)"
<example.Foo; proxy of <Swig Object of type 'Foo *' at 0x7f6a1c3632a0> >
```
However, with -builtin it gives a compilation error:
```
example_wrap.cxx: In function ‘void foo(PyObject*)’:
example_wrap.cxx:1091:83: error: ‘self’ was not declared in this scope
 1091 | #define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerObj(self, ptr, type, flags)
      |                                                                                   ^~~~
example_wrap.cxx:3563:17: note: in expansion of macro ‘SWIG_NewPointerObj’
 3563 |     auto* arg = SWIG_NewPointerObj(&foo, SWIGTYPE_p_Foo, 0);
      |                 ^~~~~~~~~~~~~~~~~~
```

For some reason with -builtin `SWIG_NewPointerObj` macro expects `self` in the current scope. This breaks uses in various extension functions. Looking at the code, `SWIG_Python_NewPointerObj` only uses `self` if `SWIG_BUILTIN_TP_INIT` was passed. In this case it does not allocate a new object at all, but uses self instead. This is used to implement init methods. It seems like it would be cleaner to have a separate function to use in init and not overload `SWIG_NewPointerObj` to do completely different things. Or at a minimum, use a different macro in init that explicitly passes `self` and not rely on its presence for normal uses of `SWIG_NewPointerObj`.

I'm currently working this around by using `SWIG_InternalNewPointerObj`, which does not use `self`, but it looks like a private function.

jschueller:
In builtin mode the SWIG_NewPointerObj macro passed 'self' to SWIG_Python_NewPointerObj, requiring 'self' to be in scope at the call site.  This broke user code in %inline blocks that called SWIG_NewPointerObj, because 'self' only exists in the generated tp_init wrapper, not in arbitrary C functions.

1. Change the builtin macro to pass NULL instead of self.
2. Guard the BUILTIN_TP_INIT code path with a self != NULL check, so tp_init still works (self is provided by the generated wrapper) while user code calling SWIG_NewPointerObj gets a normal new object.

Closes #2613

### DIFF
--- a/Lib/python/pyrun.swg
+++ b/Lib/python/pyrun.swg
@@ -18,11 +18,10 @@
 #define SWIG_ConvertPtr(obj, pptr, type, flags)         SWIG_Python_ConvertPtr(obj, pptr, type, flags)
 #define SWIG_ConvertPtrAndOwn(obj,pptr,type,flags,own)  SWIG_Python_ConvertPtrAndOwn(obj, pptr, type, flags, own)
 
-#ifdef SWIGPYTHON_BUILTIN
-#define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerObj(self, ptr, type, flags)
-#else
+/* always passes NULL for self -- user code (%inline, etc.) does not have
+   'self' in scope.  The tp_init path passes it via the SWIG_BUILTIN_INIT flag,
+   handled in SWIG_Python_NewPointerObj.  See issue #2613. */
 #define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerObj(NULL, ptr, type, flags)
-#endif
 
 #define SWIG_InternalNewPointerObj(ptr, type, flags)	SWIG_Python_NewPointerObj(NULL, ptr, type, flags)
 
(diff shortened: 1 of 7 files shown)


Lib/python/pyrun.swg
```diff
@@ -23,7 +23,11 @@
...
 
 #ifdef SWIGPYTHON_BUILTIN
-#define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerObj(self, ptr, type, flags)
+/* User code (e.g. %inline) does not have 'self' in scope, so pass NULL.
+   The tp_init path explicitly passes self through the out typemap's $owner
+   substitution (SWIG_BUILTIN_INIT), which SWIG_Python_NewPointerObj
+   handles specially when self != NULL.  See issue #2613. */
+#define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerObj(NULL, ptr, type, flags)
 #else
 #define SWIG_NewPointerObj(ptr, type, flags)            SWIG_Python_NewPointerObj(NULL, ptr, type, flags)
```
erezgeva: The two definition are identical.
Which means they can be merged.
The `#ifdef SWIGPYTHON_BUILTIN` can be removed.

jschueller: nice catch

PR #3444, opened 2026-05-20
PHP: Inline __get/__set magic methods for simple member variables

### LINKED ISSUE #2809: Make PHP magic __get and __set more efficient
_Opening a new issue for a point I noted in https://github.com/swig/swig/issues/2808#issuecomment-1955203958_

SWIG generates the PHP "magic" `__get` and `__set` methods to implement PHP attribute wrappers for C/C++ member variables (e.g. `$obj->attr`).

Currently behind the scenes these do a C/C++ `if`...`else if`... dispatch on the attribute name and make a PHP method call to `foo_get` or `foo_set` which do the actual wrapping of the member variable.  Making this call requires the method name in a PHP string zval, which gets created as a temporary on each access.

It occurs to me that we could just inline the code from the wrapped getter method into the `__get` magic method instead of generating code to make a PHP function call, which would be lower overhead.  Then we could allow suppressing the wrapped getter method, which is probably redundant in most cases.  Similarly for setters and `__set`.  

This is not really suitable for 4.2.1 though.

jschueller:
Instead of dispatching each property access through a PHP method call (foo_get / foo_set) which requires allocating a zend_string, looking up the method by name, and calling it via
zend_call_known_instance_method, the __get and __set magic methods now access simple public member variables directly through the C++ pointer.

The optimization applies to int, short, long, unsigned, float, double, bool, and char members that are public, non-const, non-array, and non-enum.  All other types (protected/private members, enums, arrays, complex types) fall back to the original method dispatch.

Const and array members now throw a clear type error in __set rather than emitting an invalid assignment.

Closes #2809

### DIFF
--- a/Source/Modules/php.cxx
+++ b/Source/Modules/php.cxx
@@ -2098,26 +2098,104 @@ public:
 
     String *v_name = GetChar(n, "name");
 
-    Printf(magic_set, "\nelse if (strcmp(ZSTR_VAL(arg2),\"%s\") == 0) {\n", v_name);
-    Printf(magic_set, "zend_string *swig_funcname = ZSTR_INIT_LITERAL(\"%s_set\", 0);\n", v_name);
-    Append(magic_set, "zend_function *swig_zend_func = zend_std_get_method(&Z_OBJ_P(ZEND_THIS), swig_funcname, NULL);\n");
-    Append(magic_set, "zend_string_release(swig_funcname);\n");
-    Printf(magic_set, "zend_call_known_instance_method(swig_zend_func, Z_OBJ_P(ZEND_THIS), return_value, 1, &args[1]);\n");
-    Printf(magic_set, "}\n");
-
-    Printf(magic_get, "\nelse if (strcmp(ZSTR_VAL(arg2),\"%s\") == 0) {\n", v_name);
-    Printf(magic_get, "zend_string *swig_funcname = ZSTR_INIT_LITERAL(\"%s_get\", 0);\n", v_name);
-    Append(magic_get, "zend_function *swig_zend_func = zend_std_get_method(&Z_OBJ_P(ZEND_THIS), swig_funcname, NULL);\n");
-    Append(magic_get, "zend_string_release(swig_funcname);\n");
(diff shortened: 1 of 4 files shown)


ojwb:
This looks wrong to me - we should use the code from the typemaps - that's what currently goes into the `__get()` and `__set()` methods rather than hard-coding handling for a subset of common types and punting on others, and what I was suggesting in #2809.

The approach here means that if a user overrides the typemaps for one of these types we'll quietly not use their code.  It also means other types won't benefit from this.

The handling for `T_CHAR` looks wrong too - SWIG standardly wraps `char` as a single character string (`signed char` and `unsigned char` are wrapped as integer types).  The user can choose to handle `char` as an integer type via `%apply` though (which is the sort of thing hard-coding the code to get/set for these types breaks).

ojwb:
The tests are failing for PHP - looks like bad generated code.

(Perhaps it'd be better to focus on one PR at a time rather than opening lots of draft PRs in parallel?)

Source/Modules/php.cxx
```diff
@@ -1669,6 +1669,46 @@ class PHP : public Language {
...
+      if (is_setter_method(n)) {
+        Replaceall(mcode, "arg2", "swig_val");
+        /* Save local var declarations (skip arg1 and args) */
+        String *mlocals = NewStringEmpty();
+        {
+          Iterator ki = First(f->localh);
+          while (ki.key) {
+            const char *k = Char(ki.key);
+            if (strcmp(k, "arg1") != 0 && strcmp(k, "args") != 0) {
+              Replace(ki.item, "arg2", "swig_val", DOH_REPLACE_ANY);
```
ojwb, line 1683: If you're really need to do these sort of replacements, I think you want `DOH_REPLACE_ID` - otherwise this risks mangling any occurrence of the substring `arg2` in a typemap.

If possible though, it be better to avoid needing to refactor code like this.  It's really hard to do this sort of thing without risk of breaking more exotic cases.

It'd also be helpful to add comments explaining what's going on.  It's certainly not clear to me (for example, why are we fiddling with `arg2` but not other `arg`_N_, and why we only do this for locals `arg1` and `args`?)

jschueller:
yes, it seems not ready

Issue #2337, opened 2022-08-10
Go module typemap support is incomplete


wsfulton:
The Go module is not fully typemap friendly. I tried to implement std::unique_ptr support after implementing the support in the other target language modules. Given the nature of Go and that it does not use the runtime, I feel the most similar implementation should be something like the Java or C# module as they are both strongly typed. Where the Java and C# have typemaps to control nearly every aspect of the code generation, Go has a lot of hard coded generation instead of typemaps. It is thus quite inflexible w.r.t. the types used in the code generation. Typemaps and controlling the code generation is absolutely fundamental to how SWIG should be working.

The improvement in the Go module that is required is to generate (at least optionally perhaps to begin with) all the types in the generated code from typemaps. From what I can make out the 'imtype' typemap is similar to the C# 'imtype' typemap. The 'gotype' typemap is similar to the 'cstype' typemap. There is no equivalent to C#'s 'ctype' typemap. The following usage of the first two typemaps is unexpected though. Consider:

```swig
%module example

%typemap (imtype) Klass * "/*imtype typemap*/ Klass"
%typemap (gotype) Klass * "/*gotype typemap*/ Klass"

%inline %{
struct Klass {};
void takeKlass(Klass *pk) {}
Klass * makeKlass() { return new Klass(); }
%}
```

This should generate the same code as if the two typemaps were not defined. But there are a number of changes in the example.go where `uintptr_t` is changed to `swig_voidp`, such as:

```go
extern uintptr_t _wrap_new_Klass_example_d7d298a69a87cd7f(void);
```
to
```go
extern swig_voidp _wrap_new_Klass_example_297296dc9f379627(void);
```

These types in the commented out section are not configurable via typemaps either from what I can make out and would need a new type typemap.

Then, looking at the generated wrapper for makeKlass in the example_wrap.cxx file:
```c++
Klass *_wrap_makeKlass_example_297296dc9f379627() {
  Klass *result = 0 ;
  Klass *_swig_go_result;
  
  
  result = (Klass *)makeKlass();
  *(Klass **)&_swig_go_result = (Klass *)result; 
  return _swig_go_result;
}
```
it should be possible to control the return type `Klass *`, which in C# is done via the ctype typemap.

I expect for classes and pointers, it would be

C#
```swig
%typemap(ctype) SWIGTYPE "void *"
%typemap(ctype) SWIGTYPE * "void *"
%typemap(ctype) SWIGTYPE & "void *"
```
Proposed for Go
```swig
[...]

PR #3445, opened 2026-05-20
Go: Fix wrapping of static constexpr const char* member variables

### LINKED ISSUE #2724: SWIG/Go generates bad code for constexpr const char* member variable
Cut-down reproducer - replace `Examples/test-suite/cpp11_decltype.i` with:

```
%module cpp11_decltype
%inline %{
struct B {
  static constexpr const char * should_be_string = "xyzzy";
};
%}
```

Then

```
$ make -s -C Examples/test-suite/go cpp11_decltype.cpptest
checking go testcase cpp11_decltype
# swigtests/cpp11_decltype
cpp11_decltype_wrap.cxx: In function ‘_gostring_ _wrap_B_should_be_string_B_cpp11_decltype_36ff4f22d88931d7()’:
cpp11_decltype_wrap.cxx:294:15: error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
  294 |   result = B::should_be_string;
      |            ~~~^~~~~~~~~~~~~~~~
      |               |
      |               const char*
make[1]: *** [../../../Examples/Makefile:444: go_cpp] Error 1
make: *** [Makefile:40: cpp11_decltype.cpptest] Error 2
```

Bad code in `Examples/test-suite/go/cpp11_decltype_wrap.cxx` is:

```
_gostring_ _wrap_B_should_be_string_B_cpp11_decltype_36ff4f22d88931d7() {
  char *result = 0 ;
  _gostring_ _swig_go_result;
  
  
  result = B::should_be_string;
  
  _swig_go_result = Swig_AllocateString((char*)result, result ? strlen((char*)result) : 0); 
  return _swig_go_result;
}
```

Should be `const char *result = 0 ;` instead.  I had a quick look but couldn't see where this was coming from.

jschueller:
Swig_VargetToFunction (Source/Swig/cwrap.c) now merges the declarator into the variable's type, so that non-constant variables get the full type including pointer/declarator.  In staticmembervariableHandler (Source/Modules/lang.cxx), qualifiers are no longer stripped from the type when a constexpr variable is wrapped as a constant, preserving the const qualifier for downstream type selection.

Previously a static constexpr const char* member generated C wrapper code using the stripped type "char", which also affected Go wrapper code generation.  The fix ensures the const qualifier is preserved.

Closes #2724

### DIFF
--- a/Examples/test-suite/cpp11_decltype.i
+++ b/Examples/test-suite/cpp11_decltype.i
@@ -73,6 +73,9 @@
 
     static constexpr decltype(&hidden_global_char) should_be_string = "xyzzy";
 
+    // Test wrapping of static constexpr const char * member variable (#2724)
+    static constexpr const char * plain_string = "hello";
+
     // SWIG < 4.2.0 incorrectly used int for the result of logical not in C++
     // so this would end up wrapped as int.
     decltype(!0) should_be_bool;
--- a/Examples/test-suite/go/cpp11_decltype_runme.go
+++ b/Examples/test-suite/go/cpp11_decltype_runme.go
@@ -1,0 +1,13 @@
+package main
+
+import "swigtests/cpp11_decltype"
+
+func main() {
+    if cpp11_decltype.BShould_be_string != "xyzzy" {
+        panic("BShould_be_string mismatch")
+    }
+
+    if cpp11_decltype.BPlain_string != "hello" {
+        panic("BPlain_string mismatch")
+    }
+}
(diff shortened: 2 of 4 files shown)


jschueller:
interferes with java apparently

wsfulton:
Changing the core is almost always the wrong thing when fixing just one language. In this case the Go typemaps are fundamentally broken and should be the focus of fixes, see #2337. I don't think there is much point in attempting Go specific fixes around types until this is sorted.

jschueller:
fair enough

PR #3446, opened 2026-05-21
Parser: emit char-valued enum items as integers, not strings

### LINKED ISSUE #1026: TypeError in method with enum argument
test_enum.i:
`enum {
  TEST_ENUM = 'A'
 };`

`$swig -python test.i
 test.i:3: Error: Type error. Expecting an int `

Same thing happens for php and I assume all others...

This is with the release version of 3.0.12. It worked fine in 2.0.1.

jschueller:
When an enum item is initialized with a character literal (e.g. `GREEN = 'g'`), SWIG incorrectly propagated the initializer's type (T_CHAR) to the enum item's type attribute. This caused Python to call `SWIG_From_char()`, which returns a Python string via `SWIG_FromCharPtrAndSize()`, producing `GREEN = 'g'` instead of `GREEN = 103`. Passing such a value to a function expecting the enum type then fails with TypeError.

Changes in Source/CParse/parser.y:

1. edecl rule -- always use T_INT for enum items regardless of the initializer expression type. The underlying type of a C/C++ enumeration is an integral type, not the type of the initializer.
2. CHARCONST production -- compute `numval` (the integer value of the character) so that `enumnumval` is available to backends that use it (Java, C#, R, D).
3. WCHARCONST production -- same treatment for wide character literals.
4. etype validation -- add T_WCHAR to the allowed integral types list so that `L'x'` literals are accepted as enum initializers.

Closes #1026

Assisted-by: Claude Code

### DIFF
--- a/Examples/test-suite/char_enum_value.i
+++ b/Examples/test-suite/char_enum_value.i
@@ -1,0 +1,12 @@
+%module char_enum_value
+
+%inline %{
+
+enum color { RED, BLUE, GREEN = 'g' };
+enum escapes {
+  NEWLINE = '\n',
+  TAB = '\t',
+  CAP_A = 'A',
+};
+
+%}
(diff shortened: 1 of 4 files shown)


jschueller:
enum is tricky

PR #3448, opened 2026-05-22
ccache: Try dropping -c argument with -E

jschueller:


erezgeva:
Excellent you try to fix :+1: 

I wonder why fixing `CCache/ccache.c‎`?
Perhaps it is better to fix `CCache/test.sh`?

Are you sure the problem is not limited to the test?

I know @wsfulton is very sensitive to changes in `ccache.c`, as we plan to merge back with `ccache` project in the future.
I am not sure what feedback, but I would suggest checking other alternative, before deciding on this one.

jschueller:
I dont think I understand this enough

wsfulton:
Yes, I wouldn't normally attempt to fix ccache, but in this case it breaks the test-suite so needs addressing asap, even though it is a compiler cache problem (not a swig related problem). The logic in @jschueller's patch here is okay, and after cleaning it up, it is committed in b6b83deb80e92be1ff7907905a9d9b2822060067. Seems low risk to me given it doesn't make much sense providing -c with -E to me and the clang compiler is saying not to combine the two arguments.

GHA CI is now green again (failing test was macos 'none clang'.

PR #3452, opened 2026-05-25
Python: Allow keyword arguments with overloaded functions

### LINKED ISSUE #2703: Python -keyword option
The documentation about -keyword is very limited:
https://www.swig.org/Doc4.1/SWIGDocumentation.html#Python_commandline

Digging deeper I realized keyword arguments are actually supported, whether or not I pass -keyword option:
```
#include <iostream>
namespace OT {
  class Mesh
  {
  public:
    Mesh(const int nVert = 0) : nVertices_(nVert) {}
    int getNVertices() const {return nVertices_;}
    void methodWithKW(int a = 0, int b = 4) {std::cout << "a="<<a<<" b="<<b<<std::endl;}
    
  private:
    int nVertices_;
  };
}

mesh = ot.Mesh()
mesh.methodWithKW(1,2)
mesh.methodWithKW(1)
mesh.methodWithKW(a=77, b=30)
mesh.methodWithKW(b=5, a=3)

a=1 b=2
a=1 b=4
a=77 b=30
a=3 b=5
```

so is this option still a thing ?

also extranative, dirvtable, castmode, nofastunpack are not self-explanatory and not documented either

jschueller:
When -keyword or %feature("kwargs") was used, keyword arguments were silently disabled for all overloaded functions and constructors. The dispatch wrapper used PyArg_ParseTuple instead of
PyArg_ParseTupleAndKeywords and the individual overload wrappers dropped kwargs support when overname was set.

Fix by removing the overname check as a condition for disabling kwargs in the individual overload wrappers and by adding a kwargs dispatch loop to the dispatch function that tries each overload sequentially, relying on PyArg_ParseTupleAndKeywords to reject mismatched keywords with TypeError (cleared between attempts). The Python proxy __init__ and method helpers now pass **kwargs through to the dispatch function.

The overloaded kwargs path is always active when -keyword / %feature("kwargs") is used. If performance with many overloads becomes a concern, a separate %feature("kwargs:overloaded") could be added later as an opt-in, but the runtime cost only applies when kwargs are actually passed (a PyDict_Size check); without kwargs, the existing argc-based fast path is unchanged.

Closes #2703

### DIFF
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -380,6 +380,7 @@ CPP_TEST_CASES += \
 	overload_complicated \
 	overload_copy \
 	overload_extend \
+	overload_kwargs \
 	overload_method \
 	overload_numeric \
 	overload_null \
--- a/Examples/test-suite/overload_kwargs.i
+++ b/Examples/test-suite/overload_kwargs.i
@@ -1,0 +1,18 @@
+%module overload_kwargs
+
+%nocopyctor;
+%feature("kwargs");
+
+%inline %{
+  struct MyStruct {
+    MyStruct() : a_(0.0), b_(0.0) {}
+    MyStruct(double a, double b) : a_(a), b_(b) {}
+    double foo() { return 0.0; }
+    double foo(double x, double y) { return x + y; }
+    double getA() { return a_; }
+    double getB() { return b_; }
+  private:
+    double a_;
+    double b_;
+  };
+%}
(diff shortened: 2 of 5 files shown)


wsfulton:
I've cancelled the GHA for now, please see #3454. I'll restart them at a later date. Sorry for inconvenience.

wsfulton:
Sort of related commit from today: 349addecbc97a05ba29c443dd72b002354274743.

I've restarted the CI jobs.

PR #3456, opened 2026-05-26
[Ruby] Migrate from Data_* to TypedData_* API

jschueller:
Ruby 3.4 marks the legacy Data_Wrap_Struct / Data_Get_Struct API as deprecated (via RUBY_UNTYPED_DATA_WARNING), generating compiler warnings that become errors under -Werror.

Replace all uses with the TypedData API:

- Add a swig_ruby_data wrapper struct that stores the wrapped pointer alongside per-object mark/free function pointers, since TypedData's static rb_data_type_t does not allow dynamic ownership transfers.
- Introduce a shared swig_ruby_data_type descriptor with generic mark/free/size callbacks that delegate through the wrapper.
- Redefine DATA_PTR to transparently unwrap through the wrapper so existing generated code and user typemaps using DATA_PTR continue to work unchanged.

### DIFF
--- a/Lib/ruby/rubyhead.swg
+++ b/Lib/ruby/rubyhead.swg
@@ -135,8 +135,8 @@
  *
  * VOIDFUNC(f) is used to typecast a C function that implements either
  * the "mark" or "free" stuff for a Ruby Data object, so that it can be
- * passed as an argument to Ruby C API functions like Data_Wrap_Struct()
- * and Data_Make_Struct().
+ * passed as an argument to Ruby C API functions like TypedData_Wrap_Struct()
+ * and TypedData_Make_Struct().
  *
  * SWIG_RUBY_VOID_ANYARGS_FUNC(f) is used for the function pointer
  * argument(s) of Ruby C API functions like rb_define_virtual_variable().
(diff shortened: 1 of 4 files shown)


Lib/ruby/rubyhead.swg
```diff
@@ -183,4 +183,45 @@
...
+ * handlers stored in the swig_ruby_data wrapper. */
+static void swig_ruby_data_mark(void *ptr) {
+  swig_ruby_data *d = (swig_ruby_data *)ptr;
+  if (d->mark && d->mark != (RUBY_DATA_FUNC)RUBY_DEFAULT_FREE)
+    d->mark(d->ptr);
+}
+
+static void swig_ruby_data_free(void *ptr) {
+  swig_ruby_data *d = (swig_ruby_data *)ptr;
+  if (d->free && d->free != (RUBY_DATA_FUNC)RUBY_DEFAULT_FREE)
```
jhawthorn, line 206: I'm not sure that's correct. If the function is `RUBY_DEFAULT_FREE` we should free the pointer in the default way (NB. that's `xfree`, not `free`). Doing otherwise is a memory leak.

That said I don't know if that ever happens, or is `free` here always known to be either 0 or a proper function.

jhawthorn:
Did you see https://github.com/swig/swig/pull/3326 ?

jschueller:
ah yes, thats totally a duplicate

PR #3458, opened 2026-05-27
chore: add generated autoconf configure script

steveratti:


ojwb:
We deliberately don't put the generated autotools files in git - it would result in conflicts and a lot of churn as developers won't always have the same autotools versions installed as each other.

Just run `./autogen.sh` after `git clone`.

PR #3460, opened 2026-05-30
tweak help messages

e-kwsm:


wsfulton:
Just wondering why there is no explanation with this pull request and why you have broken the lining up of the dashes that we had before? I expect you have a good reason, but it is out of line now.

e-kwsm:
Well, this is a trivial cosmetic PR, and I don&rsquo;t think explanation is necessary.

PR #3461, opened 2026-05-30
docs: create swig(1) man

### LINKED ISSUE #3459: feature request: man page
At the moment swig(1) is not provided.

e-kwsm:
swig.1 is created by
```
help2man --no-info --help-option=-help --version-option=-version swig
```
as of v4.4.0-255-gc94c19d29, and then manually modified

fix #3459

### DIFF
--- a/Doc/swig.1
+++ b/Doc/swig.1
@@ -1,0 +1,608 @@
+.Dd May 2026
+.Dt SWIG 1
+.Os
+.Sh NAME
+.Nm swig
+.Nd Simplified Wrapper and Interface Generator
+.Sh SYNOPSIS
+.Nm
+.Op Ar OPTIONS
+.Ar file
+.Sh DESCRIPTION
+SWIG is a software development tool that connects programs written in C and C++
+with a variety of high\-level programming languages.
+.Sh OPTIONS
+.Ss Supported Target Language Options
+.Bl -tag -width 6n
+.It Fl csharp
+Generate C# wrappers
+.It Fl d
+Generate D wrappers
+.It Fl go
+Generate Go wrappers
+.It Fl guile
+Generate Guile wrappers
+.It Fl java
+Generate Java wrappers
+.It Fl javascript
+Generate Javascript wrappers
+.It Fl lua
+Generate Lua wrappers
+.It Fl octave
+Generate Octave wrappers
+.It Fl perl5
+Generate Perl 5 wrappers
+.It Fl php7
+Generate PHP 8 or later wrappers
+.It Fl python
+Generate Python wrappers
+.It Fl r
+Generate R (aka GNU S) wrappers
+.It Fl ruby
+Generate Ruby wrappers
+.It Fl scilab
+Generate Scilab wrappers
+.It Fl tcl8
+Generate Tcl 8 wrappers
+.It Fl xml
+Generate XML wrappers
+.El
+.Pp
+Note:
+.Ql Nm Fl Ar lang Fl help
+displays options for a specific target language.
+.Ss General Options
+.Bl -tag -width 6n
+.It Fl addextern
+Add extra extern declarations
+.It Fl c++
+Enable C++ processing
+.It Fl co Ar file
+Check
+.Ar file
+out of the SWIG library
+.It Fl copyctor
+Automatically generate copy constructors wherever possible
+.It Fl cpperraswarn
+Treat the preprocessor #error statement as #warning (default)
+.It Fl cppext Ar ext
+Change file extension of generated C++ files to
+.Ar ext
+(default is cxx)
+.It Fl copyright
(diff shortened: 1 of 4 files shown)


Doc/swig.1
```diff
@@ -0,0 +1,674 @@
...
+Generate wrappers into C# namespace \fInm\fP
+.TP
+\fB\-noproxy\fP
+Generate the low\-level functional interface instead of proxy classes
+.TP
+\fB\-oldvarnames\fP
+Old intermediary method names for variable wrappers
+.TP
+\fB\-outfile\fP \fIfile\fP
+Write all C# into a single \fIfile\fP located in the output directory
```
e-kwsm: I put all the language-specific options into a single swig.1 rather than separate files like swig-csharp.1, since they will be almost the same.

Doc/swig.1
```diff
@@ -0,0 +1,674 @@
...
+.TP
+\fB\-scilab\fP
+Generate Scilab wrappers
+.TP
+\fB\-tcl8\fP
+Generate Tcl 8 wrappers
+.TP
+\fB\-xml\fP
+Generate XML wrappers
+.SS "General Options"
```
e-kwsm: Experimental features i.e. `-c` and `-ocaml` are omitted.

wsfulton:
Hi @e-kwsm , this needs to not break installations in CI nor anyone's machines in order to be considered, but it does at the moment. Any extra dependencies need to be clearly documented and detected during ./configure time.

e-kwsm:
Hellow @wsfulton, this does not introduce dependency on help2man.

https://github.com/swig/swig/actions/runs/28305929163/job/83861976437?pr=3461#step:9:890

> Installing SWIG man
> cp: /usr/local/share/man/man1/\_inst.41011\_: Permission denied
> make: *** [install-docs] Error 1

I guess this comes from the following snippet of install-sh

```sh
    dsttmp=${dstdirslash}_inst.$$_
    rmtmp=${dstdirslash}_rm.$$_

    # Trap to clean up those temp files at exit.
    trap 'ret=$?; rm -f "$dsttmp" "$rmtmp" && exit $ret' 0

    # Copy the file name to the temp name.
    (umask $cp_umask &&
     { test -z "$stripcmd" || {
         # Create $dsttmp read-write so that cp doesn't create it read-only,
         # which would cause strip to fail.
         if test -z "$doit"; then
           : >"$dsttmp" # No need to fork-exec 'touch'.
         else
           $doit touch "$dsttmp"
         fi
       }
     } &&
     $doit_exec $cpprog "$src" "$dsttmp") &&
```

and tmpfile `/usr/local/share/man/man1/_inst.$$_` or its parental directories are not writable, but I&rsquo;m not sure why this is problematic on macos.

e-kwsm:
`sudo` is necessary because the user in a macos runner is not root.

wsfulton:
> `sudo` is necessary because the user in a macos runner is not root.

If this is the case, how is it that the binaries can be installed without root? Maybe the man page install location is not honouring the correct destination directory. Could you please diagnose by showing the permissions of the destination directory - they really ought to be the same for /usr/local/bin and /usr/local/man. If they aren't then this might be a problem for Github to fix up in the images. But, whatever it is, I'd like an understanding of where the permissions are wrong as changing the installation behaviour of SWIG is now looking like a risky change.

One minor thing, can you please change the make targets to use `manual` instead of `doc` as we think of docs as the html documentation.

One last important point, I'd like to see the man page generated from the output of `swig -help` via a script instead of committing it as `Doc/swig.1`. Otherwise it'll easily get out of date, which is debatably worse than having no man page.

e-kwsm:
> > `sudo` is necessary because the user in a macos runner is not root.
> 
> If this is the case, how is it that the binaries can be installed without root? Maybe the man page install location is not honouring the correct destination directory. Could you please diagnose by showing the permissions of the destination directory - they really ought to be the same for /usr/local/bin and /usr/local/man. If they aren't then this might be a problem for Github to fix up in the images. But, whatever it is, I'd like an understanding of where the permissions are wrong as changing the installation behaviour of SWIG is now looking like a risky change.

In a macos runner, the user—`runner`—is not root; `/usr/local/{bin,share}` are owned by the user and have permission of 775, whereas `/usr/local/share/man{,/man1}` is owned by root and not writable (755).

> One minor thing, can you please change the make targets to use `manual` instead of `doc` as we think of docs as the html documentation.

Fixed.
 
> One last important point, I'd like to see the man page generated from the output of `swig -help` via a script instead of committing it as `Doc/swig.1`. Otherwise it'll easily get out of date, which is debatably worse than having no man page.

Unfortunately help2man is not perfect:

- generated man is a bit malformed due to `swig -help` output (#3460)
- help2man does not take care of &lt;placeholders&gt;

PR #3467, opened 2026-06-10
Ruby: Fix objects embedded data usage

larskanis:
Sorry, I made a mistake in the last merge request!

Two things disabled objects embedded data usage:

1. RUBY_TYPED_EMBEDDABLE is no `define`, but only an `enum` in ruby-3.3 to 4.0. It will be a define in ruby-4.1, but to enable the feature we better check for TYPED_DATA_EMBEDDED which is defined in all rubies 3.3 and following. The ruby docs are a bit misleading about this: https://github.com/ruby/ruby/blob/master/doc/extension.rdoc#c-struct-to-ruby-object

2. RUBY_TYPED_EMBEDDABLE should be OR'ed bitwise to have an effect.

### DIFF
--- a/Examples/test-suite/ruby/Makefile.in
+++ b/Examples/test-suite/ruby/Makefile.in
@@ -37,6 +37,7 @@ CPP_TEST_CASES = \
 	ruby_minherit_shared_ptr \
 	ruby_naming \
 	ruby_rdata \
+	ruby_typeddata_embedded \
 	ruby_track_objects \
 	ruby_track_objects_directors \
 	std_containers \
--- a/Examples/test-suite/ruby/ruby_typeddata_embedded_runme.rb
+++ b/Examples/test-suite/ruby/ruby_typeddata_embedded_runme.rb
@@ -1,0 +1,19 @@
+#!/usr/bin/env ruby
+#
+# Verify that wrapped objects use Ruby's embedded TypedData slot on Ruby 3.3+.
+#
+
+require 'swig_assert'
+require 'ruby_typeddata_embedded'
+
+include Ruby_typeddata_embedded
+
+thing = Thing.new(42)
+swig_assert_equal(42, thing.value, binding)
+
+# From Ruby 3.3 the per-class descriptor enables RUBY_TYPED_EMBEDDABLE, so the
+# wrapper is embedded in the object slot. Without it the wrapper is malloc'd and
+# leaked, since the free callback no longer releases it.
+if RUBY_VERSION >= "3.3"
+  swig_assert(swig_wrapper_embedded(thing), binding, "wrapper should be embedded on Ruby >= 3.3")
+end
--- a/Examples/test-suite/ruby_typeddata_embedded.i
+++ b/Examples/test-suite/ruby_typeddata_embedded.i
@@ -1,0 +1,28 @@
+%module ruby_typeddata_embedded
+
+/* Regression test for the per-class TypedData wrapper using Ruby's embedded data
+   slot (RUBY_TYPED_EMBEDDABLE) on Ruby 3.3 and later. If the per-class cext_type
+   descriptor does not request embedding, the swig_ruby_wrapped_object is malloc'd
+   and, because the free callback skips ruby_xfree when embedding is compiled in,
+   leaked on every object. The runme checks that a wrapped object really is
+   embedded so the leak can not silently come back. */
+
+%inline %{
+class Thing {
+public:
+  int value;
+  Thing(int v) : value(v) {}
+};
+
+/* True when obj's TypedData wrapper is stored inside the Ruby object slot rather
+   than a separate allocation. Always false on Ruby older than 3.3, which has no
+   embedding support. */
+bool swig_wrapper_embedded(VALUE obj) {
+#ifdef TYPED_DATA_EMBEDDED
+  return RTYPEDDATA_EMBEDDED_P(obj) ? true : false;
+#else
+  (void)obj;
+  return false;
+#endif
+}
+%}
--- a/Lib/ruby/rubyhead.swg
+++ b/Lib/ruby/rubyhead.swg
@@ -118,18 +118,18 @@
 
 /* Flags shared by every rb_data_type_t that SWIG allocates with
    TypedData_Make_Struct, so the static descriptors and the per class cext_type
-   stay in step. RUBY_TYPED_FREE_IMMEDIATELY is a (self referential) macro back to
-   Ruby 2.1, hence the '#ifdef'. Ruby 3.3+ additionally defines TYPED_DATA_EMBEDDED
-   and can store the small, fixed size swig_ruby_wrapped_object inside the object
-   slot rather than a separate malloc; that embedding also needs FREE_IMMEDIATELY. */
+   stay in step. The rb_data_type_t flags member, and RUBY_TYPED_FREE_IMMEDIATELY,
+   only exist from Ruby 2.1, so this macro (and every use of it) is guarded by
+   '#ifdef RUBY_TYPED_FREE_IMMEDIATELY' and is simply absent on older Ruby. Ruby
+   3.3+ additionally defines TYPED_DATA_EMBEDDED and can store the small, fixed
+   size swig_ruby_wrapped_object inside the object slot rather than a separate
+   malloc; that embedding also needs FREE_IMMEDIATELY. */
 #ifdef RUBY_TYPED_FREE_IMMEDIATELY
 # ifdef TYPED_DATA_EMBEDDED
 #  define SWIG_RUBY_TYPED_DATA_FLAGS (RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_EMBEDDABLE)
 # else
 #  define SWIG_RUBY_TYPED_DATA_FLAGS (RUBY_TYPED_FREE_IMMEDIATELY)
 # endif
-#else
-# define SWIG_RUBY_TYPED_DATA_FLAGS 0
 #endif
 
 /* Wrapper attached to every SWIG TypedData object. It holds the wrapped pointer
--- a/Lib/ruby/rubyhead.swg
+++ b/Lib/ruby/rubyhead.swg
@@ -116,6 +116,22 @@
 # define RTYPEDDATA_GET_DATA(x) RTYPEDDATA_DATA(x)
 #endif
 
+/* Flags shared by every rb_data_type_t that SWIG allocates with
+   TypedData_Make_Struct, so the static descriptors and the per class cext_type
+   stay in step. RUBY_TYPED_FREE_IMMEDIATELY is a (self referential) macro back to
+   Ruby 2.1, hence the '#ifdef'. Ruby 3.3+ additionally defines TYPED_DATA_EMBEDDED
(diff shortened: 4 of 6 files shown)


wsfulton:
## Review

The diagnosis here is correct and valuable — `RUBY_TYPED_EMBEDDABLE` really is invisible to the preprocessor, and the `||` vs `|` bug is real. But the fix as written is **incomplete**: on Ruby 3.3+ it suppresses the `wrobj` free without actually enabling embedding, which turns a correct no-op into a memory leak.

### Macro validation, Ruby 3.3 → 4.0

Checked `include/ruby/internal/core/rtypeddata.h` at each tag:

| Macro | 3.3.0 | 3.4.0 | master | Preprocessor-visible? |
|---|---|---|---|---|
| `RUBY_TYPED_EMBEDDABLE` | enum `= 2` | enum `= 2` | enum `= 2` | **No** (enum only) |
| `RUBY_TYPED_FREE_IMMEDIATELY` | enum `= 1` | enum `= 1` | enum `= 1` | **No** (enum only) |
| `TYPED_DATA_EMBEDDED` | `#define …2` | `#define …2` | `#define ((VALUE)1)` | **Yes** |

All three live in `enum rbimpl_typeddata_flags` (an enum since the Ruby 3.0 header refactor), so only `TYPED_DATA_EMBEDDED` can be tested with `#ifdef`. Its *value* differs across versions (2 vs 1), but that's fine because SWIG only tests its existence.

Links:
- https://github.com/ruby/ruby/blob/v3_3_0/include/ruby/internal/core/rtypeddata.h
- https://github.com/ruby/ruby/blob/v3_4_0/include/ruby/internal/core/rtypeddata.h
- https://github.com/ruby/ruby/blob/master/include/ruby/internal/core/rtypeddata.h
- Embedding contract: https://github.com/ruby/ruby/blob/master/doc/extension.rdoc — "set `RUBY_TYPED_EMBEDDABLE` in the flags", "only `RTYPEDDATA_GET_DATA` can be used", "`dfree` MUST NOT free the C struct itself", "the type must have the `RUBY_TYPED_FREE_IMMEDIATELY` flag set", and "Pointers to the C struct … MUST NOT be stored, as they become invalid when GC compaction occurs".

### Critical: leaks on Ruby 3.3+, and embedding is never enabled

The two halves of the patch use inconsistent guards.

**Free guard** (`rubyrun.swg:188`):
```c
#ifndef TYPED_DATA_EMBEDDED
  ruby_xfree(wrobj);
#endif
```
`TYPED_DATA_EMBEDDED` is defined on all Ruby >= 3.3, so `ruby_xfree(wrobj)` is compiled out on every supported Ruby.

**Descriptor flags** (`rubyrun.swg:200`):
```c
#ifdef RUBY_TYPED_FREE_IMMEDIATELY      // <-- still the old, dead guard
  , RUBY_TYPED_FREE_IMMEDIATELY
#ifdef TYPED_DATA_EMBEDDED
  | RUBY_TYPED_EMBEDDABLE
#endif
#endif
```
The **outer** guard is still `#ifdef RUBY_TYPED_FREE_IMMEDIATELY`, which (per the table) is an enum → false on 3.3+. So the whole flags initializer, including `| RUBY_TYPED_EMBEDDABLE`, is dropped and `swig_type_data_type.flags == 0`. No embedding is requested.

Net effect on 3.3+: `TypedData_Make_Struct` heap-allocates `wrobj` (flags 0, no embedding), but `SWIG_Ruby_free_swig_type` no longer frees it → `struct swig_ruby_wrapped_object` leaks on every freed object. Before this PR the free was guarded by `#ifndef RUBY_TYPED_EMBEDDABLE` (always true → always freed → correct), so this regresses "no embedding, no leak" into "no embedding, leak".

**The dominant path isn't even touched.** Most wrapped objects go through the per-class `cext_type` (`rubyrun.swg:246`), not `swig_type_data_type`. `swig_class` is a zero-initialized static (`ruby.cxx:2498`) and `handleClassName` (`ruby.cxx:2433-2436`) sets only `wrap_struct_name`, `dmark`, `dfree` — `cext_type.flags` is never set, so it's `0`. Those objects share the same free function, so they leak too after this PR, again with no embedding benefit.

### What a correct fix needs

1. Gate the flags on `TYPED_DATA_EMBEDDED`, not the dead `#ifdef RUBY_TYPED_FREE_IMMEDIATELY`. Embedding also *requires* `RUBY_TYPED_FREE_IMMEDIATELY`, so set both bits:
   ```c
   0, 0
   #ifdef TYPED_DATA_EMBEDDED
     , RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_EMBEDDABLE
   #endif
   ```
   (To keep `FREE_IMMEDIATELY` on older Rubies where it isn't `#define`d, you can't detect it via `#ifdef` — use `RUBY_API_VERSION_CODE`.)
2. **Set `cext_type.flags` for the class path** in `handleClassName` (`ruby.cxx`), e.g. emit `SwigClass%s.cext_type.flags = RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_EMBEDDABLE;` under a `TYPED_DATA_EMBEDDED` guard. Without this the common case never embeds and leaks.
3. Keep the free guard (`#ifndef TYPED_DATA_EMBEDDED`) consistent with whichever descriptors actually request embedding. Since `cext_type` and `swig_type_data_type` share one free function, both must embed or neither — otherwise the "free iff not embedded" invariant breaks again.

### Forward-looking caveat

Once embedding really turns on, the docs warn that pointers into the embedded struct move on GC compaction. Any code that caches a `wrobj*` across calls would become unsafe and must re-fetch via `RTYPEDDATA_GET_DATA`. Worth auditing `SWIG_Ruby_ConvertPtr` and the tracking code before flipping it on. (`rubyhead.swg:115-117` already maps `RTYPEDDATA_GET_DATA`→`RTYPEDDATA_DATA` only when `TYPED_DATA_EMBEDDED` is undefined, which is correct.)

### Minor

- `swig_runtime_data_type` (`rubyrun.swg:458`) and `swig_gcrefs_marker_type` (`rubyclasses.swg:84`) keep the same dead `#ifdef RUBY_TYPED_FREE_IMMEDIATELY` guard. They use `TypedData_Wrap_Struct` with externally-owned data / `dfree==0`, so they neither embed nor leak — safe to leave, but they're latent instances of the same pattern.
- No test accompanies the change; a leak like this won't fail functional tests. Consider asserting `RTYPEDDATA_EMBEDDED_P(obj)` is true in a runme on 3.3+ to prove embedding is actually engaged.

wsfulton:
That's obviously an AI review in great detail, so please check it all. Thanks.

wsfulton:
Claude is coming up with a patch and testing it atm. There are quite a few lines of change. I'll post it here for you to review.

larskanis:
The analysis is complete garbage. The AI doesn't understand that `RUBY_TYPED_FREE_IMMEDIATELY` is a macro AND an enum [here](https://github.com/ruby/ruby/blob/e83e058e8c724f9f70af1f2b2046d290e32ae226/include/ruby/internal/core/rtypeddata.h#L122). The Ruby API would be broken, if `RUBY_TYPED_FREE_IMMEDIATELY` isn't defined but `TYPED_DATA_EMBEDDED` is. 

The `RUBY_TYPED_FREE_IMMEDIATELY` check is only there to support ruby < 2.1 (releases before 2013).

Only `RUBY_TYPED_EMBEDDABLE` wasn't a macro from ruby-3.3 to 4.0, but is on master branch. This was because the feature wasn't acknowledged as public, but it is now since https://github.com/ruby/ruby/pull/16455 without any further changes. So we can use it starting with ruby-3.3 if we check for `TYPED_DATA_EMBEDDED` as proposed in this PR.

I manually checked the usages of `RTYPEDDATA_GET_DATA` for issues regarding garbage collection, but the pointer is only short living in all cases and the ruby object is GC-locked all the times. Given that there are a lot of test cases in swig, I'm pretty sure, that there's no garbage collector issue coming from `RUBY_TYPED_EMBEDDABLE`.

It makes sense to add a test case for verification per `RTYPEDDATA_EMBEDDED_P`. Do you have something already present?

wsfulton:
You'll have to bear with me as we have no Ruby maintainer and I have no other way to review these pull requests being unfamiliar with the Ruby API. I completely missed the self-referential macro and havn't come across that before, so that is interesting and quite confusing tbh.

There seems to be a real leak left over - but only on the per-class path, which the PR doesn't touch. Wrapped class objects don't use swig_type_data_type; they use the per-class SwigClass*.cext_type descriptor, whose flags are left zero. So those objects are never embedded, yet the free callback now skips ruby_xfree once TYPED_DATA_EMBEDDED is defined - leaking the swig_ruby_wrapped_object on every object. I measured ~33 bytes/object on Ruby 3.3, growing every GC round and never plateauing.

I've pushed three commits to demonstrate and fix it:

1. Avoid the undefining-allocator warning for swig_runtime_data - unrelated, but a pre-existing warning from your 4748a6c31a63c1c6c321deb85c59ce6905967452, printed by every module since the TypedData migration (rb_undef_alloc_func was called after TypedData_Wrap_Struct; undefining the allocator before the class has instances avoids it). Done first so the demo output below is clean. Please check this.
2. Add a test (ruby_typeddata_embedded) - to your point about adding a check via RTYPEDDATA_EMBEDDED_P, this is exactly that: it asserts a wrapped class instance is embedded on Ruby ≥ 3.3. On its own this commit fails, because the per-class descriptor isn't embedding.
3. Embed the per-class descriptor too - introduces a shared SWIG_RUBY_TYPED_DATA_FLAGS macro used by both swig_type_data_type and the generated cext_type, so the two can't drift apart, and sets cext_type.flags from it. The test now passes and the per-round RSS memory growth drops to zero.

To see it, on Ruby ≥ 3.3:
```bash
cd Examples/test-suite/ruby
make ruby_typeddata_embedded.cpptest   # fails at commit 2, passes at commit 3
```
I verified the sequence on Ruby 3.3.11 and 3.4.9 (embedded engages, value integrity holds across repeated GC.compact) and on 3.2 (non-embedded path unchanged). The shared macro keeps your swig_type_data_type logic intact - it just   factors it out so the class path gets the same flags. Probably best to squash the 2nd and 3rd commits into one when merging into master.

wsfulton:
I just pushed a fixup for the last commit to work on ruby-2.0.

Sigh, now [ruby gcc ucrt64](https://github.com/swig/swig/actions/runs/27448523969/job/81138802611#logs) job fails, but it is failing elsewhere too, probably an image update, I'll look at that next as that will be affecting master.

larskanis:
@wsfulton I didn't forget about this issue - just have been too busy.

larskanis:
The additional commits look very good! I found my branch again where I first tried the embedded flag and it looks very similar. But now things are more clean and tested.

If you decide to drop support for (really ancient) ruby-2.0, we could remove the `#ifdef RUBY_TYPED_FREE_IMMEDIATELY` conditions.

wsfulton:
I've squashed my 3 commits and have now merged everything in this branch with these commits: f8038e7cfc3f84f848c05181187a7f45ef0156e7...5fd1a0b1c53102298c859146b705d31c7fe553c8. ruby-2.0 was fixed so I've left it in.

larskanis:
Thank you very much! I'm looking forward to swig-4.5.

PR #3468, opened 2026-06-10
Move ccache to a separate repository based on original ccache

erezgeva:
I think this PR is the first step in the updating of SWIG-ccache

I think the best way is perform the update in a separate repository, which will allow us to do the migration in parallel to the swig development, and allow us to more easily to follow the original  `ccache` project and perform merging using the git tool.

I fork https://github.com/ccache/ccache using tag v2.4
And apply SWIG CCache differences.

This PR uses the repository in https://github.com/erezgeva/swig-ccache
As I can not add it in https://github.com/swig/
Yet the maintainers can do the fork here, and update the `repository` variable of the `Checkout CCache`.

The content of https://github.com/erezgeva/swig-ccache branch `swig_latest`, is identical to `master` `CCache` folder!

It is impossible to retain the original patches from the swig repository, due to many patches which are SWIG cross, which perform vast rebaseing and inverting of changes, which make the task nearly impossible.

v2.4 was publish in 2004-09-13, more than 21 years ago.
The last ccache version is v4.9 that published  2023-12-30

If we want to update we need some deeper changes:

- ccache moved to use cmake, we need to break the autoconf connection and generate the ccache separatly
- ccache support Windows using mingw, we can drop any "windows" code we add on top of  v2.4
- Remove the Debian folder, we do not use it. It is leftover when we import the code from Debian.
- ccache have its own GitHub tests. We can skip our `make check-ccache`
  while keep `make check-errors-test-suite` tests in `Examples/test-suite/errors/`, we can add a GHA in the ccache repository to test them in addition to the current tests in our repository.

I think we should concentrate on basic characters of the `swig-ccache` and use only `make check-errors-test-suite`, as verification tests!


erezgeva:
This PR contradict the plan for future ccache migration.
Just drop this one.

PR #3473, opened 2026-06-12
Python: Add option to generate .pyi stubs

Nerixyz:
From https://github.com/swig/swig/pull/3390#issuecomment-4664716922. To show annotations when using `-builtin`, this adds `-pyi-stub` to generate a `.pyi` file with stubs that contain the members with annotations.

I'm not sure how to test this. Python ignores the files.


wsfulton:
Thanks, I've merged this in now and taken it a bit further, see a0da96aff8dd118dec4a2b9cfaf101e7aa2acdf4...9302b31f57f34a5fd2e2ce96dd6802b6899def4b.

Regarding testing, I was thinking of using pyright or Pyrefly, but that will have to come later.

wsfulton:
I added a bit more about the testing I've added in #3528

PR #3474, opened 2026-06-13
Fix deleting invalid string when parsing Doxygen grouping comments

vadz:
This fixes a fatal regression, see https://github.com/swig/swig/issues/3403


vadz:
Is there anything preventing this PR from being merged/anything I can do to help with? This fixes a fatal problem so it looks like it ought to be merit some attention...

wsfulton:
Thanks @vadz, am squash merging it now (and added the CHANGES.current file entry).

wsfulton:
Fixed in 38e2824d487ec65afd497910da9995faefda20cc

PR #3475, opened 2026-06-13
Improvements to wrapping classes as interfaces in C#/Java/JS

vadz:
The main change here is to make any variables defined in the C++ base classes accessible via the interface in C#/Java when they were just completely lost before. For Java, any nested classes defined in these base classes are exported from the interface as well, but this is not done for C# because:

1. This is "only" available since C# 8 and so would require an option allowing to generate code requiring it.
2. Unlike Java, C# doesn't allow to refer to the nested classes defined in an interface using the derived class scope, which makes this much less useful, at least for my purposes.

Seeing that most of the recent changes in SWIG have been done by Claude, I've decided to use it for these changes too. FWIW I've reviewed all the changes in the first 3 commits but made only minor corrections to them, i.e. LLM-generated code looks fine to me.

### DIFF
--- a/Doc/Manual/Java.html
+++ b/Doc/Manual/Java.html
@@ -3649,9 +3649,17 @@ added to the interface; they are of course still available in the Java proxy cla
 </p>
 
 <p>
-<b>C# note:</b> the C# interface works the same way, except that member variables are exposed as C# properties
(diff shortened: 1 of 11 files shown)


vadz:
I've added a small change also exposing the base class member variables in JS for the classes wrapped as interfaces.

PR #3477, opened 2026-06-16
add kotlin support

TVDiasDominik:
We wanted to get completly rid of java in our code base and the swig generated java files where left.
With this swig can also generate kotlin code it should have parity with the java module


wsfulton:
HI @TVDiasDominik. My initial reaction is very nice. However, I'm trying to understand the point of this. I confess, I'm not particularly familiar with Kotlin, so I may need some technical explanations.

From what I can make out, the Java module will run perfectly well on the JVM from any Kotlin code. I need to have a really good explanation as to what is missing from the Java module in order to consider adding Kotlin. I can't accept a new module that is essentially identical when targetting the JVM. I am first and foremost concerned about maintainability. If there is nothing wrong with the Java module for Kotlin users, then I see no reason to accept this as it will soon lag behind the Java module as we have a shortage of developers willing to actively maintain SWIG modules. The sheer number of test-suite files which have had to be modified is an indicator of the extra maintenance overhead. Anyway, I hope you can provide some explanations. Thanks.

TVDiasDominik:
Hi @wsfulton it is true that kotlin can use java code.
First of kotlin is now the standard to use for android development and having still java code implies calling the java compiler just for the glue code with c++ which makes the build slower.
Currently the generated java is missing the nullsafty from kotlin which can lead to crashes if we pass in null to an actual nonenull parameter, having kotlin generated this can be prevented on compile time.
Biggest feature would be to allow to have nullsafty, kotlin Flows, kotlin coroutines, data classes usable which is currently not possible with java. Just one example what we want to do is to write a wrapper around the Kotlin Flow api to allow easy observable properties from c++.
Regarding the number of tests that where added most of them are just to make sure that the generated kotlin is in pair with what java has and works. The touched test suites are mostly just to add that they work also for kotlin i think that should just be an initial effort.

erezgeva:
The point William was trying to make is that the swig run of C/C++ code, the Java is using the result.
Which means you should be able to add Kotlin tests with the already existing Swig Java module.

If there are gaps, it is better to add them to the already existing Java module.
You can simple add the proper Kotlin tests for these missing feature.
You do not need to create a full language module for that.

After all, we may call it Java module, but it is actually C++/C to JVM.

So my suggestion, add Kotlin probing in configure.ac which works with `--with-java`.
So when we use Java, we also search for Kotlin.
Add tests in the test-suite, for Kotlin and call them from the Java `Examples/test-suite/java/Makefile.in`

Add the Kotlin tests for the features you miss and implement tham with existing Java module.
You may add the swig command line option in `Source/Modules/java.cxx` to enable this additional features, something like `-kotlin-ext` or any other proper name you may think of.

wsfulton:
Hi @TVDiasDominik,

I've been thinking about this and spun up some sub-agents to analyse in depth and 'discuss' your response. Please respond to the points below - I just don't have enough detailed knowledge of Kotlin to be absolutely sure there are not some subtle flaws in the logic.

## The maintenance problem is an important factor

The Kotlin module is essentially a fork of the Java module:

- `Source/Modules/kotlin.cxx` is 5,921 lines, of which only ~1,200 actually diverge from `java.cxx` — roughly **80% is copied-with-renames**, with no shared code path.
- In `Lib/kotlin/`, `director.swg` and `kotlinhead.swg` are **byte-for-byte copies** of the Java versions after renaming, and the STL wrappers are 75–90% copies.
- Every cross-language `%warnfilter` in the test-suite now has to carry a `SWIGWARN_KOTLIN_*` entry alongside the Java one, and there are ~234 hand-written `_runme.kt` duplicates of existing Java runners.

That means every future fix to the Java module has to be mirrored by hand into Kotlin, and the two will inevitably drift. Given SWIG's shortage of active maintainers, I don't think I can take on a second JVM backend that is ~80% redundant with the first — it's exactly the "lag behind the Java module" outcome I was worried about in my first reply.

## On the specific advantages

Looking at the benefits you listed, because if any of them genuinely required Kotlin generation, that would change the conclusion:

- **Null safety** — this is the one real gap, and it's the strongest point. But it doesn't require a Kotlin backend. Kotlin honours nullability annotations (`org.jetbrains.annotations`, JSpecify, JSR-305, androidx) from Java **bytecode at compile time** — annotated Java parameters/returns become genuine non-null / nullable Kotlin types rather than platform types. So if the **Java** module emitted `@NotNull`/`@Nullable` (derived from exactly the pointer-vs-reference information the Kotlin module already uses), your Kotlin callers would get the same compile-time null safety against the existing Java output.

- **Coroutines / Flow** — I note these aren't actually generated by the PR (I couldn't find any `suspend`/`Flow`/`kotlinx` output). And I don't think they should be: `suspend` is a compiler transform and `Flow` is a library type, both best expressed as thin **user-side Kotlin** over the generated synchronous API (`suspendCancellableCoroutine`, or `callbackFlow { … }` to bridge a native callback into a Flow — which sounds like exactly your observable-property use case). Those wrappers carry policy decisions (dispatcher, cancellation) that are better left in your code than baked into a generator.

- **data class** — `equals`/`hashCode`/`toString`/`copy` are all generatable from Java (or a Java `record`); `componentN` destructuring is minor user-side sugar. This is idiomatic polish rather than a capability gap.

- **Build speed / avoiding javac** — I'm not convinced generating Kotlin glue would be faster; `javac` is generally at least as fast as `kotlinc` for equivalent code, and your own test README notes that the 'Kotlin compiler startup overhead is significant'. A bounded `javac` pass over the glue code is a small, cacheable cost.

## Suggested ways forward

### Enhance generated Java code
The substantive gap is compile-time null safety, and I'd welcome a focused PR that adds this to the **existing Java module** — an annotations channel that emits `@Nullable`/`@NotNull` (JSpecify or JetBrains) on generated methods, parameters and returns. There's even a stubbed hook for this already (`java.cxx:2823-2834`, currently `#if 0`, with a TODO noting it should mirror the C# attributes mechanism), so it's a bounded change rather than a new backend. That gets Kotlin users real null safety today, on top of API output that already runs on the JVM, at a fraction of the maintenance cost. It will also be helpful for Java users, not just Kotlin users.

I'd rather find a way to give you what you need than say no outright — I just can't justify a second, largely-duplicated JVM module and the maintenance burden when you can have what you need via an enhanced Java module itself. If Kotlin-first ergonomics remain essential for you, I'm open to discussing what that would take, but the null safety via annotations route is the one I'd act on now.

### Generated Java to Kotlin converter
Analysing your Kotlin module points to the possibility of the C/C++ JNI code being identical to the Java module's generated code. This means we could also consider a Java to Kotlin converter for the generated .java files. But as mentioned, it's arguably pointless if you can get the same compile time advantages from Kotlin with the appropriate nullability annotations in the generated .java files.

erezgeva:
Hi @TVDiasDominik and @wsfulton,

I do understand the need to generate Kotlin only and replace all Java code.
But on the other hand William is right, maintaining a separate Language is costly, and as 90% of the C++ module code and the interface files are the same.
It make more sense to integrate the Kotlin as a sub module of Java.

Dominik think of Java as JVM base and not as a separate language.

By merging the two, we can get both goals, reduce maintenance from one side and on the other side serve our costumers whom do not wish to remove Java in their project (which I personally can understand fully).

It does bare some overhead on Java, but so is elua, and using 3 engines of Javascript. Adding new feature usually does :smile: 

At the end we want to serve our costumers the best we can and provide them with much more features we can.
And do so with the minimum effort from our side.

See: #3514 for the result

Yours
Erez

TVDiasDominik:
@wsfulton I was on vacation so sorry for the delayed answer.
Yeah, I agree most of the code is duplicated as java and kotlin share the jvm. Just from a kotlin dev it would be nicer to only have kotlin but I also understand the maintenance problem. I still need to look at the solution you proposed @erezgeva it sounds nice and if that also fits us it would be super :D

erezgeva:
> @wsfulton I was on vacation so sorry for the delayed answer. Yeah, I agree most of the code is duplicated as java and kotlin share the jvm. Just from a kotlin dev it would be nicer to only have kotlin but I also understand the maintenance problem. I still need to look at the solution you proposed @erezgeva it sounds nice and if that also fits us it would be super :D

I use the same kotlin testing scripts you wrote.
I only moved the code in `kotlin.cxx` back to `java.cxx` and `Lib/kotlin` back to `Lib/java`. And update some `Examples/test-suite/*.i`.
That is basically the change :smile:

PR #3490, opened 2026-07-09
Lua: Rename the `near` function to `cmp_flt` function that compare float numbers

### LINKED ISSUE #3464: Lua li_std_map test failure with lua-5.3
```
/usr/bin/lua5.3: ../../../../../Examples/test-suite/lua/li_std_map_runme.lua:78: iscpmap:get(i)
stack traceback:
	[C]: in function 'assert'
	../../../../../Examples/test-suite/lua/li_std_map_runme.lua:78: in main chunk
	[C]: in ?
make[1]: *** [Makefile:59: li_std_map.cpptest] Error 1
```
Seems to be a flaky test. @erezgeva, this is a new test you add a few days ago - could you try identify the flakiness, otherwise, let's delete the test.

erezgeva:
Closes #3464

And move `cmp_flt` to helpers.
Although Lua uses C floating math.
The actual used C floating functions in Lua
 might be different from the actual floating functions
 used in the original C code, which may lead to a slightly
 different floating number.
So for calculated floating numbers we should use the `cmp_flt` function in the assert.


wsfulton:
Thanks 👍

PR #3492, opened 2026-07-12
Fix boost compilation

erezgeva:
Ensure `--with-boost` works properly.

- We need override `CXXFLAGS`.
  As we define them when we use `Tools/testflags.py`.
- MacOs GHA use `--with-boost`.
- MSVC may use `-O2` on build, but omitting for testing.
- Use serialised PDB file for MSVC, to allow parallel compilation.


wsfulton (changes requested):
Looks like you missed CFLAGS, both CFLAGS and CXXFLAGS should be handled consistently.

erezgeva:
> Looks like you missed CFLAGS, both CFLAGS and CXXFLAGS should be handled consistently.

`CXXFLAGS` need a separate handling due to the boost include flag.
But we can make them same handling.
Including the override in `Examples/Makefile.in`

erezgeva (commented):
Done

PR #3493, opened 2026-07-13
Restore Lua 5.1 support / LuaJIT support, including directors

wsfulton:
Restores Lua 5.1 and LuaJIT support, which #3394 dropped, and extends it to directors,
which have never worked with Lua 5.1. Lua 5.1 is now tested in CI so this cannot regress
silently again.

## Why

74ff929bd (#3394) removed the compatibility defines for `lua_rawlen`,
`lua_pushglobaltable` and `lua_rawsetp`/`lua_rawgetp` from `Lib/lua/luarun.swg`. These are
in the runtime compiled into **every** generated Lua wrapper, not just those using
directors, so any SWIG generated Lua module now fails to compile against Lua 5.1 or
LuaJIT:

```
swigluarun.h:1555:19: error: implicit declaration of function 'lua_rawlen'
```

Found by a Fedora COPR mass rebuild against SWIG master: obs-studio's Lua bindings, which
are compiled against LuaJIT, failed to build.

#3394 removed more than its actual blocker, which was that directors use
`lua_setuservalue()`, added in Lua 5.2. Dropping Lua 5.1 was accepted on the grounds it
was old and untestable; both turn out to be avoidable. LuaJIT will not grow these
functions - it is
[API+ABI-compatible with Lua 5.1](https://luajit.org/extensions.html), and
`LUAJIT_ENABLE_LUA52COMPAT` only adds Lua level features, never Lua/C API additions.

## What

1. **Restore the compatibility defines**, for Lua 5.1 and later only (the Lua 5.0 shims
   are not restored). `configure.ac` accepts Lua 5.1 again - this only governs which Lua
   the test-suite and examples run against, as Lua is not needed to build SWIG itself.
   Also fixes the examples and test-suite where they themselves used Lua 5.2 only features
   (`LUA_OK`, `_ENV`, and `string.format("%c", ...)`, which drops NUL bytes in Lua 5.1).

2. **Director support for Lua 5.1**, using the userdata environments that uservalues
   replaced. Note `lua_getuservalue` cannot simply be defined to `lua_getfenv`: a new
   userdata has no uservalue in Lua 5.2, but in Lua 5.1 it inherits its creator's
   environment - Lua's `package` table, shared by every userdata the module creates - so
   the overrides would be shared between all objects and corrupt that table. The override
   tables are therefore marked, and an unmarked environment reported as nil.

3. **Remove two unused director helpers**. No change to generated code.

## Testing

- New Lua 5.1 job in the Linux GHA matrix.
- Not covered: running the test-suite under LuaJIT itself. That needs build system work
  (Ubuntu's `libluajit-5.1-dev` ships only a static, non-PIC library, so a shared module
  cannot link it) so we rely on the Lua 5.1 only testing.

@erezgeva, could you please review before I merge.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -122,6 +122,8 @@ jobs:
           VER: '12'
           CPPSTD: c++11
         - SWIGLANG: lua
+          VER: '5.1'
+        - SWIGLANG: lua
           VER: '5.2'
         - SWIGLANG: lua
           VER: '5.3'
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,11 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.5.0 (in progress)
 ===========================
 
+2026-07-12: wsfulton
+            [Lua] Restore support for Lua 5.1 and LuaJIT (which implements the
+            Lua 5.1 API) for non-director wrapping. Reverts part of #3394 which
+            dropped Lua 5.1 support.
+
 2026-07-11: wsfulton
             Fixed a protected or private nested class deriving from a class used
             elsewhere in the wrapped API producing a runtime upcast helper function
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -8,6 +8,13 @@ Version 4.5.0 (in progress)
 ===========================
 
 2026-07-12: wsfulton
+            [Lua] Directors now work with Lua 5.1 and LuaJIT, which do not have the
+            Lua 5.2 uservalue functions ('lua_getuservalue'/'lua_setuservalue') used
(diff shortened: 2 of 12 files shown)


erezgeva:
> Fedora COPR

The PR looks OK.

@christophe-calmejane do you have any other thoughts?

wsfulton:
With the lua-5.1 tests all passing, I'm merging to master so that it can be tested on the Fedora setup by the obs-studio project with LuaJIT. @christophe-calmejane, if there is anything to address, we can do it as an extra commit.

wsfulton:
merged, see 3c29f45e9...5124908ed

christophe-calmejane:
Looks fine to me, I'll try to build one of my big project against this commits today, just in case, and report if I find any blocker

PR #3494, opened 2026-07-14
Fix li_stdint test

erezgeva:
Converting signed to unsigned may change the value.

The test need to pass values unchanged.
As long as the values are in the type range!
Converting signed to unsigned we may convert the value to a different one if the value is outside the range.
Though the test fails.
For example on `i386`
`uint16_td(0xffff) => 0x7fff + 1`

@wsfulton 
This should solve the failure, Jitka from Red Hat found on  `i386` with `li_stdint_runme.lua`.

The reason we did not see it before is that the test was exist on PHP only, while Fedora on `i386` do not use PHP.
See:
https://kojipkgs.fedoraproject.org//work/tasks/5365/147615365/build.log

````
The SWIG test-suite and examples are configured for the following languages:
c lua octave perl5 python ruby tcl 
````

### DIFF
--- a/Examples/test-suite/li_stdint.i
+++ b/Examples/test-suite/li_stdint.i
@@ -18,10 +18,10 @@
   int16_t  int16_td(int16_t i) { return i; }
   int32_t  int32_td(int32_t i) { return i; }
   int64_t  int64_td(int64_t i) { return i; }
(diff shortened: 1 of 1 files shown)


wsfulton:
Thanks @erezgeva

PR #3497, opened 2026-07-15
Fix SWIG_CastCmpStruct to provide a strict weak ordering

jschueller:
The comparator mixed type-pointer ordering with the next-field priority, violating strict weak ordering. This could make the binary search in SWIG_TypeCheck miss valid type entries and fail at runtime.

Separate the two criteria cleanly: sort by next-field first (active entries before filtered ones), then by type pointer within each group.


wsfulton:
This seems to be a technically correct, so I'll merge it in. Shame there is no user visible fix/testcase.

PR #3500, opened 2026-07-22
[C#] Support setting the namespace via %module(csnamespace="...")

### LINKED ISSUE #3495: C# namespace in .i instead of command line.
I am not able to put a C# namespace inside the `.i` file. Is there any example, or is it not supported? Currently I am using the command-line parameter, but that is inconvenient.

### Desired usage

Declare the C# namespace in the interface file instead of (or in addition to) `-namespace` on the command line:

```swig
%module(namespace=Example.Net) example

%{
#include example.h
%}
%include example.h
```

so generated C# lands in that namespace, e.g.:

```csharp
namespace Example.Net {
  public class Circle { ... }
}
```

### Current workaround

```bash
swig -csharp -namespace Example.Net -module example example.i
```

Having to pass `-namespace` on every invoke is awkward, especially for MSBuild / IDE integration where the `.i` is the natural place to keep module and namespace settings.

sparcyu:
Fixes #3495.

Allows the C# namespace to be specified via the `csnamespace` option of the `%module` directive in the interface file, as an alternative to the `-namespace` command line option. This is convenient for build system / IDE integration (e.g. MSBuild) where the `.i` file is the natural place to keep module settings:

```swig
%module(csnamespace="Example.Net") example
```

generates code into `namespace Example.Net { ... }`, exactly as `-namespace Example.Net` does. When both are specified, the `-namespace` command line option takes precedence.

Updated after review: this now uses the existing mechanism for target language custom options passed to `%module`, as already used for `imclassname` and `csbegin`, so the change is confined to `csharp.cxx`. **`parser.y` is untouched.**

The option is named `csnamespace` rather than `namespace` because `namespace` is a C++ keyword (`Source/CParse/cscanner.c` only treats it as one when `cparse_cplusplus` is set), so `%module(namespace=...)` is a syntax error in `-c++` mode — which is the common case — and would have required a grammar change. `csnamespace` also matches the existing naming convention for target language options (`imclassname`, `csbegin`, `ruby_minherit`, `jniclassname`).

### Changes

- **`Source/Modules/csharp.cxx`**: read the `csnamespace` option from the module `options` node in `top()`, only when `-namespace` was not passed so the command line wins. Mirrors the existing handling of `imclassname`, `directors`, etc.
- **Test**: `Examples/test-suite/csharp_module_namespace.i` + runme. The C# test Makefile passes `-namespace <testcase>Namespace` to every testcase, so the global option is factored into a `CSHARP_NAMESPACE_OPT` variable and cleared for this testcase (target-specific override) so the `%module` namespace is the one exercised.
- **Docs**: documented in `Doc/Manual/CSharp.html`.
- **`CHANGES.current`** updated.

### Testing

Built swig with CMake and verified:
- `%module(csnamespace="Example.Net")` → `namespace Example.Net { ... }`, in both C and C++ mode
- `-namespace` on the command line overrides the `csnamespace` module option
- `csnamespace=""` → no namespace emitted
- no option set → unchanged (global namespace, no regression)
- the new testcase's generated C# compiles together with its runme

### DIFF
--- a/CHANGES.current
+++ b/CHANGES.current
@@ -7,6 +7,19 @@ the issue number to the end of the URL: https://github.com/swig/swig/issues/
 Version 4.6.0 (in progress)
 ===========================
 
+2026-09-11: sparcyu
+            [C#] Add support for setting the C# namespace via the 'csnamespace' option of
+            the %module directive, as an alternative to the -namespace command line
+            option. This is convenient for build system integration where the interface
+            file is the natural place to keep the module settings, eg:
+
+              %module(csnamespace="Example.Net") example
+
+            The -namespace command line option takes precedence over the csnamespace
+            module option when both are specified.
+
+            Fixes #3495.
+
 2026-09-03: larskanis, wsfulton
             #3534 Fix configure error "Tools/convertpath: No such file or directory" on
             MinGW. The removed convertpath tool was still used to compute the SWIG library
--- a/Doc/Manual/CSharp.html
+++ b/Doc/Manual/CSharp.html
@@ -197,6 +197,17 @@ Note that by default, the generated C# classes have no namespace and the module 
 </li>
 
 <li>
+The namespace can alternatively be specified in the interface file itself using the <tt>csnamespace</tt> option of the <a href="CSharp.html#CSharp_module_directive"><tt>%module</tt> directive</a>, avoiding the need to pass <tt>-namespace</tt> on the command line. This is convenient for build systems where the interface file is the natural place to keep the module settings. For example:
+
+<div class="code">
+<pre>
+%module(csnamespace="com.bloggs.widget") widget
+</pre>
+</div>
+generates code into the same namespaces as the <tt>-namespace com.bloggs.widget</tt> command line option above. If both are specified, the <tt>-namespace</tt> command line option takes precedence over the <tt>csnamespace</tt> module option.
+</li>
+
+<li>
 The <a href="SWIGPlus.html#SWIGPlus_nspace">nspace feature</a> is also supported as described in this general section with a C# example.
 Unlike Java which requires the use of the -package option when using the <tt>nspace</tt> feature, the -namespace option is not mandatory for C#.
 </li>
--- a/Examples/test-suite/csharp/Makefile.in
+++ b/Examples/test-suite/csharp/Makefile.in
@@ -27,6 +27,7 @@ CPP_TEST_CASES = \
 	csharp_features \
 	csharp_lib_arrays \
 	csharp_lib_arrays_bool \
+	csharp_module_namespace \
 	csharp_namespace_system_collision \
 	csharp_prepost \
 	csharp_swig2_compatibility \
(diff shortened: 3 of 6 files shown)


sparcyu:
Can any of the maintainers take a look?

wsfulton:
Definitely. It's not a good time right now though due to the imminent 4.5.0 release. I would have considered this patch at this late stage if it was low risk, however it changes the parser. I would expect it to only contain changes in csharp.cxx though and use the designed approach for custom attributes passed to `%module`. See for example the `ruby_minherit` implementation in ruby.cxx.

wsfulton:
Also, c# has a target language custom option called imclassname, which you'll see in csharp.cxx.

sparcyu:
Thanks, that makes sense — reworked accordingly.

The change is now confined to `csharp.cxx` and uses the existing mechanism for custom options passed to `%module`, the same as `imclassname`. `parser.y` is untouched.

The option is spelled `csnamespace` rather than `namespace`: `namespace` is only treated as a keyword when `cparse_cplusplus` is set, so `%module(namespace=...)` parses in C mode but is a syntax error under `-c++`, which is the common case. Using a non-keyword name avoids the grammar change entirely and also lines up with the naming of the other target language options (`imclassname`, `csbegin`, `ruby_minherit`, `jniclassname`).

Verified that `%module(csnamespace="Example.Net")` generates into `namespace Example.Net`, that `-namespace` still takes precedence when both are given, and that the new testcase's generated C# compiles against its runme.

No rush on this given the 4.5.0 release — happy for it to wait.

sparcyu:
Rebased onto current `master` now that 4.5.0 is out. The only conflict was `CHANGES.current` (4.6.0 changelog reset); the `csnamespace` entry is at the top of the 4.6.0 section. Implementation is unchanged: `csharp.cxx` only, `%module(csnamespace=...)`, `-namespace` still wins.

sparcyu:
Temporarily closing and reopening to retrigger GitHub Actions (macOS run was cancelled; we cannot rerun it without admin on swig/swig).

sparcyu:
@wsfulton Following up now that 4.5.0 is out.

This is the shape you asked for: confined to `csharp.cxx`, `%module(csnamespace=...)` via the same custom-option path as `imclassname`, no `parser.y` change. `-namespace` still takes precedence when both are given.

CI is green on linux, windows, macos, and cmake after a retrigger (the previous macOS run had been cancelled). The branch currently conflicts with `master` — likely `CHANGES.current` again. Happy to rebase whenever you have time to look.

sparcyu:
Rebased onto current `master`. The only conflict was `CHANGES.current`; the `csnamespace` entry is at the top of the 4.6.0 section again.

sparcyu:
@wsfulton Gentle ping — rebased onto current `master` again (only `CHANGES.current`). The patch is still `csharp.cxx` only, `%module(csnamespace=...)` on the same custom-option path as `imclassname`, no parser change. Whenever you have a moment after 4.5.0 would be appreciated.

sparcyu:
Rebased onto current `master` (only `CHANGES.current`). Implementation unchanged: `csharp.cxx` only, `%module(csnamespace=...)`, `-namespace` still wins.

sparcyu:
@wsfulton Gentle ping — rebased onto current `master` on 11 Sep, CI is green (linux/windows/macos/cmake), mergeable. Still `csharp.cxx` only, `%module(csnamespace=...)` on the same custom-option path as `imclassname`. Happy to adjust if anything else is needed.

PR #3506, opened 2026-07-22
[Python] Fix refcount underflow in director

### LINKED ISSUE #2015: Callback to target language: deletion of interned string failed
Dear swig team,
 
I am trying to to a  python to C to python callback

I am using swig version 4.02 with python 3.8 on redhat  but I ad the exact same problem with swig 3.0.12 and python 3.6 on ubuntu
```
swig -version:
    SWIG Version 4.0.2
    Compiled with g++ [x86_64-pc-linux-gnu]
   Configured options: +pcre
```
Running the following program (I attached a zip file), 
[swig_test.zip](https://github.com/swig/swig/files/6437004/swig_test.zip)
I get the following output:

```
python ./test.py
C Callback to Python 4c1457b0
Python Callback
1234567890123456789012345678901234567890
C Callback to Python 4c1457b0
Python Callback
1234567890123456789012345678901234567890
C Callback to Python 4c1457b0
Python Callback
1234567890123456789012345678901234567890
C Callback to Python 4c1457b0
Python Callback
1234567890123456789012345678901234567890
Fatal Python error: deletion of interned string failed
Python runtime state: initialized


[user2@localhost swig_test]$ 

```

Here is my Makefile:

```
build:
        /bin/swig -python -c++ -threads example.i
        gcc -O2 -fPIC -I/usr/include/python3.8/ -c example_wrap.cxx
        gcc -shared -lm -lstdc++ -o _example.so example_wrap.o

```
Now my example.i:

```
/* File : example.i */
%module(directors="1") example

/* turn on director wrapping Callback */
%feature("director") Callback;

%inline %{
    void* toVoidP(PyObject *v) {
        return (void *) v;
    }

    struct Callback {
        virtual void EthListener(PyObject* param1) = 0;
        virtual ~Callback() {}
    };

    static Callback *eth_listener_handler_ptr = NULL;

    void* ethListenerCallback(void* param1)
    {
        while(true){
            printf("C Callback to Python %x\n", param1);
            eth_listener_handler_ptr->EthListener((PyObject*) param1);
            sleep(1);
        }
        return NULL;
    }

    void eth_listener_wrapper(Callback *handler, void* param1) 
    {
        eth_listener_handler_ptr = handler;
        
        pthread_t thread;
        pthread_create( &thread, NULL, ethListenerCallback, param1);
    }
%}

void* toVoidP(PyObject *);
%constant void (*ethListenerCallbackPtr)(void*) = ethListenerCallback;
```
and now my small program test.py:
```
import time
import example

# Callback class is defined and derived from C++ class BinaryOp
class PythonCallback(example.Callback):
	# Define Python class 'constructor'
	def __init__(self):
		# Call C++ base class constructor
		self.counter = 0
		example.Callback.__init__(self)


	# Override C++ method
	def EthListener(self, param1):
		print ("Python Callback")
		print(param1)
		return



param1 = "1234567890123456789012345678901234567890"
handler = PythonCallback()
example.eth_listener_wrapper(handler, example.toVoidP(param1))

while True:
	time.sleep(1)`

```

jschueller:
Each director upcall for a method taking PyObject* created a SwigVar_PyObject, assigned the C++ parameter (a borrowed reference) to it, then let the SwigVar_PyObject destructor DECREF it on scope exit -- with no matching INCREF.  After enough calls the object's reference count underflowed, the object was freed prematurely, and the next call crashed with 'deletion of interned string failed'.

Fix: add a %typemap(directorin,noblock=1) for PyObject* that INCREFs the parameter after assignment, balancing the destructor's DECREF.

Fixes #2015.

Assisted-by: opencode (deepseek-v4-flash)

### DIFF
--- a/Examples/test-suite/director_pyobject.i
+++ b/Examples/test-suite/director_pyobject.i
@@ -1,0 +1,14 @@
+%module(directors="1") director_pyobject
+
+%feature("director") Callback;
+
+%inline %{
+    struct Callback {
+        virtual void callback(PyObject* param1) = 0;
+        virtual ~Callback() {}
+    };
+
+    void call_callback(Callback *handler, PyObject* param1) {
+        handler->callback(param1);
+    }
+%}
--- a/Examples/test-suite/python/Makefile.in
+++ b/Examples/test-suite/python/Makefile.in
@@ -26,6 +26,7 @@ CPP_TEST_CASES += \
 	callback \
 	complextest \
 	director_guard \
+	director_pyobject \
 	director_stl \
 	director_wstring \
 	file_test \
--- a/Examples/test-suite/python/director_pyobject_runme.py
+++ b/Examples/test-suite/python/director_pyobject_runme.py
@@ -1,0 +1,36 @@
+import sys
+import gc
+import director_pyobject
+
+class PyCallback(director_pyobject.Callback):
+    def __init__(self):
+        director_pyobject.Callback.__init__(self)
+    def callback(self, param1):
+        pass
+
+handler = PyCallback()
+
+# Use a bytearray (mutable, never interned, never immortal)
+obj = bytearray(b"test data")
+refs = []
+
+for i in range(100):
+    before = sys.getrefcount(obj)
+    director_pyobject.call_callback(handler, obj)
+    after = sys.getrefcount(obj)
+    if before != after:
+        raise RuntimeError(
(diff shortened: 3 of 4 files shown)


wsfulton:
Committed, this needed some additional changes, see 1f1f7731315e748c0b6920277b2c5eb181278471...ff57c2aa6bcdc87f132ec8cc8a3afe9d191c177e

PR #3510, opened 2026-07-22
ci: Cancel in-progress jobs when pushing new commits

Nerixyz:
The Swig repo runs a lot of GitHub Actions jobs on a PR to test many supported configurations/languages. When pushing commits to a PR, this will request another batch of jobs. When the previous jobs aren't done yet, they will still continue, even though their result isn't needed anymore.

With this change, any previous jobs are cancelled when pushing new commits. This follows the example from the [GitHub actions docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-using-concurrency-and-the-default-behavior).

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -28,6 +28,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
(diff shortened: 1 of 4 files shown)


jschueller (approved):

wsfulton:
I need pushed jobs to complete. The onus is on authors to push only when they have finished their work.

PR #3513, opened 2026-07-24
CI: Fix node18 install

jschueller:
fixes the javascript / node 18 / gcc 13 / c++17 job


wsfulton:
Can you explain? The `javascript node 18 gcc13 c++17` failures have gone away. I diagnosed it with Opus 4.8 yesterday and it concluded a truncated header file was the problem and it was a temporary problem. What's your full diagnosis to arrive at this pull request?

jschueller:
yes, seems its gone

PR #3514, opened 2026-07-25
[Java-Kotlin] Add Kotlin using JVM as Java sub-module 

erezgeva:
Kotlin can be used with JVM, Javascript and native.
This PR covers Kotlin with JVM only as a sub module of the existing java module.

The PR is a rebase of #3477 by @TVDiasDominik

Thanks Dominik for your work.

The PR integrate the Kotlin back into Java module and make it a sub-module.

Regarding Java to Kotlin translation, there is **no** reliable tool to perform the translation, and as Kotlin popularity increased, I think it is better to generate directly and test with Kotlin code. 

Bellow are tests performed on GHA, with Java as comparing.

All tests use C++ 20
Java runs 777 tests
Kotlin runs 739 tests
|                   |Total      |Tests      |
|-------------------|-----------|-----------|
|Linux              |           |           |
|    Java           |11:53      |10:28      |
|    kotlin         |47:05      |45:59      |
|MacOS              |           |           |
|    Java           |15:31      |10:42      |
|    kotlin         |34:35      |32:24      |
|Windows            |           |           |
|    Java-ucrt      |13:49      |10:45      |
|    kotlin-msvc    |1:11:07    |1:06:44    |
|    kotlin-gcc     |59:47      |56:17      |
|    kotlin-ucrt    |1:02:42    |59:28      |

As used with most Kotlin, GHA images use one version of last Kotlin compiler, [k2-compiler](https://kotlinlang.org/docs/k2-compiler-migration-guide.html).
But as we might need to test different Kotlin compilers in the future, we leave `VER` variable unused for now.
However the Kotlin can use different JVM versions as Java does, using the `JVM_VER` variable to set the proper `JAVA_HOME` for the test.
In comparing Java, we use the same  JVM and JDK version, so we can use the `VER` variable for both in a similar way we do with other languages.

Although the K2 compiler is faster than its predecessor,  it is much slower than Java and other languages.
So although the PR support using Kotlin on all available GHA images,
this PR add only one test on MacOS, since it is the fastest test.


erezgeva:
Doing a 2nd test and compare Ubuntu 26 and MacOS
MacOS is the fastest to compile Kotlin regardless of clang or gcc.
Perhaps JetBrains prefer MacOS for development?  
Can not find any reference to back it, but the results are clear.

2nd round
|                   |Total      |Tests      |
|-------------------|-----------|-----------|
|Linux              |           |           |
|    Java           |11:55      |10:30      |
|    kotlin-gcc       |44.58      |42:53      |
|    kotlin-clang   |47.03      |46:01      |
|MacOS              |           |           |
|    Java           |11:40      |9:03       | 
|    kotlin-clang   |28:49      |26:07      |

* using clang on Linux is slower perhaps due to lack of cache

erezgeva:
Hi @TVDiasDominik,

Manage to remove all warnings.
Some with proper `@Suppress` and some by fixing.

<del>We left with 6 tests that fail.</del>
<del>Form whom 2 lack Kotlin XXX_runme.kt source code.</del>
We left with 2 tests that fail.

The 6 tests have the same errors on #3477.
The rebase is full.

<del>The 6 tests related to 3 different errors:</del>
The 2 tests related to 1 error:

<del>`error: cannot access 'method': it is protected in 'class'`</del>
`java.lang.NoSuchMethodError` from JNI
-    cpp11_director_using_constructor.cpptest
-    director_protected.cpptest
-    <del>director_using.cpptest</del>

<del>`error: null can not be a value of a non-null type 'type'`</del>
-   <del>li_boost_shared_ptr_director.cpptest</del>
-    <del>naturalvar_onoff.cpptest</del>

<del>`error: unresolved reference: Foo`</del>
-   <del>director_shared_ptr.cpptest</del>

Please notice, as Kotlin is a sub-module it is **NOT** experimental any more, as this is a privilege of a complete module.
That means that once we merge.
All changes, Java or any other that breaks Kotlin will be adrressed.
We will probably need your assistance as most of us are novice of Kotlin.

P.S.
The tests need to pass on GitHub action, preferably without warnings.
Running tests locally is great for development, but insufficient for merging.

erezgeva:
Update, fix

- `'director_shared_ptr` 
   Need to fix: FooDerived fail to inherit callPong() and callUpcall()
 - `naturalvar_onoff`

erezgeva:
Update: Fix `li_boost_shared_ptr_director.cpptest` and `boost_shared_ptr.i‎`

erezgeva:
Update:
- Clean Kotlin test-suite Makefile
- Add filters for `SWIGWARN_PARSE_KEYWORD` in test suite interface files.

erezgeva:
Update fix `director_using`

As Java `protected` method can be used from any class in the package.
The equivalent is using Kotlin `internal`.
Update `java.cxx`

PR #3515, opened 2026-07-27
CI: Add abi3t job (abi3+free threaded)

jschueller:
see https://peps.python.org/pep-0803/

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -213,11 +213,13 @@ jobs:
           SWIG_FEATURES: -builtin -O
         - SWIGLANG: python
           VER: '3.15'
-          continue-on-error: true
         - SWIGLANG: python
           VER: '3.15'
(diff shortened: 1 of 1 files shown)


wsfulton:
We are minimally testing 3.15 atm as it is not released and we have too many jobs. You've removed the continue-on-error without explanation. The commit message needs to have some explanation as to why this is different to 3.14t testing, just referencing a PEP is too brief.

jschueller:
yes, maybe its not worth yet

wsfulton:
What was your thinking here? Do we not need the special 't' build in 3.15 now in order to test free-threaded builds? It was quite complex getting the free-threaded testing turned on, is it different in 3.15?

jschueller:
I was just checking free-threaded flavor has no conflict with the stable abi, since that combination (abi3t) will become kind of the new standard with free-threaded flavor on track to becoming the new default, or so as I understood this PEP

PR #3516, opened 2026-07-28
[Python] Customize return annotation for multiple argout matches

### LINKED ISSUE #3469: [Python] PEP 484: Improve/customize return type when argout typemaps are used
Currently, when argout typemaps are used, the return type can be incorrect. 

```swig
%typemap(in, numinputs=0) (short *short_list) { /* ... */ }
%typemap(in, numinputs=0) (size_t short_list_len) { /* ... */ }
%typemap(freearg) (short *short_list) { /* ... */ }
%typemap(argout) (short *short_list, size_t short_list_len) { /* drop result and create list */ }
%typemap(pytyping) (short *short_list, size_t short_list_len) "typing.List[int]"
size_t get_list_argout(short *short_list, size_t short_list_len);
```

Currently, this yields:
```python
def get_list_argout() -> "typing.List[typing.Union[int, typing.List[int]]]":
```
However, it should be `typing.List[int]`, because we drop the previous result.

There should be a way to tell Swig that we overwrite the original return value. Maybe an `argoutaction` with value `overwrite`.

Nerixyz:
When multiple argout matches are in a function or there is one argout match and one return type, Swig would previously concatenate them with `, `. This annotation is invalid. Furthermore, multi-argument pytyping typemaps weren't respected.

With this PR, both are fixed. By default, it's assumed the argout typemaps use `SWIG_AppendOutput` which creates a list. So by default, the annotation will look like `typing.List[typing.Union[A, B, C]]`. The prefix, joiner, and suffix can be customized. Furthermore, you can specify that the return value is discarded.

Fixes #3469.

### DIFF
--- a/Doc/Manual/Python.html
+++ b/Doc/Manual/Python.html
@@ -7149,9 +7149,19 @@ def do_something(t: "int|float") -&gt; "int":
 When <tt>argout</tt> typemaps are used, the return annotation is built from the wrapped return
 type and each matched <tt>pytyping</tt> typemap for the <tt>argout</tt> parameters. A single
 returned value keeps its type directly if the function signature returns void. Multiple returned values default to
-<tt>typing.List[typing.Union[...]]</tt>. There will be multiple returned values if the function signature returns non-void
-and there is at least one parameter that is an argout. This matches the behaviour when <tt>SWIG_AppendOutput</tt> is used.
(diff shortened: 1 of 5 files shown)


wsfulton:
Hi @Nerixyz, can you please take a look at #3532 as my preferred implementation and add any comments there.

Nerixyz:
Closing in favor of #3532.

PR #3517, opened 2026-07-28
Allow easily ignoring clang-format commit when running git-blame

vadz:
Add the file which can be used to ignore the specified commits when running git-blame and add the commit corresponding to the Great Source Reformatting to it.

Put it in the location used by GitHub web UI so that it would also be used there.

[skip ci]

----

Please consider merging this trivial PR as it makes it much more convenient to use git-blame.


wsfulton:
That is a truly beautiful trick.

I see Github picks it up automatically too when viewing the blame in Github.

Issue #2165, opened 2022-01-25
No way to get type annotations for argout typemaps


krishauser:
It appears that argout typemaps disappear from any SWIG output in Python (whether --py3 is used or not). The output type seems to disappear from the docstring and type annotations. 

So `void my_func(int& out)`

will show up as `def my_func() -> "void":`

I am trying to create a tool that will automatically generate Python-friendly type hints (https://github.com/krishauser/improve_swig_docs) and this is making it difficult to get intepretable output types.

ojwb:
See also #735 (for the benefit of other readers - I see you're aware of that)

wsfulton:
Fixed by a number of pytyping commits but the final one to fix this is #3518.

PR #3518, opened 2026-07-28
[Python] Support INOUT/INPUT/OUTPUT for pytyping

Nerixyz:
This fixes the annotations when using the INOUT/INPUT/OUTPUT typemaps. Before, this would use the name for `Type*` or `Type&`. With this PR, the name for `Type` is used.


wsfulton:
Merged with some fixes required for the python_annotations_typing testcase in cbdcd40c66912a0d181d6d5c0de4a19cb4266e76. Thank @Nerixyz. Also fixes #2165.

wsfulton:
Further related work was added in cbdcd40c66912a0d181d6d5c0de4a19cb4266e76...b0f7fce8b06d4d27ec12431a0f45434e16e34ec6

PR #3519, opened 2026-07-29
[C#] Re-implementation of AOT support from PR #1262

willadsen:
#1262 has not been touched for years, but I needed that functionality, so I've necromanced that PR in a fresh state. Will add tests


erezgeva:
@willadsen 

Although you are on draft.
I think you have too many commits on you PR.
Many are not relevant, and some are just past merged.

I do respect original authors, so I never commit someone else under my name but the original author.
But as you ask to change only 4 files, I am sure you can reduce most of these commits, whom are not related to any of these 4 files, and remain with only these commits that only contribute directly to these 4 files. You can also squish commits of from the same author, the date is less important. the history is important only if it means different parts of actual development.

willadsen:
@erezgeva @wsfulton based on the previous PR there were different opinions about what kind of tests to add, so I'll wait for feedback on that until making some

PR #3520, opened 2026-07-29
[Python][Typing] Annotate std::vector

Nerixyz:
This adds a PEP 484 (pytyping) annotation for `std::vector` and works
towards having annotations for all supported `std::` types. Before
this, `std::vector` was either annotated as `typing.Any` or as
`IntVector` (i.e. the proxy name), if
`%typemap(pytyping) SWIGTYPE "$&pytypename"` is used. The former is too
broad (we don't accept any object) and the latter is too restrictive
(any iterable is accepted).

With this, the annotation for `std::vector<T>` is
`collections.abc.Iterable[T]` and `typing.Tuple[T, ...]` as a return
value.

To be able to annotate `std::vector`, two changes were made:
- Added a special variable `${n}_targ{m}` to get the typename of the
  m-th template argument in the n-th type. This was required, because
  the vector type is declared in `std/std_vector.i`, so in
  `python/std_vector.i`, we don't have access to the value type.
  I experimented with passing the type through `%swig_vector_methods`.
  For simple types this works well, but if you have a
  `vector<vector<int, Alloc>, Alloc>`, the comments become a problem.
  Swig will wrap the type in parentheses, but then you can't pass it
  into `$typemap()` anymore.
- Added a new `$typemap(meth:attr?, ty)` syntax that looks up
  `$typemap(meth:att, ty)` and falls back to `$typemap(meth, ty)` if the
  former is undefined. This matches the behavior for the return value in
  `pytyping`.

The changes here are scoped to `std::vector` to get the direction
right. The other `std::` types will likely be similar.

### DIFF
--- a/Doc/Manual/Contents.html
+++ b/Doc/Manual/Contents.html
@@ -590,6 +590,7 @@
 <li><a href="Typemaps.html#Typemaps_special_macro_descriptor">$descriptor(type)</a>
 <li><a href="Typemaps.html#Typemaps_special_macro_typemap">$typemap(method, typepattern)</a>
 <li><a href="Typemaps.html#Typemaps_special_macro_typemap_attribute">$typemap(method:attribute, typepattern)</a>
+<li><a href="Typemaps.html#Typemaps_special_macro_typemap_attribute_optional">$typemap(method:attribute?, typepattern)</a>
 </ul>
 <li><a href="Typemaps.html#Typemaps_special_variable_attributes">Special variables and typemap attributes</a>
 <li><a href="Typemaps.html#Typemaps_special_variables_and_macros">Special variables combined with special variable macros</a>
--- a/Doc/Manual/Typemaps.html
+++ b/Doc/Manual/Typemaps.html
@@ -53,6 +53,7 @@
 <li><a href="#Typemaps_special_macro_descriptor">$descriptor(type)</a>
 <li><a href="#Typemaps_special_macro_typemap">$typemap(method, typepattern)</a>
 <li><a href="#Typemaps_special_macro_typemap_attribute">$typemap(method:attribute, typepattern)</a>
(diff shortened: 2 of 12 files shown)


erezgeva:
Hi @Nerixyz

You probably need to run `make format` to fix the `Source/Modules/python.cxx` and `Source/Swig/typemap.c‎` syntax format, than you can actually run the GHA tests.

I also recommend the change the PR to draft, until it pass all tests and have the content and descriptions you intend.

Please add a full and comprehensive explanation
1. why do we need the change?
1. What is the goal of the change?
1.  How do we achieve these goals?
1. In short there is no context, and no thread that connect the text to a story, that other developers can follow. It is more like sporadic thoughts from you mind. Do not fear from a long description, it is OK to start a PR with a long description, as long as it is coherent and not deviate from the thread of the story.
1.  Leave future continuations to the end, please do not start with the future, start with now and end with the future.
1. The lack of proper explanation is valid to `Doc/Manual/Typemaps.html`. It is too short and confusing. I also recommend to add an example, example are an excellent way to focus on what the feature provide and how to use it. It does not replace the test you add, it is double, which is important to other developer's understanding.
1. We know the keywords, but they are insufficient to understand the change, as the keywords may be used in different ways to achieve different goals. I do not refer to the title of the PR, they are short and contains keywords like yours, I refer to the reset!
1. The commit itself should also be with a description, focus on the change itself. It does not have to be long and probably a subset of the PR description that focus on the change itself, i.e. what is the commit actually change and add.

Nerixyz:
> You probably need to run `make format` to fix the `Source/Modules/python.cxx` and `Source/Swig/typemap.c‎` syntax format, than you can actually run the GHA tests.

Looks like that was a problem on master which is now fixed.

> I also recommend the change the PR to draft, until it pass all tests and have the content and descriptions you intend.

Checked that it passes the tests locally, so this is ready for review in my mind.

> Please add a full and comprehensive explanation

Updated the description with your feedback.

PR #3522, opened 2026-07-30
[Python] Pass  `SWIG_FromCharPtrAndSize` from SWIG Preprocessing to compiler

erezgeva:
Hi @wsfulton 
Follow Jitka finding.

Add test that verify the fix by adding an extend code that call  `SWIG_FromCharPtrAndSize` without explicit fragment.
That is allowed as we have a typemap above it which does use the fragment.

### DIFF
--- a/Examples/test-suite/python/Makefile.in
+++ b/Examples/test-suite/python/Makefile.in
@@ -117,6 +117,7 @@ C_TEST_CASES += \
 	file_test \
 	li_cstring \
 	li_cwstring \
+	python_charptr_fragment \
 	python_nondynamic \
 	python_varargs_typemap \
 	python_weakref \
--- a/Examples/test-suite/python/python_charptr_fragment_runme.py
+++ b/Examples/test-suite/python/python_charptr_fragment_runme.py
@@ -1,0 +1,8 @@
+import python_charptr_fragment
+
+if python_charptr_fragment.getStr() != 'test1':
+    raise runtimeerror("wrong getstr value")
+
+t = python_charptr_fragment.tst_str2()
+if t.getStr() != 'test2':
+    raise runtimeerror("wrong tst_str2::getstr value")
--- a/Examples/test-suite/python_charptr_fragment.i
+++ b/Examples/test-suite/python_charptr_fragment.i
@@ -1,0 +1,31 @@
+%module python_charptr_fragment
+
+%typemap(out,noblock=1,fragment="SWIG_FromCharPtrAndSize") struct tst_str {
+  %set_output(SWIG_FromCharPtrAndSize($1.str,$1.len));
+}
+
+%extend tst_str2 {
+PyObject *getStr() {
+    return SWIG_FromCharPtrAndSize("test2", 5);
+}
+}
+
+%inline %{
+
+struct tst_str {
+    size_t len;
+    char* str;
+};
+
+struct tst_str2 {
+    int dummy;
+};
+
+
+struct tst_str getStr()
+{
+    static struct tst_str r = { 5, "test1" };
+    return r;
+}
+
+%}
--- a/Lib/python/pystrings.swg
+++ b/Lib/python/pystrings.swg
@@ -73,7 +73,7 @@ SWIG_AsCharPtrAndSize(PyObject *obj, char **cptr, size_t *psize, int *alloc)
 }
 
 %fragment("SWIG_FromCharPtrAndSize","header",fragment="SWIG_pchar_descriptor") {
-#define SWIG_FromCharPtrAndSize(carray, size) SWIG_FromBinaryCharPtrAndSize(carray, size, 0)
+%#define SWIG_FromCharPtrAndSize(carray, size) SWIG_FromBinaryCharPtrAndSize(carray, size, 0)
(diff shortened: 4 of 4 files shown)


wsfulton (commented):
Thanks Erez. The `Lib/python/pystrings.swg` change is correct and I'd like it in for 4.5.0, but the test doesn't exercise it and the same defect is in several other UTL languages.

I've kicked off an AI agent to fix the shortcomings posted, as this is needed for the 4.5.0 release. Hopefully it'll be ready to post shortly.

wsfulton (commented):
## The fix is right

A bare `#define` at the start of a line in a brace-form `%fragment{...}` body is eaten by SWIG's own preprocessor and never reaches the generated file, so the macro only ever existed at SWIG level. `%#define` emits a real one. 

## The test doesn't test the fix

`%extend { }` bodies *are* preprocessed by SWIG, so the macro has always been expanded there. `python_charptr_fragment.i` compiles cleanly on unpatched master and the test passes with and without your `Lib` change.

Typemap bodies delimited with `%{ ... %}` are *not* macro expanded, which does give a failing test:

```swig
%typemap(out,noblock=1,fragment="SWIG_FromCharPtrAndSize") struct CharBuf %{
  $result = SWIG_FromCharPtrAndSize($1.str,$1.len);
%}
```

With `{ }` the wrapper gets `SWIG_FromBinaryCharPtrAndSize(..., 0)` and the bug is masked. With `%{ %}` the name survives, and as an implicit declaration is only a warning in C the module builds and then fails at load:

```
ImportError: _charptr_fragment.so: undefined symbol: SWIG_FromCharPtrAndSize
```

With your patch it imports cleanly.

## The same defect in other UTL languages

`SWIG_FromBinaryCharPtrAndSize` is lost the same way in `rubystrings.swg:47`, `perlstrings.swg:53`, `octprimtypes.swg:243` and `rfragments.swg:174`, as are `SWIG_FromUint8Array` and `SWIG_ToUint8Array` in `jsc/javascriptuint8array.swg:7,40` and `napi/javascriptuint8array.swg:44`.

Scilab is worst affected — seven files, including `SWIG_ConvertPtr` and `SWIG_NewPointerObj`. Every other language puts those in a verbatim runtime block (`pyrun.swg:18,22`, `rubyrun.swg:13,15`) and they reach the output; counting them in a generated wrapper gives python 4, ruby 3, perl5 3, tcl 3, scilab 0.

Two caveats there:

* `scipointer.swg` can't be fixed by `%#define` alone. Nothing ever requests those fragments — they exist purely to carry the SWIG-level `#define` — so converting them drops the definitions altogether and Scilab stops compiling. They need moving to a runtime insert.
* `scipointer.swg:28`, `scilonglong.swg:18` and `scilonglong.swg:45` must be left alone: they reference `$result`, which only works because SWIG-level expansion drops the text into a typemap body.

For reference, `javascriptstrings.swg` already does this correctly with `%insert(runtime) %{ ... %}`, and Scilab's `scidouble.swg`, `sciint.swg` and `scienum.swg` already use `%#define`.

## The test should run for all languages

Make it a common test guarded by `SWIGUTL` — the same languages provided `SWIG_FromCharPtrAndSize` in 4.4.1 as today, so only guile needs excluding, plus Tcl, Scilab and JavaScript if `SWIG_FromBinaryCharPtrAndSize` is covered too. Note that `%set_output` is itself a SWIG macro and so isn't expanded inside `%{ %}` either, and a hard coded `$result =` is wrong for Perl (`argvi++`), Tcl and Scilab. Keeping `%set_output` in a `{ }` typemap and routing just the macro reference through an alias defined in `%wrapper %{ %}` works for every target.

## Smaller points

* `raise runtimeerror(...)` is lowercase, so it raises `NameError` rather than `RuntimeError` — all 291 other Python runme files use `raise RuntimeError`.
* The messages say `getstr` where the symbol is `getStr`.
* The `.i` file uses 4-space indentation where the test suite is predominantly 2-space.
* A `CHANGES.current` entry is missing.

wsfulton:
Sorted in de573cda78890318e570611066cf4e94f4066a6e. Let me know if you see any oversights/issues.

erezgeva:
> Sorted in [de573cd](https://github.com/swig/swig/commit/de573cda78890318e570611066cf4e94f4066a6e). Let me know if you see any oversights/issues.

Do you want to add more tests for the other changed macros?

wsfulton:
> Do you want to add more tests for the other changed macros?

I think they're covered, not sure which ones you mean, or do you mean SWIG_ToUint8Array and SWIG_FromUint8Array? I'm not sure it's worth spending too much effort on all these, but on whole more tests are generally useful though.

erezgeva:
> > Do you want to add more tests for the other changed macros?
> 
> I think they're covered, not sure which ones you mean, or do you mean SWIG_ToUint8Array and SWIG_FromUint8Array? I'm not sure it's worth spending too much effort on all these, but on whole more tests are generally useful though.

Wrote one for Javascript with SWIG_ToUint8Array and SWIG_FromUint8Array.
I will PR it, and finish the saga :smile: 

Thanks for your feedback and assistance.

PR #3529, opened 2026-08-07
[Python] Show overloads in type annotations

### LINKED ISSUE #3470: [Python] PEP 484: Show overloads
If a function is overloaded, Swig doesn't add type annotations to the function and collapses the arguments into `*args`. With [PEP 484](https://peps.python.org/pep-0484/), overloads can be annotated with [`@typing.overload`](https://docs.python.org/3/library/typing.html#overload):
```python
@typing.overload
def overloaded(arg: "int") -> "None":
    ...
@typing.overload
def overloaded(arg: "str") -> "int":
    ...
def overloaded(*args):
    """Docstring goes here"""
    return _impl.overloaded(*args)
```

Some notes for implementing this:
- Default arguments also create overloads. There should be one declaration for non-overloaded functions with default arguments. For an overload with default arguments, there should only be one declaration with `@typing.overload`.
- Overloads have to be emitted from the `*Handler()` methods when called for the specific overload, because some attributes are only correct there (`Language` uses `Swig_require`/`Swig_restore`).
- [PEP 570](https://peps.python.org/pep-0570/) (Python 3.8) added a marker for positional-only arguments - anything before `/` is positional-only. It might be nice if the overloads used this. In the first example, we'd show `overloaded(arg: "int", /)`.

Nerixyz:
This PR adds `@typing.overload` annotations to overloads. It works with both the C and pytyping modes. Swig can honor [PEP 570](https://peps.python.org/pep-0570/) when `python:annotations:positional-only-overloads` is enabled (I didn't have a better name, happy for suggestions).

Before this PR, overloads would only have a return type annotation:

```swig
// test.i
short overloaded(short a);
short overloaded(short a, short b);
```
```python
# test.pyi
def overloaded(*args) -> "int":
    ...
```

With this PR, the annotations look like this:

```python
# test.pyi
@typing.overload
def overloaded(a: "int") -> "int":
    ...
@typing.overload
def overloaded(a: "int", b: "int") -> "int":
    ...
```

The declarations are generated in the respective `{staticmember,member,constructor,function}handler` methods.

Closes #3470.

### DIFF
--- a/Examples/test-suite/python/python_pyi_overloads_positional_runme.py
+++ b/Examples/test-suite/python/python_pyi_overloads_positional_runme.py
@@ -1,0 +1,33 @@
+import re
+
+from swig_test_utils import swig_check, swig_assert, swig_assert_raises
+import python_pyi_overloads_positional
+
+# The functions should work regardless of Python version.
+swig_check(python_pyi_overloads_positional.overloaded(), 42)
+swig_check(python_pyi_overloads_positional.overloaded(1), 2)
+swig_check(python_pyi_overloads_positional.overloaded(1, 2), 3)
+swig_check(python_pyi_overloads_positional.not_overloaded(1, 2), 3)
+
+with open("python_pyi_overloads_positional.pyi") as f:
+    lines = f.read().splitlines()
+
+idx = lines.index('@typing.overload')
+swig_assert(idx >= 0)
+swig_check(lines[idx], "@typing.overload")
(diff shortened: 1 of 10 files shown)


Examples/test-suite/python/python_pyi_overloads_positional_runme.py
Nerixyz: This does string parsing, because Python's `ast` <3.8 won't be able to parse the code otherwise.

Issue #3527, opened 2026-08-04
[Python] PEP 484: Fix return type when argout typemaps are used for class methods


degasus:
Hi, I'm currently trying to reproduce #3469, but I'm failing even harder to get the pep484 annotations for arg out of class-methods:
```
%typemap(in, numinputs=0) (short *short_list) { /* ... */ }
%typemap(in, numinputs=0) (size_t short_list_len) { /* ... */ }
%typemap(freearg) (short *short_list) { /* ... */ }
%typemap(argout) (short *short_list, size_t short_list_len) { /* drop result and create list */ }
%typemap(pytyping) (short *short_list, size_t *short_list_len) "typing.List[int]"

size_t function_get_list_argout(short *short_list, size_t short_list_len);

class C {
public:
    size_t method_get_list_argout(short *short_list, size_t short_list_len);
};
```

This yields
```
# As reported in the other issue
def function_get_list_argout() -> "typing.Optional[SWIGTYPE_p_short]":
    return _test.function_get_list_argout()

class C(object):
    # Even worse for methods
    def method_get_list_argout(self) -> "int":
        return _test.C_method_get_list_argout(self)
```

FYI @Nerixyz

Nerixyz:
> Hi, I'm currently trying to reproduce [#3469](https://github.com/swig/swig/issues/3469), but I'm failing even harder to get the pep484 annotations for arg out of class-methods:

Sorry, I had a typo in the `pytyping` typemap, it should've matched `(short *short_list, size_t short_list_len)`, not `(short *short_list, size_t *short_list_len)`  (notice the pointer on the second argument. I updated the other issue description.

With 5b13fbf23e050c239a61f9c74bb3d6b2c5cf3f8b, there is now support for multi-argument argout typemaps. The `/* drop result and create list */` use of argout typemaps is not supported with this - that would require #3516 or something similar. If you pull from master and rerun swig, you should now get:
```python
def function_get_list_argout() -> "typing.List[typing.Union[int, typing.Any]]":
    return _test.function_get_list_argout()
class C(object):
    def method_get_list_argout(self) -> "typing.List[typing.Union[int, typing.Any]]":
        return _test.C_method_get_list_argout(self)
```

degasus:
No need to be sorry, thanks a lot for your effort on python annotations!

It wasn't just this typo. The "int" as annotation was a different bug, but I can confirm that this bug is fixed on current master.

Thanks for your update.

Issue #3084, opened 2024-12-06
Regression: generated Python functions return a list instead of a single value


hvbtup:
I use swig 4.3.0 to generate Python wrappers for a C library.

For example, one of the C functions is defined as

    EXPORT PTError API PBMGetLevel(BookmarkHandle h, int* level);

PTError is defined as

    typedef enum PTError {
      PTSuccess = 0,
      PTNotPDF = 1001, ...
    } PTError;

And I am using the following in the *.i file:

    void create_pdf_exception(const char* api, PTError rc);

    %typemap(in,numinputs=0,noblock=1) PTError* err (PTError temp) {
        temp = 0;
        $1 = &temp;
    }

    %typemap(in,numinputs=0,noblock=1) PTError* pErr (PTError temp) {
        temp = 0;
        $1 = &temp;
    }

    %typemap(out) PTError {
        if ($1 == 0 || $1 == PTAlreadyWritten) {
            $result = SWIG_Py_Void();
        } else {
            create_pdf_exception("$symname", $1);
            goto fail;
        }
    }

    %typemap(argout) (PTError* err) {
        if (*$1 == 0 || *$1 == PTAlreadyWritten) {
            ;
        } else {
            Py_XDECREF(resultobj);
            create_pdf_exception("$symname", *$1);
            goto fail;
        }
    }

    %typemap(argout) (PTError* pErr) {
        if (*$1 == 0 || *$1 == PTAlreadyWritten) {
            ;
        } else {
            Py_XDECREF(resultobj);
            create_pdf_exception("$symname", *$1);
            goto fail;
        }
    }

With SWIG 4.1.0, calling the function `PBMGetLevel(bmh)` in Python returned a single integer `-1`.
With SWIG 4.3.0, this now returns a list with two elements: `[None, -1]`.

The same happens for all functions in the C lib:
Instead of just the expected, they return a list containing None and the expected value.

**I think that this issue is a consequence of the fix for #2905.**

Comparing the C code generated with SWIG 4.1.0 and SWIG 4.3.0, it looks like this:

    if (SWIG_IsTmpObj(res2)) {
      resultobj = SWIG_Python_AppendOutput(resultobj, SWIG_From_int((*arg2)), 0);
    } else {
      int new_flags = SWIG_IsNewObj(res2) ? (SWIG_POINTER_OWN |  0 ) :  0 ;
[...]

PR #3532, opened 2026-08-10
Let argout and out typemaps say what they return

wsfulton:
## Summary

Adds four optional typemap attributes so that a typemap can state what it returns to the
target language, instead of SWIG guessing:

| Typemap  | Attribute    | Meaning                                                          |
| -------- | ------------ | ---------------------------------------------------------------- |
| `out`    | `numoutputs` | `0` means the function return value is not returned at all        |
| `argout` | `numoutputs` | `0` means this typemap contributes no returned value             |
| `argout` | `overwrite`  | `1` means this typemap replaces `$result` rather than appending  |
| `argout` | `container`  | what the typemap appends into, or builds: `"list"` (default) or `"tuple"` |

`numoutputs` accepts `0` or `1` only; any other value is an error.

Plus the supporting machinery:

* a new `$isvoidresult` special variable, which is `$isvoid` extended to also be true for a
  return value that an `out` typemap has suppressed. `$isvoid` itself is unchanged and still
  describes the C++ return type;
* one shared helper substituting both special variables in every target language. Along the
  way this fixes Lua, whose director method emitter substituted neither, leaving `$isvoid` in
  the generated C++ where it does not compile;
* a language neutral `emit_output_summary()` that works out what a wrapped function returns,
  and a `Language::composeOutputType()` virtual for target languages to turn that into a type;
* new warning 478 when the `argout` typemaps matched for one function disagree about the
  container;
* stricter validation of `numinputs` and `numoutputs` attribute values.

Python is the first consumer: PEP 484 annotations and `.pyi` stubs now describe what a
function actually returns.

### Appending is a convention, not a requirement

SWIG has always documented multiple return values loosely. The Typemaps chapter says the
extra return values are *"often appended to return value of the function"* - often, not
always. Nothing in the code enforces it, and the library itself ships `argout` typemaps that
do not append, for example the array typemaps in `Lib/java/arrays_java.i` which write results
back into the caller's array and return nothing of their own.

In practice `argout` typemaps do one of three things:

1. append a value to `$result` - the `typemaps.i` `OUTPUT` family;
2. return nothing, for example checking an error code and raising an exception;
3. discard `$result` and replace it with something else.

SWIG cannot tell which of these a typemap does by looking at its code - the body is opaque
target language source. Until now it assumed every `argout` typemap does (1) and contributes
exactly one value. That assumption is wrong often enough to produce visibly incorrect output,
and there was no way for a typemap author to correct it.

### Why this matters now

While the assumption only affected documentation-like output it was tolerable. Since SWIG
started generating PEP 484 annotations and `.pyi` stubs it is a correctness problem: the
annotations are consumed by type checkers and IDEs, so a wrong annotation is worse than no
annotation. A stub claiming `typing.List[typing.Union[bool, typing.List[int]]]` for a function
that returns `typing.List[int]` will make a type checker reject correct user code.

The `out` typemap case is worse still, because there the wrong assumption changes what the
wrapper actually returns at runtime, not just how it is described.

## Related issues and prior work

Only #3469 is directly fixed by this work, and #3084 gains a supported way to get the
behaviour it asks for. The rest is context that shaped the design.

| Issue | Reporter | What it raised | Relationship to this work |
| ----- | -------- | -------------- | ------------------------- |
| [#3469](https://github.com/swig/swig/issues/3469) | Nerixyz | A multi-argument `argout` typemap that drops the result and builds a list is annotated `typing.List[typing.Union[int, typing.List[int]]]` instead of `typing.List[int]`. Asked for a way to say the typemap overwrites the return value. | **Fixed.** `overwrite=1` on the `argout` typemap. The reported annotation string is a regression test that fails byte for byte if the attribute is ignored. |
| [#3084](https://github.com/swig/swig/issues/3084) | hvbtup | Regression from #2905: a function using an `out` typemap that turns an error code into an exception started returning `[None, -1]` instead of `-1`. The workaround needed an `out` typemap pushing a placeholder plus a `ret` typemap unpicking it. | **Opt-in fix for Python.** `numoutputs=0` on the `out` typemap says the return value is not returned, and `$isvoidresult` tells `SWIG_AppendOutput` there is nothing to append to. The placeholder dance is no longer needed. Existing interfaces are unaffected until they add the attribute. |
| [#684](https://github.com/swig/swig/issues/684) | jroose | An `out` typemap and an `argout` typemap on the same function appearing to conflict. Closed as not reproducible, but the suggested solutions were `typemap(argout,numoutputs=0)` *by name*, and returning a tuple. | **Prior art for the syntax.** The attribute spelling and the tuple container both come from this suggestion. The attributes describe what an `argout` body does; they do not stop a body overwriting `$result` or make it build a tuple. |
| [#2165](https://github.com/swig/swig/issues/2165) | krishauser | Output types missing from Python type annotations for `argout` typemaps, blocking a type hint generation tool. | **Historical context.** The `pytyping` typemap already restored the types; this work makes the combination of them accurate. |
| [#2905](https://github.com/swig/swig/issues/2905) | vadz | Fixing `SWIG_Python_AppendOutput` to stop discarding a null return introduced the `is_void` argument, whose meaning is "is the C++ return type void". | **Predecessor behaviour.** `$isvoidresult` generalises that argument to the question the `argout` typemaps actually need answered: does the wrapper have a return value to append to? |
| [#3527](https://github.com/swig/swig/issues/3527) | degasus | Wrong PEP 484 return types when `argout` typemaps are used for class methods, called out as more severe than #3469. | **Already fixed separately.** This series adds the related constructor case: `__init__` no longer gets a return annotation built from its `argout` typemaps. |

### Attributes on the typemap, not a feature on the function

The information belongs to the typemap, because it is a property of the code the typemap
contains. A typemap that builds a tuple builds a tuple wherever it is applied. Putting the
information on the typemap means it is written once, next to the code it describes, and
travels automatically through `%apply`. A `%feature` attached to each wrapped function would
have to be repeated for every function using the typemap and could drift out of step with it.

### Language neutral core

`emit_output_summary()` in `Source/Modules/emit.cxx` walks the parameters of a function and
works out what it returns, recording on the node:

* `wrap:returnsurvives` - the function return value is one of the returned values;
* `wrap:outputparms` - the parameters contributing a returned value, in order;
* `wrap:outputcontainer` - the container they are returned in;
* `wrap:outputcontainerbuilt` - the container is built by an overwriting typemap rather than
  appended into, so a single value is in it too.

This contains no Python. A target language turns the result into a type by overriding
`Language::composeOutputType(Node *n, List *types)`. Python composes
`typing.List[typing.Union[...]]` or `typing.Tuple[...]`; a language whose `SWIG_AppendOutput`
equivalent builds something else composes whatever suits it. The base implementation handles
a single returned value, which is all a language without such support needs.

The shared part is the summary and the hook, not a finished integration: a second target
language still has to collect its own per-parameter type strings, call the summary, implement
`composeOutputType()` and wire the result into its own output. What it does not have to do is
design another set of attributes, or ask interface authors to restate the same facts.

The `container` attribute is deliberately semantic - `"list"` or `"tuple"` - rather than a
literal fragment of target language syntax. Each backend maps the concept onto its own
notation: `typing.List` / `typing.Tuple` for Python, `Array` / a fixed size tuple for Ruby
RBS, `array{...}` / `list<...>` for PHPStan, `[]` / a tuple type for TypeScript.

### Warning 478

All the `argout` typemaps appending into one `$result` must agree about what it is, so if the
matched typemaps declare different containers the type of the values returned cannot be worked
out and warning 478 is issued. Only the typemaps contributing values that survive matter: an
`overwrite=1` typemap discards everything before it, so agreement is only required from that
point on.

The warning is raised from `emit_output_summary()`, which today is only reached when Python is
generating annotations - that is, with `-typehints` or the `python:annotations` feature. Other
backends do not call the summary yet, so they do not raise it.

It also fires in the legacy C annotation mode, where the annotation is the C parameter types
concatenated and no container is involved, so there the message's mention of a catch-all type
does not describe what happens. Restricting it to the modes that use a container is left as
follow-up.

### Attribute validation

`numoutputs` must be `0` or `1`; anything else is an error. `numinputs` was previously parsed
leniently with `atoi()`, which takes any leading digits and ignores the rest, so `"abc"` was
read as 0 and silently dropped the parameter from the wrapped function while `"1x"` was read
as 1. It is now rejected in the same way. Both have errors test-suite cases. See Compatibility
below.

## Why this rather than #3516

[#3516](https://github.com/swig/swig/pull/3516) by Nerixyz targets the same problem and was
the starting point for this work. The disagreement is about where the information lives and
what shape it takes, not about whether the problem is real - it is, and Nerixyz identified it
precisely.

#3516 puts four attributes on the Python `pytyping` typemap:

```
%typemap(pytyping, argoutaction="overwrite") short *OutReplace "int"
%typemap(pytyping, argoutprefix="typing.Tuple[", argoutjoiner=",", argoutsuffix="]") double *OutTuple "float"
```

Reasons for preferring the approach in this merge request:

1. **The information describes the `argout` typemap, so it belongs on the `argout` typemap.**
   Whether a typemap appends, returns nothing or replaces `$result` is a fact about the C code
   in the `argout` typemap. In #3516 that fact is recorded on a different typemap - `pytyping`
   - which exists only to supply a Python type name. The two can be applied to different type
   patterns and can be edited independently, so they can disagree. Putting `overwrite=1` on the
   `argout` typemap keeps the statement and the code it describes in one place.

2. **It is not Python specific.** `argoutaction` and friends are attributes of a Python only
   typemap handled in `python.cxx`, so Ruby, PHP and JavaScript would each need to invent their
   own equivalent on their own typemap, and every `.i` file would have to repeat the same fact
   once per target language. Here the fact is stated once on the `argout` typemap and read by a
   language neutral function.

3. **`container="tuple"` says what the typemap does; `argoutprefix`/`argoutjoiner`/
   `argoutsuffix` say how to write a Python type.** The prefix/joiner/suffix form stores
   literal target language syntax in the interface file. `"typing.Tuple["` means nothing to a
   Ruby RBS or TypeScript backend reading the same `.i` file, so the formatting would have to
   be duplicated once per target language. A semantic attribute leaves each backend free to
   render the concept in its own notation, including languages where a tuple is not written
   with brackets at all. It is also one attribute with one consistency rule rather than three
   strings that must be kept in step across every typemap contributing to a function - which
   is why #3516 needs three separate mismatch checks.

4. **It covers the `out` typemap too.** #3516 does not address #3084: a function whose return
   value is turned into an exception still contributes a phantom value. Because the same
   `emit_output_summary()` reads `numoutputs` from both the `out` and `argout` typemaps, one
   mechanism covers both, and `$isvoidresult` makes the runtime behaviour match the annotation.

5. **Mismatch handling belongs in the shared layer.** #3516 adds Python warning 742 and
   carries on with whichever conflicting value was seen first, so the annotation is still
   emitted as a definite type that may be wrong. Warning 478 is issued from the language
   neutral layer, and the annotation degrades to a catch-all type rather than to a confidently
   wrong one.

6. **Attributes are inert where unsupported.** The three `argout` attributes never change
   generated wrapper code, so adding them to a typemap is harmless for a target language that
   does not yet read them.

The test cases from #3469 are incorporated into `python_annotations_typing`, credited in the
commit that carries them.

## Testing

```
$ make -j16
$ make -j16 check-errors-test-suite
$ make format-check
$ make -C Doc/Manual maketoc check
$ .agents/skills/swig-test/test-configured-languages.sh -j 8 -x c typemap_isvoid.cpptest
$ make -j16 -k check
$ SWIG_FEATURES=-typehints        make -j16 -k check-python-test-suite
$ SWIG_FEATURES="-pyi -typehints" make -j16 -k check-python-test-suite
```

Github Actions have been run.

Coverage added:

* `Examples/test-suite/python_annotations_typing.i` - appending, error checking, overwriting
  and container building `argout` typemaps, single and multi-argument, with void and non void
  return types, and suppressed return values with and without `argout` parameters.
* `Examples/test-suite/python_typemap_numoutputs.i` - runtime behaviour of `numoutputs=0` on
  the `out` typemap, including a user defined `void` `out` typemap.
* `Examples/test-suite/typemap_isvoid.i` - `$isvoidresult` expansion, in every target
  language: a wrong substitution order leaves `ghi_1result()` in the wrapper and fails to
  compile.
* `Examples/test-suite/errors/swig_argout_container.i` - warning 478.
* `Examples/test-suite/errors/swig_typemap_numoutputs.i`,
  `Examples/test-suite/errors/swig_typemap_numinputs.i` - attribute validation.

The symptoms reported in #3469 and #3084 are asserted as exact annotation strings and exact
runtime values, so they fail if the handling regresses.

## Documentation

* `Doc/Manual/Typemaps.html` - the new `out` and `argout` attributes, `$isvoidresult`, what
  `container` means on an overwriting typemap, and notes on when the attributes are needed.
* `Doc/Manual/Python.html` - new section "Annotations for argout typemaps" with worked
  examples.
* `Doc/Manual/Warnings.html` - warning 478.
* `CHANGES.current`.

## Compatibility

`numoutputs` defaults to 1 and `container` to `"list"`, so a typemap that does not use the new
attributes behaves exactly as before. Annotations are unchanged for such typemaps.

Two behaviour changes without an opt-in:

* **`numinputs` parsing is stricter.** A value that is not a non-negative integer is now an
  error rather than being run through `atoi()`. An interface relying on the old parsing was
  either silently dropping a parameter from the wrapped function or getting a count it did not
  write, so this turns a silently wrong wrapper into a build failure - but it is a build
  failure that did not happen before.
* **`__init__` no longer receives a return annotation** when a constructor has a parameter
  using an `argout` typemap. Every other constructor SWIG generates is annotated this way, so
  this removes an inconsistency rather than introducing one. Note that omitting the annotation
  is not the same as writing `-> None`: type checkers special case `__init__` and infer `None`
  when any parameter is annotated, which is the behaviour relied on here.

## Follow-up

* `composeOutputType()` implementations for Ruby RBS, PHPStan/Psalm and TypeScript.
* Tightening the pyrefly flags used by the test-suite: the default `basic` preset does not
  catch malformed annotations, which is why an early bug in this work had to be found by
  diffing generated output rather than by the checker.
* Warning 478 fires in the legacy C annotation mode, where its wording does not apply. It
  should either be restricted to the modes that use a container or be worded to cover both.


Nerixyz:
> ### Warning 478
> 
> All the `argout` typemaps appending into one `$result` must agree about what it is, so if the matched typemaps declare different containers the type of the values returned cannot be worked out and warning 478 is issued.

I also thought about this in my PR where I have a similar warning. Technically, we can resolve the return type. For example, if you have a `int foo(short *OutTupleAppend, bool *OUTPUT)`, then the output will be `list[union[tuple[int, int], bool]]`. So the warning is more about informing the user that they may be doing something with unexpected results.

> 2\. **It is not Python specific.** `argoutaction` and friends are attributes of a Python only
>         typemap handled in `python.cxx`, so Ruby, PHP and JavaScript would each need to invent their
>         own equivalent on their own typemap, and every `.i` file would have to repeat the same fact
>         once per target language.

I agree with this. It's better to be handled in a neutral way.

Doc/Manual/Python.html
```diff
@@ -7276,7 +7218,196 @@ <H4><a name="Python_annotations_pytyping">33.12.1.2 PEP 484 annotation types</a>
...
+the list that <tt>SWIG_AppendOutput</tt> builds:
+</p>
+
+<div class="code">
+<pre>
+%typemap(in, numinputs=0) short *OutValue (short temp) { $1 = &amp;temp; }
+%typemap(argout) short *OutValue {
+  $result = SWIG_AppendOutput($result, PyLong_FromLong(*$1));
+}
+%typemap(pytyping) short *OutValue "int"
```
Nerixyz, line 7238: It might be worth mentioning that the return value will prefer the `out` attribute if it exists. In these examples, it doesn't make a difference, because the `in` typemap always sets `numinputs=0`.

I don't know a good example. Maybe a function like `bool readStr(char *buf, size_t bufLen)` that's wrapped as `def readStr(buf: "int") -> "str"`? So you'd have a `%typemap(pytyping, out="str") (char *buf, size_t bufLen) "int"`.

wsfulton:
> if you have a `int foo(short *OutTupleAppend, bool *OUTPUT)`, then the output will be
> `list[union[tuple[int, int], bool]]`

Thanks - you are right that the values are not untyped; that describes what `foo` returns exactly.

What kept it a warning is that the type depends on the order the `argout` typemaps run in, not on
the set of containers involved. Swapping those two parameters returns `([7, True], 1)` rather than
`[(7, 1), True]`, that is `Tuple[List[Union[int, bool]], int]`, and `overwrite=1` nests differently
again. Working it out needs a nesting built up in typemap order rather than a union over the
returned values.

Three commits from your comments:

1. 064ef86187 Test cases pinning the annotation and the returned values for both orderings.

2. 1f0ab388c1 A way to supply the type where the mixture is deliberate. The new
  `python:annotations:catchall` feature gives a named function its annotation in place of
  `typing.Any`, and supplying it also suppresses the warning:

```swig
%feature("python:annotations:catchall") tuple_then_list "typing.List[typing.Union[typing.Tuple[bool, int], int]]"
```

  It supplies a PEP 484 type, so it does nothing for the C/C++ annotation types, where 478 is
  still issued.

3. b47093524a The `out` attribute being preferred for the returned value, shown on the `pytyping`
  typemap already in that section with the attribute and the body deliberately disagreeing:

```swig
%typemap(pytyping, out="int") short *OutValue "bool"
```

  As you say it makes no difference here, and the docs now say why: `numinputs=0` means the
  parameter is never annotated, so the body is never used.

Also 71c6dc62f1, dropping the parameter list copy in `collectOutputTypes()` -
`Swig_typemap_attach_parms()` only sets attributes named after the typemap method being attached,
and every generated file in the Python test-suite is unchanged by it - and 3c22ac7d61, a test for a
`container` naming something other than a list or a tuple.

PR #3534, opened 2026-08-12
Fix leftover of convertpath removal

larskanis:
Tools/convertpath was removed in favour of cygpath in commit:
  https://github.com/swig/swig/commit/f8d198bf675bfec02680335e7d7aa375372d5831

But there was one leftover which shows this error while running configure:
```
  ../swig-4.5.0/configure: line 15476: ../swig-4.5.0/Tools/convertpath: No such file or directory
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2710,8 +2710,7 @@ AC_DEFINE_DIR(SWIG_LIB, swig_lib, [Directory for SWIG system-independent librari
 
 case $build in
   # Windows does not understand unix directories. Convert into a windows directory with drive letter.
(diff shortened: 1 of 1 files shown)


erezgeva:
Hi,

I did not miss it.
I plan to remove `SWIG_LIB_WIN_UNIX`.
See #3358

larskanis:
IMHO #3358 is a candidate for swig-4.6. But the above PR fixes a bug in swig-4.5.0 and it should be fixed in swig-4.5.1.

Issue #3535, opened 2026-08-15
[python] SIGSEGV: when a type is first queried from a second live sub-interpreter


reardonia:
Reproduced on 4.4.1, 4.5.0, master

`SWIG_runtime_data_module()` caches a per-interpreter module object in a process-wide static. The second interpreter therefore attaches its capsule to the first interpreter's module, so `PyCapsule_Import` fails there, `SWIG_GetModule()` returns NULL, and `SWIG_Python_TypeQuery()` dereferences it without a check.

Any type query from the second interpreter crashes, including `std::vector<Foo*>` conversion via `swig::traits_info<T>::type_info()`. 

Note that there is a claim that this was fixed (sourceforge SWIG bug 1114: "Py_EndInterpreter invalidates SWIG modules in other sub-interpreters") but it's not. Somehow someone conflated sequential restart (init, finalize, init again) with concurrence, ie two sub-interpreters alive at the same time.

## Reproducer

`example.i`
```swig
%module example
%inline %{
bool moduleVisible() { return SWIG_GetModule(0) != 0; }
bool typeQuery()     { return SWIG_TypeQuery("int *") != 0; }
%}
```

`main.cpp`
```cpp
#include <Python.h>
extern "C" PyObject* PyInit_example(void);

int main()
{
  PyImport_AppendInittab("example", PyInit_example);
  Py_Initialize();
  PyThreadState* main_ts = PyThreadState_Get();

  Py_NewInterpreter();                     /* A, kept alive */
  PyRun_SimpleString("import example\n"
    "print('A: moduleVisible =', example.moduleVisible(), flush=True)\n");
  PyThreadState_Swap(main_ts);

  Py_NewInterpreter();                     /* B, while A is still alive */
  PyRun_SimpleString("import example\n"
    "print('B: moduleVisible =', example.moduleVisible(), flush=True)\n"
    "example.typeQuery()\n");
  return 0;
}
```

```bash
swig -python -c++ -builtin -interface example -o example_wrap.cxx example.i
g++ -std=c++17 $(python3 -c "import sysconfig;print('-I'+sysconfig.get_paths()['include'])") \
    main.cpp example_wrap.cxx -o repro -lpython3.13
```

Actual:
```
A: moduleVisible = True
B: moduleVisible = False
Segmentation fault      (SWIG_MangledTypeQueryModule <- SWIG_TypeQueryModule <- SWIG_Python_TypeQuery)
```

Expected: `B: moduleVisible = True`, no crash.

[...]

PR #3539, opened 2026-08-16
Fix null module dereference in the type query functions

reardonia:
Fix null module dereference in the type query functions

SWIG_GetModule can return null and two runtime functions dereference it unchecked. Reproduced on master, single interpreter: generate a header with swig -python -external-runtime swigpyrun.h, then call SWIG_TypeQuery before importing any wrapper module.

```
AddressSanitizer: SEGV on unknown address 0x000000000008 (READ)
    #0 SWIG_MangledTypeQueryModule
    #1 SWIG_TypeQueryModule
    #2 SWIG_TypeQuery
```
0x8 is the size member of a null swig_module_info; both functions read iter->size before testing anything. Null is a designed return: the Python runtime returns it when PyCapsule_Import fails, and the SWIG_LINK_RUNTIME stub in Lib/linkruntime.c returns it unconditionally.

The guard goes in swigrun.swg rather than at the Python call site because SWIG_Python_TypeQuery is not the only unchecked caller: Lib/runtime.swg has four more, and SWIG_Lua_init_base_class and several Octave sites do the same. Four lines cover all of them. 0 is what both functions already return for an unknown name.

Not a fix for #3535.

### DIFF
--- a/Lib/swigrun.swg
+++ b/Lib/swigrun.swg
@@ -511,6 +511,8 @@ SWIG_MangledTypeQueryModule(swig_module_info *start,
                             swig_module_info *end,
 		            const char *name) {
   swig_module_info *iter = start;
+  if (!iter)
+    return 0;
   do {
     if (iter->size) {
       size_t l = 0;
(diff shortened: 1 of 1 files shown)


reardonia:
@wsfulton apologies if I missed a PR formatting requirement or other submission rule.  Is there anything I should do to further to progress this PR?  Is there a separate mailing list I should correspond with?

PR #3547, opened 2026-08-25
[Ruby] Fix -autorename dropping std::vector partial specializations

### LINKED ISSUE #3546: -autorename (Ruby) collides std::vector partial specializations since 4.5.0, silently dropping vector<bool>
### Regression

Since SWIG 4.5.0, using `-ruby -autorename` together with `%include <std_vector.i>` causes the `bool`, `T*` and `T const*` partial specializations of `std::vector` to be silently dropped, because they all camelcase to the same default name `Vector` as the primary template and lose to it as duplicate identifiers (`Warning 302: Redefinition of identifier 'Vector' ... ignored`).

Concretely, `%template(VectorBool) vector<bool>;` compiles without error but produces **no wrapper at all** for `VectorBool` under `-autorename`, since `vector<bool, _Alloc>`'s class definition was already dropped earlier while parsing `std_vector.i`.

### Repro
```swig
%module(directors="1") test
%include <std_vector.i>
namespace std {
  %template(VectorBool) vector<bool>;
}
```
```
swig -ruby -autorename -c++ test.i
```
Output includes:
```
Warning 302: Redefinition of identifier 'Vector' (Renamed from 'vector') as std::vector< bool,_Alloc > ignored,
Warning 302: previous definition of 'Vector' (Renamed from 'vector') as std::vector.
```
and the generated wrapper defines no `VectorBool`/`Vector` methods for the bool specialization.

### Bisected to
`ce06e2fc13a3cac4fe1aa698ba8889f38c501ee7` — "Advanced %rename fixes for templates" (first in 4.5.0).

That commit changed `apply_rename()` in `Source/Swig/naming.c` to strip template arguments from `name` before substituting it into a `%s`-style rename format:
```c
String *template_name = SwigType_istemplate_templateprefix(name);
String *name_simple = template_name ? template_name : name;
...
result = NewStringf(fmt, name_simple);
```
Ruby's `-autorename` applies `%rename("%(camelcase)s", %$isclass) "";` globally, which is exactly this code path. Before the change, `%s` received the full templated name (distinguishing `vector<bool,...>` from `vector<T*,...>` etc.), so each specialization got a distinct default identifier. Now they all reduce to the same base name `vector` → `Vector`, and only the first-parsed one survives.

### Impact / workaround
Discovered via SimpleITK's Ruby wrapping build (upgrading SWIG 4.4.0 → 4.5.0 broke `VectorBool`, `NameError: uninitialized constant`). Worked around it there by explicitly `%rename`-ing the specializations before `%include <std_vector.i>` so autorename no longer needs to invent colliding names:
- Fix: https://github.com/SimpleITK/SimpleITK/pull/2677
- Backport: https://github.com/SimpleITK/SimpleITK/pull/2675

Filing upstream since this affects any `-autorename` Ruby user of `std_vector.i` (or presumably other stdlib containers with partial specializations), not just SimpleITK.

blowekamp:
Fixes #3546

Since SWIG 4.5.0 (ce06e2fc13a3cac4fe1aa698ba8889f38c501ee7 - "Advanced %rename fixes for templates"), `apply_rename()` in `Source/Swig/naming.c` strips template parameters from a node's name before substituting it into a `%s`-style rename format. Ruby's `-autorename` applies exactly this via `%rename("%(camelcase)s", %$isclass) "";`, so `std::vector`'s primary template and its `bool`/`T*`/`T const*` partial specializations - which only differ by template argument - all reduce to the same default name `Vector` under `-autorename`. Only the first survives; the rest are dropped as duplicate identifiers (`Warning 302`).

This silently broke a very common pattern:
```swig
%include <std_vector.i>
namespace std {
  %template(VectorBool) vector<bool>;
}
```
under `-autorename`: it compiles without error but produces **no wrapper at all**, since `std::vector<bool>`'s partial specialization had already been dropped while parsing `std_vector.i`.

### Fix
Restrict the template-argument stripping to non-class nodes. The feature's own test suite (`template_advanced_rename.i`, `template_methods.i`) only exercises function templates, so classes never relied on the new stripping behavior, and a class template's arguments are exactly what distinguish otherwise identically-named specializations.

### Test
Added a regression test to `ruby_naming.i`/`ruby_naming_runme.rb`, the only existing test in the suite that compiles with `-autorename`. Confirmed it fails (compile error under `-Werror`, due to `Warning 302`) without the fix, and passes with it.

### Validation
- Full `make check-ruby-test-suite`: 766/766 tests pass.
- Confirmed the original motivating case (`Quirky::funky<int>` -> `funky_void`, from `template_advanced_rename.i`) still works correctly.
- No other bundled target-language library uses this `%s` + `%$isclass` combination, so the behavioral change is effectively scoped to Ruby's `-autorename`.

Discovered via SimpleITK's Ruby wrapping build breaking after upgrading SWIG 4.4.0 -> 4.5.0 (`NameError: uninitialized constant Simpleitk::VectorBool`).


wsfulton:
Alas, this breaks other usage of advance %rename. So I'm closing in favour of an alternative I've put in. Please take a look at 03ec714c0b67cd45322354189c23e115812aeede...5872e9b4176cf59ce4e31b010c29a4cebf32960f which includes additional testing for missing advanced renames of classes. I've included your test from this merge request.

PR #3551, opened 2026-08-27
[Python] Add -abi3t flag: support PEP 803 free-threaded stable ABI

### LINKED ISSUE #3550: Add support for abi3t (PEP 803): the free-threaded-compatible Stable ABI for Python 3.15
## Summary

Python 3.15 introduces **[PEP 803](https://peps.python.org/pep-0803/) — "`abi3t`": Stable ABI for Free-Threaded Builds** (Status: Final; requires [PEP 703](https://peps.python.org/pep-0703/) free-threading, [PEP 697](https://peps.python.org/pep-0697/) opaque per-instance data, and [PEP 793](https://peps.python.org/pep-0793/) module export hooks). It extends the existing Stable ABI (`abi3`, [PEP 384](https://peps.python.org/pep-0384/)/[PEP 652](https://peps.python.org/pep-0652/)) so that a single compiled extension can be loaded on **both** a GIL-enabled build **and** a free-threaded (no-GIL) build of CPython 3.15+, by making `PyObject`/`PyVarObject` fully opaque.

SWIG currently has no way to generate `abi3t`-compatible wrapper code. This is a natural follow-up to #3121 (which added `-nogil` support so generated modules can *declare* thread-safety at runtime under a free-threaded build) and to SWIG's existing Limited API / `abi3` support — `abi3t` is the next ABI tier up, and per the PEP it is explicitly designed to be a superset compatibility mode, not a separate track requiring separate wheels.

**Key point extension authors (and SWIG, as a code generator) need to design for:** a module built for `abi3t` is loadable by *both* interpreter flavors — it is not a "free-threaded only" artifact.

> "Extensions built for `abi3t` 3.x will be compatible with free-threading builds of CPython 3.x and above... Any extension compiled for `abi3t` will, in practice, be compatible with `abi3` as well." — PEP 803

Concretely, on Linux/macOS a single `foo.abi3t.so` is loaded by *both* `python3.15` and `python3.15t`; a wheel that supports both is tagged with the **compressed tag `abi3.abi3t`** (e.g. `myproject-1.0-cp315-abi3.abi3t-linux_x86_64.whl`), per the [abi3t migration guide](https://docs.python.org/3.15/howto/abi3t-migration.html).

## Why this matters for SWIG specifically (packaging / binary distribution)

This is primarily a **binary-packager pain point**, not a language-feature request. PEP 803's own motivation section documents real wheel-count growth that projects are trying to avoid as free-threaded Python becomes mainstream:

- `cryptography` shipped **48 wheel files** in a recent release; without an `abi3t`, supporting free-threaded builds going forward was projected to roughly double PyPI storage needs.
- `pydantic-core` distributes **112 wheels**, a number "set to grow" — without `abi3t`, free-threading support undoes the wheel-count reduction `abi3` already bought them.
- `scipy` needed **60 version-specific wheel files** for one release across four CPython versions.
- Cryptography maintainer Alex Gaynor: *"as long as we have O(1) builds, that's ok. What we can't/won't do is O(n) where we need new builds for every Python release."*

SWIG wraps exactly this kind of large, template/class-heavy C++ surface, where per-version, per-GIL-flavor wheel matrices are expensive to build and maintain in CI. `abi3t` support in SWIG would let such projects collapse "one wheel per (Python minor version × GIL flavor)" down toward "one wheel per platform," the same win `abi3` already gave for minor-version proliferation, now extended across the GIL/free-threaded split too.

<details>
<summary><b>Concrete SWIG-generated example: SimpleITK wheel counts/sizes</b></summary>

[SimpleITK](https://pypi.org/project/SimpleITK/) is a SWIG-wrapped Python binding for the ITK medical/scientific image-analysis toolkit — its generated `SimpleITK.py` proxy module defines 355 Python classes, one per wrapped filter/image/support type. Numbers below are pulled directly from PyPI's JSON API for the current release (`2.5.6`):

- 20 wheel files, **~803 MB** total, made up of three distinct axes of proliferation:
  - 10 plain per-version wheels for Python 3.9/3.10 (pre-dating SimpleITK's `abi3`/Limited-API adoption) — **~401 MB**.
  - 5 `cp311-abi3` wheels, covering *all* of Python 3.11+ on a GIL-enabled build — **~201 MB**.
  - 5 `cp314-cp314t` wheels — version- **and** GIL-flavor-specific, needed just to support one free-threaded minor version (3.14t) — **~201 MB**, i.e. almost exactly duplicating the size of the *entire* `abi3` set, for a single free-threaded version.
- Free-threaded wheels only started appearing in SimpleITK's last two releases (`2.5.5`, `2.5.6`): 10 files, **~0.40 GB** combined so far. Absent `abi3t`, every future free-threaded-only CPython minor version repeats that ~201 MB / 5-wheel slice indefinitely, stacking a second, version-pinned axis on top of the `abi3` set that already exists — undoing exactly the win `abi3` bought. With `abi3t`, that whole free-threaded slice could instead merge into the existing 5 `abi3` wheels via the `abi3.abi3t` compressed tag.
- For scale, across all 41 SimpleITK releases published to date, PyPI hosts **566 wheel files totaling ~20.3 GB**; 507 of those (~18.3 GB) predate `abi3` adoption and are plain per-version wheels — the kind of build-matrix growth a project like this is actively trying to avoid repeating on the GIL/free-threaded axis.

</details>

<details>
<summary><b>Background: the relevant PEPs and what each one contributes</b></summary>

- **[PEP 803](https://peps.python.org/pep-0803/)** ("abi3t": Stable ABI for Free-Threaded Builds, Final, Python 3.15) — defines the `abi3t` ABI variant itself: makes `PyObject`/`PyVarObject` opaque, adds the `Py_TARGET_ABI3T` compile-time macro, the `Py_mod_abi` module slot, and `PyABIInfo_FREETHREADING_AGNOSTIC` flag. States explicitly that `abi3t` binaries are loadable by GIL-enabled builds too (`.abi3t.so` is preferred-fallback-loaded by GIL builds; free-threaded builds only load `.abi3t.so`, not plain `.abi3.so`, as of 3.15).
- **[PEP 697](https://peps.python.org/pep-0697/)** ("Limited C API for Extending Opaque Types," Final, Python 3.12 — already shipped) — the mechanism `abi3t` leans on: since `PyObject` becomes opaque, an extension type's instance struct can no longer embed `PyObject_HEAD` (or any base struct) as its first member. Adds `PyObject_GetTypeData()` / `PyType_GetTypeDataSize()` plus a **negative `PyType_Spec.basicsize`** convention: instead of specifying the object's *total* size, an extension specifies how much *extra* space it needs beyond the (opaque, size-unknown-at-compile-time) base type, and the interpreter computes real layout/offsets at type-creation time.
- **[PEP 793](https://peps.python.org/pep-0793/)** — defines the replacement module-init entry point, `PyModExport_<modulename>()`, because the classic `PyModuleDef`/`PyModuleDef_Init()`/`PyInit_<modulename>()` path depends on `PyModuleDef_Base`, which itself embeds object-header bookkeeping that is undeclarable once `PyObject` is opaque. `PyModExport_<modulename>()` returns a static array of `PySlot` entries (built via `PySlot_FUNC`/`PySlot_STATIC_DATA`/`PySlot_END`), including a now-mandatory `Py_mod_abi` slot.
- **[PEP 703](https://peps.python.org/pep-0703/)** ("Making the Global Interpreter Lock Optional in CPython") — the free-threaded build mode itself. Extensions must not assume the GIL protects their C-level state, and must declare their thread-safety stance via the `Py_mod_gil` module slot (already relevant to SWIG's existing `-nogil` flag from #3121); if unset, free-threaded CPython pauses all threads and re-enables the GIL as a safety net.

</details>

<details>
<summary><b>What generated code would need to change (from the abi3t migration guide + PEPs)</b></summary>

Source: [Migrating to Stable ABI for free threading (abi3t)](https://docs.python.org/3.15/howto/abi3t-migration.html), the official CPython guide for porting extensions — explicitly notes it targets hand-written C/C++ extensions and code generators, and that generator-based tools should wait until *the generator itself* supports `abi3t` (i.e., this is exactly the kind of tool-side work this issue is asking for):

- **`PyObject`/`PyVarObject` become fully opaque.** "Accessing their members is prohibited" — any place SWIG's runtime support code (`pyrun.swg`) does a raw struct-field access on a `PyObject*` needs a getter/setter call instead.
- **Wrapper-object structs can't embed `PyObject_HEAD`.** SWIG's `SwigPyObject`, `SwigPyPacked`, and `swig_varlinkobject` (and, in `-builtin` mode, every per-class generated type) currently do exactly this and are sized via `sizeof(...)` at compile time — both incompatible with `abi3t`.
- **Negative `basicsize` + `PyObject_GetTypeData()`** (PEP 697) is the required replacement pattern: define a struct with *just* the extra per-instance fields, create the type with `basicsize = -sizeof(YourStruct)`, and fetch the instance data pointer via `PyObject_GetTypeData(obj, cls)` instead of a direct cast. Caution called out in the guide: `cls` must be the *defining* class, not `Py_TYPE(obj)`, for any subclassable type — relevant since SWIG's proxy classes support inheritance.
- **Variable-sized types (`Py_tp_itemsize`) cannot be ported to `abi3t` 3.15 at all** — worth checking whether anything in SWIG's Python runtime relies on this.
- **Module init must move to the `PyModExport_<name>` hook** (PEP 793) instead of `PyInit_<name>`/`PyModuleDef`/`PyModuleDef_Init` — a structural change to `Lib/python/pyinit.swg`. Only one `Py_mod_exec` slot is allowed (existing multiple-exec-slot code, if any, would need consolidating), and a module previously handled by `PyModuleDef` needs to switch `PyModule_GetDef()`-style lookups to the new "module token" mechanism (`Py_mod_token` slot + `PyModule_GetToken()`).
- **Multi-phase init + isolation required:** "Your module should use multi-phase initialization, and it should either be isolated or limit itself to be loaded at most once per process."
- **Build-time vs. runtime version macros diverge:** `PY_VERSION_HEX`/`PY_MAJOR_VERSION`/`PY_MINOR_VERSION` no longer reflect the *runtime* interpreter under `abi3t`; use `Py_Version` for runtime checks and `Py_TARGET_ABI3T` for the build-time floor.
- **`Py_GIL_DISABLED` is always defined under `abi3t`,** regardless of whether the loading interpreter actually has the GIL disabled — code should not use it to detect the runtime GIL state, only the build-time API target.
- Guide's own caveat worth flagging to any implementer: "Your extension may become slower, since the Stable ABI prioritizes compatibility over performance," and memory-layout details (e.g. exact offsets) are explicitly *not* guaranteed to stay stable across future `abi3t` implementations — don't hardcode anything observed at one CPython version.

</details>

<details>
<summary><b>Distribution / wheel tagging policy (why "just build for abi3t" isn't quite the recommendation)</b></summary>

Per PEP 803 and the migration guide:

- A binary built for `abi3t` and loaded under a GIL-enabled build behaves as if it declared no thread-safety opt-out for that build's own extensions — but the PEP's compatibility table shows `cp315-abi3t`-only wheels are marked incompatible with *any* GIL-enabled interpreter at the wheel-selection level; only the compressed tag `abi3.abi3t` is universally installable on both flavors (3.15+).
- Recommended practice: extension authors "explicitly compile for both ABIs at once, and signal compatibility using the wheel tag `abi3.abi3t`" — i.e. the target for SWIG should likely be **one build invocation, dual-tagged**, not "an `abi3t` build" as a wholly separate artifact from the existing `abi3` build path SWIG already supports.
- `abi3t`-only wheels (tag `abi3t` without `abi3`) are described as "discouraged, but possible," and are free-threading-only (incompatible with GIL builds of the *same* minor version, per the table in the PEP).

This matters for how the eventual SWIG flag/feature should be scoped: it's not simply "add a `-freethreaded-only` mode" — the goal is a single generated `.so`/wheel that is safe and loadable everywhere `abi3` already works, plus free-threaded builds.

</details>

<details>
<summary><b>Suggested rough scope (not a full design — for discussion)</b></summary>

- A new SWIG flag (e.g. `-abi3t`), parallel to today's `-nogil` (#3121) and existing Limited-API (`Py_LIMITED_API`) support, since `-nogil` alone only makes a runtime thread-safety *declaration* — it does nothing about opaque-`PyObject` compatibility.
- Convert `SwigPyObject` / `SwigPyPacked` / `swig_varlinkobject` in `Lib/python/pyrun.swg` to the PEP 697 opaque-layout pattern (negative `basicsize`, `PyObject_GetTypeData()` accessors) when the flag is set, replacing the current embedded-`PyObject_HEAD` struct + direct-cast + `PyObject_New(SwigPyObject, ...)` allocation pattern.
- A `PyModExport_<name>`-based module-init path (PEP 793) in `Lib/python/pyinit.swg`, alongside (not replacing) the existing `PyInit_<name>`/`PyModuleDef` path used for non-`abi3t` builds.
- `-builtin` mode (per-class generated heap types) would need the same treatment separately, since it multiplies the embedded-header problem across every wrapped class — likely a bigger follow-on piece of work, possibly out of scope for a first pass.
- Test-suite coverage exercising basic pointer wrapping, multiple inheritance (the `SwigPyObject` linked-list-of-`this`-pointers mechanism), and the `cvar`/global-variable path (`swig_varlinkobject`), built and run under an actual Python 3.15 free-threaded interpreter with `-abi3t`, and separately confirmed to still load correctly under a plain GIL-enabled 3.15 interpreter.

Happy to discuss/help prototype pieces of this if useful — filing primarily to get the feature tracked given `abi3t` is now Final for 3.15 and downstream large SWIG-wrapped projects like SimpleITK (see wheel-count/size numbers above) are likely to want this once available, to avoid re-growing the per-version/per-GIL-flavor wheel matrix `abi3` was originally meant to shrink.

</details>

## References

- PEP 803: https://peps.python.org/pep-0803/
- PEP 697: https://peps.python.org/pep-0697/
- PEP 793: https://peps.python.org/pep-0793/
- PEP 703: https://peps.python.org/pep-0703/
- Migration guide: https://docs.python.org/3.15/howto/abi3t-migration.html
- Related prior issue: #3121 (added `-nogil` free-threading declaration support)

blowekamp:
## Summary

Adds a new `-abi3t` command line option to the Python backend, generating wrapper code for Python 3.15's `abi3t` — the Stable ABI variant for free-threaded builds ([PEP 803](https://peps.python.org/pep-0803/)). Fixes/closes #3550.

`abi3t` requires `PyObject`/`PyVarObject` to be fully opaque, so this touches the parts of SWIG's Python runtime support that assume a compile-time-known object layout:

- **`Source/Modules/python.cxx`** — new `-abi3t` flag (implies `-nogil`), plus an incompatibility check against `-builtin` (which isn't `abi3t`-clean, or even plain-`abi3`-clean today, due to `Lib/python/builtin.swg`'s use of the non-limited-API `PyGetSetDescrObject`).
- **`Lib/python/pyinit.swg`** — under `SWIGPYTHON_ABI3T`, module init switches from `PyInit_<name>()`/`PyModuleDef`/`PyModuleDef_Init()` to a `PyModExport_<name>()` entry point (a `PySlot[]` array carrying `Py_mod_abi` via `PyABIInfo_VAR`, `Py_mod_gil`, `Py_mod_methods`) built via `PyModule_FromSlotsAndSpec()`/`PyModule_Exec()`, since `PyModuleDef` is unavailable once `PyObject` is opaque.
- **`Lib/python/pyrun.swg`** — `SwigPyObject`, `SwigPyPacked`, and `swig_varlinkobject` no longer embed `PyObject_HEAD` under `SWIGPYTHON_ABI3T`; their heap types reserve a negative-`basicsize` "extra data" block via `PyType_FromMetaclass` ([PEP 697](https://peps.python.org/pep-0697/)), and every `PyObject*`-to-struct cast goes through `PyObject_GetTypeData()` instead of a direct cast.
- **`Examples/test-suite`** — new `python_abi3t` test case (pointer wrapping, multiple inheritance across the `SwigPyObject` "this"-chain, the `cvar` global-variable mechanism, and pointer-to-member/`SwigPyPacked`), gated on Python >= 3.15.

## Verification

Built and ran against CPython 3.15.0rc1, both free-threaded and regular (GIL-enabled) builds:

- Construction, attribute get/set, multiple inheritance, `cvar` global variables, and pointer-to-member all work correctly.
- The GIL stays disabled under the free-threaded build (confirming the module correctly declares itself thread-safe).
- The existing (non-`abi3t`) Python test suite still passes unchanged, confirming the new code is properly gated behind `SWIGPYTHON_ABI3T` and doesn't affect the default or classic `abi3` code paths.
- Beyond the SWIG test suite itself, I also validated this against a large real-world SWIG-wrapped C++ project (SimpleITK, a medical imaging library with ~350 generated wrapper classes): configured, built, and packaged an actual `abi3t`-tagged wheel with this SWIG build, then verified it end-to-end under a free-threaded 3.15rc1 interpreter — image construction, numpy buffer-protocol roundtrip, and running a real ITK filter, all with the GIL staying disabled.

## Known limitations / follow-up work (not attempted here)

- **`-builtin` mode** (per-class generated heap types in `python.cxx` + `builtin.swg`'s `SwigPyStaticVar`/`SwigPyObjectType` metaclass) needs the same opaque-layout treatment, multiplied across every wrapped class — explicitly rejected as incompatible with `-abi3t` for now via the new incompatibility check.
- **Directors** combined with `-abi3t` are untested and not yet supported.
- Only the highest-value path (default proxy-mode pointer wrapping) is covered; this is a working prototype for that path, not a complete port of every SWIG Python feature to `abi3t`.

## References

- Fixes #3550
- [PEP 803](https://peps.python.org/pep-0803/) — abi3t
- [PEP 697](https://peps.python.org/pep-0697/) — opaque per-instance data
- [PEP 793](https://peps.python.org/pep-0793/) — module export hooks
- [PEP 703](https://peps.python.org/pep-0703/) — free-threading
- [abi3t migration guide](https://docs.python.org/3.15/howto/abi3t-migration.html)


jschueller:
dont we want to enable this via macro definition (after code is generated by swig) like we did for abi3 ?

espressolee:
I took a closer look at exact draft head `6016f57649244454b9f23423700d7fe95fa0d0e2`. The core abi3t port is substantial and the default proxy path works in my independent checks: C and C++ generated modules build and run under CPython 3.15.0b4t and GIL-enabled 3.15, and strict `abi3audit` reports zero violations.

I think one contract boundary should change before this leaves draft: `-abi3t` should not automatically imply `-nogil` / `Py_MOD_GIL_NOT_USED`.

PEP 803 explicitly separates these claims: it makes extensions loadable on free-threaded Python, but "not necessarily ones that are thread-safe without a GIL". SWIG cannot infer the synchronization properties of the wrapped C/C++ library from the selected ABI.

I made that gap observable with an otherwise ordinary module generated by this exact head using `swig -python -c++ -abi3t`. It wrapped an unsynchronized native `long` and an `add_many()` function. With 24 Python threads and 1,000,000 increments per thread:

- GIL-off: 21/100 trials lost one or two million increments;
- the same binary with the GIL enabled: 0/20 mismatches.

This is not a claim that the SWIG runtime itself caused the native race. It shows that `-abi3t` currently issues a GIL-free safety certificate for arbitrary wrapped code without an explicit author opt-in.

I also found two option-surface gaps:

- `-abi3t -O` / fastproxy is accepted but the generated wrapper fails with 16 opaque-API compile errors (`PyCFunctionObject`, `PyInstanceMethod_New`, `PyStaticMethod_New`, etc.);
- directors are documented here as unsupported, but `-abi3t -directors` and `%module(directors="1")` are currently accepted and still emit the GIL-free marker.

I tested a local candidate with the following narrower contract:

- `-abi3t` selects only the abi3t module export and opaque object layout;
- explicit `-nogil` is the independent thread-safety promise;
- abi3t without `-nogil` emits `Py_MOD_GIL_USED`;
- builtin, fastproxy/`-O`, and directors are rejected with explicit diagnostics.

On macOS arm64 that candidate gives the intended matrix on a default 3.15t startup:

- `-abi3t -nogil`: GIL `False -> False`;
- `-abi3t`: GIL `False -> True`.

Both C and C++ focused paths pass strict warning flags and `abi3audit`; all 158 error tests and 833 Python tests pass. In the native-counter control, abi3t-only enabled the GIL and had 0/40 mismatches, while explicit `-nogil` stayed GIL-off and exposed the expected unsafe-library race.

This is macOS-only evidence so far; Windows and Linux checks are still separate open gates. If this contract direction matches what you want for the option, I have the small patch and regression fixtures ready and would be happy to share or adapt them rather than opening a competing PR against the draft.

blowekamp:
Thanks for the detailed writeup and the race repro. I read through PEP 803 with my human, we discussed the `Py_mod_gil` separation specifically, and also looked at the wheel packaging tag side (`abi3t` vs. the combined `abi3.abi3t` tag) to make sure the ABI-loadability claim and the thread-safety claim stay independent end to end. We agree with your core point: `-abi3t` and `-nogil` should be decoupled.

Pushed an update (head `704127be7`) that does that:

- `-abi3t` alone now only selects the ABI (module export shape, opaque object layout) and no longer implies `-nogil`. It leaves the GIL enabled by default, matching CPython's own "has not declared that it can run safely without the GIL" behavior.
- `-abi3t -nogil` is the explicit, independent opt-in that declares the module GIL-free.
- `-abi3t` now rejects `-fastproxy`/`-O` and directors on the command line (previously only documented as unsupported, not enforced).
- Added test coverage for both cases (`-abi3t` alone leaves the GIL enabled; `-abi3t -nogil` leaves it disabled), verified against real CPython 3.15.0rc1 on both free-threaded and regular builds. Full Python test suite passes with no regressions.

Thanks again for the thorough review — happy to take a look at your fixtures if you still want to share them for cross-checking.

espressolee:
I rechecked exact head `704127be7291e45f98444d28d824731a3038e9bc` against source-built CPython 3.15.0rc1 regular and free-threaded runtimes.

The contract changes you made work as intended in my checks:

- `-abi3t`: FT GIL `False -> True`;
- `-abi3t -nogil`: FT GIL `False -> False`;
- `-builtin`, `-fastproxy`, both `-O` orderings, command-line directors, and `%module(directors="1")` all fail closed;
- the unsynchronized counter control was `0/20` mismatches with ABI3T-only/GIL-enabled and `20/20` with explicit `-nogil`;
- separately, the Python suite configured against 3.15.0b4 regular was
  `833/833` passing. I did not run the full suite against my source-built rc1.

I found two blocking gaps in the new ABI3T path.

1. **Generated C++ currently requires C++20.** `pyinit.swg` emits `PySlot_FUNC`, `PySlot_DATA`, and `PySlot_STATIC_DATA`. With rc1 headers, exact generated code fails under both strict C++11 and C++17 with three `C++20-designator` errors; C++20 passes. CPython documents `PySlot_PTR` / `PySlot_PTR_STATIC` for pre-C++20 C++. Replacing only those generated macros made strict C++11 compile and preserved regular/FT loading.

2. **A repeated `swiginit` loses the linked `this` pointer.** Under ABI3T, `SWIG_Python_GetSwigThis()` returns extra data, but `pyrun.swg` passes it back as `(PyObject *)sthis` to `SwigPyObject_append()`. Forcing two generated `*_swiginit` calls on one holder returns success but leaves `holder.this.next()` as `None` in `20/20` regular-rc1 and `20/20` FT-rc1 processes. The same source without `-abi3t` links the second pointer in `20/20`. A generated-only candidate that routes the already-resolved data pointer to a data helper changed the ABI3T result to `0/20` link losses.

There are also two contract/test points worth resolving before this leaves draft.

- Two ABI3T modules create distinct `SwigPyObject` type objects. The receiving module accepts the foreign object by fully-qualified name, then calls `PyObject_GetTypeData(obj, its_local_SwigPyObject_Type())`, although the object is not an instance of that class. It works on rc1, but violates the documented precondition and is not a Stable-ABI guarantee. A same-name spoof crash also exists in non-ABI3T SWIG, so I am not attributing that pre-existing issue to this PR.
- The new Makefile tests define only `Py_LIMITED_API`. A binary built that way on regular rc1 passes there but is rejected by FT rc1 as incompatible. The same generated source built with `Py_TARGET_ABI3T` loads on both, so the committed test should exercise the actual target.

I have the small interfaces, fresh-process runners, JSON results, and generated-wrapper candidate reductions ready to share. I kept the subinterpreter check separate; rc1 rejected the module with an explicit “does not support loading in subinterpreters” error, so that arm failed closed here.

jschueller:
lets add an abi3t configuration entry in the CI jobs to actually test it

espressolee:
For that entry: the two new tests compile with `-DPy_LIMITED_API=0x030f0000` and not
`Py_TARGET_ABI3T`, so a job that runs them inherits that. In my 3.15.0rc1 checks a binary
built with `Py_LIMITED_API` alone loaded on the regular build and was rejected by the
free-threaded one as incompatible, while the same generated source built with
`Py_TARGET_ABI3T` loaded on both. Building against the abi3t target and loading the one
binary on both interpreters is what would exercise the claim `-abi3t` is making.

`linux.yml` has no `3.15t` entry today, and `-abi3t` is 3.15-minimum per `CHANGES.current`,
so the free-threaded side needs one added.

Issue #3537, opened 2026-08-15
Include pycontainer.swg in python.swg


rlaboiss:
The [xdmf package in Debian](https://tracker.debian.org/pkg/xdmf) fails to build against SWIG version 4.5.0 (See [Bug#1144265](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1144265).  In oder to fix this problem, we [patched](https://salsa.debian.org/debian/swig/-/blob/debian/master/debian/patches/include-pycontainer.patch?ref_type=heads) the Debian package for SWIG to include  the `pycontainer.swg` in `python.swg` (conditionnally for the C++ wrapper).

We would like to know whether the SWIG authors think this is an appropriate solution, or if you can suggest a better one.

wsfulton:
That is a terrible idea and should definitely not be accepted. It is akin to a `#include` of every STL container in C++ program by default - just wrong on so many fronts. It looks like it is a patch already in debian master, I'd raise as an **urgent** request to revert it before it spreads.

I'm not sure I even understand the motivation for this if it is to avoid fixing a small number backwards compatible Python 2 macros that need to be removed (the [Bug#1144265](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1144265) link you pasted. Prior to release all the affected packages in Fedora were identified and patched, see the swig-devel mailing list, but this one should be useful: https://sourceforge.net/p/swig/mailman/message/59365186/, but I don't see xdmf in that list, but maybe Fedora don't build that.

BTW, If you know of an equivalent Debian setup for someone to build all the Debian packages prior to release, they should get in touch.

rlaboiss:
Thank you for your feedback, @wsfulton.

> That is a terrible idea and should definitely not be accepted. It is akin to a `#include` of every STL container in C++ program by default - just wrong on so many fronts. It looks like it is a patch already in debian master, I'd raise as an **urgent** request to revert it before it spreads.

Fair enough, I reverted the change by removing the patch, and released version 4.5.0-3 of the swig package, while we wait for a better solution.

rlaboiss:
[...]

PR #3554, opened 2026-08-31
Fix #3553 - Fix cross-module std::vector<T> mangled-name mismatch

### LINKED ISSUE #3553: std::vector<T> parameter/return type breaks across %import'd modules when T is only forward-declared in one of them
## Summary

When a class `T` is only forward-declared in module A (which defines a method taking/returning `std::vector<T>`), and `T`'s real definition plus `%template(...)  std::vector<T>;` live in module B which `%import`s module A, the two modules generate **different mangled type names** for the same C++ type `std::vector<T>`. Since SWIG's cross-module runtime type registry merges types by exact mangled-name match, the merge silently fails: calling the method from module A with a `T`-vector object created in module B raises a `TypeError` (or, for return values, produces an object that has lost its wrapped class, i.e. an untyped `SwigValueObject`/capsule instead of the proper proxy).

This works correctly through SWIG 4.1.1 and breaks starting with 4.2.0. It is still broken on 4.4.1 and current `master` (`5872e9b4`, reporting as SWIG 4.6.0).

Found while investigating https://github.com/NatLabRockies/OpenStudio/issues/5645, where OpenStudio's SWIG-based Python/Ruby bindings hit exactly this pattern for `ShadowCalculation::addShadingZoneGroup(std::vector<ThermalZone>&)` and `ShadingControl::addSubSurfaces(std::vector<SubSurface>&)`.

## Root cause

Bisected to commit https://github.com/swig/swig/commit/e3ccabbd4dc22a7e59a9d1b93308d96a97d35328 - "Simpler names when using SwigType_manglestr for templates" (2022-11-12, `Source/Swig/stype.c`), first released in v4.2.0.

That commit changes template mangled-name generation to drop default template arguments (e.g. `std::allocator<T>`) from the mangled name **when the full definition of the template argument is visible**, but the argument is *kept* when the template argument type is only forward-declared. Since whether `T` is "fully visible" differs between module A (forward-declared) and module B (fully defined), the *same* C++ type `std::vector<T>` ends up mangled as two different strings depending on which module wraps it — e.g. `std__vectorT_Bar_std__allocatorT_Bar_t_t` in the forward-declaration-only module vs the shortened `std__vectorT_Bar_t` in the module with the complete type.

## Minimal reproduction (Ruby, but almost certainly language-agnostic — the bug is in `Source/Swig/stype.c`, not a language module)

`common.hpp`:
```cpp
#pragma once
#include <vector>
#include <string>

class Bar {
public:
  Bar() : name_("bar") {}
  explicit Bar(std::string name) : name_(std::move(name)) {}
  std::string name() const { return name_; }
private:
  std::string name_;
};

class Container {
public:
  Container() = default;
  bool addBars(const std::vector<Bar>& bars) {
    count_ += bars.size();
    return true;
  }
  std::vector<Bar> getBars() const {
    return std::vector<Bar>(count_);
  }
private:
  size_t count_ = 0;
};
```

`container.i` — module A: `Bar` only forward-declared here:
```
%module container
%{
#include "common.hpp"
%}
%include <std_vector.i>
class Bar;
class Container {
public:
  Container();
  bool addBars(const std::vector<Bar>& bars);
  std::vector<Bar> getBars() const;
};
```

`elem.i` — module B: `%import`s module A, fully defines `Bar`, and instantiates the vector template:
```
%module elem
%import "container.i"
%{
#include "common.hpp"
%}
%include <std_vector.i>
class Bar {
public:
  Bar();
  explicit Bar(std::string name);
  std::string name() const;
};
%template(BarVector) std::vector<Bar>;
```

Build both as separate Ruby extensions and run:
```ruby
require_relative 'container'
require_relative 'elem'

bv = Elem::BarVector.new
bv.push(Elem::Bar.new)

c = Container::Container.new
c.addBars(bv)   # <-- TypeError here on SWIG >= 4.2.0
```

## Actual output (SWIG 4.6.0 / current master, `5872e9b4`)

```
repro_runme.rb:9:in 'addBars': Expected argument 1 of type std::vector< Bar,std::allocator< Bar > > const &, but got Elem::BarVector std::vector<Bar,std::allocator... (TypeError)
	in SWIG method 'addBars'
```

## Expected output

```
PASS
```
(as produced on SWIG 4.1.1, and on 4.1.1 through 4.4.0-adjacent commits before `e3ccabb`)

## Versions tested

| Version / commit | Result |
|---|---|
| v4.1.1 | PASS |
| v4.2.0 | FAIL |
| v4.2.1, v4.3.0, v4.3.1, v4.4.0, v4.4.1 | FAIL |
| master @ `5872e9b4` (reports 4.6.0) | FAIL |

Build: `g++ -std=c++17`, Ruby 3.2.2, Linux x86_64.

## Suggested direction

The mangled name for a given C++ type should not depend on how much of that type's definition happens to be visible at the point SWIG mangles it in a particular translation unit — that's exactly what makes cross-module type merging unreliable for any type whose full definition lives in a different (but `%import`-related) module than where it's first used as a template argument. Either the "simpler names" optimization from `e3ccabb` needs to be suppressed for template arguments that are only forward-declared (so it falls back to the pre-`e3ccabb`, always-fully-qualified mangling in that case), or the optimization needs to be applied consistently regardless of type completeness.

jmarrec:
- Fix #3553

std::vector<T> cannot be used as a parameter or return type across an %import boundary when T is only forward-declared in the importing module while T's real definition and %template(...) live in the imported module. The two modules mangle std::vector<T> differently (the default std::allocator<T> argument is dropped only when T is fully visible, since e3ccabbd), so the cross-module runtime type registry never merges them, and passing the vector across the module boundary fails.

Added as import_vector_fwd_{a,b}.i (with a shared header) and wired into ruby and python as a known-failing multi-module test, since this is not yet fixed. Confirmed still broken on 4.4.1 and master.

### DIFF
--- a/Examples/test-suite/import_vector_fwd.h
+++ b/Examples/test-suite/import_vector_fwd.h
@@ -1,0 +1,37 @@
+/* Shared by import_vector_fwd_a.i and import_vector_fwd_b.i.
+
+   Bar is only forward-declared in import_vector_fwd_a.i (which declares Container, using
+   std::vector<Bar> as a parameter/return type). Bar is fully defined, and
+   std::vector<Bar> is %template'd, in import_vector_fwd_b.i, which %imports
+   import_vector_fwd_a.i. */
+
+#ifndef SWIGTESTIMPORTVECTORFWD_H
+#define SWIGTESTIMPORTVECTORFWD_H
+
+#include <vector>
+#include <string>
+
+class Bar {
+public:
+  Bar() : name_("bar") {}
+  Bar(std::string name) : name_(name) {}
+  std::string name() const { return name_; }
+private:
+  std::string name_;
+};
+
+class Container {
+public:
+  Container() : count_(0) {}
+  bool addBars(const std::vector<Bar>& bars) {
+    count_ += bars.size();
+    return true;
+  }
+  std::vector<Bar> getBars() const {
+    return std::vector<Bar>(count_);
+  }
+private:
+  size_t count_;
+};
+
+#endif
--- a/Examples/test-suite/import_vector_fwd.h
+++ b/Examples/test-suite/import_vector_fwd.h
@@ -10,6 +10,8 @@
 
 #include <vector>
 #include <string>
+/* Some std_vector.i variants only #include this in a %{ %} block, a no-op for a module reached via %import. */
+#include <stdexcept>
 
 class Bar {
 public:
--- a/Examples/test-suite/import_vector_fwd.h
+++ b/Examples/test-suite/import_vector_fwd.h
@@ -10,7 +10,8 @@
 
 #include <vector>
 #include <string>
-/* Some std_vector.i variants only #include this in a %{ %} block, a no-op for a module reached via %import. */
+/* Some std_vector.i variants only #include these in a %{ %} block, a no-op for a module reached via %import. */
+#include <algorithm>
 #include <stdexcept>
 
 class Bar {
--- a/Source/Swig/stype.c
+++ b/Source/Swig/stype.c
@@ -1224,8 +1224,22 @@ static String *manglestr_default(const SwigType *s) {
 
   if (SwigType_istemplate(ss)) {
     SwigType *dt = Swig_symbol_template_deftype(ss, 0);
-    String *ty = Swig_symbol_type_qualify(dt, 0);
+    SwigType *qt = Swig_symbol_type_qualify(dt, 0);
+    /* Swig_symbol_type_qualify() can collapse a template-id back down to a shorter spelling
+       (omitting default template arguments) when a %template-instantiated node happens to be
+       registered for it locally -- but whether that's the case depends on unrelated %template
+       calls elsewhere in this compilation, not on the type itself. Re-expand any defaults here
+       so the mangled name stays a stable function of the type, regardless of what's been
+       %template'd locally (see swig/swig#3553). */
+    String *ty = SwigType_istemplate(qt) ? Swig_symbol_template_deftype(qt, 0) : Copy(qt);
+    if (getenv("SWIG_TRACE_MANGLE") && strstr(Char(ss), "Bar")) {
+      Printf(stderr, "[manglestr_default] ss=%s\n", ss);
+      Printf(stderr, "[manglestr_default] dt=%s\n", dt);
+      Printf(stderr, "[manglestr_default] qt=%s\n", qt);
+      Printf(stderr, "[manglestr_default] ty=%s\n", ty);
+    }
     Delete(dt);
+    Delete(qt);
     Delete(ss);
     ss = ty;
     type = ss;
--- a/Source/Swig/stype.c
+++ b/Source/Swig/stype.c
@@ -1232,12 +1232,6 @@ static String *manglestr_default(const SwigType *s) {
        so the mangled name stays a stable function of the type, regardless of what's been
        %template'd locally (see swig/swig#3553). */
     String *ty = SwigType_istemplate(qt) ? Swig_symbol_template_deftype(qt, 0) : Copy(qt);
-    if (getenv("SWIG_TRACE_MANGLE") && strstr(Char(ss), "Bar")) {
-      Printf(stderr, "[manglestr_default] ss=%s\n", ss);
-      Printf(stderr, "[manglestr_default] dt=%s\n", dt);
-      Printf(stderr, "[manglestr_default] qt=%s\n", qt);
-      Printf(stderr, "[manglestr_default] ty=%s\n", ty);
-    }
     Delete(dt);
     Delete(qt);
     Delete(ss);
--- a/Examples/test-suite/common.mk
+++ b/Examples/test-suite/common.mk
@@ -958,6 +958,7 @@ MULTI_CPP_TEST_CASES += \
 	clientdata_prop \
 	import_share \
 	import_stl \
+	import_vector_fwd \
 	imports \
 	mod \
 	multi_import \
--- a/Examples/test-suite/go/Makefile.in
+++ b/Examples/test-suite/go/Makefile.in
@@ -189,6 +189,8 @@ clean:
 	rm -f import_stl_a.go import_stl_a.gox
 	rm -f import_stl_b.go import_stl_b.gox
 	rm -f import_stl_c.go import_stl_c.gox
+	rm -f import_vector_fwd_a.go import_vector_fwd_a.gox
+	rm -f import_vector_fwd_b.go import_vector_fwd_b.gox
 	rm -f imports_a.go imports_a.gox imports_b.go imports_b.gox
 	rm -f mod_a.go mod_a.gox mod_b.go mod_b.gox
 	rm -f multi_import_a.go multi_import_a.gox
--- a/Examples/test-suite/go/Makefile.in
+++ b/Examples/test-suite/go/Makefile.in
@@ -34,6 +34,14 @@ CPP_TEST_CASES = \
 MULTI_CPP_TEST_CASES = \
 	go_subdir_import \
 
+# import_vector_fwd_b.i only %imports import_vector_fwd_a.i for the sake of the
+# C++ runtime type registry (see import_vector_fwd_a.i); it never references any
+# of module a's wrapped Go types. SWIG's Go backend always emits a Go `import`
+# statement for every %import though, so the generated import_vector_fwd_b.go
+# ends up with an import that Go's compiler rejects as unused.
+FAILING_MULTI_CPP_TESTS += \
+	import_vector_fwd \
+
 include $(srcdir)/../common.mk
 
 INCLUDES = -I$(abs_top_srcdir)/$(EXAMPLES)/$(TEST_SUITE)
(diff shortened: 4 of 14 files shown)


Source/Swig/stype.c
```diff
@@ -1224,8 +1224,16 @@ static String *manglestr_default(const SwigType *s) {
...
     SwigType *dt = Swig_symbol_template_deftype(ss, 0);
-    String *ty = Swig_symbol_type_qualify(dt, 0);
+    SwigType *qt = Swig_symbol_type_qualify(dt, 0);
+    /* Swig_symbol_type_qualify() can collapse a template-id back down to a shorter spelling
+       (omitting default template arguments) when a %template-instantiated node happens to be
+       registered for it locally -- but whether that's the case depends on unrelated %template
+       calls elsewhere in this compilation, not on the type itself. Re-expand any defaults here
+       so the mangled name stays a stable function of the type, regardless of what's been
+       %template'd locally (see swig/swig#3553). */
+    String *ty = SwigType_istemplate(qt) ? Swig_symbol_template_deftype(qt, 0) : Copy(qt);
```
jmarrec, line 1234: Here's what the real trace data shows, and why `qt` + `ty` (re-expand) beats reverting to plain `dt`:

## What `ss → dt → qt → ty` actually do, with real values

**Module a** (`import_vector_fwd_a.i`, `Bar` only forward-declared): for `std::vector<Bar>` used as `Container::addBars`'s parameter type, all four stages agree:
```
ss = r.q(const).std::vector<(Bar,std::allocator<(Bar)>)>
dt = (same)
qt = (same)
ty = (same)
```
No `%template` for `std::vector<Bar>` is visible in this module, so `Swig_symbol_type_qualify` finds nothing to collapse — `qt` never diverges from `dt` here.

**Module b** (`import_vector_fwd_b.i`, `Bar` fully defined + `%template(BarVector) std::vector<Bar>;`): for the same type,
```
dt = std::vector<(Bar,std::allocator<(Bar)>)>
qt = std::vector<(Bar)>          <- collapsed!
ty = std::vector<(Bar,std::allocator<(Bar)>)>   <- re-expanded, matches module a again
```
`Swig_symbol_type_qualify(dt, 0)` looks up `dt` in the local symbol table (via type-equivalence, not exact string match) and — because `%template(BarVector) std::vector<Bar>;` registered a node for it — swaps in that node's literal, user-typed spelling (1 arg, allocator omitted). That's e3ccabb's whole point ("simpler names ... default template name is generated instead of one with the default template parameter"). But it means the *same type* mangles differently in module a vs. module b, purely because of an unrelated local `%template` call — hence #3553's cross-module merge failure.

## Confirmed with the full test suite, not just this test

I broadened the trace filter (temporarily, not committed) to log every `dt != qt` divergence across the entire Python test suite (~4000 tests, all passing). Result: **34 distinct pairs, every single one is this same default-template-argument collapse** (`std::allocator<T>` or similar dropped) — never a genuine namespace/typedef requalification. That makes sense: real namespace/typedef resolution for the top-level type already happened earlier in the pipeline (`SwigType_typedef_qualified(sr)`, before `manglestr_default`'s template branch even runs), so by the time `qt = Swig_symbol_type_qualify(dt, 0)` runs, its only observable job left is this instantiation-name shortcut.

## Why not just revert to `dt` (pre-e3ccabb)

That would fix #3553 trivially — but it throws away e3ccabb's actual feature (shorter mangled/generated names matching what the user wrote in `%template`, "used in various places such as the type system"), and I haven't verified nothing downstream depends on that shorter spelling in a broader sense than mangling.

The `ty` re-expansion keeps `qt`'s collapse in the (harmless, single-module) common case but only special-cases it back out: `SwigType_istemplate(qt) ? Swig_symbol_template_deftype(qt, 0) : Copy(qt)`. Since `Swig_symbol_template_deftype` only ever consults the *template's own declared defaults* (never a specific instantiation), it's a pure function of the type — idempotent, and independent of what's been `%template`'d anywhere. That's why `ty` in module b lands back on the exact same 2-arg canonical string as module a's `ss`/`dt`/`qt`, closing the cross-module mismatch, while a qualifier-collapse buried inside a *nested* type (e.g. the `NodeIT::owners` alias-qualifier case handled elsewhere in `Swig_symbol_type_qualify`, `symbol.c:1811`) is untouched since `qt` there isn't itself a bare template-id.

Net: full python suite (~4000 tests) + your earlier ~55 template/typedef tests and all multi-module tests pass with this fix, and the regression test (`import_vector_fwd.multicpptest`) now passes.

Examples/test-suite/import_vector_fwd.h
```diff
@@ -0,0 +1,39 @@
...
+   std::vector<Bar> is %template'd, in import_vector_fwd_b.i, which %imports
+   import_vector_fwd_a.i. */
+
+#ifndef SWIGTESTIMPORTVECTORFWD_H
+#define SWIGTESTIMPORTVECTORFWD_H
+
+#include <vector>
+#include <string>
+/* Some std_vector.i variants only #include this in a %{ %} block, a no-op for a module reached via %import. */
+#include <stdexcept>
```
jmarrec: workaround: added #include <stdexcept> to Examples/test-suite/import_vector_fwd.h (shared by both test modules), with a comment noting why — it's a workaround for the underlying Lib/c/std_vector.i bug where a raw %{ #include <stdexcept> %} block is a no-op for a module reached via %import. c, python, and ruby all pass now

I think std_vector.i should use a fragment instead. This is similar to this recently fixed issue:https://github.com/swig/swig/issues/3537

https://github.com/swig/swig/blob/5872e9b4176cf59ce4e31b010c29a4cebf32960f/Lib/c/std_vector.i#L9-L12

wsfulton:
Hi @jmarrec, each time you push you start a massive compute job on Github. The idea is you develop and test locally (or on your own Github account) and not on the main SWIG Github CI, then push when you have everything ready. As you are changing fundamentals in the core, please run run the full test-suite locally. Just to let you know that you're changing an extremely sensitive area of the code so please check what you are doing carefully. Also to remind you that any use of AI must be disclosed as per the agentic files in the repos that should be used.

jmarrec:
@wsfulton I can understand (some) your point of view from the maintainer hat. I can't interpret your tone from this writing, so I don't know if you're just casually laying facts, or if there's some proper frustration I'm creating (apologies if so)

But I fear you may pushing a slightly unrealistic goal here from irregular contributors such as myself, at least in the current state of documentation. 

There's no documentation saying it's forbidden/frowned upon to open a PR before you've done ALL testing locally (or on your fork) **[1]**. But noted nonetheless, doing on a fork is fine.
I **did** test everything myself locally using Ruby and Python - which passed. I don't use C#/php unless I'm forced to (don't have it on all my machines), never used go, and TBH I don't even know what lua is.


As far as my use of AI, I didn't really look at hidden folder `.agents`(and symlink `.claude`) folders and there's no mention of it **[2]**

There is nothing in this PR that I left an AI do autonomously. I did ask Code Sonnet 5 to analyze and summarize the trace calls I put around the point of insert in stype.c. And did ask it to find out why lua and go tests were still failing.

I can manually put `Assisted-by:` into these commits if you want but:

1) That'll re-trigger the CI matrix
2) My previous contributions have always been squashed-merged with a commit rewrite anyways


I'm also happy to just close this PR and let you deal with it (or not) from scratch.


----

Note: [1] and [2] would probably benefit from being added to a CONTRIBUTING.md at root, along with a link to the wiki or something

FYI, your wiki has a dead link https://github.com/swig/swig/wiki, the right one is https://github.com/swig/swig/wiki/SWIG-4.0-Development

image

wsfulton:
Hi @jmarrec. Sorry if I've come across as frustrated, I was hoping to simply communicate some development points, which I think you have quite rightly highlighted as being deficient in our documentation. That's useful feedback, I'll make some doc improvements.

I was assuming standard agentic workflows where AI agents followed the AI standards, which is for them to read the AGENTS.md file. In theory an agent should do this and then pick up all the skills referenced therein which are in the .agents directory. These agentic files spell out a lot of our development process so that agents can follow them. Without details of your agent harness, models and how you're using them, I'm not entirely sure how come this was missed, but possibly is because you are not fully agentic coding, but instead using agents as a light touch guide. I know that copilot, for example, didn't until recently read AGENTS.md. I've also seen long sessions where the agent forgets some basics. I suppose humans also need to read the agentic files when developing to check the agents are following the guidelines are being followed as most of the useful information is outlined in these files. Traditionally, this information is not well documented and would be deduced from existing code and commits. Also I would say these guidelines are also in early days/incomplete and will need constant refinement.

Regarding your two specific points.
1. Your agent really should have put the "Assisted-by" in the commit message, if the agent does the commit. Failing that during the review and squash, I have no problem adding this in myself and will do so on disclosure of AI assistance. So unless you want to make further changes, no need to just add the Assisted-by (there is also ci-skip as an option).
2. I generally decide whether to squash merge or not depending on the nature of the pull request commits, but if you have any strong preference, I can follow that.

Please keep the pull request open, contributions are most welcome. I'll be spending a concerted effort to work my way through them all in the coming weeks prior to the 4.6 release.

jmarrec:
I'm taking all your remarks in consideration for future use. Thank you for the explanations.

> Your agent really should have put the "Assisted-by" in the commit message, if the agent does the commit

I don't let agents ever commit, on purpose. I refuse to stamp something I don't control/understand. I'm still very wary of LLMs, even though I do admit they facilitate some things (with the tradeoff of perhaps increasing your cognitive load by switching your focus from tedious/repetitive implementation taks to lots of review)

PR #3555, opened 2026-08-31
CMake: Lower pcre requirement

jschueller:
could it be considered for 4.5.1 release ? this prevents compiling on almalinux 8


wsfulton:
I've removed the arbitrary minimum version altogether as mentioned in 336519f05adb82d7d3e7bb3de11e83db160f2f7b and will include it in 4.5.1.

jschueller:
thanks!

PR #3557, opened 2026-09-05
Fix make check failure with jsc

xry111:
Examples/Makefile.in checks ENGINE but it seems unassigned at all (if not explicitly overriden).  Check JSENGINE instead, which gets the value set by configure if ENGINE is not set.


erezgeva:
> ENGINE

The patch seems OK.

The explanation is a bit wrong.
We do set `ENGINE`.
We simply do it outside of this make file.

You are correct that the selection of rules should be done based on actual used engine, i.e.  `JSENGINE`.
 `ENGINE` should only be used to set `JSENGINE`!
As done on line 750.

You forget the `javascript_version` rule.
It should use  `JSENGINE`, as well.

xry111:
> > ENGINE
> 
> The patch seems OK.
> 
> The explanation is a bit wrong. We do set `ENGINE`. We simply do it outside of this make file.
> 
> You are correct that the selection of rules should be done based on actual used engine, i.e. `JSENGINE`. `ENGINE` should only be used to set `JSENGINE`! As done on line 750.
> 
> You forget the `javascript_version` rule. It should use `JSENGINE`, as well.

Addressed.

PR #3574, opened 2026-09-20
Add opt-in typed overload declarations to Python stubs

jgillis:
(this is part of a GPT-6 Astra run to port features used in CasADi's swig fork back upstream)

Python stubs currently collapse overloaded functions into catchall signatures, losing argument/return relationships. Add opt-in declarations using existing `pytyping` maps:

```swig
%include <std_string.i>
%feature("python:annotations", "typing");
%feature("python:stub:overloads") choose;
int choose(int);
std::string choose(std::string);
```

With `-pyi`, this emits separate `@typing.overload` declarations returning `int` and `str`. Runtime dispatch order is retained by default, including default-argument copies. An optional per-signature `rank="-1"` places a declaration earlier when typing requires a different order, such as `str` before `Sequence[str]`. Equal ranks remain stable; invalid ranks are diagnosed. Identical parameter signatures are combined, with union returns when necessary.

Local validation: runtime and AST tests in proxy, builtin and fastproxy modes; strict positive/negative mypy and Pyright consumers; Python 3.7 grammar and existing stub regressions. Coverage includes director constructors, named `%extend` typemaps, typemap defaults, grouped inputs, argout returns, static methods and rank diagnostics. Ranked/unranked runtime `.py` files are byte-identical.

All 202 fork CI jobs passed on 4483beb73 across Linux, macOS, Windows and CMake.


Nerixyz:
Duplicate of #3529?

